PR #2105, opened 2013-10-09
PHPDoc compatible with ST3

yvmarques:


chabinot:
Hello,
I've installed PhpDoc on my Sublime Text 3 (bulid 3156). It seems that it does not work. When i tape               /** then tab, the result is
/**
 *
*/
and that's all, no generated comments on my function.
Please, tell me what to do ?
Regards

FichteFoll:
You should report this on the package's repository.

PR #3194, opened 2014-05-07
TypeTodo made both st2/st3 compatible

NikolayRag:


THEr2d2:
Hello, I have a question and a huge request regarding to typeToDo for siblime.
Are you planning to update to version 3/4 for current builds?
I think the plugin is brilliant. At least when it comes to the idea because I can't make it work.
I've already gone through all the configuration methods, etc., and either I'm missing something or it doesn't want to work for me ;/
I even created an sql file with the database configuration because I thought there was something wrong with creating tables by the plugin, but it didn't help either. Moreover, I noticed that the plugin overwrites the configuration with an empty o.o sometimes without reason (slight surprise).
Maybe a video option on how to configure the plugin with each of the options. I don't consider myself a genius, but I'm not stupid and I failed terribly with this.
Unfortunately, I haven't touched PY and I don't know much about it....
Thanks in advance for your answer :)

[typetodo.sql.txt](https://github.com/wbond/package_control_channel/files/12873688/typetodo.sql.txt) (remove .txt)

NikolayRag:
Sorry for long answer.
I'd collected pretty much small changes, fixes and redesign motives 
So it's absolutely definitely will be upgraded to actual state, without changing the idea.
I believe not in very long time - it's on the schedule, but but will be done asap.

Thanx for feedback!

PR #3330, opened 2014-06-12
Extended Tab Switcher plugin for sublime

rajeshvaya:
**This plugin is inspired by the default tab switcher in netbeans**  which displays the open files to choose from on pressing `ctrl+alt+tab`.

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -695,6 +695,17 @@
 			]
 		},
 		{
+			"name": "ExtendedTabSwitcher",
+			"author": "Rajesh Vaya",
+			"details": "https://github.com/rajeshvaya/Sublime-Extended-Tab-Switcher",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"details": "https://github.com/rajeshvaya/Sublime-Extended-Tab-Switcher/tags"
+				}
+			]
+		},
+		{
 			"name": "ExtendScript",
 			"details": "https://github.com/seblavoie/sublime-extendscript",
 			"releases": [


greneholt:
I must say I was rather annoyed to find that my installation of GotoOpenFile had been silently replaced with Extended Tab Switcher, which has different default key bindings and doesn't even seem to work (I just get IndexError's). Please implement some sort of mechanism to alert users before replacing one of their packages with something else.

FichteFoll:
There is the functionality to provide update messages which could have been displayed, since this can actually be considered an update as well, just a more extensive one. However, many package developers do not make use of that feature and while I try to recommend such a notice, I might miss it sometimes since I've been doing this review process for quite a while now and sometimes just assume that things will turn out well. (Actually, this is the first time I've heard of problems.)

greneholt:
If Extended Tab Switcher worked I'm sure it wouldn't have been a problem. For now I've just manually installed GotoOpenFile, as it is perfectly sufficient for my needs.

Blasz:
This plugin (Extended Tab Switcher) doesn't even work on windows (if at all) - https://github.com/rajeshvaya/Sublime-Extended-Tab-Switcher/issues/9. 

GotoOpenFile is exactly what I want and having to manually install it is a pain.

FichteFoll:
@rajeshvaya, please address this issue.

rajeshvaya:
Hello All, Sorry for the late response. I will look into it and fix it soon.

PR #3357, opened 2014-06-20
Removed the other repos and added them to previous_names

rajeshvaya:
This request is in reference to previous [pull request - #3330](https://github.com/wbond/package_control_channel/pull/3330)

Added previous_names of other repos

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -695,6 +695,21 @@
 			]
 		},
 		{
+			"name": "ExtendedTabSwitcher",
+			"details": "https://github.com/rajeshvaya/Sublime-Extended-Tab-Switcher",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"details": "https://github.com/rajeshvaya/Sublime-Extended-Tab-Switcher/tags"
+				}
+			],
+			"previous_names": [
+				"GotoOpenFile",
+				"ListOpenFiles",
+				"OpenFileList"
+			]
+		},
+		{
 			"name": "ExtendScript",
 			"details": "https://github.com/seblavoie/sublime-extendscript",
 			"releases": [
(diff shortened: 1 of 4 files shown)


rrg:
Sorry for the delay, I just checked this after someone responded here http://stackoverflow.com/questions/19665615/sublime-text-text-list-of-open-files/19694952#19694952

I'm fine with the change, can you please get this package back to work. 

Thanks

FichteFoll:
@rrg: @rajeshvaya appears to be rather active an the last push on https://github.com/rajeshvaya/Sublime-Extended-Tab-Switcher was 4 days ago. Please post issues there, unless he does not respond to them anymore or other issues with maintainer changes arise (or similar).

rrg:
Sorry I do not understand what the problem is? How do I get my extension back to package manager?

FichteFoll:

-

In your particular case, you've been asked about 9 months ago about your stance with your package, if you'd like to continue developing on it, have @rajeshvaya's features be implemented into your package or whatever. Since your package has been merged into ExtendedTabSwitcher as a result of no answer, your package now qualifies as a new package and needs to satisfy the criteria of being a non-duplicate again.

FichteFoll:
After looking into your package I realize that your package actually provides different functionality from the other three packages (two of which were merged into one other) in that it provides the open files in a text form, likely for copy and paste purposes. I am not sure why I thought that your package's functionality was too similar to the other ones 9 months ago, but I certainly don't see it now. The functionality of course *could* be merged into ExtendedTabSwitcher since the have somewhat similar functionality, but it's different enough to warrant being its own package.

Additionally, after a few searches I was not able to find a similar package to this currently in the list, which means you package would satisfy the non-duplicate criteria.

I sincerely apologize for my misjudgment that I did not verify due to the assumption of the merge being correct. Let's continue this in #4169.

rrg:
Thank you, I'll take a look at the other thread, feel free to close this down.

PR #3545, opened 2014-08-22
Add a .editorconfig file to help contributors keep code styles

taylon:
Hey guys,

I did a pull request for a package a couple of minutes ago and my build failed twice because my sublime was setup for using spaces instead of tabs for indentation.

So I added a .editorconfig file to try to avoid that in the future.

EditorConfig (http://editorconfig.org/):

EditorConfig helps developers define and maintain consistent coding styles between different editors and IDEs. The EditorConfig project consists of a file format for defining coding styles and a collection of text editor plugins that enable editors to read the file format and adhere to defined styles. EditorConfig files are easily readable and they work nicely with version control systems.

So I think it is a powerful way to make sure that all contributors keep the same code style. We can change the configuration to match the style guide of the project, I just suggested some configurations that I think are currently used in the project.

It should also be added to the documentation, I didn't add it myself because Im not good with that and I'm not sure where is the best way to add it.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,15 @@
+# editorconfig.org
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+ident_size = 4
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.json]
+indent_style = tab
+
+[*.py]
+ident_style = space


evandrocoan:
> @FichteFoll I personally don't consider this to be useful, especially because those forcing their own standards on other people's projects should be punished for that

This should only force the indentation and line styles on this project. But I never used this plugin, so I do not know whether this will mess up with something. Anyways, this pull request is only redirected to the ones who already are using the `EditorConfig` package.

It has been on stack for 1214 days or 3 years, 3 months, 27 days:
1. https://www.timeanddate.com/date/durationresult.html?d1=8&m1=09&y1=2014&d2=4&m2=01&y2=2018

Can we either merge this or close this?

FichteFoll:
I guess it doesn't hurt having this around for when people *do* use the package. I tried it out once a while ago, but wasn't too happy with the user experience.

PR #3965, opened 2015-01-02
Add Git Mode package info.

markbirbeck:
Hi there.

This is YAGM (Yet Another Git Module)...but I feel it has value, because it is much easier to add new commands and features than many of the other implementations. This is because most functionality can be added by specifying the command to perform, and then setting some flags to indicate where any extra input should be gathered from (such as new branch or remote repo names), how the output should be directed (such as to a new buffer or just to a panel), and how it should be formatted (by automatically wiring in syntax files).

As I say in the ['Why' section of the README](https://github.com/markbirbeck/sublime-text-gitmode#why):

> The Git plugin is great, but:
>
> * it relies on a great deal of code for its functionality, yet a lot of this functionality is boiler-plate and could be easily factored out; GitMode uses Shell Command which is a factoring of this kind of boilerplate code, plus a few other features;
> * relying on code for functionality, rather than configuration, makes it difficult to contribute to and to customise. This is especially frustrating for small features;
> * the Git plugin uses slightly different terminology to Git itself, which seems unnecessary.
>
> The SublimeGit plugin is also nice, but it costs money and is not open source.

I hope you'll be able to add Git Mode to the channel.

Thanks!

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -271,6 +271,17 @@
 			]
 		},
 		{
+			"name": "Git Mode",
+			"details": "https://github.com/markbirbeck/sublime-text-gitmode",
+			"labels": ["git", "emacs"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Git Status Files",
 			"details": "https://github.com/mkraft/git-status-files",
 			"releases": [


FichteFoll:
I don't think the ShellCommand requirement is prominent enough, but otherwise we should be fine.

markbirbeck:
Thanks for your quick turnaround.

I will make the dependency more explicit, but it looks like even the
one-liner that I have in the 'install.txt' file is not being seen. :( I
just did a test install on a Windows VM and the 'install.txt' file was not
shown once installation was complete.

One route I'd like to take is to use the new dependency feature; this seems
ideal for saying that Shell Command is a dependency of Git Mode. But the
docs say that a dependency shouldn't be a plugin -- do you think that
requirement will ever be loosened? If not then I guess I'll need to factor
out the Shell Command core code so that it can be a dependency for *both*
Shell Command and Git Mode?

Any thoughts or advice?

FichteFoll:
The problem with your `messages.json` is that it's named `message.json` on your repo (missing `s`). Make sure to include the install message in your next update message as well as nobody has seen it so far.

You are right, package dependencies (for ST itself) are currently out of scope. This may or may not be permanent but likely won't change in the near future, talking about months here. Extracting the core of shellcommand and git mode into an own dependency might be the "most optimal" choice here but just better highlighting the dependency in install message and readme should also do if you're lazy.

markbirbeck:
Not lazy...but certainly dim! Thanks for spotting the silly error in the
JSON file name!

I think I'll take a look at doing the factoring at some point, since if
Package Control works as promised by bringing in the dependencies
automatically, then it will make for a very smooth install experience.

Thanks again for your prompt replies, they're very much appreciated.

PR #3966, opened 2015-01-02
add fibjs

ngot:


ngot:
This is a code completion plugin for [fibjs](https://github.com/xicilion/fibjs)

FichteFoll:

I don't see why your package would need a `.no-sublime-package` file, please remove it.

The package name does not indicate that this package provides "completions". Please add it (and make it for example "fibjs Completions").

ngot:
@FichteFoll Thanks for you review. 
This is the first time I submit a package. Sorry for the mistake what I've made. I have fixed the problemes you mentioned above，please checkout it out.

PR #3968, opened 2015-01-03
Add new plugin Visualforce preview

Lanceshi2:
Add a new plugin VisualforcePreview. This plugin basically enables sublime user to preview Visualforce pages inside sublime text

### DIFF
--- a/repository/v.json
+++ b/repository/v.json
@@ -392,9 +392,9 @@
 			"author": "Lance Shi",
 			"releases": [
 				{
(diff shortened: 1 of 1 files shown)


FichteFoll:

If you want your plugin to work cross-platform (which you should) you should use `os.path.split()` instead of splitting on a hardcoded backslash.

Lanceshi2:
Updated according to advises. And thank you for pointing out the issue with hardcoding '\'.

PR #3972, opened 2015-01-04
Remove space between words - SexySnippets

felquis:
I saw that space causing some issues in the URL
screen shot 2015-01-04 at 18 26 40

So I removed.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -821,11 +821,12 @@
 			]
 		},
 		{
-			"name": "Sexy Snippets",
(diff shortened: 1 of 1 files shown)


FichteFoll:

While these "some issues" is a normal thing for URIs with spaces, using none is generally preferred as it is easier to import them in Python plugins in case you needed to, but you are not using any in the package anyway.

felquis:
I added the `previous_names`.

PR #3974, opened 2015-01-05
Update i.json

jabyrd3:
Sorry about the last pull request, I forgot my fork was so old.

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -614,7 +614,7 @@
 				{
 					"platforms": "*",
 					"sublime_text": "*",
-					"details": "https://github.com/jabyrd3/sublime-ixio/tags"
+					"tags": true
 				}
 			]
 		}
--- a/repository/i.json
+++ b/repository/i.json
@@ -606,6 +606,17 @@
 					"branch": "master"
 				}
 			]
+		},
+		{
+			"name": "ixio",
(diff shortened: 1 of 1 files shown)


FichteFoll:
You can delete your `package-metadata.json` as it will be replaced on install.
You can also delete your `.no-sublime-package` file as it's not necessary and should only be included if you really need it.

I don't think it is necessary to include the requests package for just two http POST requests, that's possible with urllib as well. This wouldn't be a problem if the requests module existed as a dependency for Package Control already but we'll have to work something out about that in the future.

And in other news ... why would you do this? [`s = sublime.window`](https://github.com/jabyrd3/sublime-ixio/blob/9f8b617376d688e1bfd28f2ef886c9737f7b5411/ixio.py#L24)

jabyrd3:
Thanks, I've deleted those two files. Tomorrow I'll rework the package to use urllib again.

The `s = sublime.window` bit is an embarrassing leftover that snuck in. Thanks for spotting it.

jabyrd3:
Using urllib instead of requests now, removed .no-sublime-package and package-metadata.json. The newest release tag is 1.0.2.

FichteFoll:
Well, that's ... a lot of commits.

Anyway, looking good now. I recommend you to not load the settings at the beginning though as the user might change them while ST is running and they will not get picked up this way.

PR #3978, opened 2015-01-06
Added Diction

GeeF:
Plugin supporting GNU diction style checker

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -393,6 +393,17 @@
 			]
 		},
 		{
+			"name": "Diction",
+			"details": "https://github.com/GeeF/SublimeDiction",
+			"labels": ["linting", "stylecheck"],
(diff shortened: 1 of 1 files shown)


FichteFoll:
Does this work for Windows? If not you should restrict users from installing it there.

Also you should put your `os.environ['PATH'] += os.pathsep + '/usr/local/bin'` line in a conditional so it only gets executed on OSX.

And finally it seems that you are reading all the settings at startup. That means any configurations by the user at runtime won't get picked up.

GeeF:
Sorry, this is my first Sublime Plugin :)

Excluded windows for now. There is a cygwin version of diction but I can't test it right now. Will update, once I get the chance.

Also settings will now load at runtime.

Thanks for the suggestions.

PR #3979, opened 2015-01-07
Added "Indent and braces" package

Epskampie:
Added the https://github.com/Epskampie/sublime_indent_and_braces package.

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -231,6 +231,19 @@
 			]
 		},
 		{
+			"name": "Indent and braces",
+			"description": "Indent selection and wrap it in braces.",
+			"author": "Simon Epskamp (Epskampie)",
+			"details": "https://github.com/Epskampie/sublime_indent_and_braces",
+			"labels": ["text manipulation"],
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
I can't see anything in your package that wouldn't work on ST3 so you can enable it for that.

By the way, your `for r in copy.deepcopy(sel):` loop will not work correctly because you are modifying the view's contents in each loop and then trying to access the old locations. On ST3 this practically won't do anything at all since the `Selection` class there does not actually contain elements but forwards item accesses to an API call which can obviously not be copied.
`copy.deepcopy(view.sel())` actually *fails* on Windows ST2 with a RuntimeError because `sublime.RegionSet` is not pickleable.

Instead, I recommend you do this: `for r in tuple(reversed(view.sel())):` as that will fix all the flaws I mentioned above, *unless* there is a line with an end and a start of a selection.

Epskampie:
Thanks for your comments! 

I actually originally meant to say __only__ sublime text 3, but with your pointers I now made the package sublime text 2 and 3 compatible.

PR #3954, opened 2014-12-29
Plugin VN IME - VNI

88d52bdba0366127fffca9dfa93895:
Gõ tiếng Việt trên Sublime Text 3 - kiểu gõ VNI

PR #3980, opened 2015-01-07
Sublpress rebirth.

helmutgranda:
After doing some digital forensics I was able to revive a project that has been left unattended for a little more than a year:

https://github.com/brzrkr/Sublpress/

The creation of this repo is to provide the feature to users to be able to publish/edit content to their Wordpress blog.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -2884,6 +2884,15 @@
 			]
 		},
 		{
+			"details": "https://github.com/helmutgranda/sublpress2",
+			"releases": [
+				{
+					"sublime_text": "<3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"details": "https://github.com/facelessuser/SubNotify",
 			"releases": [
 				{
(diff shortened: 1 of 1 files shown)


helmutgranda:
I just noticed the thread below:

https://github.com/wbond/package_control_channel/pull/3954

Let me know if we should follow a similar route instead of using Sublpress2

FichteFoll:
Yes, the same procedure would be preferred for sublpress as well.

Since the original package is in an awful state as well (last commit deleted many files, `.pyc` all over the place), hasn't been fixed in more than a year and is currently not even installable on package control for some reason (probably casing of the url), I won't even wait for a response here. But for fairness's sake I will mention the original author so he at least *can* see what is being done here.

cc @brzrkr

helmutgranda:
Changes done.

FichteFoll:
Tags must be valid semantic versions (so `0.1.0` or `v.0.1.0`).

helmutgranda:
Tag release version updated to 0.1.0.

PR #3982, opened 2015-01-08
Added DustBuster

zanuka:
DustBuster includes:
- enhanced syntax definition for .dust files
- better comments for .dust files
- sublime completions of currently supported dustjs helpers, plus some helpful extras

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -907,6 +907,7 @@
 		{
 			"name": "DustBuster",
 			"details": "https://github.com/zanuka/dust-buster",
+			"labels": ["language syntax", "auto-complete", "snippets"],
 			"releases": [
 				{
 					"sublime_text": "*",
--- a/repository/d.json
+++ b/repository/d.json
@@ -905,6 +905,16 @@
 			]
 		},
 		{
+			"name": "DustBuster",
+			"details": "https://github.com/zanuka/dust-buster",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "DXL",
 			"details": "https://github.com/Adarma/DXL",
 			"labels": ["build system", "color scheme", "language syntax", "linting"],
--- a/repository/d.json
+++ b/repository/d.json
@@ -895,17 +895,8 @@
 			]
 		},
 		{
-			"name": "Dust.js",
-			"details": "https://github.com/sntran/sublime-dustjs",
-			"releases": [
-				{
-					"sublime_text": "*",
-					"branch": "master"
-				}
-			]
-		},
-		{
 			"name": "DustBuster",
+			"previous_names": ["Dust.js"],
 			"details": "https://github.com/zanuka/dust-buster",
 			"labels": ["language syntax", "auto-complete", "snippets"],
 			"releases": [
--- a/repository/d.json
+++ b/repository/d.json
@@ -898,7 +898,7 @@
 			"name": "DustBuster",
 			"previous_names": ["Dust.js"],
 			"details": "https://github.com/zanuka/dust-buster",
-			"labels": ["language syntax", "auto-complete", "snippets"],
+			"labels": ["dust", "language syntax", "auto-complete", "snippets"],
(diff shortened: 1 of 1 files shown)


FichteFoll:
How does this compare to https://packagecontrol.io/packages/Dust.js ?

zanuka:
Glad you asked @FichteFoll 
DustBuster's syntax definition makes dust.js template code much more readable. I tried the Dust.js package and it turns a whole line of dust the same color. Also, comments were improved in DustBuster. Dust.js package wasn't working properly, as they were using html comment tags , instead of dust tags {! !} 
I also added support for all of the major dust templating helpers, where the Dust.js package only had a handful. 

We've been using it on an internal dev team currently building a framework that uses dust.js templates and find that DustBuster allows us to be more productive. I will also be actively developing DustBuster to add new features.

Dust.js syntax def
dustpackage-syntax

DustBuster syntax def
dustbusterpackage-syntax

Thanks, 

Mike

zanuka:
Dust.js package - html style comments 
dustcomments

DustBuster package - dust style comments {! !}
dustcomments

FichteFoll:
cc @sntran

zanuka:
Thanks @FichteFoll 
Looking at the commit history on the current Dust.js Sublime package, it appears that @sntran is not actively maintaining this package any longer. Our team plans to continuously adapt and refine the feature-set of this package with the goal of being the most comprehensive Dust.js Sublime package available. Although we are open to working with, and taking suggestions from @sntran, we think that integrating our work into his repo may not be the best approach. In the future, we'd like to make this more than just a syntax definition with completions. We believe it's okay to have competing packages in Package Control (e.g. Handlebars, Angular, Ember, etc. all have various package alternatives)

Looking forward to hearing your thoughts

sntran:
I am okay with either solutions that @FichteFoll prefer. I set out to make my package because back then, there was none. There have not been a whole lot of requests for my package, and as I have not been working with Dust for a long time, I think @zanuka package is more well maintained at the moment, and more tailored to the need of the community using Dust.js.

As for the plan for my own package, because it merged with somebody else, I would not be able to take it down. @FichteFoll can decide to remove it from Package Control though, and replace with DustBuster. I have no problem with that :)

zanuka:
Thanks @sntran , much appreciated :-) Our team is building a framework that uses Dust.js so we have a great interest in keeping this package well-maintained.

FichteFoll:
Replacing (or taking over) an existing package while maintaining its core functionality, increasing functionality and likely providing more frequent updates is usually the best outcome of this situation. The main advantages are:

1. Old users are updated to the new package automatically.
2. New users are not confused about two packages with a very similar set of features and don't have to try both (if they actually find them).

Replacing the entry on Package Control is the easiest option here. @zanuka, you may decide whether you want to rename the package as well in the process or keep the old one.

zanuka:
Ok I've replaced the entry and added the previous_names accordingly. I will also pull in the existing trigger completions from sntrans version so that users will still have access to those. I will have a 1.0.1 release tag out today.

sntran:
If possible, please credit @gpbl for his auto completions in your package.

zanuka:
@sntran sounds good, will do.

zanuka:
1.0.1 released > https://github.com/zanuka/dust-buster/tree/version/1.0.1
@gpbl credited, previous completions merged in

FichteFoll:
Great! Wish it was always this smooth and quick :smile:

zanuka:
Excellent. Thanks @FichteFoll !

zanuka:
@FichteFoll , how long does it take for Package Control to pickup a new tag version? It's updated on the site now but seemed to take about an hour. Also, shouldn't the package auto-upgrade on ST restart? I had to go in an do a manual Upgrade Package from the command menu. Thanks

FichteFoll:
It takes approximately 15 to 60 minutes for new tags to get picked up. Package control only runs once every hour, so if you relaunch St in quick succession it would only check for updates once.

zanuka:
Ok, thanks for the clarification.

PR #3983, opened 2015-01-08
Adding GoToTest.

ahare:


FichteFoll:
I am fairly sure that I have seen a package like this before, but I can't find it. Might have been another language.

Either way, using branches for releases is deprecated and won't be accepted for new packages anymore: https://packagecontrol.io/docs/submitting_a_package#Step_4

ahare:
@FichteFoll I've updated to use tags. I wasn't able to find a package that does this for golang.

PR #3985, opened 2015-01-09
suggest Fonty

SammaySarkar:
added Fonty (https://github.com/SammaySarkar/Fonty)

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -830,6 +830,16 @@
 			]
 		},
 		{
+			"name": "Fonty",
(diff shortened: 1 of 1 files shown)


FichteFoll:
As you mention in your readme, this package is very similar to FontCycler. Maybe you and @akshaykumar90 can work something out together to combine the features of both packages into one and include compatability for ST2 and ST3?

SammaySarkar:
I'd love to. (in fact I had at one point actually removed my repo after finding that out)
But first I have to learn Python enough... :/

PR #3987, opened 2015-01-10
Added AutoRefresh entry

Waterflames:


FichteFoll:
I suggest instead of `(type(refreshRate) is not int and type(refreshRate) is not float)` you use `not isinstance(refreshRate, (int, float))`, not only because it's shorter but because it's the preferred way to check types.

Another idea is to allow multiple threads simultanously to reload multiple files by storing all threads in kind of a dict associated by the view's id.

Waterflames:
I have added the 1.0 tag (I think?), replaced the typecheck code, and implemented dictionary storage.
Any other requests?

FichteFoll:
Looks good.

The only thing I noticed is that PEP8 suggests to compare with `None` using the `is` operator and `is not` respectively, but usually you can just use automatic booleanization (is that even a word) in boolean contexts like `if` statements or binary operands (`and`, `not`, `or`). Very minor remark.

PR #3988, opened 2015-01-11
Add Terminality plugin

spywhere:


spywhere:
I'm curious, if I'm using tags as releases, is readme fetched from the release or the latest commit?

I just want to add screenshot into the readme file.

FichteFoll:
The readme is fetched from master.

The general idea of your package has been implemented already, but I think your method and approach to this are different enough.

PR #3992, opened 2015-01-13
Add Virgil language syntax

lpstein:
Language itself is not public right now, hoping to get it open sourced before too long.  I hope that's not an issue?


FichteFoll:
Fine by me. You should add a readme with general information about the repo/package though, as to what it contains or how it looks with a screenshot.

PR #3994, opened 2015-01-13
Adds Coffee2Js package

mattseymour:
CoffeeScript to Js package added to the repository list.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1155,7 +1155,7 @@
 			"details": "https://github.com/mattseymour/sublime-coffee-to-js",
 			"releases": [
 				{
-					"sublime_text": "<=3000",
+					"sublime_text": "*",
 					"branch": "master"
 				}
 			]
--- a/repository/c.json
+++ b/repository/c.json
@@ -1156,7 +1156,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"branch": "master"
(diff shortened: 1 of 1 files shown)


FichteFoll:
Don't forget to push correctly formatted tags to the repo, e.g. `v0.1.0`.

`"sublime_text": "<=3000"` doesn't make much sense for a version selector. If you remove the trailing comma in `coffeetojs.sublime-settings` it will work for both ST2 and ST3 so I suggest you replace it with `"*"`.

It would be nice if your readme actually said something about what it does exactly and how it works. I only found out by reading the source code.

mattseymour:
Apologies about the readme. Missed it out of the previous commit.

@FichteFoll by :

> Using branches for releases is deprecated and won't be accepted for new packages any more: https://packagecontrol.io/docs/submitting_a_package#Step_4

Are you talking about my use of annotated tags for release numbers?

wbond:
@mattseymour No, he means that under releases you have specified you want to use the master branch for releases (`"branch": "master"`), however we don't support that method except for grandfathered packages.

Instead it should be `"tags": true` and you should create tags that are semantic version numbers to create releases, like your `v0.1.0` tag.

mattseymour:
Changes have been made to use `"tags": true`.

Thanks for your explaination @wbond

PR #3999, opened 2015-01-17
j-commander rename

zanuka:


FichteFoll:

And FYI, you can rename repos on github which will move all issues/pull requests and set up http redirects for you.

zanuka:
yeah forgot about that... just added previous name. thanks @FichteFoll , good to know about the repo renaming

PR #4003, opened 2015-01-18
switch `branch` to tag:true

xjsender:
switch ```branch:master``` to ```tag:true```


FichteFoll:
Another tip: You can append commits to existing pull requests by pushing onto the branch you opened it from.

PR #4004, opened 2015-01-19
Adds Copy To SharePoint Root

eakron:
Adds a simple helper for working with SharePoint development in Sublime Text which lets you quickly copy files to the "hive" using a Sublime build system and a Powershell script.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1844,7 +1844,7 @@
 			]
 		},
 		{
-			"name": "Copy To SharePoint Root",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Your package name here and the name you hardcoded in your build system don't match. You have to use the same for both.

eakron:
Removed the spaces in the package name

FichteFoll:
Now you need to update the package's position in the file ;)

eakron:
Done!

PR #4005, opened 2015-01-19
Coffee2Js `tags` option was using string not bool

mattseymour:
Altered Coffee2Js option to use `"tags":true` instead of `"tags":"true"`


FichteFoll:
Oops, didn't notice that.

PR #4006, opened 2015-01-19
Added Craft Development Snippets

alecritson:


FichteFoll:
You have to push version tags to your repo, otherwise Package Control won't find a release.

alecritson:
Thanks for the heads up! I've added a version tag

FichteFoll:
Okay, but it has to be a semantic version (or at least be in the format of that) which requires three number parts.

alecritson:
okay sure, i've updated the tag :)

PR #4007, opened 2015-01-19
Adding Redacted package

nickgravelyn:
Adding my Redacted package: http://github.com/nickgravelyn/redacted

Followed https://packagecontrol.io/docs/submitting_a_package. Tests all passed.

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -454,6 +454,16 @@
 			]
 		},
 		{
+			"name": "Redacted",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Won't work with ST2 right now because `plugin_loaded` is not called there.
Just add `if int(sublime.version()) < 3000: plugin_loaded()` to your code and it will be fine.

I'll merge this ahead of the change.

PR #4008, opened 2015-01-20
Added Vorhees package v1.0.0

frankjonen:
Added the first release of the Vorhees JSON color scheme and language description.


FichteFoll:
Your `author` and `readme` keys are redundant, but that's not a problem.

PR #4011, opened 2015-01-21
Add hacklang typechecker

SiebelsTim:
Ref https://github.com/SiebelsTim/hack-sublime/issues/5

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -32,6 +32,16 @@
 			]
 		},
 		{
+			"name": "Hacklang Typechecker and Autocompletion",
+			"details": "https://github.com/SiebelsTim/hack-sublime",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Haml",
 			"details": "https://github.com/phuibonhoa/handcrafted-haml-textmate-bundle",
 			"releases": [


FichteFoll:
You should remove the `.sublime-package` from your repo since it's unnecessary data and instead have all its contents in the initial repo (which might as well just be that .py file).

Also you need to push a versioned tag.

SiebelsTim:
@FichteFoll thanks, do I *have* to remove the .sublime-package file? I like having it in there in case someone wants to install it without the package manager.

FichteFoll:
I suggest you host it in a separate branch instead or use the github releases feature that allows you to upload binary "builds". You'd have to update the download url for every release though.

SiebelsTim:
All right. Ping.

FichteFoll:
I actually meant it the other way around. Especially since using branches for releases is deprecated and won't be accepted for new packages anymore: https://packagecontrol.io/docs/submitting_a_package#Step_4

SiebelsTim:
So I can't specify a specific branch where the packagecontrol is searching for the correct tag? Does it have to be in master?

FichteFoll:
Tags are completely independant of branches. You can even create and push a tag that is not on a branch (anymore) because you reset the branch or didn't push it.

SiebelsTim:
Oh! It's never too late to learn :)

changed again, sorry for the inconvenience

FichteFoll:
Alright, but one last thing: It has to be a semantic version in the form `v1.0.0`.

SiebelsTim:
aaah, right. Sorry.

PR #4013, opened 2015-01-21
Add Fugly Syntax Highlight

diogomoretti:
Sublime Text (2 and 3) syntax highlight for [Fugly](https://github.com/pablo-cabrera/fugly-js).

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -947,6 +947,7 @@
 		{
 			"name": "Fugly Syntax Highlight",
 			"details": "https://github.com/diogomoretti/fugly.tmLanguage",
+			"labels": ["language syntax"],
 			"releases": [
 				{
 					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


FichteFoll:

diogomoretti:
Fixed branch. Waiting new review =]

FichteFoll:
Okay, but you need to push tags to your repo.

@wbond, I think the docs need to be a bit more verbose in this regard, this has happened a lot in the past. Maybe just providing example git commands and an example tag name is sufficient.

diogomoretti:
I added the labels ;) pls, check it.

FichteFoll:
While this is also good, it's not what I meant either.

I meant you have to create a tag with `git tag v1.0.0` and then push it with `git push origin v1.0.0` (or alternatively `git push --tags`). It will then be listed here: https://github.com/diogomoretti/fugly.tmLanguage/tags

diogomoretti:
Tag added! ;)

diogomoretti:
thanks ;)

PR #4016, opened 2015-01-22
added CloseOtherWindows

werkzeugh:


FichteFoll:
You should remove the `TestWindowCommand` class and the debug print in closeow command.

NicoSantangelo:
@FichteFoll @werkzeugh  Hey! I was wondering what was the difference between [CloseOtherWindows](https://packagecontrol.io/packages/CloseOtherWindows) and [CloseAllButThis](https://packagecontrol.io/packages/CloseAllButThis) ? I've been using the latter and didn't understood this addition.

Sorry if I'm not being useful just think it can be a bit confusing, even the code looks really similar :)

FichteFoll:
Basically, CloseOtherWindows closes ... all other windows and CloseAllButThis closes ... ~~all but this~~ all other windows and all other files in the last remaining window as well.

I agree, this could have been merged into a single package but I didn't find the older package before.

werkzeugh:
despite the name, CloseAllButThis closes only other tabs in the current view but does not
touch other open windows. this is why i wrote my own.

CloseAllButThis only enumerates  self.window.views() while this package enumerates sublime.windows()

but such a close-other-tabs-feature is already  built into sublime by right clicking on a tab,
but its not available as a keyboard-shortcut.

➜ so i also included a Close-Other-Tabs feature in version 1.1 (should i rename the package then, or split it in two ?) 

✔  i removed TestWindowCommand class and the debug print in closeow command.

FichteFoll:
You could rename it to "CloseOther", incorporate the functionality of `CloseAllButThis` and then replace that on PC as well.

PR #4019, opened 2015-01-22
Sublime Text version indicator set to pre 3.0 only, redundant author removed

frankjonen:
Thanks for the pointer. Fixed it.

### DIFF
--- a/repository/v.json
+++ b/repository/v.json
@@ -444,11 +444,10 @@
 			"name": "Vorhees",
 			"details": "https://github.com/frankjonen/vorhees/",
 			"labels": ["color scheme", "formatting", "language syntax"],
-			"author": "frankjonen",
 			"readme": "https://raw.githubusercontent.com/frankjonen/vorhees/master/README.md",
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": "<3000",
 					"tags": true
 				}
 			]


FichteFoll:
Damn, I just realized that you changed from `*` to `<3000` and not the other way around. This change was unnecessary since you don't define any plugins (`.py`) in your package and everything else works the same in ST2 and 3.

frankjonen:
Not everything apparently. It doesn't work in ST3. I either get a "not found" error or on a later install (just now) nothing happens and all text renders white.

So for the time being I'd rather exclude it from that branch. It's better than annoying people with a bad user experience.

Sorry for the hassle.

FichteFoll:
Okay, I did some investigation since, theoretically, there shouldn't be issues. I eventually found out that you are referencing your color scheme in the settings with `"color_scheme": "Packages/Vorhees/Vorhees (JSON).tmTheme"`, which is wrong since it must be `Packages/Vorhees/Vorhees/Vorhees (JSON).tmTheme` since you placed it in a sub directory of your repo. Since the file doesn't exist, ST shows an error message and uses its fallback "black-on-white" color scheme once you set a file's syntax to Vorhees.

Here is the thing though: It doesn't work on ST2 either. I suppose you actually didn't install it on ST2 but just tested with the "Vorhees" sub directory of your repo in your packages directory that is decoupled from the git repo.

So, I suggest you to either fixe the path in the settings or just move all the files in that folder upa level so that they are at the package root (which makes more sense since there are only 3 anyway).

frankjonen:
Oh dear. That shouldn't have happened. I have absolutely no idea why it's not using the zip file in the releases but just takes the entire repo. That also explains why I didn't get it to work with the downloaded version. When things look simple…

FichteFoll:
Well, Package Control just downloads the [zipball of your tag](https://github.com/frankjonen/vorhees/archive/1.0.0.zip), it never even looks into your releases (since the API for that was only added recently). wbond/package_control#497 contains relevant discussion and infos regarding this.

frankjonen:
If I got that right, it'd mean you'd have to support multiple avenues for looking into files on Github, which sounds like a lot of extra work that isn't all that beneficial to the end user (of PC). 

Since I don't have to publish the build source separately (it's easy enough to reverse it anyway) there's no harm done in cleaning up my repo.

Thanks for your patience. I really appreciate it and I might have learned a thing or two.

FichteFoll:
Glad you got something out of this.

frankjonen:
Yep, I now have a better grasp on how this works. Thanks again for your help.

PR #4027, opened 2015-01-24
Add PgcliSublime package

darikg:


FichteFoll:

Since you already pushed a tag, this shouldn't be much of a problem.

darikg:
Whoops, sorry about that. I just copy/pasted another package, and forget to check the docs. Fixed!

PR #4028, opened 2015-01-24
updated sass director

Stephn-R:


FichteFoll:
Any specific reason for the rename?

Stephn-R:
We decided to move the project to an organizational account. Bigger plans haha

FichteFoll:
I was mostly referring to the change of the `name` attribute. Btw, you can rename and move repositories on github itself which will preserve issues and set up http redirects.

Stephn-R:
Ah, the name change is not too specific but we would prefer it be called "Sass-Director" and not "SassDirector"

PR #4030, opened 2015-01-24
Adds sublime django package

mattseymour:
This package will be a collection of useful django snippets and commands
which will closely follow the django api and best programming practices.
Currently I am working through the django 1.7 template creating snippets.
Later it will include models, forms, and django commands executable via
shortcuts.

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -498,6 +498,16 @@
 			]
 		},
 		{
+			"name": "Django",
+			"details": "https://github.com/mattseymour/sublime-django",
+			"releases": [
+				{
+					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


FichteFoll:
How does it compare to the [other Django packages](https://packagecontrol.io/search/django)?

mattseymour:
Whilst the package is currently in the early development stages it could be classed as similar to other django packages. My aim over the coming weeks is to support a whole host of new and useful commands. With the aim of making this package more of a django specific ide very different to what's currently about.

FichteFoll:
Uhm, okay. Let's hope you deliver.

PR #4031, opened 2015-01-24
Added donate links.

davidolrik:


FichteFoll:
You might want that link refer to the english version of paypal.

PR #4034, opened 2015-01-26
Add Bunting package

xhacker:


FichteFoll:
Version must be conform to the semver syntax, which is `v<major>.<minor>.<patch>`, so three numbers.

xhacker:
@FichteFoll: updated

PR #4035, opened 2015-01-26
Add Hound Search

bgreenlee:
adding Hound Search plugin

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -568,6 +568,16 @@
 			]
 		},
 		{
+			"name": "Hound Search",
+			"details": "https://github.com/bgreenlee/SublimeHound",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"branch": "master"
+				}
+			]
+		},
+		{
 			"name": "How Do I Code Search",
 			"details": "https://github.com/jlangston/SublimeHowDoI",
 			"releases": [
--- a/repository/h.json
+++ b/repository/h.json
@@ -572,7 +572,8 @@
 			"details": "https://github.com/bgreenlee/SublimeHound",
 			"releases": [
 				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
Specifying a branch for releases has been deprecated and is not accepted for new packages anymore.

By the way, I would not like my double click to be overridden by a plugin *that doesn't even mention this*.

bgreenlee:
I'll document the double-click behavior, and specify how to remove it. It does not interfere with the default Sublime double-click, as that gets passed on. It is only a problem if there are other plugins that use double-click (AFAIK, there's no way around that currently)

bgreenlee:
My solution, btw, was to make the double-click handling off by default, toggleable by a user setting: https://github.com/bgreenlee/SublimeHound/commit/c4c6acda3ac0bc241ae2e00c3f227e9ce94d9948

FichteFoll:
If it doesn't interfere with the default implementation (didn't check if it was forwarded) it would probably be okay then, but a short mention of this would be nice.

Either way, the current way of just copying the file won't do since ST3 does not require extracted packages anymore and PC uses that. You'd need the `sublime.load_resource` API to get the virtual file's contents, but it only exists on ST3. Furthermore, you need to create the folder since it won't exist on first install. A way around this is to provide a `.no-sublime-package` file in your repo which instructs PC to manually unextract your package, but this is not desireable as the end user loses out on the overriding feature and should only used if necessary or really inconvenient otherwise.

bgreenlee:
Arg. Thanks. Sounds like I have no good options to allow the user to be able to enable/disable double-click functionality, then.

bgreenlee:
Ok, back to a default mousemap, but well-documented.

PR #4039, opened 2015-01-27
Add MacDown App Menu

diimdeep:


FichteFoll:
Branch releases have been deprecated, you must use tags for releasing.

diimdeep:
@FichteFoll sorry, maybe it is possible to check about this in unit tests(TravisCI) and/or in ChannelRepositoryTools

FichteFoll:
No, because the older packages may still use it and detecting which package has been added is not exactly trivial (there are many edge cases).

PR #4041, opened 2015-01-29
add Theme - DC

dubeg:
Add theme for Sublime Text 3

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -510,6 +510,17 @@
 			]
 		},
 		{
+			"name": "Theme - DC",
+			"details": "https://github.com/dubeg/theme-dc",
+			"labels": ["theme"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Theme - Eightlime",
 			"details": "https://github.com/HennerS/Eightlime",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Since theming hasn't really changed between ST2 and ST3 you should be able to allow this package for ST2 as well.

dubeg:
Ok, the theme should work with ST2 now.

dubeg:
Just a question because i don't know where to discuss this. Should I remove the file icons from the theme package and put them in a new one? Most icons are from the Seti UX/UI package, but I changed and added a few. Hmm.

FichteFoll:
Why would you want to do that? You need those icons in the theme, don't you?

dubeg:
Yeah, well, you can have custom file icons without a custom theme. And maybe some ppl prefer the original Seti icons to some of my versions. 

Also, I just tested this: some pref. files seem to conflict with Cobalt2. I guess because their names are slightly different and they don't get overriden. You can disable the unused theme, but meh.

I thought it would be cool to have icon packages like we have for color schemes and UI themes. Anyway, I probably won't do it.

FichteFoll:
The problem with icon packages is that they need a file with the same name as the current theme in the package root so that the icons are considered. This could be achievable with some plugin code however.

Anyway, your decision.

PR #4045, opened 2015-01-31
Add JXA Package to repository

dharma-guardian:
In OSX Yosemite you can use Javascript now to program for the Open Scripting Architecture. Because the Script Editor by Apple is not ideal I want to add JXA integration in ST.

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -330,7 +330,7 @@
 				{
 					"sublime_text": "*",
 					"tags": true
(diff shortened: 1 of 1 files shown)


FichteFoll:
Since this is "just" a build system I would prefer if the package name reflected that.

dharma-guardian:
Change package name to "JXA Build System"

PR #4047, opened 2015-01-31
add PackageGrupper

unknownuser88:


FichteFoll:
How's this different from (or: "how does it compare to") https://github.com/STPackageBundler/package-bundler?

unknownuser88:
i was searching for such kind of plugin but unfortunately haven't found one.
this is almost the same stuff. 
Thanks much!

PR #4049, opened 2015-02-02
Replacing QuantifiedDev plugin with 1self

douglas-vaz:
QuantifiedDev is now 1self. Plugin changed to reflect the new name and use the latest platform APIs.


FichteFoll:

douglas-vaz:
Thanks a lot for pointing that out, have made the changes.

PR #4051, opened 2015-02-02
Add Cfengine syntax highlighter.

astraverkhau:


FichteFoll:
FYI, you can append commits to an existing pull request by pushing to the branch you opened it from.

I'm not sure if `v.0.1.0` will work (highly doubt it), so if your package is not listed on packagecontrol.io after an hour you might want to change that to `v0.1.0`, or just do it immediately.

astraverkhau:
Tag re-named to v0.1.0. Thanks!

PR #4054, opened 2015-02-04
Add Cubescript Syntax Highlighter

srbs:


FichteFoll:
Tip: The easier procedure to set a syntax for all files with a certain extension is to just open a file with that and then select **View → Syntax → Open all with current extension as… → Cubescript**. It does essentially the same but is easier.

srbs:
You know what, that is. Thanks for the tip!

PR #4055, opened 2015-02-04
add PyBuilder package

mriehl:
Tests are passing locally.


FichteFoll:
Note that because the gif is part of your repo every user will have to download it (on every update). Consider hosting it elsewhere.

mriehl:
Thanks for the heads up! I'll move it somewhere else.

PR #4056, opened 2015-02-04
Add NetLinx plugin to package repository

amclain:


FichteFoll:
Since you're using the `sublime.load_resource` API which is not available on ST2, it won't be compatible with it either. Which is not a bad thing imo, everyone should be using ST3 beta.

Anyway, you should either make it compatible with ST2 or change the sublime_text version selector.

amclain:
My documentation says to download ST3, so I'm ok with the version selector. Thanks for the tip!

PR #4057, opened 2015-02-04
Add Crestron plugin to package repository.

amclain:


FichteFoll:
If you want this to work in ST2 you need to remove [this trailing comma](https://github.com/amclain/sublime-crestron/blob/12910bef20176d29bbd894920b2a45ac238d5ad8/Crestron.sublime-build#L12).

amclain:
Trailing comma removed. Thanks for the tip!

PR #4059, opened 2015-02-05
add C++YouCompleteMe plugin.

glymehrvrd:
I have written an plugin introducing some IDE C++ features to Sublime Text. Hopes to publish this plugin to repository channel.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -78,7 +78,7 @@
 		},
 		{
 			"name": "C++YouCompleteMe",
-			"details": "https://github.com/glymehrvrd/SYCM",
+			"details": "https://github.com/glymehrvrd/CppYCM",
 			"labels": ["c++", "code completion", "syntax check"],
 			"releases": [
 			{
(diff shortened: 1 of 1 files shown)


FichteFoll:
This looks like pure python to me and no other executables involved, which means your `.no-sublime-package` file is rather redundant. If you don't need it, please remove - or is there anything I missed?

glymehrvrd:
@FichteFoll, it indeed needs other executables. Because the execuatables are platform-sensitive and I can't compile on each platforms, so I left this work to the users. `.no-sublime-package` is necessary, users need to put compiled component into the plugin folder before it works.

FichteFoll:
Ah okay. This is not how you should handle this as the package directory is overwritten completely on each update (also by design). You should definitely use a different approach for users to specify the executables, e.g. by a setting variable.

glymehrvrd:
Done. I didn't realize the folder will be completely overwritten on update. And I changed my plugin github addr to `CppYCM`.

PR #4063, opened 2015-02-06
added ng-annotate buildsystem

ctf0:


FichteFoll:
The repository schema was updated to `3.0.0`, which is why I asked you to reset in the first place. That means: Your release must include `"tags": true` instead of a `details` key.

Furthermore, your `issues` and `platforms` keys are redundant.

@wbond, what are we doing about "small" packages like this that literally consist of just two lines? Usually I reject them since this is really not much of an effort to come up with yourself (compared to actually installing the package). I would do the same here but figured I should ask about this at some point.

wbond:
I'm not opposed to small packages. Those two lines may be useful to some people unfamiliar with ST configuration.

In my opinion we should mostly focus on helping people not submit broken packages, or 100% duplicates.

ctf0:
for issues u r right and its removed ,but for platform check https://packagecontrol.io/docs/submitting_a_package

ctf0:
@FichteFoll thanx for the merge but the package is not avail on PC.

FichteFoll:
@ctf0, you didn't push a tag. Also, please remove `.no-sublime-package` as it's completely unnecessary.

ctf0:
fixed ,should i now make a new PR or can we merge it through here ?

FichteFoll:
It will be detected automatically.

PR #4064, opened 2015-02-06
Added Tera Term Language Package

tkyjhr:
Tera Term is a terminal software (popular in Japan to some extent), and Tera Term Language (ttl) is a macro language to write a script for Tera Term.
It seems there is no package for Tera Term Language in package control, so I created a simple package and would like to add the package if acceptable.

Thank you.

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -222,7 +222,6 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"platforms": ["windows"],
 					"tags": true
 				}
 			]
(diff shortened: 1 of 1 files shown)


FichteFoll:
Haven't seen a new package on bitbucket in a while.

Anyway, this package doesn't include anything Windows-specific besides the language likely not used somewhere else, which isn't really a reason to disallow people to edit code in that language on a Linux machine. Imo you should lift that platform restriction.

tkyjhr:
Thank you very much for the review.
I agree. I removed the platform restriction.

PR #4065, opened 2015-02-06
Added ANSI package

aziz:


FichteFoll:
Interesting idea for a plugin, but I personally would prefer a name more similar to the [wiki article](https://en.wikipedia.org/wiki/ANSI_escape_code) and not just "ANSI" which basically says nothing about what it does.

aziz:
It was really difficult to name this one! I'm going with **ANSIescape** which is similar to what it's been called in the vim land *AnsiEsc.vim*

PR #4068, opened 2015-02-07
Added LessThanParse

pigeontech:
Sublime Text Plugin that will convert less than signs into the equivalent entity, which comes in handy when trying to display code samples within HTML.

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -337,6 +337,16 @@
 			]
 		},
 		{
+			"name": "LessThanParse",
+			"details": "https://github.com/pigeontech/LessThanParse",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Levels",
 			"details": "https://github.com/mazurov/sublime-levels",
 			"releases": [


FichteFoll:
https://packagecontrol.io/packages/StringEncode
https://packagecontrol.io/packages/Hasher
and https://packagecontrol.io/packages/RegReplace can also do this (and more).

pigeontech:
The first two you linked work pretty well. I just wanted to only convert the less than character so that example code displayed in HTML is still pretty readable while in the editor.

Other than source code readability, the other issue I have with the more full featured ones is that they will also convert the `&` character. That means if you run encode on the same code sample twice (because you added/edited some), The `&` goes from `&amp;` (ok) to `&amp;amp;` (not ok). You could run decode first, but it's easier not to be surprised or have to keep going back and forth. For code samples to display properly in the browser, only the the `<` character needs encoded.

Admittedly, it's a pretty narrow use case. If you want to close, that's ok!

FichteFoll:
Well, it's technically just a simple "replace `<` with `&lt;`", but on a command. I honestly don't know if it's worth to include it since it's really just that and there are the other two packages who do more than that, and RegReplace which allows you to bind arbitrary replacements (also within selection only).

Thus, I'm inclined to close.

PR #4069, opened 2015-02-07
Opinionated JavaScript Snippets - follows current best practices and strict linting standards

levbrie:
There are lots of JavaScript snippets out there, and many of them are well-done in terms of their feature sets.  Unfortunately, they don't follow best practices for JavaScript code style and fail on most linters.  These snippets don't.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -1077,6 +1077,17 @@
 			]
 		},
 		{
+			"name": "Atomized Tomorrow Color Scheme",
+			"details": "https://github.com/levbrie/atomized-tomorrow-color-scheme",
+			"labels": ["color scheme", "themes", "color", "syntax theme"],
+			"releases": [
+				{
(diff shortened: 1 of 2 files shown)


FichteFoll:
What are the exact differences to https://github.com/jprichardson/sublime-js-snippets? I looked into a few snippets and couldn't find any. If they are not much, won't a pull request suffice?

I don't know if you intended to include your Atomized Color Scheme as well, but if you did: Using branches for releases is deprecated and won't be accepted for new packages anymore: https://packagecontrol.io/docs/submitting_a_package#Step_4

Also, the ["syntax theme"](https://packagecontrol.io/browse/labels/syntax%20theme) label does not exist (→not used by any other package) and I suggest you to remove it since it means the same as "color scheme", except that it's more confusing.

FichteFoll:
@levbrie ping

PR #4073, opened 2015-02-10
Add AceJump plugin

sundy-li:
This is  AceJump Plugin inpired by emacs AceJump~

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -69,7 +69,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"branch": "master"
+					"tags": true
 				}
 			]
 		},
--- a/repository/a.json
+++ b/repository/a.json
@@ -64,6 +64,16 @@
 			]
 		},
 		{
+			"name": "AceJump",
+			"details": "https://github.com/sundy-li/AceJump",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "master"
+				}
+			]
+		},
+		{
 			"name": "ACF Snippets",
 			"details": "https://github.com/iamhexcoder/acf_snippets",
 			"releases": [


FichteFoll:
Well, https://packagecontrol.io/packages/EasyMotion has the same intention and drew inspiration from the same emacs plugin and the corresponding vim one. It does things a bit differently (e.g. it doesn't use multiple letters: https://github.com/tednaleid/sublime-EasyMotion/issues/14), but the functionality is so overlapping that I don't think it makes sense to have two of them.

Maybe you can add the different functionality to EasyMotion or you add the one from EasyMotion to your plugin and we replace EasyMotion, if its original author agrees and he does no longer maintain it or something. 

cc @tednaleid

tednaleid:
If the new AceJump plugin does the same thing as the existing EasyMotion plugin and will be actively maintained, I'm willing to have EasyMotion be deprecated and point people towards AceJump.  I haven't been using Sublime for the last year or so because I was waiting for ST3 final to be released before spending more time on the plugin.

sundy-li:
Well, I have looked into the EasyMotion source code today, and I think it has doing this jump work much better than my Acejump , thanks @tednaleid  and @FichteFoll , so generous and kindness.

There is no need to have this overlapping plugin, I am aware of that, maybe sooner I will pull some merge request to make the EasyMotion better, thanks again~

FichteFoll:
From what I see, the best would probably be to fork EasyMotion, incorporate features of AceJump (optionally as options) and replace the package here in the channel, or alternatively to grant access to the EasyMotion repo instead.

I hope you two can sort this out, but feel free to mention me when you have questions regarding the process on package control's default channel. I'm closing this issue so you can open a new one once you're done.

PR #4074, opened 2015-02-10
Add Theme - LP

dubeg:
Add UI theme to the channel.


FichteFoll:
I personally would prefer a more verbose name than just two letters (which has been happening more often recently), but we don't have a policy on that currently so it will remain personal choice for now.

dubeg:
Alright, will do in the future.

PR #4075, opened 2015-02-10
Add Oceanic Next Theme

voronianski:
https://github.com/voronianski/oceanic-next-theme

### DIFF
--- a/repository/0-9.json
+++ b/repository/0-9.json
@@ -69,7 +69,7 @@
 		{
 			"name": "6to5",
 			"details": "https://github.com/6to5/6to5-sublime",
-			"labels": ["language syntax", "snippets", "JavaScript", "ECMAScript6"],
+			"labels": ["language syntax", "snippets", "javascript"],
 			"releases": [
 				{
 					"sublime_text": "*",
--- a/repository/0-9.json
+++ b/repository/0-9.json
@@ -69,7 +69,7 @@
 		{
 			"name": "6to5",
 			"details": "https://github.com/6to5/6to5-sublime",
-			"labels": ["language syntax", "snippets"],
+			"labels": ["language syntax", "snippets", "JavaScript", "ECMAScript6"],
 			"releases": [
 				{
 					"sublime_text": "*",
--- a/repository/o.json
+++ b/repository/o.json
@@ -90,11 +90,11 @@
 		{
 			"name": "Oceanic Next Color Scheme",
 			"details": "https://github.com/voronianski/oceanic-next-theme",
-			"labels": ["color scheme", "JavaScript", "ECMAScript6"],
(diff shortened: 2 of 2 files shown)


FichteFoll:

I would also prefer if you used the term "Color Scheme" more deliberate instead of "Theme" because that is incorrect ST terminology.

I doubt that `punctuation.definition.tag.*.*` is actually a scope name and you intended to just write `punctuation.definition.tag` here?

For color scheme development there is also https://packagecontrol.io/packages/CSScheme (which is more suitable), but it does not convert existing schemes to the CSScheme format yet.

And finally, since 6to5 is not your plugin I would usually ask the developer before we make a change, but since your tags are resonable I will just mention instead. @zertosh

zertosh:
hey @voronianski, I'm cool with adding more labels, but they seem to case-sensitive: [javascript](https://packagecontrol.io/browse/labels/javascript) vs [JavaScript](https://packagecontrol.io/browse/labels/JavaScript), and there are more under javascript (all-lower-case). As for `ECMAScript6`, that's just noise - besides that's not strictly true since we're adding ES7 and Flow definitions (hopefully soon).

FichteFoll:
cc @wbond, imo labels should not be case-sensitive, or should they?

zertosh:
I agree with @FichteFoll - I always assumed labels were case-insensitive. I only tried lowercase "javascript" because there are only 4 results for "JavaScript".

wbond:
Labels are case sensitive, because the user interface for labels is URLs, and URLs are case sensitive unless you write code to do redirects and stuff like that.

voronianski:
@FichteFoll 

> Using branches for releases is deprecated and won't be accepted for new packages anymore: https://packagecontrol.io/docs/submitting_a_package#Step_4

switched to tags - https://github.com/wbond/package_control_channel/pull/4075/files#diff-36e92b7c4e2886182c4935d82f760448R97

> I would also prefer if you used the term "Color Scheme" more deliberate instead of "Theme" because that is incorrect ST terminology.

It actually uses "Color Scheme" term now - https://github.com/wbond/package_control_channel/pull/4075/files#diff-36e92b7c4e2886182c4935d82f760448R91

@zertosh, I've changed label `JavaScript` to `javascript`, hope this is OK. :smile:

FichteFoll:
>It actually uses "Color Scheme" term now - https://github.com/wbond/package_control_channel/pull/4075/files#diff-36e92b7c4e2886182c4935d82f760448R91

It did that before, actually. I was more referring to the heading in your readme and the image.

zertosh:
thanks @voronianski and @FichteFoll

voronianski:
@FichteFoll thanks :+1:

PR #4078, opened 2015-02-11
add Carmesim color scheme

sergiokopplin:


FichteFoll:
So, what exactly are the differences to the original spacegray theme? They look *very* similar to me, besides a small shift in hue.

I also don't see a reason to include the Dracula scheme since it's [available on package control](https://packagecontrol.io/packages/Dracula%20Color%20Scheme).

sergiokopplin:
It's the same idea of the amCoder Package. Optimized for the Dracula color Scheme.
I can split the dracula from the package, what you think? The user can install separate from the Carmesim.

My intention is to create a theme with the Dracula theme colors. What is a fork from the spacegray, like the amcoder is.

sergiokopplin:
I put the dracula out of the package.

FichteFoll:
>Go to `Preferences -> Color Scheme -> User` and select the `Dracula Color Scheme`.

"User" is wrong here, as indicated by the JSON that follows.

Also, you need to push a new tag with the latest commit in order to mark a new release.

sergiokopplin:
I will fix that! Thank you

sergiokopplin:
Fixed!

PR #4085, opened 2015-02-12
Added a new plugin "Django Starter"

geekpradd:
It's similar to the Flask Starter Plugin except longer and more complicated (plus more useful).
It enables Django Developers to create Django Projects directly from Sublime Text rather than using django_admin.py from the Terminal. It auto creates all python files and Django Folders.

Link: https://github.com/geekpradd/Sublime-Django-Starter

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -528,6 +528,16 @@
 			]
 		},
 		{
+			"name": "Django Starter",
+			"details": "https://github.com/geekpradd/Sublime-Django-Starter",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


geekpradd:
Travis Build is passing. My forked repo can be merged.

geekpradd:
I've added support for Sublime Text 2 in this commit.

FichteFoll:
For `djangoFileCode.py` I can recommend you to take a look at the `textwraps.dedent` function.

Furthermore, I highly suggest you to use the `os.path.join` function instead of hardcoding backslashes for file paths.

geekpradd:
@FichteFoll .. Thanks for the feedback. I've improved the code structure and used os.path.join and textwrap.dedent in the latest version.

PR #4088, opened 2015-02-12
Rename 'Shell Command' to 'ShellCommand'.

markbirbeck:
I'm loading the Shell Command package as a module in my Git Mode package and it has been pointed out by @randy3k (markbirbeck/sublime-text-gitmode#9) that this doesn't work with the space in the name. (I'd been testing with a symbolic link, which hid the problem...dumb of me!)

So this pull request is to rename the command package. I found information about the `previous_names` property in the `example-repository.json` file, so hopefully I've included everything that's necessary.

Thanks.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -890,9 +890,10 @@
 			]
 		},
 		{
-			"name": "Shell Command",
+			"name": "ShellCommand",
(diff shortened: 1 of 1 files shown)


FichteFoll:
FYI, you can use `__import__` or `importlib` to import packages with spaces, but it's definitely less convenient.

PR #4091, opened 2015-02-13
adding new sublime3 plugin "MB Warband API"

oshcherbyna:
<a href=https://en.wikipedia.org/wiki/Mount_%26_Blade:_Warband>MB Warband</a> scripting plugin.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -480,6 +480,18 @@
 			]
 		},
 		{
+			"name": "MB Warband API",
+			"details": "https://bitbucket.org/Shcherbyna/mb-warband-api",
+			"author": "Oleg Shcherbyna",
+			"labels": ["warband", "language syntax", "snippets", "build system", "auto-complete"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}				
+			]
+		},
+		{
 			"name": "MDN Search",
 			"details": "https://github.com/jackfranklin/ST2-MDN-Search",
 			"releases": [


FichteFoll:

I would also suggest to place all the snippets into a sub folder as the root is pretty messy right now. And instead of hard-deleting `.pyc` files you should add them to your `.hgignore`.

oshcherbyna:
- "You have to push valid semantic version tags (with optional leading v) to your repo." --
v1.158 is version of warband_module_system (api). 
v1.158.2 - is the version of plugin. Is mean that plugin support "warband_module_system. v1.158", not 1.166 (<a href=https://www.taleworlds.com/en/Games/Warband/Download>latest available</a>). - Tag v1.158.2 added to repo.

- "to place all the snippets into a sub folder" - thanks for suggest. Done

- "And instead of hard-deleting .pyc files you should add them to your .hgignore"
Hard-deleting *.pyc is a feature that will work on user's localhost in warband_module_system folder after this system building.. not inside plugin/repo dir (<a href=http://forums.taleworlds.com/index.php/topic,320675.0.html>link</a>).

FichteFoll:
> Hard-deleting *.pyc is a feature that will work on user's localhost in warband_module_system folder after this system building.. not inside plugin/repo dir.

You mean [here](https://bitbucket.org/Shcherbyna/mb-warband-api/src/9f8570d71c740dfb78dc2b036ad24ccab4d0ee48/mb_warband_api.py?at=default#cl-285)?

I couldn't find any reference in the code to your "cleaning_pyc", so you can probably remove this.

The reason why I'm asking is that ST3 packages are zipped by default and you can not simply access files or subfolders like you usually would (e.g. when developing a plugin).

oshcherbyna:
Oh, I understood. But how to listen Sublime events? For example: building process finish => start cleaning process... to make conveyer?

oshcherbyna:
"cleaning_pyc" folder is removed from package.

FichteFoll:
For the release notes, refer to https://packagecontrol.io/docs/messaging.

For the cleaning of pyc, I'm not quite sure what exactly you want to do, but calling `subprocess.call()` directly is synchronous, unlike the `exec` command which handles it in a separate thread. Also, if you're running Python you can add the `-B` command line argument to make it not create the binary files.

oshcherbyna:
Ok, all done. Package ready to merging.

2015-02-15 0:16 GMT+02:00, FichteFoll <notifications@github.com>:

PR #4093, opened 2015-02-13
A new plugin 'CoolFormat' added

akof1314:
A Sublime Text plugin for Source Code Formatter.


FichteFoll:
FYI, `min(sel.a, sel.b)` is equivalent to `sel.begin()`,  `for _, ch in enumerate(code):` is equivalent to ` for ch in code:`. I also suggest you to use `os.path.join` instead of hardcoding forward slashes.

PR #4095, opened 2015-02-13
added haxelang (barebones haxe language support)

ix:


FichteFoll:

However, it seems you are already pushing tags, so you just need to change it in the pull request.

Usually we are against duplicated functionalit in the package list, but you are right in that the original Haxe package is quite loaded compared to yours, so I guess that's okay. I suggest you to include the original package's `LICENSE` though.

ix:
alright then, will do

ix:
hmm, it doesn't seem like i can change the branch comparison. i'll open a new pr

ix:
so, like this?

FichteFoll:
Exactly.

cc @clemos btw

PR #4097, opened 2015-02-14
Added LogReader Plugin

deneb42:


FichteFoll:
"LogReader" is a pretty generic name while your package is only meant to be used for ios and android platform logs. I suggest something like "MobileLogReader".

deneb42:
Hello,
You're right. I changed the name to MobileLogReader as you sugested.

PR #4099, opened 2015-02-15
babel is the new 6to5

zertosh:
We've rebranded 6to5 as bebel to better reflect our mission. We moved 6to5/6to5-sublime over to babel but things broke, so we moved 6to5-sublime back to 6to5. I guess once this is merged I'll go ahead and migrate the repo over. cc: @FichteFoll 

https://github.com/babel/

### DIFF
--- a/repository/0-9.json
+++ b/repository/0-9.json
@@ -65,17 +65,6 @@
 					"branch": "master"
 				}
 			]
-		},
-		{
-			"name": "6to5",
-			"details": "https://github.com/6to5/6to5-sublime",
-			"labels": ["language syntax", "snippets", "javascript"],
-			"releases": [
-				{
-					"sublime_text": "*",
-					"tags": true
-				}
-			]
 		}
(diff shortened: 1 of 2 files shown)


FichteFoll:

zertosh:
Thanks! totally missed https://packagecontrol.io/docs/renaming_a_package.

zertosh:
@FichteFoll I'm good to go on my side. So from the user's perspective, because of the "previous_names", the transition should be seamless?

FichteFoll:
Yes.

FichteFoll:
You might want to think of an update message about the name change. See https://packagecontrol.io/docs/messaging

FichteFoll:
And, by the way, I suggest you to check out https://packagecontrol.io/package/CSScheme as it's way superior to YAML-tmTheme, imo.

zertosh:
Thanks @FichteFoll!

PR #4100, opened 2015-02-15
add HTML2CSS

jaxgeller:
add https://github.com/jaxgeller/sublime-html-to-css

Select html with cursor, run from command palette and css class + id selectors get parsed from the html and pasted into your clipboard.

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -749,6 +749,16 @@
 			]
 		},
 		{
+			"name": "HTML2CSS",
+			"details": "https://github.com/jaxgeller/sublime-html-to-css",
+			"releases": [
+				{
+					"sublime_text": ">3000",
+					"branch": "master"
+				}
+			]
+		},
+		{
 			"name": "HTML2Haml",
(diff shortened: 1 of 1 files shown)


FichteFoll:
https://packagecontrol.io/packages/CSS%20Primer
https://packagecontrol.io/packages/eCSStractor
?

jaxgeller:
These are good, but I like having it save straight to the clipboard, instead of overwriting files. Also, this works by selecting the content you need, not just the whole file.  But thanks for bringing this up, i appreciate it.

FichteFoll:
Do you think it would make sense to add that kind of functionality to an existing plugin like @hudochenkov's?

hudochenkov:
@jaxgeller, in eCSStractor you can select part you need and only selection will be processed.

Your plugin doesn't work with multiple classes.

I can add “copy to clipboard” option to eCSStractor if you [open issue](https://github.com/hudochenkov/ecsstractor/issues).

jaxgeller:
Going to close this with https://github.com/hudochenkov/ecsstractor/issues/6 . Thanks everyone.

PR #4103, opened 2015-02-16
Added sublime-xpath plugin.

rosshadden:
Added plugin for seeing/copying the current static `xpath` of the `XML` nodes the cursors are on.


FichteFoll:

rosshadden:
@FichteFoll Okay, I used tags instead.

Thanks.

PR #4106, opened 2015-02-17
Add mocha chai coffee snippets

sqrtofsaturn:


FichteFoll:
The path that you hardcoded in your plugin file to set the syntax onload differs from the package name you specified here ("Mocha Chai Coffee Snippets" vs "Mocha CoffeeScript"). You need to adjust that, otherwise the code won't work.

FichteFoll:
Also, what is the exact difference to https://packagecontrol.io/packages/Mocha%20Coffee%20Snippets ?

royvandewater:
I fixed the package name in both projects to be "Mocha Chai CoffeeScript". "Mocha Coffee Snippets" package provides snippets only for Mocha w/CoffeeScript. The snippets in "Mocha Chai CoffeeScript" include a bunch of snippets and syntax highlighting for things that come with Chair, a 3rd party expectations library. I believe many people use Mocha without Chai, and would therefore not want those snippets cluttering their autocomplete.

FichteFoll:
Still the same problem ([1](https://github.com/octoblu/sublime-text-mocha-coffeescript/blob/5340d64bb0e19b3022e8566879a050152b674fce/MochaCoffeeScriptDetectFileType.py#L14), [2](https://github.com/wbond/package_control_channel/pull/4106/files#diff-a4bc313e2f8753f02db1486ed9843aceR879)).

royvandewater:
Whoops, you're right. I should've been more careful. They should match now.

PR #4111, opened 2015-02-18
Scilab plugin added

holgern:
I added my Scilab syntax plugin for sublime 3


FichteFoll:
Since you don't provide any plugins there is no reason to not enable this package for ST2. Please adjust the version selector.

holgern:
I adjusted the version, is everything correct now?

FichteFoll:
Yes.

PR #4114, opened 2015-02-19
Renpy Python - A Renpy .tmLanguage language definition

koroshiya:
Renpy is software used to create Visual Novel video games.
The Renpy scripting language is based around python, with its own keywords and syntax outside of python blocks.

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -616,6 +616,16 @@
 			]
 		},
 		{
+			"name": "Renpy Language",
+			"details": "https://github.com/koroshiya/Sublime-Renpy",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "master"
+				}
+			]
+		},
+		{
 			"name": "Repeat Macro",
 			"details": "https://github.com/sivakumar-kailasam/Repeat-Macro",
 			"releases": [
--- a/repository/r.json
+++ b/repository/r.json
@@ -621,7 +621,6 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"branch": "master",
 					"tags": true
 				}
 			]
--- a/repository/r.json
+++ b/repository/r.json
@@ -621,7 +621,8 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"branch": "master"
+					"branch": "master",
+					"tags": true
 				}
 			]
 		},


FichteFoll:

`.no-sublime-package` and `packages.json` files are redundant, please remove them.
-

Regarding your package:

[I don't think this is how copyright works](https://github.com/koroshiya/Sublime-Renpy/commit/04e910dc18caa57672fa14787c1a347f1b694ea1#diff-9879d6db96fd29134fc802214163b95aR3)

Furthermore, you still need to rename the file and adjuste its [name](https://github.com/koroshiya/Sublime-Renpy/blob/04e910dc18caa57672fa14787c1a347f1b694ea1/PythonImproved.tmLanguage#L32), as well as base scope and fileTypes.

And finally, I actually think that you would get away with just defining your own renpy rules and then importing Python Improved (`import: source.python`), which will fortunately automatically fall back to the default Python syntax if the Improved package is not installed. That has the advantage to get updated automatically and does not require you to merge changes to PythonImproved.

koroshiya:
Thanks FichteFoll,
I've implemented your suggestions and rewritten the project from scratch.
Could you please take another look? The GitHub URL hasn't changed (deleted and started with the same name).

FichteFoll:
Alright, looks good.

However, you instructions in your install message are wrong since it will be `View -> Syntax -> Open all with current extension as... -> Renpy Language` (from package name or syntax name, not sure right now but they are the same anyway).

And you have to push a tag to your repo and refer to it by specifying `"tags": true` here in this pull request, not in your `packages.json` since that won't be looked at.

koroshiya:
Oops, right you are.
I've now changed the name in install.txt to "Renpy Language" and marked the release as 1.0.

FichteFoll:
Getting close but still something to go.

Tags but be [semantic versions](http://semver.org), so 1.0.0 would be appropriate. Also this pull request still points to your master branch.

koroshiya:
Okay, fingers crossed: https://github.com/wbond/package_control_channel/pull/4118

FichteFoll:
Nothing changed?

By the way, you can append commits to existing pull requests by pushing onto the branch you opened it from, so let's stay on this one.

koroshiya:
Hm, this is really testing my understanding of GitHub.
What I changed was the sublime-renpy repository's versioning and my package_control_channel fork's branch (in the other PR).
Did you mean I need semantic tags for package_control_channel and a different branch for sublime-renpy?

FichteFoll:
The tag is fine now, but you need to add `"tags": true` in *this* pull request by making the change in your package_control_channel fork on the `master` branch.

koroshiya:
Done.

FichteFoll:
Well, I think I wasn't clear enough but you also have to remove the branch key since you can obviously only specify one release pattern (tags or branch).

koroshiya:
Ah... gotcha.
This time for sure!

FichteFoll:
indeed

PR #4115, opened 2015-02-19
Append " - ES6 & JSX Syntax" to Babel

zertosh:
Otherwise people can't find the package when they search for ES6 or JSX :disappointed:

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -13,8 +13,8 @@
 			]
 		},
 		{
-			"name": "Babel",
-			"previous_names": ["6to5"],
+			"name": "Babel - ES6 & JSX Syntax",
+			"previous_names": ["6to5", "Babel"],
 			"details": "https://github.com/babel/babel-sublime",
 			"labels": ["language syntax", "snippets", "javascript"],
 			"releases": [


FichteFoll:
Can they not? Description is scanned as well, but only on the website

zertosh:
@FichteFoll from Package Control's "Install Package" palette menu. It seems that it only searches titles and not descriptions.

wbond:
This is not what package names are for.

zertosh:
I agree @wbond. It sucks to have to change the name just to be discoverable. But opening up "Package Control: Install Package" and searching is a common package discovery pattern.  Without this change, anyone typing "JSX" will not see us, despite having the best JSX extensions for Sublime. Had I known that the fuzzy matching only looked at names and not descriptions, I would've called the package something else from the start.

wbond:
Hmm, so perhaps this should be two different packages then? One `Babel` (since 6to5 was renamed specifically since it is NOT just ES6/JSX anymore), one `ES6-JSX`?

zertosh:
@wbond We're not going to break up the different types of syntax definitions by type, but instead add them here as they get developed. The next thing on my agenda is to add [Flow type](http://flowtype.org/) support - that's still 2-3 months away. When that happens, they package should then be called "Babel - ES6 & JSX & Flow Syntax".

My Python is pretty bad, but is fuzzy matching on the description too a feature that's possible in Package Control? Or is that a Sublime limitation?

wbond:
It is a Sublime Text limitation. A package named Babel - ES6 & JSX & Flow is kind of ridiculous. Package names are not for descriptions. I'm sorry users won't find the package by typing in words that describe it. That is what the website is for.

Anyway, aren't jsx and flow react things? Maybe an ES6 package for js developers and a react package?

zertosh:
Yeah that sucks. I would love to keep the name as simple as possible, but I also recognize that people have to find it. Anything I can do to change your mind?

There is already a great ES6-only package: https://github.com/Benvie/JavaScriptNext.tmLanguage. The JSX one provided by Facebook is not very good (see these [screenshots](https://github.com/babel/babel-sublime#screenshots) for a comparison). And there is no Flow support from anyone yet.

FichteFoll:
So yeah, package names really shouldn't be used as descriptions for the package. That's what the description is for.

It's too bad that it hinders discoverability of packages from inside, but that is not suggested anyway because you get to see a lot more on the website anyway, such as the readme. Instead of working around the issue we should encourage people to use the website for discovering packages because it's easier to customize by us and not limited by ST's API (see also https://github.com/SublimeTextIssues/Core/issues/688).

PR #4117, opened 2015-02-19
Deselect last selection package

Loamhoof:
Add a ST3 plugin to deselect the user's last selection (like with a `ctrl+d`).

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -312,6 +312,17 @@
 			]
 		},
 		{
+			"name": "DeselectLastSelection",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Good idea.

In fact, it's so good that this is already built in and goes by **Edit → Undo Selection → Undo Move** and the key binding <kbd>Ctrl + U</kbd>.

Loamhoof:
Awww how could I miss that... Sorry for the bother!

PR #4119, opened 2015-02-20
Added BEMJSON snippets

1vank1n:
BEMJSON snippets for Sublime
Write BEMSJON faster.

This snippets help write BEMJSON (file by [BEM](http://bem.info) methodology) faster.

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -275,6 +275,17 @@
 			]
 		},
 		{
+			"name": "BEMJSON snippets",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Using branches for releases is deprecated and won't be accepted for new packages anymore: https://packagecontrol.io/docs/submitting_a_package#Step_4. Since you have pushed a tag already, this shouldn't be much of an issue.

1vank1n:
@FichteFoll, fixed.

PR #4120, opened 2015-02-20
Convert HTML to SLIM

parterburn:


FichteFoll:

Your tags must be valid [semantic versions](http://semver.org), that means `1.0.0`.

parterburn:
@FichteFoll makes sense - updated!

PR #4121, opened 2015-02-20
Added package "Nbspr"

ssddanbrown:


FichteFoll:
Your tags must be valid [semantic versions](http://semver.org) (with optional leading `v`), so `1.0.0`.

ssddanbrown:
My apologies, I have changed the current tag to fit with semantic versions.

PR #4124, opened 2015-02-20
added scandir as a dependency

stdavis:


FichteFoll:
Dependencies, too, need to be [sorted alphabetically](https://travis-ci.org/wbond/package_control_channel/builds/51584555#L113).

Anyway, the repo you referenced won't work as a dependency. Please read up on the [dependencies docs](https://packagecontrol.io/docs/dependencies) and refer to the already existing ones.

stdavis:
Wow. Did not catch the subfolder requirements the first time around. My apologies. I guess that I was assuming that it would be a much simpler situation. Now I understand why no plugin devs are using the dependency management other than you and @wbond.

Thanks for your time.

wbond:
Dependencies were designed primarily for compiled Python modules, hence they need to be split by ST version, platform and architecture.

Dependencies are not like pip at all, and need to be set up with Sublime Text's architure in mind.

stdavis:
I can see that now. I think that's a shame that I can't easily pull in
python dependencies in a similar manner to pip. However, I'm sure that I
don't understand all of the implications of that. Just wanted to give you
some feedback.

Packages Control is an amazing system that I use everyday and I thank you
for your work on it.

On Mon, Feb 23, 2015 at 7:12 AM, Will Bond <notifications@github.com> wrote:

> Dependencies were designed primarily for compiled Python modules, hence
> they need to be split by ST version, platform and architecture.
>
> Dependencies are not like pip at all, and need to be set up with Sublime
> Text's architure in mind.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/wbond/package_control_channel/pull/4124#issuecomment-75546808>
> .
>



-- 
*Scott Davis*
Utah AGRC <http://gis.utah.gov/developer/>
stdavis@utah.gov
@ScottAGRC <https://twitter.com/ScottAGRC>

PR #4126, opened 2015-02-21
"Adding Package Info For Color To Color Package"

elbatico1:
Thank you for your time and great service.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1472,6 +1472,18 @@
 			]
 		},
 		{
+			"name": "Color To Color",
+			"details": "https://github.com/elbatico1/ColorToColor",
+			"labels": ["color", "color conversion", "css color"],
+			"author": "Fabio Fantini",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Colorcoder",
 			"details": "https://github.com/vprimachenko/Sublime-Colorcoder",
 			"releases": [


FichteFoll:
Please consider adding this very similar functionality to https://packagecontrol.io/packages/CSS%20Color%20Converter or getting in touch with @TheDutchCoder, particularly color names and hsv.

I do not see any particular reason to have two of these plugins, in addition to the others (which are totally redundant as well but have existed beforehand):
https://packagecontrol.io/packages/Hex-to-RGBA
https://packagecontrol.io/packages/Hex%20to%20RGB%20Converter
https://packagecontrol.io/packages/Hex%20to%20HSL%20Color%20Converter

elbatico1:
Hi, first of all thank you for your time!
I think that you should consider to add this package because it's very 
simple and in one shortcut provide all the possible color value with a 
popup window.
I understand that there are several similar packages but, for example, 
as you listed, hex to rgb or hex to rgba, hex to hsl and so on, are 
multiple package that do the same thing as mine, one package.(maybe 
consider to eliminate some very old package?? - last commit was 3 years ago)
Also css color converter do the same but with multiple action required, 
color highlighter (very good one) provide a box around the value, 
underline etc.. very good but sometime messy.
What I'm trying to do is provide a very simple and straight forward tool.
That's it, I made my case, I hope that you'll reconsider and, thanking 
you for your valuable work, I promise that this will be my first and 
last attempt to change your mind.
Thank again, keep up the good work,
Fabio
Il 22/02/15 01:46, FichteFoll ha scritto:

FichteFoll:
The latest three links are irrelevant, I just wanted to state that we have even more of those.

A single package for many conversions exists already and its the first link in my comment, https://packagecontrol.io/packages/CSS%20Color%20Converter.

TheDutchCoder:
In all honesty, I haven't had a lot of time to work on this plugin anymore and I'm not actively using it myself because of my full-time switch to Sass, where I simply don't need it.

Feel free to fork my plugin or to collaborate on it though.

FichteFoll:
In that case, @TheDutchCoder, how about @elbatico1 taking over your package on the list by replacing it with his?

We'd only need to have a seamless replace by defining the same key bindings or by using an update message using https://packagecontrol.io/docs/messaging that notifies users of the takeover.

elbatico1:
Hi @TheDutchCoder <https://github.com/TheDutchCoder>, as suggested by 
FichteFoll, if you agree, I could take over and replace your package 
with mine.
Maybe keeping the name and replacing the package link in "details" and 
using an update message because super+shift+c is in conflict with all 
the other package that work on colors.
Let me know if you don't mind.
Thank you for your time and work,
Fabio
Il 22/02/15 14:48, FichteFoll ha scritto:

FichteFoll:
@TheDutchCoder, @elbatico1, has this been settled?

PR #4127, opened 2015-02-22
Update d.json

Proyecto513:
Remove Django Click Package


FichteFoll:
Please update your fork (by rebasing).

Proyecto513:
I will update, but the matter is that I'm the new mantainer and I can't continue with the development as a separated package, I integrated it into another Django Package I'm developing

FichteFoll:
What about the [5k users](https://packagecontrol.io/packages/Django%20Click) of this package, don't you think they would be interested in an upgrade to the other package or at least a notice about it?

Proyecto513:
Yeah, actually that makes more sense, I'll work on it

Proyecto513:
Hello :smile:

FichteFoll:
Uhm, I'm a bit confused.

So, I take it that Django-Click has been incorporated into DjangoCommands by @vladimirnani. Then you forked that repo and created a separate branch for the old "django click" package and made some changes. What is the reason why this can't be done on the Django Commands package? Also, I would encourage you to upload tags and use the tag versioning scheme instead of pointing at branches.

Proyecto513:
The only difference

Proyecto513:
Hello, I've been thinking about the solutions you proposed and I think that  maybe some anouncement about the other package where Django click is already impremented and leting those 5k users to decide if they want to use it or not, and before ask why i modified the entry on django commands I'm contributing to that package and I think that changing to tags is important there too :smile: good night || evening || morning (apologize for my english)

FichteFoll:
Good call.

PR #4129, opened 2015-02-22
Added Commando package.

ericpridham:


geekpradd:
There is a slight error I suppose. In the releases object, the value for the key `sublime_text` should be `>=3000` not `>3000`.

FichteFoll:
Indeed. While it's not a huge issue, consistency with the other packages would be nice.

ericpridham:
Thanks for the tip!

PR #4000, opened 2015-01-18
Add package GraveSkip

zzh8829:
https://github.com/zzh8829/GraveSkip

PR #4131, opened 2015-02-23
Add package GraveSkip

zzh8829:
Add package GraveSkip

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -1102,6 +1102,16 @@
 			]
 		},
 		{
+			"name": "GraveSkip",
+			"details": "https://github.com/zzh8829/GraveSkip",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Sorry for the delay in #4000, but I work based on notifications and sadly github does not send any when commits are pushed (and don't intend to do in the future, I asked :< ). Thus, I depend on people commenting when they fix the requests - or they need to wait until I re-visit old pull request approximately once a month.

FichteFoll:
Huh, and in all that mess I forgot to check your tags. You have to push valid [semantic version](http://semver.org) tags (with optional leading `v`) to your repo, otherwise no release will be found.

PR #4139, opened 2015-02-24
Theme Ksix, a dark theme and color scheme.

matiasmoya:


FichteFoll:
I *highly* recommend you to add screenshots to your readme because that is the only real thing you can use as a user to judge the theme or color scheme before you install it.

matiasmoya:
Done! thanks a lot :)

PR #4142, opened 2015-02-25
Add ToggleGroup plugin

daytonn:


FichteFoll:
To me this sounds like is has many overlapping features with https://github.com/SublimeText/Origami, except that Origami works with more panes and is direction-driven and not just toggling. Do you know that package?

If you especially want a function to toggle between two (or maybe more; didn't check code) groups then you might want to support https://github.com/SublimeText/Origami/issues/29 since I feel like Oriigami is really the place where this would be useful. If you think that they solve very different problems and use a different enough method to do so, then I would probably indulge.

daytonn:
@FichteFoll I hadn't heard of Origami but I think this deserves it's own plugin. This is a simple toggle command for those who DO NOT want to focus on where and how and only want to toggle between groups (I use this mainly for toggling left and right groups). Origami seems to be quite full-featured and defaults to two-stage key bindings which is a deal breaker for me. The philosophy between these two plugins could not be further apart, which is probably why https://github.com/SublimeText/Origami/issues/29 exists.

daytonn:
Also this will toggle through more than 2 panes

FichteFoll:
Okay, your choice then. I don't exactly agree on the philosophy not possibly be further apart, but well. You can also customize all key bindings in ST and hiding bindings with key chords or not defining any at all reduces the risk of creating conflicts with other packages.

However, I have a concern regarding terminology. Toggling only works with two states and if this also works with multiple groups (which it does, the source is practically the same as the one I proposed in the referenced issue) "cycling" is more fitting.

daytonn:
I suppose cycling is more appropriate. I just think of it as toggling because I use it to go from left to right. I may have to play a bit with Origami but from a cursory glance it does not seem to suit me. I do not use arrow keys, two-stage key bindings are the devil. If I did use Origami, I would be remapping all the things so I could get to the base functionality of cycling tab focus/move. The point is I don't want to care about where things are. This plugin always means the other one. The fact that it cycles is basically a free feature of the implementation.

daytonn:
I renamed the plugin to CycleGroup

PR #4144, opened 2015-02-26
Change ScrollOffset and MouseEventListener to let them be installed on ST3

vihangm:
Plugin to provide vim style scroll off behavior.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -1211,7 +1211,7 @@
 			"details": "https://github.com/SublimeText/MouseEventListener",
 			"releases": [
 				{
-					"sublime_text": "<3000",
+					"sublime_text": "*",
 					"branch": "master"
 				}
 			]
--- a/repository/s.json
+++ b/repository/s.json
@@ -401,6 +401,16 @@
 			]
 		},
 		{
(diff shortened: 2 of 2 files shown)


FichteFoll:
There is already a very similar package like this: https://github.com/SublimeText/ScrollOffset
It's currently marked as ST2-only but I didn't find anything in the source that would prevent it from working on ST3 so it could just be updated.

Would it be possible to add possibly missing functionality to that package?

cc @adzenith

vihangm:
Yeah, I'll test it out and see if there are any egregious issues with ST3. I noticed it while sending out this PR, but didn't have the time to test it.

adzenith:
I'm pretty certain I was using it for a long time on ST3. Not sure why it's marked ST2--I probably just forgot to update it when ST3 came out :P

vihangm:
Depends on https://github.com/SublimeText/ScrollOffset/pull/11 and https://github.com/SublimeText/MouseEventListener/pull/1 being merged.

vihangm:
The PRs are merged. This should be good to go now.

FichteFoll:
Great!

PR #4145, opened 2015-02-26
renamed package

idleberg:
As mentioned [here](https://github.com/wbond/package_control/issues/890)

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -782,10 +782,10 @@
 			]
 		},
 		{
-			"name": "AppleScript",
+			"name": "AppleScript Extensions",
 			"details": "https://github.com/idleberg/AppleScript-Sublime-Text",
 			"labels": ["auto-complete", "build system", "snippets"],
-			"previous_names": ["AppleScript Completions and Snippets"],
+			"previous_names": ["AppleScript", "AppleScript Completions and Snippets"],
 			"releases": [
 				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
Since you can not remove the package using the Remove Package command I need to check later if renaming will work before merging.

FichteFoll:
Okay, I scanned the source code and there shouldn't be any issues with renaming this package on ST3. I also tested it locally and had almost no problems. On ST2 there likely *would* be two issues, but luckily AppleScript is not included in ST2 by default.

For some reason both packages, previous and new name, end up as ignored and in the "in_process_packages" list despite the events being cleared. I'm still investigating here, but it is fixed by a simple restart of ST.

The only problem right now is that any customizations from other people in the AppleScript package folder would be recognized as a package rename since PC does not check if a package is actually under PC control before renaming (I think this is a bug? @wbond). Since this is *very* likely not going to be the case because it rarely happens, we can savely do the rename. You might want to remove the previous_names entry once the rename is likely done for everyone in a few months however, just so that we are clear.

FichteFoll:
Okay, apparently ST2 **does** ship the AppleScript package, it is however removed from the Pristine Packages once installed (which is why assumed it did not).

I believe this is Package Control's work and since it happens on install, the damage has already been done anyway.

PR #4146, opened 2015-02-27
added DocPy-a docstring python autocomplete tool

RicherMans:
DocPy is a simple python function autocomplete tool, which helps writing function documentation.

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -635,10 +635,11 @@
 		{
 			"name": "DocPy",
 			"details": "https://github.com/RicherMans/Sublime3-pydoc",
+			"labels":["auto-complete","formatting","snippets"],
 			"releases": [
 				{
 					"sublime_text": ">=3000",
-					"branch": "master"
+					"tags":true
 				}
 			]
 		},
--- a/repository/d.json
+++ b/repository/d.json
@@ -633,6 +633,16 @@
 			]
 		},
 		{
+			"name": "DocPy",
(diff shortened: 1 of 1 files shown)


geekpradd:
Great Plugin.. I recently started writing docstrings in Python and it looks helpful. ST2 support would be awesome (I think i can help, but I'll have to grasp your code first)

FichteFoll:

I didn't spot anything that would be incompatible with ST2 (but no guarantees). However, @geekpradd, you should really consider upgrading to ST3 since it's a lot better.

geekpradd:
Already on ST3 @FichteFoll .. But I always like backporting stuff (especially Python 3 to Python 2). And I have ST2 installed that I use sometimes. 

And many people who bought ST2 haven't purchased ST3 (like me) because it's a beta and so ST2 still lurks in our machines.

RicherMans:
Well actually there is something not working with ST2, which is in line 121 of the plugin
```python
functionname, *args = funcdefargs
```
I have never worked with python3, but I just liked that feature so I simply implemented it. Havent even thought of using the whole plugin for ST2, but will do then.

FichteFoll:
Okay, but you need to edit this PR so that it points to your releases.

RicherMans:
Is it enough to commit the release into the current pull request or do I only need to push one commit which points at the release?

FichteFoll:
I don't exactly understand, but what you did is fine.

PR #4147, opened 2015-02-28
New plugin: Python Fix Imports

Stibbons:
Website: https://github.com/Stibbons/python-fiximports
Release: https://github.com/Stibbons/python-fiximports/releases/tag/1.0.0
For Sublime Text > 3

Signed-off-by: Gaetan SEMET <gaetan@xeberon.net>

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -1687,7 +1687,7 @@
 			"details": "https://github.com/Stibbons/python-fiximports",
 			"releases": [
 				{
-					"sublime_text": ">3000",
+					"sublime_text": ">=3000",
 					"tags": true
 				}
 			]
--- a/repository/p.json
+++ b/repository/p.json
@@ -1683,6 +1683,16 @@
 			]
 		},
 		{
+			"name": "Python Fix Imports",
+			"details": "https://github.com/Stibbons/python-fiximports",
+			"releases": [
+				{
+					"sublime_text": ">3000",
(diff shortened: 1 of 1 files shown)


FichteFoll:
For consistency, please use `>=3000` for the version selector.

I don't see the `MergeUtils` module provided in your package, which you import in `python_fiximports.py`.

I also don't see a reason for your `.no-sublime-package` file. Please use it only if it does not work otherwise.

And finally, since ST scans all Python files in the package root for plugins (i.e. subclasses of the API interface from `sublime_plugin`) I recommend you to place modules in subdirectories, e.g. move `fiximports.py` into `fiximports/__init__.py`.

Stibbons:
Isn't .no-sublime-package required for importing libs from inside my plugin?

MergeUtils comes from another plugin. I reup the 1.0.0 version

FichteFoll:
No, you can import stuff inside packages just fine, especially when using relative imports. Only when some libraries are using absolute imports you either have to add their parent folder to `sys.path`, and even then I think that ST is able to find those while inside an archived plugin. I'm not sure though; or you rewrite the library to use relative imports everywhere.

`.no-sublime-package` is mostly needed when you need to call stuff like binaries or the machine's python interpreter with a bundled script. Or bash scripts and the like.

Stibbons:
Thanks a lot

PR #4149, opened 2015-03-01
New Snippet: Placehold.it

bartholomej:
Website: https://github.com/bartholomej/sublime-placeholdit
Release: https://github.com/bartholomej/sublime-placeholdit/releases/tag/v1.0.0
For Sublime Text 2 and 3

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -897,21 +897,22 @@
 			]
 		},
 		{
+			"name": "Placehold-it Image Snippet",
+			"details": "https://github.com/bartholomej/sublime-placeholdit",
+			"labels": ["Placeholder", "placehold.it", "tag", "image"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Placehold.it Image Tag Generator",
 			"details": "https://github.com/SubZane/Sublime-Placehold.it",
 			"releases": [
 				{
 					"sublime_text": "<3000",
-					"branch": "master"
-				}
-			]
-		},
-		{
-			"name": "Placehold.it Snippet",
-			"details": "https://github.com/bartholomej/sublime-placeholdit",
-			"releases": [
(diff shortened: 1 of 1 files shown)


FichteFoll:

Please also consider not using dots in package names since they break Python plugins completely. You don't define any currently, but better be defensive here.

bartholomej:
Thank you for your quick feedback @FichteFoll and sorry for my mistake.
I know using branches is deprecated and I was prepared to use tags. Just typing error in definition ;)

Good point with python *dot problem* in the name. It's pity that I had to give up the real name of service (placehold.it)... but I just replaced *dot* with *dash*.

I hope it's OK ;)

FichteFoll:
Interestingly your latest commit made me aware of the existance of [the other placehold.it package](https://packagecontrol.io/packages/Placehold.it%20Image%20Tag%20Generator). Isn't that more flexible than your snippet? Or did you just not know about it because it is not marked for ST3 (and wouldn't work with it either since it contains the evil dot, but could probably be fixed very easily). Maybe you can help @SubZane a bit and update the package with missing functionality or take it over completely in case he doesn't want to maintain it anymore.

SubZane:
If anyone wants to take over maintenance of Placehold.​it Image Tag Generator I'd be happy since I no longer have the time to support it :)

bartholomej:
OK guys, you're right. This will be better ;)
@SubZane If you don't want to maintain it anymore, can I ask you for transfering ownership or something like that? I think you've moved over to Atom.io, right? ;) And also this will be better for future maintenace.
Firstly we want to fix it for ST3 and also we're planing some new features ;)
Thank you for your work, you did a well job ;) :+1:

SubZane:
@bartholomej I cannot transfer ownership because you have a fork and it's conflicting. I added you as a collaborator in the meantime :)

FichteFoll:
All right, I'm closing this issue then.

PR #4151, opened 2015-03-01
Dart: Update source for releases

guillermooo:
New versions of the Dart package will be released from a release-only repo.

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -90,7 +90,7 @@
 		},
 		{
 			"name": "Dart",
-			"details": "https://github.com/guillermooo/dart-sublime-bundle",
+			"details": "https://github.com/guillermooo/dart-sublime-bundle-releases",
 			"releases": [
 				{
 					"sublime_text": ">=3000",


FichteFoll:
Why?

guillermooo:
To clean up the repo from dev files. Is there a better way?

FichteFoll:
Hm. There are indeed situations where you would prefer to not include certain files in the release but still keep them in the repo for developing purposes. Because of how the packages are bundled (PC just downloads the zipball of the tag) a `.pcignore` feature like git's won't help much since the files are being downloaded anyway.

I suppose there are two ways for this, one is the separate repo that you used and the other is a separate branch that will at least keep the relevant "branches" contextually close in the same repo.  Tagging from a different branch is trivial and you could prefix them as `release-v1.5.3` or similar using the new PC feature. (I personally would prefer the branch way, similar to the "gh-pages" branch that github uses for the website representation.)

Since there is no clear winner here and both ways have their (dis-)advantages, this is your decision to take. Just hit me up with what you decided on. I was just curious.

guillermooo:
> I suppose there are two ways for this, one is the separate repo that you used and the other is a separate branch that will at least keep the relevant "branches" contextually close in the same repo. Tagging from a different branch is trivial and you could prefix them as release-v1.5.3 or similar using the new PC feature. (I personally would prefer the branch way, similar to the "gh-pages" branch that github uses for the website representation.)

That's correct, but I personally prefer not to use branches in my master repository. Really, it's just a matter of preference.

> Since there is no clear winner here and both ways have their (dis-)advantages, this is your decision to take. Just hit me up with what you decided on. I was just curious.

I'll go with the separate repo, then :)

Thanks!

PR #4152, opened 2015-03-01
added request-oauthlib and oauthlib

csch0:


csch0:
will change links first..

FichteFoll:
By the way, I changed the pypi badge in the table to the shields.io one because it matchs the other one better (and also includes the leading `v`)

img for reference

PR #4153, opened 2015-03-01
Add Promela Spin.

corbanmailloux:


FichteFoll:
Instead of requiring the user to install jspin to a certain directory you can require him to add the `jspin\bin` folder to PATH and also provide a default path extension in your build system like so:
```js
{
  "cmd": ["spin.exe", "-X",  "-u250",  "$file"],
  "path": "$path;C:\\jspin\\bin"
}
```

corbanmailloux:
Excellent. I've updated the code and made a [new tag](https://github.com/corbanmailloux/sublime-promela-spin/releases/tag/v1.0.1).

PR #4154, opened 2015-03-01
Add LogView package

FlashSystems:
Added the LogView package.
Added labels to the LDIF-Syntax package.


FichteFoll:
Do you use any specific API that's only available starting 3065? Because otherwise I suggest you to change that to 3000 for consistency with the other packages. You can change it later if you do end up using API that was added later, like the new `view.show_popup` API.

FlashSystems:
The minimum version required is 3033 because this plugin uses on_text_command and 3033 fixed a bug in on_text_command.
I've tested the plugin with 3065 because this is the last publicly availiable version. It should work with 3033 but it's untested. Should I lower the required version to 3033?

FichteFoll:
I never really thought about the recent additions because my plugins are mostly compatible with ST2 and ST3 (so I need to assure somewhat hug compatability anyway), but this is a valid concern.

PR #4156, opened 2015-03-02
Add MIT Alloy highlighting.

corbanmailloux:


FichteFoll:
FYI, you don't need to escape forward slashes or hypens outside of sets in regular expressions.

corbanmailloux:
Thanks. Just to clarify, you mean:
```
- comment: Line comment
  name: comment.double-slash
  match: \/\/.*$
```

could be replaced with:
```
- comment: Line comment
  name: comment.double-slash
  match: //.*$
```

Correct?

FichteFoll:
Yes.

And `match: \-\-.*$` with `match: --.*$`.

corbanmailloux:
Okay. Thanks for the heads up. I'll change it over.

PR #4159, opened 2015-03-02
Added LPrint.

mbr:


FichteFoll:
You need to resolve the "LPrint" subdirectory, the file should be in the repository root. Python plugins must.

mbr:
Should be fixed in the current master and version 0.2.0.

PR #4161, opened 2015-03-04
Terminal in packages

Proyecto513:
Package addition

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -225,7 +225,7 @@
 					"tags": true
 				}
 			]
-		}
+		},
 		{
 			"name": "Tera Term Language",
 			"details": "https://bitbucket.org/tkyjhr/st_teraterm",
--- a/repository/t.json
+++ b/repository/t.json
@@ -217,6 +217,16 @@
 			]
 		},
 		{
+			"name": "Terminal in Packages",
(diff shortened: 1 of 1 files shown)


FichteFoll:
1. Won't work in ST2 because `print` is a statement there and does not have a value.
2. I'm also a bit confused about `"{}".format(self.command)`, it makes no sense.
3. Instead of running the change directory command inside the terminal it's better to give a working_dir parameter to Popen.
4. It's better to access settings when they are used, not for example when a command is instantiated (as I don't actually know when that happens) because this will assure that the current setting is read live and is not outdated.

Proyecto513:
I'm really amazed by the thought that you read all the code in the packages, but it's nice to know that somebody cares about the quality of them.

I just pushed a commit :smiley:

FichteFoll:
Alright, looks good.

FichteFoll:
... almost.

Your tags must be valid [semantic versions](http://semver.org) (with optional leading `v`). And you also didn't push one for the latest changes.

geekpradd:
You are using global variables in your code. Can't you just pass TERMINAL as a parameter to CommandThread?

Proyecto513:
Just pushed another one

PR #4162, opened 2015-03-04
GDL Support

runxel:
Adds my `GDL-sublime` package which brings support for GDL on Sublime Text.
- syntax
- color scheme
- some preferences
- completions


FichteFoll:
You can remove the "Support" part of the package name, it's redundant.

runxel:
Ok, did so.

PR #4163, opened 2015-03-04
added DummyData plugin

blcook223:
This plugin allows a user to easily create random dummy JSON data within ST. See https://github.com/blcook223/DummyData/blob/master/README.md for more details.

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -916,6 +916,16 @@
 			]
 		},
 		{
+			"name": "DummyData",
+			"details": "https://github.com/blcook223/DummyData",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Duplicate Lines",
 			"details": "https://github.com/wjthomas9/duplicate-lines",
 			"releases": [


FichteFoll:
Looks very similar to https://github.com/titoBouzout/FakeDataGenerator. Did you miss that?

blcook223:
I hadn't seen that. I'll take a look when I have a chance and get back to
you.

Best,

Ben

blcook223:
I've been playing around with FakeDataGenerator. It's definitely a nice plugin with a bigger variety of built-in methods than my plugin, and it does share some functionality with DummyData.

I still think DummyData would be a valuable addition. It better meets the needs of the projects I work on, which are what spurred me to write it. Here's the biggest differences I see:

- FakeDataGenerator is geared toward interactively producing random text (not too different from [Random Everything](https://packagecontrol.io/packages/Random%20Everything)). DummyData is specifically designed to produce dummy JSON files.
- DummyData will always output valid JSON. FakeDataGenerator can easily leave you with commas in places that are invalid for JSON.
- DummyData allows contributors to a project to save data models (e.g., in a project directory) and use them to generate any number of JSON files with new random data with a single command each time. FakeDataGenerator isn't really set up to do this.
- If you want to generate a JSON file with multiple repeated sections, FakeDataGenerator requires you to highlight multiple sections of your file and issue multiple commands. DummyData lets you to do this with one command.
- FakeDataGenerator has no method for random selection among arbitrary JSON data. For example, with FakeDataGenerator, `"{{ random_element(array=({'an': 'object'}, ['an', 'array'])) }}"` produces `"an array"` (a string). With DummyData, `["{% random %}", {"an": "object"}, ["an", "array"]]` produces `["an", "array"]` (an array).

Thoughts?

FichteFoll:
> FakeDataGenerator is geared toward interactively producing random text

To me, the general appraoch sounds more promising as it allows more flexible use without invalidating specific use cases.

> FakeDataGenerator can easily leave you with commas in places that are invalid for JSON.

I never used it, so I can't say how easy or hard it is to produce invalid JSON e.g. with trailing commas. And how much of a hassle it would be to fix. [FYI, since I'm not sure how you do it, but auto-correcting many errors might also be an indicator for poor design.]

> use them to generate any number of JSON files with new random data with a single command each time. 

I'm sure this functionality could be added to FDG as well.

> If you want to generate a JSON file with multiple repeated sections, FakeDataGenerator requires you to highlight multiple sections of your file and issue multiple commands.

Can't comment on this.

> FakeDataGenerator has no method for random selection among arbitrary JSON data.

Your example looks like a bug to me (in FDG)

cc @titoBouzout

-

Since I took the effort yesterday to write down why we prefer not to add duplicated funtionality or very similar packages, I'll use the opportunity and paste that here:

blcook223:
Well, I certainly understand that there are reasons for not duplicating functionality.

I would argue that:

> the packages themselves are different enough in that they provide different use cases and 
> would attract different users. 

The goal of DD is to produce valid JSON files filled with dummy data quickly based on pre-defined data models. FDG is more interactive and more flexible in the sense that you can use it to generate CSV among other things, but my understanding is that it just isn't set up to meet the same goal as DD, the biggest reason being you can't use FDG to randomize/repeat/produce any arbitrary JSON data.

@titoBouzout, I'd definitely be interested to know if you have any thoughts about adding any or all of this functionality to FDG.

FichteFoll:
@titoBouzout ping

titoBouzout:
While the packages have some kind of duplicate behaviour, these well, offer different possibilities .... Im not against similar packages when the effort to merge these packages is considerable.

blcook223:
@FichteFoll did you have any additional thoughts about this? I'd still like to add this as a separate plugin given @titoBouzout 's response. If you feel more should be done in my documentation to distinguish DD from FDG and other packages, let me know.

FichteFoll:
No, this is fine. If nobody is (currently) willing to merge the code bases then there is no reason to stall the addition of this package indefinitely. If you decide to merge at one point we can do that here too any time.

PR #4165, opened 2015-03-05
Adding TXT2PYNB repository to appropriate .json file

jmwerner:
Addition of TXT2PYNB package to the default channel repository

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -1810,6 +1810,17 @@
 			]
 		},
 		{
+			"name": "TXT2PYNB",
+			"details": "https://github.com/jmwerner/TXT2PYNB",
+			"releases": [
+				{
+					"sublime_text": "<3000",
+					"platforms": ["osx", "linux"],
+					"tags": true
+				}
+			]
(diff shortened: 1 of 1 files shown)


FichteFoll:
I have a few issues with how this package is structured.

First, I don't see a reason to run `txt2pynb.py` using the system's Python rather than the built-in Python of Sublime Text itself. It does not import external dependencies, so it can run inside ST just fine.
Second, the way it currently stands, ST will generate a traceback on nearly every start because it attempts to load `txt2pynb.py` but will 1. fail if ST itself was not called with arguments and 2. fail if ST was started with a non-filename parameter because it obiously won't be openable. This is assuming that ST forwards its arguments to `sys.argv`.

jmwerner:
Thank you very much for the feedback, I greatly appreciate it! I will fix these problems as soon as I can and report back.

jmwerner:
I believe I have fixed all of those problems you listed, please let me know if that is not the case. Thanks for your patience!

PR #4167, opened 2015-03-05
added DotNetComments

juzzbott:


FichteFoll:
- Your tags must be valid [semantic versions](http://semver.org) (with optional leading `v`).

- `re.match('.*C#.tmLanguage', syntax_type)` is *semantically* equivalent to `syntax_type.endswith('C#.tmLanguage')`, but you should consider matching scopes instead, using `view.match_selector(0, "source.cs")`.

- `def run(self, edit, args)` is not exactly how it should be used. keyword arguments are expanded when `run` is called, so you should change this to `def run(self, edit, comment_block)` and later `view.run_command('dot_net_summary_comment', {'comment_block': comment_block})`, which gets rid of one redundant dict layer.

juzzbott:
Hi @FichteFoll, 

Thanks for the comments on some of the improvements to the plugin. I've also updated the version info for the tag.

Cheers, 
Justin

PR #4168, opened 2015-03-05
Change the name of the package

jfromaniello:
I think there is a problem with Sublime 3 when the name of the package contains a dot.

And I've always wanted to change from Node.js to CommonJS because that what really is.

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -659,6 +659,17 @@
 			]
 		},
 		{
+			"name": "Require CommonJS Modules Helper",
+			"details": "https://github.com/jfromaniello/sublime-node-require",
+			"previous_names": ["Require Node.js Modules Helper"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
Indeed, dots are not good in package names. I thought it didn't work on ST2 too, though.

May I encourage you to use tags as releasing strategy? https://packagecontrol.io/docs/submitting_a_package#Step_4

jfromaniello:
Sure, thanks

jfromaniello:
done, I created the tag 1.0.0

Sorry for the delay

FichteFoll:
No problem. You need to update the release here to include `"tags": true` instead of the branch line though.

jfromaniello:
@FichteFoll thanks for your all help. 

It is done and I merged all the commits into one.

thanks again

FichteFoll:
Oh, there is unfortunately another thing that I either forgot or you changed in the force-push, but when renaming you have to add the old name to a `previous_names` array.

jfromaniello:
@FichteFoll done

PR #4169, opened 2015-03-05
Added ListOpenFiles extension back

rrg:
This package existed in the repository before, I'm not sure how but it somehow got removed. Adding it back.

http://stackoverflow.com/questions/19665615/sublime-text-text-list-of-open-files/

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -557,17 +557,6 @@
 			]
 		},
 		{
-			"name": "ListOpenFiles",
-			"details": "https://github.com/rrg/ListOpenFiles",
-			"labels": ["File List"],
-			"releases": [
-				{
-					"sublime_text": "*",
-					"branch": "master"
-				}
-			]
-		},		
-		{
 			"name": "lispindent",
 			"details": "https://github.com/odyssomay/sublime-lispindent",
 			"labels": ["text manipulation", "formatting"],
@@ -618,6 +607,17 @@
 				}
 			]
 		},
+		{
+			"name": "ListOpenFiles",
+			"details": "https://github.com/rrg/ListOpenFiles",
+			"labels": ["File List"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": "true"
+				}
+			]
+		},		
 		{
 			"name": "Lithium Snippets",
 			"details": "https://github.com/fitzagard/li3_sublime",
--- a/repository/l.json
+++ b/repository/l.json
@@ -557,6 +557,17 @@
 			]
 		},
 		{
+			"name": "ListOpenFiles",
+			"details": "https://github.com/rrg/ListOpenFiles",
+			"labels": ["File List"],
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
It appears you have found #3357 already, so I'm closing this.

#3330 preceded this btw, for future reference.

FichteFoll:
(for context see #3357)

Your package needs to be [here](https://travis-ci.org/wbond/package_control_channel/builds/53249221#L167) to be in alphabetical order.

Also, using branches for releases is deprecated and won't be accepted for new packages anymore (which your package qualifies as now): https://packagecontrol.io/docs/submitting_a_package#Step_4

rrg:
Ok so what do you suggest I should do? I'm not sure if I understand it correctly.

FichteFoll:
1. Fix order.
2. Push a semantic version tag.
3. Change you release to include `"tags": true` instead of the branch key.

rrg:
Okay, I committed the change to my patch-1 branch, not sure if I need to create a pull request again and how. Can you please take a look?

FichteFoll:
All good.

rrg:
When do I see this show up in package manager? Also can I see all past installs as well? or all that information is now deleted?

Thanks for the help btw!

FichteFoll:
Oh, I totally forgot that you need to remove your package name from the other package's `previous_names` list. Well, I did think about it numerous times but forgot to tell you.

The installs and stats will be lost.

rrg:
How do I do that? What packges whill have my package in the previous_names list and why?

FichteFoll:
I did it for you. Because of #3330 and #3357.

rrg:
Cool, so when can I see my package in the package manager?

FichteFoll:
It should be visible by now. I can't diagnose anything, so @wbond needs to check.

wbond:
There is an error with it, but it isn't showing up since it hasn't yet once been able to fetch the info.

Most likely I'll be adding a "broken packages" page to https://packagecontrol.io today.

FichteFoll:
Ah, I just saw it as well. Sneaky. I suppose I could add a test for it since it is very unlikely that this would happen intentionally.

A broken packages page sounds like a very good idea for cases like these. Will it include "missing" ones too?

wbond:
Yes, it will include missing packages and any URLs that were not properly retrieved or parsed.

wbond:
@rrg You included `"tags": "true"` which will look for tags in the form `true1.0.0`, `true1.2.1`, etc. Instead, it should be `"tags": true`.

rrg:
I see, do I need to be doing the pull request thing again then?

wbond:
No, I just made the fix for you

rrg:
@FichteFoll, @wbond  Although I hate the fact that it was removed and I lost all the stats, thanks a lot for the help.

FichteFoll:
May this be a lesson to enable email notifications and checking them somewhat frequently.

PR #4173, opened 2015-03-06
Rename OCaml package

FichteFoll:
It is currently the same as a default package and that causes issues.

See https://github.com/wbond/package_control/issues/890 and https://github.com/wbond/package_control_channel/pull/4145.

@whitequark, please acknowledge this change.

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -365,6 +365,19 @@
 			]
 		},
 		{
+			"name": "Better OCaml",
+			"details": "https://github.com/whitequark/sublime-better-ocaml",
+			"description": "An improved version of default OCaml package",
+			"labels": ["language syntax"],
+			"previous_names": ["OCaml"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "master"
+				}
+			]
+		},
+		{
 			"name": "Better RSpec",
 			"details": "https://github.com/fnando/better-rspec-for-sublime-text",
 			"labels": ["language syntax", "snippets"],
--- a/repository/o.json
+++ b/repository/o.json
@@ -55,18 +55,6 @@
 			]
 		},
 		{
-			"name": "OCaml",
-			"details": "https://github.com/whitequark/sublime-better-ocaml",
-			"description": "An improved version of default OCaml package",
-			"labels": ["language syntax"],
-			"releases": [
-				{
-					"sublime_text": "*",
-					"branch": "master"
-				}
-			]
-		},
-		{
 			"name": "OCaml Autocompletion",
 			"details": "https://github.com/whitequark/sublime-ocp-index",
 			"releases": [


whitequark:
No. I did it deliberately. I do not remember exactly which issues were there with the package as originally named, since it was almost two years ago, but it was certainly much less usable than in the current form.

I've been using it for quite a while and have not noticed any issues; neither did any of the people I know who installed it. You'd have to be more specific on the issues it causes.

FichteFoll:
The biggest issue is that the original package is deleted on ST2 and can only be restored by obtaining from somewhere or somebody else or reinstallation. It also means that the original package data is removed before the installation. On ST3 the original files are not deleted but simply ignored completely. Since you started with the source of the default package and then iterated over it, there won't be an issue with the latter two, but the first is still there.

Also, the fact that you based your repo on the initial package (which negates the above two issues), raises a copyright concern since there was no license included. We currently do not know about the stance of the default packages regarding copyright and licensing, so I can not even tell you how it stands.

We would prefer if we, as the package control channel maitnainers, didn't need to deal with issues such as these and thus no packages overriding the default packages were allowed. I will need to talk to @wbond if we are willing to make an exception to this or not.

whitequark:
It would be perfectly fine to restrict the OCaml package to ST3. Otherwise, I strongly prefer to keep it as-is.

Copyright is a worse concern, though I do not see how it would affect you as package-control maintainers.

FichteFoll:
cc @wbond

wbond:
I don't see why we can't rename this to `Better OCaml`.

As it stands right now, Package Control has not been designed or tested to deal with overriding built in packages. Because of this, I think the two reasonable options are:

1. Rename this custom OCaml packages to `Better OCaml`
2. Someone volunteers to write tests and fix bugs related to dealing with packages that override built-in packages

So unless we have someone step up to do number 2, I think we are going to take option 1. Unfortunately I don't currently have the time to devote to dealing with package overrides.

I'll post a tweet asking for volunteers who may be interested in contributing to https://github.com/wbond/package_control/issues/890. If we don't have any movement on the issue in the next few weeks, I think we'll move forward with the rename.

whitequark:
I have submitted a PR to restrict the package to ST3: https://github.com/wbond/package_control_channel/pull/4415

After that, I don't think a rename is necessary. You're raising a barrier (and also probably breaking existing installations) while not fixing any problem that actually exists. The package has over 1K installs and no one has ever complained of any breakage.

FichteFoll:
I still don't like the idea of replacing default packages even just for namespacing reasons, but it appears I am the only one with that opinion. Because of that, restricting to ST3 will do for now.

FichteFoll:
Restricting the package to ST3 will do for now and should at least solve the technical difficulties (which is why I merged #4415). Things will be problematic when users uninstall the package however.

Either way, I don't think we should allow installation of packages with the same names as default packages (and thus overriding them) because of namespacing. Similar packags aimed at improving on a default package may coexist at the same time, but only one will receive the benefit "taking over the original package name", and we don't have any policy in regard to that currently.

Ideally there would be some way to modify and fix the shipped packages altogether (including Jon using the updated states for new ST builds) and a team that reviews changes since these packages will be widely adopted, but that still has to go a long way.

https://github.com/SublimeTextIssues/Core/issues/621
http://www.sublimetext.com/forum/viewtopic.php?f=2&t=17892#p67185
http://www.sublimetext.com/forum/viewtopic.php?f=2&t=17892&start=10#p67224

whitequark:
I acknowledge all of your points. I will be happy to see a working process for improving standard packages in place. So I am completely fine with all your decisions for now. (Regarding ST2, the damage is already done, so I guess there is not much choice.)

FichteFoll:
@whitequark FYI, the default packages have been open sourced at https://github.com/sublimehq/Packages quite some time ago and pull requests are regularly reviewed and acepted. Please consider contributing your OCaml enhancements there.

I still believe that there shouldn't be packages with the same name as a default package on Package Control, which is why I'll keep this PR open until the situation is resolved.

whitequark:
@FichteFoll Okay, it is definitely better to have the improvements upstream. I no longer actively use OCaml though so I'll let any interested party do this and then voluntarily remove my package entirely.

evandrocoan:
It is the time we should merge this because:

> 1. #FichteFoll Similar packages aimed at improving on a default package may coexist at the same time, but only one will receive the benefit of "taking over the original package name", and we don't have any policy in regard to that currently.

> 1. #whitequark it is definitely better to have the improvements upstream. I no longer actively use OCaml though so I'll let any interested party do this and then voluntarily remove my package entirely.

> 1. #wbond (a few year ago) I'll post a tweet asking for volunteers who may be interested in contributing to wbond/package_control#890. If we don't have any movement on the issue in the next few weeks, I think we'll move forward with the rename.

whitequark:
Please go ahead with the rename once you have the time for it, I am OK with it as the package maintainer.

FichteFoll:
Added a test for default package names in 362f36917990242b95a5d6effb9811a9f6247a2b.

PR #4177, opened 2015-03-07
Added Python Output Format

Tenzer:
This plugin uses Python's [tokenize](https://docs.python.org/3/library/tokenize.html) module to format `print()`'ed code from Python programs into more readable text. There's an example in the [README file](https://github.com/Tenzer/PythonOutputFormat/blob/master/README.md), using the metadata this very pull request adds :)

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -1767,6 +1767,16 @@
 			]
 		},
 		{
+			"name": "Python Output Format",
+			"details": "https://github.com/Tenzer/PythonOutputFormat",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Python Path to Clipboard",
 			"details": "https://github.com/Mimino666/SublimeText2-python-package-to-clipboard",
 			"previous_names": ["Python Package to Clipboard"],


FichteFoll:
FYI, if you use tuples for your `if ... in ...:` checks you can "save some overhead". Not exactly huge, but maybe a good habit.

Tenzer:
Thanks! Actually, it should be sets (they are hashed) as they are O(1) compared to lists and tuples which (worst case) are O(n). Here's a quick benchmark:

    $ python -m timeit '"foo" in ["bar", "baz"]'
    10000000 loops, best of 3: 0.0694 usec per loop
    $ python -m timeit '"foo" in ("bar", "baz")'
    10000000 loops, best of 3: 0.07 usec per loop
    $ python -m timeit '"foo" in {"bar", "baz"}'
    10000000 loops, best of 3: 0.0371 usec per loop
    $ python -m timeit '"foo" in ["foo", "bar", "baz"]'
    10000000 loops, best of 3: 0.0339 usec per loop
    $ python -m timeit '"foo" in ["bar", "baz", "foo"]'
    10000000 loops, best of 3: 0.0701 usec per loop
    $ python -m timeit '"foo" in ("foo", "bar", "baz")'
    10000000 loops, best of 3: 0.0331 usec per loop
    $ python -m timeit '"foo" in ("bar", "baz", "foo")'
    10000000 loops, best of 3: 0.0711 usec per loop
    $ python -m timeit '"foo" in {"foo", "bar", "baz"}'
    10000000 loops, best of 3: 0.0374 usec per loop
    $ python -m timeit '"foo" in {"bar", "baz", "foo"}'
    10000000 loops, best of 3: 0.0368 usec per loop

I'll get it changed in the code, but I don't think it's worth pushing an update for it by itself, since it's only fractions of micro seconds which are saved :)

FichteFoll:
You are right. However, set literals have been added in Python 3 and thus won't work in ST2!

FichteFoll:
Well, I got curious and it appears that the Python interpreter (or compiler) has some auto optimizations in place for the `in` operator used with lists:

```
C:\Users\Fichte>python -m timeit "['foo', 'bar', 'baz']"
10000000 loops, best of 3: 0.102 usec per loop

C:\Users\Fichte>python -m timeit "('foo', 'bar', 'baz')"
100000000 loops, best of 3: 0.0153 usec per loop

C:\Users\Fichte>python -m timeit "'foo' in ('foo', 'bar', 'baz')"
10000000 loops, best of 3: 0.0313 usec per loop

C:\Users\Fichte>python -m timeit "'foo' in ['foo', 'bar', 'baz']"
10000000 loops, best of 3: 0.0311 usec per loop

C:\Users\Fichte>python -m timeit -s "foo = ['foo', 'bar', 'baz']" "'foo' in foo"
10000000 loops, best of 3: 0.0312 usec per loop

C:\Users\Fichte>python -m timeit -s "foo = ('foo', 'bar', 'baz')" "'foo' in foo"
10000000 loops, best of 3: 0.053 usec per loop
```

Tenzer:
I just checked with Python 2.7.8 and the above calls still work for me. I tried the benchmark as well and the tests involving sets got slower, around 0.130 microseconds, whereas the lists and tuples tests stayed around the same time. So I guess the conclusion is that sets are quicker on Python 3 but lists and tuples are quicker on Python 2 - which by the way have the same speed.

FichteFoll:
Well, I stand corrected again. Set literals where added in 2.7, but they still don't work in ST2 which uses 2.6 (and where I tested this in the console before posting).

Tenzer:
Ah, okay. So the final conclusion is that using lists and tuples are the same in `in` sentences, and should be used over sets if there's a chance of Python 2.6 being involved. I guess both of us learned something from this :D

FichteFoll:
TIL indeed. Thanks for introducing the timeit module to me.

PR #4178, opened 2015-03-07
added JavaScript Ultimate (and ascii version)

JoshuaWise:
I created a package to replace the built-in JavaScript syntax language. There are two versions: a default version (for full Unicode support), and an Ascii-only version.

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -289,6 +289,28 @@
 			]
 		},
 		{
+			"name": "JavaScript Ultimate",
+			"details": "https://github.com/JoshuaWise/javascript-ultimate",
+			"labels": ["language syntax", "snippets"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
+			"name": "JavaScript Ultimate (Ascii)",
+			"details": "https://github.com/JoshuaWise/javascript-ultimate-ascii",
+			"labels": ["language syntax", "snippets"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "JavaScriptFunctionDefinition",
 			"details": "https://github.com/asdf23/JavaScriptFunctionDefinition",
 			"labels": ["JavaScript", "autocomplete"],
--- a/repository/j.json
+++ b/repository/j.json
@@ -291,18 +291,7 @@
 		{
 			"name": "JavaScript Ultimate",
 			"details": "https://github.com/JoshuaWise/javascript-ultimate",
-			"labels": ["language syntax", "snippets"],
-			"releases": [
-				{
-					"sublime_text": "*",
-					"tags": true
-				}
-			]
-		},
-		{
-			"name": "JavaScript Ultimate (Ascii)",
-			"details": "https://github.com/JoshuaWise/javascript-ultimate-ascii",
-			"labels": ["language syntax", "snippets"],
+			"labels": ["language syntax", "snippets", "javascript"],
 			"releases": [
 				{
 					"sublime_text": "*",


FichteFoll:
Is there a specific reason why you decided to host the two language definitions in separate packages? I don't really see a problem with adding the ascii version to the other package as "Javascript (ascii)".

FichteFoll:
A few more things regarding your `setup.py`:

1. Since you disable the default package for Javascript (and JSON), you might want to mention that in the readme in case the user uninstalls your package at some point and finds himself without any javascript support.
2. Due to the above, you might be interested in the new events feature of Package Control 3.0: https://packagecontrol.io/docs/events
3. Because ST3 supports reading packages from the zipped `.sublime-package` format, your `getPackageName` function will yield the wrong result for that (something like `Javascript Ultimate.sublime-package`). You can access a global named `__package__` (just like `__file__`) that will hold the package's name, but only on ST3!

JoshuaWise:
Thanks for asking about that! Ultimately, there were several factors that influenced my decision. 

First, I’d have to give a scopeName to the ASCII version, such as source.js.ascii, because otherwise some of the language preferences (.tmPreferences) would clash. It’s not really standard practice to include the encoding in language scope names.

Also, even if I did that, Sublime Text can only effectively associate one filename extension (.js) with one language grammar. Upon opening files in Sublime, if you have both the Unicode and Ascii versions installed, Sublime will always choose the same one upon opening a js file, which might be the one you don’t want. So, the user would either have to manually switch the language every time they open a JS file, or they’d have to edit the language file themselves. And practically all users would only want one version anyways. 

Given this, I think having two versions is the best option for everybody. Let me know what you think of my thought process here.

Here’s what I did to fix the other issues you mentioned:
I updated the Readme to inform users about the disabling of the default JavaScript package.
I updated install.py to use the Package Control hooks for installation and uninstallation, so the default JavaScript package should be automatically re-enabled upon removal of my package.
I fixed the getPackageName function to work in both Sublime Text 2 and 3.

Thanks for the compatibility tips. Let me know if anything else needs to be done.

Cheers,
Joshua Wise

FichteFoll:
> First, I’d have to give a scopeName to the ASCII version, such as source.js.ascii, because otherwise some of the language preferences (.tmPreferences) would clash. It’s not really standard practice to include the encoding in language scope names. 

I don't see a problem with that, imo it might very well be the standard practice. Your syntax definitions are the first that I've seen to differentiate between unicode and ascii, so you can someone choose a practice.
The other argument is that you'd otherwise have the exact same scope for two different syntax definitions. I'd even go as far as label them as `source.js.ultimate` and `source.js.ultimate.ascii`.
However I'm not sure how a `source.js` include would behave then. It either tries to resolve to a nearby scope (such as `source.js.ultimate` or will always try to choose `source.js`, in which case it wouldn't work any more since you disabled the default package). ~~Needs testing.~~ Just tested, by renaming the default JavaScript syntax's base scope to `source.js.default`. It works.

>Also, even if I did that, Sublime Text can only effectively associate one filename extension (.js) with one language grammar.

Yes, but each user can configure that locally with the **View → Syntax → Open all with current extension as…**. In case you decide to bundle them into one package, you might even want to explicitly set the unicode variant as the default. (You do that by bundling a syntax-specific settings file with `"extensions": ["js"]`.)

>Given this, I think having two versions is the best option for everybody. Let me know what you think of my thought process here.

Well, it's certainly the "easier" one, even though I suspect the combined bundle would be easier to maintain. All your issues mentioned are solveable with more or less effort, so it comes down to whether you like that or not. Personally I would, especially if I was using something like https://packagecontrol.io/packages/SyntaxHighlightTools to compile the tmLanguage files with the same code base but different identifier expressions.

JoshuaWise:
Hi, sorry for my late response. I talked with some other programmers and did some tests, and I decided to simply only release the Unicode version.
I have several reasons for doing this:
People who don’t want to use non-Ascii characters can simply not use Ascii characters. I don’t see any harm in giving them Unicode support even though they won’t be using it.
Using the Unicode version of JavaScript Ultimate doesn’t conflict with other packages that try to highlight or eliminate unwanted Unicode characters.
After doing some tests on my local machine, I found no practical performance difference between the Ascii and Unicode versions, even though common sense would lead me to believe that the Ascii version would be faster.
I’m slightly lazy, a little bit OCD, and providing both Ascii and Unicode versions together makes my head twitch. Mostly because I don’t want to add confusion/clutter for people who don’t read my Readme. I prefer not to overcomplicate things.

Anyways, I’ll send another pull request to match this decision.
I hope to hear back soon.

Thanks,
Joshua Wise

JoshuaWise:
Quoting my last email:
> Anyways, I’ll send another pull request to match this decision.
Disregard this. I wasn’t aware that pull requests automatically point to the most recent commit.
Anyways, I hope to hear back soon.

Thanks,
Joshua Wise

FichteFoll:
Very fine by me.

JoshuaWise:
I just tried installing JavaScript Ultimate through package control, but the install.py script doesn’t seem to be functioning. Is there a way to quickly debug this, without constantly pushing new versions of my package to its Github repository? Or, could you take a look at it and check if I made an obvious mistake?

Thanks,
Joshua Wise

FichteFoll:
What exactly do you mean "isn't functioning"? Is there a traceback? Maybe `events.install(packageName)` is not returning true?

You can test installation with PC by creating your own `repository.json` file and adding that to the `repositories` user setting of Package Control. These may point to local files, so `"e:\\Development\\Sublime\\dependency.json"` is valid. However, you can't install local packages, so in the actual package specification you need to point to a url that allows direct download of your package in a zipped state. You could of course also host a dummy repo for that purpose.

JoshuaWise:
I added this to my user settings of Package Control:
"repositories": [
    “/path/to/dependency.json"
]
and dependency.json looks like this
{
    "schema_version": "3.0.0",
    "packages": [
        {
	    "name": "JavaScript Ultimate Dev",
	    "details": "https://github.com/JoshuaWise/javascript-ultimate-dev/archive/master.zip",
	    "labels": ["language syntax", "snippets", "javascript"],
	    "releases": [
	        {
		    "sublime_text": "*",
		    "tags": true
	        }
	    ]
        }
    ]
}

JoshuaWise/javascript-ultimate-dev is a new repo that I created for testing purposes.

But when running Package Control: Install Package, I get:

Package Control: Invalid "base" value "https://github.com/JoshuaWise/javascript-ultimate-dev/archive/master.zip" for one of the releases of the package "JavaScript Ultimate Dev" in the repository /path/to/dependency.json.

When installing the live JavaScript Ultimate package, I get no traceback in the Sublime console, just "reloading plugin JavaScript Ultimate.install”.

FichteFoll:
`"https://github.com/JoshuaWise/javascript-ultimate-dev/archive/master.zip"` should be `"https://github.com/JoshuaWise/javascript-ultimate-dev" and `"tags": true` should be `"branch": "master"`.

JoshuaWise:
Thanks, it’s working now.
I tried adding “print” calls to install.py, but none of them are being invoked.
I even added “print('- script start -')” to the very beginning of the script, but it doesn’t print anything to the sublime console upon installation through Package Control.
What could be going wrong?

I’m sorry to be taking up so much of your time.
Joshua Wise

JoshuaWise:
Hmmm, disregrd my previous email. I restarted Sublime Text, and now the print calls are working. Strange.
Anyways, thanks.

JoshuaWise:
Okay, so I’ve located the problem in my install.py script, but I don’t know why it exists. Maybe you can help me.
The script I’m referring to is here: https://github.com/JoshuaWise/javascript-ultimate-dev/blob/master/install.py <https://github.com/JoshuaWise/javascript-ultimate-dev/blob/master/install.py>
upon installation through Package Control, plugin_loaded is invoked properly, but events.install(packageName) evaluates to False.
I double checked that  packageName is indeed the correct name of the package (in my testing case, "JavaScript Ultimate Dev”), and it is.
Any ideas at why the package_control event is evaluating false?

FichteFoll:
I don't see anything wrong currently, but I don't have the time to debug this right now. Maybe @wbond can tell you more about this since he actually wrote the code and (hopefully) tested it already.

However, I can tell you that you definitely don't need to append `.sublime-package` for the package name.

FichteFoll:
Since you are responding to emails and thus likely won't see edits: The package name parameter of your testing package obviously needs to be `events.install("JavaScript Ultimate Dev")` because that's what you specified in your local `dependency.json`. I only chose that file name to test dependencies btw, but it doesn't matter.

JoshuaWise:
Right, the ".sublime-package" appendage was just for testing. It doesn't work with simply the literal string "JavaScript Ultimate Dev" either. I'll try to get in touch with @wbond .

FichteFoll:
Does it work now?

PR #4179, opened 2015-03-07
"Preap" dark theme and syntax

equinusocio:
Added "Preap" theme package

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -1169,7 +1169,7 @@
 			"labels": ["theme", "color scheme"],
 			"releases": [
 				{
-					"sublime_text": ">3000",
+					"sublime_text": ">=3000",
 					"tags": true
 				}
 			]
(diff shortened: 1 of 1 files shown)


equinusocio:
New awesome dark theme added!

equinusocio:
@FichteFoll Hi, there's a reason why my package does not appear in the list of available packages? I done something wrong?

FichteFoll:
Ah, yes. I didn't notice it in my initial scan, but your tags currently have the format `v.major.minor.patch` whereas PC by default only finds `vmajor.minor.patch` and `major.minor.patch`. You eather need to fix the tags (I've never seen that naming scheme before) or adjust the release here to specify `"tags": "v."` so that PC will find the prefix you used.

equinusocio:
Ok! Thanks. I will simple update the release prefix. Thanks!

PR #4180, opened 2015-03-08
Fix bug in highlighting

nyx0:


FichteFoll:
The correct way to do string matching in YAML-tmLanguage format is the following:
```yaml
- name: string.quoted.double.yara
  begin: '"'
  end: '"'
```

Please remove your `package-metadata.json` from the repo, otherwise things like https://github.com/wbond/package_control/issues/900 could happen.

Also, when renaming you have to add the old name to a `previous_names` array.

nyx0:
Ok I think all it's good thank you very much for your advices.

FichteFoll:
So, uhm, I'm not sure what you're doing, but I think you're doing it wrong. `'"'` should be interpreted by the yaml parser as a single `"`, which would then be output into the tmLanguage plist. In https://github.com/nyx0/yara/commit/c5528acd9a6ab459db8c7c903362e79eeb1604d9 however it wasn't just `"` being added but the whole `'"'` sequence. Are you doing these edits by hand?

nyx0:
My bad, yes for this modifications I've done it by hand :/ I know

PR #4182, opened 2015-03-09
Mark JSCoverage as compatible with 3rd version too

genbit:


FichteFoll:
- Python 2.6's `open` does not know the `encoding` keyword argument.
- In line 75 you use a hardcoded forward slash with file paths. This will fail on Windows.
- Usually, you append a newline to your logging and not the other way around (prepend newline, then print data).
- Also, you might want to consider gating the verbose debugging behind a debug setting.

PR #4186, opened 2015-03-10
Add new plugin fecoding

pillys:
Add a new plugin named fecoding, a plugin based node.js

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -196,6 +196,17 @@
 			]
 		},
 		{
+			"name": "Fecoding",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Since you are executing a file in your package you need to add a `.no-sublime-package` file to your repo root that tells Package Control to unextract your package (for ST3).

Also, using branches for releases is deprecated and won't be accepted for new packages anymore: https://packagecontrol.io/docs/submitting_a_package#Step_4

pillys:
Thanks, now the tag is added.

PR #4192, opened 2015-03-12
Added Flask Plugin

geekpradd:
This plugin adds autocompletions and snippets for the Flask microframework. 

Everything is working however I need to add a certain feature to the plugin that I am unable to do so.
@FichteFoll, you've helped me before and so can you please tell me how do I add the name of the module in the autocompletions.

Like for example in the AngularJS plugin. 

Here to the right of the autocomplete AngularJS is written in Italics. I want to achieve the same effect but sublime-completions does not support it natively (snippets do). Currently, the autocompletion looks like this:

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1347,6 +1347,17 @@
 			]
 		},
 		{
+			"name": "Color Scheme - Dusk",
+			"details": "https://github.com/geekpradd/Dusk-Color-Scheme",
+			"labels": ["color scheme"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Color Scheme - Eggplant Parm",
 			"details": "https://github.com/mimshwright/sublime-eggplant-parm",
 			"labels": ["color scheme"],
--- a/repository/d.json
+++ b/repository/d.json
@@ -947,6 +947,17 @@
 			]
 		},
 		{
+			"name": "Dusk",
+			"details": "https://github.com/geekpradd/Dusk-Color-Scheme",
+			"labels": ["color scheme"],
+			"releases": [
+				{
+					"sublime_text": "*",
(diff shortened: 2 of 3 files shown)


geekpradd:
I've added another plugin.. This is a color scheme that I have been working on for some time.. it's called Dusk and it goes well with the Flatland theme (it's based on the color brown though)..

FichteFoll:
I suggest to rename your package to "Flask Completions" because that what its main purpose is. It's also not exactly auto-complete as that suggest context-awareness, such as the type of the variable you are accessing and providing only relevant completions or completing function signatures.

All types of completions allow hints, see http://docs.sublimetext.info/en/latest/reference/completions.html#trigger-based-completions.

geekpradd:
Ok.. I'm gonna rename it.. Also what do you mean by hints? 
Do you mean the italic text that comes to the right of the completion?

FichteFoll:
[Yes](http://docs.sublimetext.info/en/latest/extensibility/completions.html#hints).

I would also suggest you to comply to the naming convention of color schemes and themes, where this is included in their name, e.g. "Color Scheme - Dusk".

geekpradd:
Thanks.. The hint part will take some time as I have to modify the entire json file. Time to create a python script.
I'll be renaming the color scheme too. 
Everything should be done in half an hour or so.
Thanks again.

geekpradd:
@FicheteFoll, do I need to rename my repository as well or just the value of the `name` key in the JSON files?

geekpradd:
@FichteFoll .. Everything is done and changed. You can merge now.. (unless you need to change the repository name as well)

FichteFoll:
No, repository name doesn't matter. However, you need to fix the location of the packages in the json files since you renamed them.

geekpradd:
@FichteFoll..  Everything is correct now.

PR #4193, opened 2015-03-12
Add new HaoGist plugin

xjsender:
Add new HaoGist plugin

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -107,6 +107,17 @@
 			]
 		},
 		{
+			"name": "HaoGist",
+			"details": "https://github.com/xjsender/HaoGist",
+			"labels": ["gist", "cache"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "haoide",
 			"previous_names": ["Salesforce IDE"],
 			"details": "https://github.com/xjsender/haoide",


FichteFoll:
Please make use of the new dependencies feature and require `requests` instead of bundling it in your package: https://github.com/packagecontrol/requests

@codemil, can you comment on the state of your Gist package? Would you be interested in adding @xjsender as a contributor so that we don't have two packages doing very similar things?

xjsender:
@FichteFoll , thanks for your tip about dependency part, I have already used this feature in my plugin.

I remember, I requested a ```refresh gist``` feature from that plugin but no response since half a year ago.

Honestly, gist was blocked in china half a year ago, I can only use gist Api to manage my gists, however, there is no thread process and cache in that gist plugin, so I can only develop my own plugin to manage my gists.

Thanks.

FichteFoll:
I will wait a few days for @condemil to read and reply before merging.

condemil:
@FichteFoll, thanks for summoning me here. I think it will be better to keep on the collaborated development of one feature-rich plugin.

@xjsender, if your ready to merge everything together and support only one package I'm fine to add you as a contributor.

xjsender:
@FichteFoll , I am not intended to merge something to ```gist``` plugin, can you please just merge this plugin?

FichteFoll:
Well, it's sad that you are unwilling to collaborate, but I won't force you to. You should be aware of advantages and disadvantages by now.

PR #4195, opened 2015-03-12
Some tests updates

FichteFoll:
Since these are bigger-scale changes than what I do usually, I thought you might want to review this, @wbond. As always, highly suggested to review by commit.

Verified to work with ChannelRepositoryTools on ST2 and ST3 this time.

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -295,7 +295,7 @@ class TestContainer(object):
     }
 
     def _test_release(self, package_name, data, main_repo=True):
-        # Fail early and test for required keys
+        # Test for required keys (and fail early)
         if main_repo:
             self.assertTrue(('tags' in data or 'branch' in data),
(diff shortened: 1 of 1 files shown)


wbond:
I just stepped through all of the commits.

The last yield is very odd – I'm sure that will bug me until I figure out why. :smile:

The only other one I noticed was combining the package and dependency release tests. I'm up in the air as to whether adding complexity inside the method is worth removing some repetition, but it is probably generally a good step forward.

FichteFoll:
>I'm up in the air as to whether adding complexity inside the method is worth removing some repetition

Yeah, I also noticed how complexity increased (quite a bit) but not too much to warrant the repetition. It's managable currently, but hopefully we won't have to add more.

I was also very surprised about the yield thing, but didn't bother digging into why it wouldn't work because it's likely not worth the time. I should be studying anyway :S

PR #4198, opened 2015-03-12
Add Aristocat theme

ddeville:
New Aristocat theme for Sublime Text.


FichteFoll:

You latest tag `1.2` has to be `1.2.0` to get picked up.

PR #4199, opened 2015-03-12
Added Mistral package.

giampierod:


FichteFoll:
Sublime Text does not support snippets with the `.tmSnippet` extension, they have to be `.sublime-snippet`. See also http://docs.sublimetext.info/en/latest/extensibility/snippets.html

Additionally, using branches for releases is deprecated and won't be accepted for new packages anymore: https://packagecontrol.io/docs/submitting_a_package#Step_4

giampierod:
tmSnippet works fine with Sublime Text 2, I have customers who have tested it as well as myself. It has worked for years, I have two other packages that use it. The default Scala and GO packages that ship with Sublime Text 2 use tmSnippet files.

I see the part about tagging and will make the update.

FichteFoll:
Oh sorry, it appears you are right. ST does indeed pick up tmSnippet files.

Makes me wonder why `.sublime-snippet` exists in the first place.

giampierod:
tmSnippet comes from the TextMate system. There were a ton of packages made
for that text editor and the Sublime Text folks wanted to leverage all that
goodness to kick start their app. But .sublime-snippet is the native format
for Sublime.

I guess I am just old school. And I really hate those CDATA tags.

I will work on fixing the issue with my tags/branches later tonight and
send a new pull request.

FichteFoll:
Well, it would have been nice if ST had used the chance to go away from the plist format if it was introducing a different extension anyway, such as the one used by https://sublime.wbond.net/packages/SaneSnippets.

Either way, you don't need to open a new pull request, you can just append commits to this one.

giampierod:
Ok, I think I got it now.

FichteFoll:
Almost. Semantic versions have three numeric parts.

giampierod:
I swear I will get this. Updated my tag.

FichteFoll:
There we go.

PR #4205, opened 2015-03-13
Midnight theme for ST3

devottam:
A dark gray theme for Sublime text. Noticeably highlights selected item during auto-complete, command palette etc

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -735,7 +735,7 @@
 			"releases": [
 				{
 					"sublime_text": ">=3000",
-					"branch": "master"
+					"tags": true
 				}
 			]
 		},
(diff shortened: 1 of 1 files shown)


wbond:
So, you need to change `"branch": "master"` to `"tags": true` and then create at least one tag (https://github.com/devottam/midnight/tags) that is a three-part version number, e.g. `1.0.0`.

You don't need a new PR, just add a new commit with the changes.

devottam:
Updated. Thanks!

FichteFoll:
You still need to push a tag though.

devottam:
__Sorry__ I thought, I pushed all the tags but somehow it didn't.

Done.
Thanks.

PR #4210, opened 2015-03-14
Add the Meteor Autocomplete package

Slava:
Hi, I am authoring this new package that adds a headache-free autocompletion for [Meteor](https://github.com/meteor/meteor), a popular JS framework for App development.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -673,7 +673,7 @@
 			]
 		},
 		{
-			"name": "Meteor Autocomplete (Tern.js)",
+			"name": "Meteor Autocomplete (TernJS)",
 			"details": "https://github.com/slava/tern-meteor-sublime",
 			"releases": [
 				{
--- a/repository/m.json
+++ b/repository/m.json
@@ -673,6 +673,16 @@
 			]
 		},
 		{
+			"name": "Meteor Autocomplete (Tern.js)",
+			"details": "https://github.com/slava/tern-meteor-sublime",
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
You can't have dots in package names when using plugins because they will fail to load.

Selections have `.begin()` and `.end()` methods to do what your sel_end and sel_start functions do.

I don't think it's a good idea to bundle the node binary because it's considerably large and everyone has to download the binaries for the other platforms as well. Instead, just require it to be installed and available on PATH or add a setting to configure the path/specify where it is located.

Slava:
@FichteFoll I can rename the package to have no dots.

I insist on having node binaries to a) pin the version b) improve the install process. Most Meteor users never run node directly. Having a global node binary that is up to date is often a blocker and is confusing for my users.

Slava:
@FichteFoll Thank you for the review, I am addressing everything you mentioned:

- Removed the dots
- Since this package is some form of a fork of a different package (the original TernJS), I want to keep the changes in the python code to minimum to make it easier to rebase in the future
- as I said above, I think it is the best trade-off for my situation to let people download a 15mb package once and never bother to fix their broken installation ever again. That being said, I see why it looks controversial. My original package (200 stars, https://github.com/Slava/tern-meteor) worked this way and only a handful of users managed to setup everything correctly.
- I pushed the `v1.0.0` tag.

FichteFoll:
Okay. You might want to mention the 15MB download in the readme however.

PR #4214, opened 2015-03-15
added Jolie Language plugin

thesave:


FichteFoll:
You should remove your `*.tmLanguage.cache` files, they are only a cache for ST2 and will be generated, and ST3 stores them at a different location anyway.

thesave:
Thanks for the quick reply and your time.
I think I made the proper the changes.

PR #4216, opened 2015-03-16
Added pygments dependency

AndreasBackx:
I didn't really know what to put in for load_order so I just put "01" because it doesn't depend on anything. I couldn't find anything on it and I want to round this up so I can go to sleep too.

### DIFF
--- a/repository/dependencies.json
+++ b/repository/dependencies.json
@@ -45,6 +45,20 @@
 			]
 		},
 		{
+			"name": "pygments",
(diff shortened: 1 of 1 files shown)


FichteFoll:
I just put 50 because that's in the middle so that there is enough space for adding some dependencies on both sides, even though they don't depend on anything. Well, pyyaml can optionally work with libyaml, but that wasn't made available yet.

Otherwise looks fine.

PR #4223, opened 2015-03-17
Adding StraightQuotes plugin

dhirsch:
My first plugin, so please let me know if there's anything wrong

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -2199,6 +2199,16 @@
 			]
 		},
 		{
+			"name" : "StraightQuotes",
+			"details" : "https://github.com/directpartners/StraightQuotes",
+			"releases" : [
+				{
+					"sublime_text" : "*",
+					"branch" : "master"
+				}
+			]
+		},
+		{
 			"name": "Strapdown Markdown Preview",
 			"details": "https://github.com/michfield/sublime-strapdown-preview",
 			"previous_names": ["Strapdown.js Markdown Preview"],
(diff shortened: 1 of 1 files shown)


FichteFoll:
You have to push valid [semantic version](http://semver.org) tags (with optional leading `v`) to your repo. They consist of three numeric parts.

Then, your package is *very* similar to MagiclessQuotes by @daryltucker, which you referenced. I suggest you work on a pull request and collaborate with @daryltucker on this because of that. I'm sure you can also work out why it didn't work for you previously (see end of my post).

By the way, your plugin doesn't work correctly. Test with multiple one-to-many replacements of the same type, e.g. multiple `™` in your test file. It will become all messy. The reason for that is that by replacing content with `view.replace` you change the view's text extends and the previously returned regions from `view.find_all` are not updated with these changes. If you insert text before a region then it will shift by the difference of the previous to the new text.
Work around: Do replacements in reverse, that is `for rgn in reversed(self.view.find_all(...)):`. Magiclessquotes actually does that, even though differently.

@daryltucker, your `self.view.end_edit(edit)` line needs to go, that method doesn't exist in ST3 anymore. It doesn't affect funtionality since it's the very last line of the code, but it creates an ugly traceback in the console every time it is run.

daryltucker:
@FichteFoll Thanks for fix!

@dhirsch Let me know if you want to get together, or let me know what MagiclessQuotes is missing for you!

FichteFoll:
Closing since it seems you have gotten in touch on the MagiclessQuotes repo.

PR #4227, opened 2015-03-19
Update v.json for VvPhpAudoDollarSign plugin.

ZhaonanLi:
Submit VvPhpAutoDollarSign plugin.
Author: Zhaonan Li.

### DIFF
--- a/repository/v.json
+++ b/repository/v.json
@@ -483,6 +483,16 @@
 					"branch": "master"
 				}
 			]
+		},
+		{
+			"name": "VvPhpAutoDollarSign",
+			"details": "https://github.com/ZhaonanLi/VvPhpAutoDollarSign",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "master"
+				}
+			]
 		}
 	]
 }


FichteFoll:
1. Using branches for releases is deprecated and won't be accepted for new packages anymore: https://packagecontrol.io/docs/submitting_a_package#Step_4

2. You must not read settings in the top-level of a module in ST3 since the API might not be available at the time the module is initialized. `plugin_loaded` will be called (if it exists) when the API has finished loading. (To test this for yourself, just restart ST and see if there are errors in the console or the plugin doesn't work.)

3. `view.begin_edit()` and `view.end_edit()` don't exist in ST3 anymore. You need to create a TextCommand for the changes and call it with `view.run_command`.

4. The `xrange` function doesn't exist in Python 3, just use the normal `range`.

5. Additionally, it is usually recommended to read settings when you need them, so that the user can modify them on the fly (which will also avoid 2.). Since you need them so often, you should check out `Settings.add_on_change`. Don't forget to remove a listener/callback when your plugin is unloaded (i.e. disabled or removed) though. ST3 will call `plugin_unloaded` and `unload_handler` (for compatability and because of a bug) while ST2 will only call `unload_handler`. ([example](https://github.com/FichteFoll/FileHistory/blob/d6e6f4d84794f5e47c0b2ac1c775dcec978ef834/file_history.py#L792-L800))

  Raising exceptions is also not recommended. They will only be printed to the console anyway, in which case you might as well just `print`.

6. Your `VvPhpAutoDollarSign` class only contains static functions. They are marked as classfunctions, but they don't actually do anything with the class, so they might as well be just functions.

ZhaonanLi:
Hi FichteFoll, 

Thanks a lot. I have developed a new plugin based on your suggestions.
https://github.com/ZhaonanLi/VvPhpDollar
And thanks again, I learnt a lot from your comment.

Sincerely,
Zhaonan Li.

FichteFoll:
That looks better.

However, you should consider that a user may have multiple carets and inserted the sequence at multiple locations, not just `view.sel()[0]`. You should check them all.

I also remember something that I forgot yesterday: You change case of the syntax names of the settings but you don't do it for the syntax name you retrieve from the view. I suggest you don't touch case at all.

You just only to push a tag and update the repository information in this pull request.

PR #4228, opened 2015-03-19
Add ChefEncryptedDataBag package

labocho:


FichteFoll:
I suggest you to remove the `print(settings)` line since it will hardly provide useful information to anyone. I also suggest you to use `os.path.join` instead of forward slashes, even though Windows usually handles them as you'd expect.
And finally, since your `ProcessDataBagItemCommand` class is not meant to be called by ST but meant to be used as a mixin, it should subclass `object` and not `sublime_plugin.TextCommand`. Instead, the other commands e.g. `EncryptDataBagItemCommand` should subclass both. (Currently `view.run_command("process_data_bag_item")` will execute the command and result in some error message being raised.

labocho:
Thank you! I'll fix it.

PR #4231, opened 2015-03-19
simple number increment that also takes user input

Leulas:


FichteFoll:
You can do that with [Insert Nums](https://packagecontrol.io/packages/Insert%20Nums) and the following format string: `i|_+1`
You can also save the format string with a key binding for quicker access.

I'd argue this package is not needed.

Leulas:
in my opinion it's simpler to use with out any extra knowledge, is customizable with the argument for your prefrence and the user input windows alow for non standard changes easy. I made it cause i didn't want a big package just for that feature. Do as you wish =)

FichteFoll:
Depends on what you define as "big". InsertNums is only 450 lines of code and I'd consider that rather small.

I wrote the current iteration of InsertNums so I might be biased, but imo it's really simple to use. You would need to know the abstract formatting syntax for some features however (e.g. expression mode) and those can't exactly be explained in the input prompt without losing clarity. I would display an output panel with a quick reference if it was possible, you can only have one panel at a time.

It would probably be reasonable to extract the expression functionality into a different package (one package for one cause), but the code is actually very similar. Besides, there are other packages (yes, mmultiple iirc) that do evaluation stuff on selections (but not as neatly as InsertNums).

By the way, I would expect the parameter of your command to be `by` and not `steps`.

Leulas:
Thank you for your input, will change steps to by. I just prefer a smaller(as in less feature packed) package and thought it would be fun to contribute in some way. Love both sublime and package control so i gave it a shoot^^

FichteFoll:
Sure thing.

I just don't see a need for this (currently). If I wasn't the only one reviewing packages I'd ask about his opinion since I may be biased, but there is only Will currently and he likely has other things to do than review this case. Thus, I'm closing this.

PR #4232, opened 2015-03-20
Update s.json with a new plugin- Summary

geekpradd:
This plugin summarizes text by using the following algorithm: https://gist.github.com/shlomibabluki/5473521

It can be very useful when writing documents in Markdown or ReStructuredText or TXT files.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -3022,6 +3022,17 @@
 			]
 		},
 		{
+			"name": "Summary",
+			"details": "https://github.com/geekpradd/Summary-Sublime",
+			"labels": ["text manipulation"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Summerfruit Color Scheme",
 			"details": "https://github.com/thgie/Summerfruit",
 			"labels": ["color scheme"],


geekpradd:
@FichteFoll , I want your help in a plugin project. I'm creating a Sublime Plugin to switch themes directly from the GUI (like how Color Schemes can be tweaked from the GUI). I have been able to collect the Theme files (by using lots of `os.path` code) but the main problem is modifying the Menu on demand.

Since the menu code is stored in the `Main.sublime-menu` file which once submitted as a plugin becomes a part of the `sublime-package` file, I am unable to modify the menu.

Do you have any idea on how I can change the Menu file contents programmatically from Python?
I want to change the sub menu contents. (Like how Sublime Text natively does in C++)

FichteFoll:
- You need to generate the menu "on the fly" using the `json` module.

- `os.path` code won't work for ST3 because it stores schemes in zipped archives. Use `sublime.find_resource` and `sublime.load_resource` there.

- Since you need to place the Menu file in some packages directory, you probably have to create one when the package is installed on ST3 (because it won't exist there since it's installed as a zipped archive). I suggest using the plugin's name in that case. Don't forget to remove the menu file when the plugin is disabled or uninstalled (=unloaded) however, using `plugin_unloaded`. You can check out https://packagecontrol.io/docs/events.

geekpradd:
And @FichteFoll , how do i change the prefernces user file from Python?

geekpradd:
Ah, I get it.. So I will create a custom Menu file in a package folder. Just tell me how to modify preferences and I'll be done.

FichteFoll:
What kind of preferences? `sublime.load_settings`/`sublime.save_settings`?

geekpradd:
Ok, `save_settings`. I got the plugin working but I was manually overriding the Preferences file. Thanks

geekpradd:
@FichteFoll , Also is there a `on_plugin_installed` or something similar API that Package Control provides? I want to refresh the menu file whenever a new plugin is installed. Thanks.

FichteFoll:
No, there is not. You could listen to changes to the PC preferences file however and monitor the `installed_packages` list.

PR #4233, opened 2015-03-20
change DjangoManageCommands from repository to channel.json

Proyecto513:
Is there any better way to give  a specific version of the plugin depending the sublime text version?


FichteFoll:
[Yes](https://github.com/wbond/package_control/blob/d3bc04fd5c0c4bdb476a3ea798e6815508dc9bac/example-repository.json#L273-L301). This can also be done in the default repository files.

Proyecto513:
Thanks :smile:

PR #4235, opened 2015-03-20
Big Comment

iBad:


FichteFoll:
https://packagecontrol.io/packages/ASCII%20Decorator does this too, execept that it provides a lot more option. I'd argue that there is no need for this package in the list.

PR #4239, opened 2015-03-23
Added Sublime-Ascii-Art

xgvargas:
HI, It's my first plugin, so let me know if something is missing...

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -966,6 +966,16 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
See https://github.com/wbond/package_control_channel/pull/4235#issuecomment-84213381

xgvargas:
Yeah, I have to agree with you.... let me improve it!!

PR #4240, opened 2015-03-23
Adding SoDaReloaded Theme

Miw0:
Dark custom UI theme for Sublime Text 3 optimized for better readability. The theme supports the new Sublime Text 3 icons for file types out of the box!

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -931,6 +931,18 @@
 			]
 		},
 		{
+			"name": "Theme - SoDaReloaded",
+			"details": "https://github.com/Miw0/sodareloaded-theme",
+			"author": "Miw0",
+			"labels": ["theme"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Theme - Sodarized",
 			"details": "https://github.com/jrolfs/sodarized",
 			"labels": ["theme"],


wbond:
Umm, this seems to pretty clearly violate the license terms of the Soda Theme. I think it should be removed until a better name is chosen.

https://github.com/buymeasoda/soda-theme#license

Miw0:
Hm regarding other packages like "Theme - Sodarized" or "Theme - Soda SolarizedDark" I thought the name would be abstract enough to not violate the license :/

https://packagecontrol.io/search/soda

wbond:
Well, from my perspective, if someone forked my project and called it "ProjectPlus" or "ProjectReloaded" and I had asked people to make a distinct name, I would consider it a violation. That said @buymeasoda would be the final judge.

If I had seen the `Theme - Soda Solarized` come through, I may have raised issue.

The only reason I bring this up is because I think this happened in the past with another Soda variant, and I thought Ian came and voiced his opinion that it did not follow the license. However, I can't find that via search, so maybe I am just imagining.

Miw0:
I chose the name because it's not a completely new theme. I just tweaked some settings to get a cleaner look and added icon support for files. After some time of usage I thought other people would like it and that's why I wanted it to be installable via package control. 

It almost looks like the awesome Soda Theme Dark and I think it would be much more wrong to give it a completely new name because of that. I also honor the work of @buymeasoda which led me to use this name which sounds familiar to the original theme.

Don't understand me wrong, I have no problem in changing the name if it is violating the license, I just wanted to make clear why I chose this name :)

PR #4244, opened 2015-03-23
Add Theme - Darkmatter package

mpatrick:


FichteFoll:
There is no reason for this package to be restricted to ST2 and linux/osx. It's just a theme+color scheme, which are independant of platform or ST version.

You should remove `package-metadata.json`, it is supposed to be automatically generated by PC on install.

Nobody will see the update message for 1.0.0 because it's the very first version available. It also has the a wrong filename.

Your `messages.json` contains an errorneous trailing comma.

mpatrick:
@FichteFoll ok, I didn't know that this would work with ST3 and Windows because I didn't test on these environments, but thanks, I'll fix the issues and amend the commit.

mpatrick:
@FichteFoll ok, I think we are good to go now :smile:

PR #4247, opened 2015-03-24
Add FilterPipes plugin

tylerl:


FichteFoll:
While all the features that this package provides are, to my knowledge, also available as individual packages already, the compilation and similar data structuring is certainly attractive.

However, using branches for releases is deprecated and won't be accepted for new packages anymore: https://packagecontrol.io/docs/submitting_a_package#Step_4

tylerl:
Updated the version strategy.

Plugins do exist for the "execute command line" feature and "base64 encode your selection" and bits like that, but those are included more as example of use rather than core features. The real driving force behind this plugin is the idea of making it super easy to create your own filters as you need them. 

We've been using it for several months now and it's great -- you run into a situation where it would be really simple to accomplish X if you could send selected text through a certain 3-line python script. And about 90 seconds later you have your own custom command that does just that.

I have dozens of filter commands configured on my own machine for all sorts of use cases; I'm considering releasing more of them depending on how useful they may be to a wider audience.

PR #4248, opened 2015-03-24
Added Theme Switcher plugin

geekpradd:
This is the plugin I asked you for help, @FicheteFoll. 
It works nicely in ST3 on both Windows and Linux although I was unable to implement a Background Task to watch for new Theme Installation. Instead I added a GUI "Refresh Theme Cache" option.

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -1077,6 +1077,17 @@
 			]
 		},
 		{
+			"name": "Theme Switcher",
+			"details": "https://github.com/geekpradd/Theme-Switcher",
+			"labels": ["theme"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "ThemeAutomator",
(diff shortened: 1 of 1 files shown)


FichteFoll:
There are a few theme-switching packages already out there, but they are all a bit different in their usage. It would be preferrable if you could indicate the usage pattern of your package in its name, e.g. "Theme Menu Switcher" or "Themes in Menu" or similar.

https://packagecontrol.io/packages/QuickThemes
https://packagecontrol.io/packages/Themr

FichteFoll:
Using the `json` module for modifying the theme setting is error prone because it can't handle comments (which are allowed in `.sublime-settings` files). Use `sublime.load_settings` and `sublime.save_settings` instead.

geekpradd:
Okay, I changed the name and I am using the Sublime API for modifying settings.

FichteFoll:
I just noticed that you crawl the zip files manually, only search in those tha follow the `Theme - <name>` convention and don't consider the normal packages not at all.

Just use `sublime.find_resource`.

PR #4250, opened 2015-03-24
Add FBP DSL syntax highlighting

trustmaster:
This one is more simple than https://github.com/paulyoung/FBP.tmbundle and supports ST only, plus it uses different colors.

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -197,6 +197,16 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
This is about the same languages and both packages contain nothing more than the language itself (except that the other package also defines comment sequences). I suggest you get in tough with @paulyoung instead and submit a pull request with your changes.

PR #4252, opened 2015-03-24
Curiosity Theme

dmnplb:


FichteFoll:

I would also suggest to delete all tags and start with `v1.0.0` since this essentially a new package,

dmnplb:
@FichteFoll Deleted all the tags. There's only one now (v1.0.0). Thanks for the clarification ;)

PR #4254, opened 2015-03-25
Add package 'PlasticSCM'

ccll:
PlasticSCM is a commercial version control system, this package integrate some feature with Sublime Text.


FichteFoll:

PR #4256, opened 2015-03-25
C/C++/OC code autocompletions base on libclang.

lvzixun:


FichteFoll:
Since your package opens the file `lib/libcc.so` it needs to be extracted. Add a `.no-sublime-package` file to the repo that tells Package Control to do that.

lvzixun:
thanks, I create `.no-sublime-package` file at repository. and `"branch": "master"` replaced by `"tags": true`.
I should be careful to read the manual at first.  Orz

PR #4259, opened 2015-03-26
added doge

tylerl0706:
Added the Sublime Text 3 plugin: doge

This was my first Sublime Text plugin so if there is anything I did incorrectly, please let me know!

Thanks!

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -685,6 +685,16 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
Yeah ... no. Interesting idea and POC, but since this package is essentially useless I don't think adding it to the PC channel will be of use for anyone.

PR #4267, opened 2015-03-29
Added es6 navigator to repository

ryanhollister:


wbond:
You can remove the `readme` entry from your PR – Package Control will find it automatically.

Why do you have a `.no-sublime-package` file in the repo? It doesn't seem you should need that.

ryanhollister:
bear with me here as it's my first package. Removing the readme is fine. But if I remove .no-sublime-package then it seems I can't ship a default package settings file. Am I missing something?

wbond:
The `.no-sublime-package` file only controls if the package is unzipped onto disk for Sublime Text 3. The only packages that need to be unzipped are packages containing shared libraries or executable/shell scripts. The settings file can be read by Sublime Text from the zipped `.sublime-package` file. User's changes to it should be saved in a file with the same name in the `Packages/User/` folder.

ryanhollister:
Thanks, I will clean this up.



Ryan Hollister

FichteFoll:
@ryanhollister, it appears you re-added the `.no-sublime-package` file. I suggest you check out other plugins and see what they are doing in regard to settings management, since they are doing it "right".

You are also using a lot of hardcoded forward slashes for dir separators instead of `os.path.join`.

PR #4268, opened 2015-03-29
Added new plugin, HoogleSearch

s4wny:
[Hoogle](https://www.haskell.org/hoogle/) integrated into sublime text.

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -528,6 +528,18 @@
 			]
 		},
 		{
+			"name": "HoogleSearch",
+			"details": "https://github.com/s4wny/HoogleSearch",
+			"author": "Sawny",
+			"labels": ["haskell", "hoogle"],
+			"releases": [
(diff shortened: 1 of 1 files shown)


wbond:
From looking at your python, I don't think this will work in ST2, so you should update the `sublime_text` selector to `>=3000`.

Also, you should add some information to your readme about how to use the package.

FichteFoll:
You should take a look at `urllib.parse.quote_plus`.

s4wny:
Thx for the input. I have updated the readme and added `urllib.parse.quote_plus` now.

https://github.com/s4wny/HoogleSearch

s4wny:
Squashed commits into one.

joneshf:
This is everything I need in life! Thanks!

s4wny:
@joneshf Glad to hear :)

PR #4272, opened 2015-03-30
Added aeQuery package

kyletmartinez:


wbond:
The readme says the package includes syntax highlighting, but all I see are snippets.

Also, you should add a description to the GitHub repo because it will show up in the search results on packagecontrol.io.

What do you think about an alternate name, such as `ExtendScript Snippets`. Right now without the description and the current name, I don't think ExtendScript users will be able to find this. I tried googling for `aeQuery` and it didn't seem to bring up anything related to ExtendScript either.

kyletmartinez:
Great feedback! I based my original README off of another similar package so I will make those changes. I will add a description as well. I didn't quite know how Package Control worked.

Changing the name is perfectly fine with me! I suppose mimicking jQuery was more self-gratifiying than anything else. I'll make those changes and push it. I'll let you know when I am done!

kyletmartinez:
I submitted a new request with the name change, description added, and README edit. Let me know anything else that you think I need to do!

wbond:
@kyletmartinez In the future, you can just push new commits to the branch that was used for a PR instead of openeing a new one.

PR #4275, opened 2015-03-31
adding Tint to repositories

yyyc514:


wbond:
You should remove the explicit `"readme"` entry from the pull request. Package Control will automatically find it via the `"details"` entry.

yyyc514:
Removed "readme".

FichteFoll:
I do not see a specific reason for your `.no-sublime-package` file. If you don't need it, please remove it as it limits [customization by users](https://packagecontrol.io/docs/customizing_packages).

yyyc514:
Perhaps I'm confused.  Why does GitSavvy have this file?

FichteFoll:
I don't know, it doesn't look like it'd need one either. I think I just missed the file in the review.

@divmain?

yyyc514:
I thought maybe because all the files are split out (vs a single py file) that the package format wouldn't work... are there any good docs and when it's required not required?

divmain:
@FichteFoll, it was necessary in an early version but shouldn't be any more.  It'll be removed in the next release.  Thx!

FichteFoll:
@yyyc514 not yet, but https://github.com/wbond/packagecontrol.io/issues/46 should give you a good starting point.

wbond:
The reason for a `.no-sublime-package` is for shared libraries (`.dll` on Window, `.so` on Linux and `.dylib` on OS X) and executables since they can not be loaded from a zip. In some previous versions of Sublime Text 3 there were some assets related to themes that were not loading from zips, but this is no longer the case.

yyyc514:
Removed.

PR #4277, opened 2015-03-31
Clean up labels a bit and see if contributions here are welcome.

yyyc514:


yyyc514:
Maybe "c" is actually more correct - but the single letter really bug me when the wrong case... it just looks like it's wrong.

wbond:
Thanks for taking a stab at this.

Since URLs are case sensitive, obviously the `C` and `c` tags are currently different. I am wondering if we should take the approach of trying to implement "proper" casing for everything, or just force it all to lowercase? Currently it seems that most everything is lowercase.

@FichteFoll do you have any thoughts on this general topic?

FichteFoll:
Since normal words in English are lower case anyway, it seems to be the easiest choice. The only remaining words are names (e.g. for languages).
That would mean with proper case we'd have Markdown, JavaScript, Java, C, C++, Underscore.js. My concern with this is that you'd need to know the proper case of all the things and this is where things become ... harder. Especially if we wanted to test casing of labels.

I'm for lower casing all labels for mainly the following reason: It's harder to fuck up and easier to maintain. That means users can not accidentally create new labels by different casing and we can just test if labels are lower case instead of testing if the label already exists with different casing.

Either way, we should definitely unify labels. Since this would be good to do programmatically, we could as well unify indentation and whitespace usage across all files. I know that this is inconsistent currently.

yyyc514:
@FichteFoll @wbond Ok, so lowercase all labels, regardless of length?  What are the rules regarding names that are obviously camel-cased now?  CartoCSS, OpenStreetMap.  A bit of something is  lost there in the translation to lowercase I think... Do we adopt dashes or underscores or some other character mapping?

FichteFoll:
Yes, something is lost, but it's the lesser evil imo. `cartocss` or `openstreetmap` still looks reasonably well to me.

Not sure about underscores or hyphens in labels, but I prefer to not use them. Need to check if they are currently used at all.

I created #4310 with my ideas for an auto-formatter tool.

joneshf:
I'm a bit confused on the purpose of this unification. Is it to make label searching easier on the website? Couldn't the website do a case insensitive search or use some more in depth matching algorithm? Or is there another reason for it?

yyyc514:
To make searches easier, clean up inconsistencies, remove meaningless labels.  The case thing is just one aspect of the discussion (albeit the one getting all the attention).

If I remove the C/C++ case changes (going the wrong way) can this be merged?

wbond:
Well, labels aren't currently used for searching, just browsing. Unification means there won't be a list of two packages with `c` and three more under `C`.

repository/m.json
```diff
@@ -277,7 +277,7 @@
 		{
 			"name": "MarkdownEditing",
 			"details": "https://github.com/SublimeText-Markdown/MarkdownEditing",
-			"labels": ["markdown", "github flavored markdown", "github markdown", "gfm", "multimarkdown", "color scheme"],
```
FichteFoll: Why did you remove `github flavored markdown` here?

yyyc514: Judgement, the word flavored doesn't really improve on the other two.

yyyc514:
> Unification means there won't be a list of two packages with c and three more under C.

Yep, that'd be an improvement.

repository/m.json
FichteFoll: Hm okay, `github flavored markdown` provides roughly the same information as `github markdown` as a label, so we only need one. Technically the former is the correct one, but the latter is shorter.

FichteFoll:
I'm okay with merging this. It changes casing of a few labels, but we are going to change that anyway eventually. If you feel like removing those commits then go ahead, otherwise I'm with this.

yyyc514:
C++ and C commits have been removed;  Lets get it merged. :rocket:

FichteFoll:
Sneaky Tint addition here, but I just merged it anyway.

yyyc514:
> Sneaky Tint addition here, but I just merged it anyway.

Sorry, that was an oversight.  Not on purpose.  Either it got there when I rebased or it was there all along and I didn't notice because I branched from that branch.

Thanks though.

PR #4278, opened 2015-03-31
Add Laravel Forms Bootsrap to repository

redgluten:


wbond:
You'll need to change `"branch": "master"` to `"tags": true` and use tags to created releases. As mentioned in the docs, we no longer accepts branch-based releases for new packages.

Also, in your readme/description it indicates this package is for ST3, however your pull request indicates it also works with ST2. One or the other should be fixed.

redgluten:
Ok, I’ll make a new pull-request with the changes then.

PR #4279, opened 2015-03-31
Added LiveStamps Plugin Meta

TundraTech:


wbond:
You had the version operator correct the first time. What you've specified right now will only allow it to be installed on ST2 and the very first build of ST3, which I don't believe was ever available to the public.

TundraTech:
Sorry, it uses async_timeout so anything above that version is fine

wbond:
Sorry, it looks like I missed this the first pass through. We no longer accept new packages that use branch-based releases. So you'll need to change `"branch": "master"` to `"tags": true` and create a tag in your repo every time you want to create a new version for users to download.

The tags must follow the semantic versioning (semver) specification. I.e. a three part number such as `1.0.0`.

TundraTech:
Thanks will I created a 1.0.0 release and added tags key

PR #4281, opened 2015-04-01
Add Monokai JSON+ theme by @ColibriApps

mikebranski:


wbond:
@mikebranski Unfortunately we no longer accept branch-based releases. Instead, a semver tag needs to be created for each releases. E.g. a tag such as `1.0.0` or `1.0.5` need to be made for a new release to be installed on users machines.

The only other thing I would want is @ColibriApps to confirm that they are ok with this being added to Package Control.

wbond:
To further specific, changing to tag-based releases means that `"branch": "master"` needs to be changed to `"tags": true`.

mikebranski:
Hey @wbond, sorry about that. Believe it or not, I did read the contribution docs – I copied an existing recipe, but forgot to address the tagging bit before I submitted! I've updated the recipe and pinged them to tag the release.

Also, yes, @ColibriApps is all aboard: ColibriApps/MonokaiJsonPlus#1

wbond:
Once there is a semver tag, I can merge this.

benmarten:
ok, i've added the `1.0.0` tag

PR #4288, opened 2015-04-04
Add 'MantisBT' package

vboctor:
MantisBT package for administrators and developers

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -153,6 +153,17 @@
 			]
 		},
 		{
+			"name": "MantisBT",
+			"details": "https://github.com/vboctor/sublime-mantisbt",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Since this package seems to apply to only a very small group of people (developers working on the mantisbt code) I would suggest having potential users use the "Add Repository" command from package control which allows to add any package that is hosted on github/bitbucket or even custom stuff with a `repositories.json` file, even when they are not on the official list.

@wbond?

vboctor:
@FichteFoll The goal of this plugin is to also be helpful to all developers who install MantisBT on their own servers.  The plugin will help with customizations and editing of configuration files.  MantisBT is one of the most popular open source bug trackers and it would be great for developers installing it to get some additional assistance within Sublime.

It would be nice for MantisBT extension to show up in the package manager to help with discoverability.

PR #4291, opened 2015-04-05
Symfony2 Override package added

igormukhingmailcom:


igormukhingmailcom:
@wbond Hi.
Sorry for the rush, but how long can that operation take?
Thank you.

FichteFoll:
I usually review daily but I've been on vacation.

You can append commits to existing pull requests by pushing onto the branch you opened it from. Just FYI. Also, is there a certain reason why this is marked as linux-only?

igormukhingmailcom:
@FichteFoll 
Sorry, I does not noticed that branches is deprecated. Fixed now.

It marked as Linux-only because at this time it tested only on Ubuntu.

PR #4295, opened 2015-04-08
Added Theme - Numix Light

ikenfin:


FichteFoll:
In your readme you mention the package name `Theme - Numix Light` but in this pull request it is `Numix Light Theme`. I suggest you stick with the former since that is the generally accepted conevention of naming themes (and similarly color schemes).

ikenfin:
Fixed it @FichteFoll

FichteFoll:
FYI, your currently highest tag (1.0.0) does not point to the latest commit and still contains the color scheme.

ikenfin:
Yes, i have updated tag just now. Thanks!

PR #4297, opened 2015-04-08
Add whitequark/rainbowth

whitequark:


FichteFoll:
You'll need to change `"branch": "master"` to `"tags": true` and use tags to create releases. As mentioned [in the docs](https://packagecontrol.io/docs/submitting_a_package#Step_4), we no longer accepts branch-based releases for new packages.

Converns regarding your code:

1. You're leaking the `base_dir + '/Rainbowth/Rainbowth.cache'` and the first `scheme_file` file handle.
2. You could use `os.path.join(sublime.cache_path(), 'Rainbowth', scheme_name + '.cache')` instead of the weird path math that injects a folder into whatever folder of the currently used scheme, which might as well not actually exist because it resides in a `.sublime-package` archive. It also does not use "hard-coded" forward slashes.
3. Instead of `pickle` you could also use the json library since all your exported data appears to be serializable (that is, lists, dicts and strings).
4. You could defer your `on_modified` callback by using a [timeout thread](https://gist.github.com/FichteFoll/0694dadafe51eb493b47) or similar mechanisms (easiest is using `on_modified_async` and a `threading.Lock` object) which would reduce the performance penalty.

FichteFoll:
@whitequark ping

whitequark:
@jrgdiz has solved the first two issues but not second two. I don't have time to work on them right now though.

FichteFoll:
Yeah, I noticed the pull requests. The bigger issue is that you're using `"branch": "master"` however. I can merge this once you corrected that and pushed a tag.

whitequark:
Done

FichteFoll:
The value of the tags key should be a boolean true (since a string `"true"` means something different).

whitequark:
Oops, that was dumb. Fixed, rebased

FichteFoll:
Um yeah, [semantic versions](http://semver.org) consist of exactly three numeric parts. Sorry for not spotting that earlier.

whitequark:
Done

PR #4298, opened 2015-04-09
add sublime WinSCP

thecotne:


FichteFoll:
Please consider adding the package to the default repository files directly and migrating to the 3.0.0 schema. It makes maintaining easier for us, in case of a schema version update for example, since packages are in a centralized place.

PR #4301, opened 2015-04-10
Added a basic YAPF command

ondoheer:
First version of the plugin, plan to keep upgrading it

### DIFF
--- a/repository/y.json
+++ b/repository/y.json
@@ -35,6 +35,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
I just merged https://packagecontrol.io/packages/PyYapf%20Python%20Formatter (from #4285). Can you take a look at that and contribute there via pull request or by other means? ST3 support would be useful I suppose.

ondoheer:
Brilliant!will do!

PR #4303, opened 2015-04-11
Added SwigSnippets

thecodechef:
Hey Thought This Might Swig Users


thecodechef:
Help Swig Users

FichteFoll:
You'll need to change `"branch": "master"` to `"tags": true` and use tags to create releases. As mentioned [in the docs](https://packagecontrol.io/docs/submitting_a_package#Step_4), we no longer accepts branch-based releases for new packages.

PR #4306, opened 2015-04-11
Added Axosoft package to the repo

ckaznocha:
Added Axosoft package to the repo

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -2610,6 +2610,17 @@
 			]
 		},
 		{
+			"details": "https://github.com/ckaznocha/SublimeAxosoft/",
+			"homepage": "http://ckaznocha.github.io/SublimeAxosoft/",
+			"labels": ["axosoft", "ontime", "issue tracker"],
+			"releases": [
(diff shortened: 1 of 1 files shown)


FichteFoll:
You don't need the _ssl binaries since Package Control provides them to all installations now (since PC3.0.0). You can also use the [reqeusts dependency](https://github.com/packagecontrol/requests) instead of bundling your own.

That means the only library you still need is `lib/axosoft/axosoft_api/`, which you can import using relative imports (`from .lib.axosoft import axosoft_api`). In fact, since your package currently also includes many other files from the axosoft repository I suggest you just extract the relevant python files into a `axosoft_api` subdirectory of your repository root and include licensing information.

After all this, you don't need `.no-sublime-package` and don't need to mess with `sys.path` in your package.

ckaznocha:
I didn't know about the requests dependency from package control. Thanks for the tips @FichteFoll, I've implemented all your suggestions. Hopefully everything is good to merge now.

PR #4292, opened 2015-04-06
RPN Calculator

bphunter1972:
An RPN Calculator with programmer, scientific, and statistical modes.

PR #4313, opened 2015-04-13
RPN - Reverse Polish Notation Calculator

bphunter1972:
Supersedes pull request #4292. Let's hope this one's better.

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -894,6 +894,17 @@
 			]
 		},
 		{
+			"name": "RPN",
+			"details": "https://github.com/bphunter1972/RPN",
+			"labels": ["calculator"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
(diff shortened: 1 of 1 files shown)


FichteFoll:
Okay, so far so good, but [semantic versions](http://semver.org) consist of exactly three numeric parts.

bphunter1972:
Done. That’s the last time I listen to GitHub’s tagging recommendations.

FichteFoll:
It's not that github's recommendation is wrong. There are many different versioning schemes.

Also, it is better if you know what semver actually is (I suggest reading the spec it if you didn't already) before unconsciously using versions that conform its standard. It won't really matter for package control since packages can not be used as dependencies and semver is mainly targeted towards libraries that can be depended on, but I now use it for everything and try to find the best fit for each project.

bphunter1972:
Merely being facetious. Thanks for you help.

PR #4316, opened 2015-04-14
Chinese-Localization

rexdf:
Chinese Translation for Sublime Text. Support MainMenu TabMenu ContextMenu,etc

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -543,6 +543,18 @@
 			]
 		},
 		{
+			"name": "ChineseLocalization",
+			"details": "https://github.com/rexdf/Chinese-Localization",
+			"labels": ["localization", "chinese", "中文"],
+			"description": "Chinese Translation for Sublime Text. Support MainMenu ContextMenu TabMenu ,etc. 中文翻译，主菜单，右键菜单，标签菜单。",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Chmod",
 			"details": "https://github.com/kristoformaynard/SublimeChmod",
 			"releases": [
--- a/repository/d.json
+++ b/repository/d.json
@@ -216,18 +216,6 @@
 			]
 		},
 		{
-			"name": "Default",
-			"details": "https://github.com/rexdf/Chinese-Localization",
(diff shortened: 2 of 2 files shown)


FichteFoll:
You may not use package names of default packages because it may cause inintended side effects. You need to think of another way of overriding the menus.

rexdf:
Can you explain me how to hook the `plugin-finish-load` event? I find **emmet** and **package Control** have this hook. But I do not real understand.

FichteFoll:
I suppose you mean `plugin_loaded`? That is needed because the APIs provided by the `sublime` module are not always available when plugins are loaded.

rexdf:
I have tried in a different package name.  It will add the same id twice an english one and one chinese one.

Default is package namespace, if sublime find a differnet caption in different package, it create double items.

I try to load the `Main.sublime-menu` when all other package finished load just like **Package Control** checking network for updating.

Some package has an items like following. It will override my `Main.sublime-menu`. So I hope to load last. And the `Default/Main.sublime-menu` is the fit one.

```json
"caption": "Preferences",
"mnemonic": "n",
```

I have no other way to override the exist **Main.sublime-menu** by dynamic way. Maybe I should check about how **Package Control** disbale package Menu and enable it.

rexdf:
I just rename `Default` --> `Default2`.  Double items.
sublime_trans2

FichteFoll:
No, this is not what I meant. You are right, you need to override the default menu files by placing files somewhere in the "Default" package, but you need to do it in a different way than adding the package as "Default" to package control. Not only does that restrict any other localizations (since there can only be one package for each name), it also interferes with the user potentially overriding the default package on his own already. And this is all talking about ST3 only, because in ST2 this would go horribly wrong due to bad interactions with Package Control.

> e.g. a plugin that copies the menu files to the desired location.

rexdf:
Ok. I got it.

rexdf:
Add a python script to load the Main.menu

FichteFoll:
That looks better :+1: 

You can access files in ST archives like this: `sublime.load_binary_resource("Packages/ChineseLocalization/EN.zip")`. If you do that you don't even need the `.no-sublime-package` file anymore, but I will accept this beforehand.

PR #4317, opened 2015-04-14
Added Sphere Online Judge Plugin

geekpradd:
This plugin is used for displaying Sphere Online Judge (competitive programming) problem contents so that both programming and problem reading can be done together.

The plugin is ready for release but there is one feature I need assistance on. I'm using a textarea to display the problem which is fetched from a remote API. How do I display an Image in the textarea since some problems have images in them? Except that everything is built and ready for deployment.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1836,6 +1836,16 @@
 			]
 		},
 		{
+			"name": "Sphere Online Judge",
+			"details": "https://github.com/geekpradd/Sphere-Online-Judge-Sublime",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags" : true
+				}
+			]
+		},
+		{
 			"name": "Spice",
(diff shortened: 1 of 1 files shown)


FichteFoll:
There are only two ways to display images in ST currently.

1. Open an image file in ST as a dedicated file, this will trigger the preview. (>=3055)
2. `` tags in `view.show_popup`. (>=3072)

geekpradd:
The second option looks good but how do I position the view over the textarea panel? Any suggestions?

Or, can I add an onclick event over the image url in the textarea so that when the URL is clicked the image is loaded as a new file?

FichteFoll:
By textarea panel you mean an output panel? Should work just like with any other view.

Yes and no. Yes, you can add a mouse map binding but since they don't support contexts it will override the default bindings. I suppose the best way would be to do it in a context menu where you get event information, see `Default/open_context.url` for an example.

PR #4318, opened 2015-04-14
add APICloudSnippets plugin

taxusyew:


FichteFoll:
If you can you should add `<scope>` restritctors to your snippets because currently they are active for all languages.

taxusyew:
Thanks for your advice, I have changed the `<scope>` restritctors in my snippets.
And the tag has benn added.

FichteFoll:
There is no tag on https://github.com/taxusyew/apicloud-sublime-snippet/tags however.

taxusyew:
Sorry about that, my mistake, is that all right now?

PR #4321, opened 2015-04-15
ChineseLocalization support sublime text 2

rexdf:


FichteFoll:
`os.unlink` is not available on OSX. Use `os.remove` instead.
Or you could use a `io.BytesIO` object and pass that to the ZipFile contrstructor since it's a file-like object.

Since you have two separate codebases for ST2 and ST3 you don't need the version tests in the ST2 code or attempt to call load_binary_resource. Unless the user installs manually or copies the folder from an ST2 to an ST3 installation (which is 100% improper), there won't be any situation where the code would be run on ST3.

rexdf:
Thx, I'll try.  Maybe I can change it to something like i18n. But it need a time to finish all translation.

PR #4322, opened 2015-04-15
Include the repository to the Portuguese spelling language dictionary

alessandroleite:


FichteFoll:
Portuguese dictionaties are also available at https://github.com/titoBouzout/Dictionaries, but it has a rather large overhead (see titoBouzout/Dictionaries#42).

Until that issue is solved, packages for single or few dictionaries are legit. In any case, please consider adding the package directly to the channel and not via a self-hosted packages.json (that uses an outdated schema version) and switch to schema version 3.0.0 this way.

FichteFoll:
Since @titoBouzout rejected my suggestion, the only remaining issue is that you're using a custom `repository.json` where none is needed.

FichteFoll:
@alessandroleite ping

PR #4324, opened 2015-04-15
Added Tagger Script Syntax

phw:
See also http://forums.musicbrainz.org/viewtopic.php?id=5652

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -107,6 +107,17 @@
 			]
 		},
 		{
+			"name": "Tagger Script Syntax",
+			"details": "https://github.com/phw/sublime-tagger-script-syntax",
+			"labels": ["MusicBrainz", "language syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Tagify",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Out of curiousity, is this the same syntax that is also used in foobar2000 and http://www.mp3tag.de/en/index.html?

I've wanted to create a proper syntax for that for a while (3 years to be exact) but eventally lost sight of it.

phw:
I'll update the tag.

The foobar2000 syntax was the basis for Picard's scripting, so this should work in general. However I don't know foobar 2000 well enough to be sure there is no difference. Give it a try ;)

phw:
I've updated the tag

FichteFoll:
Well, the current syntax is pretty minimalistic so I'm sure it will work just fine. I had a more advanced one with deprecation warnings even and different highlighting for generally known tags (e.g. `%artist%`). I know that foobar supports `%album artist%` however, which your definition doesn't match currently.

phw:
> I know that foobar supports %album artist% however, which your definition doesn't match currently.

No, than this is one of the minor differences between Picard and foobar. If I find the time I would like to extend this syntax, maybe it could even offer variations for foobar and MP3tag.

Thanks for the quick merging :)

PR #4325, opened 2015-04-15
Update i.json

MtimerCMS:
Change repository address


FichteFoll:
Did you know that you can change your github username and even transfer ownership of a repository (which includes updating all references and http 301 redirects)?

PR #4327, opened 2015-04-15
Transparency

rexdf:
Make the windows transparency

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -1737,6 +1737,20 @@
 			]
 		},
 		{
+			"name": "Transparency",
+			"details": "https://github.com/rexdf/SublimeTextTrans",
+			"author": "vhanla",
+			"labels": ["theme"],
+			"description": "Make the windows transparency",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": "windows",
+					"tags": true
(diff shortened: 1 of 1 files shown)


FichteFoll:
Generally, if possible, please try to get in contact with the original author of the package first. There is even a ticket for this, but it doesn't seem to have gotten any attention: https://github.com/vhanla/SublimeTextTrans/issues/1.
cc @vhanla

That aside, there is a package for this already at https://packagecontrol.io/packages/GlassIt, but it uses a different and less comfortable solution for the user (by depending on an external binary that needs to be downloaded), so adding this package would be fine.

Other general tip:
- When changing indentation (as in https://github.com/rexdf/SublimeTextTrans/commit/2971a5bf491b76ac7c0d85d7054a92c9f58608eb), do it in a separate commit and don't mix code changes in. Commits like this are unreviewable.

rexdf:
I just make a fix to separate indentation and change codes. And try to get in contact with @vhanla .

FichteFoll:
Any updates?

rexdf:
No response  from @vhanla .

FichteFoll:
Can't help him then, merged the PR.

PR #4330, opened 2015-04-16
Edit txt2pynb to master branch default

jmwerner:


FichteFoll:
Branches for releases are deprecated and won't be accepted anymore. It's only applicable for old packages that were using it already.

PR #4334, opened 2015-04-16
New theme - BlueBerriesForest

Winns:


FichteFoll:
1. The package is not [in order](https://travis-ci.org/wbond/package_control_channel/builds/58746794#L153).
2. The above doesn't matter anyway because your package actually only contains a "color scheme", not a theme (by ST terminology).

FichteFoll:

FichteFoll:
@Winns  ping

PR #4336, opened 2015-04-16
Add Susy 2 Snippets package to proper json file

kyleshevlin:
New package of snippets for Susy 2, not to be confused with currently available package for Susy One.


FichteFoll:

kyleshevlin:
Updated version to 1.0.0

PR #4338, opened 2015-04-17
Added HipsterIpsum

seripap:


FichteFoll:
You must remove the `self.view.end_edit` call, it will fail on ST3.

Anyway, a package named "[HipsterIpsum](https://packagecontrol.io/packages/Hipster%20Ipsum)" exists already, so you'll need to choose a different name. I suppose it's pretty "hip" to write your own [ispum package](https://packagecontrol.io/search/ipsum).

FichteFoll:
"Hip Ipsum" and "HipsterIpsum" are still too similar, imo. I'd prefer if you could come up with a different name (and maybe also rename your files).

seripap:
@FichteFoll Just sharing this package for the fun of it- was intended to be quick and simple for my workmates and such. I'll just close it out, I don't see the point in putting anymore effort into another ipsum package.

PR #4344, opened 2015-04-20
Add productive snippets for Ruby and ERB

janlelis:
This commit adds new snippets for Ruby and Embedded Ruby.

It is intended as a replacement for the stock Ruby snippets and avoids gem specific code (e.g. code that is only meant for Ruby on Rails).

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -1346,6 +1346,28 @@
 			]
 		},
 		{
+			"name": "ProductiveSnippetsERB",
+			"details": "https://github.com/janlelis/productive-sublime-snippets-erb",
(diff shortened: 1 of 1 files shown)


FichteFoll:
What's the difference between your ERB snippets and https://packagecontrol.io/packages/ERB%20Snippets?

Also, you can remove the `issues` keys since these are the defaults, grabbed from the details url.

janlelis:
Thanks for the response. I removed the issues keys.

I agree that both ERB packages have a similar scope. Besides some details (triggers etc), the main difference is that the referenced package also includes Ruby on Rails specific snippets (about half of them), which this package avoids.

PR #4345, opened 2015-04-20
Adds TrailingPunctuation plugin

spadgos:
https://github.com/spadgos/TrailingPunctuation

> ### Trailing Punctuation
> This is a Sublime Text 3 plugin which will add or toggle punctuation at the end of a line.

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -1735,6 +1735,16 @@
 			]
 		},
 		{
+			"name": "Trailing Punctuation",
+			"details": "https://github.com/spadgos/TrailingPunctuation",
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
Interesting idea. I suggest using the `;,` list of puntiation to toggle between (or rather: to replace) as an argument so that it can be used for more than just these two characters.

spadgos:
>  I suggest using the ;, list of puntiation to toggle between (or rather: to replace) as an argument

I can't think of many other uses for it, so I'll wait to see if any user requests come in for that feature.

The alphabetisation has been fixed now.

PR #4348, opened 2015-04-21
added insert-semicolon package

Azeirah:
Created a package to insert a semicolon at the end of the line by double-tapping the ; button

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -424,6 +424,17 @@
 			]
 		},
 		{
+			"name": "Insert-semicolon",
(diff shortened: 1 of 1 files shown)


FichteFoll:
This is what https://packagecontrol.io/packages/Trailing%20Punctuation does as well, except that it doesn't create a new mark and considers comments at the end of the line (and can swap between , and ;). I just added that yesterday, but there is also https://packagecontrol.io/packages/Semicolon.

PR #4350, opened 2015-04-21
Added PHP Source package

mjkaufer:


FichteFoll:
Instead of your very weird `isPHPFile` function you can just use the following: `sourceName.endswith(".php")`

You're leaking a file handle, which causes issues on Windows. Use the `with` statement instead.

mjkaufer:
Thanks for those catches! I just implemented your suggestions. Anything else?

Thanks!

FichteFoll:
You don't have to close the file handle if you are using it in a `with` statement. In fact, I don't even know if it might actually throw an exception if you do that.

And finally, you need to push a new tag with these fixes.

mjkaufer:
I don't think python throws an exception, but I removed the redundant code regardless. Sorry about forgetting the version.

Thanks!

mjkaufer:
I actually got mixed responses on the error thing. It throws an error in the python console, but not when I run the code from the sublime plugin. Anyways, it's fixed regardless.

PR #4352, opened 2015-04-22
Add CJSX Syntax

kwood:
JSX/Coffeescript/CJSX syntax highlighting was [removed from sublime-react](https://github.com/reactjs/sublime-react/commit/bd1d7b7c059721b9baf8acff9840ba8ab9d35143) in favor of babel-sublime, which [does not intend on supporting CJSX](https://github.com/babel/babel-sublime/issues/96).

This is a port of the existing CJSX syntax highlighting with a few improvements. I plan on adding some other CJSX features to this repo over time.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -708,6 +708,17 @@
 			]
 		},
 		{
+			"name": "CJSX Syntax",
+			"details": "https://github.com/Guidebook/sublime-cjsx",
+			"labels": ["CJSX", "syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Clafer Tools",
 			"details": "https://github.com/gsdlab/ClaferToolsST",
 			"labels": ["Clafer", "Compiler", "Instance Generator"],


FichteFoll:
Before I add this I'd like to know what users of CJSX rather prefer, having their own package or adding this to the Better Coffeescript package (via https://github.com/aponxi/sublime-better-coffeescript/pull/209). Personally, I suppose I would want to have it as its own package since it doesn't appear to have much to do with regular coffeescript and it's probably easier to maintain this way, assuming that @aponxi doesn't use CSJX.

Please keep me updated on changes of the situation, e.g. in the above pull request and https://github.com/babel/babel-sublime/issues/96.

-

kwood:
It doesn't look like it's a priority to better-coffeescript, and the babel issue has been closed with this being the preferred solution.

I would push for merging.

FichteFoll:
Usually I would give @aponxi a bit more time to respond, but since currently there is essentially no CJSX syntax on package control I agree with merging earlier. We can later merge the package into better-coffeescript any time.

kwood:
Sorry about that -- tag corrected.

PR #4353, opened 2015-04-22
Add sublime-repository-snippets

frdmn:


FichteFoll:

Please consider removing the "Sublime" part of the package. I know that this is targetted towards Sublime Text package developing, but I suggest something along the lines of "Package Snippets", or just adding them to https://github.com/SublimeText/AAAPackageDev/ (which sounds like a good idea to me).

frdmn:
@FichteFoll Updated both. Is that okay?

PR #4354, opened 2015-04-22
Added Flat Chameleon UI Theme

r3a1ay:


FichteFoll:

r3a1ay:
`"tags": true` added.
Should I fix something else?

r3a1ay:
Thanks.
I have fixed some bugs in my theme and added new tags (v1.2.2 - the latest/added 2 hours ago).
[Package control](https://packagecontrol.io/packages/Flat%20Chameleon) page updated readme section twice, but has not updated the version of the package.
Is something wrong in my tags?

FichteFoll:
All the tags you pushed have dots after the `v`. You can make Package Control understand that (by setting `"tags": "v."`, but I suggest just removing the dots instead.

r3a1ay:
Thanks.

PR #4357, opened 2015-04-23
Add Chinese Lorem Ipsum plugin

cjltsod:
Add Chinese Lorem Ipsum plugin

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -567,7 +567,7 @@
 					"tags": true
 				}
 			]
-    	},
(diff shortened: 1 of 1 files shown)


FichteFoll:
You can remove the file `package.json`, it is redundant and won't be looked at.

[Semantic version](http://semver.org) consist of three numeric parts (for your tags).

cjltsod:
Thanks for the revise. I've made the changes.

PR #4359, opened 2015-04-23
add Youdao Dict entry in /repositories/y.json

divinites:
I developed a small plug-in for OS X users who want to use the command prompt window as a translator between Chinese and English.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -953,6 +953,16 @@
 			]
 		},
 		{
+			"name": "CN Dictionary",
+			"details": "https://github.com/divinites/cndict",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "COBOL Syntax",
 			"details": "https://bitbucket.org/bitlang/sublime_cobol",
 			"labels": ["language syntax", "completion"],
--- a/repository/y.json
+++ b/repository/y.json
@@ -134,8 +134,7 @@
 			"releases": [
 				{
 					"sublime_text": ">=3000",
-					"platforms": ["osx"],
-					"branch": "master"
+					"tags": true
 				}
 			]
 		},
--- a/repository/y.json
+++ b/repository/y.json
@@ -129,6 +129,17 @@
 			]
 		},
 		{
+			"name": "Youdao Dictionary",
+			"details": "https://github.com/divinites/YoudaoDict",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"platforms": ["osx"],
+					"branch": "master"
+				}
+			]
+		},
+		{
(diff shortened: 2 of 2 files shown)


FichteFoll:
`sys.exit()` don't do that in a plugin, ever! (just enter `import sys; sys.exit()` in the sublime console to see what happens)

I don't really see a reason for this being available on OSX only, the code looks very cross platform. If you just made it because you couldn't test it on the other OSes then let me assure you that it will very likely just work.

Finally, you need to change `"branch": "master"` to `"tags": true` and use tags to create releases and push valid [semantic version](http://semver.org) tags to your repo.. As mentioned [in the docs](https://packagecontrol.io/docs/submitting_a_package#Step_4), we no longer accepts branch-based releases for new packages.

FichteFoll:
I just noticed that there is a package like this already: https://packagecontrol.io/packages/Youdao%20English-Chinese%20Translate by @friskfly

How is that package different to yours, appart from not being marked as ST3-compatible?

divinites:
Hi, FichteFoll,

Many thanks for your comment.  I noticed this plugin when I decided to program my own. There are mainly three reasons for that:

1. As you mentioned, it is not sublime text 3 compatible, and the developer did not update it since two years ago.

2. We use different ways to display the result. That plug-in will initial a new view every time and make it active when it tries to translate a word, so users will have to switch back and forth between the new view and the working document.

I use the command prompt windows as the default output window, so it is less interruptive..

3. Besides the dictionary API that is used in that plug-in, my plug-in has more options. In the 0.2.0 version that I just uploaded, users can choose their preferred dictionary APIs by modifying “cndict.sublime-settings”.


best wishes  —T.W.

FichteFoll:
I don't see a reason for these two packages to be separated.

@friskfly, do you have any interest in having @divinites replace your package on Package Control or adding him as a contibutor to your repository so that he can implement his improvements there?

divinites:
I do not know if I should start a new fork or continue this one. I wrote a lightweight email-agent for sublime text, called oath-mail, which use oauth authentication  to write and send emails.

FichteFoll:
Please open a new pull request for the new package (and remove the commits from this one if you can, otherwise we could get merge conflicts).

FichteFoll:
@friskfly, please comment on this issue. Otherwise we will assume that you have no interest in having the package listed on package control anymore (I can see that you are still active on github).

divinites:
Thanks for your suggestion, the commitment has been reverted, now I will start a new pull request for another packages.

friskfly:
Sorry for seeing this issue now , since I haven't updated this program for two yeats, I agree to replace my package。

divinites:
As FriskFly has agreed to use my package to replace his, could you please merge this pull request as well? I have updated CN Dict to v.0.5.0, using output_panel, instead of command prompt to display the translation output.

PR #4361, opened 2015-04-23
add Fortran package

315234:
I'm not sure I've specified the ST version number requirement correctly. This syntax highlighting requires at least build 3084.

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -948,6 +948,16 @@
 			]
 		},
 		{
+			"name": "Fortran",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Should be `>=3084`, otherwise your package was uninstallable until the next build.

315234:
I'm not sure if you have been notified of the efc94e5 so perhaps this has fallen of the radar. Can this be merged now?

FichteFoll:
Indeed, I have not. Github does not notify about comments and did not consider adding that either after my requests.

PR #4363, opened 2015-04-24
Update repository.json

FengRR:
I want to connect sublime for Cloudcc

### DIFF
--- a/repository.json
+++ b/repository.json
@@ -30,6 +30,7 @@
 		"./repository/x.json",
 		"./repository/y.json",
 		"./repository/z.json",
-		"./repository/dependencies.json"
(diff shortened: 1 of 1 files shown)


FichteFoll:
Just add the package to one of the files in `repositories/`, use the 3.0.0 schema, push a valid [Semantic version](http://semver.org) tag with three numeric parts and remove `packages.json` from your repo.

Also, the commands you register in the menu and key binding are not defined in your plugin code (and don't exist on my ST install), so everything your plugin is doing currently is removing double line breaks from the clipboard and pasting that.

FichteFoll:
@FengRR ping

PR #4366, opened 2015-04-24
add PyCrypto dependency

jlegewie:
Currently only compiled for osx.

### DIFF
--- a/repository/dependencies.json
+++ b/repository/dependencies.json
@@ -59,6 +59,21 @@
 			]
 		},
 		{
+			"name": "PyCrypto",
+			"load_order": "50",
+			"description": "Python Cryptography Toolkit - https://www.dlitz.net/software/pycrypto/",
+			"author": "jlegewie",
+			"issues": "https://github.com/jlegewie/sublime-PyCrypto/issues",
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
You only have `st3_osx_x64/Crypto` so I suppose this should be for `"platforms": ["osx-64"]`.

jlegewie:
Done.

FichteFoll:
My bad, `osx-x64` it is.

jlegewie:
Okay, back to "osx" and I squashed the commits to keep this clean...

FichteFoll:
Why back to "osx"? It's clearly not available for osx-x86.

I mean it's not too big of a deal since the worst thing to happen is that, if specified, Package Control just installs the dependency for osx-x86 and it just does nothing, but an actual error would certainly be better.

jlegewie:
I guess I got confused. I thought Travis chocked on `osx-64` and didn't see the `x`. Now it should be fine.

wbond:
Just for the record, there is no ST3 OS X x86. There is a just a single OS X version, and it is 64bit: http://www.sublimetext.com/3.

FichteFoll:
Well ... TIL I guess.

PR #4370, opened 2015-04-27
add AutoAligner package

generall:
Package for automatic code aligment based on BNF and Needleman–Wunsch algorith.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -1195,6 +1195,18 @@
 			]
 		},
 		{
+			"name": "AutoAligner",
+			"details": "https://github.com/generall/aligner",
+			"author": "Andrey Vasnetsov aka generall",
+			"labels": ["code alignment", "alignment", "code style"],
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
`path = sublime.packages_path() + "/aligner"` will fail because the package will be saved as `AutoAligner` (as specified in this PR).

Your `Popen` call will also fail on ST3 because it is package into a `.sublime-package` there. Add a `.no-sublime-package` file to the repo to tell PC to unpack it.

```
				req_version = (2,7)
				cur_version = (sys.version_info.major, sys.version_info.minor)
				if cur_version > req_version:
```
could also be written as `if sys.version_info > (2, 7):` (even though I think you mean `>=`.

generall:
Thanks for review. Remarks are fixed

FichteFoll:
[Semantic versions](http://semver.org) have three numeric parts.

generall:
Fixed

PR #4373, opened 2015-04-28
Unity 3D Shader: Add issues and move to tags releases

petereichinger:


FichteFoll:
You don't need the issues url, it wil be detected automatically (and has been before).

petereichinger:
I removed the line (see last commit).

Maybe I didn't look in the right place but maybe the Docs section should be updated?

FichteFoll:
Not sure what you used, but it's mentioned [here](https://github.com/wbond/package_control/blob/d3bc04fd5c0c4bdb476a3ea798e6815508dc9bac/example-repository.json#L133-L135).

petereichinger:
I seem to have misread that comment...

Thanks!

PR #4374, opened 2015-04-28
Relax version constraint for Subclim

lanthaler:
It is also supported for Sublime 3 as the author of the package confirms here:

https://github.com/JulianEberius/Subclim/issues/19#issuecomment-37093846


FichteFoll:
cc @JulianEberius

PR #4375, opened 2015-04-28
ludik st3 support added

jonny64:


FichteFoll:
Is it intended that you want to restrict this package for windows only? You still have Linux key binds.

jonny64:
Yes. Our group don't test (and use) Linux version for about a year.

PR #4376, opened 2015-04-28
add package named by SublimeHydra

iamfredng:
this is a Sublime plugin of Lua

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -1019,22 +1019,22 @@
 			]
 		},
 		{
-			"name": "LuaSmartTips",
-			"details": "https://github.com/haoyue90/LuaSmartTips",
-			"releases": [
-				{
-					"sublime_text": "*",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "LuaSimpleComplete",
 			"details": "https://github.com/iamfredng/LuaSimpleComplete",
 			"releases": [
 				{
 					"sublime_text": "*",
 					"platforms":["windows", "linux", "osx"],
+					"tags": true
+				}
+			]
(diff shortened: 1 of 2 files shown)


FichteFoll:
I don't particularly see a reason for this to not be available for Linux. I assume you just didn't test it there? It shouldn't be a problem to allow linux users to install this as well.

iamfredng:
I don't have Linux. So I not sure it work for Linux.

FichteFoll:
I can assure you that this will work just fine on Linuxs. I don't have access to all the different OSes either, but since Python is cross-platform you will be fine, as long as you are't doing platform specific stuff which this does not seem to do.

iamfredng:
Ok. I will add Linux platform to My setting.  Thanks a Lot.

FichteFoll:
I just noticed it, but your package violates our [naming policies](https://packagecontrol.io/docs/submitting_a_package#Step_2).

FichteFoll:
@iamfredng ping

iamfredng:
Can I just change my package name and repush?
I want the new package named LuaSimpleComplete, Is OK?

iamfredng:
@FichteFoll I rename the old plugin to LuaSimpleComplete

FichteFoll:
Technically that is all that was necessary, but practically your latest commit seems to have caused a merge conflict with current `master`. Please fix the conflict by squashing and rebasing.

iamfredng:
@FichteFoll Thanks for removed the stale label. The conflict is fixed.

FichteFoll:
"merged" separately

PR #4378, opened 2015-04-29
Eberstark Group Inc. Channel addition.

markdleblanc:
The EGI Channel introduces the SyncR package which offers bi-directional SCP, Profile, and DIFF support.


FichteFoll:
I am not sure about the time_modified format restrictions, but I suggest you add a leading zero for the hour.

Also, please use tabs for indentation in the channel. (I will add a test for that later.)

markdleblanc:
Added the leading 0 and fixed the spacing issue as per your request.

Thanks.

PR #4379, opened 2015-04-29
add "Coq Improved" Plugin entry

izgzhen:


FichteFoll:
If this is "not just duplicate", what does it do differently then? What prevents you from filing a pull request? Why are there three more or less active forks (including original) of the same thing?

https://github.com/izgzhen/Coq-Improved-Plugin
https://github.com/whitequark/Sublime-Coq
https://github.com/mkolosick/Sublime-Coq

izgzhen:
@FichteFoll The major difference in my version is that I write the major body of tmLanguage (the original version is nearly none), and completely drop the python interface to `coqtop` (and recommended sublimeREPL instead). The major reason why I don't choose to pull request on the whiteshark's repo, is because of this. I don't think it is a good idea to pull to the original repo with a large portion of code deleted directly. Thank you.

FichteFoll:
I'd like to hear what @whitequark has to say about this.

whitequark:
@izgzhen Please send a PR with tmLanguage changes, at least. 

I don't think SublimeREPL comes even close to replacing the coqtop wrapper, so that should stay. I don't see a reason to drop it, anyway

whitequark:
Oh, and @mkolosick's repo has all my changes at the moment, so the two are equivalent. I'm not sure which one is 'canonical' but I'll at least try to keep them in sync.

izgzhen:
@whitequark OK, I got it. P.S. about the `coqtop` wrapper, I can see that a lot of excellent ideas were put into it such as "nextStatement". But it can't run more normally, so I tried sublimeREPL and that turned out to b

whitequark:
Well, SublimeREPL and SublimeCoq are not mutually exclusive, you can use whatever you see fit! But I think it would also be nice if you filed a bug at mkolosick/Sublime-Coq explaining your issue.

PR #4394, opened 2015-05-04
Add LazyTimeTracker

Bwata:
Would like to add my sublime package to Package Control. If there are any issues please let me know, this is my first publish package.


FichteFoll:

Bwata:
FichteFoll I added the tag. Is that all I needed to do?

FichteFoll:
You still need to do the first part.

Bwata:
Ok, Sorry I missed that.

PR #4395, opened 2015-05-05
Add sublime-hashids

zergin:
Add package for encoding and decoding hashids (from hashids.org)

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -203,6 +203,17 @@
 			]
 		},
 		{
+			"name": "Hashids",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Unfortunately you can not use submodules when specifying `details` to have package control auto-fetch the latest release. See https://github.com/wbond/package_control/issues/126 and https://github.com/wbond/package_control/issues/825.

zergin:
I see. I've changed submodule to subtree - should be working fine now (at least the downloaded release archive contains hashids library).

zergin:
Thanks.

PR #4396, opened 2015-05-06
Added Intellitip plugin

jbrooksuk:


FichteFoll:
You should try to read the data files using `sublime.load_resource` and not `open(some_relative_path)`, which will then allow you to remove the `.no-sublime-package` file. However, I'll add the package before you did that.

FichteFoll:
Also, you're leaking the file handle in https://github.com/jbrooksuk/Intellitip/blob/985f107b4461926ed0520420584e2ec98686d1d6/intellitip.py#L47

PR #4397, opened 2015-05-06
Add Miva Template Language (MVT) Syntax package

mghweb:


FichteFoll:
Your scope name should be different from html since it is not exactly html. Something like `text.html.mvt` should do, which will preserve potential configuration targeted at the HTML syntax.

mghweb:
The scope has been updated.

PR #4399, opened 2015-05-07
Add Cucumber Snippets e Highlight em Português

RobertoPegoraro:
Cucumber Snippets e Highlight em Português

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2663,6 +2663,16 @@
 			]
 		},
 		{
+			"name": "Cucumber Snippets e Highlight em Português",
+			"details": "https://github.com/RobertoPegoraro/cucumber-snippets-highlight-portugues",
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


FichteFoll:

@wbond was there anything regarding non-ASCII characters in package names?

RobertoPegoraro:
I created a new release in my repository with version 1.0.0, I think that's what you asked for, right?

FichteFoll:
Yes. I'd like to wait for @wbond's reply however.

wbond:
The only issue I know of with non-ascii characters is that the command palette does not strip diacritics when searching, so if you have a package named `Résumé`, you would have to type `é` and not just `e` to match it.

RobertoPegoraro:
This is a problem I need to change?

FichteFoll:
Not really. It just means that you can't find this package by searching for `Portugues`.

PR #4403, opened 2015-05-08
Added Starbound Lua syntax plugin

UnknownX7:


FichteFoll:
I suggest you rename your syntax file to `Lua Starbound.tmLanguage` and change the base scope to `source.lua.starbound`.


UnknownX7:
Hi, thanks for the help and tip. I've updated my plugin.

PR #4405, opened 2015-05-08
Update w.json

splincode:


FichteFoll:
That is the wrong file, it should be `a.json` obviously.

FichteFoll:
There also already is a package like this: https://packagecontrol.io/packages/auto-save

PR #4407, opened 2015-05-08
Auth mail branching

divinites:
This is a lightweight email agent for sublime text3. it use oauthv2-authentication so users need to register an app on corresponding website before they can use it. I also wrapped the oauth2 into the plug-in.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -953,6 +953,16 @@
 			]
 		},
 		{
+			"name": "CN Dictionary",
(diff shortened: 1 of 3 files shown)


FichteFoll:
I suggest you squash or reset this branch since you have quite a few unnecessary commits and github reports a file mode change.

divinites:
I start a new pull request #4408.

best wishes

PR #4408, opened 2015-05-08
Add oauth-mail entry

divinites:
I start a brand-new fork to initiate the pull request, hope this time it would be fine.

This is a light-weight email agent, use oauthv2 as the authentication method.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2663,6 +2663,16 @@
 			]
 		},
 		{
+			"name": "Cucumber Snippets e Highlight em Português",
+			"details": "https://github.com/RobertoPegoraro/cucumber-snippets-highlight-portugues",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Cucumber Step Finder",
 			"details": "https://github.com/danielfrey/sublime-cucumber-step-finder",
 			"releases": [
--- a/repository/o.json
+++ b/repository/o.json
@@ -13,6 +13,16 @@
 			]
 		},
 		{
+			"name": "Oauth Mail",
+			"details": "https://github.com/divinites/oauth-mail",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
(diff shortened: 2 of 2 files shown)


FichteFoll:
You'll need to change [this line](https://github.com/divinites/oauth-mail/blob/f69d7381eaa0c1786ece91f172e0d23d1f20693e/mail.py#L18) to respect the actual package name, which will be "Oauth Mail" when installed via package control. Or you just use the global `__package__` attribute which Sublime Text provides since ST3.

divinites:
Hi, FichteFoll,

Thanks for your reminder, I have committed the change. 

_SYNTAX_PATH = globals()['__builtins__']['__package__']+"/mail.tmLanguage”

best wishes

divinites:
Hi, Fichte Foll,

How about the new version? Also, the original pull request for cndict has been removed, should I restart another one?

FichteFoll:
I suggest you re-check [this](https://github.com/divinites/oauth-mail/blob/39a62fe52c1bdd489edf2475a27df12ecb3b04f3/mail.py#L23) line.

Your import management is questionable, especially inserting the package dirname into sys.path. Just use relative imports instead so you can remove the sys.path code from all three files (you can remove it from two already anyway).
Additionally, I recommend placing modules that do not use the sublime interface into subdirectories so that ST will not attempt to load them and check for commands or event listeners.

divinites:
Hi, FichteFoll,

That line you spotted is a mistake because of shortcut keys… I have corrected it.

For sys.path manipulation, I also think that it is a very ugly workaround. Now I have adopted your suggestion and removed all sys.path sentence. 

Also, I create a sub-folder “libmail" inside the package, so that only one .py file in the root path of the package. 

Thank you very much for your help, this is the first time I try to develop a formal python project for users other than myself.  I really learned a lot from you comments.
 

best wishes —T.W.

FichteFoll:
You're welcome.

There is no real documentation on these sorts of things. I intend to write a plugin dev guide on http://docs.sublimetext.info/ but haven't come around to do that yet since there's a lot of other things we need to do there.

PR #4414, opened 2015-05-12
Added KIXtart package

vipex:
Added language syntax, build system and snippets package for KIXtart language.

### DIFF
--- a/repository/k.json
+++ b/repository/k.json
@@ -286,6 +286,17 @@
 					"branch": "master"
 				}
 			]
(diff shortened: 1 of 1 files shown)


FichteFoll:
Note that by default your package won't be installable because the only available version is a pre-release. Users have to opt-in to install pre-release versions of a package using the `install_prereleases` setting.

vipex:
Good to know, next update will be production.

Thanks.

PR #4416, opened 2015-05-12
Add Monokai sidebar theme

joshuahiggins:


FichteFoll:

joshuahiggins:
Thanks for the info, I missed that before. I've added tags to the repo and updated this pull request.

PR #4422, opened 2015-05-14
Add a GN package for the GN build system used by Chromium

eseidelGoogle:
The README at https://github.com/eseidelGoogle/gn_sublime
explains what GN is and how to learn more information.

GN is used by several hundred engineers working on Chromium
and associated projects, many of whom use Sublime.  Having this
package in Package Control should make it much easier for these
users to have proper highlighting and comment control for
.gn and .gni files.

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -632,6 +632,16 @@
 			]
 		},
 		{
+			"name": "GN",
+			"details": "https://github.com/eseidelGoogle/gn_sublime",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Gnuplot",
 			"details": "https://github.com/hesstobi/sublime_gnuplot",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Actually, Package Control does not require you to host the package on github, it's just the easiest way. You may also host the package as an archive ony any (https serving) server, as well as a .json file that tells PC the correct metadata which it would fetch from the github API otherwise.

[Here](https://csch1.triangulum.uberspace.de/release/packages.json) and [here](https://packagecontrol.io/packages_3.json) (which are referenced in [`channel.json`](https://github.com/wbond/package_control_channel/blob/master/channel.json)) are some examples.

If you want to continue hosting this on github instead of managing it on your own you need to push a semver versioned tag to the repository.

eseidelGoogle:
I've pushed a 0.0.1 Tag.  Please take another look.

PR #4424, opened 2015-05-15
Added Trainz Config to packages list

Dangoo:


FichteFoll:
The readme key is redundant.

Dangoo:
Should be fixed by now, v0.1.0 added

PR #4431, opened 2015-05-18
Added FileDelete plugin

jbrooksuk:


jbrooksuk:
Hang on... I broke some formatting.

jbrooksuk:
Fixed :)

FichteFoll:
Line 8 will throw an exception if run on an unsaved view. Also, I'm not sure that I'd make the answer recognition case sensitive.

jbrooksuk:
I didn't know about Delete Current File :)

PR #4432, opened 2015-05-18
A MarkDown inspired PlainTasks theme for Sublime Text.

ronilaukkarinen:


FichteFoll:
How about adding the scheme (it's a "color scheme") to the https://github.com/aziz/PlainTasks repo directly? cc @vovkkk

vovkkk:
Sure, we accepted all proposed schemes so far.
On the other side, it is simpler for author to change it, keeping scheme in own repo.
Either way is okay for me.

FichteFoll:
Same here, I can see pros and cons for either option. Your call, @ronilaukkarinen.

ronilaukkarinen:
Hello. Yes, seems logical. This is my first package, so was not quite sure
what to do. The reason I am implementing this is that I noticed my theme
getting erased every time after updates and of course the wider audience
can be achieved this way. I am going to send this directly to PlainTasks,
so this can be closed. Thanks a bunch!

PR #4436, opened 2015-05-20
Added CenterComment

coder-mike:
Any feedback is welcome (including an existing package that I've missed that already does this)

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -312,6 +312,16 @@
 			]
 		},
 		{
+			"name": "Center Comment",
+			"details": "https://github.com/coder-mike/sublime-center-comment",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "cexio-ticker",
 			"details": "https://github.com/iceydee/cexio-ticker",
 			"releases": [


FichteFoll:
I don't know of a package that does this.

I suggest you add an option for the coulmn width and support for more than just "//"-style comments. Check out https://github.com/gkhn/SectionComment/blob/f94b87b82ff261b009a402710a43af362234376b/SectionComment.py#L6-L35, which is a rip from `Default/comments.py` (the command responsible for toggling comments).

Also, you have to push a valid [semantic version](http://semver.org) tag (with optional leading `v`) to your repo.

coder-mike:
Great, thanks for the feedback. I've made the changes you suggested, including adding a settings file with "line_width", defaulting to what I was using before which was the "wrap_width" in the global settings. I also used the code at the link you provided to dynamically extract the comment tokens for the current language. I was hoping to do this from the beginning, but couldn't find out how until you suggested that reference. I've tested it with C++, python, and HTML comments, and I presume it will extend naturally to others. I've also started using the semantic versioning format for tags, starting at v1.3.0.

FichteFoll:
Btw, on ST3 you could also just do `from Default.comment import build_comment_data` since that's where the other plugin ripped it from in the first place. It will only work on ST3 however since ST2 doesn't expose its plugins to be importable like that.

PR #4438, opened 2015-05-20
Added VintageousPluginSurround plugin

asfaltboy:
May resolve guillermooo/Vintageous_Plugin_Surround#7

### DIFF
--- a/repository/v.json
+++ b/repository/v.json
@@ -331,6 +331,16 @@
 			]
 		},
 		{
+			"name": "VintageousPluginSurround",
+			"details": "https://github.com/guillermooo/Vintageous_Plugin_Surround",
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
cc @guillermooo

You'll need to change `"branch": "master"` to `"tags": true` and use tags to create releases and push valid [semantic version](http://semver.org) tags to your repo. As mentioned [in the docs](https://packagecontrol.io/docs/submitting_a_package#Step_4), we no longer accepts branch-based releases for new packages. This also applies to package you don't own.

FichteFoll:
ping @guillermooo, @asfaltboy

asfaltboy:
@FichteFoll I've no commit bit on that repo and so cannot create a tag ... I've also pinged the repo owner over at the issue

PR #4440, opened 2015-05-20
ClangAutoComplete: Typo in platforms support field.

pl-ca:
There was a small typo in the platforms array. It is expected that the platform names be separated by quotes.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -789,7 +789,7 @@
 			"releases": [
 				{
 					"sublime_text": ">=3000",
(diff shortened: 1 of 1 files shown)


FichteFoll:
I wonder why the test suite didn't catch it. Will need to revisit.

Anyway, you can remove the platforms entry entirely since it is set to `"*"` by default (which is the same as what you want to accomplish).

pl-ca:
Thanks for the update, I have removed the platforms entirely.

PR #4444, opened 2015-05-22
Update Derby Package Location

rossedman:
- Update Locations
- Add tags support
- Add label

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -302,12 +302,13 @@
 		},
 		{
 			"name": "Derby - Bourbon & Neat Autocompletions",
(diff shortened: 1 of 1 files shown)


FichteFoll:
While you.re at it, you should consider naming the package just "Derby" since the package name is not a location for a description (that's what the description from your package repo is for, currently "Bourbon & Neat Sublime Text Snippets & Auto Completions "). See https://packagecontrol.io/search/derby.

rossedman:
@FichteFoll Awesome. Thanks for point that out. Opened a PR [here](https://github.com/wbond/package_control_channel/pull/4445)

PR #4447, opened 2015-05-24
Add Code Cast to package manager repository.

mraiur:


FichteFoll:
Unfortunately, 1.0.0.0 is not a valid [semantic version](http://semver.org).

As a tip, a few words describing your packages (and setting a repo description) would help discoverability and help users understand quickly what it does, without watching the video.

mraiur:
Updated the tags from v1.0.0.0 to 0.1.0 ( its a minor release ) and updated the README. 
Thanks for for the feedback!

PR #4449, opened 2015-05-24
Adds Ember CLI snippets

jhliberty:


FichteFoll:
How are these different from https://packagecontrol.io/packages/Ember.js%20Snippets ?

jhliberty:
@FichteFoll - Thanks for the response! I've changed branch master to tags true, I'm already using tagged releases so that works great for me!

They are different in the fact that these snippets are for users utilizing Ember CLI instead of plain Ember.js - Ember-CLI utilizes es6 modules instead of global variables and `export default` based on the file name and typically requires an import statement.

I'm also going to be adding additional snippets for Ember-CLI that non CLI ember users won't even have access to such as ember-cli-http-proxy/ember-cli-mock-proxy for instance.

PR #4450, opened 2015-05-25
Adding apigen

dans98:
I've created a simple plugin that allows anyone to quickly generate php source code documentation, from within Sublime. It allows a user to run ApiGen from the context menu, & the side bar.

The plugin allows users to have multiple apigen.neon files thought-out their project. This allows them to generate small sections of documentation quickly during the development process.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -785,8 +785,8 @@
 			"details": "https://github.com/dans98/Sublime-ApiGen",
 			"releases": [
 				{
-					"sublime_text": "*",
-					"branch": "master"
+					"sublime_text": ">=3000",
+					"tags": true
 				}
 			]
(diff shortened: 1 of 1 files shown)


FichteFoll:
You'll need to change `"branch": "master"` to `"tags": true`. As mentioned [in the docs](https://packagecontrol.io/docs/submitting_a_package#Step_4), we no longer accepts branch-based releases for new packages.

dans98:
It's updated now.

Sorry I missed that, I was following [example-repository.json](https://raw.githubusercontent.com/wbond/package_control/master/example-repository.json) that said it was only deprecated.

 I also limited it to 3, even though it would probably work fine in 2.

PR #4451, opened 2015-05-25
Add CnDict Entry

divinites:
because the previous cndict branche (CN<=>EN dictionary) has been deleted, so I am afraid pull request #4359  is no longer available, so I create a neat pull request here. I've made some improvements since then.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1054,6 +1054,7 @@
 		{
 			"name": "Cndict",
 			"details": "https://github.com/divinites/cndict",
+			"previous_names": ["Youdao English-Chinese Translate"],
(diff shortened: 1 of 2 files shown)


FichteFoll:
If you want to properly replace @friskfly's package you need to:

1. Delete his package from the json files ("Youdao English-Chinese Translate").
2. Add the package name to your package's `previous_names` array.
3. Prepare an update message for users migrating to your package so that they get notified about the change. (refer to https://packagecontrol.io/docs/messaging)

divinites:
Have done what you said, Thanks a lot for your help!

PR #4460, opened 2015-05-29
Added Chatscript syntax repo the the channel

joshuasnowball:
Allo. Here's a pull that include my syntax definition to the default channel. Cheers!

https://github.com/joshuasnowball/chatscript-tmlanguage

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -467,6 +467,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
0.9 and 0.8 are not valid [semantic versions](http://semver.org). Package Control will not see those releases.

PR #4464, opened 2015-05-30
Add SelectionPythonEval to Package Control.

jfcherng:
https://github.com/jfcherng/SelectionPythonEval

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -630,6 +630,17 @@
 			]
 		},
 		{
+			"name": "SelectionPythonEval",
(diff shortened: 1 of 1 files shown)


FichteFoll:
https://packagecontrol.io/packages/Insert%20Nums can do this just as well, except that it has additional syntax and features.

I'm also quite sure that there is some other [eval](https://packagecontrol.io/search/eval) package.

jfcherng:
Just tried [insert Nums](https://packagecontrol.io/packages/Insert%20Nums ) and it worked even better!
I agree with your point.

PR #4467, opened 2015-06-01
Added StyleTooltip dependancy

huot25:


FichteFoll:
That's cool, but won't work.

1. `readPlistFromBytes` is Python 3, same as `sublime.load_resource`
#. The ST name for this is "popup" (due to `view.show_popup`) so it would be preferrable to use that instead of "tooltip", at least in a dependency/module.
#. Sublime will never load the Python file by itself since dependencies are not plugins. The file will only be executed if a plugin uses `import styled_tooltip`, in which case the commands will not be exposed and not be made available by ST. Same for `plugin_loaded`.  
Instead, you should provide an API that goes like `def stylized_popup(view, content, *args, **kwargs):`, where args and kwargs are forwarded to the following `view.show_popup` call. That way developers can also provide on_click callback, location or basically any other parameter.
#. `run_on_scheme_change` in `Preferences.sublime-settings` is a very generic name that does by no means provide context about what it is actually about. Same for the other settting.
#. You don't add event listeners to settings keys, you add event listerners to entire files and provide a key to remove the listener later (using `settings.clear_on_change`).
#. The usage of the term "theme" in various places is very misleading since you are actually parsing the active color scheme.
#. Color schemes can be changed by syntax-specific, project-specific or view-specific settings. You only consider the global settings file however. Use `view.settings().get("color_scheme")`.
#. In that regard, instead of parsing the scheme file on startup or Preferences.sublime-settings setting change you should store a hash of the currently used color scheme of the view the API I mentioned earlier is executed on, check if you currently have an entry for this in some temp place and regenerate the .css file if the scheme's content changed or it's a new scheme.

Hopefully you can use this.

huot25:
Thanks @FichteFoll for the feedback!

1. You're absolutely right about this. This plugin was supposed to be for ST3 only, but I overlooked the Sublime Version when I added my entry in package control. I'll be sure to update that.
2. Good point about the naming. I've been so focused on getting the logic working that I wasn't watching the naming. Too much Twitter Bootstrap recently that tooltip got stuck in my brain.
3. I think I am going to run with your idea here and make that a method to be called vs a command. All of the view.show_popup parameters are supported by my code currently, but I definitely see making this independent of the sublime plugin flow as a better option. Plus using **kwargs will make this less coupled to the current function parameters.
4. I will do some research to see how other developers are naming their settings. I had just thrown these together, and should spend more time naming them. 
5. I see now. I had implemented the setting changed logic last week and I read the API as key within the file to watch, not a key to identify the callback. Thank you for clarifying.
6. I will make sure I have updated references to theme so it is less confusing.
7. I never change the color scheme aside from defining the default color scheme and I would have never thought of testing for view specific color schemes. That's really good to know, Thanks!
8. Based on your feedback I think I have an approach that will work. I've some notes to implement the changes and will post and update once they are done.

huot25:
Changed this plugin to work as an importable module based on @FichteFoll's recommendation. The plugin now parses the active view's color scheme if the color scheme has not already been parsed and stores the value down to a hash table.

FichteFoll:
Looks pretty good. Few things:

- I suggest you don't hook into `sublime_plugin` however to register an event hook and instead compare hashes when `show_popup` is called.
- You allow for a `view` parameter in `SchemeParser.load_color_scheme` but you don't supply one when calling it. I highly suggest working on "arbitrary" views and only falling back to `sublime.active_window().active_view()` if none was supplied in the initial `show_popup` call, if at all. I think you can trust the user here to only call it with valid view objects. I would only pass the color scheme path to the Parser anyway.
- Your method name "parse_plist" is misleading since the plist has already been parsed in `load_color_scheme` and you are working on pure Python objects from there on.
- Instead of accessing `os.sep` you could use `os.path.join`.
- Since your "depencency" is obviously only available for ST3 I highly suggest renaming the subfolder from "all" to "st3" to better reflect that.

FichteFoll:
@huot25 ping?

I guess the module is functioning, but these few improvements would be nice. I won't block merging this, but I'd appreciate if you at least commented on my feedback.

huot25:
@FichteFoll 

I plan on getting back to this soon and I am going to implement your suggestions. My son has been in and out of the hospital for the past few weeks and that has taking priority over everything else which is why I haven't responded. He is doing better but still rough going. I should have time over the next few days to get back to this provided everything continues to go well.

FichteFoll:
Sorry to hear that.

No worries, I'm just cleaning up old issues like I do every once in a while.

huot25:
I've just implemented @FichteFoll's suggestions as well as refactored most of the code to be easier to follow.

FichteFoll:
Looking good here. However, it seems that a remote change caused a merge conflict with this, so could you please rebase?

And finally, one last concern: You never pop anything from the stylesheets cache. It can potentially grow very huge. I advice adding some limitations for the number of stylesheets to cache in the file at any time, maybe something like 10, and then filter by LRU (least recently used). That requires either an additional key that stores a timestamp, or rearranging the stylesheet data into an array where new stylesheets get inserted at the top, moved to the top when used and the last (oldest) is poped when it grows too large.

FichteFoll:
Another note: The Python convention is to name modules in `snake_case`. I highly suggest complying to that.

huot25:
Rebased my pull request as well as implemented the changed recommended by @FichteFoll. Implemented a cache limit setting which is 5 by default.

FichteFoll:
That rebase looks more like a merge to me.

You definitely need to push a new tag however, and one that is a valid [semantic version](http://semver.org). Preferrably without a prerelease too, because I don't know how PC will interact with that for dependencies, although I suspect it won't be an issue.

huot25:
@FichteFoll 
I've just created 1.0.0 tag. 

I had tried to follow this [rebase guide](https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request), but my existing commits cause a conflict with the dependancies.json file. I merged the entry for my dependency in to correct the issue. 

I am not very familiar with git as I've only used other version control systems(TFS) until now. Do you have a better resource to reference? I am sure I will run into this again and I want to make sure I am doing everything properly when submitting to the repo.

Thank you for all of your help getting the module up to snuff.

FichteFoll:
The rebase guide looks fine. Are you sure you solved merge conflicts correctly and used `git rebase --continue`?

If I find the time tomorrow, I will cherry-pick your commits - or I will apply the changes manually, that will probably be easier.

huot25:
@FichteFoll 

I must have screwed something up along the way. Just now, I reverted to '71bcaa6', merged the changes from wbond's repo to get my copy up to speed. Then I reapplied my changes which I believe allows git to just do a fast-forward merge. I hope this makes your job a little easier and you wont have to do any extra work to merge this through. Sorry for all of the confusion.

PR #4469, opened 2015-06-01
Add Path Translator

hbhakhra:


FichteFoll:
I forgot to check, but [semantic versions](http://semver.org) consist of exactly three numeric parts.

hbhakhra:
I updated the version tag from v1.0 to v1.0.0

PR #4473, opened 2015-06-02
Added new package "Package Sync"

utkarsh9891:
Added a new package to the Sublime package listing.
This package takes back up of the currently installed packages on the
user's machine and allows him to port the same on other devices as well.
This also caters to preserving the user settings for ech package.

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -33,7 +33,7 @@
 			]
 		},
 		{
-			"name": "PackageSync",
+			"name": "Package Sync",
 			"details": "https://github.com/utkarsh9891/PackageSync",
 			"labels": ["backup", "restore", "sync", "package"],
 			"releases": [
--- a/repository/p.json
+++ b/repository/p.json
@@ -33,6 +33,18 @@
 			]
 		},
 		{
+			"name": "Package Sync",
+			"details": "https://github.com/utkarsh9891/PackageSync",
+			"labels": ["backup", "restore", "sync", "package"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"platforms": ["osx", "windows"],
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
`sublime.packages_path()` is not always available when the module is loaded. You need to place it in a global `plugin_loaded` function.

This looks more like a backup and rollback plugin that something that "syncs" packages.

utkarsh9891:
This plugin is still under development. I have currently posted the backup & restore part only.
Would be adding the _auto-sync packages_ part within the next two days.

Would update the `sublime.packages_path()` code as per your comments.

utkarsh9891:
This plugin is still under development. I have currently posted the backup
& restore part only.
Would be adding the *auto-sync packages* part within the next two days.

Can you please provide a plugin that actually uses the
sublime.packages_path() code inside a globalplugin_loaded function as per
your comments.

--
Thanks & Regards,
​Utkarsh Upadhyay
​,
Email: utkarsh9891@gmail.com​ | ​Phone​: +91-9953219784

FichteFoll:
Easy: https://github.com/wbond/package_control/blob/d65406ee9ff65032d4a55689f602c704dc6cb70b/Package%20Control.py#L120-L128

If you're curious, just restart ST and see in the console how `sublime.packages_path()` will return `None`, probably resulting in an exception thrown afterwards (don't remember your exact code).

utkarsh9891:
I have updated the package. Take a look if you need to.

As for the syncing part, would be adding the code in a few days time. But I need to test the installation aspects of the package as well.
Would be great if the pull request can be completed soon.

utkarsh9891:
Any estimate as to how much time it would take for this pull request to be approved???

FichteFoll:
FWIW, you can test installation before it's added to the defautl channel by specifying a URL to a custom `repository.json` or to the github repository directly using the "Add Repository" command.

PR #4475, opened 2015-06-03
updated MiniPy package info

vim-zz:
updated MiniPy info as it doesn't appear when searching the default channel


vim-zz:
it needs another change

PR #4476, opened 2015-06-03
updated MiniPy package info

vim-zz:
fixed github url due to user name change and some package details


vim-zz:
added another package

PR #4482, opened 2015-06-04
Add Smart Region

gbaptista:
Open and search files directly from your source code.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1433,6 +1433,17 @@
 			]
 		},
 		{
+			"name": "Smart Region",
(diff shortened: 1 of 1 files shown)


FichteFoll:
[this](https://github.com/gbaptista/sublime-3-smart-region/blob/6a1ebc59af7928431a061dd6cad6847090b47b62/SmartRegion.py#L34) should be `continue`. `next` is just a built-in function in Python.

gbaptista:
@FichteFoll tks! I will change it. =)

PR #4484, opened 2015-06-05
add scala syntax plugin

gwenzek:
I created a syntax plugin for Scala, using the recent .sublime-syntax format (build 3084).

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -178,6 +178,17 @@
 			]
 		},
 		{
+			"name": "Scala Syntax",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Variables in sublime-syntax have technically only been added in 3085, but that doesn't really matter.

However, [semantic versions](http://semver.org) consist of exactly three numeric parts.

gwenzek:
Thanks for the feedback.
I renamed the first release to [0.1.0](https://github.com/gwenzek/scalaSublimeSyntax/releases/tag/0.1.0)

PR #4494, opened 2015-06-07
Pull request for Redlime plugin

tosher:
Redlime - Redmine manager plugin for Sublime Text 3

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -514,7 +514,7 @@
 			"releases": [
 				{
 					"sublime_text": ">=3000",
-					"branch": "master"
+					"tags": true
 				}
 			]
 		},
--- a/repository/r.json
+++ b/repository/r.json
@@ -508,7 +508,7 @@
 			]
 		},
 		{
-			"name": "Redlime - Redmine manager",
+			"name": "Redlime",
 			"details": "https://github.com/tosher/Redlime",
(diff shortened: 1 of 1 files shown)


FichteFoll:

Also, the package name is more to be though of as a unique identifier and not a description of what the package does, thus I suggest to name the package just "Redlime". Your package will still be searchable on https://packagecontrol.io/search/ via "redlime manager" since it's in your repo's description.

tosher:
Without "redmine" in name, it will not be searchable in Sublime text via "Package control/Install package" by "redmine" substring :(
Redlime  - it's only funny name for Sublime/Redmine words, but it's [Redmine](http://redmine.org) manager :)

FichteFoll:
I know that. It's unfortunate that SublimeText does not allow us to have the second line searchable in a quick panel (https://github.com/SublimeTextIssues/Core/issues/688), but we would rather encourage people to use [the website](https://packagecontrol.io/search/) for discovering packages instead of searching within SublimeText itself, since it has numerous advantages. You get a lot more information about a package there compared to within ST, for example the target ST versions, the latest version for each release and when it was released and the readme.

tosher:
Ok, will waiting for quick panel upgrade :) Name was changed to "Redlime".

FichteFoll:
Thanks.

tosher:
Thanks!

PR #4495, opened 2015-06-08
adding AsciiConverter

penginryo:
sorry to push several times before I notice my mistakes...

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -1084,6 +1084,17 @@
 			]
 		},
 		{
+			"name": "AsciiConverter",
+			"details": "https://github.com/penginryo/AsciiConverterPlugin",
+			"labels": ["ASCII", "binary", "converter", "decimal", "hexadecimal"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"details": "https://github.com/SublimeText/AsciiDoc",
 			"releases": [
 				{


FichteFoll:
There is also https://packagecontrol.io/packages/hextoASCII, which is very similar but formats hex digits slightly differently.

I propose to add the functionality from that package to your package and then replace that package with yours, since it's more feature rich but has differences in formatting. A few options would probably be a good idea.

@ALLZ, what do you think?

penginryo:
Thank you for replying. It sounds good, however, it seems @ALLZ doesn't working anymore. If he doesn't reply, what should I do?

FichteFoll:
We'll wait for a little bit more. Considering vacations and other stuff, I expect most people to reply within 2 weeks, so 5 more days.

penginryo:
5 days have passed and @ALLZ still doesn't reply to us. What should I do?

FichteFoll:
Actually, the 2 weeks would be counting from the first mention (where he would get an email), not from the submission date of the pull request. That way my bad with the 5 days, so they actually still have another day to replay.

I just sent them a mail (found in readme) that points to this issue and gave them another 7 days. Replacing a package on the list is a serious matter.

In any case, you, @penginryo, can already implement the features from https://github.com/ALLZ/hextoASCII.

ALLZ:
Well, I found and read your mail, @FichteFoll, and still I don't get your problem, guys. If you want to use any functionality from my package, or fork it, or else - it's OK, feel free to do it. I haven't got any time now for support and upgrade it because of, well, different reasons. But I don't think it's important to replace my package since users can choose which one of them to use. You guys just acting like kind of raiders, you know?;)
If you find my package outdated, and **really** want to replace it  - okay, I'll respect your wish, but then you must remove package's entry from the proper .json in Package Cotnrol. Anyway, good luck!

penginryo:
First of all, I want to aplogize to you, @ALLZ, for trying to replace what you created without your consent. Since there are users using your package, and I'm the latter, let's not to replace your package, that's not fair to you and your users. We should think up other solutions regarding this matter, and so far I've got following solutions

1. Placing your package and my package individually.
2. Merging my package into your package and make it single one(I'll send a pull request).
3. Placing decimaltoASCII and binarytoASCII(ones I created) respectively so that your package won't be included.

If you have better ideas, please leave a message. I want to hear from both of you, @ALLZ and @FichteFoll, thanks.

ALLZ:
@penginryo, I don't blame you) At last, this isn't great work that worth fighting for. I haven't got any better solutions than you offered. I think, the way to place your package alongside with mine is good enough.

FichteFoll:

1. If the feature provided is trivial, the user is prompted with different packages all achieving practically the same thing and confused about which they should use.
2. If the difference is trivial, the user is again confused about which they should use, but the package authors or maintainers are also effectively doubling their work (of maintainance). If they worked together on the same package (either by collaborating or by proposing pull requests), the community would be.
3. If one of the authors or maintainers can not keep maintaining his version of a package, it is abandoned. With two or more authors collaborating the chance of this happening is reduced.

There are exceptions to this rule in cases where developers simply can't or don't want to work together due to certain reasons, or the packages themselves are different enough in that they provide different use cases and would attract different users. However, there might also still be the option of combining these kind of packages by including functionality of one in the other.

Either way, there is no definitive objective criteria here.

FichteFoll:
Do you have anything to say regarding my essay on duplicated packages? Imo, adding @penginryo as a contributor to @ALLZ's package sounds like the best plan currently.

Otherwise I'll just merge this in a while because I don't intend to block packages from existing, I just want to redirect people to contribute to existing packages if possible (or convince existing package authors to transfer their repos if they can't or don't maintain it anymore).

penginryo:
I'm totally ok with your opinion, @FichteFoll, but the problem is that if @ALLZ will accept it or not.

PR #4501, opened 2015-06-09
Added VEX package

teared:
Houdini VEX language definition, syntax highlighting and autocompletions for Sublime Text editor.

### DIFF
--- a/repository/v.json
+++ b/repository/v.json
@@ -190,6 +190,16 @@
 			]
 		},
 		{
+			"name": "VEX",
+			"details": "https://github.com/teared/vex",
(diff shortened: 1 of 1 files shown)


FichteFoll:
How does this compare to https://packagecontrol.io/packages/VEX%20Syntax?

teared:
This one is pretty experimental - I'm still working on it and plan to make a plugin with a small set of tools working with whole application, while some people may want only the language definition.

FichteFoll:
Please report back once that is the case.

teared:
I'll create a new request later.

PR #4502, opened 2015-06-09
Adding IntelXDK

likhter:


FichteFoll:
Is there a specific reason why you included ` .no-sublime-package`? If you use relative imports in ST3, such as `from . import XDKCore`, you don't need to mess with `sys.path`.

Also, your latest versioned tag is from march, so you might want to tag a more recent commit as a new release.

likhter:
@FichteFoll, thanks for you response,

``.no-sublime-package`` is presented since there is also file ``xdk_plugin.conf`` and plugin is going to store something there.
``from . import XDKCore`` does not work for ST2 unfortunately.

As soon as this pull request is approved we will add new README.md and new version tag.

FichteFoll:
It is not recommended to store temporary files in the package's directory if that data is supposed to persist across updates of the package. The reason is simply that Package Control removes all files in the package when it's updated. Instead, use the User package. (`os.path.join(sublime.packages_path(), "User", "xdk_plugin.conf")`).

```py
if int(sublime.version()) >= 3000:
    from . import XDKCore
else:
    import XDKCore
```

likhter:
Thanks again. Fixed everything you mentioned: now it's using "User" folder, removed path mess and .no-sublime-package file.

likhter:
Thanks, @FichteFoll

PR #4503, opened 2015-06-09
added LiveReloadMake

alessandro-pezzato:
please add my package "LiveReload Make extension for Sublime Text 3 "

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -663,6 +663,17 @@
 			]
 		},
 		{
+			"name": "LiveReloadMake",
+			"details": "https://github.com/alessandro-pezzato/LiveReloadMake-sublimetext3",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"platforms": ["osx", "linux"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "LiveScript",
 			"details": "https://github.com/paulmillr/LiveScript.tmbundle",
 			"releases": [


FichteFoll:
You can import a package's plugin code in ST3 via `from LiveReload import LiveReload` (as in `from <packagename> import <modulename>`. That doesn't require you to mess with the path (and maybe doesn't even work if LiveReload is in a .sublime-package archive (I'd need to test that)).

alessandro-pezzato:
Thank you, I've found it in [LiveReload plugin](https://github.com/dz0ny/LiveReload-sublimetext2/blob/devel/LESSPlugin.py). A simple `import LiveReload` is working fine.

FichteFoll:
Ugh, that's totally unnecessary. It's the equivalent of

```py
if int(sublime.version()) < 3000:
    import LiveReload
else:
    from . import LiveReload
```

cc @dz0ny

FichteFoll:
@alessandro-pezzato As a note, your code will likely fail to execute properly if @dz0ny fixes his importing code, since the way he does it currently overrides the internal `sys.modules["LiveReload"]` that is supposed to be the entire package with just the `LiveReload.py` module. Under normal circumstances you would have to use `from LiveReload import LiveReload`. The problem is that the normal code fails here since the overritten LiveReload module does not contain a submedule or attribute named "LiveReload".

Hope this is understandable.

alessandro-pezzato:
This is my first plugin for sublime and I've a little experience with python. So... which is the way? Your snippet isn't working for me (sublimetext 3083), only `import LiveReload` is working.

FichteFoll:
Yes, I noticed that after you mentioned it doesn't work. The reason is that LiveReload uses this method by itself and that interferes with Sublime Text's module management.

I suggest you use 
```py
try:
    import LiveReload
except ImportError:
    from LiveReload import LiveReload
```
to 1) work right now and 2) still work when the issue I mentioned is fixed upstream. Added https://github.com/dz0ny/LiveReload-sublimetext2/pull/111.

alessandro-pezzato:
Ok, just fixed as you suggested, thank you.

FichteFoll:
Ah sorry, I messed up. It should be the other way around since `import LiveReload` will work in any case.

```py
try:
    from LiveReload import LiveReload
except ImportError:
    import LiveReload
```

Merging anyway since it works currently.

PR #4516, opened 2015-06-14
Add package `DocBlockr Python`

adambullmer:


FichteFoll:
https://github.com/KristoforMaynard/SublimeAutoDocstring does this too. It supports different styles, including `'google'`, so I propose looking there first.

Additionally, did you consider adding this to DocBlockr directly? If yes, why didn't you? https://github.com/spadgos/sublime-jsdocs/issues/113

cc @spadgos

adambullmer:
**TL;DR** _Python is so drastically different than C based languages that it would require an almost entirely new plugin, or at the very least a rewrite to the inner workings of jsdocs to make it possible._

I thought about trying to create a new parser to fit in the jsdocss schema, but decided to roll my own to get it up and running first. Python offers a few more features that would require their package to be a little more robust: Keyword arguments, Decorators, Extending Classes, Yielding. It would also require a document parser that will read in both directions of the docstring: above for decorators and module definitions, and below to try and interpret class / module variables, return / yield values. jsdoc determines scope by brackets, as is the convention of the languages it currently supports, but python is an indentation based language, and needs different considerations to determine end of function or class.

I'm not opposed to contributing to jsdocs as it would greatly benefit this project, but I think this plugin needs more time to mature and be more feature complete. Also, a lot of collaborative work from the guys at @spadgos would be needed to rework the basic components they have established to support non C based languages.

I tried AutoDocstring, but I think it isn't quite as feature complete for generating the docstring tempalte as I was hoping this plugin to be, and would definitely prefer the plugin to function more like the jsdocs plugin does.

FichteFoll:
I agree with your reasoning. Maybe @spadgos has something to say about this, too?

>I tried AutoDocstring, but I think it isn't quite as feature complete for generating the docstring tempalte as I was hoping this plugin to be

What prevents you from filing a pull request against AutoDocstring?

>and would definitely prefer the plugin to function more like the jsdocs plugin does.

What do you mean exactly?

FichteFoll:
(submitted too early)

If you can, please try to extend existing packages first before working on your own. There are reasons that would prevent collaborating on a project, but in most cases it's just a combination of laziness or unwillingness to contribute to a non-owned repository. See bottom for copy-pasted reason for preventing duplicates

cc @KristoforMaynard 

---

spadgos:
It's true that it would take a significant effort to get non c-style languages supported in docblockr.

*completely tangential side note: you should consider using non-gendered pronouns when speaking of people of unknown gender (eg: users). Instead of 'he'/'his' just use 'they'/'their'.*

KristoforMaynard:
After a (very) cursory glance at the `DocBlockr Python` plugin, the functionality that I saw that it supplies in addition to those already in `AutoDocstring` are:

1. Inserting a snippet instead of editing the buffer. From a usability perspective I like this approach and think that would be a great feature to merge into AutoDocstring.
2. Extra parsing, such as pulling out the parent class from a class defintion. This seems useful.
3. An alternative (not Google or Numpy) formatting for the docstring.

If there is interest, I would love to discuss pull requests into `AutoDocstring` for these features, and ones that I no doubt missed.

abathur:
@spagdos :+1: for the pronouns aside.

FichteFoll:
I already fixed my template, so thank you for the aside, @spadgos.

Regarding this PR, @adambullmer, please get in contact with @KristoforMaynard and work on the existing plugin to make it the best Python docstring generator Sublime Text has ever seen.

Thus, I'm closing this.

PR #4518, opened 2015-06-15
Enabled Sublime Text 3 support for Snippet Destroyer

twolfson:
While transitioning to ST3, I found a solution to supporting Snippet Destroyer. As a result, I would like to allow installation for Snippet Destroyer in ST3.

Repo: https://github.com/twolfson/sublime-snippet-destroyer

Documentation on ST3: https://github.com/twolfson/sublime-snippet-destroyer/tree/0.1.1#sublime-text-3-setup

In this PR:

- Enabled all Sublime Text versions for Snippet Destroyer

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1558,7 +1558,7 @@
 			"labels": ["snippet", "delete", "destroy"],
 			"releases": [
 				{
-					"sublime_text": "<3000",
+					"sublime_text": "*",
 					"tags": true
 				}
 			]


FichteFoll:
FWIW, there are ways to "destroy" snippets in ST3 without having to extract all packages. See http://docs.sublimetext.info/en/latest/extensibility/packages.html#customizing-or-overriding-packages.

Basically, you'll want to use `sublime.find_resources()` to spot snippet files and then create empty files of the same names in the Packages dir to override them.

twolfson:
Ah, that is a better approach. I will definitely keep it in mind. Thanks =)

twolfson:
As a heads up, I eventually took on the override approach. It was a little more finnicky than desired (e.g. Sublime Text 2 needs to a snippet in a `.sublime-snippet` so we must unlink still, Sublime Text 3 requires something plist-esque in a `.tmSnippet` file although no snippet required) but it worked.

Thanks for the advice :+1:

FichteFoll:
Thanks for the info in edit. I'll make sure to add this to the [unofficial docs](http://docs.sublimetext.info).

PR #4519, opened 2015-06-15
Add PHP-Codebeautifier

Ennosuke:
Add the possibility to use phpcbf from within Sublime Text.

I know there is another beautifier for PHP but it's only for ST3 and the used beautifier has stalled development.


FichteFoll:
Since this package uses a different external beautufier, it's legit.

PR #4523, opened 2015-06-16
Add Emoji package

ethanal:
This is a plugin that allows users to insert emoji from the command palette.

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -391,6 +391,16 @@
 			]
 		},
 		{
+			"name": "Emoji",
+			"details": "https://github.com/ethanal/SublimeEmoji",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "EmoKid Color Scheme",
 			"details": "https://github.com/tdheff/EmoKid",
 			"labels": ["color scheme"],


FichteFoll:
`path = os.path.join(os.path.dirname(os.path.abspath(__file__)), "emoji-data.txt")` will fail when installed via package control, since that ships packages in `.sublime-package` files. ([docs](http://docs.sublimetext.info/en/latest/extensibility/packages.html#sublime-package-packages))

You should be using `sublime.load_resource("/".join("Packages", __package__, "emoji-data.txt")` instead (always forward slash is intended).

ethanal:
@FichteFoll I just made this change.

FichteFoll:
Additionally:

1. Your "insert_emoji" command is functionally equivalent to the built-in "insert" command.
2. Most of these don't render correctly on Windows (for obvious reasons):

  

3. Use `view.window()` instead of `sublime.active_window()`

ethanal:
1. True, but OS X's special character/emoji picker doesn't work in sublime
2. Would you recommend making it OS X only?
3. Fixed

FichteFoll:
1. >True, but OS X's special character/emoji picker doesn't work in sublime

  I don't understand that. OSX's emoji picker should be irrelevant for whether you are using the built-in "insert" or your redundant custom "insert_emoji" command.

2. I think it's still reasonable to have this on Windows since it does at least show some of the symbols and you can still insert them (which is the core functionality of this package).

ethanal:
Can you clarify what you mean by the built-in "insert" command/how it's accessed?

FichteFoll:
```py
view.run_command("insert", {"characters": "↔⇌↓←:⃗↑↕"})
```

ethanal:
Ah ok changed it to use `insert`.

FichteFoll:
In the end, I forgot to check the tags page ...

[Semantic versions](http://semver.org) consist of exactly three numeric parts. Also, you haven't pushed a tags for these fixes yet.

ethanal:
Fixed the tags.

PR #4530, opened 2015-06-18
PxRemTranslate plugin

youing:
My PxRemTranslate plugin was uploaded,please check it and confirm,thanks.

### DIFF
--- a/repository.json
+++ b/repository.json
@@ -5,7 +5,7 @@
 	"packages": [
 
 		{
-			"name": "Alignment",
+			"name": "PxRemTranslate",
 			"details": "https://github.com/youing/PxRemTranslate",
 			"author": "Paul Wang",
 			"releases": [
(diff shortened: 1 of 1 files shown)


FichteFoll:
This is obviously a fork of https://github.com/BsWiM/REM-PX, which seemingly has the same functionality and has been on Package Control for 2 years. **If** that wasn't the case, I **would** have told you this (consider it for the next time):

- Something went wrong completely, please restart from scratch.
https://packagecontrol.io/docs/submitting_a_package#Step_6

- You have to push a valid [semantic version](http://semver.org) tag (with optional leading `v`) to your repo.

- Please remove the `.pyc` file from your repo, it will be generated by ST2 on load and is not used by ST3.

- Please also remove `package-metadata.json`, this file is supposed to only be generated by Package Control.

PR #4534, opened 2015-06-18
Monkeypatch unittest on function call, not on load

FichteFoll:
Doing changes like these on module load is very intransparent and the
explicit call of the function telegraphs this way better.

Inspired by requests:
https://github.com/kennethreitz/requests/blob/9bbab338fdbb562b923ba2d8a80f0bfba697fa41/requests/__init__.py#L53-L54

cc @wbond


wbond:
Looks good to me. I pulled it down and ran the tests via ChannelRepositoryTools.

PR #4537, opened 2015-06-19
Added "Dota Lua VScript Snippets" Package

hex6:


FichteFoll:
What you are providing in your package is more commonly referred to as "completions" (with basic function signature) instead of snippets. Using `.sublime-completions` you can also combine all in one file and do not need to branch into >1500 other files.

See http://docs.sublimetext.info/en/latest/reference/completions.html

hex6:
Had no idea about that, thought that this was the only (but very ugly way) to do it. Thanks for the tip!

FichteFoll:

PR #4541, opened 2015-06-21
Pushing ModoPluginBuilder

alexwidener:
ModoPluginBuilder for generating plugins automatically within Sublime Text.


FichteFoll:

alexwidener:
Submitted and should be fixed and working now. Tag on repo is 1.0.0. 

Apologies, it is my first package. Thanks for the information.

PR #4543, opened 2015-06-21
added ZScript Highlighter

sonictk:
ZScript syntax highlighting for ZBrush for Sublime Text


FichteFoll:

You don't need the `.no-sublime-package` file, please remove it.

Your `packages.json` is also redundant, since it will not be looked at.

sonictk:
@FichteFoll Hi, thank you for the information, I have made the necessary changes to both the repository file and my own bitbucket repository as well. Please let me know if there is anything else I may have set up incorrectly for submission to Package Control. Thank you!

PR #4544, opened 2015-06-22
io Node Framework for nrf24le1 SoC chip by Nordic Semiconductor

justind000:


FichteFoll:
```
PLUGIN_NAME = "i/o node"
# [...]
sublime.load_settings(PLUGIN_NAME + ".sublime-settings")`
```

Yeah, idk about that.

You should search for templates using `sublime.find_resources` and `sublime.load_binaty_resurce` instead of opening the files directly since that

1. requires the package to be unpacked (which limits configurability)
2. will remove all user-provided templates on each update of your package. This will not happen if the package is still in a `.sublime-package` archive.


I don't know what you intend with ` .no-sublime-package.gitignore`.

justind000:
I learned the plugin system by paring down a similar plugin. Anyway, I fixed the `PLUGIN_NAME` thing. 

The oddly named file (thanks Windows), has been renamed to `.no-sublime-package` according to the Submitting a Package page, step 5.4 because there are executable files in the package. Which leads me to how the templates are loaded and some confusion. 

I am assuming that since `.no-sublime-package` is included, it will have to be unpacked. The template is itself a zip file that will be unpacked by the plugin, creating a small build environment for an obscure piece of hardware. 

I was not aware of your second point, but I had not anticipated users installing their own templates. Perhaps template is a bad word, as I mentioned above, it's closer to a build environment. 

Please let me know if I'm still confused and any additional changes that may need to be made.

FichteFoll:
Okay, so here is how I would do it:

First, there actually is no executable file in that package. There is `io.py` and `Default.sublime-commands`, which are both loaded by ST, and there is `templaces/Basic.zip`. The latter is read by `io.py`. 
Second, any file that you process in a Python plugin can be read via `sublime.load_(binary_)resource` and then passed to other functions. This allows us to not need a `.no-sublime-package` file which limits the user's override capabilities (see [docs](http://docs.sublimetext.info/en/latest/extensibility/packages.html#customizing-or-overriding-packages)).
Last, 

```py
templates = sublime.find_resources("/".join(__package__, "templates", "*.zip")
tmpl_names = [tmpl.rsplit("/", 1)[1][:-4].replace("_", " ") for tmpl in templates]
# ...
template_data = sublime.load_binary_resource(templates[index])
buffer = io.StringIO(template_data)
with zipfile.ZipFile(buffer) as zf:
    zf.extractall(location)
```

is roughly how I would do it. This also allows users to add their own templates to `Packages/io Node Framework/templates/`, while the actual package resides in `Installed Packages/io Node Framework.sublime-package`.

justind000:
This will have to wait. I don't know enough python to make your code work.

FichteFoll:
Okay, I will merge as is. Just keep in mind that using `.no-sublime-package` is suboptimal and prevents users from [overriding](http://docs.sublimetext.info/en/latest/extensibility/packages.html#customizing-or-overriding-packages).

justind000:
Added. Thanks for being patient.

PR #4547, opened 2015-06-22
Added BehaveToolkit

mixxorz:
BehaveToolkit provides integration between Sublime Text 3 and Behave.

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -286,8 +286,8 @@
 		},
 		{
 			"name": "Behave Toolkit",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Your tags don't have the `st-` prefix, as specified here. I suggest to just change it to `"tags": true`.

mixxorz:
Hey. Thanks for clearing that up! I read the docs and I didn't quite understand what the tags key meant. :stuck_out_tongue: 

I changed it to tags: true

PR #4550, opened 2015-06-23
add sublime plugin: Flow Built-In Type Completion

xinchaobeta:
add sublime plugin: Flow Built-In Type Completion

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -813,6 +813,16 @@
 			]
 		},
 		{
+			"name": "Flow Built-In Type Completion",
+			"details": "https://github.com/xinchaobeta/flow-builtin-type-completion",
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
Your importing only works on ST3, ST2 does not allow relative imports. You either need to adjust the import code to use try-catch or restrict the package for ST3.

Also, the second part of `'source.js' in scope_name_arr or 'source.js.jsx' in scope_name_arr` is redundant, since it will never be true when the first part wasn't either.

xinchaobeta:
Now I have make modified the code to suit ST2, and removed the 'source.js.jsx' . [0659c47](https://github.com/xinchaobeta/flow-builtin-type-completion/commit/0659c47b1d029195126ce6d9afb1504bf1832422)

PR #4552, opened 2015-06-24
add Uncrustify package

obxyann:
This is a source code beautifier/formatter plugin that allows the user to use Uncrustify to format the C-like languages in Sublime Text.

### DIFF
--- a/repository/u.json
+++ b/repository/u.json
@@ -74,6 +74,17 @@
 			]
 		},
 		{
+			"name": "Uncrustify",
+			"details": "https://github.com/obxyann/Sublime-Uncrustify",
+			"labels": ["formatting", "c", "c++", "java", "formatter", "beautifier", "code style"],
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
Your executable variable is named `DefaulExecutable`. I assume this is a typo, but it's consistent.
> ```py
def open_file(window, file_name):
	window.open_file(file_name)
```

You prefer functional programming over object-oriented?

obxyann:
1. Yes I will fix the typo in later version.
2. I think I forgot the to add a description text in the pull request. now it is "No description provided"! how can i fix it? I will add following text:
"A source code beautifier/formatter plugin that allows the user to use Uncrustify to format the C-like languages."

obxyann:
Thanks, I will fix it in later version.

FichteFoll:
I should also mention that it's *very bad practice* to name variables with TitleCase, which is reserved for class names. In this situation you should be using ALL_CAPS.

PR #4553, opened 2015-06-24
SendToSandbox plugin add

DavidGerva:
Pull request to add this new package SendToSandbox


FichteFoll:
If you intend to display the 1.0.0.txt message on installation, you need to declare it as the `install` message. PC won't "update" to 1.0.0 (since it's the first version) and thus not show the message.

DavidGerva:
Thank you @FichteFoll I corrected it!

PR #4556, opened 2015-06-24
Add UI Tweak for my Material Theme 

equinusocio:
I've made this package addon for overwrite some UI elements that some Material Theme user want.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -439,6 +439,17 @@
 			]
 		},
 		{
+			"name": "Material Theme - White Panels",
+			"details": "https://github.com/equinusocio/material-theme-white-panels",
+			"labels": ["theme", "color scheme", "material design"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Materialized CSS Snippets",
 			"details": "https://github.com/ayinloya/materialized-css-snippets",
 			"labels": ["css", "Materialized", "Materialized CSS"],


FichteFoll:
In the `.sublime-theme` file you refer to files in a `Material Theme White Panels` directory, but your package will be named `Material Theme - White Panels`.

equinusocio:
Ok thanks! I will rename now.

equinusocio:
Mmmm do think why if the package name is "Material Theme White Panels" he works and if named "Material Theme -  White Panels" doesn't? He doesn't override the settings in the second case.

Is there a chance to get the package named "Material Theme White Panels" or how i can fix this? I think the package is loaded before the original Theme by ST.

FichteFoll:
Hm, this is weird. It works when `Material Theme - White Panels` is unexctracted and in the Packages folder. It doesn't work for me if I name it `Material Theme White Panels` however, which it also shouldn't since that would still be at the same position relative to the original.

So I tested with `Material Thema` and `Material Themez` and discovered, that error messages (because I didn't adjust the folder names in the sublime-theme file) are only displayed for Themez. This implies that ST does indeed sort `Material Theme ...` before `Material Theme` and thus doesn't load the theme file properly. This would also be in line with the package loading in the Packages folder, since that is always loaded after installed packages.

FichteFoll:
I've determined this to be a Sublime Text bug and reported it at https://github.com/SublimeTextIssues/Core/issues/910.

Possible solutions for you:
1. Rename the package somehow so that it really gets loaded after "Material Theme", by making sure at least one letter is different.
2. Add `.no-sublime-package` to the package to turn it into a normal package, which are loaded correctly.

equinusocio:
Ok thanks man. I will ad the ```.no-sublime-package``` to the repository. I think it's better.

PR #4557, opened 2015-06-25
added ContextualMove plugin

davidson16807:
Ctrl+IJKL to move. Other keys set move behavior. You can move by characters, words, search results, and more.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2012,6 +2012,16 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
You have to push a valid [semantic version](http://semver.org) tag (with optional leading `v`) to your repo.

davidson16807:
Tagged 1.0.0

PR #4559, opened 2015-06-25
Update spec finder repo url

rogeriochaves:
The sublime plugin is no longer available on the original repository and I used it a lot, so I forked and improved the plugin (to find the specs faster and with more precision).

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1891,11 +1891,11 @@
 		},
 		{
 			"name": "Spec Finder",
-			"details": "https://github.com/Resellers/sublime-spec-finder",
+			"details": "https://github.com/rogeriochaves/sublime-spec-finder",
 			"releases": [
 				{
 					"sublime_text": "*",
-					"tags": true
+					"branch": "master"
 				}
 			]
 		},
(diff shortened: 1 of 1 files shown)


FichteFoll:
@resellercore, what happened there?

@rogeriochaves, for the same reason we don't allow new packages with branch based releases, we don't allow switching back to it either: they have been deprecated.

rogeriochaves:
@FichteFoll okay then, done!

I searched but I couldn't find any other ways to contact @resellercore, neither any explanation why they deleted the repo.

FichteFoll:
We'll wait a few days for an answer. Until then, you can push a valid [semantic version](http://semver.org) [tag](http://git-scm.com/book/en/v2/Git-Basics-Tagging) (with optional leading `v`) to your repo.

rogeriochaves:
@FichteFoll sorry, I forgot to push the tags, is everything ok now?

FichteFoll:
Yes, all okay. We'll wait for another few days however, imo 2 weeks are a minimum for these issues.

bfocht:
the repo on the resellers group was deleted.  I am not able to approve pull requests for @wbond

FichteFoll:
@bfocht, I am unsure what you mean with that. If the package was deleted and there is no replacement for it, I will just replace the record on Package Control with the fork by @rogeriochaves (in 1-2 days).

PR #4560, opened 2015-06-25
Add Umajin package

AdamHarte:


FichteFoll:
You have to push a valid [semantic version](http://semver.org) tag (with optional leading `v`) to your repo.

AdamHarte:
@FichteFoll Sorry about that. I have never had to push a tag before, and though it would just happen with a normal git push. But the tag is there now.

FichteFoll:

AdamHarte:
@FichteFoll I have created a new 1.0.0 tag for my repo. Do I need to do anything else for Package Control to see this new version?

FichteFoll:
No, Package Control will automatically find the new version up to one hour after you pushed it.

PR #4563, opened 2015-06-27
TotgaMathText initial check in

timluther:
TotgaMathText
=============
This package contains useful functions for manipulating selected text with mathematical functions.
i.e. selecting a series of numbers and then applying a formula (selected via a drop down menu) to those selections. 

This is my first attempt at writing a package control compliant sublime plugin - I'd appreciate any help in ensuring this package is compliant, useful and doesn't duplicate existing functionality.
I intend to further develop this module if submission proves successful.

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -96,7 +96,7 @@
 					"branch": "master"
 				}
 			]
-		},
+		},		
 		{
 			"name": "TADS3",
 			"details": "https://github.com/VoidPhantom/sublime-tads3",
@@ -1783,6 +1783,17 @@
 					"sublime_text": "*",
 					"platforms": "windows",
 					"tags": true
+				}
+			]
+		},
+		{
+			"name": "TotgaMathText",
+			"details": "https://bitbucket.org/timluther/totgamathtext",
+			"description": "Useful mathematical functions that can be applied to selected text",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "master"
 				}
 			]
 		},


FichteFoll:
All functionality in your package and more is also provided by https://packagecontrol.io/packages/Insert%20Nums, except for a preset or recently used storage (which is somewhere on my todo list).

Anyway, I have a few suggestions for your (next) package:

1. Don't mess with `sys.path` unless necessary. ST3 supports relative imports and ST2 sets the working directory to the package's directory, so that "absolute" imports check the package's folder too. You're not importing anything anyway.
2. Don't place files directly in the Packages folder. Use `Packages/{User}/` instead.
3. Your "insert_my_text" command is unused and redundant anyway. Sublime Text provides an "insert" command already.
4. By convention, commands added to the command palette are built like `PackageName: Command Name/Description`.
5. Your readme is extremely lackluster. It provides almost zero useful information. I had to analyze the source code in order understand what it *actually* does, and you can't expect that from everyone.
6. The `.sublime-package` file is unnecessary and only adds junk to the download archive. It won't be used anywhere.
7. I'm not exactly sure what your unused `RebuildLines` function (which uses class naming convention btw, bad practice) is supposed to do, but I guess it's `"\n".join(lines)`.

timluther:
Hi,
  Thanks for the tips and for the heads up regarding InsertNums - I had
looked for a package containing this sort of functionality but obviously
missed the keywords.
  Some of the points you raised (1.3 & 7) were down to simple sloppiness on
my part, leaving in code and tests from previous projects that had no
relevance here. I'll be more diligent in the future.
  I had thought the 'user' directory was only for user created add-ons,
whereas my impression was that the root 'packages' folder was for third
party add-ons. This project started off life in user, I moved it to the
root thinking that was correct. Could you possibly clarify for me under
what circumstances should a package be in the root 'packages' directory?
  The .sublime-package file is a left over as I was trying to work out how
to submit the package as, at first, my git repo didn't work (was missing
tags).
  As for the readme - again, apologies for the sloppiness. I'm afraid too
much of my effort was spent (prematurely) on trying to figure out how to
submit packages that the package itself wasn't as polished as It should
have been.
I've figured out private repos now so I'll host, test and polish these
privately before submitting another one.
Many thanks for your feedback, that's what I was looking for.

timluther:
I've been re-factoring some of my code and have moved some code in to a shared module. Unfortunately, the same problem I've had with python before not finding modules in the same directory as the importing file is in has appeared, necessitating the need for this line:
sys.path.append(os.path.dirname(__file__)) 
If appending to sys.path is frowned upon, are there any other work-arounds for this problem?

I've been using InsertNums and it's very functional and useful. However, perhaps there's also room for a plugin with a smaller scope and simpler syntax?

FichteFoll:
> Could you possibly clarify for me under what circumstances should a package be in the root 'packages' directory?

I'm talking about your json config file here, which are usually placed in the User package since that is guaranteed to not be touched (i.e. delete random files) by program code. Other than that, I would suggest you to use ST's settings interface which does it similar, except that users can easier only override the settings they want to. See http://docs.sublimetext.info/en/latest/customization/settings.html.

The package itself will go to either Installed Packages or Packages, depending on the ST version.

> it seems the problem was that sublime or python wasn't properly updating when the sub-module was edited and saved

Sublime Text does not reload submodules when they change, only top-level .py plugin files (meaning only in the package root).

> I've been using InsertNums and it's very functional and useful. However, perhaps there's also room for a plugin with a smaller scope and simpler syntax?

I don't really think so. You can easily replicate your package's functionality by referencing the command with a 'format' parameter in a keybind or command palette definition. I should probably add a section in the readme for that, together with the other thing I mentioned.

timluther:
Ah, I understand you now: I'd already moved the json file earlier today. Thanks for the pointers :)

FichteFoll:
So, can I close this?

timluther:
Sure, if you like.

PR #4569, opened 2015-06-29
Add TracRPC for The Trac Project

lyrixderaven:


FichteFoll:

Also, it seems you are not using semver correctly.

lyrixderaven:
Valid point, thanks for the hint! I'll correct the issue with a proper semver tag for 1.0.0.

PR #4570, opened 2015-06-29
Add LuaSmartTips

haoyue90:


haoyue90:
Add LuaSmartTips

FichteFoll:
In the requiresmarttips completions you compare against the syntax file used while in the other you use match_selector. I suggest using match_selector for both.

PR #4573, opened 2015-06-29
add package CustomPATH

victorhaggqvist:


FichteFoll:
It should be noted that this only modifies the path of the plugin container, even though pretty much everything (including build systems) goes through the plugin container.

Also, when you use `os.pathsep` instead of hardcoding the colon, this also works on Windows.

victorhaggqvist:
Thanks, incorporated the feedback in https://github.com/victorhaggqvist/SublimeCustomPATH/tree/v0.1.2.

PR #4575, opened 2015-06-30
added RailsBase16 color scheme to the package list

tompave:
RailsBase16 is a customization of the [Base16 color scheme](https://github.com/chriskempson/base16), which aims to provide better syntax support for Ruby and Ruby on Rails.  

[Readme and examples](https://github.com/tompave/rails_base_16/blob/master/README.md)

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -149,6 +149,17 @@
 			]
 		},
 		{
+			"name": "RailsBase16 Color Schemes",
+			"details": "https://github.com/tompave/rails_base_16",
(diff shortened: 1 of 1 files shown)


FichteFoll:
What do you think of applying these changes upstream? cc @chriskempson

tompave:
We did [talk about it](https://github.com/chriskempson/base16-builder/issues/120), but I don't think we followed it up with anything concrete.

I would still be super happy to contribute to the original theme :-)

tompave:
Thanks!

PR #4576, opened 2015-06-30
Added new package: MatlabAutoComplete

joepmoritz:
This new package adds Matlab filenames (without extension) in currently open folders to auto complete, when editing a Matlab file. This is useful as Matlab forces filenames to be the same as the function name, so this effectively adds all the user's custom functions to autocomplete.

This is slightly different from existing packages that autocompletes filenames and paths, as this is restricted to completing *.m files, does not add the extension, and triggers everywhere inside a Matlab file (but nowhere else).

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -461,8 +461,8 @@
 			]
 		},
 		{
-			"details": "https://github.com/joepmoritz/MatlabAutoComplete",
-			"labels": ["auto-complete"],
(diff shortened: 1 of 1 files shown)


FichteFoll:
I would appreciate a different package name, since it really only completes filenames and nothing more.

FichteFoll:
@joepmoritz ping

joepmoritz:
Have been very busy lately, will fix it now.

PR #4577, opened 2015-06-30
Added package SublimeREPL Babel.

ggrks:
SublimeREPL-babel extends the SublimeREPL package with a javascript REPL provided by [babel](https://babeljs.io), which allows you to execute next generation javascript code in a command line prompt since most of the syntax are not wildly supported by major platforms.
The original [SublimeREPL](https://github.com/wuub/SublimeREPL) repository seems to be inactive for 4 months, so I think it's better to create a standalone package.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -2989,7 +2989,7 @@
 			]
 		},
 		{
-			"name": "SublimeREPL Babel",
+			"name": "SublimeREPL-babel",
 			"labels": ["repl", "javascript", "babel", "es6", "es7"],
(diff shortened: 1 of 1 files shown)


FichteFoll:
Dependencies unfortunately don't work like this. You can only depend on actual "dependencies", not on packages.

Besides that, I really do not understand what your package actually does. It defines a .sublime-command file but doesn't actually provide a command by itself. If you're extending on the SublimeREPL package, maybe you should file a pull request for that instead?

You didn't push any tag either.

PR #4578, opened 2015-06-30
Added package Entropy

danechitoaie:
Package that allows you to do certain tasks for Demandware sandboxes using Sublime Text 3


FichteFoll:
`COMMAND/CONTROL + SHIFT + P` is called "command palette" not goto anything.

danechitoaie:
Ah, right. I'll update the readme. Thanks.

PR #4579, opened 2015-06-30
Add SimpleCov package for ST3

sentience:
This package is a fork of [Ruby Coverage](https://packagecontrol.io/packages/Ruby%20Coverage), which is already in Package Control; however, Ruby Coverage is ST2-only, whereas this package is ST3-only.

I [proposed releasing this fork under the original package name](https://github.com/integrum/SublimeRubyCoverage/pull/4); however, I’ve had no reply from the original package authors. I suspect it is no longer being maintained. Given that the old package is ST2-only and this one is ST3-only, there should be no confusion to Package Control users, who will only see one or the other when searching for a Ruby coverage package.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1215,6 +1215,16 @@
 			]
 		},
 		{
+			"name": "SimpleCov",
+			"details": "https://github.com/sentience/SublimeSimpleCov",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "SimpleMarker",
 			"details": "https://github.com/jugyo/SublimeSimpleMarker",
 			"releases": [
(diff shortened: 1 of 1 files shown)


sentience:
Sorry about the failed test. I have write access to package_control_channel and accidentally pushed to master in this repository rather than my fork. I undid that push right away, but that caused the associated build on Travis CI to fail, which is the fail visible above. Hopefully the passed build on the PR itself is acceptable.

FichteFoll:
Sounds very good to me. The only thing I'm missing is the relation to Ruby in the package name. "SimpleCov", to me, implies that it's compatible with multiple languages.

sentience:
SimpleCov is the name of the Ruby coverage tool that this package integrates with: https://github.com/colszowka/simplecov So I'd say putting Ruby in the package name would be redundant. I could see mentioning it in the package description so that searches for ‘Ruby’ match, though. Sound good?

FichteFoll:
Alright, didn't notice that.

PR #4582, opened 2015-07-01
Add new package "IPython Pick"

marciomazza:


FichteFoll:
You must move your test files to a subdirectory. Otherwise, ST will load them automatically and override `sys.modules["sublime_"]`, which will result in numerous failures for other plugins.

marciomazza:
Thanks for the revision.

I moved the files to the subdirectory "tests" and made some more provisioning on the sublime modules mocking.
Also changed the tag to 0.1.0

PR #4585, opened 2015-07-02
Add FT Origami

kaelig:
screenshot 2015-07-02 11 01 03

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -1105,6 +1105,16 @@
 			]
 		},
 		{
+			"name": "FT Origami",
+			"details": "https://github.com/Financial-Times/FTOrigamiSublime",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "master"
+				}
+			]
+		},
+		{
 			"name": "FTPSync",
 			"details": "https://github.com/NoxArt/SublimeText2-FTPSync",
 			"labels": ["ftp", "sync"],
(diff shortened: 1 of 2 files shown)


FichteFoll:
So, what is the difference to ColorHighlighter?

kaelig:
This is for the Financial Times Origami colour scheme. Actually, maybe I should create a FT Origami package instead, since we might add stuff to it later such as autocompletion and HTML snippets.

What do you think?

kaelig:
This is what Origami is all about: http://origami.ft.com/
And the color module's documentation is here: http://registry.origami.ft.com/components/o-colors@3.0.0

kaelig:
@FichteFoll I've renamed the repo and made it clear what it was. We we use Sublime a lot at the Financial Times, and a lot of third party partners also do.

FichteFoll:
Okay.

However, your repo url 404s. And you'll need to change `"branch": "master"` to `"tags": true` and use tags to create releases and push valid [semantic version](http://semver.org) tags to your repo. As mentioned [in the docs](https://packagecontrol.io/docs/submitting_a_package#Step_4), we no longer accepts branch-based releases for new packages.

kaelig:
Thank you @FichteFoll. I have updated the file and released a v1, hope everything is ok now.

kaelig:
Thank you very much @FichteFoll.

PR #4586, opened 2015-07-02
Use Travis's container base architecture

kaelig:
This results in significantly faster builds but removes the ability to use `sudo`.


FichteFoll:
LGTM, thanks.

PR #4590, opened 2015-07-04
Add Better Stylus Snippets package to repository

lnikell:
Hello! 
I used your guides about creating packages and I know that your list of the packages have already contain Stylus Snippets, but that package did not have changes from 2013 and that package use old syntax with colon, which not supported in many companies.

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -398,16 +398,6 @@
 			]
 		},
 		{
-			"name": "Better Stylus Snippets",
-			"details": "https://github.com/lnikell/better-stylus-snippets",
-			"releases": [
-				{
-					"sublime_text": ">=3000",
-					"tags": true
-				}
-			]
-		},
-		{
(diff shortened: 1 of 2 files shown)


FichteFoll:
You closed this PR, but I reviewed it anyway, i case you intend to reopen it.


"Better Stylus Snippets" is a very subjective name, because what exactly makes it "better" other than not including colons, and that's personal preference. I suggest something along "Stylus Clean Completions" or similar, which hints the non-use of colons.

Also, since these are actually completions and not snippets, it's more fitting.

lnikell:
Sorry, I saw it too late. I closed pull request because I saw my mistake.

FichteFoll:
You can append commits to existing pull requests by pushing onto the branch you opened it from.

lnikell:
Thank you! Got it and fixed.

PR #4591, opened 2015-07-04
Add Better Stylus Snippets package

lnikell:
Hello! 
I used your guides about creating packages and I know that your list of the packages have already contain Stylus Snippets, but that package did not have changes from 2013 and that package use old syntax with colon, which not supported in many companies.

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -397,6 +397,16 @@
 				}
 			]
 		},
+    {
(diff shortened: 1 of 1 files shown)


lnikell:
Thank you for your previous comment. I will change the name of repository and package and will create pull request again.

PR #4593, opened 2015-07-05
Add Copy Code Snippet

socsieng:


FichteFoll:
I would have been nice to have an example of how the code is actually formatted. There doesn't seem to be a similar package however.

You shouldn't be playing your `setup.py` (whose reasoning I question) in the root folder as ST will load it and attempt to find classes that interface its API.
You also shouldn't mess with `sys.path`, unless necessary. ST3 supports relative imports, ST2 uses absolute. Just use those instead. For ST3, you need to prepend `__package__ + '.'` to the module name before checking if it's in sys.modules.

FYI, ST2 uses Python 2.6.

socsieng:
Thanks for the feedback @FichteFoll.

I'll get to work on addressing your feedback on the items above. Where I'm ultimately going with this package is I want to be able to copy and paste code snippets with rich text formatting to be used in things like email communication and documentation.

With regard to `setup.py`, I'm still very new to Python as a language/environment, where `setup.py` came from was it was provided as part of a `yeoman` generator for Python which came packaged with `makefile`, unit tests and `travis` integration. Are you able to suggest any alternatives? I'm really keen to maintain `travis` integration though.

FichteFoll:
Hm, you might get problems with running `make setup` then, which calls `pip install ...`, but since you don't actually have any dependencies this could just work (if you remove the records). Don't quote me on that however.

Your test setup seems quite complex for a simple plugin like yours, so I can't simply propose how to do it differently. You could check out https://packagecontrol.io/packages/UnitTesting however, which is a probably even more complex approach but operates on the real Sublime Text.

socsieng:
`make setup` with the `pip install` is working (at least locally and on `travis`). Ironically, the only dependencies that are installed are there to support unit testing.

I agree that this is a somewhat complicated process for executing unit tests for this package as it stands, but I do expect the scope and complexity to grow when I introduce copy as rich text feature. This testing approach was basically a copy/paste of another [Sublime package (IndentX)](https://github.com/socsieng/IndentX) that I created which I think worked well for me.

FichteFoll:
You might be able to move everything related to running tests (notably `setup.py`, potentially `Makefile` too) down a directory and adjust paths accordingly. Be vary of Python's importing mechanics however.

FichteFoll:
For the short-term solution I suggest wrapping your `setuptools.setup` call in `setup.py` in a `if __name__ == '__main__':` block so it doesn.t get executed by Sublime Text when loading the plugin files.

socsieng:
I've wrapped up the contents of `setup.py` with `if __name__ == '__main__'`, removed support for Sublime Text 2, and started using relative imports (removed `sys.path` hack)

PR #4595, opened 2015-07-06
Fix renamed repository of compass package

renestalder:
Seems no longer to be visible in the package control, even github should redirect all repository names correctly. So here is the fix.

### DIFF
--- a/channel.json
+++ b/channel.json
@@ -121,7 +121,7 @@
 		"https://raw.githubusercontent.com/vkocubinsky/sublime_packages/master/packages.json",
 		"https://raw.githubusercontent.com/wallysalami/QuickLook/master/packages.json",
 		"https://raw.githubusercontent.com/weslly/sublime_packages/master/packages.json",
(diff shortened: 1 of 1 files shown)


FichteFoll:
I wonder why it doesn't work though, github seems to properly return the file with a 200 and I doubt PC is doing anything else than downloading that file.

cc @wbond

FichteFoll:
You might want to update to the [new 3.0 schema](https://github.com/wbond/package_control/blob/master/example-repository.json) btw, or migrate the file to the default channel completely so that the schema can be updated automatically. You still have to change the address when renaming the repo however.

wbond:
It seems to be present: https://packagecontrol.io/packages/Compass.

renestalder:
@wbond well, it wasn't after the repository rename as you can see in the statistics. There's a big hole from 2th until 7th of July.

@FichteFoll I'll look into the schema change. Thank you.

wbond:
That isn't a bug in PC. GitHub does not provide redirects for the API, so there is nothing to do there.

PR #4599, opened 2015-07-07
Add Show ASCII Code package

borislubimov:
Show ASCII Code - Simple plugin for displaying ASCII code of the current character in the status bar


FichteFoll:
The package name is not exactly correct, since it will also handle unicode symbols just fine (Python's ord does that). How about "Character Code"?

borislubimov:
@FichteFoll Thanks for advice! I've changed the package name and added tag for my repo.
Also I installed package Char Value and had a look on it. The goal of it is different - to substitute characters by their codes, while my plugin doesn't change text - just shows character codes in status bar and allows to compare codes of different characters (which could look the same, as on plugin demo https://github.com/borislubimov/ShowCharacterCode)

FichteFoll:
I am aware of the difference, I was mainly referencing Char Value because it is similar in regard to charater codes and does not use "ASCII".

PR #4601, opened 2015-07-07
Added NetSuite Bundle Package

ththev:
The NetSuite Bundle for Sublime provides code completion for NetSuite APIs and a quick panel menu with NetSuite records internal IDs and script templates.

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -191,6 +191,16 @@
 			]
 		},
 		{
+			"name": "NetSuite Bundle",
+			"details": "https://github.com/ththev/NetSuite-Bundle-for-Sublime",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Nette",
 			"details": "https://github.com/Michal-Mikolas/Nette-package-for-Sublime-Text-2",
 			"releases": [
--- a/repository/n.json
+++ b/repository/n.json
@@ -195,7 +195,7 @@
 			"details": "https://github.com/ththev/NetSuite-Bundle-for-Sublime",
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": ">=3000",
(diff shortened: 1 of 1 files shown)


FichteFoll:
1. You're leaking file handles.
2. You won't be able to access local files in a ST3 installation like that, because the package is bundled as a `.sublime-package` zip archive there. You'll need to use `sublime.load_resource` instead, which is only available once the API has loaded (when it calls the `plugin_loaded` function).
3. Instead of defining your own insert command, just use the built-in "insert" command with the `characters` parameter.

ththev:
Thank you for the comments.
I've made some changes and decided to make the package available only for sublime 3 for now.
Can you please check it?

FichteFoll:
Your `load_resource` calls will fail because Package Control installs the package as "NetSuite Bundle", as indicated by this pull request. You can use `sublime.load_resource('/'.join("Packages", __package__, "RecordTypes.json"))`.

I don't undestand why you added [this](https://github.com/ththev/NetSuite-Bundle-for-Sublime/commit/f71bc192c9dedfb3f51dd18d023040555e3ec542#diff-525ea0a238cd08f7e3520f25ca9444c7R55) however and are not using "insert_snippet" anymore. Sublime Text should use whatever line endings are set for the currently edited file, unless it's https://github.com/SublimeTextIssues/Core/issues/858.

ththev:
Thank you again for the suggestions.

The reason why i removed end lines was because I was using JS files as templates and CR characters were being added to the view (probably because of the issue that you mentioned before).
I researched a bit more and I'm now using snippets again as I think this is the best option.

I also made the other required changes and created a new tag. Can you please check it again?

Thank you!

FichteFoll:
Interesting. I discovered the issue from some default snippets that where using the `.tmSnippet` extension with CRLF linebreaks, but was unable to reproduce with `.sublime-snippet` files.

PR #4603, opened 2015-07-07
Updated support of QuickRails to ST2+ST3

danpe:


FichteFoll:
Are you *sure* this still works on ST2? Because it doesn't look like it to me.

For imports in ST3 I highly suggest using relative imports like `from .QuickRails import QuickRailsWindowCommand, rails_root`, while in ST2 you must use `from QuickRails import QuickRailsWindowCommand, rails_root`. Just wrap one of these in a try block.

danpe:
Hey, Thanks.
I fixed it, works on both.

FichteFoll:
I'm still not sold on that this will work on ST2. Did you verify this?

The imports should look like this:
```py
try:  # ST3
  from .QuickRails import QuickRailsWindowCommand, get_idea
  from .QuickExec import ProcessListener
except ValueError:  # ST2
  from QuickRails import QuickRailsWindowCommand, get_idea
  from QuickExec import ProcessListener
```

FichteFoll:
@danpe ping

PR #4604, opened 2015-07-07
added remify

brousalis:
Converts

margin: 4px

to

+rem(margin, 4px)

Very specific use case, assuming you are using .sass files and a mixin called rem to convert pixels to ems.


FichteFoll:

brousalis:
Sorry! I've updated the repo.

PR #4605, opened 2015-07-08
added initial VoiceCode package

unboundmusic:
Added a sublime package to work with voicecode.io for complete voice control of sublime text.


FichteFoll:
`@exec "subl --command 'goto_line {\"line\": 10}'"` does not work, at least not for me.

PR #4446, opened 2015-05-22
Added JaiLang Package

gingerBill:

PR #4606, opened 2015-07-08
Added Jai Syntax Plus to repository lists.

jasonericson:
I went back and forth on whether I should add this, as there's a Jai syntax highlighter already existing in Package Control. I eventually decided that my version is different enough that both mine and the existing one have merit for different reasons (mine is fairly "aggressive" in it's scoping resulting in a lot of coloring, which some people might not like).

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -78,6 +78,16 @@
 			]
 		},
 		{
+			"name": "Jai Syntax Plus",
+			"details": "https://bitbucket.org/jasonericson/jaisyntax",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Jamon",
 			"details": "https://github.com/drhayes/SublimeJamon",
 			"releases": [


FichteFoll:
Your repo url 404s.

Maybe @GreenLightning can comment on https://bitbucket.org/jasonericson/jai-syntax-plus.

FichteFoll:
Related: #4446.

jasonericson:
Hey @GreenLightning !

I'd be happy to contribute to your repo instead of adding this one, if you're interested. Would you mind taking a look at what I've built when you have the chance?

The biggest variation that I'm not sure how people will feel about is my choice to highlight variable names at their definition. I thought it would be a nice visual cue since Jai doesn't have the visual cue of a variable name being preceded by a type at its definition - but it might be too much information for some people.

Everything else I'm pretty confident about, minus a couple bugs I've listed in Issues in my repo.

GreenLightning:
As I said, I'm happy to accept pull requests!

https://github.com/GreenLightning/sublime-jai

If you want, you can also open an issue with something I missed (like functions) and we can figure out a way how to do it together.

I would like my syntax highlighter work correctly for as many programs as possible though. This is rather hard, because the syntax is very context sensitive, which is difficult to capture using regular expressions, and there is no official specification or compiler released yet.

I've stumbled upon similar problems as you have described in your issues section, which is why I haven't created more sophisticated syntax highlighting patterns yet.

jasonericson:
You make a good point about the changing nature of the language. How about this: I'll keep developing my syntax highlighter in parallel, as a sort of "experimental branch", so I can play around with different approaches to understanding the context as the language changes. In the meantime, I'll create pull requests for your repo for bigger, more obvious things such as missing keywords and such. And yeah, I'll open an issue about functions specifically (maybe structs), as identifying those will allow for Goto shortcuts.

Once the language is released, I'll start making pull requests for yours that incorporate more of the sophisticated parsing I've figured out, since it will be more stable at that point. How does that sound?

GreenLightning:
Sounds good.

FichteFoll:
Sounds good here too. I take it this can be closed then?

jasonericson:
Yes, I'll do it myself.

PR #4607, opened 2015-07-08
Add Fuse package to the repository.

Tapped:


FichteFoll:
I suggest using relative imports instead of absolute (and not basing on "Fuse").

You should really take a look at PEP8 and the naming conventions it suggests.

Tapped:
We have tried to be consistent in our style, but if it matters that the style is not "proper" python style, we can fix that.

FichteFoll:
It does not matter, but naming functions in TitleCase is quite concerning, for me at least, since that is normally reserved for classes. I also couldn't actually find a style since you were using lowercase and camelCase, as well as TitleCase for method names (in the `BuildResults` class for example)

Tapped:
Yeah, build_results, fuse_parseutils and fuse_util has not yet been refactored for its function names. 

However it is okey, with camelCase and__camelCase for method names?

FichteFoll:
__ actually has a special meaning that is not limited to simply naming (and I hope you are aware of that). If not, you should look it up. It should also be mentioned in PEP8.

I'm more concerned about relative imports.

Tapped:
@FichteFoll 
Hey, I have done some cleanups. Thank you for your time.

PR #4609, opened 2015-07-08
Add phpunit-snippets; Closes gerardroche/sublime-phpunit-snippets#1

gerardroche:
I'm aware there are similar packages.

This package is one of three individual packages for PHPUnit support:

* https://github.com/gerardroche/sublime-phpunit
* https://github.com/gerardroche/sublime-phpunit-snippets
* https://github.com/gerardroche/sublime-phpunit-completions

PR #4610, opened 2015-07-08
Add phpunit-completions; Closes gerardroche/sublime-phpunit-completions#2

gerardroche:
I submitted this package before but it was rejected because it was too similar to other packages. The other packages are either unmaintained or similar but not the same.

Up to you. Merge, close, pull changes into another package, it's ok with me.

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -918,6 +918,17 @@
 			]
 		},
 		{
+			"name": "phpunit-completions",
+			"details": "https://github.com/gerardroche/sublime-phpunit-completions",
+			"labels": ["php", "testing", "phpunit", "completions"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Pipe Dream",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Well, I can see that you have quite a few more completions for different aspects compared to https://github.com/tkowalewski/phpunit-sublime-completions, but I still don't really see something that would prevent you from adding the snippets there. Maybe @tkowalewski could take a lot at yours and say if he'd be fine with them being added to his repo, or giving you push access.

Besides, the names are way too similar.

(for future reference: https://github.com/wbond/package_control_channel/pull/3649)

gerardroche:
ping @tkowalewski what do you think?

These completions render most of the snippets in the [PHPUnit Snippets](https://packagecontrol.io/packages/PHPUnit%20Snippets) package redundant. The snippets package will cause duplicate auto completes and unexpected behaviours. \cc @florianeckerstorfer

tkowalewski:
@gerardroche Push access granted, feel free :)

FichteFoll:
@gerardroche did you settle this? (and #4609)

gerardroche:
Cool.

PR #4611, opened 2015-07-08
Add php-grammar; Closes gerardroche/sublime-php-grammar#1

gerardroche:


FichteFoll:
I can't merge this on github, likely because of the other package from you I merged earlier. Please rebase.

Regarding the package itself, you can remove `package-control.json` since it won't be used anywhere -- unless you need it for personal reference. I also don't think it's too good of an idea to automatically load the `tests.py` tests on every installation since users will hardly need them, but it would be quite hacky to load them dynamically since you are loading commands there.

gerardroche:
The tests stuff should be exlcuded from package control installations see the .gitattributes file. As far as I understand package control exports the package from github which will use git archive and will use the gitattributes ignore rules. Unless something in package control has changed.

The package-control.json is just for reference.

Rebasing now.

FichteFoll:
Alright then, didn't look inside .gitattributes.

PR #4612, opened 2015-07-08
Added Easy Open package

mjkaufer:


FichteFoll:
You're defining a TextCommand, yet you're not actually doing anything with the view. Either switch to WindowCommand or use `self.view.file_name()` instead of extracting variables. That function has only been added recently, so it won't work in ST2 obviously.

mjkaufer:
Whoops. I'll let you know when I fix that. Thanks!

mjkaufer:
@FichteFoll I believe I fixed it. Any other issues you see, or are we good to go?
Thanks!

FichteFoll:
Looks good now.

You could take a look at `os.path.splitext`.

mjkaufer:
Haha, I'll use that next time. Thanks.

FichteFoll:
@mjkaufer can you please find a different name for your package? I think I forgot to check, but there already is a package named "[EasyOpen](https://packagecontrol.io/packages/EasyOpen)" and that is **way** too similar.

PR #4613, opened 2015-07-09
Add AceJump plugin

ice9js:
I'm aware of the existing EasyMotion plugin, however there are some fundamental things that my version does differently (mutiple modes, multiple views) and that's the reason why I created a new package for it.
I'll let you be the judge whether it's a worthy successor for EasyMotion in st3 as st2 is not supported right now.

cc @tednaleid

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -66,8 +66,6 @@
 		{
 			"name": "AceJump",
 			"details": "https://github.com/ice9js/ace-jump-sublime",
-			"readme": "https://raw.githubusercontent.com/ice9js/ace-jump-sublime/master/README.md",
-			"issues": "https://github.com/ice9js/ace-jump-sublime/issues",
 			"labels": ["code navigation", "text navigation"],
 			"releases": [
 				{
--- a/repository/a.json
+++ b/repository/a.json
@@ -64,6 +64,19 @@
 			]
 		},
 		{
+			"name": "AceJump",
+			"details": "https://github.com/ice9js/ace-jump-sublime",
+			"readme": "https://raw.githubusercontent.com/ice9js/ace-jump-sublime/master/README.md",
+			"issues": "https://github.com/ice9js/ace-jump-sublime/issues",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Looks different enough to warrant its own package imo, but this can also easily be combined since the functionalities are not exclusive. Let's see what @tednaleid can say about adding these features to his package or other collaboration efforts.

On the pull request in particular:
- I'm definitely not a fan of overriding default key bindings, which `ctrl+shift+l` is.
- I don't think using `view.end_edit` is a good idea since it will get called again by the TextCommand wrapper, but the token is reset to `0` before that. Do you even need it for the undo command? Maybe use `sublime.set_timeout`? Otherwise, if it works ...

tednaleid:
I've only got experimental support for ST3 in EasyMotion, and I don't really have any future plans for expanding on what's already in the plugin, so I think having a successor plugin with official support for ST3 makes sense.   

I'd probably rather deprecate EasyMotion and have @ice9js's plugin take over rather than merge the two.  Making EasyMotion "correct" for ST3 would take more time than I'm willing to put in to the plugin, especially as I don't use ST3.

FichteFoll:
I see, that sounds reasonable.

In that case I suggest deprecating the st3 releases from EasyMotion (preferrably with an update message) and pointing to @ice9js's AceJump, which is ST3-only.

@tednaleid, take a look at https://packagecontrol.io/docs/messaging. You'll need to switch the release strategy to tags however, at least for the st3 branch. Just push a `st3-v0.1.0` tag from the commit that adds a message for the `0.1.0` version and change your release entry in the package_control_channel repo from `"branch": "st3"` to `"tags": "st3-v"`.

@ice9js, I'm merging this if then if you don't have any objections.

ice9js:
I removed the unnecessary keys from the entry in the packages json file.
Thanks for the other tips as well, I'll try to figure those out.

Ready to merge.

PR #4614, opened 2015-07-09
Add Envy

sailorswift:
The project has to do some very nasty stuff to ensure the environment updates before any other package uploads (e.g. not using sublime.settings and avoiding package_loaded).  Let me know if this is okay!

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -529,6 +529,16 @@
 			]
 		},
 		{
+			"name": "Envy",
+			"details": "https://github.com/sailorswift/Envy",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "EPM",
 			"details": "https://github.com/ethereum-package-manager/epm-sublime",
 			"releases": [


FichteFoll:
I don't see a reason why you need to load the settings before `plugin_loaded`. Mind explaining?

sailorswift:
Yes!  plugin_loaded was called after packages that depended on the environment variables supplied by the shell/Envy.  One example of this was GoSublime which depends on $GOPATH and $GOROOT.  When using plugin_loaded, GoSublime would fail because because $GOPATH and $GOROOT did not exist.

The settings file is important because it's where the user specifies their shell and its config file that are used to load the environment variables.  Since plugin_loaded does not happen fast enough to set the environment variables before other packages are loaded, I needed to write my own method to load settings.

I tried naming my package AAAEnvy and then use plugin_loaded but it still ran into the same issue.  From what I could see, there is no way to specify the order in which plugins are loaded.

FichteFoll:
See http://docs.sublimetext.info/en/latest/extensibility/packages.html#merging-and-order-of-precedence for how packages are ordered. I'm not sure in which order their `plugin_loaded` callbacks are called though.

In any case, since your package will be installed as `.sublime-package`, you won't be able to read files using `open` - at least not files in your package. You need to use `sublime.load_resource` for that.
Also, users should **never** have to edit the package in place since Package Control will replace the entire package on an upgrade and thus remove all user edits. Instead, refer to the user folder and do it like settings are handled by ST usually.

sailorswift:
load_resource doesn't work unless it's called in or after plugin_loaded() :disappointed:.  I can't think of a way around it since open cannot be used, so I'll be closing this and just rely on a git clone as hacky as that is.

PR #4617, opened 2015-07-11
Add Laravel 5 Snippets

Lykegenes:
Hi!

There is currently another similar package in the channel, but the author deleted his Github account so it's unavailable. The missing package name is "Laravel Five Snippets" if you want to remove it.

Thanks!

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -108,17 +108,6 @@
 			]
 		},
 		{
-			"name": "Laravel Five Snippets",
-			"details": "https://github.com/brnlbs/laravel-five-snippets",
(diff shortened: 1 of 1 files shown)


FichteFoll:
You are right, the other package author (and package alike) seems to be gone.

I suggest you move the old users to your new package by adding its name to a "previous_names" array in your package info.

Lykegenes:
I made the changes, the package is ready!

FichteFoll:
I forgot to mention that you need to remove the old package from the json as well.

Lykegenes:
Done, Thanks!

PR #4621, opened 2015-07-12
Adding HSP Package.

potato4d:
This is the language definitions HSP Language(http://hsp.tv/).


FichteFoll:

This package will also work perfectly fine on ST3, so I suggest you change the version selector to `"*"`.

potato4d:
Thank you for your advice.
Done, please check it.

PR #4624, opened 2015-07-12
Remove space in package Show Texdoc

CareF:


FichteFoll:
You are using relative imports in the top level of your package. These will fail on ST2. You either have to wrap them in a "try-catch" block and use the absolute style in ST2 (`from 
from tex_package_recognizer import GetPackagenameInline`) or remove the compatability flag for ST2 by changing the `sublime_text` version selector.

CareF:
Thank you for your advice! I have changed the imports in my package.

PR #4629, opened 2015-07-14
Add DoxyDoxygen to the official packages list

20Tauri:
This package simplifies writing comments.

It have a good language comprehension.
May probably replace DoxyDoc package (except for the license), because even if DoxyDoxygen is not dedicated to C++ it provide a better support that it...

DoxyDoxygen have a different approch than the great DocBlockR. Due to his better knowledge of language, DoxyDoxygen may guess type without any configuration. It also provide a lot of long time requested functionaly that cannot be easily implentated to DocBlockR (like comments reparsing)

Here an example of generated documentation:
```JavaScript
/**
 * { description }
 *
 * @param      {number}  errorCode  { description }
 *
 * @return     {string}  { description_of_the_return_value }
 */
function error2msg(errorCode) {
    if (errorCode == 404) {
        return "page not found";
    }

    return "...";
}
```

Please visit the official [DoxyDoxygen page](http://20tauri.free.fr/DoxyDoxygen) for further informations

Thanks you for your time,
Thanks you for your review.

### DIFF
--- a/channel.json
+++ b/channel.json
@@ -2,12 +2,12 @@
 	"schema_version": "3.0.0",
 	"repositories": [
 		"./repository.json",
-		"http://20tauri.free.fr/DoxyDoxygen/DoxyDoxygen.json",
 		"https://bitbucket.org/jjones028/p4sublime/raw/tip/packages.json",
 		"https://bitbucket.org/klorenz/sublime_packages/raw/tip/packages.json",
 		"https://csch1.triangulum.uberspace.de/release/packages.json",
 		"https://eberstarkgroup.com/releases/packages.json",
 		"https://packagecontrol.io/packages_2.json",
+		"https://raw.githubusercontent.com/20Tauri/DoxyDoxygen/master/DoxyDoxygen.json",
 		"https://raw.githubusercontent.com/accerqueira/sublime/master/packages.json",
 		"https://raw.githubusercontent.com/afterdesign/jshintify/master/packages.json",
 		"https://raw.githubusercontent.com/afterdesign/MacTerminal/master/packages.json",
--- a/channel.json
+++ b/channel.json
@@ -2,6 +2,7 @@
 	"schema_version": "3.0.0",
 	"repositories": [
 		"./repository.json",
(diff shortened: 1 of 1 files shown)


FichteFoll:
1. For security reasons you must use the https protocol for hosting your own .json repository.
2. `previous_names` is redundant and could possibly confuse PC because it contains the same name as it actually has. Please remove.
3. it's better to just not include the `donate` key than setting it to `""`
4. The `details` url only supports links to github or bitbucket repos and is just there to let the crawler auto-fill the remaining keys. You can either change this to `https://github.com/20Tauri/DoxyDoxygen` and remove `readme`, `issues`, `homepage` (if you add the homepage to your package's "url", next to its description), `description` (fetched from repo description) or just remove the details key.
5. `readme` has to refer to the raw readme file.
6. `date` fomatting is invalid. It must be `YYYY-MM-DD HH:MM:SS` and should be in UTC.

Install the [ChannelRepositoryTools](https://packagecontrol.io/packages/ChannelRepositoryTools) package and run the tests on your .json file until they pass.

Because your package is not open source I would not want to replace the DoxyDoc package.

20Tauri:
Thanks you for this great review.

Repository file has been fixed and tests passed with your patch.

JSON Repository file is now hosted on github (https)... As github certificates seem to be checked (to prevent man in the middle attack), I will also switch packages on it... Don't seem to be blocking for Package_Control, but it's better for final users.

Regards,

20Tauri:
Change to secure all URLs

FichteFoll:
Alright, this looks good.

Before I merge this however, you might want to check out [github releases](https://github.com/blog/1547-release-your-software) and the ability to upload arbitrary binary files.

20Tauri:
Thank you for this hint !
It's a more logical way to use GitHub (and using the Git Api is easy to maintain).

The only "issue" is that  Asset cannot have the same download name nor been hidden.
So "Sublime-Packages" have to been renamed to be used manually...

Due to this restriction, I've also slighty modified the repository ("url" field)

Thanks again for all

PR #4634, opened 2015-07-16
Add Postap

LPGhatguy:
All tests pass!

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -1157,6 +1157,16 @@
 			]
 		},
 		{
+			"name": "Postap",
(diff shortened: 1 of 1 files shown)


FichteFoll:
I suggest flagging your package as a theme, by adding the "theme" label or by including "Theme" in the package name, e.g. "Theme - Postap".

You should also add a screenshot to your readme, users will hardly install a theme just to find out how it looks.

LPGhatguy:
I've included the 'theme' and 'color scheme' tags; I'll be adding a screenshot to the README in just a moment.

PR #4635, opened 2015-07-16
Added Glorified Copy 'n' Paste

teddybeermaniac:


FichteFoll:
I honestly don't think "Copy 'n' Paste" describes the functionality of this package, even remotely.

I also doubt you need to fiddle with `sys.path`, as it only seems to relevant for importing jinja2 from `thirdparty/` and you can easily do that with a relative import (`from .thirdparty import jinja2`).

teddybeermaniac:
Well, jinja2 uses absolute imports and if I understand python importing logic correctly, it won't work that way. But it turns out I can load it via importlib, which I already use in my plugin. So I fixed it this way - https://bitbucket.org/teddy_beer_maniac/sublime-text-glorified-copy-n-paste/commits/c212f900e7e0f3b3a7ecd64444e76671788760b3#Lgcnp.pyF37T37

Oh, and I honestly think "Glorified Copy 'n' Paste" describes plugin's functionality spot on. But it's just a name, and describing functionality is description's job, which it does. ;-)

FichteFoll:
That's an interesting way to do it. I only recently looked into how importing works in Python. I didn't think about jinja2 using absolute imports (and I hate it when packages do that, particularly for this reason).

It's your package, so you get to chose the name, unless it's not compliant to our standards. I just thing that "copy and pasting" is usually used for a different workflow, while yours is more akin to templating. In the end, it's only used for identification anyway.

FichteFoll:
Did you test the importing in `.sublime-package` state btw? I'm a bit worried about the `os.path.join(os.path.dirname(__file__), 'thirdparty')` part because as far as I understand, it wouldn't work with the ZipLoader. `load_module` and `find_module` also expect the fullname afaik, which would be `"jinja2.ext"` in the last import.

teddybeermaniac:
Yes, I tested it while packaged as a `.sublime-package` and it works. AFAIK [zipimport](https://docs.python.org/3/library/zipimport.html) module is used by `ZipLoader` and its documentation stated that 
> an archivepath of foo/bar.zip/lib will look for modules in the lib directory inside the ZIP file foo/bar.zip

so that's why I used `os.path.join`. Also `__file__` contains a path structured like that.

About that fullname, suprisingly it didn't work with it. But maybe I did something wrong, while testing it.  I'll check again.

Also thank you for all feedback and merging. :-)

FichteFoll:
I think I understand it now, but maybe you can confirm.

Basically, if you do `import jinja2` and hthen print `jinja2.__loader__` it should *not* be an instance of `sublime_plugin.ZipLoader` because it gets loaded by the internal `zipimport` module instead.
The reason is that `sublime_plugin.MultiZiploader.find_module` does not find `...sublime-package/anything` modules due to how it works (check `sublime_plugin` for implementation).

teddybeermaniac:
First, I don't know why, but when you wrote `ZipLoader` in your previous comment I thought that is the name of class from Python's libraries. And to answer your question, yes. Of course if jinja2 would use relative imports and it wouldn't require markupsafe then your suggestion from first comment would work and ZipLoader would be used. 

I think I should package those both as dependencies so that I won't require all this magic.

FichteFoll:
That is a very good idea, especially considering that jinja2 isn't the smallest package.

PR #4636, opened 2015-07-16
Add GoSnippets

dericofilho:


FichteFoll:
You'll need to change `"branch": "master"` to `"tags": true`. As mentioned [in the docs](https://packagecontrol.io/docs/submitting_a_package#Step_4), we no longer accepts branch-based releases for new packages.

I highly suggest adding a readme to your package that informs (potential) users which snippets are added.

dericofilho:
> You'll need to change "branch": "master" to "tags": true. As mentioned in the docs, we no longer accepts branch-based releases for new packages.

PR #4642, opened 2015-07-19
Added "Theme - Interstellar"

flovan:
Hi,

As a little side project, I made these planet-inspired color schemes, accompanied by a fitting UI theme. I think I've got it all done properly, but since I've been testing with a symlinked folder on my computer, I don't know if I am linking to my UI theme's images correctly.
For example here: https://github.com/flovan/interstellar-package/blob/master/Interstellar.sublime-theme#L43

Please let me know if I need to change anything to make this repo function properly as a ST package.

- F

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -845,6 +845,17 @@
 			]
 		},
 		{
+			"name": "Theme - Interstellar",
+			"details": "https://github.com/flovan/interstellar-package",
+			"labels": ["theme", "color scheme"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "master"
+				}
+			]
+		},
+		{
 			"name": "Theme - itg.flat",
 			"details": "https://github.com/itsthatguy/theme-itg-flat",
 			"labels": ["theme"],


FichteFoll:
Well, technically the Sun and Pluto are not planets ...

Anyway, you are correct with your concern regarding the pathing: It has to read `Theme - Interstellar/...`, as by this pull request.

Furthermore, [semantic versions](http://semver.org) consist of exactly three numeric parts.

(And I don't know if it's intended that your readme only shows `⋮` between the scheme names.

flovan:
I thought Pluto was recently renamed as a planet? But you're right about the sun, I might expand this to be "planets and stars" That last thing was done on purpose though! 

Thank you for taking look; I'll try pushing the updates later today.

flovan:
@FichteFoll I updated the paths and added a proper 3-part tag. I'll revisit the semantics of how I refer to the themes as planets in a later stage, probably after fully testing the package installation and releasing v1.0.0.

Is there anything else that needs to be addressed?

PR #4643, opened 2015-07-21
Added jinja2 and markupsafe to dependencies.json and to Glorified Copy 'n' Paste's dependency list

teddybeermaniac:
Is there by any chance possibility for dependencies to have dependencies themselves? I didn't feel like putting markupsafe along jinja2 was a good idea, but jinja2 will not work without it, so they both have to be added as dependencies in packages.

### DIFF
--- a/repository/dependencies.json
+++ b/repository/dependencies.json
@@ -31,6 +31,34 @@
 			]
 		},
 		{
+			"name": "jinja2",
+			"load_order": "50",
+			"description": "Python Jinja2 module",
+			"author": "teddy_beer_maniac",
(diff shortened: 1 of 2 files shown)


FichteFoll:
At the moment, dependencies can not have dependencies themselves. Packages must list both dependencies.
If one depedency depends on another dependency, it should have a "higher" load order so that it gets loaded later.

Your dependnecy repos do not have a valid semver tag.

I also recommend specifying dependencies in the a `dependencies.json` file in the root of the package because that will allow you to adjust dependencies along with code changes, while specifying dependencies on the channel is remote and needs to be accepted.

teddybeermaniac:
Fixed, fixed and noted.

PR #4645, opened 2015-07-22
Added SwitchCSSColorModel

Aarononomous:


FichteFoll:
This seems to do the same as https://packagecontrol.io/packages/CSS%20Color%20Converter. Please add any missing features to that package via pull requests.

I have to say you are using an interesting settings model however, allowing to specify project or syntax-specific settings as a dict and then utilizing the `.get` method, which is available to both dicts and the Settings object from `load_settings`.

I think your problem occurs if you do have a "SwitchCSSColorModel" dict in `view.settings()` that does not contain all settings values, as the ones who are not contained will yield `None`. You need to be merging the overrides with the defaults from `.load_settings`.

FichteFoll:
@Aarononomous ping

PR #4655, opened 2015-07-26
Add "openresty lua snippets" support.

membphis:


wbond:
You should delete package.json from your repository. That is only used for custom hosting, and is not actually being used with your PR here.

Also, you tags needs to be semantic version numbers, so instead of `v1.0` it will need to be `v1.0.0`.

membphis:
thanks for your help. i have removed the file package.json and add the new tag v1.0.0.

please check if its fine.

PR #4656, opened 2015-07-26
Add SublimeTek plugin

hug33k:


wbond:
You don't currently have any semantic version tags in the repository. An example of one is `1.0.0`.

Is this only to be used for students in one specific program? If so, do you think it is worth having it in the default channel, or do you think it makes more sense being added as a repository to each student's install?

hug33k:
Alright, tag is now here.

I think it would be easier for new students to find this plugin with Package Control because most of students use ST and most of them are looking for a plugin like this. They don't want to spend hours and days to find the right plugin on Internet or something else.

FichteFoll:
We have decided that we would prefer if the package accepting process was automated in the near future anyway and that we won't be trying to convince authors of a package with a farily limited and closed target audience (i.e. students at your college/univ) to use the "Add Package/Repository" commands of Package Control, which would allow installing arbitrary packages. Thus, we will accept your package.

However, your name is not valid https://packagecontrol.io/docs/submitting_a_package#Step_2.

hug33k:
Tek is already used for one of my projects.
SublimeTek is the implementation for ST.
So, I don't know what to use for the plugin name right now.

FichteFoll:
You don't need to rename the repo but change the "name" in the package info you added in this PR. It will be the "new" name of the package.

FichteFoll:

FichteFoll:
@hug33k ping

PR #4658, opened 2015-07-27
Add Asciidoctor (AsciiDoc)

jirutka:


FichteFoll:
There already is an "AsciiDoc" package: https://github.com/SublimeText/AsciiDoc. It seems unmaintained, but I can add you as a contributor. Interested?

cc @mrchief

jirutka:
I know about [this package](https://github.com/SublimeText/AsciiDoc), but the syntax highlighting is totally broken and as you said, it’s unmaintained (there are already 2 unmerged PRs).

I’ll be happy to merge [asciidoctor/sublimetext-asciidoc](https://github.com/asciidoctor/sublimetext-asciidoc) to the current one, it will solve naming problem and related confusions for users. However, I think that we prefer to keep it under the [Asciidoctor](https://github.com/asciidoctor/) organization (same as [Atom](https://github.com/asciidoctor/atom-language-asciidoc) [packages](https://github.com/asciidoctor/atom-asciidoc-preview) etc.)… @mojavelinux?

FichteFoll:
Oh I see, I didn't notice this was part of something bigger. In that case, I think we can merge this even if it is a duplicate because it is part of a bigger brand and has a distinctive name. It would also not confuse current users of AsciiDoc with the big differences (at least to me).

Only suggestion I have is to remove the ` (AsciiDoc)` part of the name, which is redundant semantically since it's already a part of "asciidoctor", and thus can easily be searched. Even if not, the description is also considered when searching on https://packagecontrol.io.

jirutka:
@FichteFoll By the way, [Asciidoctor](http://asciidoctor.org/) is the current _implementation_ of the AsciiDoc _format_. The [previous implementation](http://asciidoc.org/) in Python, called AsciiDoc, is de facto deprecated. That’s the reason why I’ve named it “Asciidoctor (AsciiDoc).”

So, should I rename it to just “Asciidoctor”?

mojavelinux:
> called AsciiDoc

I like to call it AsciiDoc Python just to be super clear.

Asciidoctor is not only the leading implementation, it's now the official AsciiDoc language maintainer, so technically it would suffice to just call it AsciiDoc. Even GitHub & Gitlab now use Asciidoctor 1.5.2, so Asciidoctor is AsciiDoc.

FichteFoll:
>So, should I rename it to just “Asciidoctor”?

Imo yes, but I can only give recommendations.

>Asciidoctor is not only the leading implementation, it's now the official AsciiDoc language maintainer, so technically it would suffice to just call it AsciiDoc.

The issue here is that the "AsciiDoc" package exists already. Technically, I have full reign over it however, so I could decide whether it should remain in the package list or Asciidoctor gets to repalce it (with any name). This would be the preferred action since it will not result in a duplicate and will get the updates to all users of the current "AsciiDoc" package.

There are two concerns with this, however, that you can hopefully solve for me:

1. Asciidoctor is flagged for ST3 only. The only relevant parts (Python plugins) seem to be compatible with ST2 however, so after a potential brief test on ST2 you should be free to flag it as compatible with everything.
2. Current users of the AsciiDoc package might not like the differences between the current and the potential replacement. Could you elaborate how the two packages differ or maybe help me with the decision whether users will get used to the changes?

jirutka:
> Asciidoctor is flagged for ST3 only.

Yes, because, to be honest, I really don’t wanna bother with ST2 and Python 2.x _especially_. The package is in an early stage of development, I have intention to add many more plugins and also build actions. The decision to support ST2 may complicate the next development.

> Could you elaborate how the two packages differ.

Well, this is not easy to define, because these packages have basically nothing in common. The current AsciiDoc package is based on [asciidoc.tmbundle](https://github.com/zuckschwerdt/asciidoc.tmbundle) by Christian Zuckschwerdt. I’ve started from [AsciiDoc-TextMate-2.tmbundle](https://github.com/mattneub/AsciiDoc-TextMate-2.tmbundle) by Matt Neuburg. He rewrote the syntax definition from scratch, because:

> An advantage of writing the grammar entirely from scratch is that I was able to make a completely fresh start. Many of the problems with the previous AsciiDoc bundle were caused by incorrect assumptions, mostly having to do with the fact that it was based on the Markdown bundle, which was a huge mistake, as AsciiDoc is not particularly like Markdown and is certainly not related to it.

I modified some parts, dropped deprecated syntax and fixed some issues. It’s not finished yet, there are still some issues I need to fix, but it already works better than syntax definition in the current package (in documents that doesn’t use deprecated syntax).

I also rewrote all snippets; some of them are equivalent to those in the current package, some are replaced with more convenient keymaps, some are intentionally missing (e.g. those with deprecated syntax).

I believe that the new package will provide significantly better user experience (my goal is something like [MarkdownEditing](https://github.com/SublimeText-Markdown/MarkdownEditing)). However, it’s (intentionally) not backward compatible with the previous one, so we can’t exclude that some users will not like this change (e.g. those who still uses old AsciiDoc Python).

I’m not sure; it would be probably better to release the new package as Asciidoctor, keep the old one and maybe replace the old one someday in the future.

FichteFoll:
Thank you for your comment, this made the decision much easier for me. Yes, you should release this as a new package imo.

The only thing remaining is the package name now, I suppose. I'm still in favor of just "Asciidoctor" because it's simple, describes everything you need and does not contain a "description" in its name, which is what the actual `description` is for. Names should be just that, names.

jirutka:
> I'm still in favor of just "Asciidoctor" because it's simple, describes everything you need.

I agree. @mojavelinux, is this okay for you as well?

mojavelinux:
Yep. In light of the new information, I agree.

jirutka:
@FichteFoll Okay, I've already changed name to Asciidoctor, could you please merge it?

mrchief:
This is great @jirutka @FichteFoll @mojavelinux! 

I had created the ST2 package when I was working on my book and it was supposed to be a quick hack more than anything else. I haven't been able to spend a lot of time on the ST2 package but I'll be happy to hand over the reins to anyone who's interested. 

AsciiDoctor was news to me (really tells how up-to-date I'm on) and I agree with the discussion here. If there's lot of interest with ST2 I can help fix things but otherwise, I would say ST3 is the way to go.

mojavelinux:
:+1:

PR #4659, opened 2015-07-28
Added Pushbullet Plugin for Sublime Text 3

xarnze:


FichteFoll:
`requests` is available as a dependency, so you don't have to bundle it: https://github.com/packagecontrol/requests.

xarnze:
I've now changed this to be a dependency

FichteFoll:
[subtle pointer](https://github.com/xarnze/sublimepushbullet/blob/547f180b47ea682ca5c1cae9780db9ca82e0c0ad/sublime_requests.py#L3-L4)

I am also unsure how this is supposed to work on ST3 (as you flagged) because of your absolute imports (of `commandline` and `sublime_requests`). Did you perhaps intend to flag this as `<3000`?

xarnze:
These imports have now been removed

PR #4661, opened 2015-07-28
Added Java Bytecode repository/package

xdrop:


FichteFoll:
There already is a package for Javap: https://packagecontrol.io/packages/Javap, but since your package provides an additional syntax this is okay.

PR #4666, opened 2015-07-29
Adding WebSequenceDiagram Package

gaffo:


FichteFoll:

You have quite some debug prints in the code that the user will hardly appreciate. I suggest masking this with a debug setting (or the logging module).

I highly suggest adding a readme to tell your users (and me) what the package is about.

FichteFoll:
@gaffo ping

PR #4667, opened 2015-07-30
modify xtemplate details path

noyobo:
* rename github reprository path


FichteFoll:
homepage, author and readme keys are obsolute and fetched from the details url automatically (that's its only purpose).

noyobo:
@FichteFoll  Thank you for pointing out my mistake :smiley:

PR #4668, opened 2015-07-30
Update "ARIA attributes and roles" package

indrekpaas:
* Point to new repo
* Add more labels
* Update ST version

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -989,11 +989,11 @@
 		},
 		{
 			"name": "ARIA attributes and roles",
-			"details": "https://github.com/indrekpaas/ARIA",
-			"labels": ["aria", "auto-complete", "html"],
+			"details": "https://github.com/indrekpaas/ARIA-attributes-and-roles",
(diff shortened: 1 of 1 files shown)


FichteFoll:
I don't see a reason to restrict this package to ST2, .sublime-completion files haven't changed.

indrekpaas:
> I don't see a reason to restrict this package to ST2, .sublime-completion files haven't changed.

After an extensive research and tests I haven't been able to get `.sublime-completion` extensions to work on ST3. More and more plugin developers use python for writing autocompletion plugins on ST3 or choose to use snippets.

Since python don't fit in my toolbox and until I rewrite this plugin using snippets, I'd like to limit ST3 users installing a non-functional plugin.

FichteFoll:
Taking a closer look, I think the reason why the completions won't show up in ST3 because the HTML completions are provided by a Python plugin that hides all other completions by default. You are correct in that it is better not not allow installation on ST3 due to that.

PR #4670, opened 2015-07-30
Add Splunk Conf File Syntax Highlighting package

shakeelmohamed:


FichteFoll:
https://packagecontrol.io/packages/INI covers this quite nicely (.conf uses an INI-like syntax), so I don't see the need for this package. If you want .conf files to automatically open with the INI syntax, either propose a pull request that adds the extensions to that package or just configure it manually via *View > Syntax > Open all with current extension as…*.

shakeelmohamed:
Thanks for the suggestion, I tried it out the INI package but it doesn't do quite what I want.

We also want this package for discoverability. Plus, there's a special warning for default/example Splunk .conf files which you can see in the below screenshot:

screenshot

FichteFoll:
>I tried it out the INI package but it doesn't do quite what I want.

Care to explain?

shakeelmohamed:
Stanzas aren't highlighted, see the screenshot (this is with the same theme)

capturfiles_1

FichteFoll:
I suppose this boils down to personal preference, although I have to agree that the scoping is a slight bit weird (see [here](https://github.com/clintberry/sublime-text-2-ini/blob/master/INI.tmLanguage)). However, the stanzas *should* be highlighted as `entity.section.ini`, so maybe your color scheme doesn't define a color for the `entity` scope?


Note that the INI syntax matches everything (except some whitespace) and that you can assign custom colors in your color scheme to pretty much any token, if you like.

shakeelmohamed:
I'm using my own theme, monokai best, but switching to monokai didn't change anything.

I still think this is a reasonable addition to package control for discoverability, if nothing else.

FichteFoll:
I don't necessarily agree on discoverability, since everyone should know that this is definitely an INI-like format. I wouldn't even consider searching for "Splunk Conf" because of that.

That said, if you *really* think that this needs discoverability, then so be it...

shakeelmohamed:
> That said, if you really think that this needs discoverability desperately, then so be it...

Don't make me beg! :smile:

FichteFoll:
>Don't make me beg!

Oh, I'm sure you would like to.

PR #4672, opened 2015-07-31
Add Responsys Built-In Function

tatky:
This is the language definitions for Oracle Responsys Built-In Function.

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -779,22 +779,22 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:

Some file sytems are case sensitive, so you should adjust the casing of your message reference in `messages.json`.

tatky:
Thank you for your review.
Fixed a version to 1.0.0.
File name of README.md modified to lower case.

PR #4674, opened 2015-07-31
adds Browser Sync package

iamdjones:


FichteFoll:
Your package is structured in a weird way that won't work properly because plugin files are not loaded. They must be in the parent directory.
You can develop the "branches" independantly by doing what your redundant `package.json` suggests: split package into three real branches, move everything from `st3_*` to the root and push differently prefixed tags from the branches. Alternatively you can just put everything in one directory and adjust the code to only run the OS-specific binaries.

iamdjones:
@FichteFoll I agree. I'm going to take the route you suggested lastly, choose the specific OS binary at runtime. I appreciate your comments.

PR #4675, opened 2015-08-01
Add MDL Language package

ardenpm:
Add new language definition for the NVIDIA MDL (Material Definition Language) Language. This language defines the appearance of materials for photorealistic renderers such as the NVIDIA Iray renderer. MDL is a free, published language for making materials.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -547,6 +547,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
You can allow installation on ST2 without fear, the file format hasn't changed.

PR #4676, opened 2015-08-01
Wildlife Sublime theme

tushortz:
A Vibrant Sublime text theme with cool and rich markup syntax highlighting


FichteFoll:
1. This PR is a total mess. Please squash (after you did the following changes).
2. Don't add the package in a custom repository.json and again as a random file in the repo. Just add it to `repository/w.json`.
3. Your labels are obscure. I highly suggest browsing through https://packagecontrol.io/browse/labels and adopt a few of these instead.

tushortz:
Hi @FichteFoll thanks for the reply and trying to put me in the right path. I actually havent done this before and struggled a lot. Do I need o put a package.json in my original wildlife repository and

After adding my json code into the repository/w folder is that enough or should I also put the raw.githubcontent/username/themename/master/package.json in the channel.json file

FichteFoll:
No, you should only list the package in `repository/w.json` and not reference your `package.json`.

The tests must pass however: https://packagecontrol.io/docs/submitting_a_package#Step_7

tushortz:
I made changes and reran the test but now I get an error

 AssertionError: Error while reading './repository/w.json': ValueError('No JSON object could be decoded',)

FichteFoll:
Alright, this looks good now. However, you made many unnecessary changes and commits in the process. If you can, please squash the commits into a single one, or start all over on a fresh branch if you don't know how and don't want to read into it.

tushortz:
@FichteFoll  i'll be interested in knowing how to squash them please if you can explain

FichteFoll:
Honestly, google can explain this better than I could.

tushortz:
I think ill start afresh but one more question. I added 
https://raw.githubusercontent.com/tushortz/wildlife-sublime-theme/master/packages.json

to the channel.json     is that allowed

FichteFoll:
Allowed yes, but unfavorable because we do not have access to that file and if the package schema happens to get updated (which has happened a few times before), we can not migrate the current data automatically. There are still packages from the 1.0 and 1.1 days that do not have a readme specified because that was only possible in 2.0 and the authors didn't update their `packages.json`. So unless it's necessary, which it is not for you, please refrain from adding a custom `packages.json` file to `channel.json`.

tushortz:
ok thanks

FichteFoll:
Continued in #4689

PR #4650, opened 2015-07-23
Added Copy Filename package

shivapoudel:
Previously there is a package *Copy File Name* just below of this package configs which works on Sublime Text - 2 and hasn't been updated for last 2 years and PR was not reviewed since 2014. So I have created this package. Please have a review and give it a merge.
This is my first package volunteered if something is missing and let me know xD

This works on Sublime Text - 3 ;)

PR #4680, opened 2015-08-03
Added Copy Filename package

shivapoudel:
Previously there is a package Copy File Name just above of this package configs which works on Sublime Text - 2 and hasn't been updated for last 2 years and PR was not reviewed since 2014. So I have created this package. Please have a review and give it a merge.
This is my first package volunteered if something is missing and let me know xD

This works on Sublime Text - 3 ;)

@wbond Reference #4650

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2102,17 +2102,6 @@
 			]
 		},
 		{
-			"name": "Copy File Name",
-			"details": "https://github.com/nwjlyons/copy-file-name",
-			"previous_names": ["copy-file-name"],
-			"releases": [
-				{
-					"sublime_text": "<3000",
-					"branch": "master"
-				}
-			]
-		},
-		{
 			"name": "Copy Filename",
 			"details": "https://github.com/shivapoudel/sublime-copy-filename",
 			"releases": [
--- a/repository/c.json
+++ b/repository/c.json
@@ -2113,6 +2113,16 @@
 			]
 		},
 		{
+			"name": "Copy Filename",
+			"details": "https://github.com/shivapoudel/sublime-copy-filename",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Copy Filepath With Line Numbers",
 			"details": "https://github.com/theskyliner/CopyFilepathWithLineNumbers",
 			"releases": [


FichteFoll:
1. Remove `package-metadata.json`, it is supposed to be created by Package Control on installation.
2. Remove `.no-sublime-package`, you don't need it and it limits override capabilities.
3. Your changelog seems ... wrong?
4. This is the same functionality as https://github.com/nwjlyons/copy-file-name, except that places the menu item at a different location and has not been flagged to support ST3, even though it does. cc @nwjlyons

shivapoudel:
@FichteFoll above listed 1, 2 & 3 are solved:
https://github.com/shivapoudel/sublime-copy-filename/commit/6f7ef512f39f00eee447c6b23d9c9269ba059ac8
https://github.com/shivapoudel/sublime-copy-filename/commit/7068511e4c7644ab15d7708075451ab74cc7b2cf

shivapoudel:
@FichteFoll ping :)

FichteFoll:
What about 4.?

shivapoudel:
@FichteFoll I got confused what you mean to say in 4. Please can you explain briefly :worried:

FichteFoll:
I mean that there already is a package that provides the exact same functionality. You probably only missed it because it has not been flagged to support ST3 (because it is older than ST3), but it should work perfectly fine.

I would rather just flag the existing package as ST3-compatible than add another package with the same functionality.

shivapoudel:
@nwjlyons unfortunately I am really very very sorry to remove that package. Because the `Sid Bar.sublime-menu` is annoying because its not working perfectly.

shivapoudel:
I am sorry but since there is no any updates to that previous package I have to solve 4 by removing previous package. Can you please give this a merge. @nwjlyons if we are going to do wrong then just ping and say to stop us.

@FichteFoll can you fix this conflict and give a merge as we need this feature badly in our team members with keymap enable. Otherwise we have to copy file path and only copy basename which is time consuming.

nwjlyons:
Is the problem here a namespace issue?

I think all I need to do is update the package_control_channel to specify my package works with ST3. I can do that in the next few days.

shivapoudel:
@nwjlyons I will send some PR for you package tonight :)

FichteFoll:
Exactly, @nwjlyons. I will close this PR then.

PR #4683, opened 2015-08-03
Add 'Open Search in a New Tab' plugin

everyonesdesign:
The plugin makes Sublime open each find in folder search in a separate tab


everyonesdesign:
@FichteFoll 
OK, I added semver tag and changed main branch field to tags: true
Thank you for help

PR #4688, opened 2015-08-04
Add 'Jinja2 Better Syntax' plugin

everyonesdesign:


FichteFoll:
Well, this is obviously a fork of https://github.com/kudago/jinja2-tmbundle, as you also noted in https://github.com/mitsuhiko/jinja2-tmbundle/issues/13. It really does seem abandoned though. Imo, it makes more sense to "just" replace the reference to the original repo with your fork, but I won't do that just because.

@mitsuhiko, if you would like @everyonesdesign to take over your package on Package Control and thus keep users updated, or don't reply within ~14 days, I will initiate steps to do so.

everyonesdesign:
@FichteFoll 
Please, consider the merge, more than 2 weeks have passed by now in silence.

FichteFoll:
Okay.

Please replace the original "Jinja2" package instead, by removing your addition in this PR and changing the original's url and `master` to `tags`. You might also want to provide an [update message](https://packagecontrol.io/docs/messaging).

everyonesdesign:
OK, I changed the plugin's details, please, review it
I don't think I can provide an informative message, maybe later I'll make a changelog

PR #4694, opened 2015-08-07
Add open in eclipse plugin

hbhakhra:


FichteFoll:
This package just opens executes an OS command with `subprocess.Popen`. There are several packages that support a superset of this, such as https://packagecontrol.io/packages/Commando, and thus make this package very redundant.

PR #4698, opened 2015-08-12
add CSSfmt package

Redknife:


FichteFoll:
If you're calling "external" programs using bundled scripts or binaries, you need to add a `.no-sublime-package` file to the repo. Otherwise the package will be inside a .sublime-package archive.

FichteFoll:
@Redknife ping

Also, you need to rebase to current master.

PR #4701, opened 2015-08-13
added Scoggle package

ssanj:
Created a new Scala package to allow switching between production and test sources. Has three built-in matching strategies and has facilities for the creation of additional matchers.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -293,6 +293,18 @@
 			]
 		},
 		{
+			"name": "Scoggle",
+			"details": "https://github.com/ssanj/Scoggle",
+			"labels": ["testing", "scala"],
+			"releases": [
+				{
+					"sublime_text": ">=3083",
+					"platforms": ["osx", "linux"],
+					"tags": true
+				}
+			]
+		},			
+		{
 			"name": "Scold",
 			"details": "https://github.com/Xion/SublimeScold",
 			"labels": ["code_sharing", "email", "remote_collaboration"],


FichteFoll:
Use `from . import scoggle_types as stypes` for relative imports in ST3 that don't require messing with sys.path.

ssanj:
Hey @FichteFoll,

When I do that it breaks all my dynamic class loading from the matchers package. Also the tests which are in their own folder break. I tried prepending all package paths with "Scoggle." but that doesn't seem to work for the dynamic loading. I can create a branch with what I've got for you to review if you like?

I find Python packages themselves quite confusing. I presume when I modify the system path (which I agree is not good) it adds the current directory onto the package path so I can reference matchers classes as matchers.python_class_name.ClassName. If I don't I presume that ST3 adds the plugin directory to the package path, so then I need to reference classes as Sublime_Plugin_Name.matchers.python_class_name.ClassName.

Any help is appreciated.

ssanj:
Hey @FichteFoll,

I've removed all calls to sys.path. Please have a look.

FichteFoll:
ST adds the Packages path to `sys.path` (and a custom loader to `sys.meta_path` for that matter, which handles the loading of `.sublime-package` files), so every module in a package is accessible with `PackageName.path.to.module`. For your tests (or more specifically everything that is in a sub package, i.e. subdirectory) you need to use two dots with relative imports so that you go up a level, such as `from .. import scoggle_types as stypes`.

For the dynamic class loading, you can use `__package__` instead of `Scoggle.matchers` because the two are equivalent in the average situation and the first will still be correct if the package was saved under a different name. I'm not exactly sure about tests, but they should work as well. There are other methods available for doing what you intend to do (e.g. importing the classes into `matcher.__init__` and then using `getattr`), but yours should work "just fine".

ssanj:
Thanks for the feedback and the merge @FichteFoll. :)

PR #4702, opened 2015-08-13
Add CSSfmt

dmnsgn:
Plugin for https://github.com/morishitter/cssfmt

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2638,6 +2638,17 @@
 			]
 		},
 		{
+			"name": "CSSfmt",
+			"details": "https://github.com/dmnsgn/sublime-cssfmt",
+			"labels": ["css", "formatting"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"details": "https://github.com/lcdsantos/CSSFontFamily",
 			"releases": [
 				{


morishitter:
Hi I am @morishitter, author of CSSfmt. Please accept this PR as the official ST3 plugin.

FichteFoll:
I must have missed this somehow since I don't remember checking this package, nor had unread notifications.

@dmnsgn, `
BIN_PATH = join(sublime.packages_path(), dirname(realpath(__file__)), 'cssfmt.js')` on the module level will fail on ST3 because the API is not populated at module load time. Just remove `sublime.packages_path()` from the call, it doesn't make sense anyway. (I wonder how this works for you.)

While I personally like the usage of an iterator for sublime.Region, monkeypatching of `sublime.Region` raises the following concerns:

1. Because the sublime module (just like any module) is shared in the entire plugin environment, other plugins may monkeypatch the module differently and depend on their implementation. This is quickly apparent because you made the iterator iterate over `.a` and `.b` while other people may prefer `.begin()` and `.end()` (which is also how you appear to be using it).
In other news: Just define `has_selection(self)` as `return any(not reg.empty() for reg in self.view.sel())`, which is functionally equivalent, way shorter and does not need a monkeypatch.

dmnsgn:
Thanks @FichteFoll 
Will do asap

dmnsgn:
@FichteFoll https://github.com/dmnsgn/sublime-cssfmt/blob/master/CSSfmt.py is updated with your previous comments (no monkey patch and updated BIN_PATH). Though I am not sure what you mean by "re-raising the exception after you showed the eror message dialog".

FichteFoll:
I was referring to [this section](https://github.com/dmnsgn/sublime-cssfmt/blob/c5109a14cad68b6ba6d509c28b3a706103d55a30/CSSfmt.py#L37-L40).

PR #4704, opened 2015-08-14
Added perl-Test-Class extension

jonasbn:
Another Perl syntax package for Perl's Test::Class I created a long time ago, but I forgot to request its possible addition to the package control channel.

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -472,6 +472,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:

jonasbn:
I have made the changes requested and done the same for perl-Test-More, for which I am also the author.

PR #4706, opened 2015-08-15
Added Xkcd plugin

eivind88:
Added new plugin [Xkcd](https://github.com/eivind88/xkcd)

### DIFF
--- a/repository/x.json
+++ b/repository/x.json
@@ -65,6 +65,16 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
`"sublime_text": "<3000"` indicates that this package should only be installable for ST2.

eivind88:
Thanks for catching that!
I've updated it to ```"sublime_text": ">=3000"```.

PR #4714, opened 2015-08-18
Add GotoWindow package

ccampbell:
This pull request adds a new sublime package I wrote to quickly jump around between open sublime windows using the keyboard.  It works like `super + t` except for open windows instead of files in the active window.

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -1181,6 +1181,16 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
Please remove `package-metadata.json`, it is supposed to be generated by Package Control on installation.

ccampbell:
@FichteFoll done!

PR #4716, opened 2015-08-18
Add "Test Plier" package

asfaltboy:
Sublime Test Plier is a simple yet "smart" python test picker/runner, unlike other test runners uses AST instead of regex.


FichteFoll:

asfaltboy:
@FichteFoll updated version to 1.0.0

PR #4717, opened 2015-08-18
Add CFDocs plugin to package control

mikesprague:


FichteFoll:

The `readme` key is supposed to point to a raw file, not a html page. It will be automatically fetched if you simply remove it.

Is there a certain reason why you didn't name your package "CFMLDocs"?

mikesprague:
Thanks for the quick feedback. I have made the suggested updates and pushed a v1.0.0 tag to the package repo. 

I chose to name it CFDocs (versus CFMLDocs) because the resource it searches is also named CFDocs. If you see this as an issue for any reason, please let me know. Thanks!

FichteFoll:
>If you see this as an issue for any reason, please let me know.

No, I was just curious.

PR #4720, opened 2015-08-19
Package for AspectJ

pchaigno:
Package is hosted at [pchaigno/sublime-aspectj](https://github.com/pchaigno/sublime-aspectj).
It only takes care of syntax highlighting for now.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -1179,6 +1179,17 @@
 			]
 		},
 		{
+			"name": "AspectJ",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Note that, by default, your package is not installable because it only has a "pre-release" version and that requires users to explicitly opt-in into installing.

pchaigno:
@FichteFoll Thanks! I'll release a first official release in a couple weeks ;)

PR #4722, opened 2015-08-20
commit package QuickOpen

zsytssk:
https://github.com/zsytssk/QuickOpen


FichteFoll:

PR #4724, opened 2015-08-21
add quick open

zsytssk:
add quick open


FichteFoll:

zsytssk:
changed

PR #4725, opened 2015-08-21
Added IMSwitchForVintage Package

zhongxingdou:


FichteFoll:
`package-metadata.json` is supposed to be created by Package Control on installation, so you should delete it from the repo.

If your package works only on OSX you should restrict installation to that OS, using the `platforms` key.

FichteFoll:
You need to add a `.no-sublime-package` file to the repo because you are trying to execute the `im` file in your package, but packages are stored in zip archives by default.

PR #4731, opened 2015-08-23
add React Templates

individuo7:


FichteFoll:
Before I merge this I want to mention that your release schema is flawed because it's impossible to release more than one version on a single day. You may want to consider this before I merge, because it's pretty much irreversible (since you're starting with 2015 as major number).

individuo7:
you are right, I fix that adding a number at the end

FichteFoll:
Unfortunately that won't do either because we require you to use semver-like versions, which have exactly three numeric identifiers. You should check it out: http://semver.org.

individuo7:
ok I got it, then I change the version number to v1.0.0

PR #4732, opened 2015-08-23
added Padawan.sublime(Padawan.php integration) package

mkusher:


FichteFoll:
Due to the way Package Control downloads repositories (via zip source archives) and github not including submodules there, you can not use submodules like this. Consider using `git subtree` instead.

mkusher:
ok, sorry. Will fix this soon

mkusher:
@FichteFoll fixed submodule problem and moved to semantic versioning

FichteFoll:
You should not be using dots in the package name, this will cause major issues with loading the Python plugin files.

mkusher:
Damn I, sorry. Fixed version and remove `.php` from package name

PR #4733, opened 2015-08-24
Updated repo link address Coffee2Js

mattseymour:
I have / am renaming the repo so the repo name and plugin name are the same.


FichteFoll:
FYI, you can rename repositories on github which preserves all data (and even sets up redirects) instead of creating a new one.

PR #4735, opened 2015-08-24
Add IAS package

rmobis:


FichteFoll:
You use the `.sublime-syntax` format, which has only been added in the latest dev cycle (and fleshed out in 3092), so please change your version selector to `"sublime_text": ">=3092"`.

rmobis:
@FichteFoll My bad, I noticed this after opening the PR but forgot to push the fix. Done now.

PR #4736, opened 2015-08-25
Add "Today Color Scheme" package

juliotrigo:
Add a new color scheme package called "Today Color Scheme".

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -551,7 +551,7 @@
 					"tags": true
 				}
 			]
-		},	
+		},
 		{
 			"name": "Theme - Bamboo",
 			"details": "https://github.com/gzhihao/bamboo-theme",
@@ -562,7 +562,7 @@
 					"tags": true
 				}
 			]
-		},		
+		},
 		{
 			"name": "Theme - Brackets",
 			"details": "https://github.com/leszekr/Theme-Brackets",
@@ -1564,6 +1564,17 @@
 				{
 					"sublime_text": "<3000",
 					"branch": "master"
+				}
+			]
+		},
+		{
+			"name": "Today Color Scheme",
+			"details": "https://github.com/juliotrigo/today-sublime-color-scheme",
+			"labels": ["color scheme"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
 				}
 			]
 		},


FichteFoll:
So, uhm, what are the differences to the original?

juliotrigo:
I originally started to work on this new color scheme, based on the *Tomorrow Night Eighties* theme, in order to have a color scheme to use at work, since I could not find any theme which I was completely happy with: for the ones I tried, either I did not like the colours or I did not like the way colours were applied to the different languages.

Even though *Tomorrow Night Eighties* and the new *Today* color schemes share some of their colours (not how they are applied), they are two completely different themes. You can find some of the main differences between the two themes below (you can see some image examples in the *Today* repository):

- **The rules that apply colours to the different languages / formats are different**. Having a look at the source code of both them, you will find that *Today* introduces more new and different rules. Some examples are:
  - The main color for both language reserved words and operators is orange. I think it helps readability combining those using the same color. Functions, arguments, class names, parent classes and several other basic parts are also highlighted differently and accordingly.
  - Method special variables like `this` or `self` share the same color with method arguments. This way they are easy to spot and follow. They are also in *Italic*, like method arguments, to keep consistency.
  - All the function calls, sub-calls and their arguments (things like `_sort_data(stripped_info).strip_args(params)`) are not coloured using the same color now (blue in *Tomorrow*). I don't think this helped readability.
  - Integers always use the same color (Turquoise), even if they are part of the key=value pair argument section.
  - **HTML / XML**: *Tomorrow* uses the red color for both element names and attribute names, making more difficult to spot the difference. *Today* uses blue for HTML elements and red for attributes, making it easier and nicer to read in my opinion.
  - **CSS**: both schemes use different colours here. Things like `1.5em` are now split into 2 differentiated parts with different colours. The main color contrast between basic structures like `footer#bottom-bar` is now blue-red, which I think it reads better.
  - **SQL**: the main colours here are orange-yellow-purple, which I find nicer. Operators and reserved words use the same color now (orange, which is the main color of the theme). Things like `pa.product_code`, where we use table attributes are now highlighted in yellow. They were originally white, like `product_code`, which does not include the table namespace. I think this was not good and nice to read.
- **Comments**: they are in *Italic* and use a different grey color, less bright, which helps readability and let you focus on the rest of the code better.
- **Font styles**. Some parts of the code are in *Italic*, like inherited classes, function arguments, comments, language reserved words and functions (`None`, `set`, `r`, `unicode`, 'self', 'this', etc.), etc., which I think helps readability.

With *Today*, I've mainly focused on the following languages / formats: Python, CSS, HTML, SQL, JavaScript and JSON; but all the languages that I have tried with it look nice and, at least, better than using the other schemes I've tried.

As I said, I originally started modifying the *Tomorrow* theme, but over the months, the number of differences in styles and how and where colours were applied continuously increased, until this fork became a completely different theme. This is the theme I have been using for moths and I thought I could make it public so more people could try it out and benefit from it.

Please let me know anything else you would like to know about this.

FichteFoll:
Thanks for that (very) long list, it makes my decision easier. You should probably add that to your readme.

juliotrigo:
Thank you for merging it and for your suggestion.

PR #4737, opened 2015-08-26
Adding HackerRankHelper Plugin

dineshkumar-cse:
I've added my plugin https://github.com/dineshkumar-cse/SublimePluginHackerRank by adding packages.json to channel and tested with ChannelRepositoryTools.

### DIFF
--- a/channel.json
+++ b/channel.json
@@ -35,6 +35,7 @@
 		"https://raw.githubusercontent.com/damccull/sublimetext-SolarizedToggle/master/packages.json",
 		"https://raw.githubusercontent.com/danielmagnussons/orgmode/master/packages.json",
 		"https://raw.githubusercontent.com/danielobrien/sublime-DLX-syntax/master/packages.json",
+		"https://raw.githubusercontent.com/dineshkumar-cse/SublimePluginHackerRank/master/packages.json",
 		"https://raw.githubusercontent.com/enriquein/JavaSetterGetter/master/packages.json",
 		"https://raw.githubusercontent.com/farcaller/DashDoc/master/packages.json",
 		"https://raw.githubusercontent.com/FichteFoll/sublime_packages/master/package_control.json",
--- a/channel.json
+++ b/channel.json
@@ -35,7 +35,6 @@
 		"https://raw.githubusercontent.com/damccull/sublimetext-SolarizedToggle/master/packages.json",
 		"https://raw.githubusercontent.com/danielmagnussons/orgmode/master/packages.json",
 		"https://raw.githubusercontent.com/danielobrien/sublime-DLX-syntax/master/packages.json",
-		"https://raw.githubusercontent.com/dineshkumar-cse/SublimePluginHackerRank/master/packages.json",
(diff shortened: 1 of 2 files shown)


FichteFoll:
- Are you sure that you ran the correct tests with ChannelRepositoryTools? Because your data is invalid.

- Please consider adding the package to the `repository/*.json` files directly instead, because it gives us more control over potentially updating your package to a newer schema version or correcting other errors. There are so many custom json files linked that still use 1.2 or 1.1 schema, while we are at 3.0.0.
- Instead of bundling the requests package, you could make use of the new dependencies feature and include requests following [these instructins](https://github.com/packagecontrol/requests#how-to-use-requests-as-a-dependency). It also won't require messing with `sys.path`.
By the way, due to a bug `__init__.py` in a package root is loaded twice by Sublime Text.

dineshkumar-cse:
Hi, i've added my package in repository.json and [Updated my plugin ot include requests as depnedency] (https://github.com/dineshkumar-cse/SublimePluginHackerRank/commit/771dac79b665b5934584cf504560f0e59896adde?diff=unified)
* I thought adding remote json file might help me not to ask for pull requests in case of future change.
* what data is invalid? can you explicitly mention just to know it.
* I tested with Test Default Channel (passes), TestRemoteRepository (not current use) passed, but with remote it fails coz `https://raw.githubusercontent.com/jvantuyl/sublime_diagram_plugin/master/packages.json` is missing. or do we need to run any other tests?

FichteFoll:
>  I thought adding remote json file might help me not to ask for pull requests in case of future change.

You will very rarely have to change the package data in the future, most likely only if you want to change the version contraints or rename the package.
Judging from your incredibly weird `packages.json`, it seems like a good idea to have your changes reviewed anyway.

> what data is invalid? can you explicitly mention just to know it.

It should be quite obvious once you actually checked out the [`example-repository.json`](https://github.com/wbond/package_control/blob/master/example-repository.json) file, but:

- Packages must be placed in a `packages` array.
- `home_page` is invalid, it must be without the underscore. It's redundant anyway because that's the default.
- You can only provide either a branch or a tag key in a release.

> I tested with Test Default Channel (passes), TestRemoteRepository (not current use) passed, but with remote it fails coz https://raw.githubusercontent.com/jvantuyl/sublime_diagram_plugin/master/packages.json is missing. or do we need to run any other tests?

"Test Local Repository (This File)" is what you should have run, while editing your `packages.json`.

---

Now, there are a few additional issues:

1. Please use the same indentation style as everywhere else in the file. It is really ignorant to not follow established styles (indentation or coding style in general) in pull requests, especially if they are *very* obvious.
2. You'll need to change `"branch": "master"` to `"tags": true` and use [tags](http://git-scm.com/book/en/v2/Git-Basics-Tagging) to create releases and push valid [semantic version](http://semver.org) tags to your repo. As mentioned [in the docs](https://packagecontrol.io/docs/submitting_a_package#Step_4), we no longer accepts branch-based releases for new packages.

PR #4738, opened 2015-08-26
Add package kroman

cheunghy:
Kroman is a package that romanize korean.

Features:

Romanize region.
Romanize file.
Romanize file and put text into new window.


FichteFoll:

Also, please consider adding a readme to your package to inform users what it is about.

cheunghy:
I've updated semantic version and added a readme.

cheunghy:
Thanks!

PR #4739, opened 2015-08-26
Addition of parva language syntax highlighting

EternalDeiwos:
https://github.com/muskatel/Parva_Syntax


EternalDeiwos:
@FichteFoll so 0.1.0 would be valid?

FichteFoll:
Yes.

PR #4743, opened 2015-08-28
Added the package "CaretJump".

r-stein:
The package CaretJump adds the ability to start search from every caret and jump with the caret or create a new caret at the target selection.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -251,6 +251,17 @@
 			]
 		},
 		{
+			"name": "CaretJump",
+			"details": "https://github.com/r-stein/sublime-text-caret-jump",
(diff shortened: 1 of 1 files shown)


FichteFoll:
There already are the similar packages https://packagecontrol.io/packages/JumpTo and https://packagecontrol.io/packages/SelectUntil, except that they only add a caret and not a selection as far as I can tell. Maybe you could consider adding some of the different functionality to one of those?

r-stein:
Thanks, I should have seen the package [JumpTo](https://packagecontrol.io/packages/JumpTo). My features would nicely integrate into it. Hence I close this pull request.

PR #4744, opened 2015-08-28
Add Exalt

eerohele:
A plugin for wrangling XML.

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -745,6 +745,18 @@
 			]
 		},
 		{
+			"name": "Exalt",
+			"details": "https://github.com/eerohele/exalt",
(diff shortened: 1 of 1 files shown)


FichteFoll:
The name "Markup" seems a bit too general to me since you are "only" validating and formating the XML (and SGML) family, which is quite a limited subset of all the markup languages. I suggest finding a different name that is more specific.

eerohele:
That's fair enough.

I suppose the name doesn't have to be too "literal". Would you accept "Exalt"? For, uh... "**X**ML v**al**idation and forma**t**ing".

FichteFoll:
Sounds good to me. Just make sure to include all those keywords in the description so that they are searchable on packagecontrol.io.

eerohele:
Renaming done.

PR #4748, opened 2015-08-31
Add One Command Syntax package

yrsegal:


FichteFoll:
The lack of a readme for the actual "OneCommand" repo is kind of a bummer.

FichteFoll:
Oh, I actually forgot to mention this, but you can use relative imports in Sublime Text, via `from .OneCommand import oneCommand` (ST3) and `from OneCommand import oneCommand` (ST2). That way you don't need to mess with sys.path.

yrsegal:
Yeah, agreed (but it's technically @destruc7i0n's project so I don't want to touch it)

yrsegal:
Fitchte, it wasn't working for some reason so I used sys.path.

PR #4751, opened 2015-09-01
Added SublimeLinter - CatGutterTheme

m10l:
Created a SublimeLinter gutter theme. :heart_eyes_cat: 

screen shot 2015-09-01 at 14 18 02

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -3035,8 +3035,8 @@
 			"labels": ["linting"],
 			"releases": [
 				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
I am fairly certain that gutter themes are only for SublimeLinter 3, which only works on ST3. So you have your version selector reversed.

m10l:
@FichteFoll Ah thanks for the help. I got a lil confused by the guide I was following. Updated as per your recommendations now.

PR #4752, opened 2015-09-01
Add "Lessy" Plugin

FabioFuschi-SNK:
Added the Lessy Plugin, that convert less inline to CSS

### DIFF
--- a/channel.json
+++ b/channel.json
@@ -36,7 +36,7 @@
 		"https://raw.githubusercontent.com/danielmagnussons/orgmode/master/packages.json",
 		"https://raw.githubusercontent.com/danielobrien/sublime-DLX-syntax/master/packages.json",
 		"https://raw.githubusercontent.com/enriquein/JavaSetterGetter/master/packages.json",
-		"https://raw.githubusercontent.com/FabioFuschi-SNK/lessy-sublimeplugin/master/packages.json"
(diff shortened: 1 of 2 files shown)


FichteFoll:
Please remove your addition to `channel.json`. That method is outdated, your package schema is outdated and applied inproperly.

Instead, you'll need to change `"branch": "master"` to `"tags": true` and use [tags](http://git-scm.com/book/en/v2/Git-Basics-Tagging) to create releases and push valid [semantic version](http://semver.org) tags to your repo. As mentioned [in the docs](https://packagecontrol.io/docs/submitting_a_package#Step_4), we no longer accepts branch-based releases for new packages.

Furthermore, I think you are using `dependencies.json` improperly because there is no `lessc` dependency available. Read here for information about dependencies: https://packagecontrol.io/docs/dependencies

FichteFoll:
Also, the way you invoke lessc is 1) unflexible because you hardcode the path, and 2) incompatible with Windows.

FabioFuschi-SNK:
Hello,

thank you for the help. i made the followed changes now.

- removed channel.json changes
- change "branch": "master" to "tags": true
- remove dependencies.json

on the Plugin:
- remove hardcoded path
- add windows support


i hope i done it right now.

PR #4753, opened 2015-09-01
Add FileReloader package to f.json

petermac-:


FichteFoll:
"revert" is a view command, so you don't need to go the exra way of singling out groups and going over views there while always ensuring the view is focused. You can just

```py
for view in self.window.views():
    view.run_command("revert")
```

PR #4756, opened 2015-09-03
Update Ensime packge branches and labels

rorygraves:


FichteFoll:
The provided `issues` link is already the default, so it is redundant.

Would you be interested in using tags-based releases instead of commit-based (i.e. all commits on a branch)? It allows better version transparency and update messages and commit-based releases are deprecated and not accepted for new packages anymore.

Besides, maybe you want to allow ST2 users to install releases from your ST2 branch, or rather from specially tagged commits that have a `st2-` prefix? Example: https://github.com/FichteFoll/sublime_packages/blob/42615c564147a2337544ffd0066983a5392975c0/package_control.json#L5-L18

rorygraves:
@FichteFoll  Thanks for the detailed feedback, I have updated the PR.  I have moved to using tags as you suggest (I missed that it had been deprecated).  
We have dropped ST2 support at this point as we simply do not have the resource to keep them both running at the moment.

Thanks

PR #4761, opened 2015-09-06
Wildlife gutter theme first release

tushortz:


FichteFoll:
I suggest using "gutter theme" in the package name, like these similar packages: https://packagecontrol.io/search/gutter%20theme

tushortz:
I have renamed it to wildlife-gutter-theme if that's better

PR #4762, opened 2015-09-06
Add ATS2 Build System Plugin

ShengYun:
AddedATS2 Build System Plugin.

Currently there are no build systems for ATS2, so I created one. Please merge:)

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -1255,6 +1255,18 @@
 			]
 		},
 		{
+			"name": "ATS2 Build System",
+			"details": "https://github.com/ShengYun/ATS2-Build-SublimeText3",
+			"labels": ["build system"],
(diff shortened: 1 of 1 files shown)


FichteFoll:
You should check out build variants: http://docs.sublimetext.info/en/latest/reference/build_systems/configuration.html#variants

ShengYun:
Hi @FichteFoll, thanks for the information of variants, I just pushed a new version of my plugin, but when I'm using Package Install, it still shows the old 1.0.0 version, instead of my new 1.1.0 version. Is there anything I need to do? or just wait?

ShengYun:
@FichteFoll Just found it in the documents, it will update the packages in approximately 1 hour:) I should RTFM first:)

PR #4771, opened 2015-09-08
Updated version requirements for rosshadden/sublime-xpath.

rosshadden:


FichteFoll:
Are you sure you want to flag the package as ST2-only?

rosshadden:
No, haha my bad. Thanks for pointing this out.

rosshadden:
Updated.

FichteFoll:
Disregarding that branch-based releases are deprecated, this won't work because tag releases are always higher than commit releases in precedence.

You need to use different tag prefixes for the ST2 release.

rosshadden:
Updated.  I was doing what other packages had done in the json files, I didn't realize any of it was deprecated.  The [docs](https://packagecontrol.io/docs/submitting_a_package) don't say anything about setting `tags` to a string, but some other packages do this so I'm once again just kind of copying what I see.

Does this look good?  Thanks!

FichteFoll:
From the docs:

>For other options, please read the documentation in the example-repository.json file.

PR #4772, opened 2015-09-09
Add fluent-logger-sublimetext

airtoxin:


FichteFoll:
Parameterless `super()` only works in Python3 and thus ST3.

airtoxin:
@FichteFoll thank you for your review. I fixed it.

FichteFoll:
Also, missed it earlier but your name is inappropriate: https://packagecontrol.io/docs/submitting_a_package#Step_2

airtoxin:
To be honest, I knew the name is not good, but this plugin is logging tool for fluentd.
Plugins like this which is logging tool usually are named `fluent-logger-XXX`. (XXX is source.)
So, I just followed that.

airtoxin:
@FichteFoll any response?

FichteFoll:
I did a little reseach and didn't find the "fluent-logger-XXX" naming scheme for other plugins. I found "fluent-logger-<language>" for repo names on the fluent-logger org, but that makes sense for namespacing. It's not necessary in this isolated Sublime Text environment.

As an example, the package is named just [fluent-logger](https://pypi.python.org/pypi/fluent-logger) on pypi.

airtoxin:
@FichteFoll I see.

airtoxin:
↑ dependencies are writen in dependencies.json

FichteFoll:
No, pip (and pypi) is not used to resolve dependencies. Dependencies have to be manufactured uniquely for usage with Package Control, but you can do that. Refer to https://packagecontrol.io/docs/dependencies.

airtoxin:
@FichteFoll I tried zipped plugin. but its following command of 'open settings file' did not work. It opens empty file.

```
{
    "caption": "fluent-logger",
    "children":
    [
        {
            "command": "open_file",
            "args": { "file": "${packages}/fluent-logger/fluent-logger.sublime-settings" },
            "caption": "Settings - Default"
        },
        {
            "command": "open_file",
            "args": { "file": "${packages}/User/fluent-logger.sublime-settings" },
            "caption": "Settings - User"
        }
    ]
}
```

FichteFoll:
It works just fine for my packages. I don't really see an obvious error with your snippet either.

Are you sure the file is named `fluent-logger.sublime-package`?

airtoxin:
@FichteFoll really? My way is wrong? hmm. I did that way.

1. remove `.no-sublime-package`
2. `$ zip -r fluent-logger.sublime-package fluent-logger/`
3. move zipped package to `Installed Packages` directory
4. open `User/Package Control.sublime-settings` and add "fluent-logger" to installed_packages
5. then, restart ST3 and open __Sublime Text__ > __Preferences__ > __Package Settings__ > __fluent-logger__ > __Settings - Default__

it shows always empty file...

FichteFoll:
The way you create `.sublime-package` is errorneous. It creates an additional directory within the zip file, which will change the path of the files to `${packages}/fluent-logger/fluent-logger/fluent-logger.sublime-settings` and cause none of the Python files to be loaded because they are not in the package root.

Also, step 4 is redundant.

airtoxin:
@FichteFoll I removed `.no-sublime-package` file and it works all well. thank you!

PR #4776, opened 2015-09-09
Added Peg.js LiveScript syntax plugin.

tgrospic:
Language syntax for PEG.js with embedded LiveScript code.

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -432,7 +432,7 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
Please do not use dots in your package name. They are not causing problems right now, but will prevent Python plugins from being loaded on ST3 if you were ever to add one and using dots for new package names won't be accepted anymore.

tgrospic:
I hope it is sorted as expected.

PR #4778, opened 2015-09-10
change apicloud snippet plugin

taxusyew:
we need change this plugin's details to another place.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -728,8 +728,8 @@
 			]
 		},
 		{
-			"name": "APICloudSnippets",
-			"details": "https://github.com/taxusyew/apicloud-sublime-snippet",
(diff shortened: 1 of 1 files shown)


FichteFoll:
The APICloudLoader uses `shutil.copytree` on itself, which requires it to be unpacked on ST3. Add a `.no-sublime-package` file.

zhou-xinghai:
hi FitchetFoll,
  The shutil is a internal module support by python3 like module path,os. Is it require to unpack on ST3? If so, what content should be in .no-sublime-package file? I am a newbie here, so can you pls share more details.
Thanks

taxusyew:
The .no-sublime-package file had been added into the folder. please check that.

PR #4786, opened 2015-09-12
Add SublimeEasyPluginDevelopment.

leoxnidas:


FichteFoll:
I do not see anything particularly useful in this package, for the following reasons:

1. The package is required to use it and anything using it will just not work if it's not available.
2. It adds a level of more or less intransparent metaclassing to the code that doesn't really help much. Instead of `sublime_plugin.WindowCommand` you write `easy_plugin.SublimePlugin` - and that's it.
One of Python's conventions is "explicit is better than implicit", and this violates it greatly.
3. Your usage of `cwindow` is errorneous.
4. I don't understand why you are binding random functions from the sublime module to the class, and less so why you do it in the meta class.

Beyond that, the package is in an inacceptable state that I can't be bothered to go over now until the above is resolved.

Actually, I just refuse to add this. I hope you understand my decision.

PR #4787, opened 2015-09-13
add php dotenv highlighter package to the repository

kylestev:
Followed the instructions and the tests passed. :sunglasses:

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -699,6 +699,17 @@
 			]
 		},
 		{
+			"name": "PHP DotEnv Highlighter",
+			"details": "https://github.com/kylestev/dotenv-highlighter",
+			"labels": ["php", "dotenv", ".env", "language syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "PHP Form Builder",
 			"details": "https://github.com/migliori/sublime-phpformbuilder",
 			"labels": ["auto-complete", "bootstrap", "form", "php"],


FichteFoll:
This is just an INI-like format, isn't it? Seems redundant to add another syntax definition for essentially the same syntax.

kylestev:
@FichteFoll Thanks for the comments, I do believe that it is counterproductive to add new packages that provide the same functionality to Package Control.

The reason why I believe my package isn't just another INI highlighter is because the [vlucas/phpdotenv](https://github.com/vlucas/phpdotenv) (1,500+ Stars) format is a bit different than INI files, though similar. The real reason I created the package is because it is heavily utilized (by default) in the [laravel/laravel](https://github.com/laravel/laravel) PHP framework which has 18,600+ Stars on Github and it would be nice to have a package that provides basic linting and highlighting support for this specific file type.

I tried using the [mattbanks/dotfiles-syntax-highlighting-st2](https://github.com/mattbanks/dotfiles-syntax-highlighting-st2) package but it didn't provide the depth of features that I wanted -- which is fairly typical of packages with a broad target, even if they do work great for general cases.

FichteFoll:
>is a bit different than INI files, though similar.

From https://github.com/vlucas/phpdotenv/blob/master/tests/fixtures/env/ I found out that the notable differences are: `EXPORT` and `IMMUTABLE`/`MUTABLE` keywords, or some sort; variable interpolation in values like `${NVAR1}`. However, you syntax does not support these. If you add support for the special features of the .env syntax then I am willing to merge this.

PR #4791, opened 2015-09-13
Add vuejs complete repo

BrainBuzzer:
__Please__ merge this pull request.

### DIFF
--- a/repository/v.json
+++ b/repository/v.json
@@ -519,6 +519,18 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


repository/v.json
```diff
@@ -519,6 +519,18 @@
 			]
 		},
 		{
++			"name": "Vuejs Complete",
```
kylestev: You have committed `+` characters which is not valid JSON.

PR #4792, opened 2015-09-13
Add codic plugin

airtoxin:


FichteFoll:
Please use the requests dependency instead of bundling your own: https://github.com/packagecontrol/requests

Your install instructions are wrong.

airtoxin:
@FichteFoll i fixed them. I didn't know `dependencies.json` but it is very cool.

PR #4793, opened 2015-09-13
RhetosDSL syntax highlighting package

Hugibeer:
This package provides syntax highlighting for [Rhetos](https://github.com/rhetos/rhetos) development platform.

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -917,6 +917,18 @@
 			]
 		},
 		{
+			"name": "RhetosDSL",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Your package is new, it never had a name in Package Control's channel. You should remove the previous_names entry.

`.cache` files are redundant on repositories since they are just that, cache files, and regenerated on installation.

Hugibeer:
I hope this is OK now. 
Thank you.

PR #4799, opened 2015-09-16
added sublimious

dvcrn:
added sublimious package to package control


FichteFoll:
The tests don't pass: https://packagecontrol.io/docs/submitting_a_package#Step_7

dvcrn:
Must have overseen that. All good now!

PR #4802, opened 2015-09-17
Add iCalendar package

kimsey0:
Very simple [iCalendar](https://en.wikipedia.org/wiki/ICalendar) syntax highlighting.

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -48,7 +48,7 @@
 		},
 		{
 			"name": "iCalendar",
-			"details": "https://github.com/kimsey0/package_control_channel",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Your details url should point to https://github.com/kimsey0/iCalendar-sublime.

kimsey0:
Oops about the details URL. I've fixed that now.

Shall I omit the 'v' in front of the version number, which is traditionally used on GitHub?

FichteFoll:
The 'v' is optional, Package Control will detect and filter it.

kimsey0:
Great. Thanks!

PR #4810, opened 2015-09-18
Add BeautifyLatex

ketan:
Add a plugin to prettify latex code using latexindent.pl

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -213,6 +213,15 @@
 			]
 		},
 		{
+			"details": "https://github.com/ketan/BeautifyLatex",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "master"
+				}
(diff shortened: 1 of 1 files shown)


FichteFoll:

You do not run executables shipped in your package, thus you do not need `.no-sublime-package`. Please only use it when necessary because it is not possible to override package files that way.

ketan:
Thank you for getting back so soon.

> You'll need to change "branch": "master" to "tags": true and use tags to create releases and push valid semantic version tags to your repo. As mentioned in the docs, we no longer accepts branch-based releases for new packages.

Done - I also just created tag `0.0.1` under my repo.

> You do not run executables shipped in your package, thus you do not need .no-sublime-package. Please only use it when necessary because it is not possible to override package files that way.

Done!

PR #4816, opened 2015-09-20
Add Local Packages plugin

spywhere:


wbond:
How is this different than creating a `dependencies.json` for your package and running the `Satisfy Dependencies` command?

spywhere:
Well, the plugin goal is not to install the dependencies for the current project rather it install the packages in which missing from the user's installed packages.

The use case for me is when I work with my project, I required some packages to be installed in order to continue the work. Such as when I switch to different machine, some packages already installed, some are not, so this plugin will help in this use case. This could also contains some details configuration such as platform-dependent package or vice versa.

But if `dependencies.json` can fulfil my need, I am happy to remove the plugin from the channel and use the proper one instead.

wbond:
Ah, it could be terminology that is throwing things off. Package Control uses the term "dependency" to mean a specific type of Python package that is loaded before all other packages, with facilities to allow for OS/architecture specific shared libraries, and similar elements. And Package Control 3.1 added a file named `.sublime-dependency` that causes Package Control to recognize a local folder as a dependency to allow it to be installed. This helps during the development process.

So when I saw `.sublime-local-dependencies` I was curious as to what problem it was solving that was different.

It sounds like in your situation it is so you can define what Packages are necessary to properly work on a project.

spywhere:
Yup.

And yeah, the name `.sublime-local-dependency` is come up from `.sublime-dependency` since I cannot come up with a proper name other than "dependency". Sorry for making you confused with that.

PR #4818, opened 2015-09-20
Add PyFusion package

pdilyard:


FichteFoll:
Note that your `packages.json` file is redundant and won't be looked at.

PR #4819, opened 2015-09-20
add IndentTips

QS20199:
Demo

https://github.com/QS20199/IndentTips_sublime_plugin

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -327,6 +327,18 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
Please consider hosting the image elsewhere and not in the repository since Package Control will download it for every user and place it somewhere where it's not easily visible.

QS20199:
OK, I fixed the problem, please check it again~thanks~

PR #4823, opened 2015-09-20
Added 'kill' package to repository.json

goldsborough:
This is a plugin to kill rogue processes. The point is that when you have an endless loop in Python or Ruby (and I can imagine anywhere else) you cannot stop that process and it will run in the background hogging resources. This plugin looks for the process corresponding to the current file and kills it.

It requires the [*psutil*](https://github.com/giampaolo/psutil) python package as a dependency.

### DIFF
--- a/repository/k.json
+++ b/repository/k.json
@@ -132,6 +132,17 @@
 			]
 		},
 		{
+			"name": "Kill",
+			"details": "https://github.com/goldsborough/kill",
+			"labels": ["build system", "process", "build"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Kimbie Color Scheme",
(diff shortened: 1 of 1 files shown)


FichteFoll:
`psutil` is not available as a dependency: https://github.com/wbond/package_control_channel/blob/master/repository/dependencies.json

goldsborough:
And there's nothing to change that?

wbond:
@goldsborough You would need to package it (psutil) up as a PC dependency for all of the OS/architecture combinations you want to support. That is assuming it contains more than pure Python. In that case, you have to create a repo with all of the Python in a folder named `all`.

FichteFoll:
See https://github.com/packagecontrol/requests for reference.

goldsborough:
Ok thanks guys, I'll do that.

FichteFoll:
@goldsborough ping

PR #4824, opened 2015-09-21
Add ctranslator plugin

lovesnow:
Auto translate Chinese to English or English to Chinese.


FichteFoll:

In order to use dependencies (and not include the literal source code), follow the steps at https://github.com/huot25/StyledPopup#how-to-use-styledpopup-as-a-dependencey.

PR #4827, opened 2015-09-22
Update c.json

lovesnow:
add ctranslator tool


FichteFoll:
#4824 for reference

PR #4828, opened 2015-09-22
added last used tab

dvcrn:
added last used tab

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -159,6 +159,16 @@
 			]
 		},
 		{
+			"name": "LastUsedTab",
+			"details": "https://github.com/dvcrn/sublime-last-used-tab",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "LaTeX Accents",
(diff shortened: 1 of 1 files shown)


FichteFoll:
I assume the difference to the default ctrl+tab behavior is that this plugin also goes over tabs in different groups?

dvcrn:
The difference is, ctrl + tab cycles through the last visited buffers in stack. This command here will always select the previous tab. 

Assuming you are opening the following files: `a.py`, `b.py`, `c.py`, `d.py`. Executing this command multiple times will always result in this behavior: `d.py` -> `c.py` -> `d.py` -> `c.py` since the other tab will always be the tab you visited right before. 

To make this behavior work with the ctrl+tab stack command, you'd need 2 keybindings: One for cycle forward and then one to cycle backwards.

dvcrn:
Looks like I misunderstood the `next_view_in_stack` command by a good amount. A user on reddit explained me how to use it and where my mistake was so yeah, I guess it really is quite similar to this. 

I'm closing this PR unless I find some behavior where I really need this is instead.

PR #4831, opened 2015-09-23
submitting DupLines

thiagoh:


FichteFoll:
Your feature list is partly wrong.

>It extends and improves the duplicate_line sublime's plugin so you can also use ctrl+shift+d to duplicate the current line to below and even if you have some text selected the whole line will be copied down.

While it does mention having "selected text" will also duplicate the line, it removes the ability to duplicate certain text segments which the default implementation does. Well, it is built-in anyway but deserves to be mentioned imo.

>You can duplicate multiple lines at the same time.

The default implementation can do this too.

>You don't need to select the whole line in order to duplicate it. Any line with a selection will be duplicated.

same

>Just press the shortcut key ctrl+alt+up to duplicate the current line to above and ctrl+alt+down to duplicate it to below.

These are differently-purposed default bindings, please do not override them.

And finally, this is pretty much the same that https://packagecontrol.io/packages/Duplicate%20Lines does already.

thiagoh:
The eclipse, for example, has those both shortcuts for duplicating lines up and down and I imagine that there is a lot of people who uses eclipse and sublime as well and would really enjoy this feature being accessed in the same way. Don't you think that could be a good idea letting the user decide by himself whether use the default implementation or mine one?

About the selection issue is that if the user select any part of a text and uses the duplicate line's default implementation, only the selected text will be duplicated and with mine the whole line will be.  This way the user doesn't need to worry about deselecting any text before using the command.

FichteFoll:
> Don't you think that could be a good idea letting the user decide by himself whether use the default implementation or mine one?

I'm perfectly fine with that.

However, my main concerns were:

1. The description is (slightly) inaccurate, especially regarding the default implementation.
2. It overrides default bindings. Like the above, this is not something that will prevent you from getting the package listed, but a headsup from me because users may use the bindings you are overriding and might get annoyed at that. I, for example, use these bindings. This is why many packages do not define keybindings at all and instead just list example bindings in their readme. 
3. https://packagecontrol.io/packages/Duplicate%20Lines does this already. Like, exactly the same.

thiagoh:
I see, and I agree with you on these points. So what if I:

1. correct the description about the default implementation
2.1. remove the ctrl+shift+d binding and let the user use the default implementation with this binding. 
2.2. I've been trying to find the ctrl+alt+up/down sublime default's binding and I confess I didn't find it at all. 
3. The package Duplicate Lines you listed doesn't do the same. There is no duplicate line to above, the mouse cursor only goes down. With my plugin when you duplicate line to above, the mouse cursor goes along with the line.

FichteFoll:
2.2) In `Default (Windows).sublime-keymap`:

```js
 	{ "keys": ["ctrl+alt+up"], "command": "select_lines", "args": {"forward": false} },
	{ "keys": ["ctrl+alt+down"], "command": "select_lines", "args": {"forward": true} },
```

3) I honestly didn't test the other package, but the source indicates that it can insert the duplicated line before or after the "currently selected lines": https://github.com/wjthomas9/duplicate-lines/blob/master/duplicate_lines.py#L8 It's just not mentioned in the readme.

FichteFoll:
@thiagoh ping

thiagoh:
@FichteFoll  so, what if I remove/change the windows shortcuts so that you can approve this plugin?

FichteFoll:
I'm more concerned about the duplicated functionality, which you haven't commented on again yet. But that is a requirement as well.

thiagoh:
@FichteFoll so.. what do you suggest?

FichteFoll:
I suggest that you should contribute to @wjthomas9's package if you have any features to add and I will close this pull request.

thiagoh:
ok

wjthomas9:
FYI, this was added to my README a week or so ago. https://github.com/wjthomas9/duplicate-lines/pull/8

PR #4835, opened 2015-09-24
Compile Sass automatically without the need to install Ruby or LibSass.

jakerb:
Version 1.1.0 Sassy.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -129,6 +129,17 @@
 			]
 		},
 		{
+			"name": "Sassy",
+			"details": "https://github.com/jakerb/Sassy",
+			"labels": ["sass", "sassy"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
That's an interesting approach, but it is flawed: URLs have a maximum length, which is reached pretty easily when sending entire files over GET parameters. Use POST instead.

Furthermore:

1. You must remove `package-metadata.json`. That file is supposed to be generated by Package Control upon installation.
2. You have to push a valid [semantic version](http://semver.org) [tag](http://git-scm.com/book/en/v2/Git-Basics-Tagging) (with optional leading `v`) to your repo.
3. The tests don't pass: https://packagecontrol.io/docs/submitting_a_package#Step_7
4. A readme explaining what this package does (especially that it sends all text over the network) is appropriate.
5. I used http://178.62.31.223/s/jakebown/sassy/?sass=.hi%0A%20%20color:%20a to test it, but it doesn't display anything.

FichteFoll:
@jakerb ping

PR #4836, opened 2015-09-24
Added OpenFileInCurrentFolder package to channel

barneywilliams:


FichteFoll:
I think there was a package like that already, but I can't find it. Either it disappeared or it is so badly discoverable that it probably deserves to not be found.

Anyway, You'll need to change `"branch": "master"` to `"tags": true` and use [tags](http://git-scm.com/book/en/v2/Git-Basics-Tagging) to create releases and push valid [semantic version](http://semver.org) tags to your repo. As mentioned [in the docs](https://packagecontrol.io/docs/submitting_a_package#Step_4), we no longer accepts branch-based releases for new packages.

Also, since you alread pushed a tag, [semantic versions](http://semver.org) consist of exactly three numeric parts.

barneywilliams:
Ah ok, I will update the config and be sure to use semantic version release tags.

Yeah, I was surprised not to find another package like this when I took a gander...

Thanks,
Greg

barneywilliams:
I have switched to tags and re-release as 1.0.0 per semantic versioning requirements. Thanks!

PR #4842, opened 2015-09-25
rename sublime WinSCP to WinSCP and make compatible with all platforms

thecotne:


FichteFoll:

thecotne:
@FichteFoll can you please merge it? and is it better if i squash thous two commits?

FichteFoll:
It's okay, the commit history is fucked up anyway.

I don't get notifications from commits, which is why I didn't see your update earlier.

PR #4848, opened 2015-09-26
Update g.json

SumeetSinha:
Changed the name of plugin from gmESSI-gmsh tools  to gmESSI-gmsh Tools


FichteFoll:

PR #4849, opened 2015-09-26
Add facebook material theme

ridewn:


FichteFoll:
Please delete `package-metadata.json` from your repo. It is supposed to be generated by Package Control only.

Also, if you renamed your package files, most notably "Material.sublime-theme", you could probably install this alongside the original.

ridewn:
I deleted the `package-metadata.json` file, I tried renaming the theme so that it could be installed alongside of the original, but for some reason the icons wouldn't show up once i edited the paths in the theme file, there's probably more files that I need to change. From what I can tell by the people who are using it, they are using it as a complete replacement for the theme, but if it becomes an issue I can always look into it and make it work. Hope that's fine. Thanks!

PR #4852, opened 2015-09-28
add a ConEmu cmd package for windows user

kiliwalk:
launch [conemu](https://github.com/Maximus5/ConEmu) from the current dir or the project folder, only for sublime text 3 and windows!

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1093,6 +1093,18 @@
 			]
 		},
 		{
+			"name": "Cmd",
+			"details": "https://github.com/kiliwalk/sublime_cmd",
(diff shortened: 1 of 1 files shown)


FichteFoll:
1. Your repo should not contain ` package-metadata.json`, please remove it.
2. The name "cmd" is highly inappropriate since this is specifically targetting "conemu".
3. I recommend you to use `window.folders()[0]` instead of the project file name because that allows users to place their project files in a different directory than their actual project.

kiliwalk:
thanks for your advices, i wil change these

PR #4857, opened 2015-09-30
Add NoTutorial package

chadnewbry:
A small package I wrote that suggests better key bindings to use based on the current key bindings a person is using. 

* Currently only recommends using move line up /down instead of deleting, moving a line, and then pasting.

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -489,6 +489,17 @@
 			]
 		},
 		{
+			"name": "NoTutorial",
+			"details": "https://github.com/chadnewbry/NoTutorial",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"platforms": ["osx"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "NP-Complete",
 			"labels": ["completions"],
 			"details": "https://github.com/Nax/sublime-np-complete",


FichteFoll:
- [This](https://github.com/chadnewbry/NoTutorial/blob/fda20830e8084ce8029ef2f93a048a34adb24ce2/NoTutorialMappingController.py#L9-L10) is problematic because packages are installed into `Installed Packages` as a `.sublime-package` zip archive and thus the folder you're trying to access does not exist.

  Instead, use `sublime.load_resource("Packages/NoTutorial/tutorial_mappings.txt")`.

- I do not see a reason to limit this to osx only, besides the lack of key bindings for the other platforms.

- As a structure recommendation, I suggest placing everything that does not interact with the `sublime_plugin` interface in a subdirectory and importing it from the files that do, most notably `NoTutorial.py`. Example: https://github.com/FichteFoll/CSScheme/commit/010d9f1b8f77520d9dc1e0b71764b336d0e53b69

- You should add a readme.

chadnewbry:
Hi FichteFoll,

Thanks for your prompt reply and insight into how better to structure the package. I haven't written a sublime plugin before so that was much appreciated. 

Per your comments I did the following: use sublime.load_resource, moved all files that don't interface with the sublime_plugin API to subfolders, and added a README. I haven't added mappings for other platforms yet as currently I don't know if there is any interest and I'm on a Mac. I will if someone comments on the issue I created.

Let me know if I missed anything. Thanks again for your help :)

FichteFoll:
Note that you haven't pushed a new tag to the repository, so people will still install the very first version from 11 days ago.

PR #4862, opened 2015-10-02
Add StoryToJasmine package

gallegogt:
SublimeText plugin for BDD that retrieves a Story from Pivotaltracker
and inserts it to the current document as a jasmine spec.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -2467,7 +2467,7 @@
 			"releases": [
 			{
 				"sublime_text": "*",
-				"branch": "master"
+				"tags": true
 			}]
 		},
 		{
--- a/repository/s.json
+++ b/repository/s.json
@@ -2461,6 +2461,16 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


wbond:
You'll probably want to rename `plugin.py` in your root folder. The reason for this is that it is a generic name, and in Sublime Text 2, packages are loaded in such a way that this plugin will be added to `sys.modules` as just `plugin`. `story_to_jasmine.py` would be far less likely to conflict.

FichteFoll:
You didn't push a new tag for the latest changes yet, so they won't be installed.

Please also use a consistent indentation style.

FichteFoll:
@gallegogt ping

gallegogt:
@FichteFoll Hi, this plugin have some problem?

FichteFoll:
@gallegogt, no, not the package itself. Please address the inconsistent indentation style in this pull request.

PR #4863, opened 2015-10-03
Added React ES6 snippets

mboperator:
Same keybindings as Facebook's ReactJS Snippets, updated syntax.
I would submit a PR to them but they're sticking with ES5 syntax for maximum support.

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -350,6 +350,16 @@
 			]
 		},
 		{
+			"name": "React ES6 Snippets",
+			"details": "https://github.com/mboperator/sublime-react-es6",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "master"
+				}
+			]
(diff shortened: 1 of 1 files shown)


wbond:
So this is an ES6 alternative to https://packagecontrol.io/packages/ReactJS%20Snippets?

wbond:
It looks like you should clean up the readme and messages since this is a fork. Getting rid of references to the syntax definition would be good. Updating the readme with the new name and explaining that it is an ES6-fork of the original would be good also.

Are the animated gifs accurate any longer with the new snippets?

mboperator:
It's actually a fork of Facebook's https://github.com/reactjs/sublime-react (just titled "ReactJS on Package Control)

I made some changes to the readme and updated both of the gifs. Let me know if you see anything else that should be changed.

PR #4864, opened 2015-10-03
Update Hide Menu package

p3lim:


wbond:
This seems to be along the same lines as https://packagecontrol.io/packages/Hide%20Menu, however your version seems more robust since it has an event listener.

Perhaps contact @pcorey and see if you guys can merge these two very similar packages into one super-package? :-)

p3lim:
His seems only to toggle the "Hide Menu" option on any startup, which would show the menu if it was hidden previously, and the menu would be shown on every new window as well.

The option that already exists in ST _works_, the part that doesn't work is making the setting "sticky" (in practice) across windows, which is what mine does.

There is nothing to merge in my opinion.

wbond:
@p3lim Right, I realize that yours seems to more likely be what people are looking for (from my understanding). What I am trying to accomplish is have a package for this task that works, rather than one that doesn't and one that does, and have users try to install both to figure it out.

Maybe all of your code could replace what is currently in "Hide Menu" as a new version with a release note. I don't know, I just figured it would be good to try and have a conversation with @pcorey to see if some sort of collaboration or merging or name reused makes sense.

If neither of you are interested, or they solve subtly different problems, I will certainly add this in. I'm just trying where I can to help improve the overall package experience for Sublime Text users.

p3lim:
I wouldn't mind taking over the name, but he'd have to agree to that :)

And I'm not trying to be difficult here, maybe his package worked properly at one point, maybe it serves a purpose in ST2 (I wouldn't know), but in the current version of ST3 it seems to do nothing but to add an annoyance.

wbond:
Understandable. Let's give @pcorey a day or two to see if he gets any notifications about this. If he does and is interested in such a route, great. If not, no hard feelings and I think it would make sense to merge this in.

p3lim:
Agreed.

pcorey:
My package is outdated, broken and neglected. Feel free to take over the name!

p3lim:
Thanks @pcorey.

@wbond I'll take the name, want me to rebase this to rename and also remove the old plugin?

wbond:
Yeah, so you'll need to update the PR to replace the existing entry with your repo's details. You'd probably also want to add a message for your initial version letting any existing users know the change.

p3lim:
Updated the PR and added upgrade message in https://github.com/p3lim/sublime-hide-menu/commit/3cbdeeba4a36ac201660b34391d3690d8b386ae5 as suggested.

PR #4866, opened 2015-10-03
add CHeaders

leoxnidas:


wbond:
A few items I noticed:

 1. You shouldn't need a `.no-sublime-package` file since you are not including shared libraries or executables in the package
 2. We no longer support new packages using `branch`-based releases. Instead you must use tags that are semantic versions and set `"tags": true` in place of `"branch": "master"`.
 3. Is there a technical limitation to not supporting OS X?
 4. Your readme currently does not make it clear exactly what is does, or how to use it. I can kind of guess based on the name and first line of readme, but it would be good to try and give some examples and what commands the package provides.

leoxnidas:
there's No support for IOS X , because I have no way to test CHeaders in a IOS X machine.
ADD CHeaders... try again...

wbond:
You'll have to address the other issues before I can merge this.

leoxnidas:
ok..=p

leoxnidas:
ADD CHeaders

FichteFoll:

leoxnidas:
thanks @FichteFoll

leoxnidas:
add CHeaders ..

PR #4868, opened 2015-10-04
Wanderlei Color Scheme

JeremySaks:
A harmonious dark-themed color scheme for Sublime Text 3.

### DIFF
--- a/repository/w.json
+++ b/repository/w.json
@@ -35,6 +35,16 @@
 			]
 		},
 		{
+			"name": "Wanderlei Color Scheme",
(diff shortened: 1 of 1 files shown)


wbond:
Is this tested in Sublime Text 2? I just noticed your readme and description mention ST3, but the `sublime_text` key in the pull request indicates is works for ST2 and ST3. We should just make those consistent, one way or the other.

JeremySaks:
Yes, it supports both 2 and 3. I've updated the README and description to reflect support for both. Default Channel tests pass for both 2 and 3.

PR #4870, opened 2015-10-05
add phpDocLookup package

Miw0:
Similar package like https://github.com/Miw0/jQueryDocs just for PHP.

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -875,6 +875,18 @@
 			]
 		},
 		{
+			"name": "phpDocs",
+			"details": "https://github.com/Miw0/phpDocs",
+			"author": "Miw0",
+			"labels": ["php", "docs"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "phpDocumentor",
 			"details": "https://github.com/benmatselby/sublime-phpdocumentor",
 			"previous_names": ["DocBlox"],
--- a/repository/p.json
+++ b/repository/p.json
@@ -875,8 +875,8 @@
 			]
 		},
 		{
-			"name": "phpDocs",
-			"details": "https://github.com/Miw0/phpDocs",
+			"name": "phpDocLookup",
+			"details": "https://github.com/Miw0/phpDocLookup",
 			"author": "Miw0",
 			"labels": ["php", "docs"],
 			"releases": [


wbond:
Since there is already a package named `PhpDoc`, and this differs only by adding a `s` at the end, what do you think about tweaking the name?

Perhaps something like `phpDocLookup`?

Miw0:
Well I knew you would say this .. Honestly I would like to keep this name because this package follows the same naming pattern like my other package and both do virtually the same thing except for different languages. I also thought about publishing more packages for more languages which then will follow the same naming convention so the user can identify what each package will do just by reading the name.

I also see that the names differ too less which can confuse people or lead to install the wrong package. Honestly I feel a bit stuck here :D

wbond:
I understand why you want to name it that way, it just seems the benefits of changing the name outweigh the consistency.

Miw0:
Hm well you're right. I'm gonna rename it.

FichteFoll:
Also, why is this package needed? There are a bunch of "look this up on *" or "look this up on php.net" packages already. What makes your package different?

Miw0:
@FichteFoll which packages? Can you give some examples?

My package works as follows: You select a php function and press a key combination and your default browser will pop up loading php.net/FUNCTION which is way more informative than showing information about the function in Sublime Text and way more up-to-date. 

Originally I created jQueryDocs which works the same just for jQuery and api.jquery.net because I found myself often loading the online docs and searching for explanations about some functions and this made my workflow way more fast. As time went by I found myself writing more and more php where I had the same problem: Searching on php.net for explanations or just to look something up so I wanted the same functionality for php as I had for jQuery. :blush:

I think this package will help lots of users who are at the beginning of writing php and need to look up things way often because it gives you a fast and easy way to get the information you are looking for on the website which was made for that.

FichteFoll:
Most notably https://packagecontrol.io/packages/GotoDocumentation, although https://packagecontrol.io/packages/FindDoc could also easily made to work with proper keybinds. I probably rejected the other proposals because this type of package is rather common, especially for many different languages since many think each language needs its own package.

Miw0:
Well, I didn't know one of these packages exist and I understand your point.
In the end it is not my decision if you merge this package or not. I have nothing more to add and I think I don't need to defend my package by all means. :)

FichteFoll:
Alright, in that sense I'm closing this.

PR #4871, opened 2015-10-05
Added SCA syntax and settings package

osecluna:
This is a package to correct syntax highlighting for JavaScript templating used on Netsuite's SCA platform

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -158,16 +158,6 @@
 			]
 		},
 		{
-			"name": "SuiteCommerce Advanced syntax and settings",
-			"details": "https://github.com/venda/SCA-SublimeText/",
-			"releases": [
-				{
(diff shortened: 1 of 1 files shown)


wbond:
Perhaps add a little description (and maybe a link) to the readme of what SCA is, in case someone it looking for a different project that uses the same acryonym? (I don't know if that is an issue or not, it just wasn't clear to me looking at the readme and googling it). I realize this pull request mentions Netsuite, it is just that the readme does not.

Also, your tags need to be semantic versions, which means they need to have three numeric parts, e.g. `1.0.0` instead of just `1.0`.

osecluna:
Hi,
Thanks for your feedback. I have made the package name and the README more explicit and also changed the tag to 1.0.0

PR #4872, opened 2015-10-06
Updated x86_64 assembly package info

13xforever:
Seeing as it was already submitted and accepted, let's try to do it right

PS: it is not clear if a simple case adjustments in the package name should be treated as a rename

### DIFF
--- a/repository/x.json
+++ b/repository/x.json
@@ -23,13 +23,13 @@
 			]
 		},
 		{
-			"name": "X86 and X86_64 assembly",
+			"name": "x86 and x86_64 Assembly",
 			"details": "https://github.com/13xforever/x86-assembly-textmate-bundle",
-			"labels": ["syntax", "at&t", "gas"],
+			"labels": ["syntax"],
 			"releases": [
 				{
 					"sublime_text": "*",
-					"branch": "master"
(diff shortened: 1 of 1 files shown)


wbond:
You are correct that it is not extremely clear about case changes and renames. In this case I can tell you that in order to actually change the name for existing users and have them get new releases, the old capitalization will need to be added in a `previous_names` list for the package.

13xforever:
Alright, I've updated the package info with the old name

13xforever:
Also, while we're on the topic of docs, it would be nice to have an explicit info on what tags are recognized as a package version (i.e., only "1.2.3" or "v1.2.3" works as well, etc).

wbond:
For tag versions, see https://packagecontrol.io/docs/submitting_a_package#Step_4. Specifically, any tag that is a [semantic version number](http://semver.org/) will work.

13xforever:
Yes, I know follow the semver in my own projects, but I prefer to do tags with prefix instead of a bare number. Prefixes are not a part of semver specification per se.

wbond:
https://github.com/wbond/package_control/blob/master/example-repository.json#L264-L288

PR #4873, opened 2015-10-06
add AutoSaveBuildLatex

jgonzalezr:
Requires latextools package

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -1549,6 +1549,16 @@
 			]
 		},
 		{
+			"name": "AutoSaveBuildLatex",
+			"details": "https://github.com/jgonzalezr/AutoSaveBuildLatex",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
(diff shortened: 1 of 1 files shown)


wbond:
Your readme and install messages state that ctrl+shift+s should be used to enable auto build, however you also say to use that key combo for auto save. From reading further on, I think you intended to have users bind auto build to ctrl+shift+b.

jamesfzhang:
This is a fork of my plugin [auto-save](https://github.com/jamesfzhang/auto-save) with some modified code. Can you explain what you're trying to do @jgonzalezr?

jgonzalezr:
I tried to define an auto-save and auto-build plugin only for .tex files, using ctrl+s and ctrl+b respectively. I've updated the README file. 
Use auto-build helps to review the output file while you are typing with Latex.

FichteFoll:
You can just combine auto-save and https://packagecontrol.io/packages/SublimeOnSaveBuild instead, which is also more flexible.

PR #4874, opened 2015-10-07
add file header creator

matthewgao:
Automatically add file header while create a new file or save a empty file

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -321,6 +321,17 @@
 			]
 		},
 		{
+			"name": "File Header Creator",
+			"details": "https://github.com/matthewgao/FileHeaderCreator",
+			"labels": ["sublime-enhanced","filename"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "master"
+				}
+			]
+		},
+		{
 			"name": "File Rename",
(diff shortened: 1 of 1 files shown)


wbond:
A few notes:

 1. Sublime Text will only load python files in the root of a package directory
 2. We no longer accept branch-based releases for packages. You'll have to create a git tag for each release (the name should be a semantic version number, e.g. `1.0.0`) and change `"branch": "master"` to `"tags": true`.
 3. The `sublime-enhanced` label is really a construct created by @shagabutdinov to tie the collection of his packages together. I don't think it makes sense to add here.

However, in the grand scheme of things, it almost seems you are re-creating https://packagecontrol.io/packages/FileHeader. How is this different?

FichteFoll:
Also, what are notable differences to the other "header" or "template" packages.

FichteFoll:
@matthewgao ping

matthewgao:
Sorry, I missed this notification.

Yes, I see there has already had a FileHeader add-on, I didn't notice that before I implement mine. And mine makes no difference with FileHeader, so just ignore me. :)

FichteFoll:
Okay.

PR #4875, opened 2015-10-07
Add tags property to package

wjthomas9:


wbond:
It looks like there may have been a merge or change that caused this change to have a conflict. You'll need to rebase upon the master branch here.

Also, you'll want to delete the `details` key from your `releases` now that you have `tags`. Additionally, your tags will have to be semantic version numbers to be picked up by Package Control. So instead of `1.0` it should be `1.0.0`. http://semver.org/ has details.

wjthomas9:
I think we're good now!

wbond:
Somehow a commit from another user got pulled into your copy.

I'll commit your branch -> tags change on master and close this.

wbond:
ddf3c83092f808235bbe1574c40085c4a4b10e54

PR #4877, opened 2015-10-07
added SnippetX to repository

ColinRyan:


wbond:
Since your only current tag is `0.1.0-alpha`, which is a pre-release version, only uses to explicitly add the SnippetX package to their list of pre-release packages will be able to install this.

Instead, I would recommend just a tag like `0.1.0`.

ColinRyan:
Thanks, I've made that adjustment. Do you think there's any value setting this as a pre-release? Asking because I'm not up on the standards or etiquette for sublime package releases, this being my first.

wbond:
I think pre-release stuff in the Sublime world is mostly useful for popular packages where you want to roll out changes to people who are willing to be beta testers. For new packages, or those with few users, it is probably not all that useful.

wbond:
I should also say, the GitHub "pre-release" checkbox doesn't actually affect Package Control in any way. The only way PC considers a package pre-release is if it uses a semver `-` suffix. For instance, `1.0.0-beta` is prerelease since it has `-beta` after the `1.0.0`. If you add a `+` suffix (or no suffix), that is not considered pre-release.

ColinRyan:
Ok, thanks you! Considering that, I've altered the package so that it's not pre-release. I appropriate your help! :)

ColinRyan:
ahhhh ok. Good to know.

PR #4878, opened 2015-10-08
add my plugin difinitions.

ekazyam:
It'll add a separator file to open file list .

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -191,16 +191,6 @@
 			]
 		},
 		{
-			"name": "SidebarSeparator",
-			"details": "https://github.com/ekazyam/SidebarSeparator",
-			"releases": [
-				{
-					"sublime_text": ">=3000",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "Advanced Aria Templates snippets pack",
 			"details": "https://github.com/dpreussner/advanced-at-snippets-pack",
 			"labels": ["snippets"],
--- a/repository/a.json
+++ b/repository/a.json
@@ -191,8 +191,8 @@
 			]
 		},
 		{
-			"name": "AddSeparate",
-			"details": "https://github.com/ekazyam/AddSeparate",
+			"name": "SidebarSeparator",
+			"details": "https://github.com/ekazyam/SidebarSeparator",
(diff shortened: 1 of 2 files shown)


wbond:
I think a snippet would be more appropriate for this instead of a whole python-based package.

ekazyam:
Thanks. I'm going to try it.

ekazyam:
Which part should i cut for snippet ?

ekazyam:
Thanks. I'm going to try it.
Which part should i cut for snippet ?

ekazyam:
Thanks. I'm going to try it.
In Python rather , do you think that there is a good plug -ins using a snippet?

ekazyam:
If you use a snippet , you can not use a temporary file using the scratch buffer . I thought you do not want to save when you close the source code .

FichteFoll:
I do not understand the purpose of this package. It creates a new file, makes it a scratch view and then adds some separator line to it. What for?

ekazyam:
I will open a lot of files while working . Therefore , in the case of the tab display , management is troublesome . Therefore , I have to hide the tab . List of files I'm sure in the OPEN FILES of sidebar . At that time , since the request had colleagues that want a separator to identify the files for each application , it has been created . However, in management by the tab display , this plug -in is meaningless .

FichteFoll:
Ah, I understand. In that case I would rather use a name like "SeparateOpenFiles" or "SidebarSeparator" and use `view.set_name` directly instead of adding a new line. Also, I would advise to additionally set `view.set_read_only(True)`.

FichteFoll:
Oh, and you must not provide `package-metadata.json` in your repository. Please remove it.

ekazyam:
Thanks.
I will do the modification .

ekazyam:
Modification has been completed . Because you have changed the package name , my repository also has been changed . Whether it is better that sent again pull request ?

ekazyam:
Modification has been completed . Because you have changed the package name , my repository also has been changed . Whether it is better that sent again pull request ?

ekazyam:
Modification has been completed . Because you have changed the package name , my repository also has been changed .

PR #4881, opened 2015-10-09
PyQt Autocomplete

tushortz:
Autocomplete feature for PyQt with Sublime text

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -1840,6 +1840,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


wbond:
It looks like you need to rebase your commit on top of master.

tushortz:
I don't think this is a duplicate @FichteFoll because I closed the old pull request and the new pull request has two packages added instead with name changes from previous packages.

FichteFoll:
This is how I label them.

PR #4882, opened 2015-10-09
Add new 'PHP comment Wrap' package

canufeelit:


FichteFoll:
Your `.no-sublime-package` file is not needed, please remove it.

Also, you can directly provide arguments to the `insert_snippet` command without going through a snippet file.

canufeelit:
Hi @FichteFoll,

First, thanks for the tips.

I've fixed the versions and removed the `.no-sublime-package` file.
Also, I've ended up rewriting the plug-in in python, so it will be easy adding more options this way.

Looking forward to your comment on this.

Regards

FichteFoll:
Your currently pushed tag is still named `v1.0`, which is not a semantic version since those exist of exatly 3 numeric parts.

In your `messages.json` you define a key for `v1.0.0`, however Package Control will look for the unprefixed `1.0.0` version here. It won't matter because nobody will be able to upgrade to 1.0.0 anyway (it's the earliest version), but for future versions it might.

canufeelit:
It should be fixed now. Corrected the `messages.json` and pushed the `1.0.0` tag.

PR #4885, opened 2015-10-10
add package: Javascript Snippets for IMWeb

longyiyiyu:
add the "Javascript Snippets for IMWeb" package to the default channel

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -328,7 +328,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"branch": "master"
+					"tags": true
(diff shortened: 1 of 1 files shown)


FichteFoll:
Uhm, idk about this. It only contains two one-line snippets?

longyiyiyu:
These is the initialization of the pkg and that two snippets are just for test. We will collect our team members' snippets and make a standard list of these.

FichteFoll:
I will never understand why people submit their heavily WIP packages before actually adding any substantial functionality ...

PR #4887, opened 2015-10-11
Added PyQt and  Qt autocomplete ...

tushortz:
* Added PyQt autocomplete
* Added Qt autocomplete
* Renamed Wildlife 
* Added label for JavaIME

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -193,7 +193,7 @@
 		{
 			"name": "JavaIME",
 			"details": "https://github.com/tushortz/JavaIME",
-			"labels": ["java","snippets", "auto-complete", "completion", "autocomplete", "snippet", "completion"],
+			"labels": ["java","completions", "auto-complete", "completion", "autocomplete", "snippet", "Completion"],
 			"releases": [
 				{
 					"sublime_text": "*",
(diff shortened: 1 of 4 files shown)


tushortz:
I realised that a pull request of mine was marked as duplicate even after I had closed it earlier before opening a new one that includes two new packages

FichteFoll:
You should definitely take a look at completions files instead of providing more than 10000 snippets in a package, each in their own file: http://docs.sublimetext.info/en/latest/reference/completions.html

tushortz:
Thanks, I didn't know something like this existed. I'll definately have a
look

FichteFoll:
@tushortz, since you replied by mail, I edited my above comment shortly after submission to include the following:

>Furthermore, in the conventional sense "autocomplete" refers to the static or dynamic analysis of source code and only suggestion completions for the available variables and their available attributes according to their type. Your packages are about the traditional "completion" instead.

tushortz:
Thanks again. And from what you mentioned now, is it possible to do that on
sublime text as well.

tushortz:
I'll close this pull request and create a new one for the updated packages

PR #4891, opened 2015-10-13
add 'eRube XML newline remover package'

calebfoster:


FichteFoll:
What is this package for? It seems to remove a few weird newline constructs and calls the reindent command, but that's it.

FichteFoll:
@calebfoster ping

PR #4893, opened 2015-10-14
gherkin_auto_complete_plus

austincrft:
Added [gherkin_auto_complete_plus](https://github.com/austincrft/sublime-gherkin-auto-complete-plus) package

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -229,6 +229,17 @@
 			]
 		},
 		{
+			"name": "gherkin_auto_complete_plus",
+			"details": "https://github.com/austincrft/sublime-gherkin-auto-complete-plus",
+			"labels": ["auto-complete", "language syntax", "testing", "gherkin", "cucumber"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"details": "https://github.com/hypernovagama/Gherkin-highlight-sublimetext",
 			"labels": ["language syntax"],
 			"releases": [
--- a/repository/g.json
+++ b/repository/g.json
@@ -229,23 +229,23 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
The usual naming convention is to use PascalCase with optional whitespaces between words. Is there a specific reason why you chose to not follow that?

austincrft:
In the [Submitting a Package](https://packagecontrol.io/docs/submitting_a_package) doc, it says "Use CamelCase or underscore_notation", so I assumed it was okay to use it. Also, it makes it easier to use [import statement in gherkin_event_listener.py](https://github.com/austincrft/sublime-gherkin-auto-complete-plus/blob/master/gherkin_event_listener.py#L4). Is there an easier way to do this? I'm okay with changing the name.

FichteFoll:
>In the Submitting a Package doc, it says "Use CamelCase or underscore_notation", so I assumed it was okay to use it.

I see, I never noticed.

>Is there an easier way to do this?

Yes, just use a relative import.
These kind of imports only work on ST3 btw (on ST2 they need to be absolute but without the package name, oddly). Either you need to fix the imports or change the ST compatability.

austincrft:
Ah, I didn't realize those imports would only work in ST3. I'll fix the import and then rename, because it seems underscore notation is an unpopular choice. Thanks!

austincrft:
Okay, @FichteFoll. I think this should be satisfactory.

FichteFoll:
You specify `"tags": "st3-"`, but there are actually no tags with this prefix on your repo. You want to set this to `"tags": true`.

austincrft:
Okay, apparently I misunderstood their use in the docs. Got that fixed.

FichteFoll:
Here's what you can do with them: https://raw.githubusercontent.com/FichteFoll/sublime_packages/master/package_control.json

PR #4895, opened 2015-10-16
Add Sequoia Studios package channel

alexggordon:
This adds the sequoia studios sublime package channel for DB1.

### DIFF
--- a/channel.json
+++ b/channel.json
@@ -129,6 +129,7 @@
 		"https://raw.githubusercontent.com/yangsu/sublime-octopress/master/packages.json",
 		"https://raw.githubusercontent.com/zfkun/sublime-kissy-snippets/master/packages.json",
 		"https://release.sublimegit.net/packages.json",
+		"https://sequoiastudios.io/sublime_packages/packages.json",
 		"https://sublimerge-bfcloud.rhcloud.com/packages.json",
 		"https://wuub.net/packages.json"
 	]


FichteFoll:
Please do not use "packages_cache" and instead provide the information in "packages".

I should probably add tests for that, don't you think, @wbond?

Furthermore, if you structure the json file as intended and run the tests using the ChannelRepositoryTools package, you will notice that your "date" timestamps are badly formatted.

alexggordon:
Heh, I actually did run the [tests](https://gist.github.com/alexggordon/49fc3ec9eddcbd6dde70), and no errors arose. In addition to that, I almost exactly followed the format [demonstrated](https://github.com/wbond/package_control/blob/master/example-channel.json) (notice the ```"packages_cache"``` json key there also) for schema version 3.0. 

However, after looking through the channels, I found [this example](https://csch1.triangulum.uberspace.de/release/packages.json) package channel that appears to be the model I should be following. I've updated my json to follow that example, which then caused the tests to fail. 

Either way though, I've corrected the date formatting!

FichteFoll:
The problem here is that you are not providing a channel but a repository file, which does not have this key. (I know, the differentiation is counter intuitive.)

alexggordon:
Gotchya. I understand now. The json file I'm submitting is the repository file, and I'm adding it to the channel. 

File has been updated!

PR #4904, opened 2015-10-17
jQuery mobile demos

coderyang:


FichteFoll:
I think you are using `gen_commands.py` locally to generate `Default.sublime-commands` since it currently would be automatically loaded by ST and not work.

coderyang:
Oh, This is my first time to write plug-ins for sublime-text, and I am not familiar with the workflow of publishing a package. I think I have fixed the problem. Thanks for your reminder.

FichteFoll:
Don't forget to push a new tag after the changes.

coderyang:
Ok!

PR #4905, opened 2015-10-17
added PyQt5 and Qt completions

tushortz:
* Added PyQt5 completions
* Added Qt Completions
* Renamed Wildlife


FichteFoll:

PR #4909, opened 2015-10-20
Added "Bash from Context Menu"'s entry

sathvikp:
This is a simple package to open a bash terminal in the current file's directory from the Sublime Text Context Menu.

Requires you to have `bash` in your system path.
Windows users can get it through MINGW

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -161,6 +161,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
Differences to https://packagecontrol.io/packages/Terminal ?

FichteFoll:
@sathvikp ping

PR #4911, opened 2015-10-20
Add Jam Language package

BenjaminSchaaf:
(disclosure: I'm also the author of the language itself)

Just created a simple syntax highlighter and build system for [Jam](http://lets-jam.org)
Although the language is still in its infancy, it is nice to have syntax highlighting. So why not add it to the package manager while I'm at it?

Links (for convenience)
- The [repo](https://github.com/BenjaminSchaaf/jam-sublime-syntax)
- The current latest [tag](https://github.com/BenjaminSchaaf/jam-sublime-syntax/releases/tag/1.0.0-a)

Should I also add any labels as well? (like `language syntax`)

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -80,6 +80,7 @@
 		{
 			"name": "Jam Language",
 			"details": "https://github.com/BenjaminSchaaf/jam-sublime-syntax",
+			"labels": ["jam", "language syntax", "build"],
 			"releases": [
 				{
 					"sublime_text": "*",
--- a/repository/j.json
+++ b/repository/j.json
@@ -78,6 +78,16 @@
 			]
 		},
 		{
+			"name": "Jam Language",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Yes, feel free to add labels.

Also, note that by default pre-releases won't be installable since Package Control only installs stable versions unless the user specifically adds your package to the install_prereleases list.

BenjaminSchaaf:
Thanks, I added some labels. Although the tag version is `alpha`, I haven't marked it as pre-release.

FichteFoll:
Prerelease status is only taken directly from the tag name and not from what you configured fit the release on github. The presence of `-a` is all it takes to define a prerelease version.

BenjaminSchaaf:
Ah ok. I've removed the `-a` from the tag.

PR #4916, opened 2015-10-22
Added Churchkey to c.json

mrjax:
Churchkey is a local file opener with configurable search directories.  Additional features include better selection parsing and URL domain options.  Planned features: configurable program selection for opening, more complex selection parsing and file finding, default behavior configuring.  Please enjoy

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -760,6 +760,18 @@
 			]
 		},
 		{
+			"name": "Churchkey",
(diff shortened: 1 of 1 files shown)


FichteFoll:
https://packagecontrol.io/packages/Open-Include does this already, and there is also https://packagecontrol.io/packages/Open%20URL which you mentioned in your readme already.

PR #4917, opened 2015-10-22
add underscorejs completions

coderyang:


FichteFoll:
How is this different from the [other packages](https://packagecontrol.io/search/underscore), which errorneously chose to implement this in the form of snippets?

FichteFoll:
@coderyang ping

PR #4919, opened 2015-10-23
Add CopyWithLineNumbersReloaded package

crash5:


FichteFoll:
Well, this is pretty much exactly https://github.com/curiousstranger/copy_with_line_numbers.sublime-package with a different name and a few more options to invoke the command. But that's not on Package Control. And the license situation is pretty weird since you did include it but it didn't contain a copyright holder in the first place.

crash5:
These three project almost the same with some modifications, every one with GPL >=2 and only the theskyliner version has copyright header.
This version based on the freeella version, with the modiciations from curiousstranger. So I get the license text from freeella (whic is the same in curiousstranger).

 - https://github.com/theskyliner/CopyFilepathWithLineNumbers/blob/master/readme.md
 - https://github.com/freeella/copy_with_line_numbers.sublime-package/blob/master/copy_with_line_numbers/Readme.txt
 - https://github.com/curiousstranger/copy_with_line_numbers.sublime-package/blob/master/copy_with_line_numbers/Readme.txt

Can you help me what can I write to the copyright holder?

FichteFoll:
I would probably just list the people whose code I used, and myself. This is just a really weird situation.

PR #4926, opened 2015-10-24
Octopress Snippets

SunDi3yansyah:
Add octopress-snippets

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -365,7 +365,7 @@
 			"releases": [
 				{
 					"sublime_text": ">=3000",
-					"platforms": ["windows"],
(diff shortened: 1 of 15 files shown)


FichteFoll:
You'll need to change `"branch": "master"` to `"tags": true` and use [tags](http://git-scm.com/book/en/v2/Git-Basics-Tagging) to create releases.

Either way, there is already a package with this exact name on Package Control, but it's currently flagged for ST2 only errorneously (snippets are always compatible): https://github.com/yangsu/sublime-octopress.

cc @yangsu

SunDi3yansyah:
Okay thanks, I can revision

PR #4927, opened 2015-10-25
add package w/ strict tmLanguage for hosts files

tijn:
I generated some very long regular expressions with a Ruby script (included in the repository) to attain really precise syntax highlighting for hosts files.

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -664,6 +664,8 @@
 		{
 			"name": "Hosts",
 			"details": "https://github.com/tijn/hosts.tmLanguage",
+			"labels": ["language syntax"],
+			"previous_names": ["hosts"],
 			"releases": [
 				{
 					"sublime_text": "*",
--- a/repository/h.json
+++ b/repository/h.json
@@ -671,6 +671,16 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
This has a shared scope with @rodrigoflores' https://github.com/rodrigoflores/hosts package. Since  your package has more precise patterns, I suggest to replace the existing hosts package. @rodrigoflores, are you with me on this?

rodrigoflores:
@FichteFoll :+1:, please feel free to change, I'm not maintaining this package anymore. 

Sorry for taking two weeks to respond.

tijn:
@FichteFoll the previous entry did not contain a "name" entry but I added one. I hope it can be a seamless upgrade for everyone but I really don't know how Sublime handles this, do you?

Anyhow, I added a mesage, and if you agree with it, it can be merged.

FichteFoll:
Package names are case-sensitive in most situations (due to case-sensitivity on file systems, but not all), so you need to add the lower-case version to a `previous_names` array if you want to switch to the "Hosts" version.

>I hope it can be a seamless upgrade for everyone but I really don't know how Sublime handles this, do you?

Yes, once you did the above, the upgrade should be seamless.

PR #4934, opened 2015-10-28
adding ChapelLang

acrosby:
Please let me know if this wasn't done properly or if I need to fix something. Cheers!

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2464,6 +2464,16 @@
 			]
 		},
 		{
+			"name": "ChapelLang",
+			"details": "https://github.com/acrosby/sublimetext-chapel",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
The tests don't pass: https://packagecontrol.io/docs/submitting_a_package#Step_7

acrosby:
Perhaps you can explain why. The documentation isn't particularly helpful. Testing the local channel passes, but I have no idea how that might relate to the default channel. Thanks!

FichteFoll:
The package is not in lexicographical order. It should be here: https://travis-ci.org/wbond/package_control_channel/builds/87895411#L154

Test failures explain why they failed. If you run them properly, they should also fail for you locally since it's the same code being run.

acrosby:
Oh I see I pushed a change before i sorted , sorry for my confusion

FichteFoll:
Oh yeah, and uncommited/unpushed changes is also a possibility.

PR #4937, opened 2015-10-29
Add Maude Syntax Highlighting package

jpsikorra:


FichteFoll:

While you're at it, you could also remove the `homepage` key, which is redundant.

jpsikorra:
Done @FichteFoll

PR #4939, opened 2015-10-29
Add Makefile Plus to repository list

Altomare:


FichteFoll:
What are the differences to https://packagecontrol.io/packages/Makefile%20Improved?

Altomare:
The main difference is that it's somewhat more complete as a coloration.
 - Indented blocks
 - Function calls not being interpreted as targets (ex: "$(info OS: $OS)" breaks Makefile Improved)
 - Proper target coloration
 On Makefile Improved, any indented definition isn't colored, as well as some keywords.

Altomare:
makefile plus
makefile improved

FichteFoll:
Since the author of Makefile Improved hasn't been active at all lately (about a year or so), I doubt you'd be able to collaborate with him on this.

So instead, I'll merge this as an alternative.

PR #4942, opened 2015-10-30
Adding Better Bookmarks to package repository.

dusk125:
Better Bookmarks integrates Sublimes bookmarks with the ability to have layers, save bookmarks and open then later.


FichteFoll:
Your latest release states "unix compatability", but here you still restrict the package to Windows only. I suggest opening another PR with the platform restriction removed.

PR #4948, opened 2015-11-01
Adding Grepmark to Package Control Repository.

dusk125:


FichteFoll:
I'm very positive that `sublime.error_message` is not available at plugin load time immediately after startup, meaning your error message will not be shown most of the time, if at all.

PR #4951, opened 2015-11-02
Update e.json

julux:
Just add EasyCatalog syntaxes and funtions


julux:
why is not update?
Please help me :-)

PR #4955, opened 2015-11-03
Added TwelfTools Package

nfeltman:
Adding a new package called TwelfTools, which is intended to support the Twelf theorem prover.  

Some quick searches suggest that there is no other sublime package intended to support Twelf.

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -2222,6 +2222,16 @@
 			]
 		},
 		{
+			"name": "TwelfTools",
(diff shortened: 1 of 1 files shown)


FichteFoll:

FichteFoll:
@nfeltman ping

nfeltman:
Oops.  Apparently I didn't push the tag, but that's done now.  Is that sufficient?

Thanks for your patience.

PR #4961, opened 2015-11-04
Added CodeCells

jesseengel:
CodeCells is a simple package I've been using for a while to facilitate copying and pasting chunks of code from a python script into a REPL like IPython. This allows you the easy of working with cells of code that you might get from an IPython Notebook, but with the workflow of having separate panels for your code and it's output. The use of cells is taken from the Spyder IDE `#%%`, which in turn took it from Matlab `%%`. It could be extended to other languages, but I thought it might be useful to others in it's current form.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1279,6 +1279,16 @@
 			]
 		},
 		{
+			"name": "CodeCells",
+			"details": "https://github.com/jesseengel/CodeCells",
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


FichteFoll:

You are also free to remove `CodeCells.sublime-package` from your repo since it serves no purpose.

jesseengel:
Great, I've removed the sublime-package and added a version tag. Thanks for the pointer :)

PR #4964, opened 2015-11-06
Added CloseFolder

aviaryan:
Sublime Text plugin to close all open files of current directory or all open files inside some directory.
https://github.com/aviaryan/CloseFolder


FichteFoll:

PR #4967, opened 2015-11-08
Added Hugofy

akmittal:
Hugofy is a package to use Hugo(Static site generator) commands from Sublime text.

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -1046,6 +1046,16 @@
 			]
 		},
 		{
+			"name": "Hugofy",
+			"details": "https://github.com/akmittal/Hugofy",
+			"releases": [
+				{
+					"sublime_text": ">3000",
+					"tags": true
+				}
+			]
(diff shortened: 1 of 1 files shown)


FichteFoll:
Few issues:

- Check out the `os.path` module, especially `os.path.sep` and `os.path.join`.
- Your get theme(s) commands are problematic because they seem to rely on the current working directory (`subprocess.Popen("git clone --recursive https://github.com/spf13/hugoThemes.git themes",...)`). You should specify an explicit path there.
- You have to push a valid [semantic version](http://semver.org) [tag](http://git-scm.com/book/en/v2/Git-Basics-Tagging) (with optional leading `v`) to your repo.

akmittal:
Hey,
* changed all the path generation using ```os.path.join```
* Specified the path explicitly to download themes.
* Added semantic version tag.

PR #4975, opened 2015-11-10
Added yaml2json package

maral:
New yaml2json package helps with automatical conversion from YAML to JSON.

### DIFF
--- a/repository/y.json
+++ b/repository/y.json
@@ -12,6 +12,16 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
https://packagecontrol.io/packages/SerializedDataConverter does this without external dependencies.

Also, your keybindings are flawed since pressing ctrl+s will prevent the file from being saved, and most importantly, is active for *all* yaml files.

PR #4978, opened 2015-11-10
Added CloseTabsLeft

deXterbed:
There's no support for closing tabs on the left in ST2 as of now. TBH i never use the existing "close tabs on the right" command and I was a bit shocked to find out that there wasn't an option for the opposite, so there goes my first stab at python and sublime plugins. Would love some feedback from you guys. Thanks a lot for the docs and all the help!

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1072,6 +1072,16 @@
 			]
 		},
 		{
+			"name": "CloseTabsLeft",
+			"details":"https://github.com/deXterbed/CloseTabsLeft",
+			"releases": [
+			{
+				"sublime_text": "<3000",
+				"tags": true
+			}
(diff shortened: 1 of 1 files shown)


FichteFoll:

deXterbed:
@FichteFoll done!

FichteFoll:
In your package info you define the tag prefix as `"st2-"`, but you don't use that prefix for your tags. Either you set the tag prefix to `true` (which is either none or `"v"`) or push a prefixed tag to the repo.

deXterbed:
thanks @FichteFoll, updated PR (replaced "st2-" with true)

deXterbed:
thank you so much :)

PR #4986, opened 2015-11-13
Add new packages and dependencies

facelessuser:
- Add ColorHelper package
- Add markdown dependency
- Add mdpopups dependency

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1820,6 +1820,15 @@
 			]
 		},
 		{
+			"details": "https://github.com/facelessuser/ColorHelper",
+			"releases": [
+				{
+					"sublime_text": ">=3080",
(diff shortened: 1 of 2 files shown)


facelessuser:
Okay, I need to rename `markdown` dependency to something like `sublime-markdown` so the folder name doesn't clash with the `Markdown` package name on windows.  Really didn't think about that until now.

facelessuser:
Okay renamed `markdown` to `python-markdown`.  Also added `backrefs` dependency.

FichteFoll:
Good stuff :+1: 

> so the folder name doesn't clash with the Markdown package name

Neither did I think of it in the tests. Case sensitivity I mean.

We should really store dependencies somewhere else ... in the near future.

PR #4989, opened 2015-11-15
add ShenMa plugin

molee1905:
1.a scaffold for shenma 
2.integrate some tools for developing in future

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1047,6 +1047,17 @@
 			]
 		},
 		{
+			"name": "ShenMa",
+			"details": "https://github.com/molee1905/ShenMa",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"platforms": ["osx"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Shinken",
 			"details": "https://github.com/Frescha/shinken-sublime",
 			"labels": ["language syntax", "snippets"],


FichteFoll:
I'm having problems understanding what this package does. Do you mind explaining for me? You might also want to consider updating the readme with that.

There will also be problems regarding the distribution of this package since you are performing file operations on a sub-folder, but the distribution will be served as a `.sublime-package` zip archive (on ST3). It is probably easiest to provide a `.no-sublime-package` file in your repo to explicitly tell PC to unextract your package.

Also, check out `sublime.executable_path()`.

molee1905:
Thanks for checking, i can explain and add it to my README.md:

it's a plugin to create projects skeleton based on defined scaffolds( see the template folder), in my work, i call these projects "sc card" (shortcut). Through this plugin, for example when i input `news_detail`, it will generates folders(black) and files(italic) like below: 
* **news_detail**
  * **css**
	  * *sc-advanced_news_detail.css*
  * **data**
  * **img**
  * **js**
	  * *sc_news_detail.js*
  * **res**
  * **tmpl**
  * *.eslintrc*
  * *index.html*
  * *README.md*

all i want is the folders mentioned just now, it will be appeared in my workspace (e.g. `/Users/jack/xxx/sc/shortcuts/`), the `news_detail` is for users not for sublime, and the `news_detail` is called **sc name**, the destination of the folder is called **sc path**

As it is declared in the repository.json, this plugin just works on sublime text 3 and osx, `sublime.executable_path()` will return `/Applications/Sublime Text.app/Contents/MacOS/Sublime Text`, the func `get_subl_executable_path()`
will return `/Applications/Sublime Text.app/Contents/SharedSupport/bin/subl`,  so, there is no problem with `sublime.executable_path()`

As you suggested, .no-sublime-package is added, thanks a lot

PR #4990, opened 2015-11-16
add Googleapi2Useso

BruceZhang1993:


FichteFoll:
I think we can assume users to know about the Find and Replace tool that Sublime Text provides to perform this small task.

If they want to automate the replacement (with a key binding for example), they should use https://packagecontrol.io/packages/RegReplace.

PR #4992, opened 2015-11-16
Search Stack Overflow and get code snippets from answers directly to your Sublime Text

rinas7:


FichteFoll:
Pretty good idea imo.

You can show multiple rows in a quick panel item, by the way. Shouldn't include too many lines, but will still be better readable than currently.

rinas7:
Thanks for the advice, @FichteFoll! I've updated the plugin to show snippets in multi-row items as you suggested. They're much more readable now.

PR #4993, opened 2015-11-16
Adding JSON Region to package control

dusk125:
JSON Region allows plugin developers to easily encode and decode sublime.Regions to a json object.

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -845,6 +845,17 @@
 			]
 		},
 		{
+			"name": "JSON Region",
+			"details": "https://github.com/dusk125/sublime-jsonregion",
+			"labels": ["region", "json", "encoder", "decoder"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "JSON Reindent",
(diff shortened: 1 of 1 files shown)


FichteFoll:
I question the usefulness of this "package". Can you provide an example?

Either way, this seems to be more suited to be provided as a dependency instead of a package, if anything

dusk125:
No problem, my use case was attempting to save bookmark regions in my plugin Better Bookmarks, because Regions aren't JSON serializable, I had to figure out how to do what this project does. I thought I would share it so others could easily do this. I can instead submit this as a dependency if you think it's better suited there. Attached is a gist of the saving and loading of bookmarks from my plugin: [BetterBookmarks.py lines 97:110](https://gist.github.com/dusk125/2add421d2d475d3f20d1)

FichteFoll:
Sorry, it appears I forgot to answer.

I still find this to be a very specific use case that can hardly be used, but if you want to make this available to other packages this should *definitely* be implemented as a dependency. (I also can't merge this PR anymore.)

Thus, I'm closing this. Please open a new PR with this as a dependency or not. Your call.

PR #4997, opened 2015-11-18
added Sprak Syntax & Completions for scripts

Eforen:
added Sprak Syntax & Completions for scripts from the game else Heart.Break() which has scripting in it and this gives users of Sublime the ability to handle scripts copy and pasted from or too the game.

Signed-off-by: Ariel Lothlorien <git@arielkauai.com>

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -2222,7 +2222,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"branch": "master"
+					"tags": true
 				}
 			]
 		},
--- a/repository/s.json
+++ b/repository/s.json
@@ -2215,6 +2215,18 @@
 			]
 		},
 		{
+			"name": "Sprak Syntax & Completions for else Heart.Break()",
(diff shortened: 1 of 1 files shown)


FichteFoll:

Eforen:
Oh ok will do

FichteFoll:
Okay, good. However, there is another problem that I just noticed: You have a dot in your package name. This will prevent any Python plugins from loading (on ST3) and even thoug you don't have any currently, we don't accept any new packages with dots.

Eforen:
Oh I was unaware that was a problem I have yanked the dot.

Eforen:
I made a change to my package and taged it v1.0.1 do I need to tell packagecontrol.io somehow that there is a new version? if not how long does the automated build process take?

FichteFoll:
No, just pushing the tag is enough. It may take up to 1 hour though.

Eforen:
Oh ok cool thx it is updated now

PR #5004, opened 2015-11-24
added Network Config Syntaxt to repository

ghost:


FichteFoll:
Both packages you reference, https://github.com/tunnelsup/sublime-cisco-syntax and https://github.com/nprintz/junos-sublime-pkg, are already on Package Control. I do not see a purpose for adding these as a single package. Please contribute your changes to the individual packages as pull requests instead.

PR #5007, opened 2015-11-25
Added JSX Comment to repository

Briix:


FichteFoll:
This is a really barebones package, but why not use "built-in" comment support from JSX syntax definitions like https://github.com/allanhortle/JSX or https://github.com/babel/babel-sublime? (the latter defines `{/* ...*/}`-style comments while the first defines usual javascript ones)

Briix:
@FichteFoll You're absolutely right. I had originally looked at the JSX packaged, but I completely missed the babel one. There's no use for my package, my bad.

PR #5013, opened 2015-11-28
Add Project Planner

pedrokost:


FichteFoll:
1. You can not use `shutil.copyfile` on ST3 because package are in archived state (`.sublime-package`) there by default. Either use `sublime.load_binary_resource` on ST3 only or specify a `.no-sublime-package` file, which tells PC to specifically unextract your package. Note that this is suboptimal.
What is the benefit of copying the default settings to the user folder anyway? No other package does this.
2. Your relative imorts in the root folder (e.g. `from .utils import ....`) will fail on ST2. They must be absolute there instead (`from utils import ...`).
3. Try not to mess with `sys.path`, but especially do not insert anything in the front and not remove it when you are done. Just use relative imports (should work in your case).
4. Please use the `requests` Package Control dependency instead of bundling your own: https://github.com/packagecontrol/requests
5. Your `lib/trollop/.hg` folder can safely be removed.

pedrokost:
Hi @FichteFoll 

I've handled the mentioned issues, except issue number 2. I don't know how to eliminate the relative imports and since I haven't tested the plugin for ST2, I would rather not have people try it.

I've made a new release with these and other changes (v1.1.0)

FichteFoll:
If you don't want people to install your package on ST2, you should adjust the `sublime_text` version selector in this pull request to `>=3000`.

pedrokost:
I have limited it to ST3 only

PR #5015, opened 2015-11-30
add cssnext

zhouwenbin:


FichteFoll:
1. You'll need to change `"branch": "master"` to `"tags": true` and use [tags](http://git-scm.com/book/en/v2/Git-Basics-Tagging) to create releases and push valid [**semantic version**](http://semver.org) tags to your repo. As mentioned [in the docs](https://packagecontrol.io/docs/submitting_a_package#Step_4), we no longer accepts branch-based releases for new packages.

2. Please remove the `package-metadata.json` file from your repo. It is supposed to be auto-generated by Package Control on installation.

3. Please remove all `.pyc` files from your repo and add them to `.gitignore`.

4. Your Python code is copypasta'd and partly flawed, but I already ranted over it somewhere else and can't find it right now.

5. `sublime.packages_path()` may not be available at the time you call it and just return `None`, making everything fail.

6. Because you are invoking Node, you need a `.no-sublime-package` file.

FichteFoll:
Actually, what is the difference to the original, https://github.com/sindresorhus/sublime-autoprefixer, in the first place?

zhouwenbin:
Thanks a lot.It contains autoprefixer and other useful plugn.[http://cssnext.io/features/](http://cssnext.io/features/)

FichteFoll:
Okay, so one uses the `autoprefixer` node module, while yours uses `postcss-cssnext`, which actually *also* wrappes `autoprefixer` but does additional things. Correct?

You should really link cssnext.io in your readme.

zhouwenbin:
Many thanks. I had link cssnext in the License section.

PR #5023, opened 2015-12-03
use master banch

xiewandongqq:


FichteFoll:
For the same reason we don't accept new packages with branch-based releases anymore, we also don't accept packages switching back to it. It's deprecated.

PR #5026, opened 2015-12-04
LaTeXtools entry

msiniscalchi:
Changed entry to use tags rather than branch (finally! I should have done this much earlier!)

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -247,7 +247,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"branch": "master"
+					"tags": true
 				}
 			]
 		},


msiniscalchi:
Question: since "branch" syncing was used up to today, will the switch to "tags" create issues for users? I have created a tag that corresponds to the most up-to-date commit in the master branch, so pulling that would not change anything from the users' point of view. Hope that's OK!

FichteFoll:
The change will be completely transparent for users (as in: they won't notice). Package Control will re-download the package as if there was an update (technically, this is an update).

Curious, but did your previous update messages actually show to anyone? You'd need to hit the exact timestamp for commiting the change, I suppose.

msiniscalchi:
@FichteFoll great! Thank you.

And yes, messages were shown. At first my very hacky workflow was like thus: push all functional commits, then check the github timestamp of the last commit and do a git commit -c (or --amend) with just the messages.json entry, with the correct timestamp. However, I then figured out that, in practice, the timestamp in messages.json needn't be exact; using the correct day and hour was enough. So in my last functional commit I also updated messages.json with an approximate timestamp, and that worked fine.

Obviously, very hacky. Happy to have finally switched to tags!

FichteFoll:
>the timestamp in messages.json needn't be exact

Good to know :+1:

PR #5030, opened 2015-12-05
Added Ruby Completions

tushortz:
* Added support for "require gems" suggestions
* function completion suggestions


FichteFoll:
You can append commits to existing pull requests by pushing onto the branch you opened it from.

PR #5035, opened 2015-12-07
Add EPL as a new package

StefanE:
Add basic support for EPL programming language, and symbol navigation. 

Should be using tags instead of branch now.


StefanE:
Hmm reusing old pull request instead...

PR #5036, opened 2015-12-07
Added `External_Programs`

Hibou57:
It's neither the same as `External Command` nor `External REPL`. See the README file for differences and rationals.

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -1077,6 +1077,7 @@
 		{
 			"name": "External_Programs",
 			"details": "https://github.com/Hibou57/External-Programs-ST3",
+			"donate": "https://pledgie.com/campaigns/30727",
 			"labels": ["text manipulation", "utilities"],
 			"releases": [
 				{
--- a/repository/e.json
+++ b/repository/e.json
@@ -1075,6 +1075,17 @@
 			]
 		},
 		{
+			"name": "External_Programs",
+			"details": "https://github.com/Hibou57/External-Programs-ST3",
+			"labels": ["text manipulation", "utilities"],
+			"releases": [
+				{
+					"sublime_text": ">=3083",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Have you taken a look at https://packagecontrol.io/packages/Commando?

Hibou57:
I just had a look at it, and `build_like` may seems to overlap it. However, this differs a lot with this (from the wiki):
> $input - The input of the command (usually the output of the previous command.

That's a bit too match for something simple, and it may be good to keep things simple for multiple reasons.

Later the wiki even says:
> More to come ...

Something I personally won't install.

Hibou57:
Any way, I just added a mention to it in the README file.

FichteFoll:
Still not entirely sold on this. It does seem a lot like "basically the same functionality as others, but with different configuration".

Variable expansion in the `build_like` command seems useful, but could also be achieved by defining a build system and then calling that using the `build` command and parameters (after build system rework - to be documented), so I suppose that's still an ease of configuration.

After all, configuration *does* qualify as a feature so it it's different enough, which I think it is, we can accept this package.

Hibou57:
> and parameters (after build system rework - to be documented)
Yes, I know a rework is pending which is not yet documented. If after the rework and the documentation, the `build_like` looks obsolete, then I will just remove it and tell with a PackageControl message.

For the links, that's because I'm using MultiMarkdown primarily. I will try the relative anchors so that it works with simple Markdown. Thanks for the comment on this.

FichteFoll:
The build system rework (which is mostly about "build selection") is already complete. It's just nowhere documented.

You can investigate some parameters by experimenting with it and having `sublime.log_commands(True)` enabled, which is what I did so far.

PR #5045, opened 2015-12-09
Add AltUp package

germtb:


FichteFoll:

germtb:
Should the tag be on the git repository of the package? (if so I thought I had added it already) or should it be in my fork of the package control channel?

FichteFoll:
Should be on the package repo. I don't remember if the tag had been added already. Since I copy-paste most of my messages, it could have been that I just forgot to check or didn't delete the section.

PR #5048, opened 2015-12-10
Enable Open Related package for ST3

denniskempin:
The package is pretty much abandoned but works well on ST3.

### DIFF
--- a/repository/o.json
+++ b/repository/o.json
@@ -372,7 +372,7 @@
 			"details": "https://github.com/vojtajina/sublime-OpenRelated",
 			"releases": [
 				{
-					"sublime_text": "<3000",
+					"sublime_text": ">=3000",
 					"branch": "master"
 				}
 			]


FichteFoll:
Actually, as [this comment](https://github.com/vojtajina/sublime-OpenRelated/commit/7fd16589a4168d7330b6c9ab24682390e8da9c3d#commitcomment-3062749) mentions, it doesn't actually work on ST2 anymore.

It should be as follows:

```
if int(sublime.version()) >= 3000:
    from . import converter
else:
    import converter
```


cc @vojtajina

denniskempin:
I never tested it on ST2. I'll send a pull request along his way. This plugin does such a minor little thing... but it really helps my workflow :)

FichteFoll:
This doesn't seem to go either way. I suggest just setting this to `>=3000` until the PR is accepted.

denniskempin:
I apologize for the messy change. I've never worked with GitHub pull requests before and it's too late for me to be doing this ;)

Let me know if this is ok or if there is a way to squash this series of commits into a single change. I usually do history editing for this, but I think GitHub does not like that.

FichteFoll:
You can do history editing all you want, actually. Squashing or rebasing and then force-pushing is common practice for pull requests. You can try it here.

Tip for next time: Create a new branch and create a pull request from that instead of working on `master`. It will make things easier in case you are creating multiple PRs simultaneously.

denniskempin:
Perfect! I've been told it's poor practice on github, but this fits well with the workflows I'm used to. Dankeschoen!

FichteFoll:
Well, it depends on the pull request's and the commits' scope, but for the kind of changes to this repo, squashing is mostly better than issueing more commits.

PR #5050, opened 2015-12-11
add BeautifyRust package for ST3

vincenting:
BeautifyRust is a binding for Sublime Text to the Rustfmt.

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -231,6 +231,16 @@
 			]
 		},
 		{
+			"name": "BeautifyRust",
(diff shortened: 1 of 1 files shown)


FichteFoll:

Also, this is not how `Settings.add_on_change` is used. The string parameter is an identifier to be used later with `Settings.clear_on_change`. So this should be something that can identify your plugin, e.g. `__name__`. The callback then gets called whenever any change *could have happened*, i.e. when settings had been reloaded.

vincenting:
Thank you for reminding. I have fixed them both.

PR #5051, opened 2015-12-13
Add GobuildEx plugin package json

jerrytaffy:
This is a Sublime Text plugin with Golang develop. He difference is that, before it will be killed your process. He is a very very very simple plugin.

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -1621,6 +1621,18 @@
 					"branch": "master"
 				}
 			]
+		},
+		{
+  			"name": "GobuildEx",
+  			"details": "https://github.com/jerrytaffy/GobuildEx",
+  			"releases": [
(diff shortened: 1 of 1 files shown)


FichteFoll:
Tests are failing: https://travis-ci.org/wbond/package_control_channel/builds/96558693

Why do you not add this to an existing package instead?

FichteFoll:
@jerrytaffy ping

jerrytaffy:
ping??? How ping??? Terminal ping command???

FichteFoll:
@jerrytaffy I was simply asking you to address the issues I raised in my earlier comment. "pinging" is usually done to check whether your communication partner is still available, which is normally common for software but also occasionally used for person-to-person communication.

PR #5052, opened 2015-12-13
Internal plugin 

mockingbird87:
Plugin for MediaMonks to assist in our grunt workflow.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -1131,6 +1131,17 @@
 			]
 		},
 		{
+			"name": "MM Banners",
+			"details": "https://github.com/mockingbird87/MM-Banners",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"platforms": ["windows"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Moai Debugger",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Is this supposed to be a package for a proprietary audience? If yes, you might want to consider adding this to your peers' `repositories` setting using the "Package Control: Add Repository" command.

Other things:

1. You specify this package as ST3-only, but are doing ST3-incompatible things in your code. Notably, `begin_edit` and `end_edit`. Use the built-in "insert" command instead. Furthermore, you should define a `plugin_loaded` function which will be called once the API methods are available, otherwise you package may not be properly initialized (referring to your `set_timeout` call at the end).

2. You specify the package as Windows-only but have keybindings for all OSes in your package.

3. OS-specific settings are not a thing for anything that is not a `Preferences.sublime-settings` file. I.e., you need to interpret your `... (Windows).sublime-settings` manually with plugin code.

4. You have to push a valid [semantic version](http://semver.org) [tag](http://git-scm.com/book/en/v2/Git-Basics-Tagging) (with optional leading `v`) to your repo.

FichteFoll:
@mockingbird87 ping

PR #5055, opened 2015-12-14
Commit package BM Applescript Toggle Comments

bammoo:


FichteFoll:
Why don't you just use the standard method of defining a [`.tmPreferences` file with the comment sequences](http://docs.sublimetext.info/en/latest/reference/comments.html), as is also mentioned in the [corresponding bug report](https://github.com/SublimeTextIssues/DefaultPackages/issues/44)

bammoo:
@FichteFoll  Oh, I didn't see it before. Thank you for pointing out.

PR #5056, opened 2015-12-15
Added DocBlockr_Python to channel.

adambullmer:
Port of the DocBlockr project but for Python docstrings. Uses similar architecture and strategies for detecting docstrings and definitions. as the DocBlockr project, but the engine to get the definition of the class/function had to be rewritten entirely to allow for parsing python code and not C based code.

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -693,6 +693,26 @@
 			]
 		},
 		{
+			"name": "DocBlockr_Python",
+			"details": "https://github.com/adambullmer/sublime_docblockr_python",
+			"releases": [
+				{
+					"platforms": "*",
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			],
+			"labels": [
+				"python",
+				"docstring",
+				"automatic",
+				"formatter",
+				"generator",
+				"auto-complete",
+				"snippets"
+			]
+		},
+		{
 			"name": "Docker Based Build Systems",
 			"details": "https://github.com/domeide/sublime-docker",
 			"labels": ["build system","build"],


FichteFoll:
1. In https://github.com/adambullmer/sublime_docblockr_python/commit/25566261b916fe3f4565ae3c75cf998baac042b4 you say you fixed a namespace issue, but this will not help here. It actually makes things worse since the package will be installed as `Docblockr_Python` (and yes, Python is case-sensitive).
However, there is a proper way to do this: relative imports.

2. `readme`, `issues`, `donate` and `author` keys are redundant, because Package Control expands these automatically. You can just remove them.

adambullmer:
Trimmed this repository keys as requested.

You're absolutely right about the import name. I actually have a separate branch ready for this, but I had to use the current `sublime_docblockr_python` name for anyone who installed this by adding the repo to package control manually.

I have an open [PR](https://github.com/adambullmer/sublime_docblockr_python/pull/7) I can merge and tag a release if that's all that is keeping this from being approved.

FichteFoll:
No really, just use *relative imports*. They work in any folder.

adambullmer:
- [x] Adjusted to use relative imports.

Merged into master, tagged and cut a release

FichteFoll:
Btw, I'm not entirely sure why you are using `import_module` directly with a hard-coded string [here](https://github.com/adambullmer/sublime_docblockr_python/commit/46586b15663b9b75681fd8d7061e96dc43e981cd#diff-ab930a63cc886128b6d9ae0191520cfcR26), but I'll merge it anyway since this likely won't be an issue.

adambullmer:
Good look! I think that's left over from earlier import errors I was getting during development, and just never took it out. I'll get on fixing that.

PR #5057, opened 2015-12-15
Add DiffView package

CJTozer:
Side-by-side diff viewer for VCS comparisons.
* Currently supports Git only.
* Only tested on Sublime Text 3 (hence restricted to ST3 for now).

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -461,6 +461,17 @@
 			]
 		},
 		{
+			"name": "DiffView",
+			"details": "https://github.com/CJTozer/SublimeDiffView",
+			"labels": ["Git", "diff", "compare", "difference", "comparison"],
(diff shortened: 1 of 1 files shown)


FichteFoll:
Differences to [other diff packages](https://packagecontrol.io/search/diff)?

CJTozer:
None of the other diff packages show changes from VCS in a side-by-side way with difference highlighting.

I think the existing packages fall into X camps:
* those that just show raw `git diff` output
* those that show diffs between 2 files, but require the user to input the 2 files manually
* [paid, closed source](http://www.sublimerge.com/)

I've updated the tags to be semantic.

FichteFoll:
Okay, I'll trust you on this one.

PR #5058, opened 2015-12-16
added "String Case" package

oriSomething:
it adds a package that help to easily change case of selected text by keybindings.

the package:
https://github.com/oriSomething/string-case-sublime

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -2599,6 +2599,17 @@
 			]
 		},
 		{
+			"name": "String Case",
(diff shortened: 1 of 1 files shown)


FichteFoll:
What about [this other package](https://packagecontrol.io/packages/Case%20Conversion) with 24k installs and easily searchable keywords in its name?

oriSomething:
wow, thanks. i have no idea why i didn't find this before i build this package, because i searched it for awhile

PR #5059, opened 2015-12-16
Added Highlight Dodgy Chars repo

TuureKaunisto:
This is a plugin that highlights non-ascii chars making eg. spotting zero width spaces much easier.

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -432,6 +432,16 @@
 			]
 		},
 		{
+			"name": "Highlight Dodgy Chars",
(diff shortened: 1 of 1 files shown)


FichteFoll:
If you are whitelisting scandinavian special characters by default, you might want to consider other european special characters such as in French or German. Otherwise you might as well not define any whitelist characters by default and have the user decide this on his own, which would be more consistent imo.

TuureKaunisto:
A good point! I'll add European special characters to the whitelist to the next version. The Scandinavian characters were just an example but it would be nice to have a wide set of characters in the default settings.

Issue #4310, opened 2015-04-12
Write a formatter utility


FichteFoll:
The formatter should:

- [ ] Unify labels and translate them to lower case (#4277)
- [ ] Unify repository urls and remove trailing slahes for example
- [ ] Remove redundant keys that would be equivalent to what PC finds from `details`, notably `readme`, `author`, `base`
- [ ] Remove redundant keys that specify the general default value, notably `platforms`
- [ ] Unify whitespace usage. I'm thinking of what [`example-repository.json`](https://github.com/wbond/package_control/blob/master/example-repository.json) uses, which is:
  - Block style by default, that means a new line for each value
  - Inline style (in same line) for `labels`, `author` and `platforms` keys
  - The same indentation/whitespace style (I don't think I need to explain this)
  - No trailing whitespaces
  - Newline at EOF
- [ ] Add a `name` key for github repos if none present? We could make this a soft requirement in the process.

More?

yyyc514:
Seems reasonable... are contributors going to be the ones required to run this tool after making any changes or adding a plugin?  Not sure we want to make that process any more difficult for newcomers.

FichteFoll:
No, I don't think we can reasonably require package authors to run it, but it will be a part of the ChannelRepositoryTools package so that they can do it optionally. There doesn't seem to be a reasonable way to include this in the travis ci build either.
I mainly intend this to be a "run once in a while" think for maintainers.

twolfson:
For Travis CI, we might be able to run the formatter then diff to see if there are changes/not. If there are changes (i.e. formatter wasn't run since last commit), then we could complain and exit with a non-zero code.

robfrawley:
@twolfson +1

FichteFoll:
Some of the formatter features have been added as tests in #7853, so the user can fix the issue by themselves.

FichteFoll:
Better idea: We can have a github action run the formatter on pull requests and push to the branch when there are changes and run the tests afterwards.

michaelblyons:
In case lurkers missed it: see #9378

PR #5063, opened 2015-12-18
Update DiffView tags

CJTozer:
Didn't realise they were case-sensitive on packagecontrol.io


FichteFoll:
This is in discussion to get changed btw: #4310

PR #5064, opened 2015-12-18
add MoveMore package

sflip:
This package makes it possible to move the cursor up or down more than one line at a time. As described in the readme file, this is a missing but wanted feature.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -1601,6 +1601,18 @@
 			]
 		},
 		{
+			"name": "MoveMore",
+			"author": "sflip",
+			"description": "Move cursor up or down more than one line",
+			"details": "https://github.com/sflip/SublimeMoveMore",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
https://packagecontrol.io/packages/LineJumper

sflip:
Oh,  I did not find that! 

However, I see one difference: With my plugin you can use the same key binding to extend the selection if there is one and do not create one if there isn't. This is the desired behaviour for vim (Vintage/Vintageous) users like me.

FichteFoll:
Easy:
```js
[
    {
        "keys": ["alt+up"],
        "command": "line_jumper",
        "args": {
            "number_of_lines": 10,
            "cmd": "up_select"
        }
    },

    {
        "keys": ["alt+up"],
        "command": "line_jumper",
        "args": {
            "number_of_lines": 10,
            "cmd": "up"
        }, "context": [
            { "key": "selection_empty", "operator": "equal", "operand": true, "match_all": true }
    }
]
```

And that way it is even user-configurable.

sflip:
All right.

PR #5065, opened 2015-12-19
Added sublime-hdlproject parameters to h.json.

bootsiaz:
Request to add a VHDL/Verilog based plugin for Sublime Text 3.

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -266,7 +266,7 @@
 			]
 		},
 		{
-			"name": "HDL Project",
+			"name": "HDLProject",
 			"details": "https://bitbucket.org/bootsiaz/sublime-hdlproject",
 			"releases": [
 				{
--- a/repository/h.json
+++ b/repository/h.json
@@ -266,6 +266,16 @@
 			]
 		},
 		{
+			"name": "HDL Project",
+			"details": "https://bitbucket.org/bootsiaz/sublime-hdlproject",
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
1. Your repo structure is incompatible with package control. Python plugins need to be at the root of the repository.

2. Your imports will fail after installation because you hard-coded `HDLProject` as the top level for imports, yet your package will be installed as `HDL Project` (with a space). You should use *relative imports* where possible, i.e. everywhere.

3. Your `HDL_PROJECT.py` entry point file is available in both `.py` and `.pyc`. Please remove the pyc variant.

4. Just in case you didn't do this intentionally: this package is closed source. Usually this is only done when you want to sell a package and have license restrictions in the code. Otherwise, all packages are open source to allow for easier contributions. Your decision.

bootsiaz:
1. I have updated my plugin to be at the root of the repo, thanks. 
2. I can't use relative imports because this package is not on the sys.path. I will stick with paths relative to the Sublime package directory.  I have removed the space in the name of the project. 
3. I have removed the .pyc file you mentioned. 
4. I am intentionally keeping it closed source for now.

bootsiaz:
I have updated my Sublime Text 3 package to use relative paths for imports as you requested. 

I am still thinking of making this a commercial package in the future, so I am not comfortable revealing the source. 

I have updated the documentation as well. 

Let me know if there is anything else you need.

PR #5068, opened 2015-12-20
add Typora Markdown App (OSX)

Asoul:
add Typora.app for markdown editor, forked from [Mou Markdown App (OSX)](https://github.com/rwoody/mou-markdown-sublime)


FichteFoll:

Asoul:
ok, I've used tag instead.

PR #5069, opened 2015-12-20
TomcatMavenHotReload

francescou:


FichteFoll:

The package doesn't work on Windows. Either use `os.path.join` and `os.path.sep` or restrict the package to OSX and linux only.

francescou:
thank you for your help, it should be fine now

PR #5072, opened 2015-12-22
add google dictionary plugin

houcheng:
Hi,
I've added my google dictionary plugin into g.json. Please help to merge this commit.
Thanks a lot,
Signed-off-by: Houcheng Lin <houcheng@gmail.com>

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -922,6 +922,17 @@
 			]
 		},
 		{
+			"name": "Google Dictionary",
+			"details": "https://github.com/houcheng/GoogleDictionary",
+			"labels": ["dictionary", "google", "translator"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "master"
+				}
+			]
+		},
+		{
 			"name": "Google Reader",
 			"details": "https://github.com/speg/SublimeGReader",
 			"releases": [


princemaple:
I'm not a collaborator on package control, just a walk by comment: `ctrl+b` conflicts with default sublime key map of `build` command. It's a pretty common command. Could you consider removing the default key map, instead just provide it as an example so if anyone wants to use it, they could just take what you have provided?

princemaple:
@FichteFoll just saying, I think it'd be really great if all the package developers can leave out the shortcuts and context menus, instead just provide them as examples & commands. Unless the whole point of the plugin is to key map something, e.g. vim.

princemaple:
@houcheng I think your plugin is great, it will be even better if you could utilize the tooltip functionality of ST3 instead of pushing more text onto the already crowded status bar. I love this [colorhelper](https://github.com/facelessuser/ColorHelper) plugin, you can see how tooltips can come in handy.

houcheng:
@princemaple 
Thanks for your comment, it is great and helps me a lot on plug-in design. 
The colorhelper's tooltip effect is what I need, I 'll reference this plugin and
display the translated text as tips. I'll also remove the default key binding.

princemaple:
@houcheng :+1: :+1:

FichteFoll:
As @princemaple rightfully pounted out, ctrl+b is a very commonly used key binding and your package will not be accepted as long as it's overriding it.

Furthermore, you'll need to change `"branch": "master"` to `"tags": true` and use [tags](http://git-scm.com/book/en/v2/Git-Basics-Tagging) to create releases and push valid [**semantic version**](http://semver.org) tags to your repo. As mentioned [in the docs](https://packagecontrol.io/docs/submitting_a_package#Step_4), we no longer accepts branch-based releases for new packages.

@princemaple, I usually call package authors out when I notice them overriding default bindings, but I don't have a policy for whether or not I would be accepting a package if the author didn't change the bindings or remove them. I am very supportive of not defining key bindings by default and will recommend this in our package development guide at http://docs.sublimetext.info/ (when we finally get to it).

FichteFoll:
Oh, and I almost forgot. Please do not modify `sys.path` when it can be avoided. Use relative imports for ST3 instead and change them to absolute imports for package-revel modules on ST2. You can check my plugins for examples.

houcheng:
@FichteFoll 
Sorry for late to response, as I went to Japan for new year vacation. The tool tips and semantic version are cool, I'll reference this design and follow version rule to tag.
The modification to sys.path is hard to avoid, I've tried to use relative import but in third party library, there are a lot of direct import in various import form.

PR #5074, opened 2015-12-22
Added the SodaSeti theme to the repository

domibarton:


FichteFoll:

You can not use submodules to install packages via Package Control. They are not supported.

domibarton:
You mean "RTFM" - sorry my fault ;)

I removed the git submodule, created a new v1.0.0 tag in my repo and updated your repo file to use tags instead of the master branch.

Hope that's OK?

PR #5078, opened 2015-12-26
'Select Regex' package

mvoidex:
https://github.com/mvoidex/SublimeSelectRegex
It also replaces 'Select By Regex', which is not maintained anynore: https://github.com/tarwich/SelectByRegex/issues/5#issuecomment-167193286

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -644,7 +644,7 @@
 		},
 		{
 			"name": "Select By Regex",
-			"details": "https://github.com/mvoidex/SublimeSelectRegex",
+			"details": "https://github.com/mvoidex/SublimeSelectByRegex",
 			"releases": [
 				{
 					"sublime_text": "*",
--- a/repository/s.json
+++ b/repository/s.json
@@ -643,7 +643,7 @@
 			]
 		},
 		{
-			"name": "Select Regex",
+			"name": "Select By Regex",
 			"details": "https://github.com/mvoidex/SublimeSelectRegex",
 			"releases": [
 				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
>Will it replace package on update (which may not be good)?

Yes, it would be replaced. Which is good imo, unless you removed significant functionality. The old package is discontinued anyway. Thus, I suggest adding the old name to a `previous_names` array if you don't wish to continue the package name.
I suggest preparing an update message for the takeover however.

I don't need extra confirmation from @tarwich btw, the referenced issue is enough.

FichteFoll:
Also, the rename caused the package to be out of order.

mvoidex:
> I suggest preparing an update message for the takeover however.

Ok, I'll add update message

mvoidex:
Seems, that 'Select By Regex' used incorrect tags: 'v1.1.2' instead of '1.1.2' (or is it acceptable too?)
If I use tag '2.0.0' for my plugin, will it show update message correctly?

mvoidex:
I've renamed repo name, added tag '2.0.0' and update message.
Please, let me know, If there's anything I have to do

FichteFoll:
>Seems, that 'Select By Regex' used incorrect tags: 'v1.1.2' instead of '1.1.2' (or is it acceptable too?)

Package Control will automatically remove the `v` prefix from tags if it finds one. For other prefixes, one would have to specify a string with the prefix for the `tags` key.

>If I use tag '2.0.0' for my plugin, will it show update message correctly?

Yes.

mvoidex:
Ok, thanks for help!

PR #5080, opened 2015-12-27
add Yeti UX & UI

ctf0:


ctf0:
could someone tell me what exactly is the error ???

FichteFoll:
1. [Semantic versions](http://semver.org) consist of exactly three numeric parts. This applies to all repos.
2. Yeti_UI does not need a `.no-sublime-package` file. Please remove it.
3. Yeti_UI does not contain a color scheme. Please remove the label.
4. Personally, I find the naming convention of "UI" vs "UX" **very** confusing, but the labels kind of help.

ctf0:
- files removed.
- i left the color scheme tag as i will add the original schemes port later once they are done, so instead adding them and sending a new PR i kept the tags.
- the naming is to differentiate the theme from the color scheme

FichteFoll:
>as i will add the original schemes port later once they are done

But wouldn't that leave the _UX packages redundant since the color schemes are included in the UI packages already? In that case you'd need to remove those anyway and could as well add the label to the UI packages later.

ctf0:
the UX package is different, its more like a personal mod of the original scheme + i didn't want to force it over the user so i thought making a different package would be better for both worlds.

FichteFoll:
Ah okay, that wasn't transparent to me at all. (Maybe because of the package names? I'm starting to see a pattern :see_no_evil: )

Well, whatever drives your boat.

ctf0:
big thanx :+1:

PR #5087, opened 2015-12-29
Swift Foundation Framework Autocompletion Snippets

hatunike:
A SublimeText autocomplete snippet package for the Swift Foundation framework. For use in .swift files that have access to Foundation.

Note : the snippets were generated from the open source version of Foundation. As long as the api's are synonymous with the closed Foundation api, these snippets can be used without issue.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -3755,6 +3755,16 @@
 			]
 		},
 		{
+			"name": "Swift Foundation Autocomplete Snippets",
+			"details": "https://github.com/wbond/sublime_alignment",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Swig",
 			"details": "https://github.com/enagorny/sublime-swig",
 			"releases": [
--- a/repository/s.json
+++ b/repository/s.json
@@ -3755,7 +3755,7 @@
 			]
 		},
 		{
-			"name": "Swift Foundation Completion",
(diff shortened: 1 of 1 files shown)


FichteFoll:
You should totally take a look at Completions files: http://docs.sublimetext.info/en/latest/reference/completions.html

Also, "auto completion", by convention, implies context-aware completions based on the variable type or available variables in general. Please remove the "auto" part from your name.

hatunike:
Nice, those completion files look much better. I'll make the suggested change in the name and add converting to snippet files to the road map. 

As for additional scoping and true autocompletion do you have any further recommendations?

FichteFoll:
You mean for writing an auto-completion plugin? Well, I can tell you how to provide completions dynamically (by linking to [docs](http://docs.sublimetext.info/en/latest/reference/api.html#sublime_plugin.EventListener.on_query_completions)) and I can refer you to other packages that do this already (such as SublimeCodeIntel).

hatunike:
I've changed the name of the plugin. Yes, you correctly inferred my question above. And thanks or the helpful resources. The tool that I wrote to generate the snippet files from Foundation should be easy enough to convert to output the completion files. But true autocompletion support for swift development in Sublime Text is the longer term goal. First for the Foundation Framework - then for imported modules installed from the package manager, and finally for the local source code of the project.

In the meantime, i've found just using the Snippets as submitted in their current form to be useful while writing swift.

PR #5093, opened 2016-01-03
Add ROS snippets

ckrooss:


FichteFoll:
Please remove the "auto-complete" label since that is something different.

PR #5099, opened 2016-01-05
rename `escape & unescape tool` to `Unicode Escape`

iahu:
rename `escape & unescape tool` to `Unicode Escape`


FichteFoll:

iahu:
Sorry.
I have added the previous name to `previous_names`.

PR #5100, opened 2016-01-05
Add SelectExact package

spywhere:


FichteFoll:
This might be surprising to you, but this exact functionality already exists in Sublime Text, although it's fairly hidden.
Basically, you have to open the find panel and assure that whole-word search is active. Then you proceed with placing a caret inside the word you want to expand but not select it! If you then repeatedly use the "find under expand" command, it will behave as your plugin.

However, since this is "fairly hidden" and you might want to be able to expand the exact selection you made and not rely on ST's automatic word expansion, this package is still valuable.
There probably is at least one package that performs a similar task, but I honestly can't find it right now and if I can't, then most other people can't either, meaning it's probably a lost cause.

spywhere:
Thank for the comment.

Wow, that's something I did know. I've tried searching through the forums on this functionality and quite surprise that there're no post about it. I do expected that ST should have this feature hidden somewhere but yeah, that's a "fairly hidden".

I've know that there's a package allow to search using RegEx, but that package is still not in my scope of work since that package will ask for the RegEx on the first selection (while mine is just go ahead and search it right away).

Anyway, thanks for the merge. :)

PR #5102, opened 2016-01-06
remove space between the name of `Direct Edit`

UniFreak:
After the installation, the folder name in Data/Package will be `Direct Edit/`, and this cause my package from functional. so I removed the space between the name. hoping this will end the fold name be `DirectEdit`.


FichteFoll:
Yes, it will result in a folder named "DirectEdit" (or in `DirectEdit.sublime-package`).

That said, please remove `.no-sublime-package` from your repo. It isn't necessary and should not be used if possible.

PR #5103, opened 2016-01-06
Add linux to ImportJS package platforms

trotzig:
This plugin has been tested on a Linux Mint 64 Bit machine, and it's
working. So I'm adding yet another platform to the list. Windows might
come later down the line, but I haven't had the chance to test yet.

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -253,7 +253,7 @@
 			"releases": [
 				{
 					"sublime_text": ">=3000",
-					"platforms": ["osx"],
+					"platforms": ["osx", "linux"],
 					"tags": true
 				}
 			]


FichteFoll:
By the way, your package will most likely "just work" on Windows too since you aren't using something OS-specific besides forward slashes for file paths in line 88, which Windows auto-translates to backslashes internally (although using `os.path.join` is recommended).

Another by the way, bundling plugins for all editors along with the executable itself in one repository will result in at least ST users having to download all files they do not have any interest in (such as the executable, which they will install with `gem` separately anyway). Thus, I recommend splitting the repositories.

trotzig:
The python plugin itself is probably ready for Windows, it's the executable that I worry about. Before I have time to look into making that work for Windows, I'd rather not allow people to install the ST plugin.

I've definitely thought about splitting the repo into one plugin per editor. There are benefits of keeping it all contained (easier to test, no need to duplicate documentation, etc). It will likely happen some day, but until I get more people to use the plugin I'll probably keep it in one repo. I also want to stabillize the tool before splitting too. 

Thanks for the help!

PR #5105, opened 2016-01-06
Ascii Graph Plug-in

divinites:
A wrap for "graph-easy", making ascii graphs with a simple DSL.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -1213,6 +1213,18 @@
 			]
 		},
 		{
+			"name": "AsciiGraph",
+			"details": "https://github.com/divinites/AsciiGraph",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Your `ascii_graph_region` command seems to use a weird method for providing stdin to the process you want to spawn. You can provide a string to `Popen.communicate`, which will then be sent to the process's stdin.

The method will also most likely fail on Windows because it would require `shell=True`.

divinites:
Thanks for your comment! I would like to provide windows support in the future, but at this moment I  windows environment, but your suggestion make sense, I will change ascii_graph_region in the next version.

PR #5106, opened 2016-01-07
Add Exclude Paths package

kemayo:


FichteFoll:
Couple tips:

1. Don't just replace "." as a path but use `os.path.normpath(os.path.join(os.path.dirname(project_file_name), project_path))` instead (`project_path` should really be named `folder_path` for less confusion).
2. `sidebar_selection` will be detected as a command by Sublime Text's plugin collector and attempting to run it will fail with an AttributeError. To prevent this, you can use multi-inheritance and use the `SidebarSelectionCommand` class as a mixin.
3. Hard-deleting the folders' exclude patterns will also remove potentially user-configured entries. A way to work around that is to additionally save the plugin-inserted values in a different key next to those settings and only erase the patterns in that setting from the actual setting interpreted by Sublime Text.
4. If you are defining a `WindowCommand`, as is the proper way for this kind of plugin, access the appropriate window from `self.window` and do not use `sublime.active_window()`. With the former approach, the commands can still be used if the window it is used on is not active.
5. There's a debug print in your code.

kemayo:
I agree with all of those but 3, and have adjusted the plugin accordingly. I disagree with 3 because I think there's no particular point in distinguishing between the two types.

*Really* this entire plugin isn't particularly focused on projects which actually have a project-file you could edit. It's conceptually more aimed at "I opened a directory in `subl` and want to tweak what's included a bit".

FichteFoll:
>I disagree with 3 because I think there's no particular point in distinguishing between the two types.

That's okay.

Re 5: You added a new debug print while working on 1 ;)

Re 2: Sublime Text detects commands by interspection of the modules attributes (i.e. globals) and checking classes whether they are subclasses of any of the `*Command` classes that it provides. The Class name is only used for naming the command and the trailing "Command" can actually be omitted.
The only three ways to prevent `sidebar_selection` from being picked up are:

1. Not subclass `sublime_plugin.WindowCommand` and use the class as a mixin instead (as I mentioned earlier). (recommended)
2. Define all commands in a different file that is not a Python file on the package root level (i.e. a submodule in a subdirectory) and only import the commands in a file exposes these commands to Sublime Text's module loader.
3. Use `del SidebarSelection` before the end of the file (not recommended).

PR #5107, opened 2016-01-07
Add GLSL Compiler Package

sindney:
Adds a .glsl file build-system, build glsl code with shipped binary, and print errors if there's any.

It will search for keywords "vs", "vertex", "fs", "fragment", "ps", "pixel" to determine if it's vertex shader or not.

If no key words were found. It will pass #define VERTEX_SHADER / FRAGMENT_SHADER and compile the same shader twice.

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -645,7 +645,23 @@
 					"tags": true
 				}
 			]
-		},
+		}, 
+		{
+			"name": "GLSL Compiler", 
+			"details": "https://github.com/sindney/GLSLCompiler",
+			"releases": [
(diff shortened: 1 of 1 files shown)


FichteFoll:
You need to add a `.no-sublime-package` file to both your branches because you are executing bundled files and ST3 will, by default, have the package in archived `.sublime-package` state, meaning the OS won't find the files.

You should probably use forwards slashes in the build system for OSX too.

sindney:
@FichteFoll I'v fixed those issues, thanks.

PR #5109, opened 2016-01-07
Add the UnicodeCharacterInsert package

Sorbing:
**Sublime Text 3 Plugin for insert Unicode Character**

Press <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>U-I</kbd> to open popup panel with Unicode Characters table.

Example screen:
The Unicode Characters popup window

### DIFF
--- a/repository/u.json
+++ b/repository/u.json
@@ -151,6 +151,16 @@
 			]
 		},
 		{
+			"name": "Unicode Character Insert",
+			"details": "https://github.com/Sorbing/sublime-unicode-character-insert",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "master"
+				}
+			]
+		},
+		{
 			"name": "Unicode Escape",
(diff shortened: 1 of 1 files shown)


FichteFoll:

The package won't work on ST3. Because packages are installed in an archived state there by default and you need to access package file contents using the `sublime.load_resource` API, which is only available to ST3.

Sorbing:
Thanks for the tip. I fix inclusion the html resource through `sublime.load_resource`. Tell me please how to properly create a tag? At first I create a lightweight tag `v0.0.1`. And just now (after fix `load_resource`) I create a new tag `v0.0.2` ([see releases](https://github.com/Sorbing/sublime-unicode-character-insert/releases)). What did I do wrong? Thanks)

FichteFoll:
Looks all right to me.

princemaple:
interesting

PR #5112, opened 2016-01-08
add completions for cocos2dx

wshxbqq:
add code auto-complete for cocos2dx

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1233,17 +1233,6 @@
 			]
 		},
 		{
-			"name": "cocos2dx js api",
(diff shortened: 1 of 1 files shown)


FichteFoll:

wshxbqq:
set tags true and labels completions

FichteFoll:

You should not include `package-metadata.json` in your repo. This file is will be automatically created by Package Control on installation.

wshxbqq:
remove package-metadata.json
Semantic versions consist of exactly three numeric parts.

PR #5116, opened 2016-01-10
plugin add Open URL Here

dkavraal:
Sublime Text plugin to open URLs inside the editor. To mention, the URL is downloaded into a temporary file to be able to open. So saving the file after an edit inside Sublime Text practically does not affect the remote file.

### DIFF
--- a/repository/o.json
+++ b/repository/o.json
@@ -433,6 +433,16 @@
 			]
 		},
 		{
+			"name": "Open URL Here",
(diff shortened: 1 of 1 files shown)


FichteFoll:
https://packagecontrol.io/packages/Open-Include does this already, and more.

dkavraal:
Thanks. Couldnot have found that so written myself. That is really much better worked out.

PR #5117, opened 2016-01-10
Rename plugin and repository URL

adashko:


FichteFoll:

adashko:
previous_names was added

PR #5119, opened 2016-01-11
Add Materialize-White-Panels

saadq:


FichteFoll:
???

saadq:
Did I do something wrong? What's the problem?

FichteFoll:
1. The package url is the same as for the Materialize package (your package)
2. The package name is *very* similar to the Material Theme - White Panels package by @equinusocio
3. The package specified in `previous_names`, material-theme-white-panels, does not exist.

saadq:
Sorry for my mistakes! I misunderstood what `previous_names` was for, I renamed the repo so I thought I had to include that. The reason why the package name is so similar is because it's literally a fork of Material - Theme White Panels. Unfortunately, his White Panels is not compatible with the themes I made for [Materialize](https://github.com/saadq/Materialize), so I received a request to create a version that worked for it which is what this package is for. Does this updated pull request look better?

FichteFoll:
I see now. Thanks for the clarification.

I suppose it's not possible to make a single one for both themes?

saadq:
Well my Materialize repository is basically a collection of 20+ themes that are based on his original Material Theme. So this package basically modifies a bunch of settings on each one of these themes, and I basically had to create 20+ identical files labeled <theme-name>.sublime-theme which overwrite their respective themes. 

I suppose the only way to make a single repo for both his Material theme and my collection of themes would be to add in all of these files to his White Panels repository. Although this might still be problematic since his theme has diverged a bit from the way my themes are, so if he makes any modifications their may be compatibility issues.

FichteFoll:
I see. That is not feasable indeed.

saadq:
Thanks!

PR #5127, opened 2016-01-13
Add DocPHP: show official php document

acabin:


FichteFoll:
There is already a package names "PhpDoc". The package names are too similar. Please change your name to be more different.

You also might want to consider looking into the `Wiew.show_popup` API.

acabin:
Thanks for suggestions.
The View.show_popup is not in the official document, it hurts...

FichteFoll:
Well, yes, popup documentation is kind of lacking at the moment.

Because ST3's API is not necessarily available at load time, you need to call things like `sublime.load_settings` in a global `plugin_loaded` function, which gets called once the API is loaded. Otherwise your plugin will fail miserably.

acabin:
Loading fixed.
plugin_loaded() looks like an event but it's not.
I got confused.

PR #5134, opened 2016-01-15
Added Antescofo Package

arshiacont:
Package for Antescofo Language communication between ST2/3 and Max/Pd widely used in real-time Computer Music Composition


FichteFoll:
For potential future releases, [semantic versions](http://semver.org) consist of exactly three numeric parts.

PR #5142, opened 2016-01-18
Add yasi package

nkmathew:
Home: https://github.com/nkmathew/sublime-yasi
Integrates: https://github.com/nkmathew/yasi-sexp-indenter

### DIFF
--- a/repository/y.json
+++ b/repository/y.json
@@ -73,7 +73,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"branch": "master"
+					"tags": true
 				}
 			]
 		},
--- a/repository/y.json
+++ b/repository/y.json
@@ -67,6 +67,17 @@
 			]
 		},
 		{
+			"name": "yasi",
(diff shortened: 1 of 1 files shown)


FichteFoll:
It appears that you are importing `yasi.py` in your code. You don't actually need to add it to `sys.path`, but you need to import differently based on the ST version this is running.

On ST2, you need to do `import yasi`.
On ST3, you need to do `from . import yasi`.

If you tested the above and it failed, I'd gladly look into the details on why it is failing (because it shouldn't).

FichteFoll:
Also, You'll need to change `"branch": "master"` to `"tags": true`.

nkmathew:
Should be good i think https://github.com/nkmathew/sublime-yasi/commit/cdff27b905567e129df51c6866141928e01bfeb4

FichteFoll:
Imports would have worked without the rename if you either tested the relative import first or determined the correct import statement based on `sys.versioninfo` or `sublime.version()`, but the name makes sense, so it doesn't really matter.

PR #5150, opened 2016-01-21
Pattern Password Plugin

0rez91:
New plugin for generate password with pattern.


FichteFoll:
I suggest adding this as a feature to https://packagecontrol.io/packages/Generate%20Password

cc @iambibhas

PR #5154, opened 2016-01-23
Added UnicodeTeX

neilanderson:


FichteFoll:
Your import of `unicode_symbols` will be failing on both ST2 and ST3.

1. Your will be installed as "UnicodeTex", as is inferred from your repository name.
2. ST2 expects absolute imports from the root package level whereas on ST3 can use relative imports.

FichteFoll:
Furthermore, you'll need to change `"branch": "master"` to `"tags": true` and use [tags](http://git-scm.com/book/en/v2/Git-Basics-Tagging) to create releases and push valid [**semantic version**](http://semver.org) tags to your repo. As mentioned [in the docs](https://packagecontrol.io/docs/submitting_a_package#Step_4), we no longer accepts branch-based releases for new packages.

neilanderson:
Thanks! I think I've added tags with semantic versioning and updated u.json accordingly. I've also fixed the importing so that it's done in the right way depending on sublime.version().

FichteFoll:
I suggest using relative imports on ST3 whenever possible, but other than that everything's fine.
(`from .unicode_symbols import greek_dictionary, math_dictionary`)

PR #5156, opened 2016-01-24
add 'q KDB' package

komsit37:


FichteFoll:
1. Note that `__init__.py` will be loaded twice because of a bug (?).
2. When you include styled_popup directly and not use it as a dependeny (for whatever reason), you must not *prepend* it to `sys.path` because that will either cause all other packages to use your package or the module will have been loaded already anyway and you'll use the dependency anyway.
Either use the actual dependency or relative imports.
3. Maybe you just want to provide numpy as a dependency instead? It is the ideal use case. You would only need `sys.path` for qpython then.
4. You have several `.pyc` files in you repo.
5. In build systems `"$file_path/$file_name"` is equivalent to `"$file"`. Your env there is also purely for UNIX, so you might want to consider using platform-specific overrides: http://docs.sublimetext.info/en/latest/reference/build_systems/configuration.html#platform-specific-options
6. I suppose most of your methods in the `Settings` class would be better used as classmethods instead of staticmethods.
7. If including a generally wide-spread command like `chain`, you should consider prefixing it in a way that will not override potential user modifications to that commands.

komsit37:
Thanks for your comments. I will try to fix all of them.

komsit37:
raised https://github.com/wbond/package_control_channel/pull/5168 for adding numpy dependency in a separate package

FichteFoll:
Similarly to the numpy dependency PR, you should restrict the OS architecture in your platform specifications.

komsit37:
Fixed  all above issues in master https://github.com/komsit37/sublime-q
Let me know if you have anything else you want me to fix

komsit37:
One more task

- [ ] restrict the OS architecture in your platform specifications

PR #5161, opened 2016-01-25
RSS Plugin for Sublime 3

JaredMHall:
A simple RSS reader for Sublime Text 3

As a lover of Emacs, this has been a hard transition to Sublime Text for me. Especially hard without an RSS reader! From what I found this is the first available, I hope to make a version for Sublime Text 2 as well.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -3291,6 +3291,16 @@
 			]
 		},
 		{
+			"name": "SublimeRSS",
+			"details": "https://github.com/JaredMHall/SublimeRSS",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"details": "https://github.com/jarhart/SublimeSBT",
 			"labels": ["repl", "scala", "testing"],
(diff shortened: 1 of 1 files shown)


FichteFoll:
You can remove your `__init__.py` since it won't matter, but since it's empty it also doesn't hurt. Note that the file will get loaded twice currently, due to a bug.

Instead of modifying `sys.path`, you should be using relative imports in ST3, i.e. `from .feedparser import feedparser`. If you rename `feedparser/feedparser.py` to `feedparser/__init__.py` you can use `from . import feedparser` instead.

Because Packages are distributed in ST3 in an archived state, you will not be able to use `open` to read files. Instead, you would have to use the `sublime.load_resource` API, which is capable of handling archived and normal packages. In your case, you should consider using a `.sublime-settings` file instead with two setting keys: one with the default feedlist and one with extensions for the user. You could also oly use one setting, but then users would have to copy the default list into their user list if they only wanted to *add* feed urls.

FichteFoll:
@JaredMHall ping

PR #5164, opened 2016-01-25
Added LangDocs

tushortz:
Preview documentations for several languages right from sublime text. Supported languages are:

* CSS
* HTML
* Java
* JavaScript
* Matlab
* Plain text (Dictionary)
* Python
* Ruby

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -24,7 +24,7 @@
 		{
 			"name": "LangDocs",
 			"details": "https://github.com/tushortz/LangDocs",
-			"labels": ["CSS", "css", "docs", "documentation", "html", "java", "javascript", "js", "language", "matlab", "python", "ruby", "snippets", "st3", "sublime", "text", "utilities",],
+			"labels": ["CSS", "css", "docs", "documentation", "html", "java", "javascript", "js", "language", "matlab", "python", "ruby", "snippets", "st3", "sublime", "text", "utilities"],
 			"releases": [
 				{
 					"sublime_text": "*",
--- a/repository/l.json
+++ b/repository/l.json
@@ -22,6 +22,17 @@
 			]
 		},
 		{
+			"name": "LangDocs",
+			"details": "https://github.com/tushortz/LangDocs",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Notable differences to other doc packages?

You should probably take a look at `View.show_popup`, just like https://github.com/wbond/package_control_channel/pull/5127.

Also, is there a particular reason why you are including `.no-sublime-package`?

tushortz:
No. I just copied some necessary files for quick start from some of my
previous work. I'll update it.

tushortz:
@FichteFoll I could update the package to the show pop up thing but i guess it wouldn't be compatible with sublime text 2. Is that right and where can I find the documentation or a great example implementing these features

FichteFoll:
@tushortz that is correct, this API has been added around build number 3075. You could check out the other package to see how they handle it.

You have basically three options: 1. popups, 2. output panel and 3. message dialog (what you currently use). You could this configurable or select one for the user based on availability.

However, what I'm more concerned about is my first question. If you need a list for compariosn, start with this: https://packagecontrol.io/search/doc

tushortz:
@FichteFoll . I've seen some examples of how the popup is used and its really cool. especially using stylesheets to manipulate its appearance.

One problem though, is that I am implementing it for Sublime Text 3 and I want Sublime text 2 to use the Message Dialogue instead.


For the lists of Docs, I must be honest, there are lots. But some of the ones I have seen are for writing doc comments. Some display documentation in a new window, some however actually opens a web browser to view the docs. Some does something similar to mine and the list goes on.

I don't want to justify mine but mine is different in a number of ways and supports soome common languages. I'm also making the update atm to use popups.

FichteFoll:
>I've seen some examples of how the popup is used and its really cool. especially using stylesheets to manipulate its appearance.

You should totally check out the [mdpopups](https://github.com/facelessuser/sublime-markdown-popups) Package Control dependency.

>One problem though, is that I am implementing it for Sublime Text 3 and I want Sublime text 2 to use the Message Dialogue instead.

I don't see a problem there. Conditional statements are a very basic programming concept.

>I don't want to justify mine but mine is different in a number of ways and supports soome common languages. I'm also making the update atm to use popups.

That's okay. I just wanted to hear your view on this after I specifically confronted you with it. I draw my own conclusions from that.

PR #5165, opened 2016-01-26
Added Python Completions

tushortz:
Extended Python completion features for Sublime text with extra imports made easy

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -1509,6 +1509,17 @@
 			]
 		},
 		{
+			"name": "More Python Completions",
+			"details": "https://github.com/tushortz/More-Python-Completions",
+			"labels": ["completions", "completion", "Completion", "python"],
+			"releases": [
+				{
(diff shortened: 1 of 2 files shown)


FichteFoll:
There is also this package named "Python Auto-Complete", although it doesn't actually provide *auto-completion*, but it only includes completions for the stdlib module names, not their attributes. Thus it has less noise. I'd prefer if that package was named "Python Completions" and yours "More Python Completions" or similar, which more precisely represents their scopes. What do you think?

cc @eliquious

Btw, I personally wouldn't prefer your package over actual auto-completion, but there may be people who like this.

eliquious:
@FichteFoll I'm ok with renaming my plugin to something more accurate.

FichteFoll:
Thanks, @eliquious, done via e110ea43ec29e1c17e152730e628c97f6a59dc93

PR #5166, opened 2016-01-26
Add Gemma repository

skinnyfit:
Adds Gemma repository for easily creating Gemfile entries


FichteFoll:
Regarding ST2-compatability: There isn't anything obvious that would prevent it from working, but iirc the API is not thread-safe in ST2 so that might be a show-breaker.

PR #5167, opened 2016-01-26
RDate Editor

Rivfadersp:
Helps to edit dates in russian language.

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -340,6 +340,17 @@
 			]
 		},
 		{
+			"name": "RDate Editor",
+			"details": "https://bitbucket.org/SergeyEg/date-editor",
+			"releases": [
+				{
+					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


FichteFoll:

Ctrl+Shift+H is a default key binding. Please either choose a different binding or don't define a binding at all and instead offer a menu item and command palette entry (and suggest a binding in your readme).

Rivfadersp:
Thank you for your remarks! I just learning to program, so hope that i did everything correctly this time.

FichteFoll:
You will know when you successfully pushed a tag to the repository when it appears [here](https://bitbucket.org/SergeyEg/date-editor#tags) in that list. There are online resources on how to do this, for example the one I linked.

I'm pretty sure you can not trigger a key binding like `"ctrl+c+d"`- at least I can't.

Rivfadersp:
Done.

PR #5168, opened 2016-01-27
add numpy dependency

komsit37:
numpy 1.8.1 built for python 3.3 is required for my other plugin https://github.com/komsit37/sublime-q.
Original PR https://github.com/wbond/package_control_channel/pull/5156

### DIFF
--- a/repository/dependencies.json
+++ b/repository/dependencies.json
@@ -200,7 +200,7 @@
 					"base": "https://github.com/komsit37/sublime-numpy",
 					"tags": true,
 					"sublime_text": ">=3000",
-					"platforms": ["windows", "osx"]
+					"platforms": ["windows-x64", "osx-x64"]
 				}
 			]
 		},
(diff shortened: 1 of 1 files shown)


FichteFoll:
Since you only provide the binaries for 64bit platforms, you should restrict the availability to those. Refer to the `lxml` dependency.

Your tag `python3.3-numpy1.8.1` is not a valid semver and will not get picked up by Package Control. You can specify the prefix you used, however, like so: `"tags": "python3.3-numpy"` instead of just setting `true`.

FichteFoll:
Furthermore, you need to place the actual numpy module in a `numpy` sub-folder under `stver_platform_arch` because that folder will be added to `sys.path`.

komsit37:
Updated the following. let me know if you found anything else. Thanks!
1. created numpy sub folder i.e. https://github.com/komsit37/sublime-numpy/tree/master/st3_osx_x64/numpy
2. created a new release tag https://github.com/komsit37/sublime-numpy/releases/tag/1.8.1
3. updated platforms to 64bit only

PR #5169, opened 2016-01-27
Add ESpec syntax highlight package

retgoat:
ESpec is a BDD framework for Elixir.
ESpec based on awesome SublimeText/RSpec but has it's own snippets related espesially to ESpec.

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -705,6 +705,17 @@
 			]
 		},
 		{
+			"name": "ESpec syntax highlighting",
+			"details": "https://github.com/retgoat/ESpec",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": ["osx", "linux"],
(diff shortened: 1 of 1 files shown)


FichteFoll:
Instead of absolute imports like `from ESpec import shared` you should use relative `from . import shared`. Note that neither of these work on ST2, where you would have to write `import shared` instead. Also note that all plugins in ST2 share the same global scope in regards to module importing, so any module with the same name as `shared` could cause the setup to get weird quickly.

Either restrict the package to ST3 only or add the two import flavors to each file, depending on Python/ST version, and rename the `shared` module to something more unique, e.g. by prefixing with a namespace.

retgoat:
@FichteFoll,
many thanks for your comments.
Actually I'm not a Python developer, but made the version dependent import as you adviced.
Please have a look.

PR #5172, opened 2016-01-28
Add 	"Django Lookup Snippets"

icycandle:


FichteFoll:

icycandle:
Thanks for reminding, tags has pushed now. : )

PR #5177, opened 2016-01-28
Added OmnisStudioIndenter and EdifactHighlighter

Frogli:


Frogli:
Added another plugin for Sublime Text. Test in Sublime Text via the ChannelRepositoryTools are OK, but I can not create a new pull request and the Travis integration tool is not fire up. Is that all right?

FichteFoll:
Pull requests are based on branches. You can create a pull request from each branch. You can then append commits to existing pull requests by pushing onto the branch you opened it from.

The url `https://sublime.wbond.net` is deprecated. Use `https://packagecontrol.io` instead.

What you are doing in the `omnisstudioindenter` command can also be done by setting metadata: http://docs.sublimetext.info/en/latest/reference/metadata.html and then utilizing the `reindent` command.

Frogli:
Hi Fichtenoll,

Thank you for your answer. I had looked in to this, but somehow did not find this page. I will look into this again!

Grtz Henk

Frogli:
One more question: Where do I point to https://sublime.wbond.net? I am not aware that I mention this url somewhere 😊

FichteFoll:
In your readme.

Frogli:
Ah yeah thanks! Fixed!

PR #5182, opened 2016-01-29
Added FreckleTimer

andela-bkagia:
This plugin allows you to manage freckle timers.

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -1188,6 +1188,17 @@
 			]
 		},
 		{
+			"name": "FreckleTimer",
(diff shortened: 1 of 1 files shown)


FichteFoll:
super+f is already bound to the find panel. Please do not override default bindings.

andela-bkagia:
@FichteFoll I've fixed the keyboard issue, but unforunately I've come up against a blocker [#953](https://github.com/SublimeTextIssues/Core/issues/953). Freckle recently swicthed to **TLSv1.2+** and Sublime has not been updated to support that.

I might try again with a pure python TLS implementation, but for now I'll close this PR.

PR #5183, opened 2016-01-29
Delphi API

JeisonJHA:


FichteFoll:
ctrl+shift+up and ctrl+leftclick are both commands that already exist in the default bindings and I would be really upset about a package overriding them. Please suggest key bindings in your readme or at least choose free (default) bindings instead of forcing them onto your users.

JeisonJHA:
I used these shortcuts to be similar to Delphi IDE. My intention is that those who use it don't feel the impact when using the ST instead of Delphi.

FichteFoll:
Well, we don't have set policies on whether or not packages may override default bindings, because user experience is usually very horrible when suddenly a commonly used binding has been overridden and the user is left completely clueless why and has no idea how to fix it, unless he is an advanced user.
Mentioning these in the readme *could* suffice for this case. You need to make them stand out more, however.

JeisonJHA:
I had changed the readme to warning the users that there are shortcuts that override ST bindings.

FichteFoll:
Make the warnings bold, please.

PR #5184, opened 2016-01-29
Register RailsGoToSass package

xijo:
A package to switch between views and sass files in a rails project

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -182,6 +182,16 @@
 			]
 		},
 		{
+			"name": "RailsGoToSass",
+			"details": "https://github.com/xijo/sublime_rails_go_to_sass",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "RailsGoToSpec",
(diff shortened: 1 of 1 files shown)


xijo:
A package to switch between views and sass files in a rails project.

FichteFoll:
This is a bit similar to https://packagecontrol.io/packages/Rails%20Related%20Files but for a more specific use case.

xijo:
Hello @FichteFoll,

thanks for your response. Indeed, the Rails Related Files is similar, but I really wanted a plugin that focusses on those specific sass files, within large rails codebases having seeing all related files can be quite chaotic.

What do you mean with change `"branch": "master"` to `"tags": true`? I created a 0.1.0 tag but can't yet change or create a new PR "for" this tag. The documentation for `Step 8` is not very specific regarding the creation of the PR.

Thanks for your help and your work on the package control!

xijo:
@FichteFoll nevermind, I found the `tags: true` flag and added it. `v0.1.0` tag is created as well.

FichteFoll:
@xijo it appears you didn't push your changes to the PR.

PR #5185, opened 2016-01-30
Added Minify on Save package

eltonvs:
Added Minify on Save package


FichteFoll:

eltonvs:
@FichteFoll thanks for the feedback.
I haven't seen this thing about tags instead branch, sorry.
I've updated this, you can take a look again, please?

PR #5186, opened 2016-01-30
Added LangDocs

tushortz:
Preview documentations for several languages right from sublime text. Supported languages are:

CSS
HTML
Java
JavaScript
Matlab
Plain text (Dictionary)
Python
Ruby

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -22,15 +22,16 @@
 			]
 		},
 		{
- 			"name": "LangDocs",
-  			"details": "https://github.com/tushortz/LangDocs",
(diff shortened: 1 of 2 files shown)


tushortz:
I've also added pygame completions in my p.json. This is to prevent recreating a pull request @FichteFoll  separately

FichteFoll:
I would have preferred if you had updated the other PR (#5164) with LangDocs and opened a new one for pygame, but this works.

You need to push a tag to pygame completions.

tushortz:
Oh, no worries. Maybe next time. And I've added the tags for the pygame completion

PR #5192, opened 2016-02-02
Add Franca IDL syntax highlighter

arthurfait:


FichteFoll:
Note that [semantic versions](http://semver.org) consist of exactly three numeric parts, so your `v0.1` tag will not be found, but `v0.0.1` will.

PR #5195, opened 2016-02-03
C11 syntax definition package added

petervaro:
A modern and 100% strictly standard conforming C language definition package added

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -109,6 +109,17 @@
 			]
 		},
 		{
+			"name": "C11",
+			"details": "https://github.com/petervaro/C11",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "C99",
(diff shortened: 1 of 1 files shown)


FichteFoll:
There have been a few changes to the C syntax definition that ships with Sublime Text by default recently. Can you compare the differences to the current C syntax and see if this addition is still needed? Alternatively, I suggest adding [upstream](https://github.com/sublimehq/Packages/pulls).

FichteFoll:
@petervaro ping

petervaro:
@FichteFoll sorry for being disappeared, I'm here now ;)

If you look at my other packages as well which I'm maintining, you could see, that all my syntax definitions are extremely detailed and strict, and are all built from scratch, to be efficient and easy to maintain. Unfortunately, that's not the case with the original versions, nor most of the already available definitions out there.

I don't want to add this to the upstream, because it is strictly C11 specific, contains tons of stuffs that are only available in this standard -- which is still not as popular as C89 and C99. Having a *generic* (though not so precise) C definition on its own is very likely a befenit.

PR #5201, opened 2016-02-05
Deviot package for IoT development

gepd:
[Deviot](https://github.com/gepd/Deviot) is a plugin based in stino for Sublime Text 2 & 3 to the IoT development using [PlatformIO](http://platformio.org/) ecosystem and supporting **~200** boards!

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -402,7 +402,6 @@
 		{
 			"name": "Deviot",
 			"details": "https://github.com/gepd/Deviot",
-			"readme": "https://github.com/gepd/Deviot/blob/master/README.md",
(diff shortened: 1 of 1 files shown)


FichteFoll:
You don't have a single tag that PC will recognize as available by default.

By default, pre-release versions are not installed by Package Control. This has to be enabled in its settings before attemption to install your package. I suggest releasing an non-prerelease version.

Your only other tag is not a valid [semantic version](http://semver.org) tag.

gepd:
I've changed the first version to v1.0.0 according to the semantic version page.

PR #5205, opened 2016-02-05
Added tags to sublime-xpath.

rosshadden:


FichteFoll:
I should note that searching on the package control site only scans the package name and its description, not its tags.

This could be added as a feature though, couldn't it, @wbond?

rosshadden:
Oh.... thank you for pointing that out.  I guess it was obvious that I was adding the tags for this purpose!

PR #5207, opened 2016-02-06
Add Libsass Build and Sassc dependency

blitzrk:


blitzrk:
Just fixed some problems and squashed the intermediate commits. Now all tests are passing.

FichteFoll:
You don't seem to have a particular need for `.no-sublime-package`, please remove it.

Also please use relative imports for ST3 and absolute imports for root-level modules for ST2 instead of modifying `sys.path`.

blitzrk:
Removed `.no-sublime-package`.

Removed dependency from `l.json` so it is now only in `dependencies.json` in my plugin's repo.

I think I fixed the relative imports for both ST2 and 3, but I can't test it right now, because I hit Github's API's rate limit. Will test ASAP.

Good enough to merge?

FichteFoll:
Python files in the root folder look good, but you have to use relative imports in your `libsass` package as well in order to have it work on ST3. Since this is a sub-package, relative imports will work on ST2 too.

blitzrk:
Fixed and tested on 2221 and 3103. Thank you for your help.

I'm going to start drafting a Python 2/3 compatibility guide for packagecontrol.io's developer documentation. I'll mention what I've learned here and some other compatibility tidbits like the fact that the print function is automatically imported from __future__.

FichteFoll:
We also have that planned for the unofficial documentation, but there just is not enough time for us.

PR #5211, opened 2016-02-08
add Atomizr and deps

idleberg:


FichteFoll:
I suggest specifying dependencies only in one place (`dependencies.json`) to avoid confusion. You'll have to add speg to that file in that case.

idleberg:
I created `.no-sublime-package` because I had some troubles with the package without it, but maybe that was related to the dependencies. I made both changes you suggested, let's see how this works.

FichteFoll:
>I made both changes you suggested, let's see how this works.

It appears you didn't push your changes to the PR though.

idleberg:
Not sure what this PR has to do with it. I [removed](https://github.com/idleberg/sublime-atomizr/commit/e22fe4852266d344cc03961e4cb65918be325710) the `.no-sublime-package` file and [updated](https://github.com/idleberg/sublime-atomizr/commit/800fbd512be3b0b206c8fd37638f67f81c85a9ea) the dependencies. What else is there to do?

FichteFoll:
You have specified dependencies in the channel itself with this PR but also have them in your repository itself. One (channel) will take priority and will leave you wondering why changing your `dependencies.json` does not work, so I suggest removing them here.

idleberg:
Hope everything is fine this time

PR #5215, opened 2016-02-09
adding block alphabetizer

jasongornall:
A really simple block alphabetizer that lets you select multiple blocks and then alphabetize just those blocks

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -372,6 +372,16 @@
 			]
 		},
 		{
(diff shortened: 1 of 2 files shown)


gepd:
you should add your package in the `m.json` file instead of `a.json`

Look the full instructions to submit a package [here](https://packagecontrol.io/docs/submitting_a_package)

jasongornall:
ahhh thankyou

PR #5219, opened 2016-02-09
add MarkdownHiglighting package

braver:


FichteFoll:
What is the purpose of this package exactly?

braver:
Most scopes in the Markdown syntax are specific to that language (e.g. `quote`, `heading`). None of the default color schemes and none but a handful of the community schemes do any sort of highlighting for those scopes. This package adds common keywords like `function` and `variable` to some of those scopes. This is semantically incorrect, so I won't propose the original repository to follow this example. However, it does enhance Markdown files in color schemes that don't take that language into account, as illustrated in the [before/after screenshot](https://github.com/braver/MarkdownHighlighting#markdownhighlighting)

PR #5223, opened 2016-02-10
add LessImproved

braver:


FichteFoll:
Is there a specific reason why you didn't propose this as a pull request to @danro's original?

braver:
- The original is designed around pattern recognition, my fork uses keyword lists.
- It also uses some sort of build/make process that makes it hard to contribute.
- The fork is also compatible only with the latest Sublime 3 builds. 
- The original repo is stale and the author hasn't chimed in on [simple feature requests](https://github.com/danro/LESS-sublime/issues/72), [PRs for improvements](https://github.com/danro/LESS-sublime/pulls) or discussions of the [issues resulting from it's design](https://github.com/danro/LESS-sublime/issues/77).

So, yes this was originally forked from his work, but I doubt he would welcome a PR that changes *everything*, or would even respond to it. It is however a big improvement over the original because of the change to keyword lists (and the fact I'm currently actively maintaining it).

I applaud you for looking out for packagecontrol and agree fragmentation should be limited. However, I believe this moves Less support forward for Sublime 3 and Sublime 2 can keep the original, so it's perhaps not so bad.

FichteFoll:
It appears to be true that the package is hardly maintained, so I agree with your reasoning.

braver:
Awesome, thanks :+1:

danro:
Actually, had you asked @braver, I would have been open to a larger PR (possibly even giving you push access), although I must admit I am unclear as to how package control deals with v2 legacy support. @FichteFoll can you point to any docs explaining sublime v2/v3 support?

braver:
@danro Hey, that's totally cool, sorry for making assumptions and not getting in touch. I'm definitely for contributing and cooperating, but I would be barging in with a PR that basically says "hey I changed everything, how do you like this", so I chose not to do that. So, sorry about that.

If I understand package control well enough, the non-v2-compatible version would have to be a separate entry in the channel that points to a different branch or repo. The existing entry should stay so it doesn't break anything for anyone, but it could be changed to point to a different branch. But it doesn't seem like you could easily pull existing installs towards the new non-v2 version, so you might end up with two repos anyway?

We could also keep it as it stands right now, and I'll add you to the new repo and I am totally not opposed to for instance weeding out the issues and pr's in your repo. @danro, I think it's your call as for what to do with this, I'm totally fine with everything.

FichteFoll:
@danro https://github.com/wbond/package_control/blob/5643f4fa9a335a4591482a8ecdf6e0d7413fbaa1/example-repository.json#L260-L288

That said, since this is *probably* a complete rewrite of the syntax, many changes might be included as well and as such keeping the new version as a separate package makes sense for users that want to keep the current behavior. The rewrite also only works for builds >=3092 (technically 3084, but pretty much unusable, so the proper build with initial support would be 3092), so everyone with an older build would need the old syntax anyway. They can co-exist in the same location and the .sublime-syntax file would be prioritized.

The only downside is probably discoverability here, in that users of the LESS package will most likely not look for "LessImproved" even though it appears to be better. Maybe you can work something out together so that all users get the best of it. I suggest just adding the .sublime-syntax file to the LESS package and evaluating the differences between them.

braver:
Well, the changes probably aren't even that apparent. I didn't change the basic structure or scopes, it just solves a bunch of issues by approaching some problems differently. End users shouldn't notice it, not in the way the JS changes have been "noticed" recently.

I could do a pull request that adds the new .sublime-syntax file (and additions to the completions) so anyone who uses the latest beta build gets the improvement. It seems like you could silently deprecate the old syntax and anyone who has issues with it can move to a newer version of Sublime and reap the benefits.

braver:
Anyway, I opened a PR and we'll take it from there. Thanks @FichteFoll for your input and information!

PR #5226, opened 2016-02-10
Add Tron Color Scheme

bcomnes:
Added a standalone version of the Tron and Tron Legacy color schemes.

https://github.com/bcomnes/sublime-tron-color-scheme


FichteFoll:
The proper color scheme tag is `"color scheme"`.

bcomnes:
Fixed

PR #5239, opened 2016-02-11
add ImagePaste

robinchenyu:
add ImagePaste Package to paste image from clipboard and save image to a file with inserting file path.


blitzrk:
FYI this commit is 3285 changes behind master and follows the old schema (now use `"tags": true` in "releases").

robinchenyu:
merge fixed

PR #5244, opened 2016-02-12
Changes in package

akalongman:


FichteFoll:
While I am reviewing this, please remove your `.no-sublime-package` file from the repo. It is not necessary and instead complicates things for users.

akalongman:
```.no-sublime-package``` not supported anymore?

FichteFoll:
It is supported, but it is not necessary in your package.

akalongman:
Ok. I will test and after remove that file. Thank you

PR #5246, opened 2016-02-12
added laravel-helper-completions

james2doyle:


FichteFoll:
How is this different from the other [laravel packages](https://packagecontrol.io/search/laravel), notably the "Laravel 5 Snippets"?

james2doyle:
There is no overlap between this package and those Laravel snippets packages. There are a couple in one of the Blade packages, but they have a different scope and also inject `{{ }}` template tags. 

Those snippets packages simply provide completions for classes. This package is for the helper functions that are included in all Laravel files. The helpers often change at a different rate than the class names, those tend to remain unchanged. 

I think it makes sense to split them up, because depending on your style, you may like calling classes more. 

Does that clarify?

Also, I can make those changes as well.

FichteFoll:
Okay, thanks for the explanation.

We can merge this once you changed to tag-based releases.

james2doyle:
This should be good now I think?

PR #5247, opened 2016-02-13
adding Patch Apply plugin

kkujawinski:


FichteFoll:

kkujawinski:
I've changed versioning.

PR #5248, opened 2016-02-14
add a new package named CUDA cnippets

LitLeo:
This package adds some CUDA functions snippets for Sublime Text.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3159,6 +3159,17 @@
 			]
 		},
 		{
+			"name": "CUDA Snippets",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Your `generate_snippets.py` file will be detected as a plugin file and loaded by Sublime Text automatically. Because this is certainly not what you want (and will most likely fail on ST3 anyway), please prevent this from happening by either moving it to a sub-directory or wrapping the file's action in an `if __name__ == "__main__":` block.

FichteFoll:

LitLeo:
I have deleted generate_snippets.py file and add a valid semantic version tag in my repo, please recheck it. Thanks.

PR #5251, opened 2016-02-14
added tomorrow night italics color scheme

kennethreitz:
:sparkles: :cake: :sparkles:

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -1929,6 +1929,17 @@
 			]
 		},
 		{
+			"name": "Tomorrow Night Italics Color Scheme",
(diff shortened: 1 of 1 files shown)


FichteFoll:

wbond:
@kennethreitz We use a standard versioning scheme so that we can derive some meaning from versions and thus allow users to opt-in to pre-release versions, and hopefully in the future prompt users for major breaking changes in a package.

Since we use git tags, hopefully it shouldn't really be too much work to just create a new tag whenever changes are ready for users to consume.

FichteFoll:
@kennethreitz ping

PR #5252, opened 2016-02-14
Add Riemann autocomplete plugin

jarretth:
Generated from http://riemann.io/api.html


FichteFoll:

jarretth:
Updated, thanks

PR #5255, opened 2016-02-15
Add AVA

sindresorhus:
https://github.com/sindresorhus/sublime-ava

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -1716,6 +1716,17 @@
 			]
 		},
 		{
+			"name": "AVA",
+			"details": "https://github.com/sindresorhus/sublime-ava",
+			"labels": ["snippets", "javascript", "nodejs", "testing", "auto-complete", "completions"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "AVM2 Pcode Syntax Highlighting",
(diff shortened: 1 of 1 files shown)


FichteFoll:
You'll need to change `"branch": "master"` to `"tags": true` and use [tags](http://git-scm.com/book/en/v2/Git-Basics-Tagging) to create releases and push valid [**semantic version**](http://semver.org) tags to your repo. As mentioned [in the docs](https://packagecontrol.io/docs/submitting_a_package#Step_4), we no longer accept branch-based releases for new packages.

FichteFoll:
Also, I can say that I would prefer the name "AVA Snippets", as used in #5282, since that more approproately reflects what the package provides. Unless you intend to add more features of course (although I'd question which).

sindresorhus:
> You'll need to change "branch": "master" to "tags": true and use tags to create releases and push valid semantic version tags to your repo.

Sorry, missed that. Fixed now.

> Also, I can say that I would prefer the name "AVA Snippets", as used in #5282, since that more approproately reflects what the package provides. Unless you intend to add more features of course.

We plan on adding more things beyond snippets very soon, like autocompletion and ability to run tests directly from Sublime. So I prefer `AVA`, so I don't have to come back here soon again and rename it.

GitScrum:
👏👏👏

PR #5260, opened 2016-02-16
add otto theme

oivva:


FichteFoll:
Be careful with providing syntax definitions that only include other syntaxes. tmLanguage syntax definitions can not include sublime-syntax ones, meaning your package breaks highlighting for the syntaxes you attempt to include starting around build 3092, which is roughly where all default packages' syntax definitions were converted. You can prevent this by additionally providing a `.sublime-syntax` variant, which has higher priority than `.tmLanguage` and *can* include other `.sublime-syntax` files. See also https://github.com/lefoy/cyanide-theme/issues/40.

It might cause problems in combinations with other themes doing the same, too.

oivva:
Thank you for your clear and helpful explanation. I removed `.tmLanguage` files.

PR #5261, opened 2016-02-16
Add Multi line paste package

shuky19:


FichteFoll:
I find the readme to be very much non-descriptive and could only figure out the package's functionality by reading the source code. Please do something about that.

PR #5272, opened 2016-02-21
Added SQLTools repository

mtxr:


FichteFoll:
Currently your package only works on ST3 due to your usage of relative imports. I suggest to update the version selector to reflect that.

mtxr:
Just fixed it.

I want to support ST2 but I need to run more test when inserting data to a view.
So, for now, I'm supporting just ST3 or newer.

Thanks.

PR #5284, opened 2016-02-26
Remove MinimalFortran package (deprecated)

315234:
This package is no longer maintained

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -1076,15 +1076,6 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
You should consider shooting an update message to your users or potentially migrating the old MinimalFortran into a different branch of the Fortran package, which will only be installed for ST versions <3092.

315234:
I closed the pull request to stop this from happening in the end. I'm going to think about how to start migrating them over, I didn't realise so many people were still using v2.

PR #5285, opened 2016-02-27
Add Materialize-Appbar

saadq:


saadq:
In case there is any confusion, Materialize-Appbar is a fork of equinusocio's Material-Appbar, but this new one is compatible with all of the themes in [Materialize](https://github.com/saadq/Materialize).

FichteFoll:
Why does this package have a `.no-sublime-package` file?

cc @equinusocio

saadq:
I pretty much just followed the way that equinusocio did it, and I'm not too sure but I think it's because this add-on to the themes is just used to overwrite a few of the settings on the theme files and would need to go into the `Packages` folder to do so (this is my guess).

equinusocio:
It's right. Like Material Theme White Panels the package overwrite some of the original theme rules.

FichteFoll:
Hm, okay. Themes are the one thing I never touched, so I have to rely on your information here.

PR #5286, opened 2016-02-28
added Easy Import to e.json

tushortz:
Allows to import almost all packages


FichteFoll:
You use relative imports in your Python code, which is incompatible with ST2. Please adjust the version selector or tweak the code.

tushortz:
I've removed support for ST2 thanks @FichteFoll

PR #5287, opened 2016-02-28
add Snippet Lister

math2001:
Hi !
This is a plugin for open snippet _from_ Sublime. You can browse your file, but it automatly open the right folder. You just need to respect logic "rules".

Hope you'll accept it.

Signed-off-by: math2001 <australie.p@gmail.com>

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1782,6 +1782,17 @@
 			]
 		},
 		{
+			"name": "Snippet Lister",
+			"details": "https://github.com/math2001/snippet-lister/",
+			"labels": ["snippets"],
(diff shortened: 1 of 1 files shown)


FichteFoll:
I don't really get the use cases for this package (compared to just using a project for your snippets or something like PackageResourceViewer), but ... to each their own. It's different at least.

You should never commit `.sublime-workspace` files to your repository.

I also don't recommend relying on `os.chdir` and family, if only because the "working dir" is something that all plugin code in ST shares.

FichteFoll:
@math2001 ping

PR #5289, opened 2016-02-29
	add UndoRedoDontMoveCursor

titoBouzout:
add UndoRedoDontMoveCursor package

### DIFF
--- a/repository/z.json
+++ b/repository/z.json
@@ -219,7 +219,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"trags": true
+					"tags": true
 				}
 			]
 		}
--- a/repository/z.json
+++ b/repository/z.json
@@ -213,7 +213,7 @@
 			]
 		},
 		{
-			"name": "Zzzz0 titoBouzout-UndoRedoDontMoveCursor",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Why the weird prefix?

titoBouzout:
I like the name for the package. There I added and pushed the tag.

FichteFoll:
Leading zeroes are invalid for semantic versions.

titoBouzout:
A pain in the ass :P Think is fixed.

wbond:
I removed this package because it has a crappy name and was using version numbers bigger than 32bit integer, thus causing issues with the site and channel.

titoBouzout:
Now what

FichteFoll:
>version numbers bigger than 32bit integer

Well, I couldn't expect that. :joy:

wbond:
Apparently there is some issue in constructing psycopg2 result objects with integers bigger than 32bits. I can't say I would have expected it either. However, it will have to wait until I have time to figure out exactly what is causing it. It isn't obvious to me from the exception and stack trace.

Until then, version numbers < 2 billion will have to do.

PR #5294, opened 2016-03-01
Added new plugin ImsiEncoder to the Default Channel

luisillo26:
Plugin that allows to convert IMSI (international mobile subscriber identity used in telco industry) to NAI format.

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -268,6 +268,16 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
There is no reason for the `ImsiEncoder.sublime-package` file to be within the repo. It will be completely ignored on ST's end due to how PC installs it.

PR #5299, opened 2016-03-02
Add HTML Typographer

einomi:
HTML typography plugin based on Lebedev Typograph web service.

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -868,6 +868,16 @@
 			]
 		},
 		{
+			"name": "HTML Typographer",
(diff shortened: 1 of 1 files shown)


FichteFoll:
https://packagecontrol.io/search/entities

einomi:
@FichteFoll I don't think that these plugins have the same functionality. My plugin also sets up correct word wrapping.
For example, `I like riding a bike`  will be transformed to `I&nbsp;like riding a&nbsp;bike`. Then `I like` and  `a bike` will be always on one line.

FichteFoll:
Hm, I see.

PR #5300, opened 2016-03-02
Add "Simutrans dat Syntax" package

An-dz:
A small package to help development for a simulation game called [Simutrans](http://www.simutrans.com/).


An-dz:
For unknown reasons, GitHub has not forked my branch correctly.

PR #5304, opened 2016-03-05
Google Chrome JavaScript API Auto-complete package added

ismnoiet:
Offer Google Chrome JavaScript API auto-complete.

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -900,6 +900,16 @@
 			]
 		},
 		{
+			"name": "Google Chrome JavaScript API Auto-complete",
+			"details": "https://github.com/ismnoiet/Google-Chrome-JavaScript-API-Auto-complete",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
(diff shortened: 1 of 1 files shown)


FichteFoll:
This repository does not offer any file that interfaces as a Sublime Text resource.

ismnoiet:
You are right @FichteFoll  i've tried to contact @Kronuz to give me more details about [SublimeCodeIntel](https://github.com/Kronuz/SublimeCodeIntel) plugin  because i really want to help with it, But apparently  he is not responding, i hope he is doing good.
The reason why i want to push it to [package control](packagecontrol.io) is that there is a lot of developers out there who are looking to see it added to sublime text tool box.

FichteFoll:
I do not agree with adding something to Package Control that doesn't use its architechture at all. You'll need to think of a different way to promote your sublimecodeintel addition or add plugin code to automate the process.

ismnoiet:
@FichteFoll  you're right i'm thinking about another way to include it, anyway thank you for your time :+1:

PR #5306, opened 2016-03-05
Add AppendNewline

jeffhandley:


FichteFoll:
From the default key bindings:

```js
{ "keys": ["ctrl+enter"], "command": "run_macro_file", "args": {"file": "res://Packages/Default/Add Line.sublime-macro"} },
```

jeffhandley:
This is different from the default binding which also moves the cursor. AppendNewline adds a new line after the current line without moving the cursor.

This has been useful for me while writing JavaScript.

FichteFoll:
I see. I also saw this from the source code, but I wanted to hear from you specifically whether this is what you value in the package.

I suggest explicitly mentioning this difference in the readme.

jeffhandley:
Very good feedback. Thank you for reviewing and merging it!

PR #5315, opened 2016-03-08
Add KSP (Kontakt Script Processor)

nojanath:


FichteFoll:
I would gate `default_syntax.py` behind a setting rather than requiring the user to overwrite the file.

You're leaking a file handle in `build/update_completions.py`.

PR #5318, opened 2016-03-08
Add Confluence, Remove Markup Jira Confluence

mlf4aiur:
The plugin "Confluence" is rewrite based on Atlassian Confluence REST API, and "Markup Jira Confluence" is deprecated.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2226,6 +2226,18 @@
 			]
 		},
 		{
+			"name": "Confluence",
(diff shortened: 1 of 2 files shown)


FichteFoll:
You could rename the package and include an update message for users, if you want.

mlf4aiur:
The new plugin is totally rewrite, APIs and configurations have changed, and it not compatible with the old one, so I don't want rename it.

Also I just tested new plugin on ST3, maybe it not works on ST2.

PR #5327, opened 2016-03-14
Add Protobuf Syntax Highlighting

VcamX:


FichteFoll:
How is this different from https://packagecontrol.io/packages/Protocol%20Buffer%20Syntax?

VcamX:
@FichteFoll This one uses [the latest mechanism](https://www.sublimetext.com/docs/3/syntax.html) for syntax highlighting, it's more accurate and powerful but only supports Sublime Text 3 (>=3103). I think it's totally different from that one although they highlight the same language.

VcamX:
@FichteFoll Comparing the new one and the old one

New:
screen shot 2016-03-16 at 17 10 07

Old:
screen shot 2016-03-16 at 17 10 01

FichteFoll:
Hm, I see.

The names are a bit too similar for my taste. Maybe you could collaborate with @vihangm on his package and provide both, the updated and the old syntax side by side? ST will prioritize the new format if two files with the same name exist. Or split release channels on ST build 3092 (which is the first to properly support the new format).

FichteFoll:
@VcamX ping

vihangm:
I'm happy to incorporate the new syntax file or split channels. @FichteFoll, do you have thoughts on which one of those you'd prefer?

VcamX:
Having consistent syntax highlighting may be good and these two solutions still have some differences. I'd like to choose splitting channels. What do you think? @FichteFoll @vihangm

FichteFoll:
@VcamX 

> Having consistent syntax highlighting may be good and these two solutions still have some differences

In that case, I suggest not splitting channels but rather add this as a separate package. I'm okay with that option if you two (or even one) think the differencs are too big for a linear migration of users.

VcamX:
@FichteFoll so what should I do next If adding it as a separate package?

FichteFoll:
Well, the package names are still quite similar but I can't think of an alternative that still conveys the necessary information.

I'll just merge this and you can create a new PR with a rename if a better name sparks your brain.

PR #5328, opened 2016-03-14
Support for the BrainwareGroup Columbus scripting language.

heartsekai:
Thank you for your job!.


FichteFoll:
`.sublime-syntax` is only available effectively since 3092. Please change that in the version selector.

heartsekai:
Ok i updated the Version.

PR #5334, opened 2016-03-17
Add EddyMalou LoremIpsum package

Metrakit:
Add https://github.com/Metrakit/eddy-malou-lorem-ipsum package repository

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -200,6 +200,16 @@
 			]
 		},
 		{
+			"name": "EddyMalou LoremIpsum",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Because packages are, by default, installed in an archived `.sublime-package` file on **ST3**, you may not use `open()` to load resources in your package. Instead, use `sublime.load_resource()`.

Metrakit:
Voila, thanks for your feedback. I have just changed my plugin now with ```sublime.load_resource()``` and precise than this plugin is only for Sublime Text 3 now. then I gave updated my current PR with ">=3000". Thanks.

Issue #5288, opened 2016-02-29
Semantic versioning requirement is nonsense


titoBouzout:
My packages include no dependencies, the requirement for semantic versions looks like non sense to me. 

You are breaking the purpose of "semantic versions" when requiring to new packages, to follow "semantic versioning" while at the same time you require ST version. Im wrong?

gepd:
I think you're misunderstanding how PC works. `semantic versions` is required to know what version is delivered to the user by the PC. At the same time it will allows the user to know what is the version of the software they are currently using and what is the last version released. The `ST version` is used to know the compatibility of your package with ST, if your package is only available for ST 2 or 3 you can specify that. Same as the S.O.

titoBouzout:
Well, these are not the reasons on why semantic versioning exists. Is not to solve the thing you described. If for something else. All of my current  pages, does not include any versioning system, and these are working just fine. I dont want to maintain versions, so this requirement makes non sense.  IIRC packages based on branches, were getting the version from the commit date, which is perfect for me. and I would like to keep it that way. Is simple and goes to the point.  

Semtnic versioning is for something else, go and read it http://semver.org/

gepd:
There not need to make a mountain out of a molehill, you can use a tag v1.0.0 always pointing at the top of your branch, I don't think it's a big problem. You can read [this](https://github.com/wbond/package_control_channel/issues/4072) to know more about this decision.

titoBouzout:
Mmh, if you don't think is a big problem, Im happy for you. I don't want to maintain versions numbers on tags for no apparent reason. To me the date is enough, which should be enough for you too to do whatever you want to do with my package. Do you understand that? Package control is capable to keep the system that originally provided and still provides. I see no problem with it. I don't want to complicate it for no gains.

You suggesting to tag the tip as 1.0.0 and push always there? So users will never received updates? that's the best thing you can come with?

titoBouzout:
For simplicity I suggest to make semantic versioning a requirement for packages that include dependencies and allow anyone that has no problem with it to use it. For everyone else just use branches as now that is working just fine.

FichteFoll:
Semantic versions are indeed meant to be used for dependencies primarily, since the concept of an "API" is very abstract for end-user tools or applications since it is not specified. That does not mean semver can't beused that way; you just need to use your own judgement there.

[...]

PR #5335, opened 2016-03-18
Remove some packages

titoBouzout:


FichteFoll:
Is this meant to be a boycott due to #5288?

titoBouzout:
Boycott is a word with strong meaning and somewhat negative connotation. 

I just don't like the new requirement for any of my newer packages, so I want all of them removed. When distribution channels change "rules" and I dont like these Im free to move. Thanks.

FichteFoll:
Well, that is pretty much exactly what a boycott is, but whatever. It's your choice.

PR #5341, opened 2016-03-18
Request to publish SwiftKitten, a new Sublime package

johncsnyder:
SwiftKitten is a Swift autocompleter for Sublime Text, via the adorable SourceKitten framework.

For more info see: https://github.com/johncsnyder/SwiftKitten

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -3836,6 +3836,17 @@
 			]
 		},
 		{
+			"name": "SwiftKitten",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Aw yeah, another Python package using absolute imports for no reason except making including it as sub-packages more complex (ijson). cffi and pypaser do it right, meaning if you laid out the folder structure properly you wouldn't need to mess with `sys.path`.

PR #5344, opened 2016-03-19
add ImageCMS.tpl

ROBsoer:
add ImageCMS.tpl syntax plugin

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -217,6 +217,17 @@
 			]
 		},
 		{
+			"name": "ImageCMS.tpl",
+			"details": "https://github.com/ROBsoer/sublime-imagecms.tpl",
+			"author": "ROBsoer",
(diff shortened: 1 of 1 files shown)


FichteFoll:
You must not use dots in package names (older packages may still have this however). Just use "ImageCMS".

[Semantic versions](http://semver.org) consist of exactly three numeric parts and there must not be a dot after the `v`.

ROBsoer:
changed plugin name

FichteFoll:
Please also address the last section of my earlier comment.

ROBsoer:
have updated relise version

FichteFoll:
I don't see anything having changed here: https://github.com/ROBsoer/sublime-imagecms.tpl/tags

FichteFoll:
@ROBsoer ping

PR #5345, opened 2016-03-20
Added CopyOnSelect Package

chrifpa:


FichteFoll:
I suggest joining multiple selections with a newline character to mirror ST's copy behavior. I also suggest storing the strings in a list and then using `"\n".join()` because it's faster.

PR #5347, opened 2016-03-20
add troubleshooting package

guillermooo:
@FichteFoll


FichteFoll:
I think the `author` field is case-sensitive, so please change that to "FichteFoll" :smile:

guillermooo:
Done 😆

PR #5348, opened 2016-03-20
Sourcepawn completions package

ppalex7:


FichteFoll:
You seem to have the entire Nil theme mixed in your repo. I don't think this was intended? (Although you have syntax-specific setting to set the Tubnil color scheme and I do personally use that color scheme as well, I doubt that everyone would appreciate you forcing them onto your favorite color scheme.)

Is there a particular reason why you included a `.no-sublime-package` file? I do not see a particular need for it (yes, `sys.path.insert(0, os.path.dirname(__file__))` should work fine, although you should probably *not* prepend the path but append).

ppalex7:
> You seem to have the entire Nil theme mixed in your repo. I don't think this was intended?
My package is a port of https://github.com/austinwagner/sublime-sourcepawn for ST3. Tubnil theme comes from original plugin. Should i remove theme, syntax-specific settings or both of this?

>  although you should probably not prepend the path but append
fixed.

> Is there a particular reason why you included a .no-sublime-package file?
I check plugin again, working fine without it. File was removed.

FichteFoll:
So it appears you indeed have the entire "SourcePawn Syntax Highlighting" package in your repo as well. I suggest to remove that entirely and instead recommend users to install that package along with yours.

@austinwagner you might want to take a look at my earlier comment.

austinwagner:
Huh yeah. Looks like the other people I gave control to the project put that stuff in there. I'll clean it up.

ppalex7:
Removed in ppalex7/SourcePawnCompletions@6900dbfba94c6566f17f98ba8c0dbe6e7ccc5c10

FichteFoll:
Thanks.

>So it appears you indeed have the entire "SourcePawn Syntax Highlighting" package in your repo as well. 
> I suggest to remove that entirely and instead recommend users to install that package along with yours.

Any comment on this?

ppalex7:
> Any comment on this?

Removed too.

FichteFoll:
Great, thanks.

PR #5349, opened 2016-03-20
Cipher Package

depaolism:
A package that includes common functions used when deobfuscating various texts.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -859,6 +859,16 @@
 			]
 		},
 		{
+			"name": "Cipher",
+			"details": "https://github.com/depaolism/Cipher",
+			"releases": [
+				{
+					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Except for string concatenation, most of these utils are already available in at least one other package.
Maybe you could add that functionality to https://packagecontrol.io/packages/StringUtilities?

cc @akalongman

akalongman:
Pull requests hightly welcomed

depaolism:
Thanks, I found various packages that performed base64 encoding/decoding when I searched through the available packages, but I didn't come across StringUtilities (bad search skills). I plan on continuing to add different string transformations as I come across them. StringUtilities may be the place for them.

FichteFoll:
Since this seems resolved, I'm closing.

PR #5350, opened 2016-03-20
Add ESLint package

polygonplanet:


FichteFoll:
`sublime.packages_path()` may not be available at the time you are invoking it and just return `None`. You need to check it in a`plugin_loaded` callback for ST3.

polygonplanet:
Thank you for review.
I've fixed to use a `plugin_loaded` for version compatibility, and change eslint settings.

PR #5351, opened 2016-03-21
Add Fantom package

mgiannini:


FichteFoll:
Packages using `.sublime-syntax` must restrict installation to ST versions of 3092 or higher.

mgiannini:
Pushed fix to require ST version 3092 or higher

PR #5353, opened 2016-03-22
Add package -- AutoSetSyntax

jfcherng:
Plugin repo: https://github.com/jfcherng/Sublime-AutoSetSyntax

Discussions on ApplySyntax: https://github.com/facelessuser/ApplySyntax/issues/97
Discussions on ST forum: https://forum.sublimetext.com/t/automatically-set-view-syntax-according-to-first-line/18629

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -1558,8 +1558,8 @@
 			]
 		},
 		{
-			"name": "AutoSetNewFileSyntax",
-			"details": "https://github.com/jfcherng/Sublime-AutoSetNewFileSyntax",
+			"name": "AutoSetSyntax",
+			"details": "https://github.com/jfcherng/Sublime-AutoSetSyntax",
 			"releases": [
 				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
You don't need to modify `sys.path` for your code. Just use a relative import, i.e. `from .SyntaxMappings import SyntaxMappings`. Besides, I suggest to move the `SyntaxMappings.py` file into a subdirectory (e.g. `SyntaxMappings/__init__.py`) because it doesn't interface with ST directly but will still be scanned for commands or event listeners.

For the yaml module, check out https://github.com/packagecontrol/pyyaml.

jfcherng:
Nice! Didn't know it can be done like that. Fixed :+1:

FichteFoll:
Oh, and another "good to know": ST3 provides a global `__package__` variable which contains the actual package name as it is stored on the hard drive.

PR #5354, opened 2016-03-23
Lipsum  for Chinese

yinxianwei:


FichteFoll:
You should not include `package-metadata.json` in your repo. This file is will be automatically created by Package Control on installation.

I would also suggest providing these as snippets instead (more transparent for user since no key bindings are overridden and snippets are exposed in both the command palette and the auto completion popup), but that's just a recommendation.

yinxianwei:
remove package-metadata.json

PR #5359, opened 2016-03-24
FileMerge for OS X based off of WinMerge plugin

thtliife:


FichteFoll:

PR #5363, opened 2016-03-25
Rename XpressEngine, change repo location

bnu:


FichteFoll:
Please refrain from using dots in package name since they are not properly supported for Python plugins. You are not using any, but eventually we want all packages not to use dots in their names.

FichteFoll:
@bnu ping

PR #5365, opened 2016-03-26
Added Simple Import Package

vini175pa:


FichteFoll:

vini175pa:
Is it ok now?

FichteFoll:
Note that there is also a `1.0.0` tag on your repo, which would get picked up as the most recent version, although it is not the most recent commit (which is tagged as `v0.1.0`).

vini175pa:
Oh, sorry. I deleted it. I didn't realize that, because I actually edited that release, but Github created another one.

PR #5368, opened 2016-03-27
Change SublimeGit from custom repo to github

miped:
SublimeGit is going open source. Going forward it will be developed on github, and released from there.

This pull request removes this SublimeGit repo, and adds it as a standard package instead.


jrappen:
@miped awesome, thanks!

PR #5371, opened 2016-03-28
Added DarkMosquito Color Scheme

Mosquito13:


FichteFoll:
FYI, I personally find the slight pink for `support.function` (I think that's what it is) to be too subtle compared to the default text color.

Mosquito13:
Thanks, I'll check this.

PR #5375, opened 2016-03-30
Added CFMLFunctionFinder

MFernstrom:


FichteFoll:

Does Goto Symbol not work for ColdFusion syntax? It would be better to make that work or use it over a custom implementation using popups imo.

MFernstrom:
I fixed the semantic versioning to use three numerics.

I had no idea that Sublime already had that built in. It seems to work just fine with CFML too, so no need to this plugin. Oh well, at least I learned new things.

jcberquist:
@FichteFoll: with the [plugin](https://github.com/jcberquist/sublimetext-cfml) I am maintaining Goto Symbol works just fine. I deliberately gave all function names a consistent scope though, whether they were written in script or in tags. 

The older ColdFusion [plugin](https://github.com/SublimeText/ColdFusion) doesn't give all function names a consistent scope, however I did check and they still show up in the Goto Symbol list. It doesn't have indexed symbols set up as far as I can tell, so Goto Symbol in Project and Goto Definition don't work, but that doesn't affect this pull request.

MFernstrom:
@FichteFoll @jcberquist I see no point in adding this plugin.

PR #5378, opened 2016-03-31
Add Groupie

noct:


FichteFoll:
There is no `plugin_loaded` defined in your plugin, meaning that the file will fail to load on ST2 as it cannot find the function.

noct:
Fixed in latest release.

PR #5384, opened 2016-04-03
Adding Switch Settings Package

t8m8:
This Sublime Text package makes it possible to switch another settings (User/Preferences.sublime-settings).

If you want to change settings depending on the situation, this package provides the function to store some settings and to quickly change these settings.


FichteFoll:

t8m8:
I'm done. Thanks.

PR #5385, opened 2016-04-03
Add AngularJS Jasmine Boilerplate package

namoscato:


FichteFoll:
Because you are execting files from your package, you need to specify a `.no-sublime-package` file in order to tell PC to explicitly unpack it.

namoscato:
@FichteFoll, thanks; this is now resolved.

PR #5388, opened 2016-04-04
Add TabSnippets package

oleksiyk:
This simple SublimeText module will make sure TAB will insert the matching snippet (based on snippet's tab trigger) or TAB (\t) if no snippet matches. Other completions can be inserted with ENTER while autocomplete panel is visible.

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -91,6 +91,16 @@
 			]
 		},
 		{
+			"name": "TabSnippets",
+			"details": "https://github.com/oleksiyk/SublimeTabSnippets",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "TabsShortcuts",
 			"details": "https://github.com/ejoubaud/SublimeTabsShortcuts",
 			"releases": [


FichteFoll:
Wondering why the behavior you try to achieve is not already accomplished by the settings you referenced (`"auto_complete_commit_on_tab": false, "tab_completion": false`), I investigated a bit with `sublime.log_commands(True)` and came to the conclusion that this is probably a bug.

Both times, when the completion popup is visible or not, `command: insert_best_completion {"default": "	", "exact": true}` is executed, although it does insert the currently selected completion anyway if the popup is open.

oleksiyk:
@FichteFoll You are right, it is a bug and so think a lot of ST users, but not Jon Skinner unfortunately:
https://forum.sublimetext.com/t/setting-tab-completion-confusion/5964 :(

vovkkk:
If you set `"auto_complete_commit_on_tab": true` then <kbd>enter</kbd> will insert `\n` and won’t commit completion when pop up is visible.
`"tab_completion"` defines behaviour when pop up is not visible.

So it is not a bug, just a bit confusing.



> Both times, when the completion popup is visible or not, command: insert_best_completion {"default": " ", "exact": true} is executed, although it does insert the currently selected completion anyway if the popup is open.

Right, because app justly assumes that the one chosen by user is the best one; how it can be different? Inserting random ish while popup is visible with obvious selection is not user-friendly.

oleksiyk:
While its not a bug in its default understanding, it is a design issue. It doesn't allow to distinguish between completing snippets and inserting word completion. It forces user to close popup window which might not be yet visible at the moment user decides to hit TAB (because of `auto_complete_delay` option). In real life this causes a lot of false completions.

vovkkk:
> It doesn't allow to distinguish between completing snippets and inserting word completion

Good idea for feature request, not a reason for ‘dude, your app is broken FIXITNOW!!!!11’
I mean that link to forum.

oleksiyk:
I think that forum thread discussion was clearly a feature request and it was clearly denied by ST author.

vovkkk:
WAT?

> There are like 3 different auto complete settings, and none of them work. Nice design.

I’m not even sure what we discussing here? 
I mean come on.

oleksiyk:
@vovkkk Calm down, man. If you have something to say to the man who posted those lines (not me, btw) - login to forum and say there, not here. This is not a forum and not a discussion who should have done what.

vovkkk:
I’ve commented here to explain why exactly current behaviour is not a bug in case someone decide to add it into unofficial bug tracker as a bug.

You said that many think it is bug, author don’t want to fix it, then you said it was request not a bug, now you tell me what I must to do. Chap! That’s brilliant! Just perfect! not

PR #5390, opened 2016-04-05
Add Hungry Backspace plugin

xdrop:
Adding the [Hungry Backspace](https://github.com/xdrop/Hungry-Backspace/blob/master/README.md) plugin, for maintaining indentation level on backspace presses.

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -1089,6 +1089,17 @@
 			]
 		},
 		{
+			"name": "Hungry Backspace",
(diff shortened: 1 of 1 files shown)


FichteFoll:
[Line 5](https://github.com/xdrop/Hungry-Backspace/blob/02a90fc240009e29dae0664ca7de1eb13196e715/hungry_backspace.py#L5) should be removed.

PR #5391, opened 2016-04-05
Rename Test RSpec to TestRSpec package

astrauka:
Rename Test RSpec to TestRSpec package

Using `TestRSpec` package name to load theme files and want to have single worded package name


FichteFoll:

astrauka:
@FichteFoll updated

PR #5394, opened 2016-04-06
Add ChineseTranslator plugin

toaol:
This is translate plugin for Chinese.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -745,6 +745,17 @@
 			]
 		},
 		{
+			"name": "ChineseTranslator",
+			"details": "https://github.com/toaol/sublime-translate",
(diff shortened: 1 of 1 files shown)


FichteFoll:

`ctrl+q` is already bound by default to the `toggle_record_macro` command. Please do not override default key bindings and either choose a free binding or suggest your users to set this binding on their own.

toaol:
@FichteFoll Thanks, I have already updated them.

FichteFoll:
Well, `ctrl+shift+q` is also a default binding - for `run_macro`.

toaol:
@FichteFoll I am sorry about it because in my system the key binding of `run_macro`  is `ctrl+alt+shift+q`, and now, this might be OK.

PR #5397, opened 2016-04-06
Add QuickPanelEnhanced package

jaytiar:


FichteFoll:
I don't think your method of finding the quick panel is very reliable. I suggest at least also checking the `view.settings().get("is_widget")` setting (hoping this works).

jaytiar:
Thanks. is_widget is useful. In fact it turns out that is_widget is true for the overlay, but panel_has_focus is not, so by combining those, you can do the whole context detection in the key-bindings and I don't need to track act/deact events at all - much better! I've pushed a new release of the package with this change. And that now works properly with the console/find etc. visible at the same time.

PR #5400, opened 2016-04-07
Rename C Complete to CComplete

ibensw:


FichteFoll:

ibensw:
Added the old name to previous_names

PR #5401, opened 2016-04-07
Add LanguageTool package

gtarawneh:
[Previous version of this comment overwritten by mistake]

This is a package for integrating LanguageTool into Sublime Text.

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -64,6 +64,16 @@
 			]
 		},
 		{
+			"name": "LanguageTool",
+			"details": "https://github.com/gtarawneh/languagetool-sublime",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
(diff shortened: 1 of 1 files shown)


FichteFoll:
As a suggestion, you should allow setting other languages for the httprequest.

Also, please do not override default key bindings, such as `ctrl+shift+r` and `alt+left/right`.

gtarawneh:
Thanks for the heads up, didn't even realize I was overriding existing bindings (feels like all possible bindings are exhausted these days). I changed the plugin bindings and have also enabled a LanguageTool language autodetect feature for now.

FichteFoll:
`alt+f` is an accelerator (at least on Windows) to the "File" main menu entry.
The other bindings should be fine. I can't comment on OSX.

PR #5406, opened 2016-04-09
Add KeyboardSelection

robertcollier4:
Add KeyboardSelection


FichteFoll:
You say "Keyboard movement and selection to custom delimeters.", but the delimiters seem to be hardcoded?

PR #5407, opened 2016-04-10
Add FlowIDE

tptee:
_From the description:_

Add first-class Flow support to Sublime Text! FlowIDE provides autocomplete, diagnostics, type hints, and jump-to-definition for the Flow static type checker for JS. FlowIDE provides the majority of Flow-related features implemented in Facebook's Nuclide.

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -942,6 +942,17 @@
 			]
 		},
 		{
+			"name": "FlowIDE",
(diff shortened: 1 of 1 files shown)


FichteFoll:
You should mention that having `flow` available on the system path is a requirement, or add a "flow_path" setting. ;)

tptee:
Thanks, will do!

PR #5409, opened 2016-04-10
Org agenda

robinchenyu:
add org agenda for orgmode.

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -858,7 +858,7 @@
 				}
 			]
 		},
-		{
+		{	
 			"name": "iRules",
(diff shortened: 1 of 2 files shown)


FichteFoll:
The url you point to 404s.

robinchenyu:
have fixed it. Thanks

FichteFoll:
`.sublime-syntax` files are only available starting build 3092. Please update the ST requirements to that.

I also don't see a particular need for your package to only be available to Windows, so you might want to change that too.

robinchenyu:
fixed

PR #5410, opened 2016-04-10
Adding Share File Plugin

dineshkumar-cse:
plugin for sharing files.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1025,6 +1025,17 @@
 			]
 		},
 		{
+			"name": "Share File",
+			"details": "https://github.com/dineshkumar-cse/ShareFile",
+			"labels": ["share", "file", "code", "upload"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Shebang",
 			"details": "https://github.com/samizdatco/sublime-text-shebang",
 			"releases": [


FichteFoll:
1. Instead of defining your own command, you could just use the built-in `insert` command.
2. From a design perspective, `ServiceApi` should really not be a class but a module with functions, since it doesn't share state. Alternatively, use staticmethods.
3. Is `"http://localhost:8080/"` really the url you want to upload to?
4. The ST API is not thread-safe on ST2. Did you verify that your plugin works there?

dineshkumar-cse:
Hi, Thanks for the feedback.
- I had to put another command as insert required `edit` param, and my class was extending `WindowCommand` and i didn't have edit, so as a workaround i did this. If there's a better way i would love to change it. (i couldn't find it from doc, sublime, sublime_plugin code)
- I've updated the ServiceApi to have classmethods,  importing module was troublesome to keep it consistent across python3 and python2, so i've kept it in same class. 
- sorry about the url, i 've missed to push local changes.
- I've not tested in ST2, but to make it compatible with ST2 is in my Todo's. 
- I might need more inputs on thread safety with my requirement, as i didn't really foresee any issues with uploading it multiple time, or downloading the content similarly as i don't really maintain and state and play with it.

dineshkumar-cse:
I would really love some IRC for our plugin development, or a place where i can get comments on my code separately and later submit a pull request. 
Not sure whether just submitting pull request directly is right, as i think there's some collaborators and i value their time.

FichteFoll:
You can always find me (and others) in ##sublimetext on freenode.

Regarding your reply:

1. That's why I said to use `view.run_command("insert", {"characters": content})`.
2. Yeah, I feel you. From a design point, they should still be staticmethods however, since they do not rely on anything in the class (the `cls` parameter). They are entirely autonomous.
3.  -
4. Just a heads up that your package will be installable for ST2 at this moment. If you say you have ST2-support on the road (I honestly wouldn't waste time for that anymore, personally), I'll accept it like that however.
5. You can always use `sublime.set_timeout` to run something on the main thread again.

FichteFoll:
```
[20:32:30] <dineshkumar> using insert command alters the content that i want to paste to the new view (tab), is there any way to get around with that?
```

There's two ways to prevent that. One is to run `view.settings().set('auto_indent', False)` and the other is to use the insert_snippet command like so:

```py
view.run_command("insert_snippet", {"contents": "$somevar", "somevar": content})
```

dineshkumar-cse:
I've pushed the [changes](ae1570a56d2810db08b1992e137f2a3c4e09080d).

PR #5418, opened 2016-04-14
Separate HTML Typograph versions

einomi:


FichteFoll:
Note that the latest update is broken for ST2 and you should push a `st2-1.0.8` tag with the last commit that still worked on ST2.

You may and can not switch from tag-based to branch-based versioning. Please either set the ST3 channel to use `"tags": true` or `"tags": "st3-"` and push an appropriate tag.

einomi:
Ok, I've pushed a fix.

PR #5419, opened 2016-04-14
Updated project page

mtxr:
Just updated the project page

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -2345,7 +2345,8 @@
 		},
 		{
 			"name": "SQLTools",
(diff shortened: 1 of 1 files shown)


FichteFoll:
This change is not necessary. By default PC fetches the "homepage" link from your repository configuration, which is `http://mtxr.github.io/SQLTools` currently, which in return forwards to `http://code.mteixeira.me/SQLTools/`.

Removing the `details` or even changing it to a non-github or non-bitbucket url breaks everything.

mtxr:
Hmmmm. Nice. Didn't know that.

Thanks @FichteFoll

PR #5425, opened 2016-04-16
renamed plugin from "Jolie Language" to "sublime-Jolie"

thesave:


FichteFoll:
1. Your package does not follow our [package name requirements](https://packagecontrol.io/docs/submitting_a_package#Step_2).

2. When renaming you have to add the old name to a `previous_names` array.

3. You don't seem to need a `.no-sublime-package` file in your package. Please remove it.

thesave:
I should have fixed points 1. and 2.
Regarding point 3. the plugin uses the executable `LaunchOSXTerminal.scpt`, hence the `.no-sublime-package`.

Is it ok?

FichteFoll:
Ah, I see, you're calling that in your build system. Yes, then it's okay (required even).

thesave:
Perfect, thanks.

PR #5426, opened 2016-04-16
Add Oceanic Next Italic color scheme

gkonterman:


FichteFoll:
Please note that your newer tag has a lower version number than your older one.

jnijboer:
@FichteFoll We have removed the newer tag that has a lower version. cc: @gkonterman

PR #5432, opened 2016-04-18
added  SnipLime

yasintoy:


FichteFoll:
You should be using literal tab characters in snippets in order to honor the user's indentation settings, because tab characters will be replaced upon inserting snippets.

FichteFoll:
@yasintoy ping

PR #5433, opened 2016-04-18
Add "Git Commit Message Auto Save"

aristidesfl:


FichteFoll:
You need to test whether `view.file_name()` is `None` because otherwise the following attribute access raises an exception.

aristidesfl:
@FichteFoll thanks, fixed! 👍

aristidesfl:
How long does it take until it is available through package control?

FichteFoll:
Up to an hour.

PR #5442, opened 2016-04-21
Added "Drafts" plugin

eLindros:
Added the infos to my Drafts plugin.


FichteFoll:
[Semantic versions](http://semver.org) consist of exactly three numeric parts. Leading zeros on individual parts are also not allowed.

FichteFoll:
@eLindros ping

PR #5443, opened 2016-04-21
add GoGuru

alvarolm:
:)

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -887,7 +887,7 @@
 				{
 					"sublime_text": ">=3000",
 					"platforms": ["*"],
-					"branch": "master"
+					"branch": "tags"
 				}
 			]
 		},
--- a/repository/g.json
+++ b/repository/g.json
@@ -887,7 +887,7 @@
 				{
 					"sublime_text": ">=3000",
 					"platforms": ["*"],
-					"branch": "tags"
+					"tags": true
 				}
 			]
 		},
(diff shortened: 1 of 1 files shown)


FichteFoll:

You can import modules relative to your package (in ST3) using a relative import, such as `from .dep import shellenv`. Also note that `dep/shellenv/` contains `.pyc` files, which will be ignored by ST.

show_quick_panel can show entries with multiple lines. Just pass a list of lists of strings instead of a list of strings (thinking of having the description in the second line).

alvarolm:
done :)

repository/g.json
```diff
@@ -880,6 +880,18 @@
...
 		},
 		{
+			"name": "GoGuru",
+			"details": "https://github.com/alvarolm/GoGuru",
+			"labels": ["go", "guru"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"platforms": ["*"],
+					"branch": "tags"
```
FichteFoll: `"tags": true`, not `"branch": "tags"` ;)

FichteFoll:
What about my second paragraph? The third is not important.

alvarolm:
at the time isnt a critical matter, but it will be implemented **(https://github.com/alvarolm/GoGuru/issues/11),** thanks for the feedback mate .

FichteFoll:
Yes, it is not critical but I didn't want this to go unnoticed, which is why I mentioned it again.

PR #5445, opened 2016-04-22
Add String Generator

0rez91:
Generate alphanumeric strings with a pattern

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -2683,6 +2683,16 @@
 			]
 		},
 		{
+			"name": "String Generator",
+			"details": "https://github.com/0rez91/String-Generator",
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
For future reference, this is a reopen of https://github.com/wbond/package_control_channel/pull/5150.

@iambibhas, again, what do you think of adding this functionality to https://packagecontrol.io/packages/Generate%20Password? Imo it is not necessary to have two packages with such similar scope and usage.

FichteFoll:
@iambibhas has been unresponsive (for far too long), so I'm willing to add this package regardless.

However, @0rez91, you have to change`"branch": "master"` to `"tags": true`.

FichteFoll:
@0rez91 ping

PR #5448, opened 2016-04-24
add PaneNavigation

ahuff44:


FichteFoll:
The popular [Origami](https://github.com/SublimeText/Origami) package does this just fine, and more. It seems redundant to add a package with just that subfeature, especially when it already exists as a ST2 version.

Besides, you package would be too similar to the original. I would ask you to add the ST3 version as a different release or replace the original instead, since yours seems to support both versions.

ahuff44:
Does Origami do this? I just installed it and I don't think it does. Origami is all about moving tabs between panes; this package is about moving between tabs within a single pane.

This has made me realize the package name is bad; I think I'll call it BetterTabCycling instead.

FichteFoll:
Ah, I understand now. No, that's not something that Origami does, you're right. I like the new name better, too.

PR #5450, opened 2016-04-26
Cadence Skill addet to the repository list

leoheck:


FichteFoll:
We require acknowledgment of the package author, @noisyass2, before referencing their repo here. If @noisyass2 doesn't respond, you could fork the repo and add that instead.

Furthermore, there needs to be a [semantic version](http://semver.org) [tag](http://git-scm.com/book/en/v2/Git-Basics-Tagging) (with optional leading `v`) in repo.

leoheck:
Ok, I sent an email to him a few days ago but he didn't answer yet. So, how many days we should wait his answer?

FichteFoll:
Technically no day, since this pull request is about positive permission rather than negative (like a removal or takeover) and you are allowed to fork anything on github.

Usually we give users 14 days to respond to mentions.

PR #5453, opened 2016-04-27
Added SublimeInChains - modified:   repository/s.json

hc-12:


FichteFoll:
Your package name violates our naming guidelines: https://packagecontrol.io/docs/submitting_a_package#Step_2

hc-12:
Oh that's a shame. I really like that name! Somehow just  __"In chains"__ doesn't have the same ring to it.

Anyway to make this as exception? I'm also open for ideas.

FichteFoll:
I do not think the term "Sublime" provides anything significant to this package's name, so no exception here.

I though of "DontBreakTheChain" or maybe just "Chain".

hc-12:
Will rename to chains

PR #5455, opened 2016-04-28
add BatchFile to b

robertcollier4:


FichteFoll:
What does this package provide that the default package "Batch File" does not?

Besides, your name is too similar to the default package.

robertcollier4:
I should have attributed more that `Batch File.tmLangage' is from the default package. This package adds  This package adds the `Comments.tmPreferences` for hotkey "REM " add ability and some snippets. I had previously disabled my default package Batch File and had forgot I had done so - I will cancel this pull request and instead post the recommend additions to the SublimeText forums for a possible update in future version of official 'Batch File.sublime-package'

FichteFoll:
I suggest just creating an issue or pull request at https://github.com/sublimehq/Packages instead.

PR #5456, opened 2016-04-28
Added ansible vault package

adambullmer:
Tool for viewing, encrypting, and decrypting ansible vault files

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -715,6 +715,18 @@
 			]
 		},
 		{
+			"name": "Ansible Vault",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Please remove your `setup.py` from the repo or wrap it in an `if __name__ == '__main__':` block. Otherwise it would get loaded and scanned for plugins by ST, which it clearly is not intended for.

adambullmer:
Wrapped in that if statement. Needed the file to exist so I could continue using tox for testing

PR #5458, opened 2016-04-28
Added CodeIniter

arpitgogia:
Plugin to initialize files with obvious code snippets based on the file type. Fully customized.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1512,6 +1512,16 @@
 			]
 		},
 		{
+			"name": "CodeIniter",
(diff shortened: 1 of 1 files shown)


FichteFoll:
These two packages provide this functionality already:

- https://packagecontrol.io/packages/FileTemplates
- https://packagecontrol.io/packages/FileHeader

I don't think there is need for another one, do you?

arpitgogia:
Apologies. I just wrote this script for fun and thought of uploading it to Package Control. The extensions you referenced do the same job. Hence, I don't think there is a need for another one.

PR #5460, opened 2016-04-28
Added Sublime Tutor package

jaipandya:
Inspired from vimtutor, Sublime Tutor is an interactive in-editor keyboard shortcuts tutorial for Sublime Text. The tutor introduces all the frequently used shortcuts and features of ST in 6 easy to read modules.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -2971,7 +2971,17 @@
 				{
 					"sublime_text": ">=3000",
 					"platforms": "osx",
-					"branch": "master"
+					"tags": "osx-"
+				},
+				{
+					"sublime_text": ">=3065",
+					"platforms": "windows",
+					"tags": "win-"
+				},
+				{
+					"sublime_text": ">=3000",
+					"platforms": "linux",
+					"tags": "linux-"
 				}
 			]
 		},
--- a/repository/s.json
+++ b/repository/s.json
@@ -2965,6 +2965,17 @@
 			]
 		},
 		{
+			"name": "Sublime Tutor",
+			"details": "https://github.com/jaipandya/SublimeTutor",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"platforms": "osx",
+					"branch": "master"
+				}
+			]
+		},
+		{
 			"name": "Sublime Tweet",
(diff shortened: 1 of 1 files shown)


jaipandya:
screen shot 2016-04-29 at 2 19 43 am
screen shot 2016-04-29 at 2 20 04 am

FichteFoll:
In our [docs](https://packagecontrol.io/docs/submitting_a_package#Step_2), we ask package authors to consider use package names not starting with the name "Sublime", however I do see the semantic appeal in having this specific package being an exception. Still, I would like to ask you to think of an alternative name that does not contain the word "Sublime". If you absolutely can not come up with a suitable alternative, we should be able to merge regardless.

jaipandya:
Hi @FichteFoll,

I read the docs before coming up with this name, I thought about it and then finalized with the name because of a couple of reasons:

1. The package was intended to be on the lines of `vimtutor`. For an alternative in the sublime world, it was obvious for me think of `sublime tutor` as the name.

2. The reasoning given in the docs for not prepending `sublime` to any package name:
> Every package available via Package Control is for Sublime Text. Using the word Sublime just adds noise to the list when trying to find packages.

   I agree. I have also seen package names like `Sublime Bookmarks`, `sublime-github`, `SublimeGerrit`, which could totally do away with the word sublime. With this reasoning, I thought to just keep `Tutor`. But then it could be a tutor for anything, resulting in a very generic term. `Sublime Tutor` semantically made sense, since it is a tutor for Sublime Text. Yes, couldn't come up with a better or an alternative name.

FichteFoll:
@jaipandya I'm okay with that reasoning, so I'll merge.

I was on vacation when you pinged me on IRC and you're not online, so I'll reply here:

> [12:16:28] <jaiZ> Also, I had a question regarding the usage of ST3 vs ST2. Do we maintain some sort of analytics which can tell the percentage of total users still using an older version than ST3?

The latest numbers I know of are from [March](https://forum.sublimetext.com/t/sublime-text-2-not-developed-anymore/18680/4?u=fichtefoll) and indicate a 20% share of ST2, from Package Control users. It's most likely not worth to waste time on ST2 support nowadays.

PR #5468, opened 2016-05-02
sublime boto3 python package

revmischa:
🐍 boto3 for AWS

### DIFF
--- a/repository/dependencies.json
+++ b/repository/dependencies.json
@@ -31,20 +31,6 @@
 			]
 		},
 		{
-			"name": "bz2",
-			"load_order": "02",
-			"description": "Python bz2 module",
-			"author": "wbond",
-			"issues": "https://github.com/codexns/sublime-bz2/issues",
-			"releases": [
-				{
-					"sublime_text": "*",
-					"base": "https://github.com/codexns/sublime-bz2",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "boto3",
 			"load_order": "02",
 			"description": "Python boto3 module for Amazon Web Services",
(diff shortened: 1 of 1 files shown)


FichteFoll:
I suggest removing `all/setup.py` from the repo to prevent potential issues for users trying to `import setup` (for whatever reason).

revmischa:
https://github.com/revmischa/sublime-boto3/commit/d01d62e371cd9d3d0141d579163247be984bda20

FichteFoll:
By the way, what about the packages that boto3 requires?

```py
requires = [
    'botocore>=1.3.0,<1.4.0',
    'jmespath>=0.7.1,<1.0.0',
]
```

revmischa:
i don't know, does it not chase down dependencies? or dependencies of dependencies? or dependencies of dependencies of dependencies?

FichteFoll:
Not at all. Package Control doesn't have a dependency (version) resolver, nor a way for dependencies themselves to have other dependencies. Packages that need one (or two) would have to specify all packages as dependencies.

That said, it would probably make sense to include `botocore` in the `boto3` depedency as well, but idk about `jmespath`.

revmischa:
I believe all required dependencies are in there now. 

```
import sys
sys.path.append('all')
import boto3
```
works without emitting any errors.

I'm sure there is some wisdom in managing packages in this manner instead of via pip that I am uninformed about. I find it rather discouraging because I can't really test my code until this PR is accepted.

revmischa:
https://github.com/revmischa/sublime-boto3/commits/master

FichteFoll:
Yes, looks good now.

You can use the ["Install Local Dependency" command](https://packagecontrol.io/docs/usage) for local testing to limited extend, I suppose.

revmischa:
Any reason I still get:
`Package Control: The dependency specified, boto3, is not available`
I've tried everything I can think of.

FichteFoll:
What are you doing exactly?

revmischa:
I'm working on https://github.com/revmischa/sublime-awslambda
I figured out I had to create a version tag for boto3. It seems to find the dep now.

FichteFoll:
Ah yes, my bad. I forgot to check whether you pushed a tag for the dependency repo.

revmischa:
I suggest making that step more visible in the docs, I was banging my head against the wall for a while

FichteFoll:
Please open an issue for that on https://github.com/wbond/packagecontrol.io/issues, if you were so kind.

revmischa:
Donezo

PR #5471, opened 2016-05-04
Add Convert Full Half Width package.

naoyukik:
Thanks!!

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2395,6 +2395,17 @@
 			]
 		},
 		{
+			"name": "ConvertFullHalfWidth",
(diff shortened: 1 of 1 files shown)


FichteFoll:
This is also provided by https://packagecontrol.io/packages/Zen2Han, except that does not convert non-kana characters (i.e. numbers and latin characters). Any thoughts on adding that functionality to that package or replacing the existing instead?

cc @KazIgu

naoyukik:
I would like you to newly registered.
My package used library "jaconv". (https://pypi.python.org/pypi/jaconv)
If you add the function to "Zen2Han", it will be reworked.

I plan to add functionality.
It convert "Hiragana" to "Katakana" or "Katakana" to "Hiragana".

PR #5472, opened 2016-05-04
add EasyClangComplete

niosus:
adding package EasyClangComplete for clang enabled autocomplete. There is a number of packages that do approximately the same, but all of them are hard to configure or lack features. This one aims to be easy to start using and easy to hack.

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -77,6 +77,18 @@
 			]
 		},
 		{
+			"name": "EasyClangComplete",
+			"details": "https://github.com/niosus/EasyClangComplete",
+			"labels": ["auto-complete", "linting", "language syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": ["osx", "linux", "windows"],
+					"tags": "st3-"
+				}
+			]
+		},
+		{
 			"details": "https://github.com/facelessuser/EasyDiff",
 			"releases": [
 				{


FichteFoll:
You can use relative imports in ST packages (as you already do [here](https://github.com/niosus/EasyClangComplete/blob/master/EasyClangComplete.py#L42)), which should eliminate the need to provide `.no-sublime-package` as you are not invoking something with `subprocess`.

niosus:
Thanks for the advice! I think I did not fully get what you say, though.
> as you are not invoking something with `subprocess`
I do use `subprocess` but I invoke something that is not part of my code (`clang`, to be precise). Does that count? Are there particular benefits of having `.no-sublime-package` removed?

FichteFoll:
I edited my comment to be more precise about the `subprocess` part, but yes, you only need `.no-sublime-package` if you are executing a file *whithin* the package as it would otherwise be packed in a ZIP archive. See also http://docs.sublimetext.info/en/latest/extensibility/packages.html.

In the above documentation you will also find the benefit of using the .sublime-package archive format which is the ability to override certain files with "override packages" without losing changes on a package update.

niosus:
What you say makes sense. Thanks for clarifications! I will remove the `.no-sublime-package` file from the repo.

niosus:
By the way, is there a way to test if the plugin is actually working while being packaged?

niosus:
Hey @FichteFoll with commit 689cf3b3c76c9d91c3fe46724e74f1bc6d83e515 I have moved everything (apart from tests, but they are only needed when cloned) to use relative includes. Now everything should be ready, right?

niosus:
I will be off to an important work meeting the whole next week and I am desperately willing to write this package from my todo list until then. So if there is anything I can assist you with @FichteFoll just let me know. Please tell me should there still be issues with my package so I can fix them until the end of this week. Thanks a lot!

FichteFoll:
>By the way, is there a way to test if the plugin is actually working while being packaged?

Sure, you can just zip it by yourself, change the extension to `.sublime-package` and place that into your Installed Packages folder. Or you add the repository url to Package Control and have it install the package for you: https://packagecontrol.io/docs/usage. Make sure to remove it when you're done though.

niosus:
@FichteFoll that is awesome. Thanks for the tip and for merging!

PR #5476, opened 2016-05-05
Added: Python dependency PyZMQ 15.2.0

randy3k:


randy3k:
@wbond, by the way, is there any good way to browse existing python dependencies other than [dependencies.json](https://github.com/wbond/package_control_channel/blob/master/repository/dependencies.json).

Adding the list of dependencies in https://packagecontrol.io/docs/dependencies may work. In this way, developers could use existing dependencies rather than compiling their own. (It was a pain to get all platform's binary done.)

wbond:
I think FichteFoll made https://packagecontrol.github.io/. I have not built anything for such browsing.

In general there hasn't seemed to be much uptake of the dependencies functionality. Most package developers seem to think it is a way to have one package require another package, and others expect it to be a full PyPi client.

Compiling for all different platforms is a pain. I don't see that going away, ever.

FichteFoll:
Yes, I made https://packagecontrol.github.io/ a while ago and haven't updated it in a long time, which means it's definitely incomplete. Currently browsing `dependencies.json` is your best option, but I'm open to pull requests for https://github.com/packagecontrol/packagecontrol.github.io.

FichteFoll:
Here's the corresponding issue on the homepage repo: https://github.com/wbond/packagecontrol.io/issues/55

PR #5478, opened 2016-05-05
Added "Theme - Windows 10" package.

andreibsk:


wbond:
Does this work with Sublime Text 2? If not, you'll need to change the value of `sublime_text` to `>=3000`.

andreibsk:
You're right, my bad. Version updated with f6c9399722c5c8b47aee44b814d69dae5f6cdc3b.

PR #5481, opened 2016-05-06
Added syntax highlighting support for SINUMERIK 840D CNC

deathaxe:


FichteFoll:
`.sublime-syntax` support has been added with build 3092, so please adjust the version selector.

I suggest not using "Syntax" as the first word or even at all in your package name.

deathaxe:
I renamed the package to "CNC ... language support" and set version selector to >=3092.

Is it a better solution? 

I found several packages beginning with 'Syntax' as well as 'Theme' and so on. That's why I started with Syntax as well. Hope the repository name is ok as it is? I added sublime- as a prefix there because I maybe want to add more repos for other projects there.

FichteFoll:
>I found several packages beginning with 'Syntax' as well as 'Theme' and so on. That's why I started with Syntax as well. 

It makes sense for "Theme" (and even is a convention) because themes usually have arbitrary names while syntax support packages normally include the name of the syntax they provide support for or consist of it entirely. The default packages for example are named "Python" and "Java" instead of "Syntax Java" or "Syntax Python".

>Hope the repository name is ok as it is? I added sublime- as a prefix there because I maybe want to add more repos for other projects there.

The repository name is overridden by the "name" key in JSON and doesn't matter.

PR #5483, opened 2016-05-07
add GoRename

alvarolm:
:)


FichteFoll:
Please see my comments in https://github.com/wbond/package_control_channel/pull/5443.

alvarolm:
it has the bare minimum to be accepted, isn't ? the relate import isn't something critical but it will be implemented (https://github.com/alvarolm/GoGuru/issues/11)

PR #5484, opened 2016-05-07
add goodmind/sublime-flowstate package

goodmind:


FichteFoll:
I highly suggest not [using `sublime.status_message`](https://github.com/goodmind/sublime-flowstate/blob/7f2fefb5b1dceb580c3c4121fb3fd1d69370667c/flowstate.py#L5) to show that your plugin is [initializing](https://github.com/goodmind/sublime-flowstate/blob/7f2fefb5b1dceb580c3c4121fb3fd1d69370667c/flowstate.py#L15) as that will be displayed on every ST restart and provide little to no usable information to the user. This is better suited to just be printed to the console.

PR #5488, opened 2016-05-09
adds `Base Snippets` to "repository/b.json"

jonschlinkert:


FichteFoll:

You should use literal tabs for indentation in snippets because Sublime Text will convert them according to the user's indentation settings when inserting.

jonschlinkert:
done on `"tags": true` and git release. I'll take the indentation under advisement. thanks.

PR #5489, opened 2016-05-10
Add AVA Standard

sotojuan:


FichteFoll:
What are the differences to https://packagecontrol.io/packages/AVA?

sotojuan:
It's the same but without semicolons to fit the JS Standard style.

See [here](https://atom.io/packages/ava-standard) for the Atom version.

FichteFoll:
You should probably add that to the readme.

PR #5491, opened 2016-05-10
add Verilog Gadget package

poucotm:


poucotm:
Some files are missing, I will request pull again

PR #5492, opened 2016-05-10
Add CoberturaCoverage to respository

jeffrand:
Signed-off-by: Jeff Rand <jeff.rand@artivest.co>

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1252,6 +1252,16 @@
 			]
 		},
 		{
+			"name": "CoberturaCoverage",
+			"details": "https://github.com/jeffrand/CoberturaCoverage",
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
`ctrl+shift+n` is a default binding. Please choose a different one or don't define one at all and let users set their own.

FichteFoll:
Remember that you need to push a  new tag in order to publish a release.

jeffrand:
@FichteFoll done! Let me know if there's anything else I need to do.

FichteFoll:
Your `messages.json` file contains a trailing comma, which is invalid JSON.
It won't matter currently since nobody will see the update messages (they only start seeing these once they installed the package and it's not yet installable).

PR #5494, opened 2016-05-11
Add px rem tooltip

tbremer:
Just adding my new package (https://github.com/tbremer/px-rem-tooltip-sublime) to the repo.

Thanks!

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -2055,6 +2055,16 @@
 			]
 		},
 		{
+			"name": "PxRemTooltip",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Popups are not available in ST2.

tbremer:
@FichteFoll I just pushed an update to use only versions `>3070` where `show_popup` [was introduced](https://forum.sublimetext.com/t/dev-build-3070/14538).

tbremer:
Thanks @FichteFoll!

PR #5496, opened 2016-05-12
Add "JSON Key-Value" package

aurule:


FichteFoll:
Your package relies on [PackageDev](https://github.com/SublimeText/PackageDev/blob/22a6e24faf52287a51d9ecc91b62b245cc60d06a/Syntax%20Definitions/JSON%20Generic%20Array%20Elements.JSON-tmLanguage), which is a bad idea.

Ideally, Sublime Text by itself would highlight keys different from regular strings. There already are different meta scopes, but they don't allow a selector to specifically target them. I opened an issue upstream for this: https://github.com/sublimehq/Packages/issues/421.

If you remove the unnecessary dependency on PackageDev I'll merge the package because it is also compatible with ST2.

aurule:
OK. I'll remove the dependency so that this can be used in the interim before a fix, and for ST2. Thanks!

FichteFoll:
Please open a new PR once you are done with updating.

PR #5497, opened 2016-05-12
added "Conditional ColorSchemes" plugin & updated "SlideNav" to ST v3+

Enteleform:


FichteFoll:
You don't need to go around to run the `set_setting` command. Just directly use `view.settings().set`.

PR #5498, opened 2016-05-12
Add HTML (C#)

michaelblyons:
If I did something incorrectly, don't hesitate to say so.


michaelblyons:
Fixed alphabetical order for symbols.

PR #5503, opened 2016-05-14
Add GNU GAS listing file repository entry.

abcminiuser:
Adds a new simple SublimeText 3 package that provides syntax highlighting for GCC GAS listing files, generated by many build systems (or manually by users).

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -101,7 +101,7 @@
 			"labels": ["language syntax"],
 			"releases": [
 				{
-					"sublime_text": ">=3000",
+					"sublime_text": ">=3092",
 					"tags": true
 				}
 			]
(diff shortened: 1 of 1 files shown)


FichteFoll:
`.sublime-syntax` support has been added with build 3092. Please adjust your version selector accordingly.

FichteFoll:
How does it compare to https://packagecontrol.io/packages/GAS-x86%20%28or%20x64%29%20highlighting%20scheme?

abcminiuser:
The other package is far, far more complete - but only works for x86 architectures. My syntax highlighting assumes that the generated instructions are always valid and highlights the sections accordingly. The intention of this package is to make reading machine-generated listing files easier, rather than to aid the manual writing of GAS code.

PR #5506, opened 2016-05-14
Added Interactive Data Language (IDL) syntax highlighting

cschreib:
A new package to provide IDL syntax highlighting.
Proposed for inclusion in Package Control after discussion with @wbond (https://github.com/sublimehq/Packages/pull/372).

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -651,6 +651,16 @@
 			]
 		},
 		{
+			"name": "Interactive Data Language (IDL)",
+			"details": "https://github.com/cschreib/sublime-idl",
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
This should work just fine on any build starting with 3092, so I don't see much merit of locking users of 3103 out of it.

cschreib:
There have been continuous updates to the syntax highlighting implementation in Sublime since 3092, and I have not checked if the present package works correctly with either 3092 or 3103. Chances are that it will, indeed, but I'd like to check this beforehand.

FichteFoll:
Whatever you say~

PR #5511, opened 2016-05-15
Add gerardroche/sublime-phpunit

gerardroche:
This is similar to stuartherbert/sublime-phpunit which may at some point incorporate the features in this package, it has already added some features but it supports ST2 which means its codebase is much very complicated. It was easier to write a plugin from scratch only supporting ST3 with the features I wanted.

Originally the package name was phpunit but that conflicts with the other package so I renamed it to "php_phpunit" because honestly I didn't know what else to name it.

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -1187,6 +1187,17 @@
 			]
 		},
 		{
+			"name": "phpunitkit",
+			"details": "https://github.com/gerardroche/sublime-phpunit",
+			"labels": ["php", "phpunit", "test", "testing", "unit", "unit test"],
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
"phpunit" and "php_phpunit" are too similar for my taste. There is also "SimplePHPUnit" already.

Maybe something like "phpunitizer" or similar a variation.

gerardroche:
How about phpunit_tools?

FichteFoll:
Either would work.

gerardroche:
Thanks. I went with `phpunitkit`.

gerardroche:
ping @FichteFoll This ready to merge. Thanks.

PR #5520, opened 2016-05-17
new status bar file path package

kushalhalder:


FichteFoll:
https://github.com/unphased/SublimeStatusbarPath provides a *very* similar feature, but the package is only marked to support ST2 (which it does not since the code is definitely compatible).

I suggest adding features to @unphased's package and adjusting the version restriction in the channel here.

FichteFoll:
@kushalhalder ping

kushalhalder:
I'll do that.

FichteFoll:
@kushalhalder any update?

PR #5523, opened 2016-05-18
Add SuperCollider ST3 Package

acarabott:
There is an existing ST2 only package that is now abandoned, hence the ST3 in my package name. Alternative suggestions welcome.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -3749,6 +3749,17 @@
 			]
 		},
 		{
+			"name": "SuperCollider ST3",
+			"details": "https://github.com/acarabott/supercollider-sublime",
+			"labels": ["language syntax", "snippets"],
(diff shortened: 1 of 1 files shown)


FichteFoll:
The names are very similar. How different are the features provided by both packages?

acarabott:
They both have the same goal, the main difference is ST2/3 support. Each package only supports one version, so is written in Python 2 for ST2 and Python 3 for ST3.

My package is a rewrite from scratch that adds a number of features, so an integration would be a lot more work. 

Is ST2 support considered desirable these day?

FichteFoll:
Thanks for the explanation. That's enough of a reason to accept this package (with this name).

I personally wouldn't bother with ST2 anymore.

PR #5533, opened 2016-05-23
Add ToolTip-Helper to the repository

doobleweb:
Add ToolTip-Helper to the repository


FichteFoll:
If you use the `sublime.load_resource` API for ST3, you don't need to have a `.no-sublime-package` and require the package to be unextracted.

PR #5534, opened 2016-05-23
Add LuaExtended

viluon:
[LuaExtended](https://github.com/viluon/LuaExtended) is a better Lua syntax definition and snippet set for Sublime Text 3.

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -1234,6 +1234,19 @@
 			]
 		},
 		{
+			"name": "LuaExtended",
+			"details": "https://github.com/viluon/LuaExtended",
+			"labels": [
+				"auto-complete", "language syntax", "snippets", "lua"
+			],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "LuaProjectHelper",
 			"details": "https://github.com/ken48/LuaProjectHelper",
 			"labels": ["lua"],
--- a/repository/l.json
+++ b/repository/l.json
@@ -1241,7 +1241,7 @@
 			],
 			"releases": [
 				{
-					"sublime_text": ">=3000",
+					"sublime_text": ">=3092",
 					"tags": true
 				}
 			]


FichteFoll:
`.sublime-syntax` support has been added with build 3092. Please adjust your version selector accordingly.

Regarding linter support, you should check out the [`syntax_map` setting](http://www.sublimelinter.com/en/latest/global_settings.html?highlight=syntax_map#syntax-map), which makes this a lot easier.
You can also voice your support of being able to provide this information in the package that provides the additional syntax here: https://github.com/SublimeLinter/SublimeLinter3/issues/431

viluon:
Thanks for the heads-up @FichteFoll, I have fixed the version requirement and updated the [README](https://github.com/viluon/LuaExtended/blob/master/README.md) with a new, simpler guide on linters.

FichteFoll:

viluon:
... Yet GitHub itself suggest the 'v' prefix
> It’s common practice to prefix your version names with the letter `v`.


(Please ignore the custom theme)

[Semantic Versioning](http://semver.org/) uses it for tagging *its own* versions


and Package Control supports 'v' prefixes just fine (until https://github.com/wbond/package_control/issues/887 is closed as accepted, if that happens).

It is therefore a question if packages *should* change their `tags` fields or not, and given that there is no clash with Package Control itself, it seems just fine to use `v`'s for version prefixes, at least from my standpoint @FichteFoll.

FichteFoll:
>To my understanding, using v in GitHub release tags is completely okay.

It is. The Package Control crawler detects versions prefixed with `v` just fine, just like with not prefix when `tags` is set to `true`. You can choose a different prefix for your tags if you with to use another one, for example `"tags": "version-"`, and then tag all your releases as `version-1.2.3`.

The reason why I complained initially was because you only had major and minor parts but not patch.

viluon:
@FichteFoll Oh, I am sorry! Not sure how I missed that :confused:. Well, thanks or merging!

PR #5535, opened 2016-05-23
Added Galen v2+ syntax highlighter package

davidrv87:


FichteFoll:
What does this do compared to https://github.com/professant/sublime-galen and why can't propose the changes in a pull request?

davidrv87:
Version 1 of Galen uses a complete different syntax. I didn't want to do it via pull request just in case someone is still using some Galen v1 code

PR #5537, opened 2016-05-24
Nav Panel

sahibalejandro:
Also removed my very old package QuarkPHP, it's no longer used, no support, and no users.


FichteFoll:
Isn't this the same as https://packagecontrol.io/packages/Open%20File%20in%20Current%20Folder?

sahibalejandro:
No, is not the same, with Nav Panel you can also navigate through folders, is something like vim's netrw.

PR #5539, opened 2016-05-25
Add Sourcegraph

luttig:
Sourcegraph: See global cross-references and live usage examples for Go code, as you type. For Sublime Text 3.

Thanks guys!

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -2017,6 +2017,17 @@
 			]
 		},
 		{
+			"name": "Sourcegraph",
+			"details": "https://github.com/sourcegraph/sourcegraph-sublime",
+			"labels": ["go"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "SourcePawn Completions",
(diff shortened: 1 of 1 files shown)


FichteFoll:
[This](https://github.com/sourcegraph/sourcegraph-sublime/blob/ebc9712c347dfbeffa985e0d781905f6d4408c66/sourcegraph_lib.py#L69) most likely won't go well on Windows.

luttig:
@FichteFoll thanks for merging! One last question – the link to our homepage [here](https://packagecontrol.io/packages/Sourcegraph) seems to be broken, it should be http://www.sourcegraph.com. Is that something we can fix, or should it happen on the Package Control side?

FichteFoll:
The homepage url is inferred from your repository settings. It's not visible on https://github.com/sourcegraph/sourcegraph-sublime because it actually does contain `htttp://`, as I verified with the API:

```
C:\Users\Fichte>curl https://api.github.com/repos/sourcegraph/sourcegraph-sublime
{
  "id": 56555481,
  "name": "sourcegraph-sublime",
  "full_name": "sourcegraph/sourcegraph-sublime",
  ...
  "homepage": "htttp://www.sourcegraph.com",
  ...
}
```

PR #5544, opened 2016-05-27
Syntax highlighting for cp2k input files

nholmber:


FichteFoll:
You should probably remove your `v1.0.1` and `v1.0` tags. Currently, PC will install `v1.0.1`, which does not contain the latest commit however.

PR #5546, opened 2016-05-27
Make commit for PPCL in p.json

blandfbt:


FichteFoll:

Your package doesn't seem to do anything windows-specific, so I suggest opening installation for other platforms as well.

blandfbt:
I believe I addressed all the issues, with tagging, platforms, etc..  Is there more I need to do to get his accepted?

Thanks

FichteFoll:
I don't receive notifications for commits unfortunately.

PR #5548, opened 2016-05-28
Add iTimeTrack package

itimetrack:
Adding package iTimeTrack, which is a fork of wakatime but logs heartbeats to the itimetrack.com website and generates billable time entries.

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -911,6 +911,17 @@
 			]
 		},
 		{
+			"name": "iTimeTrack",
+			"details": "https://github.com/itimetrack/itimetrack-sublime",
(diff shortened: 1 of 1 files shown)


FichteFoll:
In a VCS-controlled enviroment (like git), you don't need to work with `*.bak` files.

FichteFoll:
Also, you should set `"tags": "itt"` so that your tags (releases) are picked up and not the old ones from wakatime.

itimetrack:
Added the tag prefix...

itimetrack:
Hi @FichteFoll ,  should I close this PR and create a new one?  This one should now have the tag prefix added.

PR #5549, opened 2016-05-29
Add Retina theme

braver:
The default theme recreated with @2x graphics

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -1340,6 +1340,17 @@
 			]
 		},
 		{
+			"name": "Theme - Retina",
+			"details": "https://github.com/braver/SublimeRetina",
+			"labels": ["theme"],
+			"releases": [
+				{
+					"sublime_text": ">3100",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Theme - Soda",
 			"details": "https://github.com/buymeasoda/soda-theme",
 			"labels": ["theme"],


FichteFoll:
I *highly* suggest moving all assets down to a subfolder for your own (and my) sanity. And adjusting the paths in the theme file ofc.

Other than that, I don't understand the purpose of this package. The default theme already ships with high-dpi icons and assets. It was even the first theme of all to support the new high-dpi assets, for obvious reasons.

braver:
I'm sorry, but you're coming of as annoyed by this package. I put my spare time in it and want to share it. There is no ill intent. 

> I highly suggest moving all assets down to a subfolder for your own (and my) sanity

My sanity is best served with the files in the exact same place as the original default theme. This makes it trivial to match changes to the default theme simply by diffing the two folders. If this adversely affects your sanity I'm sorry.

> The default theme already ships with high-dpi icons and assets.

It has some yes, but a lot of vital assets don't have @2x variations. These include all images for the tabs and tabbar, the x icon and all icons for the buttons on the search&replace panel. If you use tabs (instead of the open file list) on a Retina Macbook, the lack of resolution in these assets is glaringly obvious. They're rough and blurry. I've discussed this on the forums and github and apparently the original artwork is not high resolution enough to create @2x variations.

> It was even the first theme of all to support the new high-dpi assets, for obvious reasons.

This might be true, but it didn't stop them from adding new low resolution icon for the new panel switcher. Obviously no one is committed to keeping the assets retina compatible. 

> Other than that, I don't understand the purpose of this package.

I know most people, especially developers, don't care about rough or blurry graphics. However, people used to think Sublime was a beautiful editor. If you open it now on a Retina Macbook, it looks like and old relic. 

BTW. I reached out to @sublimehq and @wbond about this and offered them the high dpi assets and the source files for them. No response.


So that's the purpose of this package.

FichteFoll:
>It has some yes, but a lot of vital assets don't have `@2x` variations. 

Thanks for that clarification. This was not obvious to me through your PR description or the readme.

I was not offended by your package. We just have some quality standards regarding packages on the default channel that I would like to keep up for as long as it's managable. Few package managers can say that about themselves since they are mostly not reviewed at all.

braver:
> Few package managers can say that about themselves since they are mostly not reviewed at all.

Definitely, and it's a key strength of packagecontrol that someone (i.e. you) is putting in the effort. So sorry about assuming things and thanks a bunch!

Issue #5543, opened 2016-05-27
Why package marked missing


rexdf:
https://packagecontrol.io/packages/ChineseLocalization

I updated a new version and packgecontrol is missing.

rexdf:
Is an issue of packagecontrol server or github or missing infomation in message.json?

FichteFoll:
Everything looks alright from the outside. @wbond will need to step in here.

rexdf:
Why do it take so long?  I begin to take a look at https://github.com/wbond/packagecontrol.io

rexdf:
I will make it private for a while.

rexdf:
memory leak or insufficient memory
packagecontrol.io server

facelessuser:
I'm seeing this too.

facelessuser:
FYI, the plugin missing in action is BracketHighlighter.

Kozuno:
Hello, I don't understand how to rename ChineseLocalization to make BracketHighlighter install available again in package control.
Package control works great when I install it but I don't see BracketHighlighter listed in package control & my attempts to install it manually failed. I also tried to reinstall ST deleting ~/.config/sublime-text-3 folder but problem remains, I can't use this great plugin again :(
Please h3lp 4 816 n008 :v

facelessuser:
This is the same issue, but renaming ChineseLocalization won't help BracketHighlighter.  If I renamed BracketHighlighter in Package Control it might help it show up again, but I don't want to rename my package, and that is a hack to get it to come back.  Unless someone can tell me I've done something wrong, I think Package Control should fix this.

Kozuno:
@facelessuser Thank you for the feedback, I really appreciate. Let's hope Package Control will fix this integration problem, because honestly your package is really awesome to read more easily a code and I am not sure if I would like to stay on ST without this great utility. Sometimes I am really lame to find where brackets begin and BracketsHighlighter helps a lot. Probably one of the best package to me on ST, so congratulations for the code !

facelessuser:
@Kozuno, thanks!  You should be able to install the package manually as outlined here https://github.com/facelessuser/BracketHighlighter/issues/351#issuecomment-228601428.  Alternatively you can use git which I don't cover as not everyone is comfortable with git or has it installed on their machine.  You still need Package Control to install the dependencies though, but that is covered in the linked post.

rexdf:
I tried all kind of ways. 

[...]

PR #5552, opened 2016-05-30
rename ChineseLocalization to ChineseLocalizations to solve #5543

rexdf:
#5543 

I rename package name and change github url.


rexdf:
memory leak or insufficient memory
packagecontrol.io server

PR #5554, opened 2016-06-01
Added Core dna snippets

synthet1c:


FichteFoll:
You should add `*.cache` to your `.gitignore` and remove the file from the readme. It serves no purpose as ST will auto-generate it (and ST3 will place it in a different folder).

synthet1c:
Hi FitcheFoll,

Thanks for your advice. I have removed the cache file from the repo and added it to the `.gitignore`. And updated the tags to reflect the new version.

PR #5561, opened 2016-06-03
add lebab plugin for sublime

inkless:
This is a plugin to use lebab (http://lebab.io/) to convert ES5 code to ES6 code

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -421,6 +421,16 @@
 			]
 		},
 		{
+			"name": "lebab",
+			"details": "https://github.com/inkless/lebab-sublime",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "master"
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


FichteFoll:

The file name for messages is supposed to be `messages.json`, with an s.

Your code is not comaptible with Python 2 (and thus ST2) because you define "exec" as a function. Please rename the function or adjust the version requirements.

A general Python tip: You should be using modules for namespacing, not classes and staticmethods.

wbond:
- We no longer accept branch-based releases. You'll have to switch to use semver tags.
 - Your keybinding overrides the key binding for splitting a multi-line selection into multiple cursors. Probably best not to provide a key binding and just a command palette entry.

FichteFoll:
@inkless ping

inkless:
@FichteFoll  Thanks for reminding, didn't notice this for a while. 
Just updated as request, please let me know if there's still something not right. :)

PR #5568, opened 2016-06-04
Magento2 Theme Snippets

parthviroja:
Magento2 Theme Snippets

PR #5569, opened 2016-06-04
Magento2 Theme Snippets

parthviroja:
Magento2 Theme Snippets


wbond:
Your JSON file is completely invalid. It doesn't seem to be based on the example at https://github.com/wbond/package_control/blob/master/example-repository.json.

wbond:
Custom repository JSON isn't really necessary any longer, especially since you are hosting on GitHub anyway, so I am going to close this in favor of #5568

PR #5571, opened 2016-06-05
Added SAS Programming package

rpardee:
Thanks very much!

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -73,6 +73,17 @@
 			]
 		},
 		{
+			"name": "SAS programming",
+			"details": "https://github.com/rpardee/sas",
+			"releases": [
+			{
+				"platforms": ["windows", "linux"],
+				"sublime_text": ">=3000",
+				"tags": true
+				}
+			]
+		},
+		{
 			"name": "Sass",
 			"details": "https://github.com/nathos/sass-textmate-bundle/tree/sublime",
 			"labels": ["language syntax"],
--- a/repository/s.json
+++ b/repository/s.json
@@ -72,7 +72,6 @@
 				}
 			]
 		},
-
 		{
 			"name": "SAS Programming",
 			"description": "For using SAS Institute's Analytics & Data Management system." ,
--- a/repository/s.json
+++ b/repository/s.json
@@ -72,6 +72,7 @@
 				}
 			]
 		},
+
 		{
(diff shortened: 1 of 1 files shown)


repository/s.json
```diff
@@ -73,6 +73,19 @@
...
 		{
+			"name": "SAS Programming",
+			"description": "For using SAS Institute's Analytics & Data Management system." ,
+			"details": "https://github.com/rpardee/sas",
+			"labels": ["language syntax", "build-system", "snippets"],
+			"author": ["rpardee", "friedegg", "seemack"],
+			"releases": [
+			{
+				"sublime_text": ">=3000",
+				"tags": true
```
wbond: This and the two lines above all should be indented one more level.

wbond:
Snippets should use the tab character for indentation. If the user has specified indentation should happen via spaces, Sublime Text will convert the tabs to the appropriate number of spaces.

wbond:
I believe you should be able to use the built-in built system error navigation. You need to specify the regex to search for as a setting on the view named `result_file_regex`. You can see an example at https://github.com/golang/sublime-build/blob/master/golang_build.py#L958.

FichteFoll:
Or refer to some [docs](http://docs.sublimetext.info/en/latest/reference/build_systems/configuration.html#capturing-build-system-results).

wbond:
@FichteFoll It would be nice if you guys killed http://sublimetext.info/docs/en/ and forward it to relevant info.

FichteFoll:
Agree. cc @guillermooo

rpardee:
Thanks much for the feedback!  I have fixed the snippets (and hopefully the indentation now too).  

I'm unclear on the error nav though--build results get written to a separate .log file on disk, rather than come back from stdout, say.  So that file gets opened independently & is not a "results view", I don't think.  (At any rate, if I open one of these guys, the command palette doesn't show a 'next error' command, and hitting F4 doesn't give me any error or warning in the console--just does nothing.)  

You buy that? 

Thanks again!

FichteFoll:
Besides the logging output, this is pretty much exactly what the default build system target, the `exec` command, does. You can check its implementation in `Default/exec.py`. You can also refer to that for the `result_file_regex` setting.

rpardee:
Very cool--thanks!  I'll have a look at exec.py & see what I can see.

PR #5573, opened 2016-06-06
DotTemple

sophatvathana:


ahmadawais:
@sophatvathana I think your PR is failing because `"branch": "master",` is not allowed now. You should only use tags. 

The tests for your package didn't pass: https://packagecontrol.io/docs/submitting_a_package#Step_7

You'll need to change `"branch": "master"` to `"tags": true` and use [tags](http://git-scm.com/book/en/v2/Git-Basics-Tagging) to create releases and push valid [semantic version](http://semver.org/) tags to your repo. As mentioned [in the docs](https://packagecontrol.io/docs/submitting_a_package#Step_4), we no longer accepts branch-based releases for new packages.

wbond:
Duplicate of #5574

PR #5574, opened 2016-06-06
Dot Temple

sophatvathana:


ahmadawais:
@sophatvathana I think your PR is failing because `"branch": "master",` is not allowed now. You should only use tags. 

The tests for your package didn't pass: https://packagecontrol.io/docs/submitting_a_package#Step_7

You'll need to change `"branch": "master"` to `"tags": true` and use [tags](http://git-scm.com/book/en/v2/Git-Basics-Tagging) to create releases and push valid [semantic version](http://semver.org/) tags to your repo. As mentioned [in the docs](https://packagecontrol.io/docs/submitting_a_package#Step_4), we no longer accepts branch-based releases for new packages.

repository/t.json
```diff
@@ -2709,3 +2709,5 @@
 		}
 	]
 }
+	]
+}
```
wbond: These should not be here, but instead should be at line 1260.

repository/d.json
```diff
@@ -1248,6 +1248,16 @@
...
 			]
+		},
+		{
+			"name": "DotTemple",
+			"details": "https://github.com/sophatvathana/temple-sublime.git",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "master",
```
wbond: This needs to be removed, you can't release from a branch and from tags.

repository/d.json
```diff
@@ -1248,6 +1248,16 @@
 					"branch": "master"
 				}
 			]
+		},
+		{
+			"name": "DotTemple",
+			"details": "https://github.com/sophatvathana/temple-sublime.git",
```
wbond: Remove the `.git` suffix from the URL.

wbond:
The following files should be removed from your repository:
 
 - https://github.com/sophatvathana/temple-sublime/blob/master/packages.json
 - https://github.com/sophatvathana/temple-sublime/blob/master/TEMPLE.sublime-package
 - https://github.com/sophatvathana/temple-sublime/blob/master/repository.json

This folder is unnecessary:
 
 - https://github.com/sophatvathana/temple-sublime/tree/master/.vscode

Also, you seem to have two copies of the `TEMPLE.tmLanguage` in your repo.

FichteFoll:
@sophatvathana ping

wbond:
@sophatvathana This is a final reminder you need to make changes before we can accept your PR. :smile:

PR #5575, opened 2016-06-06
Add addons for the Boxy Theme

oivva:
Adds SublimeLinter theme and addons for changing font size of the inputs and the console panel text.

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -927,6 +927,78 @@
 			]
 		},
 		{
+			"name": "Boxy Theme Addon - Font Face",
+			"details": "https://github.com/oivva/boxy-addon-font-face",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Are you sure you need `.no-sublime-plackage` files for these? I don't think so.

oivva:
Agree, I'll delete `.no-sublime-package` files for the all addons, except **Boxy Theme Addon - Font Face**, because user should customize font family manually.

FichteFoll:
Just remember that any update of the package will erase all user changes when the package is extracted.

oivva:
After some experiments, I can say that `.no-sublime-package` should be removed only from **Boxy Theme Addon - Linter Theme**, because font size and face isn't overriding when the packages aren't extracted.

PR #5577, opened 2016-06-07
WhizzML package: DSL language syntax and auto-complete

osroca:


FichteFoll:

`.sublime-syntax` support has been added effectively with build 3092. Please update your version requirement.

osroca:
Request updated.
I've retagged our first version as v0.1.0, I guess that it is what you meant with your comment about Semantic Versions, if it is not, please can you clarify it?

PR #5578, opened 2016-06-08
Add Log Highlight

poucotm:


FichteFoll:
You don't need to add an on_change listener for settings if the only thing you do is get a new reference to the settings object. It is updated anyway when changes are detected.

poucotm:
Thank you. I've removed on_change listener for all my plugins.

PR #5592, opened 2016-06-13
Adding TooltipLSL

Makopo:
A Sublime Text tooltip reference system for LSL/OSSL, which uses Sublime 3070+ show_popup() function.
The project is [here](https://github.com/Makopo/sublime-text-tooltip-lsl). Thanks.

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -2078,6 +2078,17 @@
 			]
 		},		
 		{
+			"name": "TooltipLSL",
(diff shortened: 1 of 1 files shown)


FichteFoll:
In archived packages, you must use the `sublime.load_resource()` API instead of opening files directly.

Makopo:
Thank you. I fixed both issue now.

PR #5593, opened 2016-06-13
updated the repository since plugin ownership changed (https://news.ycombinator.com/item?id=11896851)

nanch:


FichteFoll:
You have to push a valid [semantic version](http://semver.org) [tag](http://git-scm.com/book/en/v2/Git-Basics-Tagging) (with optional leading `v`) to your repo. The latest release recorded on packagecontrol.io is `803.9.1`, so yours would need to be higher.

nanch:
Thank you for explaining that a git tag needed to be applied to the commit.

I have updated a more recent commit with a tag of v803.10.2 which is greater than the previous version 803.9.1.

Please let me know if there's anything else I can do and thank you for your assistance.

nanch:
Thank you for the quick merge, much appreciated by all of the developers that depend on this. Cheers!

PR #5594, opened 2016-06-14
Package MBXTools, for MathBook XML authors

daverosoff:
[MathBook XML](https://github.com/rbeezer/mathbook) extensions and customizations.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -742,6 +742,18 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
`ctrl+l` is a default binding, but I'll let it slide because you re-add the command with a slightly different binding. Just make sure to mention this in the readme!

PR #5597, opened 2016-06-16
Update t.json

russianidiot:


FichteFoll:
It appears you added another package in this PR.

The proper PR practice is to create a new branch in your fork from the upstream master branch, push that to github and create a PR from it. You can then make changes to the PR by pushing commits to that branch.

Note that neither of the tags you pushed point to the latest commit. Also note that `0.0.2` will be chosen over `v0.0.1`, even though 0.0.1 points to a more recent commit.

Also, please fix the JSON.

wbond:
@russianidiot Ping, don't forget you need to make some changes to this PR

PR #5598, opened 2016-06-17
add Meav package

linlymatsumura:
Meav is made for coding CSS from HTML.
Extract HTML class attribute and format CSS class selector.


FichteFoll:

linlymatsumura:
thanks. i fixed release type. please review again.

PR #5607, opened 2016-06-19
Add "microbit Build System"

stefanbates:
A BBC micro:bit build system for flashing a micropython file to a BBC micro:bit.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -1025,7 +1025,7 @@
 			]
 		},
 		{
-			"name": "micro:bit Build System",
+			"name": "microbit Build System",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Does your build system work?

I would expect it has to be either:
```
"cmd": ["python", "${packages}/microbit Build System/microbit.py", "$file" ]
```
or
```
"shell_cmd": "python \"${packages}/microbit Build System/microbit.py\" \"$file\""
```

You should also mention that your build system requires Python 2 and `python` to be mapped to that.

nanalan:
> Does your build system work?
It does.

PR #5608, opened 2016-06-19
Add package for CNC BoschRexroth MTX language support

deathaxe:
Please add my new package for syntax highlighting of NC programs written for the BoschRexroth MTX numerical control to the package repository right next to the package for SINUMERIK 840D.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1287,6 +1287,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
I forgot to check this, but your package should not have listedt the "auto-complete" tag since those are usually context-aware and need to be provided by plugins.

montycarlos:
deathaxe:master

PR #5610, opened 2016-06-20
Adding GitHub syntax theme to package control

jonrohan:


buildersbrewery:
What is the enhancement by adding this in comparison to the two examples below?

* [Theme - Primer](https://packagecontrol.io/packages/Theme%20-%20Primer)
* [Materialize](https://packagecontrol.io/packages/Materialize) (see Primer option there)

I get it you're Github staff and you'd like to spread your style via a plugin you have access to.

But if that's your main drive, wouldn't creating a repo for **specifically for Sublime Text** itself not make more sense? Compare the two examples above, where the Materialize plugin also delivers an accompanying theme along with the color scheme.

Good examples for this approach would be [equinusocio/material-theme](https://github.com/equinusocio/material-theme) and even more so [briles/gruvbox](https://github.com/briles/gruvbox), which has a modular approach and includes scripts to build from source files.

buildersbrewery:
Was asking out of curiosity, no offense implied. Why did you close this pr?

jonrohan:
Sorry, no offense taken. The main reason I submitted was that we just released the stylesheets that we do use for the syntax highlighting. There's been a few GitHub-like styles to be shipped out there, but they've been reverse engineered.

The TextMate theme file was something we've had for a while, just not open source. So I was going to submit it, as is, and start working on improvements. But, the reason I closed it was, it's not even in the ballpark of the material-theme ui themes of the ones you shared. Working on making it at the same level is not a priority for me right now. So I'll close this and let someone organically add/use the official styles.

PR #5611, opened 2016-06-20
Add MozillaQuery

saadq:


FichteFoll:
You should take a look at `urllib.parse.quote_plus`/`urllib.parse.quote`.

saadq:
I don't really know much Python, so thanks for that. I'll check it out.

PR #5614, opened 2016-06-21
add a mobile ui framework for sublime

tangbindu:


tangbindu:
add a mobile ui framework for sublime

tangbindu:
add a mobile ui framework for sublime      fixed a bug

tangbindu:
add a mobile ui framework for sublime fixed a bug    help me

saadq:
The tests are failing because you need to keep the packages in alphabetical order. Your package which is named **CUI** should go after the **CUE Sheet** package and before the **Curly Syntax Definition** package. 

Also, you'll need to change `"branch": "master"` to `"tags": true` and use tags to create releases and push valid semantic version tags to your repo. As mentioned in the docs, branch-based releases are no longer accepted for new packages.

FichteFoll:
What @saadq said.

Additionally, instead of providing completions via plugin code you could as well define a `.sublime-completions` file, since your plugin doesn't do anything dynamic at all. Read more here: http://docs.sublimetext.info/en/latest/reference/completions.html

wbond:
@tangbindu This is just a reminder that your PR needs work before we can accept it. :smile:

PR #5616, opened 2016-06-22
Add package gerardroche/five_easy_color_schemes

gerardroche:


FichteFoll:
What is the motivation behind this?

Monokai is part of the default distribution, as is Solarized. There are no other of the "five schemes" you advertise.

gerardroche:
I haven't decided what the other schemes will be yet. It will be a maximum of five light/dark schemes. 

The motivation is to group several schemes in one package and make them good, polished, and well tested. The default color schemes are not very good. I'm not happy with any alternatives. I was using Monokai Extended, which was good enough, but there have been too many changes recently. The Monokai and Solarized schemes in this package will be the best schemes available for sublime.

FichteFoll:
> The Monokai and Solarized schemes in this package will be the best schemes available for sublime.

My scheme will be better :stuck_out_tongue_winking_eye: 

I suggest you go over this motivation in your readme somewhere.

PR #5624, opened 2016-06-24
chore: updated name & repo path of the chatscript syntax

kuzyn:
Changed the repository path (user name) and recapitalized the name


FichteFoll:

kuzyn:
Oops. Added it

PR #5628, opened 2016-06-27
Add Git Time Metric package

mschenk42:


FichteFoll:
You'll need to change `"branch": "master"` to `"tags": true` and use [tags](http://git-scm.com/book/en/v2/Git-Basics-Tagging) to create releases and push ***valid*** [**semantic version**](http://semver.org) tags to your repo.

mschenk42:
How long does it take to become available to be installed via Package Control?  I see it on packagecontrol.io but I can't find it when trying to install via Package Control from within Sublime.

wbond:
If you already have it installed, it won't show as available to install.

mschenk42:
I'v tried a fresh install within a VM of Sublime and Package Control.  I still can't get it to show up under Install Package.  Would you mind trying to see if you can find it using Package Install?  I tried searching on GitTimeMetric and gtm.

wbond:
You released it as a beta, so you'll have to add the package name to the `install_prereleases` setting to get it to show in the Install list.

Unless you only want this release to go to a limited number of testing users, you should probably not release it as a prerelease (with a `-` in the tag name). Perhaps `0.9.0` would be better.

mschenk42:
Well I guess it's ready to go anyways.  I'll just update the tag in the repo.  Do I need to submit another pull request?

wbond:
Nope, just push a non-prerelease tag to the repo. The crawler runs once an hour, so it should become available soon afterwards.

PR #5632, opened 2016-06-27
SortList -- Cross-language plugin for sorting items in comma-separated lists

kylebebak:
The GitHub repo has a demo gif that shows how it works:

<https://github.com/kylebebak/sublime_sort_list>

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1979,6 +1979,16 @@
 			]
 		},
 		{
+		  "name": "SortList",
+		  "details": "https://github.com/kylebebak/sublime_sort_list",
+		  "releases": [
+		    {
+		      "sublime_text": "*",
+		      "tags": true
+		    }
+		  ]
+		},
+		{
 			"details": "https://github.com/Kentzo/SortLinesByColumn",
 			"releases": [
 				{
--- a/repository/s.json
+++ b/repository/s.json
@@ -1979,21 +1979,21 @@
 			]
 		},
 		{
-		  "name": "SortList",
-		  "details": "https://github.com/kylebebak/sublime_sort_list",
-		  "releases": [
-		    {
-		      "sublime_text": "*",
-		      "tags": true
-		    }
-		  ]
-		},
-		{
 			"details": "https://github.com/Kentzo/SortLinesByColumn",
 			"releases": [
 				{
 					"sublime_text": "<3000",
 					"branch": "master"
+				}
+			]
+		},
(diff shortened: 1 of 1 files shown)


FichteFoll:
```python
    try:
        l = eval(s)
    except:
        raise SyntaxError("List could not be parsed")
    if not hasattr(l, '__iter__'):
raise SyntaxError("Not iterable")
```

[Really?](https://github.com/kylebebak/sublime_sort_list/blob/a84d5c750881649ede8b36ef5d60b72363c58be0/sorted_string.py#L26)

Just wrap the string in brackets and use `json.loads` instead, that's safer.
Also, `SyntaxError` is not the appropriate exception type you raise in this situation. It would be `ValueError` (= the value provided cannot be worked with).

kylebebak:
@FichteFoll 
Thanks for the advice on raising `ValueError` instead of `SyntaxError`. The `1.0.1` release is now up with this change.

Wrapping the string in brackets and calling `json.loads` isn't equivalent to calling `eval`. This would require me to rewrite the plugin and it probably lose some desired functionality.

I understand the danger of running eval on untrusted strings, but this plugin is only going to be invoked by the user.

Here's an example of a plugin that I use, and that I installed via package control, that invokes `eval`: <https://github.com/jbrooksuk/Sublime-Evaluate>

FichteFoll:
>I understand the danger of running eval on untrusted strings, but this plugin is only going to be invoked by the user.

Still, the least you can do is clarify that you are in fact *evaluating* the selection as Python code. A package named "Evaluate" makes this very clear from the beginning, but yours doesn't mention it at all.

Furthermore, I don't really think you actually lose a lot of "desired functionality", unless you can explain some of that to me. The only notable differences I can think of are:

- `null` instead of `None`
- doesn't evaluate expressions like `6 * 10` (not advertised)
- doesn't handle special Python-flavored strings like `r"raw\strings"` (not advertised)

These would be features if your package said it "evaluates a list of Python expressions and sorts the values", but that's not what it says. Declaring it as being cross-language while everything needs to be a valid Python expression is also questionable.

kylebebak:
OK, using `json.loads` now instead in release `1.1.0`, I thought it would be more work to switch than it was.

Anything else, or do you think it's ready to go?

kylebebak:
@FichteFoll 

Big thanks for vetting the package and merging it! You taught me something useful with `json.loads` =)

PR #5638, opened 2016-06-29
change then 'Lua Jump Definition' name

RuiZhuo:


wbond:
You should delete the `packages.json` file from your repository and add information to your readme describing what exactly the package does, and how to use it.

RuiZhuo:
I have been revised according to your opinion.

PR #5639, opened 2016-06-29
add AlgoLang package

qraynaud:
simple syntax highlighter for an algorithmic language used in lessons in France

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -363,6 +363,18 @@
 			]
 		},
 		{
+			"name": "AlgoLang",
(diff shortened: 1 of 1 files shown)


wbond:
You'll need to create at least one tag in the repo that is a [semantic version number](http://semver.org). A good starting point is `1.0.0`.

qraynaud:
Done. I was sure I had already created one. My bad. Thanks.

qraynaud:
@wbond : any news on this ? thanks

PR #5642, opened 2016-06-29
Adding Angular-io-Code Color Scheme

NoHomey:


wbond:
You'll need to remove the addition to channel.json. That is duplicating the work of adding it to `a.json`, and is really only for people who know they need to do it that way.

Also, be sure to remove the packages.json from your repository.

NoHomey:
Both are done.

PR #5644, opened 2016-06-30
Added JAS-Sublime Java Assembly

cloudcrypt:


wbond:
This is definitely not for all version of Sublime Text (as you indicated with the `*`) since you are using a `.sublime-syntax` file, and those are only for build 3084 or newer.

PR #5646, opened 2016-07-01
This is a package for Caliper Corporation developed scripting language : "GISDK"

xuke-fang:


wbond:
You would need to unzip the files from the `.sublime-package` file and place them in the root of the repository. As it is current configured, it will not do anything when installed.

xuke-fang:
I have unzipped the files. Do I have to do something else? Thanks

wbond:
The `.tmPreferences` files seem to have an odd character in the filename. It shows as a box on my machine. Perhaps just remove that char? The part before `.tmPreferences` doesn't really matter, as long as it can be represented on all filesystems. Thus, ASCII is the best choice.

xuke-fang:
Hi wbond, I have fixed the names. Please let me know if there are any other problems. Thanks

wbond:
My only other recommendation is to add some info to the readme to describe what the package contains and what it does.

xuke-fang:
Thanks for merging the pull request. But when I try to install the package through package control and upzip the .sublime-package file. The top level still contains the old unzip GISDK.sublime-package. But I have already updated the git repository.  Am I missing something here? Thanks

PR #5649, opened 2016-07-03
Add Language - German package

gmmedia:


wbond:
You'll need to add a [semver](http://semver.org) tag to your repo before this can be installed. I suggest something like `1.0.0`.

wbond:
Just so you know, only UTF-8 encoded dictionaries are supported by Sublime Text.

PR #5655, opened 2016-07-06
add github issue

divinites:
I developed a plugin to directly manipulate github issues in sublime text buffer, with markdown extension.


divinites:
Since FitchFoll only reviews packages on Mondays, so I changed the package name from "Github Issue" to "GithubIssue" in order to avoid possible problems and delay :)

PR #5657, opened 2016-07-06
Add dependency Beautiful Soup 4.4.1

jlegewie:
Add Beautiful Soup 4.4.1 to dependencies. https://www.crummy.com/software/BeautifulSoup/

### DIFF
--- a/repository/dependencies.json
+++ b/repository/dependencies.json
@@ -45,6 +45,21 @@
 			]
 		},
 		{
+			"name": "bs4",
+			"load_order": "51",
+			"description": "Beautiful Soup is a Python library for pulling data out of HTML and XML files - https://www.crummy.com/software/BeautifulSoup/",
+			"author": "jlegewie",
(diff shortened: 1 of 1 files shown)


divinites:
Great!  I want beautiful soup as a dependencies for months!

jlegewie:
Good to hear that! I realized that I am not entirely sure whether this actually works. Can anyone test it? Otherwise, I will post in the forum and ask for advice.

divinites:
I want to use BS4 to enhance my quickmail plugin, but it is unlikely to be done recently.  Posting in the forum seems to be a good idea anyway.

jlegewie:
Ok, everything should work now. I have tested it with a private repository and Beautiful Soup seems to be running perfectly!

wbond:
Your tag is going to need to be [semver](https://semver.org)-compatiable. `1.0` won't work, but `1.0.0` will.

jlegewie:
Sorry for that. Just changed it...

PR #5659, opened 2016-07-06
Add WordPress Completions

23r9i0:


wbond:
Please read the directions start at [https://packagecontrol.io/docs/submitting_a_package#Step_4](Step 4).

23r9i0:
Create Tags for package

wbond:
Adding a `-` to a semver version makes it a prerelease. Users won't see prereleases unless they've specifically added the package to the setting https://packagecontrol.io/docs/settings#setting-install_prereleases.

23r9i0:
Upps, Sorry by not reading all documentations, before publish this package, I would say this right now, I just wanted to keep the version of WordPress as the package and I added the underscore to monitor possible changes in the package, Now I will otherwise.

wbond:
Unfortunately `4.5.3.0.1` isn't a semantic version number either.

Semver requires three numbers, with an optional prerelease suffix (starting with `-`) or optional build info (starting with a `+`).

In your case, I believe it would need to be `4.5.3+0.1` or `4.5.3+1`.

23r9i0:
Ok, already change the formatting for tags, thanks

PR #5660, opened 2016-07-07
Update EasyClangComplete

niosus:
I only account for sublime text 3, but this is not reflected in the setup. It will be from now on.

Question: can I drop the st3-prefix for tags now?


wbond:
Yes, you can drop `st3-` and start creating non-prefixed version tags. However, Package Control won't detect them until you send a PR changing the info to `"tags": true`.

niosus:
Thanks! Opened a new pull request for that.

PR #5662, opened 2016-07-08
add a plugin for apicloud-wifisync

zhou-xinghai:
The plugin is from APICloud for wifi sync.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -980,6 +980,17 @@
 			]
 		},
 		{
+			"name": "APICloudWifiSync",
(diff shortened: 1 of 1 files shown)


wbond:
You will need to create a [semver](http://semver.org) tag on the repo in order for users to be able to install it.

Since you have a number of files that need to be executed in the package, you'll need to add a file named `.no-sublime-package` to the root of the repository. Otherwise, the files will be zipped up and installed as `APICloudWifiSync.sublime-package`.

zhou-xinghai:
thanks for your remind. I have created a new tag and add the file as you required. if it's ok,then pls merge the pull request. thanks.

PR #5664, opened 2016-07-08
Add syntax plugin for QL

EdoDodo:
Hi,

I'd like to add my syntax plugin for Semmle's QL (see [here](https://semmle.com/products/semmle-ql/)) to Package Control please.

Thanks,
Edoardo

### DIFF
--- a/repository/q.json
+++ b/repository/q.json
@@ -24,6 +24,17 @@
 			]
 		},
 		{
+			"name": "QLSyntax",
+			"details": "https://github.com/EdoDodo/QLSyntax",
+			"labels": ["language syntax"],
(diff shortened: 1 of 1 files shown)


wbond:
You'll need to limit this to build 3084 or newer since that is when support for `.sublime-syntax` was added.

You'll also need to create a [semver](http://semver.org) tag on the repo so users can install it. I suggest starting with `1.0.0`.

EdoDodo:
Have done both, I've gone for 0.1.0 rather than 1.0.0 as an initial version to reflect this is still in early development and subject to changing, as recommended by the page you linked.

PR #5668, opened 2016-07-09
Added Marc Mentat Proc Highlighting package

WesleyPeijnenburg:
Provides Syntax highlighting for Marc Mentat procedure files. Also
highlights *some* syntax errors.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -270,6 +270,17 @@
 			]
 		},
 		{
+			"name": "Marc Mentat Proc Highlighting",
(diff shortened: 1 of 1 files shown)


wbond:
This will only work for builds 3084 and newer.

You probably also want to add a `name` key with the name of the syntax near the beginning of the syntax file.

wbond:
To be clear, to mark it as only compatible with 3084, you need to change `"sublime_text": "*"` to `"sublime_text": ">=3084"`.

WesleyPeijnenburg:
The name key and compatibility lines are now included. Thanks for the feedback!

PR #5672, opened 2016-07-11
Add TtcnComplete

HuiMi24:


wbond:
You haven't created any [semver](http://semver.org) tags on the repo yet, so no one will be able to install it. I would recommend starting with something like `1.0.0`.

HuiMi24:
Sorry, I forgot to add the tag.
The tag 1.0.0 added.

PR #5673, opened 2016-07-11
Bitbake syntax plugin

Driim:
Signed-off-by: Dmitriy Falko <DFalko89@mail.ru>


wbond:
You'll want to change `sublime_text` to `>=3084` since `.sublime-syntax` was introduced with that build.

Driim:
Changed

PR #5675, opened 2016-07-12
Add wkhtmltopdf plugin

jrappen:
* see commits tab for details

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -4401,6 +4401,22 @@
 					"tags": true
 				}
 			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


wbond:
You'll want to:

 1. Delete `sublime-` from the name, and then move into the appropriate .json file
 2. Delete the `author`, `issues` and `readme` values since those are the defaults
 3. Ideally delete the `description` value and use the description field of the repo on GitHub
 4. Delete the `platforms` value in `releases` since the default is all platforms

jrappen:
@wbond Done, see #5698

PR #5676, opened 2016-07-12
add gbk support for sublime >3000

chengsu:
add GBK encoding support for sublime_text 3

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -108,6 +108,16 @@
 			]
 		},
 		{
+			"name": "GBK Support",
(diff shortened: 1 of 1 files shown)


wbond:
This seems like the functionality is already covered by https://packagecontrol.io/packages/ConvertToUTF8. Does this package do something different?

chengsu:
ConvertToUTF8 sometimes get you messy text for GBK encoding. This package could be more reliable.

wbond:
My only other suggestion would be to describe how to use the package in the readme.

PR #5679, opened 2016-07-13
Add Theme - Seti Monokai

poucotm:


wbond:
Did you test the package without a `.no-sublime-package` file? Ideally packages should be created as a `.sublime-package` since it allows users to override things.

poucotm:
In my local computer, it works. And many old packages use like that. I don't know differences. Is there any difference of reading speed between them (compressed files vs. non-compressed files) ? Anyway, I've removed .no-sublime-package and committed again.

wbond:
A `.no-sublime-package` will always work, but it makes it impossible for end-users to override or tweak parts of a package. This functionality was added in Sublime Text 3.

The only reason you really need a `.no-sublime-package` file is if you have `.dll`, `.so`, `.exe`, `.sh` or other executable files in the package that need to be executed.

poucotm:
Thank you for the information ~

PR #5682, opened 2016-07-14
Add LLVM TableGen package

whitequark:


wbond:
The current tag you have (`v3.8`) is not a valid [semver](http://semver.org), so no one will be able to install the package. I suggest `3.8.0` or `v3.8.0`.

whitequark:
`v3.8.0` pushed

PR #5687, opened 2016-07-16
add NormalizeLineLength to package control channel

kylebebak:
A plugin to normalize/limit line length for multiple lines of selected text, e.g. for comments, docstrings.

[Here's the repo](https://github.com/kylebebak/sublime_normalize_line_length). The gif gives a good idea of how it works, as do the tests!

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -540,6 +540,16 @@
 			]
 		},
 		{
+			"name": "NormalizeLineLength",
+			"details": "https://github.com/kylebebak/sublime_normalize_line_length",
+			"releases": [
+				{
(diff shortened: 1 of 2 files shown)


wbond:
This seems very similar to https://packagecontrol.io/packages/Wrap%20Plus. Is there some missing functionality you could possibly collaborate on?

kylebebak:
@wbond 

I wasn't aware this existed, I looked for plugins with "line", and "normalize", and "limit" in their names.

I installed Wrap Plus and played with it a bit. It looks like it does everything mine does and then some. Thanks for reviewing, I'll close this pull request.

PR #5688, opened 2016-07-16
added python dependency PyWin32

randy3k:
It is an attempt to address https://github.com/SublimeText/Pywin32/issues/3 to bring pywin32 to package control.

The binary files of PyWin32 are updated to version 220.

### DIFF
--- a/repository/dependencies.json
+++ b/repository/dependencies.json
@@ -364,6 +364,21 @@
 			]
 		},
 		{
+			"name": "python-pywin32",
+			"author": "randy3k",
+			"load_order": "50",
+			"description": "Pywin32 module",
+			"issues": "https://github.com/randy3k/sublime-pywin32/issues",
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


daviwil:
Would be really helpful to have this added to Package Control!  Nice work @randy3k

repository/dependencies.json
```diff
@@ -364,6 +364,20 @@
...
+			"name": "python-pywin32",
+			"author": "randy3k",
+			"load_order": "50",
+			"description": "Pywin32 module",
+			"issues": "https://github.com/randy3k/sublime-pywin32/issues",
+			"releases": [
+				{
+					"base": "https://github.com/randy3k/sublime-pywin32",
+					"tags": true,
+					"sublime_text": ">=3000"
```
wbond: You probably want `"platforms": ["windows"],` here, right?

randy3k:
thank you for the catch. Fixed and rebased.

PR #5689, opened 2016-07-16
added linux dependency xdotool

randy3k:
This provies the [xdotool](https://github.com/jordansissel/xdotool) binary to Sublime Text, so developers could use this to automate things on linux without asking users to install xdotool.

### DIFF
--- a/repository/dependencies.json
+++ b/repository/dependencies.json
@@ -614,6 +614,21 @@
 			]
 		},
 		{
+			"name": "xdotool",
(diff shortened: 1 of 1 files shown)


wbond:
This should probably be restricted to linux via the `platforms` key, right?

randy3k:
thank you for the catch. Fixed and rebased.

randy3k:
I don't understand why it says that "this branch has conflicts..."

PR #5692, opened 2016-07-18
Added new package SendToJsonBlob

abhishekchavan:
This package allows user to upload a json file to jsonblob.com to easily share with peers

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -885,8 +885,8 @@
 			"details": "https://github.com/abhishekchavan/sendtojsonblob",
 			"releases": [
 				{
-					"sublime_text": "*",
-					"branch": "master"
+					"sublime_text": ">=3000",
+					"tags": true
 				}
 			]
 		},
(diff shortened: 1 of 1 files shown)


wbond:
You should follow the directions, starting with [Step 4](https://packagecontrol.io/docs/submitting_a_package#Step_4). In short, we no longer accept new packages using branch-based releases.

Also, https://github.com/abhishekchavan/sendtojsonblob/blob/master/SendToJsonBlob.py#L6, will make this not work properly in Sublime Text 2. Did you test it with ST2? Considering the number of ST2 users, it may make more sense to just mark it as only compatible with ST3.

abhishekchavan:
I have made the required changes to make it available only for ST3 and also for tagged releases.

wbond:
A semver version number needs to have three parts. So you'll need `1.0.0` or `v1.0.0`.

abhishekchavan:
Sorry about that. This is my first pull request. Added a new tag 1.0.0

PR #5697, opened 2016-07-19
added react_native_suggestion_snippet

polodev:


wbond:
You need to fix the test failures, and we no longer support branch-based releases. I would recommend following the directions starting with [Step 4](https://packagecontrol.io/docs/submitting_a_package#Step_4).

PR #5698, opened 2016-07-19
add wkhtmltopdf

jrappen:
fixes #5675


wbond:
This should be removed from your repo: https://github.com/jrappen/sublime-wkhtmltopdf/blob/master/packages.json

PR #5699, opened 2016-07-19
Adding Creative Crocodile package reference

Abban:
This is a Sublime Text plugin to generate placeholder names in the style of MGSV's soldiers. To generate one hit `ctrl-shift-m`.

## Examples
```
Gusty Goose
Indecorous Insect 
Staunch Squirrel 
Naive Nightingale 
Conservative Caterpillar 
Accommodating Alligator 
Euphoric Elephant 
Gifted Gerbil 
Zealot Zebu 
Obedient Opossum 
Deferential Donkey 
Responsive Rattlesnake 
Confident Chinchilla 
Innocent Iguana 
Yourself Yeti 
Faithful Fish 
```

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2936,7 +2936,7 @@
 			]
 		},
 		{
-			"name": "Creative Crocodile",
+			"name": "CreativeCrocodile",
 			"details": "https://github.com/Abban/CreativeCrocodile",
 			"releases": [
 				{
--- a/repository/c.json
+++ b/repository/c.json
@@ -2936,6 +2936,16 @@
 			]
 		},
 		{
+			"name": "Creative Crocodile",
+			"details": "https://github.com/Abban/CreativeCrocodile",
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


wbond:
In ST3, your package will be zipped up and installed as a `.sublime-package` file, so reading the `.json` files via `open()` won't work.

Instead, you should use `sublime.load_resource("Packages/Creative Crocodile/example.json")` to get the contents, and then use the `json` module to parse it. Note that you requested the package have a space in the name, whereas your code expects there to be no space.

http://www.sublimetext.com/docs/3/api_reference.html

Abban:
Thanks for the advice Will.

I updated the package name and changed the json read/parse code to the following:
```
data_file = sublime.load_resource("Packages/CreativeCrocodile/adjectives.json")
data = sublime.decode_value(data_file)
```
I hope this is correct.

wbond:
If it is working when the package is extracted, it should work when it is packed also.

PR #5701, opened 2016-07-20
Add Intel HEX syntax

seppestas:
Added a package for highlighting [Intel HEX format](https://en.wikipedia.org/wiki/Intel_HEX).

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -635,7 +635,7 @@
 			"labels": ["language syntax", "hexadecimal"],
 			"releases": [
 				{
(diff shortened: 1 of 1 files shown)


wbond:
`.sublime-syntax` support was not added until build 3084, thus the `sublime_text` key should be `>=3084`.

seppestas:
Ah, I was not aware of this. I'll update this PR.

seppestas:
Do you by any chance know how I can make the syntax_test file work when it is installed using Package control?

PR #5704, opened 2016-07-21
Add Koala KSP Library - Snippets to repo

magneto538:
https://github.com/magneto538/Koala

### DIFF
--- a/repository/k.json
+++ b/repository/k.json
@@ -252,6 +252,18 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


wbond:
The `author` key can be removed since that is your GitHub user name.

You need to create a semver tag - right now you have none, so no one will be able to install: https://github.com/magneto538/Koala-snippets/releases.

magneto538:
@wbond done.

PR #5709, opened 2016-07-22
Removed Intel XDK

likhter:


wbond:
Hi @likhter, you don't seem to be a member of the organization that the repository is owned by. I see you've worked on the repo, but I want to make sure that removals aren't sent in by people not authorized to do so.

Can I get a confirmation from you, @xmnboy?

On a side note, I'm just curious why you are removing the package? Are you no longer planning on maintaining it?

xmnboy:
I'm a colleague of @likhter at Intel and agree with removal of the package -- we both work on the Intel XDK project, which is what that package supports.

We are trying to focus our engineering efforts and have had to make some hard decisions about which projects we will continue to provide ongoing support. Unfortunately, this editor package is one of those projects that fell below the continued support line. Given the ongoing changes to the Intel XDK project, keeping it around would guarantee a continued support effort, so we have no choice but to remove it from circulation.

wbond:
Ok, thanks!

PR #5710, opened 2016-07-23
Add Skins support package to Sublime Text

deathaxe:
Inspired by Theme-Switcher and QuickThemes I created a little script to change both 'color_scheme' and 'theme' with one command from command pallet.

The intention with this package is to enable theme developers to provide a *.sublime-skins file in their package, whose content is displayed in the "Select Skin" command pallet or let the user save its current settings as skin.

No need to manually edit settings files any more in order to quickly change the design of Sublime Text.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1608,7 +1608,7 @@
 			"labels": ["themes", "utilities"],
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": ">=3000",
 					"tags": true
 				}
 			]
--- a/repository/s.json
+++ b/repository/s.json
@@ -1603,6 +1603,17 @@
 			]
 		},
 		{
+			"name": "Skins",
+			"details": "https://github.com/deathaxe/sublime-skins",
+			"labels": ["themes", "utilities"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Skript",
 			"details": "https://github.com/piratjsk/ST3_Skript",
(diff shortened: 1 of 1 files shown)


wbond:
It is probably best not to use the file extension `.sublime-*` since those are used by Sublime Text itself.

You also didn't create a tag: https://github.com/deathaxe/sublime-skins/tags.

Additionally, you use some API functions which are only available in Sublime Text 3, so you should at the least mark this as `"sublime_text": ">=3000"`.

deathaxe:
1. Do you have any suggestions for a good extention without breaking some naming conventions? All config files I found have sublime-... or tm... or json extentions and I didn't want to start something completely new ... ok somhow I did. But as I want to allow more than one skin-file in a folder being merged as files like the Main.sublime-menu the extention is important. Maybe something like *.themeset, *.skinset or *.skins would be ok? ... or even maybe a feature request to make a functionality like that a default capability of Sublime Text? ;-) This plugin was born due to the lack of changing the whole UI (theme and color scheme) quickly without editing the config files by hand. 

2. TAG - ooups, my false, will be fixed with extentions.

3. I never used Sublime Text 2, so I am not aware of which APIs changed, but mark it for ST3 will be done.

wbond:
I think for item 1 you could to something like `.st-skins` or `.skins`. I think either one is unlikely to run into a conflict with another package, since these files are Sublime Text-specific.

deathaxe:
I really understand this possiblie conflict issue with naming files like .sublime-... . I even thought about this on my own. So I will use the .skins extention. It looks a bit cleaner than the first one.

PR #5716, opened 2016-07-26
Add file icons package

oivva:
Details: https://forum.sublimetext.com/t/sublime-text-3-file-icons-in-sidebar/21134

I think this is the best solution as possible currently.


oivva:
@wbond I publish it as package. I think there is no need in dependency. User can install this package if he want enhanced icon support.

PR #5720, opened 2016-07-29
add coati package

st4ll1:


wbond:
You should remove the `.no-sublime-package` file from your repo. That is for situations where an executable is included in the package, and it needs to be unzipped so it can be run.

Also, you need to create a [semver](http://semver.org) tag on your repo so it will appear to users as installable.

st4ll1:
thanks for the feedback, 
 
`.no-sublime-package` is removed and repo is tagged as `0.1.0`

PR #5721, opened 2016-07-29
Add SublimeMagic

mreq:
I know that "Sublime" in plugin name is frowned upon, but just "Magic" sounds bad ;)

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -3464,6 +3464,16 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


wbond:
Perhaps just improve the description of the repo on GitHub since that is shown on packagecontrol.io and in the list of packages when installing?

mreq:
did that, thanks

PR #5735, opened 2016-08-05
add Mac Set Lang package

oyyq99999:
set the LANG for ST running on OS X to prevent some encoding issues(e.g., unicode literals for python3).

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -23,6 +23,17 @@
 			]
 		},
 		{
+			"name": "Mac Set Lang",
+			"details": "https://github.com/oyyq99999/SublimeMacSetLang",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": ["osx"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "MacDown App Menu",
 			"details": "https://github.com/diimdeep/sublime-text-macdown-app-menu",
 			"labels": ["markdown", "preview", "editor", "macdown.app"],


wbond:
Can you provide an example of a bug that this solves?

The reason I ask is that there:

 1. May be a better way to solve the issue
 2. It may be something we need to solve in ST itself

oyyq99999:
I wrote a python script which outputs non-ascii characters, but it fails with the LANG set as the default.

wbond:
Can you paste the output?

wbond:
I just tried the following in ST3 and it worked no problem on El Capitan:

```python
# coding: utf-8

print(u'Résumé')
```

oyyq99999:
Sorry it's not non-ascii, it should be non-latin.
for example `√`, the symbol for square root

wbond:
I just wanted to leave a note here that I am planning on looking at this at some point soon to hopefully resolve any issue that exists in Sublime Text itself.

oyyq99999:
Actually it's not about Sublime Text, it's about OS X, see [here](https://github.com/int3h/SublimeFixMacPath#fix-mac-path)

wbond:
Ah! There should actually be an official fix for that in Sublime Text in the next release. The plan is to load your environment variables from your default shell.

FichteFoll:
I take it this package is not needed and bound to be fixed by ST itself in a future release. Thus, I'm closing this.

PR #5737, opened 2016-08-05
Add the ExtendedTemplates package

FailedCode:
A package to generate folder structures and files with content from multiple sources.

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -1223,6 +1223,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


wbond:
Ideally `.no-sublime-package` would be removed and the code would use [`sublime.load_resource()`](http://www.sublimetext.com/docs/3/api_reference.html#sublime).

wbond:
Please open a new PR for this if you get around to making the fixes.

PR #5739, opened 2016-08-08
Add More-HTML-Email-Snippets to package list

crdelapazz:
Following are this plugin's difference from `HTML Email Snippets`
- This plugin is specialized for a purpose — **coding non-responsive email with least quirks possible** — while the other one is more flexible on its use.
- Since the purpose is assumed, it has conventions (bits of codes) enforced on snippets, allowing the developers **of similar purpose** type less on their own.
- It also have more snippets like buttons and html template.

*It would be a huge help to our team and other developers of similar purpose (that's why I prefer the public channel). Thanks!*

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -1728,6 +1728,16 @@
 			]
 		},
 		{
+			"name": "More HTML Email Snippets",
+			"details": "https://github.com/crdelapazz/More-HTML-Email-Snippets",
+			"releases": [
+				{
+					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


wbond:
You haven't yet created any [semver](http://semver.org) tags (https://github.com/crdelapazz/More-HTML-Email-Snippets/tags). Without such a tag, no users will be able to install the package.

wbond:
Please open a new PR if you get around to making these fixes.

PR #5740, opened 2016-08-08
Add timestamp converter

Zlitus:


wbond:
pytz isn't part of a standard Sublime Text install (https://github.com/Zlitus/sublime-text-timestamp/blob/master/timestampDecode.py#L1).

You would need to specify it as a dependency. See https://packagecontrol.io/docs/dependencies for general info, and https://github.com/wbond/package_control_channel/blob/master/repository/dependencies.json#L411-L424 to see that it is available as a dependency.

wbond:
Please open a new PR if you get around to making these fixes.

PR #5741, opened 2016-08-09
Update c.json

divinites:
simply change the name of cndict to chinese-english bilingual dictionary

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -733,6 +733,17 @@
 			]
 		},
 		{
+			"name": "Chinese-English Bilingual Dictionary",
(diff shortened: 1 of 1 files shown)


repository/c.json
```diff
@@ -733,6 +733,17 @@
 			]
 		},
 		{
+			"name": "Chinese-English Bilingual Dictionary",
+			"details": "https://github.com/divinites/cndict",
+			"previous_names": ["Youdao English-Chinese Translate"],
```
wbond: You'll want to add `Cndict` as a previous name, otherwise users will get stuck on the version they have installed.

PR #5742, opened 2016-08-09
Adjustments to submit the package AutoItScript

RonildoSouza:


repository/a.json
```diff
@@ -1746,6 +1746,17 @@
 			]
 		},
 		{
+			"name": "AutoItScript",
+			"details": "https://github.com/AutoIt/SublimeAutoItScript",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": ["windows"],
+					"branch": "master"
```
wbond: We no longer support branch-based releases for new packages. Instead, change this to `"tags": true` and create a [semver](http://semver.org) tag on the repo each time you want a new version to be released.

wbond:
The file https://github.com/AutoIt/SublimeAutoItScript/blob/master/packages.json should be removed from the repo.

FichteFoll:
Note that I do not have a clear image of who is in charge of this repo, although I do have commit access myself, so I guess I could in theory approve of this change.

RonildoSouza:
but I'm not allowed to create version tags.

wbond:
@RonildoSouza In that case you probably shouldn't be submitting the repo to be park of the official channel.

wbond:
I just looked at this a bit more.

This package is already part of the main channel, and is already using tags - see https://github.com/AutoIt/SublimeAutoItScript/tags and https://github.com/AutoIt/SublimeAutoItScript/blob/master/packages.json.

This PR should be changed as I mentioned previously to use `"tags": true` like the existing `packages.json` from the repo.

Then this PR needs to be adjusted to remove https://github.com/wbond/package_control_channel/blob/master/channel.json#L17.

Once this PR has been merged, the file https://github.com/AutoIt/SublimeAutoItScript/blob/master/packages.json should be deleted from the repo.

wbond:
You'll need to follow the directions at https://packagecontrol.io/docs/submitting_a_package#Step_7 to ensure your PR isn't failing the test suite.

wbond:
The entry in channel.json still needs to be removed. Additionally, I'll need confirmation from someone who has commit access to the package repo.

RonildoSouza:
Ok. thanks.

FichteFoll:
I opened https://github.com/AutoIt/SublimeAutoItScript/issues/25 18 days ago in order to find someone that is "officially" in charge of the repo. Since most of the commits are from @robertcollier4, I addressed him specifically. However, he doesn't seem to be active at this time.

I'll give him a grace period of 7 days and will approve of the change unless he (or anyone else?) objects.

wbond:
@FichteFoll I believe we have reached the deadline.

PR #5744, opened 2016-08-10
Adding Visual Settings.

darenju:
Hello,

I have created this package that allows user to tweak their configuration with a simple web app.

### DIFF
--- a/repository/v.json
+++ b/repository/v.json
@@ -441,6 +441,16 @@
 			]
 		},
 		{
+			"name": "Visual Settings",
+			"details": "https://github.com/darenju/visual-settings",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Visual Studio Bold Color Scheme",
 			"details": "https://github.com/Trismegistos84/visual-studio-bold",
 			"labels": ["color scheme"],
@@ -604,16 +614,6 @@
 			"name": "VvPHPDollar",
 			"details": "https://github.com/ZhaonanLi/VvPhpDollar",
 			"author": "Zhaonan Li",
-			"releases": [
-				{
-					"sublime_text": "*",
-					"tags": true
-				}
-			]
-		},
-		{
-			"name": "Visual Settings",
-			"details": "https://github.com/darenju/visual-settings",
 			"releases": [
 				{
 					"sublime_text": "*",
--- a/repository/v.json
+++ b/repository/v.json
@@ -610,6 +610,16 @@
 					"tags": true
 				}
 			]
+		},
+		{
+			"name": "Visual Settings",
+			"details": "https://github.com/darenju/visual-settings",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
 		}
 	]
 }


wbond:
It doesn't seem that you have CORS implemented. Leaving out such an essential security feature, and not configuring it properly could theoretically lead to something like http://blog.saynotolinux.com/blog/2016/08/15/jetbrains-ide-remote-code-execution-and-local-file-disclosure-vulnerability-analysis/. I haven't read all of your server code to know what it is possible to do, but my guess is that if you allow modifying settings, the JS to a malicious page could do something like:

 1. Send a JS request to settings
 2a. Add a package to the `"installed_packages"` list for Package Control
 2b. (Or) Add a channel to the Package Control settings with a malicious package version of one in the default package
 3. When the user restarted Sublime Text, the missing package would be installed or upgraded
 4. The package could do anything Python can, like install a trojan, leak all source code being edited, etc.

This could all be done just by visiting a random webpage.

I want to be clear, I don't know for a fact that your current implementation would allow this, but running a local webserver can be a huge gaping security hole when combined with a scripting language that has free access to the filesystem of the machine.

wbond:
This has some discussion of their specific vulnerabilities and possible mitigations: https://news.ycombinator.com/item?id=12292148.

darenju:
What you're pointing here is interesting and I had never even thought about it.

I will work on some security measure and maybe rethink how it should all take place. Thank you.

wbond:
This would still be vulnerable to DNS hijacking since you are not using a secure connection to the website.

FichteFoll:
Is there a particular reason why you can't just let the user host the webserver on his own? It's less convenient for users since they most likely would need to have node.js installed but definitely more secure. And if the backend isn't complicated, which I did not check, you could also just host it using Python's stdlib - most likely.

darenju:
That's what I thought at first, hosting the web app locally. Do you have any tip on how to achieve this, please?

This would make the overall security problem go away, I suppose.

wbond:
The issue isn't who is hosting the webserver, but that something inside of Sublime Text is listening for connections from outside of Sublime Text that can be used to change the settings. Ideally the code that is sending settings changes to Sublime Text is sharing a secret key to authenticate that it can be trusted. The key should expire in a reasonable amount of time (6 hours?), should be random (so it can't be guessed), and should be protected from being accessed by JS in other tabs in the user's browser.

Basically, you need to make sure only tabs that the plugin opens can send requests back to it, and that the secret info can't be leaked. Modern browsers should have various options such as CORS headers to ensure that other websites aren't grabbing JS data from your tabs. However, you also need to ensure that other processes aren't sending HTTP requests to your server.

FichteFoll:
>something inside of Sublime Text is listening for connections from outside of Sublime Text that can be used to change the settings.

It's not like Sublime Text does anything special to protect its settings. Almost any program can directly change ST's configuration files if it has write access on them, and user applications do. Imo this security concern is over-eager.

wbond:
> It's not like Sublime Text does anything special to protect its settings. Almost any program can directly change ST's configuration files if it has write access on them, and user applications do. Imo this security concern is over-eager.

Yes, anything *on your machine* can change the settings. The problem is allowing anything that can make a TCP connection to the machine to be able to write to the hard drive. That enables a remote exploit.

Did you read the article I posted about the vulnerability that all of the JetBrains IDEs suffered from? This plugin has the potential to open such a similar backdoor into your machine. JetBrains paid a $50,000 bug bounty for that…

I haven't heard of any reason why using an expiring session token should be a problem.

FichteFoll:
>Did you read the article I posted about the vulnerability that all of the JetBrains IDEs suffered from?

No, I didn't, but I will do now.

>The problem is allowing anything that can make a TCP connection to the machine to be able to write to the hard drive.

I clarified that statement in an edit. Can Javascript (from a malicious website) create arbitrary TCP connections?

wbond:
> I clarified that statement in an edit. Can Javascript (from a malicious website) create arbitrary TCP connections?

Websockets come to mind, although I'm not familiar with the current state of cross-domain security for those. If no protection is built into the server, then it is leaving the security up to the client software being run on the user's machines. In general, browsers try to mitigate such attacks via JS. However, in the big picture, we are talking about exposing a way for a web page to change settings, so by design there is a web server sitting there, allowing HTTP connections that write to disk.

The big browsers, in general, send CORS headers to validate cross-domain JS. However, depending on how the settings webpage is architected, it may be possible to use cross-domain request forgeries to post a form to http://localhost:8080/save_settings with data `{"installed_packages": "my_backdoor", "channels": "https://sublimehacks.com/channel.json"}`.

I'm not trying to be a jerk, but having spent many years doing web development and learning a lot about the security related to web projects, this project raised concerns based on the description. It may be the case that the exact implementation does not currently allow for a remote exploit. However, any implementation of this concept should take into consideration things like CORS, CSRF, DNS spoofing, websockets, etc.

darenju:
I will try to implement a secret key system. Hosting the web app locally would be a bad pattern, since the remote app could be updated easily for everyone.

For now, the server in Python does check where the connection comes from. I may understand it is not enough, but it is already a first step in securing the whole thing.

FichteFoll:
@darenju what's the current status on this?

darenju:
I am currently busing working on freelance job and do not have much time for my app. I will try to do something, but if someone feels like implementing CORS, feel free to do it.

FichteFoll:
I'll close the PR for now. 

Feel free to comment on this issue once you did that.

PR #5748, opened 2016-08-12
Add Streamer Mode

nicholastay:
https://github.com/nicholastay/Sublime-StreamerMode

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -2800,6 +2800,16 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


wbond:
Perhaps in the readme explain how to use it?

wbond:
I didn't notice any features that require version 3116. Perhaps at least change that to 3114 so beta users will be able to install it?

wbond:
There is a newer version of the beta, so this should be fine now.

PR #5751, opened 2016-08-14
Add 'Hide Sidebar When Not Focussed'

akrabat:
Very small package that does just one thing. The only other similar package I could find targets the idea of showing the sidebar "when you need it" rather that just ensuring that it's nearly always hidden.

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -409,21 +409,21 @@
 			]
 		},
 		{
-			"name": "Hide Sidebar When Not Focussed",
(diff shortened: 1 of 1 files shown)


wbond:
You'll need to fix the order of your package so the tests pass.

Additionally, this should be marked as compatible with build 3098 or newer since you are using APIs added in that build.

akrabat:
Thanks for the feedback @wbond. Can't believe that I didn't notice the CI check…

wbond:
No problem. Sorry it took so long to get to reviewing these…

PR #5752, opened 2016-08-15
add AddRowIndex

shede333:
add AddRowIndex: add digit and letter index in the beginning of every row


wbond:
This seems like it is very similar to https://packagecontrol.io/packages/Insert%20Nums.

wbond:
This seems to be a duplicate and no response from the author.

PR #5754, opened 2016-08-15
add Scuggest plugin

ssanj:
This is a plugin for importing Scala types.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -561,6 +561,18 @@
 			]
 		},
 		{
+			"name": "Scuggest",
+			"details": "https://github.com/ssanj/Scuggest",
+			"labels": ["imports", "scala"],
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


repository/s.json
```diff
@@ -551,6 +551,18 @@
 			]
 		},
 		{
+			"name": "Scuggest",
+			"details": "https://github.com/ssanj/Scuggest",
+			"labels": ["imports", "scala"],
+			"releases": [
+				{
+					"sublime_text": ">=3083",
+					"platforms": ["osx", "linux", "windows"],
```
wbond: This line can be removed since the default is to support all platforms.

ssanj: @wbond Thanks. I have removed the platform element and resubmitted.

ssanj:
@wbond Thanks! :)

PR #5755, opened 2016-08-15
Compass: Remove from channels, add to repository

renestalder:
Removes the Compass package from the channel.json and adds it to the repository since it is hosted on Github anyway. I guess that's how it should be.


wbond:
Great! You should now be able to remove https://github.com/whatwedo/sublime-text-compass/blob/master/packages.json.

renestalder:
@wbond Thank you. I will.

PR #5756, opened 2016-08-15
add nirum syntax

disjukr:


wbond:
Perhaps add a link to a webpage about nirium as a readme in the repo?

PR #5759, opened 2016-08-16
Remove package JuliaCompletions

ViktorQvarfordt:
The package JuliaCompletions is superseded by the package Julia, which now
contains unicode completions implemented in a superior way. Furthermore,
JuliaCompletions lists the package Julia as dependency, so this removal should
be completely safe for all users.

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -1014,16 +1014,6 @@
 			]
 		},
 		{
-			"name": "JuliaCompletions",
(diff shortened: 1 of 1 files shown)


wbond:
@jakeconnor Do you agree with the removal of this package?

jakeconnor:
Yup. Cool with me.

PR #5763, opened 2016-08-18
CCL Syntax Highlighting

daftscience:


wbond:
This file can probably be removed: https://github.com/daftscience/ccl_tools/blob/master/ccl.py.

wbond:
This also seems like a strange addition since you don't document it anywhere, and the package seems to be primarily about adding a syntax for CCL: https://github.com/daftscience/ccl_tools/blob/master/ccl.tmTheme.

daftscience:
Thanks, I've gone ahead and removed those extra files.

PR #5770, opened 2016-08-22
Add Bonsucesso color scheme

sergiovilar:


repository/b.json
```diff
@@ -796,6 +796,17 @@
 			]
 		},
 		{
+			"name": "Bonsucesso",
+			"details": "https://github.com/sergiovilar/bonsucesso",
+			"issues": "https://github.com/sergiovilar/bonsucesso/issues",
```
wbond: This line can be removed since it is the default `issues` URL based on the `details` URL you provided.

sergiovilar:
@wbond updated!

PR #5772, opened 2016-08-24
Move ASCII Replacer to tags

pjlamb12:
Instead of the ASCII replacer plugin being branch based for updates, moved it to tags

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -1353,7 +1353,7 @@
 			"releases": [
 				{
 					"sublime_text": ">=3000",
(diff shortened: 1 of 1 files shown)


wbond:
This is great. The only thing you'll need to make sure is that your tag names are valid [semver](http://semver.org) versions. Right now you have `v1.0`, but that is not a semver. It would need to be `v1.0.0`.

pjlamb12:
I have updated the release version to 1.0.0

PR #5774, opened 2016-08-25
CodeTimeTracker

victoreduardobarreto:
Helo,

It is a new sublime plugin to measure the time spent on projects.

Thanks.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1704,6 +1704,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


wbond:
The file https://github.com/victoreduardobarreto/CodeTimeTracker/blob/master/packages.json should be able to be removed.

victoreduardobarreto:
sd

PR #5777, opened 2016-08-28
add plugin for github code/issue search 

kanghj:
A plugin so that users can perform a github code/issue search from highlighted code in Sublime Text.

For easy access https://github.com/kanghj/sublime-text-github-search-plugin

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -500,16 +500,6 @@
 			]
 		},
 		{
-			"name": "Github Code Search",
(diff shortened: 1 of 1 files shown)


wbond:
You should probably escape the text being placed into the URL. Right now the current implementation could lead to invalid URLs, or something similar.

kanghj:
I have made the changes. Thanks for the check.

PR #5778, opened 2016-08-28
Master

dingo-d:
Added CSS Table of contents plugin

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3168,6 +3168,16 @@
 			]
 		},
 		{
+			"name": "CSS Table of Contents",
+			"details": "https://github.com/dingo-d/CSS-Table-of-Contents",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "CSS To SASS And Stylus Converter",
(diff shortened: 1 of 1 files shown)


wbond:
You don't currently have any valid [semver](http://semver.org) tags on the repo (https://github.com/dingo-d/CSS-Table-of-Contents/tags).

The bigger issue I see is that you are trying to license this as GPL, which is very messy since it is loaded at runtime with a bunch of other code you have no rights over.

I think the LGPL is probably a better fit for what you are looking for.

dingo-d:
I have added tags to my git, and changed licence to LGPL. 

Can you explain a bit with other code? You mean python packages that are used in the package? I have little knowledge on licences, and usually just slap GPL licence on it just to have some licence on it. Can it have any big repercussions? 

Thanks for the advice.

wbond:
Here is some info about the GPL and LGPL and about "linking": http://programmers.stackexchange.com/a/167781.

Scripting languages are effectively doing "dynamic" linking at run time. In short, it is a very muddy part of the GPL. Since your intent is to require users to contribute back to your project, the LGPL feels appropriate. This is especially true since Sublime Text, the whole reason your code exists, is closed source.

PR #5780, opened 2016-08-30
Add Koala and Koala - Snippets to repo

magneto538:
Add Koala and Koala - Snippets to repo


magneto538:
Removed the previous pull requests. Now links, tags and stuff should be correct.

PR #5787, opened 2016-09-02
Cadence Skill and 3D Tool Syntax Hilight

leoheck:


repository/0-9.json
```diff
@@ -46,6 +46,17 @@
 			]
 		},
 		{
+			"name": "3D Tool Syntax Hilight",
+			"details": "https://github.com/leoheck/sublime-3d-tool",
+			"labels": ["3d-tool", "burst-mode", "async"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "master"
```
wbond: Please see https://packagecontrol.io/docs/submitting_a_package#Step_4

repository/c.json
```diff
@@ -131,6 +131,16 @@
 			]
 		},
 		{
+			"name": "Cadence Skill",
+			"details": "https://github.com/leoheck/sublime-cadence-skill",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
```
wbond: Your [current tag](https://github.com/leoheck/sublime-cadence-skill/tags) is not a valid [semver](http://semver.org).

wbond:
These files should be removed:

 - https://github.com/leoheck/sublime-3d-tool/blob/master/packages.json
 - https://github.com/leoheck/sublime-cadence-skill/blob/master/.sublime-package
 - https://github.com/leoheck/sublime-cadence-skill/blob/master/packages.json

You should add a description of what the 3D Tool package does. Perhaps a link to the product/project it works with.

You should add a link to the Cadence Skill readme to describe what cadence skill is.

leoheck:
Thank you for the feedback.

leoheck:
Hi @wbond , is it possible to merge this pull request now?

wbond:
You are still pointing the 3D package to the master branch, not tags.

leoheck:
Hm, Now it is only possible with tags? Ok, I will fix then.

leoheck:
Both 3D and Skill were updated using tags with semantic versioning.  I think I covered all your requests now.

PR #5789, opened 2016-09-03
AddPackage "Sherumite - Theme"

gsheru:
Add Package data to repository/s.json


wbond:
Your tag needs to be a [valid semver](http://semver.org). In short, it needs three numeric parts. So `v1.0.0` instead of `v1.0`.

gsheru:
Done!

PR #5791, opened 2016-09-05
add Wilddog snippets

dongkai1993:


FichteFoll:

```sh
$ git tag v1.0.0
$ git push origin v1.0.0
```

wbond:
@dongkai1993 You need to add a tag before we can accept this.

PR #5792, opened 2016-09-05
Add Subpry plugin

BarakHarizi:
A plugin for quickly setting and removing pry breakpoints


FichteFoll:

```sh
$ git tag v1.0.0
$ git push origin v1.0.0
```

BarakHarizi:
@FichteFoll I've added the tag, is it OK now?

PR #5795, opened 2016-09-06
Add dicttoxml

idleberg:


idleberg:
Reopened, since I'm going to use this in a different package

PR #5799, opened 2016-09-07
Simple LaTeX preview plugin for ST

alexkorovkov:
A semvar tas has been created

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -492,6 +492,21 @@
 			]
 		},
 		{
+			"name": "TeXPreview",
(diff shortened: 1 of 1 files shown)


wbond:
You'll need to create a [semver](http://semver.org) tag at https://github.com/alexkorovkov/TeXPreview/tags so users may install the package.

FichteFoll:
By the way, you don't need to pass the current environment to `subprocess.Popen` since it alread inherits the env from the launching process.

PR #5803, opened 2016-09-08
Add ReqsLang package

petarslovic:
Add syntax highlighting for [ReqsLang](https://github.com/petarslovic/reqs-lang).

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -882,6 +882,17 @@
 			]
 		},
 		{
+			"name": "ReqsLang",
+			"details": "https://github.com/petarslovic/sublime-reqs-language",
+			"labels": ["requirements", "todo"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Request",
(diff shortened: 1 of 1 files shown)


repository/r.json
```diff
@@ -882,6 +882,17 @@
 			]
 		},
 		{
+			"name": "ReqsLang",
+			"details": "https://github.com/petarslovic/sublime-reqs-language",
+			"labels": ["requirements", "todo"],
```
wbond: You probably want to add the `language syntax` label

petarslovic: It is a syntax highlighting plugin for my own language, so not sure if I should add it as a label?

wbond: Since it is a `.tmLanguage`, I think it makes sense.

petarslovic: Right, cool, thanks for the feedback

wbond: Sorry, I meant literally the label `language syntax`: https://packagecontrol.io/browse/labels/language%20syntax

petarslovic: Ah, okay, got it now! :D

PR #5804, opened 2016-09-09
Add VueFormat package

lanbin:
It's a plugin for Vue file format.

### DIFF
--- a/repository/v.json
+++ b/repository/v.json
@@ -571,18 +571,6 @@
 			]
 		},
 		{
-			"name": "VueFormat",
-			"details": "https://github.com/lanbin/vueformat",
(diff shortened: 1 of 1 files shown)


wbond:
You'll need to create a [semver](http://semver.org) tag at https://github.com/lanbin/vueformat/tags so users may install the package.

lanbin:
@wbond thanks for your prompting

FichteFoll:

FichteFoll:
Furthermore, your code indicates support for all platforms but here you restrict the package to only be installable by OSX users. Further-furthermore, you should use `os.path.join` (and similar) when working with path strings.

FichteFoll:
@lanbin ping

PR #5813, opened 2016-09-17
+ GitTodo.

hipkiss91:
Allows developers within SublimeText 3 to automatically publish issues to a public repo on github through the use of properly formatted TODO's.

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -693,6 +693,17 @@
 			]
 		},
 		{
+			"name": "GitTodo",
+			"details": "https://github.com/hipkiss91/GitTodo",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": ["windows"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "GitUp",
 			"details": "https://github.com/braver/SublimeGitUp",
 			"labels": ["git"],


wbond:
You'll need to create a [semver](http://semver.org) tag at https://github.com/hipkiss91/GitTodo/tags so users may install the package.

hipkiss91:
Done.

hipkiss91:
@wbond Is this now possible to merge? :)

FichteFoll:
- Is there a particular reason why your package is marked as Windows-only? Besides a couple hard-coded backslashes as path separators, I didn't spot anything.
- You should use the `cwd` kwarg to `subprocess.Popen` instead of chaging the process's working dir with `os.chdir`.
- You might also want to consider Github's 2FA, which is based on tokens, and OAuth.
- Since the pacakge is primarily targetted at "Github", it makes sense to include that in its name.

A minor Python suggestion: Use the `in` operator instead of  `str.find` if you want a boolean result.

hipkiss91:
@FichteFoll Thanks for the feedback. I'll be sure to include these amendments :). Also, currently yes it is only targeted at github, but I plan on expanding to bitbucket and others in the future. Thanks for the python tip, I am new to it (this is my first python project).

I marked it as windows only just because I didn't have a mac to test it on so didn't want to commit to that. However, I'll dynamically include the path separators to fix this.

FichteFoll:
@hipkiss91 ping

hipkiss91:
@FichteFoll  Hey, Unfortunately progress has been slow recently. However, an update for this will be available within a week or two.

PR #5819, opened 2016-09-18
[octopress-snippets] added octopress-snippets to o.json and removed i…

yangsu:
…t from channel.json


wbond:
Package Control uses [semver](http://semver.org), so your version would need to have three parts, e.g. `v0.1.0`.

yangsu:
Oops. Updated.

PR #5833, opened 2016-09-22
host json for SublimeREPL and requirementstxt packages on GH

wuub:
reference to  https://wuub.net/packages.json was removed


wbond:
It is probably easier to just drop these in the appropriate `.json` file in the `repository/` folder of this repo.

wuub:
I expect there is some more fiddling ahead as I want to freeze/drop support for olders SublimeText versions. Once the content of those files stabilise enough I'll send another PR including them in `repository/` if that's acceptable of course :)

wbond:
Sounds good

PR #5837, opened 2016-09-26
Add Paste as String

mom1:


mom1:
something wrong?

FichteFoll:
[This import](https://github.com/mom1/PasteAsString/blob/93eeac80445cefa3322c28119763b251d7c13b06/PasteAsString.py#L8) will fail once installed via Package Control, because it will be installing the package as `Paste as String`, as specified in the `name` field. I suggest to use a relative include (and optionally renaming the package to PasteAsString, since that's the preferred naming convention for package names (due to being importable in Python)).

You might also want to check out the `view.match_selector` API function, which performs a scope selector match against a position in a view instead of doing a string contention check.

mom1:
Done:

1. Change name to PasteAsString
1. Use view.match_selector

PR #5844, opened 2016-09-28
Add ArmaBuild build system

jonpas:


FichteFoll:
The build system is only compatible with ST3.

Previously, we used to reject low-effort or low-value packages like this (it's a 2-line file), but when moving twoards an automated package review process, this will most likely go through anyway, so we're relaxing this requirement.

jonpas:
Package should now be compatible with ST2 as well (tested in portable version of ST2).

jonpas:
Thank you!

PR #5847, opened 2016-10-01
Added anaconda_go to the default channel

DamnWidget:


DamnWidget:
Weird PR, I will do it later again

PR #5849, opened 2016-10-01
Add RustFormat package

Rypac:


Rypac:
@wbond Anything extra I need to do to have this package added?

PR #5850, opened 2016-10-02
Add PersistentPython

mottosso:
Hi, I would love for my project to be accessible via Package Control.
- [Link](https://github.com/mottosso/PersistentPython)

It is my first pull-request to here, so let me know if there are any problems.

Thanks.


mottosso:
Updated README and initial comment, from prototype to functional package.

PR #5855, opened 2016-10-04
Add FoldJsFunctions package

HernanZh:
Pull request for FoldJsFunctions plugin (ST3 only).
Repo: https://github.com/LuckyKat/sublime-fold-functions
Semver tags: https://github.com/LuckyKat/sublime-fold-functions/tags

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -1110,7 +1110,7 @@
 			"details": "https://github.com/LuckyKat/sublime-fold-functions",
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": ">=3125",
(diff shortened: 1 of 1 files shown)


FichteFoll:
You should restrict installation to ST3 only.

HernanZh:
I added "sublime_text": ">=3125"
Is that enough to restrict it or is there more? @FichteFoll

FichteFoll:
No, that's certainly enough (might even be too restrictive since it only covers the latest beta and dev releases respectively). Since your package is very reliant on the JavaScript default package however, this seems alright.

PR #5857, opened 2016-10-04
Added CircleCi Status

drazisil:
- Link to your code repository: https://github.com/drazisil/sublime-circleci/
- Link to the tags page with at least one [semver](http://semver.org) tag:  https://github.com/drazisil/sublime-circleci/releases/tag/v.0.5.0-beta

Also make sure you:
1. Used `"tags": true` and not `"branch": "master"` ([versioning docs](https://packagecontrol.io/docs/submitting_a_package#Step_4)) Yes
2. Ran the tests ([tests docs](https://packagecontrol.io/docs/submitting_a_package#Step_7)) Yes - Pass

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -940,7 +940,6 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"branch": "master",
 					"tags": true
 				}
 			]
--- a/repository/c.json
+++ b/repository/c.json
@@ -940,7 +940,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"platforms": ["osx", "linux"],
+					"branch": "master",
 					"tags": true
 				}
 			]
--- a/repository/c.json
+++ b/repository/c.json
@@ -935,6 +935,17 @@
 			]
 		},
 		{
+			"name": "CircleCi Status",
+			"details": "https://github.com/drazisil/sublime-circleci/",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": ["osx", "linux"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Cirru",
 			"details": "https://github.com/cirru/sublime-cirru",
 			"labels": ["language syntax"],


joshua-golub:
Drazisil-

I am reviewing this package but have no idea what it actually **does**. Would you please update the `README.md` in your repository to provide a useful explanation?

@FichteFoll, this plugin is using a version number of `v.0.5.0-beta`. I know that `v0.5.0-beta` is valid, but is it also valid to have a period (`.`) after the `v`?

-Josh

FichteFoll:
@joshua-golub, technically any prefix is accepted, but it needs to be specified in the `"tags"` field of the release. By default, PC only recognizes no prefix and `v`.

@drazisil, as mentioned above, please add a tag that does not have a dot after `v`.
Furthermore, note that, by default, pre-release versions are not installed by Package Control. This has to be enabled in its settings before attemption to install your package. I suggest releasing an non-prerelease version.

drazisil:
README updated and also tagged a version v1.0.0

FichteFoll:
Due to `self.view.file_name().rpartition('/')[0]`, your package won't work on Windows. It will if you use `os.path.dirname(self.view.fil_name())` however.

drazisil:
@FichteFoll I didn't mark windows because I couldn't test on that. If thats all it takes though. What about https://github.com/drazisil/sublime-circleci/blob/master/CircleCi.py#L43?

FichteFoll:
>What about https://github.com/drazisil/sublime-circleci/blob/master/CircleCi.py#L43?

That operates on a URI, no? There are no differences for platforms in that regard.

>Also, self.view.fil_name() or self.view.file_name() ?

Ah, `file_name` of course.

drazisil:
Done

repository/c.json
```diff
@@ -935,6 +935,17 @@
 			]
 		},
 		{
+			"name": "CircleCi Status",
+			"details": "https://github.com/drazisil/sublime-circleci/",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "master",
```
FichteFoll: You must remove this line

PR #5859, opened 2016-10-06
DW Site Tools: Package to import/export Adobe Dreamweaver site files (.STE)

Tardog:
- Code repository: https://github.com/Tardog/dw-site-tools
- Tags page: https://github.com/Tardog/dw-site-tools/tags

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -1260,6 +1260,16 @@
 			]
 		},
 		{
+			"name": "DW Site Tools",
+			"details": "https://github.com/Tardog/dw-site-tools",
+			"releases": [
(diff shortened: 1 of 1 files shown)


FichteFoll:
`sys.path.append(os.path.join(os.path.dirname(__file__), "lib", "tkinter"))` does effectively nothing, since a) that directory does not exist inside your repo, and b) your package would be installed as a `.sublime-package` zip archive anyway. Either way, you're not using tk in your code (it's commented) and for good reason.

The `st_version` variable can never be something other than `2` or `3`.

PR #5860, opened 2016-10-07
Add Angular2 HTML Syntax

princemaple:
Repo: https://github.com/princemaple/angular2-html-syntax
Releases: https://github.com/princemaple/angular2-html-syntax/releases

This fixes html syntax to correctly highlight Angular2 `input`, `output`, `template`, `reference` and `interpolation` syntaxes in the html template

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -705,6 +705,17 @@
 			]
 		},
 		{
+			"name": "Angular2 HTML Syntax",
+			"details": "https://github.com/princemaple/angular2-html-syntax",
+			"labels": ["language syntax", "angular"],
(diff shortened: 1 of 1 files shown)


deathaxe:
Why do you want to create another package for HTML syntax? I would prefere a pull-request to the [Default](https://github.com/sublimehq/Packages) package?

princemaple:
Because it's not standard HTML.

princemaple:
sp161027_082319
those things don't exist in normal HTML, even if I did send a PR to default HTML package, it wouldn't be accepted.

PR #5865, opened 2016-10-08
[Rust] moving from commits to tags, plus name change

Jayflux:
Moving from commits to tags, plus name change and Sublime Text 3+


Jayflux:
pings @wbond

PR #5867, opened 2016-10-09
Tiny ST3 plug-in to easily switch between: no spell-check, spell-check for French, and spell-check for English

Naereen:
- Code: [Naereen/SublimeText3_SwitchDictionary](https://github.com/Naereen/SublimeText3_SwitchDictionary)
- Tags: [v0.0.3](https://github.com/Naereen/SublimeText3_SwitchDictionary/releases/tag/v0.0.3)
- [More details on this small plugin](https://github.com/Naereen/SublimeText3_SwitchDictionary#about) (remark: it is very early stage).

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -4307,6 +4307,17 @@
 			]
 		},
 		{
+			"name": "SwitchDictionary",
+			"details": "https://github.com/Naereen/SublimeText3_SwitchDictionary",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"platforms": ["osx", "linux"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "SwitchLanguage",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Your `messages.json` file contains a trailing comma and is not valid JSON.

Instead of adding your package folder to the path, use a relative import (in ST3, absolute in ST2). It's more robust and will continue to work even when in a `.sublime-package` zip archive.

Naereen:
Hi,
Thanks for your suggestions.
Sorry about the mistake on the JSON file (that's the problem when you see JSON as a Python dict... it's not as flexible!).
I adapted the import, to be a relative import (and back to the first trick in case the relative important fails).

FichteFoll:
Remember that you haven't pushed a new tag since your last release last month (iirc). I'll go ahead and merge it anyway since the package should generally work.

PR #5871, opened 2016-10-10
Add the Compile Jasper Template package

amsoell:
This is a very simple package that runs the current JRXML file through the JasperStarter compiler.

Code repository: https://github.com/MettleUp/CompileJasperTemplate
Code tags / releases: https://github.com/MettleUp/CompileJasperTemplate/tags

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2468,6 +2468,17 @@
 			]
 		},
 		{
+			"name": "Compile Jasper Template",
+			"details": "https://github.com/MettleUp/CompileJasperTemplate",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": ["osx", "linux"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Compile Selected ES6",
 			"details": "https://github.com/xinchaobeta/compile-selected-es6",
 			"releases": [


joshua-golub:
@FichteFoll, I have reviewed this package and it looks good to me.

FichteFoll:
Instead of resolving the path to the executable yourself, you could just let the OS handle it (and specify `shell=True` on Windows so that it uses `cmd /c ...` and doesn't require an extension. Unless "jasperstarter" isn't compatible with Windows in the first place.

amsoell:
@FichteFoll: Forgive me, but Python isn't a language I know _super_ well, so I'm not sure exactly how to implement this. I know "jasperstarter" is available on Windows, but I haven't used it in that environment myself. I'm happy to update it, but the pull request as-is does work.

FichteFoll:
I just wanted to say that you could replace the entire `CompileJasperTeplate.py` file with this:

```py
import subprocess

import sublime_plugin


class CompileJasperTemplateCommand(sublime_plugin.TextCommand):
    def run(self, edit):
        subprocess.Popen("jasperstarter compile \"{0}\"".format(self.view.file_name()), shell=True)
```

and it will work on all platforms. Either way, as you mentioned correctly, this package works as-is. I just wanted to confront you with the option of opening this for Windows as well. Since you replied to my comment, I assume that you considered my remarks and am now willing to merge it, even in its current state.

And while we're at it, the proper way to do something like this is with a build system, which would be as simple as this:
```
{
    "shell_cmd": "jasperstarter compile \"$file\""
}
```
and would be invoked using ctrl+shift+b (to select the build system once), after which you can use ctrl+b.
More on build systems here: http://docs.sublimetext.info/en/latest/reference/build_systems.html.

amsoell:
@FichteFoll Thanks for taking the time to write that out for me—as I said, I'm fairly new with Python but really wanted to have a way to build Jasper compilation into ST3. So this was a pretty quick-and-dirty version, but I'm definitely interested in making it better. I'll work on implementing this for version 2.0 😃

PR #5872, opened 2016-10-11
Add Lua Assist

litianqi:
- code repository: https://github.com/litianqi/LAssistX
- tag: https://github.com/litianqi/LAssistX/tree/v0.3.1

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -1244,6 +1244,16 @@
 			]
 		},
 		{
+			"name": "Lua Assist",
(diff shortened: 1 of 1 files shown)


litianqi:
tags: https://github.com/litianqi/LAssistX/tags

joshua-golub:
Litianqi-

Is there a particular reason why you added these two Lua assist tools into a new package, rather than extending one of the various Lua-related packages that already exist for Sublime Text?

If so, please explain.

-Josh

PR #5874, opened 2016-10-11
Add LC-2K Assembly Syntax Plugin to l.json

BGR360:
Syntax Highlighting and auto-completion for the LC-2K assembly ISA.

https://github.com/BGR360/lc2k-sublime-plugin

https://github.com/BGR360/lc2k-sublime-plugin/releases/tag/0.0.0

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -369,6 +369,18 @@
 			]
 		},
 		{
+			"name": "LC-2K Assembly",
+			"details": "https://github.com/BGR360/lc2k-sublime-plugin",
+			"author": "Ben Reeves",
+			"labels": ["auto-complete", "language syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "LC3 Assembly",
 			"details": "https://github.com/wufufufu/Sublime-LC3",
 			"releases": [


joshua-golub:
BGR360, please provide a more appropriate semver tag than `0.0.0`. I think that either `1.0.0` or `0.1.0` would be better.

BGR360:
Done.

joshua-golub:
@FichteFoll, this package looks good to me. I recommend that it be merged.

FichteFoll:
@BGR360, please remove `LC-2K Assembly.sublime-package` from the repo. It serves no purpose (you don't even use it for your manual download instructions) and just increases the repo size. Don't forget to push a new tag for that.

A recommendation: Use `'` single quotes for strings in YAML when specifying regular expressions. Single-quoted strings don't support any escape sequences besides `''` and are much better for something as backslash-heavy as regular expressions.

Also, using a `0.0.0` semver tag is perfectly valid and some argue for it to be the initial version, although it is uncommon. The interested reader might want to take a look into https://github.com/mojombo/semver/issues/221.

BGR360:
Will do. Why do you recommend using single quotes? To avoid having to do \\ for every escape sequence?

BGR360:
Almost all of the documentation I've seen for sublime-syntax files uses the double-quotes-and-double-backslashes style. I will make note of this for future packages I make though :)

FichteFoll:
>Why do you recommend using single quotes? To avoid having to do \ for every escape sequence?

Exactly. Actually, in most cases you don't need quotation marks at all.

> Almost all of the documentation I've seen for sublime-syntax files uses the double-quotes-and-double-backslashes style. 

Probably lack of knowledge by the author(s).

BGR360:
Removed the sublime-package file. Here is the new semver tag: https://github.com/BGR360/lc2k-sublime-plugin/releases/tag/v1.0.0

BGR360:
Thank you! :))

BGR360:
How can I be notified when this becomes publicly available through Package Control's "Install Package" command?

FichteFoll:
You can't, but it usually happens within an hour. Okay, there is https://twitter.com/SublimePackages, but that's pretty late. Might as well just refresh https://packagecontrol.io/browse/new if it's important for you.

vovkkk:
There is rss http://wbond.net/sublime_packages/community/rss

FichteFoll:
I was looking  for an rss meta tag on the packagecontrol.io site but couldn't find one. Might be worth adding an issue over it at the corresponding repo.

PR #5876, opened 2016-10-11
add MLFi package for Sublime Text

li-vu:
- Link to your code repository: https://github.com/li-vu/st-mlfi
- Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/li-vu/st-mlfi/tags

Also make sure you:

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -1352,6 +1352,17 @@
 			]
 		},
 		{
+			"name": "MLFi",
+			"details": "https://github.com/li-vu/st-mlfi",
+			"labels": ["language syntax", "mlfi", "csml"],
+			"releases": [
+				{
+					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


joshua-golub:
@FichteFoll, I've reviewed this package and see no issues. If you agree, please merge.

FichteFoll:
As a suggestion, you might want to move any code that doesn't directly interface with SublimeText (specifically, creates a subclass of one of the `sublime_plugin` classes) to sub-modules that can be imported from the main files. This would bring better structure into your package at the loss of automatic-reloading of sub-modules.

Note that importing from the package root behaves differently for ST2 and ST3.

li-vu:
Thank you, @FichteFoll, for your suggestion. I will consider it in the next iteration of the package.

PR #5878, opened 2016-10-12
add Kustom Color Schemes

kcmr:


joshua-golub:
@FichteFoll, this package looks good to me.

PR #5879, opened 2016-10-13
Change name and homepage of package

Kristinita:
### Changing package name

Yes, I understand, which is [not desirable to use the word `Sublime` in the name of the package](https://packagecontrol.io/docs/submitting_a_package#Step_2), but I decided to still use that word. Reasons:
1. I do not want confusion in names. Users of the package may get confused if the package is always called SashaSublime, but in Package Control Package will have a different name.
2. I wish for greater recognition to use the word `Sublime`. `Sublime` in my package is not a name package only for Sublime Text, `Sublime` — an integral part of package name. Even if my package ever be in other editors/IDE, it will be called `SashaSublime`.
### Changing homepage

I am change homepage, because GitHub it limits the possibility of design README.MD file. For example, [I can not use JavaScript on README.MD page](http://stackoverflow.com/questions/21340803/embed-javascript-in-github-readme-md).

Thanks.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -96,8 +96,11 @@
 			]
 		},
 		{
-			"name": "SashaS",
+			"name": "SashaSublime",
+			"previous_names": ["SashaS"],
 			"description": "Theme and color scheme, where all elements are good visible",
+			"donate": "http://kristinita.ru/Sublime-Text/SashaSublime#donate",
+			"homepage": "http://kristinita.ru/Sublime-Text/SashaSublime",
 			"details": "https://github.com/Kristinita/SashaSublime",
 			"author": "Sasha Chernykh",
 			"labels": ["theme", "color scheme"],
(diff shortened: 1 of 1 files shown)


Kristinita:
@wbond , please, могли бы вы рассмотреть запрос в ближайшее время? My package now have many bugs, because in paths of files of my package `SashaSublime`, not `SashaS`. [Example](https://github.com/Kristinita/SashaSublime/blob/master/SashaSublime.sublime-theme#L10). Thanks.

FichteFoll:
I'm not sold on your reasoning for the name change, personally.

Kristinita:
@FichteFoll , all right, but may you accept this pull-request?

Thanks.

FichteFoll:
If I'm against this, I won't merge it, but because I'm not too strong on the opposing side I'll let @wbond overrule me here, although he has traditionally followed similar judgement. 

So yeah, we'll wait for @wbond's comments on the name change.

Kristinita:
@FichteFoll , is not the author of the plugin to decide how it will be called a plugin if plugin name not violates the rules?

Thanks.

Kristinita:
@FichteFoll , all right, I [**create new pull-request**](https://github.com/wbond/package_control_channel/pull/5964), now branch has no conflicts.

Thanks.

PR #5880, opened 2016-10-13
Node DevTools

winter95:
- Link to your code repository: https://github.com/winter95/node-devtools-sublime
- Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/winter95/node-devtools-sublime/tags

Also make sure you:

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -486,6 +486,18 @@
 			]
 		},
 		{
+			"name": "Node DevTools",
+			"details": "https://github.com/winter95/node-devtools-sublime",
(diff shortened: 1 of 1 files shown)


FichteFoll:
You don't need a `.no-sublime-package` file in your package. Please remove it.

With how your package tries to launch the executable (`'node-devtools.cmd'`), it won't work on any non-windows platform. Instead, remove the `.cmd` suffix and add `'shell: True` to the `exec` command call.

FichteFoll:
@winter95 ping

PR #5882, opened 2016-10-14
Add cert tools package

Gui13:
- Link to your code repository: https://github.com/Gui13/sublime-certtools
- Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/Gui13/sublime-certtools/tags

Also make sure you:

I did all that!

This package allows users to decrypt PEM and CRT files with a right click.
It is still very limited but will evolve with user feedback.
It requires the openssl command to be available on the system (will make the path configurable in future version) and only supports the first certificate when parsing a cert chain for the moment (this is a TODO).

But this has saved me countless minutes of opening the console to parse a PEM in many nginx configurations :)

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -402,6 +402,17 @@
 			]
 		},
 		{
+			"name": "Cert Tools",
+			"details": "https://github.com/Gui13/sublime-certtools",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": ["osx", "linux"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "cexio-ticker",
 			"details": "https://github.com/iceydee/cexio-ticker",
 			"releases": [


wbond:
Just so you know, recent versions of package control some with asn1crypto and can fully decode and display information about PEM certificates. This is a pure-Python library and doesn't require the user have `openssl` installed, and doesn't require starting a subprocess. It is used by Package Control to export and encode certificates to be used with the OpenSSL-based Python _ssl module for secure connections.

Just figured I'd mention it since any one who is using Package Control will have that available, and the API makes it a bit easier to use than trying to parse the output of the OpenSSL command line tools.

Gui13:
Hi Will,

This is perfect! I'll try to do that for next version. I really went for the straight version on this, cause I needed something working fast. 
But using the pure python version would make this cross platform as well, so I'll try to do that. I have lots of ideas actually (allow to verify certificate signatures, allow to sign a cert with another etc.. from within Sublime).

Thanks for dropping by and commenting on this.

wbond:
If you are interested in signing and verification, you may want to look at oscrypto – that's the cross-platform cryptography library built on top of asn1crypto. Then there is certbuilder, certvalidator, etc which solve a number of the features you wanted to add.

Gui13:
I'm quite.. new to this package development, and I can't understand how I am supposed to import asn1crypto from my python file.
Am I supposed to import like this:

    from asn1crypto import pem, x509

Or should I import through the package control plugin, somehow like this:

    from PackageControl.asn1crypto import pem, x509

(neither work)

wbond:
You can import the version from Package Control via:

```python
from package_control.deps.asn1crypto import pem, x509
```

Gui13:
Yeah, just figured it out. Thank you Will!

Gui13:
So.. will you accept this merge request? 
I have successfully used asn1crypto to decode some PEM files, but I need some time to format the python certificate object to obtain something alike openssl output. In the meantime, maybe this one can be acceptable? And I'll update with a 0.2 version later on for true multi platform.

Thanks again for the input on this.

wbond:
Yes, this is in the queue to be reviewed

PR #5884, opened 2016-10-16
SwiftSyntax - syntax highlighting for Swift 3.0 in .sublime-syntax format

GregoryBL:
- Link to your code repository: [https://github.com/GregoryBL/SwiftSyntax](https://github.com/GregoryBL/SwiftSyntax)
- Link to the tags page with at least one [semver](http://semver.org) tag: [https://github.com/GregoryBL/SwiftSyntax/tags](https://github.com/GregoryBL/SwiftSyntax/tags)

Also make sure you:

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -4226,6 +4226,17 @@
 			]
 		},
 		{
+			"name": "SwiftSyntax",
+			"details": "https://github.com/GregoryBL/SwiftSyntax",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Swig",
 			"details": "https://github.com/enagorny/sublime-swig",
 			"releases": [


joshua-golub:
GregoryBL-

Would you please clarify for me what your Swift syntax provides that other similar packages (for example, [this one](https://github.com/quiqueg/Swift-Sublime-Package)) don't provide?

-Josh

GregoryBL:
@joshua-golub ,

As far as I am aware it’s the only (more-or-less complete, c.f. https://github.com/colinta/Swift-for-f-ing-sublime) one that uses the modern sublime-syntax format as opposed to the older tmLanguage format, which is harder (at least for me) to work with. I haven’t looked at the one you linked in great detail, but it doesn’t seem to use the meta-descriptors to give as rich of semantic meaning to, for example, class definitions. 

Perhaps that’s not enough given all of the additional tooling around that project, but I thought this one would give a nicer path forward for development. I’m sure you have a better ability to gauge that than I do.

Thanks for the review,
Greg

joshua-golub:
@FichteFoll, this package looks good to me.

FichteFoll:
FYI, Swift is peding addition as a default package: https://github.com/sublimehq/Packages/pull/11, however the current state of this is unclear. Maybe you are interested in picking that up and making sure everyone gets proper Swift support?

For now, unless you can explain the differences between the existing two pakages (maybe with pictures), I don't really see a need for this.

Btw, your readme link to Package Control is backwards.

GregoryBL:
Thanks, I'll work on contributing to those other projects.

PR #5885, opened 2016-10-16
add cslint for sublime

molee1905:
- Link to your code repository: https://github.com/cslint/sublime-cslint
- Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/cslint/sublime-cslint/releases/tag/v0.1.0

Also make sure you:

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3110,6 +3110,21 @@
 			]
 		},
 		{
+			"name": "cslint",
+			"details": "https://github.com/cslint/sublime-cslint",
+			"labels": ["linting", "formatting", "code style"],
+			"author": "molee1905",
+			"readme": "https://github.com/cslint/sublime-cslint/blob/master/README.md",
+			"issues": "https://github.com/cslint/sublime-cslint/issues",
+			"releases": [
+				{
+					"sublime_text": ">=3124",
+					"platforms": ["osx"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "CSON Converter",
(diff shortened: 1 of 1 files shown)


joshua-golub:
Molee1905-

There are a lot of [lint packages](https://packagecontrol.io/browse/labels/linting) that already exist for Sublime Text. Would you please clarify what your package does that the others do not?

-Josh

molee1905:
this is just for [cslint](https://github.com/cslint/cslint) which support to lint and fix js and css in the same time, as you can see [why cslint](https://github.com/cslint/cslint#why). 

In terms of sublime plugin, for example, the best sublimelinter, if you want to lint js and css,  you may install
* sublimelinter
* sublimelinter-contrib-eslint / sublimelinter-contrib-eslint-d
* sublimelinter-contrib-stylelint / sublimelinter-contrib-stylelint-d 

You must figure out what‘s the difference between **xxx** and **-d** before you start use.

and the output format without rule links shows on command line which is not so friendly

FichteFoll:
Arg, merged too early.

@molee1905, you don't need a `.no-sublime-package` file in your package. Please remove it.

Also, "lint" and "fix" are verey generic names for commands. Remember that the command namespace is shared among all plugins. Finally, is there a reson why you only provide an OSX keybinding (and not for all platforms or none)? Either way, it overrides a default binding and should be removed.

```
{ "keys": ["super+ctrl+f"], "command": "toggle_full_screen" },
```

PR #5658, opened 2016-07-06
Add dependencies pathtools, watchdog, for FileBrowser

vovkkk:
https://github.com/vovkkk/sublime-pathtools
https://github.com/vovkkk/sublime-watchdog

cc @aziz

PR #5895, opened 2016-10-20
Add dateutil dependecy for PlainTasks

vovkkk:
https://github.com/vovkkk/sublime-dateutil
https://github.com/vovkkk/sublime-dateutil/releases

cc @aziz

### DIFF
--- a/repository/dependencies.json
+++ b/repository/dependencies.json
@@ -134,6 +134,20 @@
 			]
 		},
 		{
+			"name": "dateutil",
+			"load_order": "50",
+			"description": "The dateutil module provides powerful extensions to the datetime module available in the Python standard library",
+			"author": "vovkkk",
+			"issues": "https://github.com/vovkkk/sublime-dateutil/issues",
+			"releases": [
+				{
+					"base": "https://github.com/vovkkk/sublime-dateutil",
+					"tags": true,
+					"sublime_text": "*"
+				}
+			]
+		},
+		{
 			"name": "dicttoxml",
 			"load_order": "50",
 			"description": "Converts a Python dictionary or other native data type into a valid XML string - https://github.com/idleberg/sublime-dicttoxml",
--- a/repository/p.json
+++ b/repository/p.json
@@ -1303,7 +1303,8 @@
 			"releases": [
 				{
 					"sublime_text": "*",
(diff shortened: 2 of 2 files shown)


joshua-golub:
@vovkkk, I am reviewing this package but don't quite understand the value that it provides to the Sublime Text user. Would you please elaborate in your README file?

vovkkk:
I believe my readme contains all needed information; it is for developers who want to use dateutil module in their package(s)

I guess you are not familiar with dependencies in PC, my readme contains link to docs https://packagecontrol.io/docs/dependencies

FichteFoll:
I agree that the readme is more than enough.

However, I suggest to use an in-repo `dependencies.json` file for specifying dependencies since that will allow updating the dependencies along with new releases instead of having to update the channel.

vovkkk:
> use an in-repo dependencies.json file for specifying dependencies since that will allow updating the dependencies along with new releases instead of having to update the channel.

I do not understand.
I have experience of adding dependencies #5658 and I do know how version update works; there is no need to update channel. Dependency will be updated in case it has new version _and_ package that depends on it has new version.

Surely, once this pr will be merged I will add dependencies.json file to repo for manual installation.

I hope that setting dependency in channel will allow to avoid restart ST after installation (if it is not this way then it could be in future), i.e. when user install PlainTasks then PC can see dependency right away install it, load it, and then install and load PT.

FichteFoll:
>Dependency will be updated in case it has new version and package that depends on it has new version.

I meant that the *package* update can change the dependencies it uses along when its code is updated as well.

>Surely, once this pr will be merged I will add dependencies.json file to repo for manual installation.

When specifying a package's dependencies in the channel, its `dependencies.json` file will be ignored.

>I hope that setting dependency in channel will allow to avoid restart ST after installation (if it is not this way then it could be in future), i.e. when user install PlainTasks then PC can see dependency right away install it, load it, and then install and load PT.

Dependencies are always installed before the package (I myself made sure of that).

vovkkk:
I’ve made change you suggested.

> When specifying a package's dependencies in the channel, its dependencies.json file will be ignored.

That’s disappointing and non-intuitive, TBH.

PR #5897, opened 2016-10-20
Add SVG to JSX converter to s repo

tylergould:
https://github.com/tylergould/sublime-svg-to-jsx
https://github.com/tylergould/sublime-svg-to-jsx/tags

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -4148,6 +4148,17 @@
 			]
 		},
 		{
+			"name": "SVG to JSX",
+			"details": "https://github.com/tylergould/sublime-svg-to-jsx",
+			"labels": ["svg", "jsx", "react"],
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


joshua-golub:
@tylergould, I am reviewing this package and I think it would be useful to include an example of what your converter will do. For example, what the before (SVG) and after (JSX) will look like. Would you please add this to your README?

FichteFoll:
Please use a less generic command name than `convert`. Remember that all plugins share the same command namespace.

tylergould:
Thanks for the feedback. I will make the updates in the next couple of days.

tylergould:
@FichteFoll and @joshua-golub -- in the latest version, I've updated the README with example input & output, and I've changed the name of the command to avoid collisions.

Thanks again for the feedback.

PR #5901, opened 2016-10-21
Vintage Solid Caret

muffinmad:
Plugin sets `caret_style` to solid in command mode and sets it back in insert mode.
- Link to code repository: https://github.com/muffinmad/VintageSolidCaret
- Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/muffinmad/VintageSolidCaret/tags

### DIFF
--- a/repository/v.json
+++ b/repository/v.json
@@ -321,7 +321,7 @@
 			"author": "muffinmad",
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": ">=3000",
 					"tags": true
 				}
 			]
--- a/repository/v.json
+++ b/repository/v.json
@@ -316,6 +316,17 @@
 			]
 		},
 		{
+			"name": "Vintage Solid Caret",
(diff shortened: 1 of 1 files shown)


joshua-golub:
@muffinmad, is there a reason why you created a separate plugin rather than just enhancing the [Vintage](https://github.com/sublimehq/Vintage) command mode itself?

muffinmad:
@joshua-golub I don't think that https://github.com/sublimehq/Vintage is for current version of Vintage. It's not even with Vintage sources in my ST installation. That's why I refused to make PR there. However I think this feature must be the Vintage part.

FichteFoll:
In your `plugin_unloaded` handler you should be calling `s.erase` instead of trying to restore the value from the default/user settings file. That way you also account for syntax- and project-specific settings. Same for `_set_insert_mode_caret`.

Also, this package only works on ST3, so please adjust the ST version restrictions accordingly.

muffinmad:
@FichteFoll, thanks! Fixed and adjusted.

PR #5903, opened 2016-10-22
Add ATG Syntax package to a.json

nolanar:
  [https://github.com/nolanar/ATG-Syntax-Sublime](https://github.com/nolanar/ATG-Syntax-Sublime)
- Link to the tags page: 
  [https://github.com/nolanar/ATG-Syntax-Sublime/tags](https://github.com/nolanar/ATG-Syntax-Sublime/tags)

This package adds syntax highlighting for Coco/R #C style ATG files.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -1518,6 +1518,17 @@
 			]
 		},
 		{
+			"name": "ATG(CocoR C#) Syntax",
+			"details": "https://github.com/nolanar/ATG-Syntax-Sublime",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": ">=3084",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Atomic Color Scheme",
 			"details": "https://github.com/gerardbm/Sublime-Atomic-Scheme",
 			"labels": ["color scheme"],


joshua-golub:
@nolanar, since I am not Coco/R-savvy, how does your package compare to [this](https://packagecontrol.io/packages/Coco%20R%20Syntax%20Highlighting) existing one?

nolanar:
@joshua-golub, before creating this package I had indeed tried the package you linked. I found it to barely be usable, with no syntax highlighting being preferable. It has the following issues:

- The choice of scope names do not adhere to the recommended best practices. This results in a color scheme that is not consistent with other packages, which can be quite jarring.
- Some key features of Coco R ATGs are not highlighted correctly. Most noticeably the section at the top of a file for global fields and methods. For example [this](http://i.imgur.com/Cwuytwn.png) should all be highlighted as C#. Features such as IF resolvers have similar issues.
- Strings are poorly handled. See the left side [here](http://i.imgur.com/3MbnshW.png) on lines 53 and 62. The double quotes should be captured by the single quotes surrounding it, treating it as a character. Instead there is some unpredictable interaction between the two. The right hand side shows how it is handled in my package.

My package offers a better experience, as it does not suffer the above issues. It is very usable in it's current state, but there are one or two features that could do with being revisited. I considered updating the other package but decided to create one from scratch for several reasons:

1. It was much easier to design it from the ground up correctly rather than try modify the old file. Also, I wished to use the new .sublime-syntax format rather than the .tmLanguage format. The downside here is my package isn't compatible with Sublime Text 2, though I feel we should be encouraging the switch.
2. The existing package also includes a version for Java and C++ (though I suspect they are equally unsatisfactory). I simply wanted to focus on C# (for now), and did not want the other versions to fall out of step with the C# version.
3. The existing package has not been updated in 3 years, and the author appears to be quite inactive. My worry was that a pull request might not be noticed. I would prefer to have control over the package.

Please let me know if any changes are needed to my package before it can be considered. Further improvements to the package are planned when I have some spare time.

nolanar:
Version 1.0.0 has been released!
It now fully supports the Coco/R syntax as defined in the [official user manual](http://www.ssw.uni-linz.ac.at/Coco/Doc/UserManual.pdf#page=34).

FichteFoll:
@nolanar, that you for the detailed writeup regarding the differences between the packages and your motivation. This helps a lot and I wish everyone would do it like this (even when I have to ask).

PR #5905, opened 2016-10-23
Added Test Jumper to t.json, this allows jumping back and forth betwe…

tomcharles:
  https://github.com/tomcharles/test_jumper
  https://github.com/tomcharles/test_jumper/tags

Also make sure you:
- [x]  Used `"tags": true` and not `"branch": "master"` ([versioning docs](https://packagecontrol.io/docs/submitting_a_package#Step_4))
- [x]  Ran the tests ([tests docs](https://packagecontrol.io/docs/submitting_a_package#Step_7))

…en source and implementation files

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -454,6 +454,17 @@
 			]
 		},
 		{
+			"name": "Test Jumper",
+			"details": "https://github.com/tomcharles/test_jumper",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true,
+					"platforms": ["osx"]
+				}
+			]
+		},
+		{
 			"name": "Test Plier",
 			"details": "https://github.com/asfaltboy/Sublime-Test-Plier",
 			"labels": ["python", "testing", "build system"],


joshua-golub:
Tom,

I am reviewing this package. Two things:

1. Have you looked at the many **switch** and **test** packages to verify that this functionality doesn't already exist somewhere?

2. Assuming it doesn't, does your default key binding conflict with any of Sublime's out of the box bindings?

-Josh

tomcharles:
After looking through the _switch_ and _test_ packages, I still feel like this one is different.  The switch packages all seem to deal with switching between header and implementation files and of the test packages they seemed very basic (i.e., I saw one simply for go, and it just looks for another file named similarly and goes to it).  This could technically be a pull request to the Go package, but its name is specifically tied to Go and this would be an entire rewrite of that, I'd say.

With all of that, this is a completely customizable solution that would work for header/impl, test/impl, x/y scenarios, and I don't believe something like this exists yet (if you do already know of one, let me know, I'll check that out).

And to the second point -- you're right, this did conflict with a default sublime keybinding. I changed the keybinding to cmd+shift+. which does not exist yet -- that's pushed and there's a new tag referencing the latest commit.

Thanks,
Tom

FichteFoll:
[This](https://github.com/tomcharles/test_jumper/blob/master/src/file_checker.py) seems like over-engineering ;)

Another note, it seems you only interpret suffixes for test file names, but in Python the convention is to prefix test files with `test_`. Can your package handle that?

tomcharles:
@FichteFoll, probably is :) I don't come from python and didn't know how to mock something like that in tests, so I went the way I knew I could make work.

And I'm not sure that your second case would work out, but I'm on it!

PR #5906, opened 2016-10-23
Add emojitations

idleberg:

Also make sure you:

No, this ain't no joke 😁


FichteFoll:
In the upstream repo's readme it mentions being Python 3 only (which makes sense to me since unicode in Python 2 is a mess), so I suggest making this dependency only available to ST3.

FichteFoll:
That kinda went backwards...

PR #5908, opened 2016-10-24
Add MQL5 package

rodrigopandini:
  https://github.com/rodrigopandini/sublime_mql5
  https://github.com/rodrigopandini/sublime_mql5/releases/tag/0.1.0

Also make sure you:

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -1915,6 +1915,17 @@
 			]
 		},
 		{
+			"name": "MQL5",
+			"details": "https://github.com/rodrigopandini/sublime_mql5",
+			"labels": ["language syntax", "snippets", "build"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Mreq Color Scheme",
 			"details": "https://github.com/mreq/mreq-theme",
 			"labels": ["color scheme"],


FichteFoll:
- In your build system, you specify to run `"$packages\\User\\sublime_mql5\\buildMQL5.bat"`, but your package will be installed as `"$packages\\MQL5"`.
- The `^.*(^.*)` regex pattern is quite pointless.
- Since you want to execute a file that you distribute with your package, you will need to tell Package COntrol specifically to unextract it by adding a `.no-sublime-package` file to your repo, because it would otherwise be installed as a `.sublime-package` archive.
- Telling users to set their paths in the .bat file is problematic because every Package update will cause changes to be reset.

Instead, you have two options (that do not require a `.bat` and `.no-sublime-package`):

1. Fetch configration from a `.sublime-settings` file and run the actual executabe from a Python plugin (see the `target` parameter for build systems)
2. Run multiple commands in build system by using `shell_cmd` and combining those with `&`. Then tell users to adjusts paths in the build system using the PackageResourceViewer package, for example.

FichteFoll:
@rodrigopandini ping

rodrigopandini:
@FichteFoll I still don't have a time to accomplish this, sorry.
Soon as possible I will try to do.
Did you have a simple example about how to use TextCommand class?
My knowgements in Python and Sublime plugin API is still very limited.

FichteFoll:
Try looking into other packages, but here's an example from one of my packages: https://github.com/FichteFoll/InsertDate/blob/f7b0d1b99a242c2658641753124516b7605c398d/insert_date.py#L172-L202

PR #5911, opened 2016-10-25
Add Nunjucks Extended Syntax

thecodechef:
- https://github.com/thecodechef/nunjucks-extended
- [0.0.2](https://github.com/thecodechef/nunjucks-extended/releases/tag/0.0.2) tag:

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -569,6 +569,17 @@
 			]
 		},
 		{
+			"name": "Nunjucks Extended",
+			"details": "https://github.com/thecodechef/nunjucks-extended",
+			"labels": ["language syntax"],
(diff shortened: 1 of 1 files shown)


FichteFoll:
What are the differences to the other nunjucks package and was was your motiviation behind creating this?

In your build system, the windows overrides are redundant since you use `shell_cmd` already. The linux overrides are missing a key name (`path`).

FichteFoll:
Also, since `.sublime-syntax` is only available since ST build 3092, you have to restrict installation of this package to that version or higher.

FichteFoll:
@thecodechef ping

PR #5922, opened 2016-10-30
SLAB

increpare:
https://github.com/increpare/slab-markup
https://github.com/increpare/slab-markup/releases

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1692,6 +1692,17 @@
 			]
 		},
 		{
+			"name": "SLAB",
+			"details": "https://github.com/increpare/slab-markup-sublime",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


FichteFoll:

You should most certainly *not* specify `source.c` as your base scope for the syntax and instead use `source.slab` (or `text.slab`).

increpare:
Hi, thanks for getting back - thanks for the feedback.

I'm slightly confused, because it's been so long, but these should fix things I believe/hope:

I fixed the scope.  I also added a tag to slab-markup-sublime ( https://github.com/increpare/slab-markup-sublime/releases/tag/0.2.1 )

Also, I can't remember the conventions exactly, but maybe I linked to the wrong project in my first post (I linked to the original project rather than the plugin branch) - in which case

https://github.com/increpare/slab-markup-sublime/
https://github.com/increpare/slab-markup-sublime/releases/tag/0.2.1

FichteFoll:
>but maybe I linked to the wrong project in my first post

You did indeed, but I noticed.

PR #5931, opened 2016-11-02
Rename package

idleberg:

Also make sure you:

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2641,6 +2641,16 @@
 			]
 		},
 		{
+			"name": "ConvertChineseCharacters",
+			"details": "https://github.com/stormtea123/ConvertChineseCharacters",
+			"releases": [
+				{
+					"sublime_text": ">3000",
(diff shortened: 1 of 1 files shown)


FichteFoll:
"Converter" is a very *general purpose* name, something that doesn't really fit what your package actually does (JSON, CSON, XML). I know this is hard because I have been there myself, but please try to find a different less general but still fitting name.

idleberg:
With names like DataConverter and SerializedDataConverter already taken, there aren't many options left. Would you agree to something like ConvertThisToThat?

FichteFoll:
That's true. How about "JSON_CSON_XML_Converter"? Afaik your package is the only one to support CSON and XML, so that would make it stand out.

FichteFoll:
Actually, it's still the only one that supports CSON. Either way, this PR is conflicting with the master branch currently.

idleberg:
I'll stick with the old name

PR #5936, opened 2016-11-04
TASK: Update location of Package

DanielSiepmann:
* As package has moved over to github


DanielSiepmann:
@Kitzberger after merging this, the new repository at github should be used by package control.

PR #5941, opened 2016-11-06
Added rainmeter to r.json for sublime text 3 on windows with new maintainer

thatsIch:

 - Link to your code repository: https://github.com/thatsIch/sublime-rainmeter
 - Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/thatsIch/sublime-rainmeter/releases/tag/2.0.0

Also make sure you:

 1. Used `"tags": true` and not `"branch": "master"` ([versioning docs](https://packagecontrol.io/docs/submitting_a_package#Step_4)): done
 2. Ran the tests ([tests docs](https://packagecontrol.io/docs/submitting_a_package#Step_7)): 

Ran 7151 tests in 0.323s
  
OK

### DIFF
--- a/channel.json
+++ b/channel.json
@@ -74,7 +74,6 @@
 		"https://raw.githubusercontent.com/Medalink/laravel-blade/master/packages.json",
 		"https://raw.githubusercontent.com/mekwall/obsidian-color-scheme/master/packages.json",
 		"https://raw.githubusercontent.com/Mendor/sublime-erlyman/master/packages.json",
-		"https://raw.githubusercontent.com/merlinthered/sublime-rainmeter/master/packages.json",
 		"https://raw.githubusercontent.com/mischah/Console-API-Snippets/master/packages.json",
 		"https://raw.githubusercontent.com/Mozillion/SublimeSpeech/master/packages.json",
 		"https://raw.githubusercontent.com/naomichi-y/string_counter/master/packages.json",
--- a/repository/r.json
+++ b/repository/r.json
@@ -251,7 +251,7 @@
 					"tags": true
 				},
 				{
-					"sublime_text": ">=3000",
+					"sublime_text": ">=3092",
 					"platforms": ["windows"],
 					"tags": true
 				}
--- a/repository/r.json
+++ b/repository/r.json
@@ -251,6 +251,11 @@
 					"tags": true
 				},
 				{
+					"sublime_text": "<3092",
+					"platforms": ["windows"],
+					"tags": "st3092-"
+				},
+				{
 					"sublime_text": ">=3092",
 					"platforms": ["windows"],
 					"tags": true
--- a/repository/r.json
+++ b/repository/r.json
@@ -240,6 +240,18 @@
 			]
 		},
 		{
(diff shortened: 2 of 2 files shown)


FichteFoll:
You need to remove the other entry from [here](https://github.com/wbond/package_control_channel/blob/afe64f80d714a5fb3051158509a6727c770944f8/channel.json#L77)
or, alternatively, you could add the original package as the ST2 release so that the name "Rainmeter" would point to two different packages depending on the ST version.

I would need confirmation from @merlinthered either way.

merlinthered:
@thatsIch has taken over the development of the package, so I think it would be best to have his fork as the sole repository for both ST2 and ST3.

FichteFoll:
That's the best option certainly.

@thatsIch, please allow installation for all ST versions and remove the other entry then.

thatsIch:
@FichteFoll this is not compatible with ST2 thus I wanted to keep the old repository for all ST2 users.

How would I do that? I only say the use of the "base" tag.

In the documentations [1] I see no mentioning about sharing the same name for different repositories. In the examples I see mostly using different branches.

---

[1] https://packagecontrol.io/docs/submitting_a_package

thatsIch:
ok, I think I found the solution in the `example-repository.json`. I hope I used it correctly.

thatsIch:
I just wanted to ask if anything is wrong with this PR

FichteFoll:
`.sublime-syntax` support has been added with build 3092. Since this seems to be pretty integral to your package, please only allow installation of your version from this build onwards.

thatsIch:
Is there a way to tell the users to use Version/Tag X of the plugin if it is below version 3092?
I just switched to the new `sublime-syntax` and did not even release that (just via commit). If not I will just increment the minimal version. 

If there is no sane way you can pull that.

Thank you very much

FichteFoll:
>Is there a way to tell the users to use Version/Tag X of the plugin if it is below version 3092?

Yes, you could create prefixed tags and then specify a release channel for builds <3092 that will only pick up tags with this prefix, for example the tag `tmLang-0.1.0` would be selected with a configuration like `"tags": "tmLang-"`.

Alternatively, you could just provide the tmLanguage file alongside the sublime-language file (with the same name). Builds that understand it will select the sublime-syntax file while other builds will just pick up the legacy one.

thatsIch:
Thank you for the pointer. I used the tag prefix solution.

> Ran 7153 tests in 0.370s

FichteFoll:
Hm, I think this didn't go as planned. The commit you tagged as `st3092-2.5.2` doesn't have any tmLanguage files (https://github.com/thatsIch/sublime-rainmeter/tree/8209e728d2e0d22b1df461345bd1379b82ad8b65).

thatsIch:
You are right.. how silly of me

I replaced the synthetic tag to 2.4.0 to the point of https://github.com/thatsIch/sublime-rainmeter/tree/6ceae7e97a6d7e003a3bd201bd2270dc5dc14cfa

This is the last commit where the `tmLanguage` are in the repo

PR #5949, opened 2016-11-10
Added Electric Imp Developer package

ppetrosh:
Requested information:

 - Link to your code repository: https://github.com/electricimp/ElectricImp-Sublime
 - Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/electricimp/ElectricImp-Sublime/tags

All tests from [tests docs](https://packagecontrol.io/docs/submitting_a_package#Step_7) passed on the branch.

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -351,6 +351,16 @@
 			]
 		},
 		{
+			"name": "Electric Imp Developer",
+			"details": "https://github.com/electricimp/ElectricImp-Sublime",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Elementor",
 			"details": "https://github.com/andresdominguez/elementor-sublime",
 			"releases": [


ppetrosh:
This is a plugin for Squirrel developers for Electric Imp Platform (https://electricimp.com). It simplifies the development process as the only alternative available is the Web based IDE and also allows people to use source control for their projects. More information can be found at https://github.com/electricimp/ElectricImp-Sublime.

FichteFoll:
You probably don't want to have the debug setting enabled by default.

Please do not override default key bindings.

Please do not modify `sys.path`, unless necessary. You can use relative imports within ST3 packages.

The `sublime` API may not be available at module load time. You must wrap your `update_log_windows` call within  `plugin_loaded` to ensure that it will succeed.

ppetrosh:
Thank you @FichteFoll for the detailed review! All your concerns are valid and I'll address them.

What's your recommendation re the menu items. What would be the best fit from your POV? A submenu for the Project menu?

FichteFoll:
Firstly, you should definitely add a `.sublime-commands` file, because I, for one, prefer using the command palette over menu juggling.

Secondly, it appears that most packages insert their custom commands in one of the existing main menu entries, depending on what fits best. I've seen them in "Edit", "Selection" and "Tools", where "Tools" appears to be the catch all. Many packages don't add their commands to a menu item at all however and just rely on the command palette and key bindings, however.

I personally considered establishing a new main menu item named "Packages" where all packages could add menu items to a sub menu under their name in a centralized place, but we're way beyond the point where this could become standard. I'm still debating whether it's worth doing or not.

ppetrosh:
Got it, thank you! I'll update the sources to address all the issues that you reasonably pointed out and will get back to you.

FichteFoll:
@ppetrosh ping

ppetrosh:
@FichteFoll I'm sorry for the delay. I plan to update the pull request by the end of this or early next week. Will ping you when done. Thanks!

PR #5963, opened 2016-11-24
Add Solium Gutter

sey:

 - Link to your code repository: https://github.com/sey/sublime-solium-gutter
 - Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/sey/sublime-solium-gutter/releases

Also make sure you:

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -2078,6 +2078,17 @@
 			]
 		},
 		{
+			"name": "Solium Gutter",
+			"details": "https://github.com/sey/sublime-solium-gutter",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"platforms": ["osx", "linux"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Sonic Pi",
 			"details": "https://github.com/fbehrens/sublime_sonic_pi",
 			"labels": ["music"],
--- a/repository/s.json
+++ b/repository/s.json
@@ -2083,7 +2083,6 @@
 			"releases": [
 				{
 					"sublime_text": ">=3000",
-					"platforms": ["osx", "linux"],
 					"tags": true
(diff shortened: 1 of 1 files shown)


joshua-golub:
@sey, a couple of things:

1. We really try to avoid having alpha-quality code in Package Control

2. What does "gutter" refer to? In an editor, "gutter" means something to me other than linting.

-Josh

sey:
Thanks for your feedback. I totally understand your first point. This is now fixed.

Gutter means that the plugin uses the Sublime Text gutter next to the line numbers. I took example on the JSHint Gutter plugin that does the same for Javascript files. You can see what it does on the JSHint homepage (https://github.com/victorporof/Sublime-JSHint).
If you think it is better I can rename it (just Solium or Solidity Linter for example).

joshua-golub:
Ok, I understand now why you used "Gutter". That's fine with me. I'll make one more pass through your package shortly.

FichteFoll:
Can I just rant a bit about npm on why [this](https://github.com/sey/sublime-solium-gutter/blob/master/scripts/node_modules/ansi-regex/index.js) is a thing and not a part of [this](https://github.com/sey/sublime-solium-gutter/blob/master/scripts/node_modules/strip-ansi/index.js)?

sey:
You can but no one will hear you here ;-).  
Though it may seem over engineered the ansi regex package may be used by different packages for different things. It makes sense to only write it once in my opinion.  
It's part of the Unix philosophy that each software should do only one thing and do it well.  
Thanks for the merge.

FichteFoll:
I'm just gonna say "leftpad".

I prefer using proper (my own) judgement on whether or not a self-contained package would be better than collecting trivial parts from multiple locations, but that is subjective of course.

PR #5979, opened 2016-12-01
add MarkdownLivePreview

math2001:
Hello!

Here's my repo: <https://github.com/math2001/MarkdownLivePreview>

My tags: <https://github.com/math2001/MarkdownLivePreview/tags>

This package shows the markdown you're typing in a phantom in a view (moved to a new group).

Matt

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -445,6 +445,17 @@
 			]
 		},
 		{
+			"name": "MarkdownLivePreview",
+			"details": "https://github.com/math2001/MarkdownLivePreview",
+			"labels": ["markdown", "preview"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
The functionality that `escape_amp.py` provides is also available in the stdlib as `html.escape` (since 3.2).

math2001:
> I'm not sure why you would need the functionality

it's for escaping this kind of thing:

```
hello&world
```
I looked quickly in the doc, didn't find it, so I built my own.

> it should already be using HTML entities.

Well, now that you said that, I did some tests, and yes, it, works...

```python
>>> import lib.markdown2 as md
>>> md.markdown('test &test', extras=['fenced-code-blocks', 'no-code-highlighting'])
'<p>test &amp;test</p>\n'
```

This is weird: https://github.com/math2001/MarkdownLivePreview/issues/2#issuecomment-265995831

I've probably done some changes so that it works. I'll have a look at what was the code at this state, and I'll let you know.

PR #5980, opened 2016-12-02
Update Jimbly's Clipboard History for ST3

Jimbly:
Also switching to use tags.  I made a new tag v1.1.0, will that actually supersede the old version (v2016.12.02.02.27.26) people have installed, or is something else required to migrate from "branch" to "tags"?

code repository: https://github.com/Jimbly/SublimeClipboardHistory
tags page:  https://github.com/Jimbly/SublimeClipboardHistory/releases

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -524,8 +524,8 @@
 			"details": "https://github.com/Jimbly/SublimeClipboardHistory",
 			"releases": [
 				{
-					"sublime_text": "<3000",
-					"branch": "master"
+					"sublime_text": "*",
+					"tags": true
 				}
 			]
 		},


FichteFoll:
This is fine.

By the way, I suggest masking all your key bindings with a simple context that compares against a setting value. That way users can easily enable or disable the override bindings, depending on what you decide to default to. Feel free to ask if this is not clear to you.

Jimbly:
Thanks!  That is not clear (I pretty much blindly stumble through Python and Sublime plugin writing trying to get these to work and stop fumbling when it works ; ).  Just naming a (non-giant) example plugin that does it the "right way" would point me in the right direction =).

FichteFoll:
I don't think there is a plugin that does it as simple as I imagine here (I know Emmet has flags to disable key bindings, but they go way over the head). It basically goes like this:

**\*.sublime-keymap** (in your package)
```
[
    { "keys": ["ctrl+v"], "command": "jch_paste" ,
      "context": [
        { "key": "setting.clipboard_history.enable" }
    ]
  }
]
```

**Preferences.sublime-settings** (in the User package)
```
{
    "clipboard_history.enable": true
}
```

If you want to enable it by default, you would change the context line to `{ "key": "setting.clipboard_history.disable", "operator": "not_equal" }`.

PR #5982, opened 2016-12-03
Added LightningComponentCompletions package

amorek:

   https://github.com/amorek/LightningComponentsCompletions
   https://github.com/amorek/LightningComponentsCompletions/releases

Also make sure you:

Done
Done

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -681,6 +681,16 @@
 			]
 		},
 		{
+			"name": "LightningComponentsCompletions",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Instead of using absolute imports like `from LightningComponentsCompletions.aura_tags import tag_dict as aura_tags`, use a relative one: `from .aura_tags import tag_dict as aura_tags`. That is more flexible (in case the package is saved under a different name) and less verbose.

PR #5986, opened 2016-12-04
Add Base16 Eighties Dark Color Scheme package

pgaspar:

 - Link to your code repository: https://github.com/pgaspar/base16-eighties-dark
 - Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/pgaspar/base16-eighties-dark/tags

Thank you!

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -151,6 +151,17 @@
 			]
 		},
 		{
+			"name": "Base16 Eighties Dark Color Scheme",
+			"details": "https://github.com/pgaspar/base16-eighties-dark",
+			"labels": ["color scheme"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Base64 Fold",
 			"details": "https://github.com/DaQuirm/base64-fold",
 			"releases": [


joshua-golub:
@pgaspar, how does your color scheme compare to [these](https://packagecontrol.io/search/Base16) others? They look pretty similar to me.

-Josh

pgaspar:
@joshua-golub they are very similar. It's a variation of the base16-eighties that I used as my theme on Atom (repo [here](https://github.com/tyre/base16-eighties-dark), by @tyre) and that I wanted to replicate now that I'm back to Sublime.

Here's a comparison gif:

comparison

The name of the Atom theme was [base16-eighties-dark](https://atom.io/themes/base16-eighties-dark), so I tried going with something similar to that so people coming from Atom could find it (it has 3.8K downloads).

My intention is not to cause confusion, though, so I'll gladly take any suggestions in that regard!

FichteFoll:
Not exactly a fan of appending the suffix "-dark" to a color scheme that is already dark and in fact only changes a few selectors (not even colors), but the name being used for an Atom package is a valid argument. I don't feel strong enough on the opposing side to reject this name, however.

PR #5987, opened 2016-12-05
Added PythOnSelection package

lgiordani:

 - Link to your code repository: https://github.com/lgiordani/PythOnSelection
 - Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/lgiordani/PythOnSelection/releases

Also make sure you:

---
This change is [Reviewable](https://reviewable.io/reviews/wbond/package_control_channel/5987)

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -2572,6 +2572,16 @@
 			]
 		},
 		{
+			"name": "PythOnSelection",
+			"details": "https://github.com/lgiordani/PythOnSelection",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "PythonStautsBarShowSymbol",
 			"details": "https://github.com/scriptkitz/PythonStautsBarShowSymbol",
 			"releases": [


math2001:
Good readme, a demo would be very welcome though...

Why do you have a `PythOnSelection.sublime-settings` empty?

In your file `PythOnSelection.sublime-commands`, you have some trailling comma. It is not going to work on st2. (I don't know if you planned to support it, but anyway). 

Because it is really annoying, I've made a really simple plugin that [removes trailing commas when you save](https://gist.github.com/math2001/f0b52d8860f6b347301b74e15c0dfcca). Copy the content, create a file in your `User` folder and create a file named `remove_trailling_comma.py` and paste the content. Save and you're done!

lgiordani:
@math2001 thank you for the comment. I wasn't sure if the settings file was required or not, this is why I included an empty one. I am going to strip the commas, thanks for the plugin. Any suggestions on the demo? Do you mean an animated GIF?

lgiordani:
Can you check it now? Thanks

math2001:
@lgiordani Well, now you got a lot of demo, good :+1:

@FichteFoll I haven't tested it, but it looks good to me...

FichteFoll:
https://packagecontrol.io/packages/Insert%20Nums does pretty much the same thing with a different syntax (and different variables). The readme needs some work and GIF animations and it's probably also not that easy to discover, but other than that every feature that your package provide should be available in that package as well.

lgiordani:
Thank you for pointing it out. I did not find it.

PR #5992, opened 2016-12-06
Add ZenScript package

way2muchnoise:

 - Link to your code repository: https://github.com/way2muchnoise/ZenScript-Highlighter
 - Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/way2muchnoise/ZenScript-Highlighter/tags

### DIFF
--- a/repository/z.json
+++ b/repository/z.json
@@ -75,6 +75,17 @@
 			]
 		},
 		{
+			"name": "ZenScript",
+			"details": "https://github.com/way2muchnoise/ZenScript-Highlighter",
+			"labels": ["language syntax", "auto-complete"],
+			"releases": [
+				{
+					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


FichteFoll:

You seem to have some duplication with your `.sublime-completions` file:

https://github.com/way2muchnoise/ZenScript-Highlighter/blob/master/ZenScript.sublime-completions
https://github.com/way2muchnoise/ZenScript-Highlighter/blob/master/scripts/ZenScript.sublime-completions

Both would be read by ST when loading the package.

way2muchnoise:
the scripts folder is excluded on export when tagging as defined in the `.gitattributes` or does the whole repo get included?

FichteFoll:
No, this is fine. I edited my comment shortly after I noticed it.

PR #5995, opened 2016-12-08
Add Unit Test File Toggler

mjj2000:

 - Link to your code repository: https://github.com/mjj2000/SublimeUnitTestFileToggler
 - Link to the tags page: https://github.com/mjj2000/SublimeUnitTestFileToggler/releases

### DIFF
--- a/repository/u.json
+++ b/repository/u.json
@@ -222,6 +222,16 @@
 			]
 		},
 		{
+			"name": "Unit Test File Toggler",
+			"details": "https://github.com/mjj2000/SublimeUnitTestFileToggler",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "UnitJS",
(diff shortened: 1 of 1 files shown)


FichteFoll:
You might want to take a look at the built-in `switch_file` command:

```js
{ "keys": ["alt+o"], "command": "switch_file", "args": {"extensions": ["cpp", "cxx", "cc", "c", "hpp", "hxx", "hh", "h", "ipp", "inl", "m", "mm"]} },
```

mjj2000:
Thanks for your feedback.

`switch_file` command is to switch between two files with the same name and **different extensions**.

This plugin is to facilitate switching between source and test script with **the same extension** and supports custom filename suffix of test file. You don't have to creating test file manually so you can focus without interruption.

FichteFoll:
Reading the source code of this was more clarifying. I suggest adding both an example of which filenames the plugin maps and the language (or testing framework) this is mostly targetted at. No language I use uses this kind of naming scheme for test files, which is why I got confused.

PR #5998, opened 2016-12-08
added Bitbucket plugin

dtao:
Repository for the plugin itself:

https://github.com/dtao/SublimeBucket

I searched around and the closest thing I could find was [GitLink](https://github.com/rscherf/GitLink). However, that plugin seems to have one specific purpose: linking to source files in a code hosting service (GitHub, Bitbucket, or Codebase).

The Bitbucket plugin aims to be more specific to Bitbucket and offer more features as a result. For example, the latest version allows you to run a command (`find_bitbucket_pull_request`) to go from a line of code to the Bitbucket pull request where that line was last modified. It also supports Mercurial repos, which GitLink doesn't (I mean, "Git" is right there in the name).

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -611,6 +611,18 @@
 			]
 		},
 		{
+			"name": "Bitbucket",
+			"details": "https://github.com/dtao/SublimeBucket",
+			"labels": ["bitbucket", "git", "mercurial"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": ["osx", "linux"],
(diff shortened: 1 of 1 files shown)


dtao:
For some reason it says "1 pending check", but when I click on the link to Travis it says the build passed. What gives?

dncrews:
@dtao to trigger a rebuild, you have to either close/reopen the PR, or just push a new commit

dtao:
Looks like that worked, thanks @dncrews!

PR #6004, opened 2016-12-10
Added EditInNewTab plugin

borjacampina:
A simple plugin for edit a select text in a new tab ([original idea](https://twitter.com/Wakkos/status/807319827266367493))

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -268,16 +268,6 @@
 			]
 		},
 		{
-			"name": "Edit Preferences",
-			"details": "https://github.com/sublimator/EditPreferences",
-			"releases": [
-				{
-					"sublime_text": ">=3000",
-					"branch": "master"
-				}
-			]
-		},
-		{
 			"name": "Edit in new tab",
 			"details": "https://github.com/borjacampina/sublimetext-edit_in_new_tab-plugin",
 			"author": "Borja Campina",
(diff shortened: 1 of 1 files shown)


math2001:
In python, the convention is to use snake_case, not camelCase: `selection_text`

Also, here's a little tip:

```python
selection_text = selection_text + 'whatever'
# is the same as
selection_text += 'whatever'
```

To get the window, because you have the view `self.view`, you can do `self.view.window()` so you're sure that you have the window of the view where the text has been selected from. In this case, I don't think it changes anything because the `active_window` will always be the window in which there's the `self.view`, but it's just to let you know that you can get the `window` from a `view`, for an other plugin (because you'll probably get caught up :wink:)

Otherwise, your plugin looks good!

FichteFoll:
>In this case, I don't think it changes anything because the active_window will always be the window in which there's the self.view

This is false. Commands can be invoked on windows and views that are not currently active (using `.run_command`) and there is absolutely no reason not to use `view.window()` unless you're encountering a bug. In this situation it is unlikely to occur, because the plugin is  primarily invoked from a context menu, but it's not impossible not to.

FichteFoll:

```sh
$ git tag v1.0.0
$ git push origin v1.0.0
```

borjacampina:
Updated. Thank you @math2001 @FichteFoll

PR #6005, opened 2016-12-10
Add AutoLISP package

divtiply:
AutoCAD AutoLISP programming language package for Sublime Text.
repo: https://github.com/divtiply/autolisp-sublime
tags:  https://github.com/divtiply/autolisp-sublime/tags

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -1863,7 +1863,7 @@
 		},
 		{
 			"name": "AutoLISP",
-			"details": "https://github.com/divtiply/SublimeAutoLISP",
+			"details": "https://github.com/divtiply/autolisp-sublime",
 			"labels": ["language syntax", "lisp"],
 			"releases": [
 				{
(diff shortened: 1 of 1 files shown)


math2001:
Why did you added the `AutoLISP.sublime-settings` with the `extension` option? The `fileTypes` key in the syntax definition doesn't work?

Could you also add a screenshot in your `README`?

divtiply:
Fixed. Please review again.

math2001:
👍 

I've look at the build system: why is it only available on windows? (sorry for not asking you this earlier)

I'm not able to tell if a syntax definition is good or not, but for the rest, it looks good! 😄

FichteFoll:
Where does `autolisp.bat` come from? It's not in your repo and not mentioned as requirement in the readme.

divtiply:
Added [autolisp-console](https://github.com/divtiply/autolisp-console) repository. Updated readme.

PR #6025, opened 2016-12-22
Master

alexkorovkov:
I suppose that after deleting this line read will be shown without additional html code (see https://packagecontrol.io/packages/TeXPreview)

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -492,6 +492,21 @@
 			]
 		},
 		{
+			"name": "TeXPreview",
+			"details": "https://github.com/alexkorovkov/TeXPreview",
+			"readme": "https://github.com/alexkorovkov/TeXPreview/blob/master/README.md",
+			"labels": ["tex", "latex", "preview"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"version": "1.0.0",
+					"branch": "master",
+					"tags": true
+				}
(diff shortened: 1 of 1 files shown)


FichteFoll:
The https://packagecontrol.io/packages/LaTeXTools package also provides this feature. Do you believe your package is needed next to that and if so, why?

FichteFoll:
@alexkorovkov ping

alexkorovkov:
@FichteFoll I know, that LaTeXTools provides the same but there are some differences:
1. TeXPreview can show not only an equation but block of latex code, for example: tables, theorems and etc.
2. Preview can be shown in build window (as for me it's better when you write code because following lines don't "jump")
3. You can see the preview in external program

Actually I don't know if someone uses it.

FichteFoll:
Okay, so it seems that github was playing a trick on me because their PR diff only showed you adding a package entry and not modifying one, which is why I thought this was a package addition. However, the package is already listed.

I can't see right now what you actually changed and am to lazy to check out the commit myself, so let's just see what github makes out of it when I merge.

FichteFoll:
Well, nothing was changed ...

alexkorovkov:
Yes. It was my fault. I wanted to delete only a "readme" line. Create a new request.

PR #6028, opened 2016-12-24
Add ImportHelper

unlight:
- Link to code repository: https://github.com/unlight/sublime-import-helper
 - Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/unlight/sublime-import-helper/releases

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -271,6 +271,16 @@
 			]
 		},
 		{
+			"name": "ImportHelper",
+			"details": "https://github.com/unlight/sublime-import-helper",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "ImportJS",
 			"details": "https://github.com/Galooshi/sublime-import-js",
 			"releases": [


math2001:
Did the tests passed? 

Since you're plugin is only available on Sublime Text 3, for your settings, you could use the `edit_settings` command.

unlight:
> Did the tests passed?

Yes

> you could use the edit_settings command

Done.
unlight/sublime-import-helper@3b60c23

math2001:
Why did you put a single children? You could just add the command to edit the settings on your package's name:

```json
[
    {
        "caption": "Preferences",
        "mnemonic": "n",
        "id": "preferences",
        "children":
        [
            {
                "caption": "Package Settings",
                "mnemonic": "P",
                "id": "package-settings",
                "children":
                [
                    {
                        "caption": "Import Helper",
                        "command": "edit_settings",
                        "args": {
                            "base_file": "${packages}/ImportHelper/import-helper.sublime-settings"
                        }
                    }
                ]
            }
        ]
    }
]
```
I think that looks better, because you're in the menu `Package Settings`, clicking on the name of a package should naturally open the settings.

unlight:
> because you're in the menu Package Settings, clicking on the name of a package should naturally open the settings.

Done. unlight/sublime-import-helper@4cbc409

FichteFoll:
There are a few packages like this already, but I don't think any of these use phantoms or provide a UX like this, so I give it a pass.

Except that `alt+i` is already used as an accelerator to the "Find" main menu item. However, because that is only active for files where it matters and the binding itself seems convenient, I'll let it slide for now.

PR #6031, opened 2016-12-26
add new -- CesiumHelper

Fav:
my code repository：https://github.com/Fav/sublime-cesium
and it has test by Package Control Channel

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -413,6 +413,17 @@
 			]
 		},
 		{
+			"name": "CesiumHelper",
+			"details": "https://github.com/Fav/sublime-cesium",
+			"labels": ["Cesium"],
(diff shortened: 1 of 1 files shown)


math2001:
You should write your `README` in english.

FichteFoll:
You should name your package "Cesium Completions", since that's exactly what it is. An English readme is not necessary, but I must be able to review your package and if I wasn't (e.g. if there was lots of Python code in it), i would require you to either make an English readme or explain to me what the package does. Since this package just provides completions, it's fine.

```sh
$ git tag v1.0.0
$ git push origin v1.0.0
```

FichteFoll:
@Fav ping

PR #6033, opened 2016-12-27
Add GraphQL Syntax

dncrews:
Adds GraphQL Language Syntax Highlighting

- https://github.com/dncrews/GraphQL-SublimeText
- https://github.com/dncrews/GraphQL-SublimeText/tags

## Overview

This adds GraphQL Language syntax highlighting. There was already an [existing package](https://github.com/graphql-python/GraphQL-SublimeText), but the maintainer does not seem to be responsive to the fact that it doesn't actually work.

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -1621,10 +1621,15 @@
 		},
 		{
 			"name": "GraphQL",
-			"details": "https://github.com/graphql-python/GraphQL-SublimeText",
+			"details": "https://github.com/dncrews/GraphQL-SublimeText3",
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": ">=3092",
+					"tags": true
+				},
+				{
+					"sublime_text": "<3092",
+					"base": "https://github.com/dncrews/GraphQL-SublimeText2",
 					"tags": true
 				}
 			]


FichteFoll:
If maintainers of packages are unresponsive, it is possible to replace their packages if they don't object within at least two weeks. However, since that package uses the `.tmLanguage` format, it is compatible with ST builds <3092 and thus still interesting for some users.

You may also choose to only replace the other package (with the generous "GraphQL" name) for ST builds >=3092 while having users of older builds download the other package instead. This is accomplished using a different `base` key for a second release that targets `"sublime_text": "<3092"`. This would be ideal.

cc @graphql-python

dncrews:
@FichteFoll Thanks. I've made that change, since he [appears to be AWOL](https://github.com/graphql-python/GraphQL-SublimeText/issues/1). Guess we check back in a couple weeks to see if he has any complaints.

Also, I'm having trouble finding great documentation on how to deal with the old tmLanguage format, so that I can attempt backwards-compatibility. Is [the old TextMate documentation](https://manual.macromates.com/en/language_grammars) the best?

FichteFoll:
The old format has a fairly good introduction [here](http://docs.sublimetext.info/en/latest/extensibility/syntaxdefs.html) (biased).

FichteFoll:
Also, that might be too basic for you, since you're already familiar with the newer format, so you probably want to check out the reference page instead: http://docs.sublimetext.info/en/latest/reference/syntaxdefs.html

dncrews:
@FichteFoll I went ahead and wrote a ST2+ package as well, pointing to that for <3092. It's functionally equivalent to the 3093+ version. I couldn't find a way for different branches to be different versions, so I've got a separate repo. Now we wait.

FichteFoll:
Feel free to ping me in three days, otherwise I'll just merge when I get to it.

Note to self that the PR passed all checks.

PR #6034, opened 2016-12-27
added abcss plugin

MarkNorman:
Link to code repository: https://github.com/tenforwardconsulting/abcss
Link to the tags: https://github.com/tenforwardconsulting/abcss/tree/0.1.3

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -33,6 +33,18 @@
 			]
 		},
 		{
+			"name": "ABCSS",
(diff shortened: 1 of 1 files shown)


FichteFoll:
If you only define a key binding for OSX which overrides a default binding, you might as well not define any binding at all. You could, however, make this key binding only active in sass files (and add corresponding bindings for the other two platforms).

MarkNorman:
I added the additional key binding files and scoped them to SASS.

PR #6036, opened 2016-12-28
Add Sienna Theme package

johansatge:
Hi,

 - Code repository: https://github.com/johansatge/sienna
 - Tags page: https://github.com/johansatge/sienna/releases

---

This is a theme for `macOS`. thx!

_[edit] Even if it was designed for macOS, it should also work on other systems, that's why I didn't restrict the `platforms`._

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -1457,6 +1457,17 @@
 			]
 		},
 		{
+			"name": "Theme - Sienna",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Themes are always platform-independant.

johansatge:
Ok it's noted @FichteFoll.

thx for the merge!

PR #6042, opened 2016-12-30
Strace syntax

djuretic:

 - Link to your code repository: https://github.com/djuretic/SublimeStrace
 - Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/djuretic/SublimeStrace/tags

Also make sure you:

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -2967,6 +2967,17 @@
 			]
 		},
 		{
+			"name": "Strace Syntax",
+			"details": "https://github.com/djuretic/SublimeStrace",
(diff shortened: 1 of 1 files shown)


math2001:
Because you're using `.sublime-syntax`, you're plugin will only work with build `3103` and higher (you need to update the `s.json`. You can use PackageDev to convert it into a `.tmLangage`

djuretic:
Thanks for the feedback, I've pushed a new commit

FichteFoll:
The actual required build number would have been 3092, but that doesn't really matter.

PR #6048, opened 2017-01-03
JavaScript Ultimate now requires build >= 3103

JoshuaWise:
`JavaScript Ultimate` is dropping support for builds that don't support the `.sublime-syntax` format.

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -371,7 +371,7 @@
 			"labels": ["language syntax", "snippets", "javascript"],
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": ">=3103",
 					"tags": true
 				}
 			]


math2001:
Why? PackageDev doesn't do the job anymore?

JoshuaWise:
In order to keep advancing the sophistication of the package, and to support more features without causing unwanted side-effects or reducing maintainability, I need the `push/pop/set` functionality that `.sublime-syntax` offers.

FichteFoll:
👍 for `.sublime-syntax`

That said, have you considered contributing to the default JavaScript syntax definition so that everyone gets to enjoy *perfectly accurate* JavaScript syntax highlighting?

JoshuaWise:
@FichteFoll To be honest, I didn't even know I *could* contribute to the default  JavaScript syntax definition. That said, many of the ideas that JavaScript Ultimate employs could be considered invasive for a default syntax definition. For example, JavaScript Ultimate now walks through your code, keeping track of exactly which context it's in at all times, so that it can highlight any errors in the code. However, since code is necessarily invalid while it's being written (for example, when you're half-way through writing a statement or keyword), it uses intelligent lookaheads across multiple lines (via a trick I came up with) to know whether or not something is an actual error, or if it's just partially-written code. This is pretty experimental, and a huge depart from a normal syntax definition (it requires preprocessing the `.sublime-syntax` file, so I can have reusable "mixins" with customizable arguments)

FichteFoll:
Sounds interesting and way too much for me to investigate right now.

FWIW, the default packages have been hosted on https://github.com/sublimehq/Packages for quite a while now.

PR #6051, opened 2017-01-05
Add FileManager

math2001:
Hi!

Repo: <https://github.com/math2001/FileManager>
Tags: <https://github.com/math2001/FileManager/releases>

### Tests

- [x] ChannelRepositoryTools
- [x] package_reviewer

### What is does - real quick

Create nested files/folder from the side bar, has auto completion, move/rename etc. 
Create from context menu (when triggered on a path)
It's a mixed between SideBarEnhancement and AdvancedNewFile, much more optimised. (10 items in the context menu)

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -508,6 +508,16 @@
 			]
 		},
 		{
+			"name": "FileManager",
+			"details": "https://github.com/math2001/FileManager",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "FileMerge",
 			"details": "https://github.com/thtliife/FileMerge",
 			"labels": ["file merge", "utilities", "diff", "difference", "merge", "filemerge"],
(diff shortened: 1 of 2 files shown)


FichteFoll:
Do you know the `global` keyword? [It does pretty much this](https://github.com/math2001/FileManager/blob/ea4eb362b6f5597108f678231ac5f49b3d1e2329/sublimefunctions.py#L9).

You have several Python-2-specific code blocks in your code (and `_push-to-st2.py`) although your package only supports ST3 (uses relative imports, for example).

math2001:
> You have several Python-2-specific code blocks in your code (and _push-to-st2.py) although your package only supports ST3 (uses relative imports, for example).

yes, I need to clean this up

> Do you know the global keyword? It does pretty much this.

I'm not sure if it makes it global to every packages, that's why I've done it this way. But with your answer, I guess it doesn't. I'll test and correct it.

Thanks for merging it in anyway!

FichteFoll:
No, `global` only operates on the current module (or file, so to say).

PR #6054, opened 2017-01-06
Inspr

wzhix:
Link to the code repository: https://github.com/wzhix/inspr
Link to the tags page: https://github.com/wzhix/inspr/releases

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -596,7 +596,7 @@
 		},
 		{
 			"name": "Insert to Column",
-			"details": "https://github.com/wzhix/inspr",
+			"details": "https://github.com/abluescarab/sublime_insert-to-column",
 			"labels": ["formatting", "text manipulation"],
 			"releases": [
 				{
@@ -618,8 +618,8 @@
 		},
 		{
 			"name": "Inspr",
-			"details": "https://github.com/sethlopezme/InspiredGitHub.tmtheme",
-			"labels": ["color scheme"],
+			"details": "https://github.com/wzhix/Inspr",
+			"labels": ["translate", "chinese", "naming"],
 			"releases": [
 				{
 					"sublime_text": ">=3000",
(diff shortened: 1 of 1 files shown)


wzhix:
Is there something wrong with my pull request?

math2001:
Could you write your `readme` in English please?

You have trailing commas in you `.sublime-keymap` files. I've done a plugin [JSONComma](https://github.com/math2001/JSONComma) that deals with them for you

wzhix:
I have edit README.md in English and removed trailing commas in `.sublime-keymap` files.

FichteFoll:
You don't have the "tab" binding in your Linux keybindings file. Furthermore, it seems that your tab key bindings might be too general because it will be active in all widgets that are overlays and not panels (i.e. quick panels and goto anything). Since you usually don't need a tab character in these, it will *probably* be fine, but I want to mention it just in case since it will prevent other plugins from trying to captue the tab character in these.

For FileHistory, we implemented our [own context key](https://github.com/FichteFoll/FileHistory/blob/069ea3d3a20c20ca96b380e4a26f63d8496d2efb/Default%20(Windows).sublime-keymap#L8) for this purpose.

wzhix:
Sorry about that. I fixed the "tab" binding in Linux and added my own context key `inspr_result_overlay_visible` to prevent other plugin capturing the tab key.

FichteFoll:
Unfortunately, your context key is useless because it always returns `True`. You only want it to be true while your quick panel is open. (For that, you set a global variable to true before opening it and set it to false when it has been closed, i.e. your `on_done` callback was called.)

wzhix:
Thanks for pointing out. Is this OK?
[set a global variable](https://github.com/wzhix/inspr/blob/master/Inspr.py#L148)
[set true if quick panel is opened](https://github.com/wzhix/inspr/blob/master/Inspr.py#L239-L240)
[set false if on_done is called](https://github.com/wzhix/inspr/blob/master/Inspr.py#L296-L297)

FichteFoll:
Yes, this should do.

PR #6058, opened 2017-01-09
Add Laravel 5 Docs light color scheme

yos-virtus:
Link to the code repository: https://github.com/yos-virtus/laravel_docs_light_color_scheme
Link to the tags page: https://github.com/yos-virtus/laravel_docs_light_color_scheme/releases

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -116,6 +116,17 @@
 			]
 		},
 		{
+			"name": "Laravel 5 Docs Light Color Scheme",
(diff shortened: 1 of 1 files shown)


math2001:
Your package looks good.

> At the moment the color scheme is designed to be used only with PHP code. With other languages it might look pretty messy so be carefull.

Maybe try to make it look good with `html/css/js`, because when you code in `PHP`, you'll often code in them to.

PR #6062, opened 2017-01-11
Add Click To Partial package.

alvesjtiago:
Link to code repo: https://github.com/alvesjtiago/click-to-partial
Link to tags page: https://github.com/alvesjtiago/click-to-partial/tags

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1180,6 +1180,16 @@
 			]
 		},
 		{
+			"name": "Click To Partial",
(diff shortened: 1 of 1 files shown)


FichteFoll:
You only specify a mouse binding for OSX (macOS). Either add bindings for all platforms or none and let the user add their own bindings to his User package.

alvesjtiago:
Hi @FichteFoll, thank you for pointing that out. I've just pushed an update to the repo adding key bindings for Linux and Windows as well as updated README instructions and tag.

PR #6065, opened 2017-01-11
Added entry for the package qasm_highlighting

gtaifu:

 - Link to your code repository: https://github.com/gtaifu/qasm_highlighting
 - Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/gtaifu/qasm_highlighting/tags

### DIFF
--- a/repository/q.json
+++ b/repository/q.json
@@ -20,8 +20,7 @@
 			"releases": [
 				{
 					"sublime_text": ">=3103",
-					"tags": true,
-					"dependencies": ["Monokai Theme"]
+					"tags": true
 				}
 			]
 		},
--- a/repository/q.json
+++ b/repository/q.json
@@ -19,7 +19,7 @@
 			"labels": ["language syntax"],
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": ">=3103",
 					"tags": true,
 					"dependencies": ["Monokai Theme"]
 				}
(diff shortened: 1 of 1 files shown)


math2001:
You're suppose to fill in the pull request message...

I'm not sure you need `packages.json`. Why did you add it? (and shouldn't it be `package.json`, without a `s`?)

And because you're using `.sublime-syntax`, you're package's only working with Sublime Text 3 `>=3103`

gtaifu:
I have added the pull request messages.

It the tutorial [Submitting a Package](https://packagecontrol.io/docs/submitting_a_package), it is said that a **packages.json** (not **package.json**) is required (maybe the tutorial requires update). However, I misunderstood this paragraph since only one of the two options are required and I am using GitHub so I need not provide the package.json file. I have deleted it.

I am fine that the package can only work with Sublime Text 3. Do I need to change any place in my project to add this restriction? I have searched online but no related information was found. Thanks!

math2001:
> I am fine that the package can only work with Sublime Text 3. Do I need to change any place in my project to add this restriction?

Not in your package, but in your pull request. In the `releases` object, change the `sublime_text` value to `>=3103` (instead of `*`)

BTW, why did you add the dependency of `Monokai Theme`?

gtaifu:
@math2001 Thanks for your instruction. I have updated in the release object the version control and removed the dependency on ```Monokai Theme```, which should be optional.

PR #6068, opened 2017-01-12
Add CodeStats to default channel

Nicd:

 - Link to your code repository: https://github.com/Nicd/code-stats-sublime
 - Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/Nicd/code-stats-sublime/releases

Also make sure you:

 - [x] Used `"tags": true` and not `"branch": "master"` ([versioning docs](https://packagecontrol.io/docs/submitting_a_package#Step_4))
 - [x] Ran the tests ([tests docs](https://packagecontrol.io/docs/submitting_a_package#Step_7))

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -568,7 +568,7 @@
 					"tags": true
 				}
 			]
-		},		
+		},
 		{
 			"name": "ChaiScript",
 			"details": "https://github.com/ChaiScript/sublimetext-chaiscript",
@@ -1830,6 +1830,17 @@
 				{
 					"sublime_text": ">=3000",
 					"branch": "master"
+				}
+			]
+		},
+		{
+			"name": "CodeStats",
+			"details": "https://github.com/Nicd/code-stats-sublime",
+			"labels": ["analytics", "statistics"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
 				}
 			]
 		},


repository/c.json
```diff
@@ -568,7 +568,7 @@
 					"tags": true
 				}
 			]
-		},
```
Nicd: Not sure what happened here but it shouldn't matter ¯\_(ツ)_/¯

math2001:
Why do you have a `.no-sublime-package`? Could you add a demo to your README?

Nicd:
I tried without `.no-sublime-package` but it messed up the paths to settings files. Is it something that shouldn't be used?

There's not much to demo. You add the API key to the settings file (opened when you install) and that's it. There's only a little blip shown in the status bar when the plugin is uploading data.

math2001:
> I tried without .no-sublime-package but it messed up the paths to settings files

you have to use the `find_resources` and `load_resource` `sublime`'s methods: [doc](http://www.sublimetext.com/docs/3/api_reference.html) (both of them work even when your package is not zipped (in a `.sublime-package`)

For the demo, my bad, I confused your plugin with an other one, sorry 😕

Nicd:
I'm using this to locate my settings file:

```
default_settings_file = os.path.join(sublime.packages_path(), 'CodeStats', 'Settings', 'CodeStats.sublime-settings')
```

which is why the Sublime packaging breaks the plugin. Then I'm using `shutil.copyfile()` to copy it to `os.path.join(sublime.packages_path(), 'User', 'CodeStats.sublime-settings')`.

Should I instead use `load_resource('Packages/CodeStats/Settings/CodeStats.sublime-settings')` and then write the contents to user settings file?

What's bad about `.no-sublime-package`, out of interest, or is it just convention?

math2001:
> Then I'm using shutil.copyfile() to copy it to os.path.join(sublime.packages_path(), 'User', 'CodeStats.sublime-settings')

That's not the right way to do it :oops:. The default settings (the `.sublime-settings` in your package) should be filled with all the settings your plugin's using, with some comments to explain. The `User` settings, don't even bother touching it. (you shouldn't touch any file in the `User` directory from your plugin). Leave it unexisting, Just use this:

```python
settings = sublime.load_settings('CodeStats.sublime-settings')
settings.get('my_settings') # looks up in the *user* setting file, and then, if it doesn't have the key 'my_settings', 
# it takes the one set in your *package* setting file.
```

Because your package is only working with `ST3` (and that 99.9% of them have the latest build), you can use the `edit_settings` command in your menu. You give the the `base_file` (the settings file in *your package*, and it'll know which *user* setting file open (the one with the same name). You can give a default content, if the file doesn't already exists. Here's a [little example](https://github.com/math2001/FileManager/blob/master/.sublime/Main.sublime-menu#L10-L13).

Tell me how you go.

> What's bad about `.no-sublime-package`, out of interest, or is it just convention?

It's cleaner (in the `$packages` folder, you only get the package you *have* to tweaked), I'm not sure if it's faster (if it'd be slower, I don't think it'd be implemented)

Nicd:
Ah, thank you very much for the example! It's sometimes difficult to find the correct ways to do things. Some things I copied from other plugins but it seems they were wrong. I already use `load_settings()` but I will remove my file copying and use `edit_settings` instead.

I will refactor my plugin and reopen this pull request when I have updated it to work like  you suggested. :)

Nicd:
Hey, do you think it looks better now? It now opens the settings file with `edit_settings` on startup if the user settings file does not exist yet. It also works as a packaged version.

math2001:
Great! That's much cleaner! 

There's still something to fix though 😄 You shouldn't put in the `default`'s value all the settings (or the command). Maybe just something that says `Put your CodeStats Settings in this file`, and with an empty object (`{}`), but that's it. The *default* settings file are here to show the different options and explain them. In the `User` settings file should be only composed of the settings only, without any comment.

Nicd:
How about now? :) Diff: https://github.com/Nicd/code-stats-sublime/commit/3ea4ccfd40456f73362988229281c317bbd332e1

math2001:
That's better, but (:sorry:) showing up the settings window as soon as ST start, it's a bit weird. Maybe show a message in the status bar saying `cannot send data: please edit your settings` if the settings are wrong.

```python
call_func(lambda: call_me())
# same as
call_func(call_me)
```

PS: Thanks for this, I didn't know this trick

```python
if view.settings().get('is_widget'):
      return
```

Nicd:
Ah, I've seen plugins do that and I just copied that style. The plugin is
totally useless without the API key so that's why I made it pop up first.

I'll make it just complain in the status bar and console until the key is
set, but that will have to wait for tomorrow at least.

On Jan 14, 2017 01:47, "Mathieu PATUREL" <notifications@github.com> wrote:

That's better, but (:sorry:) showing up the settings window as soon as ST
start, it's a bit weird. Maybe show a message in the status bar saying cannot
send data: please edit your settings if the settings are wrong.

call_func(lambda: call_me())# same as
call_func(call_me)

PS: Thanks for this, I didn't know this trick

if view.settings().get('is_widget'):
      return

—
You are receiving this because you modified the open/close state.
Reply to this email directly, view it on GitHub
<https://github.com/wbond/package_control_channel/pull/6068#issuecomment-272578438>,
or mute the thread

Nicd:
Is this better? Diff: https://github.com/Nicd/code-stats-sublime/compare/1.0.2...1.0.3

math2001:
Yes, it is

FichteFoll:
To display messages on installation, the usual procedure is to use [Package Control's messaging system](https://packagecontrol.io/docs/messaging) with an `"install"` key. Alternatively, you could also show a quick prompt that can be cancelled (I do that for my InsertDate plugin).

But these are just different ways to do the same thing.

PR #6069, opened 2017-01-13
Add Prettier package.

jonlabelle:
- **Description:** A Sublime Text Plug-in for [Prettier](https://github.com/jlongster/prettier), the opinionated JavaScript formatter.
- **Code repository:** <https://github.com/jonlabelle/SublimeTextPrettier>
 - **Tags page:** <https://github.com/jonlabelle/SublimeTextPrettier/tree/1.0.0>

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -1772,6 +1772,17 @@
 			]
 		},
 		{
+			"name": "Prettier",
+			"details": "https://github.com/jonlabelle/SublimeTextPrettier",
+			"labels": ["javascript", "formatting", "format"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Pretty JSON",
 			"details": "https://github.com/dzhibas/SublimePrettyJson",
 			"labels": ["json", "pretty", "lint", "minify", "validate", "query", "json2xml"],


math2001:
In python:

```python
from subprocess import PIPE
from subprocess import Popen
# same as
from subprocess import PIPE, Popen
```

`Prettier` is too generic, maybe `JSPrettier`? (same for the command `PrettierCommand`)

### Package improvement

Why do you have a `.no-sublime-package` ? 

You need to add a `messages.json` to the root, pointing to your `install.txt`, otherwise, it won't be taken into account:

```json
{
    "install": "messages/install.txt"
}
```


For the `printWidth`, maybe you could set that if it's set to `ruler[0]`, it uses the width of the `rulers` list...

### Python improvement

Also, you can optimise this a bit:

```python
def get_config(self):
        settings = self.get_settings()
        config = settings.get('config')
        return config
def get_config(self):
        return self.get_settings().get('config')
```


```python
PLUGIN_NAME + ": couldn't find node.js. Make sure it's in your " + 
                            '$PATH by running `node -v` in your command-line.'

# you can remove the second '+' because it's string after string
PLUGIN_NAME + ": couldn't find node.js. Make sure it's in your "
                            '$PATH by running `node -v` in your command-line.'

```

And, I don't understand this:

```python
sublime.set_timeout(lambda: sublime.status_message(PLUGIN_NAME + ': code formatted.'), 0)
```

```python
['node', PRETTIER_PATH] + [config, folder]
# you can just do
['node', PRETTIER_PATH, config, folder]
```

jonlabelle:
Thanks for the quick feedback and suggestions @math2001!

### Plug-in Name Change

I can rename the plug-in and repo, url's etc. to `JSPrettier`. Would it be to easier to just submit a new pull request with the updates... and disregard this PR (#6069)?

### Package improvement

**Why do you have a .no-sublime-package?**
*Because of the node.js dependencies which install in the package/plug-in directory.*

**Missing messages.json file**
*messages.json file has now been added to the repo. root.*

### Python improvement

- The `get_config()` method has been optimized.
- The plug-in exception message has been optimized.

**And, I don't understand this:**

```python
sublime.set_timeout(lambda: sublime.status_message(PLUGIN_NAME + ': code formatted.'), 0)
```

*In my testing, it's the only way I've been able to get the status message to appear after the region is replaced w/ the transformed text.*

jonlabelle:
You can disregard this pull request, and I've created a new pull request for [JsPrettier](https://github.com/wbond/package_control_channel/pull/6076).

PR #6070, opened 2017-01-13
Adds "Select Between" and "Go To Char" plugins

arsham:
Hi,

 - Link to repositories: [Go To Char](https://github.com/arsham/GoToChar) and [Select Between](https://github.com/arsham/SelectBetween)
 - Link to the tags pages: [Go To Char](https://github.com/arsham/GoToChar/releases/tag/0.1.1) and [Select Between](https://github.com/arsham/SelectBetween/releases/tag/0.1.1) 

Also make sure you:

 1. Used `"tags": true`: Done!
 2. Ran the tests: Done!

Thank you,
Arsham

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -931,6 +931,20 @@
 			]
 		},
 		{
+				"name": "Go To Char",
+				"author": "Arsham Shirvani",
+				"details": "https://github.com/arsham/GoToChar",
+				"readme": "https://raw.githubusercontent.com/arsham/GoToChar/master/README.md",
+				"issues": "https://github.com/arsham/GoToChar/issues",
+				"labels": ["search", "text navigation"],
+				"releases": [
+					{
+						"sublime_text": ">=3000",
+						"tags": true
+					}
+				]
+		},
+		{
 			"name": "Go To Node Module Homepage",
 			"details": "https://github.com/jfromaniello/sublime-goto-node-module",
 			"releases": [
--- a/repository/s.json
+++ b/repository/s.json
@@ -747,6 +747,20 @@
 			]
 		},
 		{
+			"name": "Select Between",
+			"author": "Arsham Shirvani",
+			"details": "https://github.com/arsham/SelectBetween",
+			"readme": "https://raw.githubusercontent.com/arsham/SelectBetween/master/README.md",
+			"issues": "https://github.com/arsham/SelectBetween/issues",
+			"labels": ["text navigation"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Select By Regex",
 			"details": "https://github.com/mvoidex/SublimeSelectByRegex",
 			"releases": [


math2001:
Why are those two plugin separated? They could have been in one, no? 

And do you know about the vintage mode in Sublime Text?
`ti` goes to the next `i`, `Ti` goes to the previous `i`, etc...
 `vi'` selects inner single quotes, `ci'` removes all the content and switch to insert mode, etc...

Also, you have created a `.keymap` for each platform, but they're all the same. You can just create a `Default.sublime-keymap` to make it work on every platform.

arsham:
@math2001 They were initially one plugin, but one of my colleagues suggested separating them would be a good idea because some people might not want both functionalities. I have made these plugins for those like me, who don't use vintage mode. 
Thanks for the `.keymap` point, I didn't know that.

math2001:
> They were initially one plugin, but one of my colleagues suggested separating them would be a good idea because some people might not want both functionalities.

You're both right: some people might not want every functionality, but one plugin is, I think, better.  So, you could use the `context` key, and check in the settings if the command is enabled. Because there is only 4 commands, it shouldn't make the setting file too messy. But if you'd like to make something cleaner, have a look at what Emmet is doing. 

If you want me to help you, don't hesitate to ask me 😉 

> I have made these plugins for those like me, who don't use vintage mode.

Just curious: why not?

arsham:
Will have a look at Emmet, cheers for the suggestion.

> If you want me to help you, don't hesitate to ask me

Thanks mate!

Do you reckon it's better to make a `dependency` package and refactor the shared functionality into that package?

> Just curious: why not?

I'm just comfotrable with vim. I have tried it (also ST vintage mode a few years ago) but I found myself way more productive in normal ST3 mode.

math2001:
> I'm just comfotrable with vim. I have tried it (also ST vintage mode a few years ago) but I found myself way more productive in normal ST3 mode.

yeah, I know what you mean. I love vintage mode (although I'm a beginner), but it's a annoying that it doesn't take into account *at all* the capacity of Sublime Text (just multi cursors for example).

> Do you reckon it's better to make a dependency package and refactor the shared functionality into that package?

Oh, sorry for not being clear. I think *one* package would be great. You don't need to create a dependency. All you need to do is add a feature so that the user can disables the keybindings with a settings. 

So, you'd have to use `on_query_context`, that'd, from the view's settings, returns `true` or `false`.

Example of settings

```json
{
"your_package_name.disabled_commands": ["select_between"]
}
```

In your key binding

```json
{
    "keys": ["whatever"],
    "command": "command1",
    "context": [
        { "keys": "your_package_name.command1" }
    ]
}
```

Because this wouldn't work on it's own, you'd have a to create Listener that'd do the job:

```python
class YourPackageListener(ListenerCommand):

    def on_query_context(self, view, key, operator, operand, match_all):
        return operand not in view.settings().get('your_package_name.disabled_commands', [])
```

I haven't tested it (it might work actually), but that's the idea.

FichteFoll:
There are a couple packages with that or similar functionality already:

https://packagecontrol.io/search/expand
https://packagecontrol.io/packages/SelectUntil
https://packagecontrol.io/packages/JumpTo

FichteFoll:
@arsham could you comment on differences between your package(s) and the ones I referenced?

FichteFoll:
@arsham ping

PR #6071, opened 2017-01-13
Added CaretPositionNavigation package

rohanrhu:
- Github: https://github.com/rohanrhu/CaretPositionNavigation/

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -286,6 +286,17 @@
 			]
 		},
 		{
+			"name": "CaretPositionNavigation",
+			"details": "https://github.com/rohanrhu/CaretPositionNavigation",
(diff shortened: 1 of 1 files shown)


math2001:
You should use `tags` instead instead of `branch` in the `c.json`

You shouldn't have a `package-metadata.json` in your package. Package Control will generate it for you.

Also, you're overwriting OSX keymaps. Maybe you could use `cmd+shift+left/right` (in the keymap, replace `cmd` by `super`)? It's not used by default.

You should also use [package control's messaging](https://packagecontrol.io/docs/messaging)

math2001:
Why have you closed this PR?

rohanrhu:
I opened new pull request: https://github.com/wbond/package_control_channel/pull/6075

PR #6072, opened 2017-01-13
Add Angular CLI for Sublime Text

4ern:
This plugin allows you to run the Angular CLI commands using the Sublime Text interface.

- Link to repositories: [Angular CLI](https://github.com/4ern/angular-cli)
- Used "tags": true: Done!
- Ran the tests: Done!

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -739,8 +739,8 @@
 		},
 		{
 			"name": "Angular CLI",
-			"details": "https://github.com/4ern/angular-cli",
-			"issues": "https://github.com/4ern/angular-cli/issues",
+			"details": "https://github.com/4ern/angular-cli-for-sublime-text",
+			"issues": "https://github.com/4ern/angular-cli-for-sublime-text/issues",
 			"homepage": "http://4ern.de",
 			"author": "4ern",
 			"labels": ["angular", "cli", "angular2"],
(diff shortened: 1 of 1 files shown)


math2001:
Because you have trailing commas in your `.sublime-commands` file, it won't work in ST2. You can use [`JSONComma`](https://github.com/math2001/JSONComma) to get them automatically removed/added

The convention for python files is to use 4 *spaces* to indent your code. You're not obliged to do this, but it'd be  a bit better.

Why did you add this?

```python
def __init__(self, *arg, **kwargs):
	super(AngularCliCommand, self).__init__(*arg, **kwargs)
```

Maybe because you needed it before, but you can remove it now, it's not needed.

```python
except (IOError, OSError):
			sublime.status_message('IOError or OSError - command aborted')
```

You could use `sublime.error_message` instead, and you could show the error:

```python
except (IOError, OSError) as e:
    sublime.error_message('IOError or OSError - command aborted. Error: {}'.format(e))
```

4ern:
Hi math2001 thanks for your feedback.
I have implemented all your suggestions.

- remove trailing commas
- remove init function + change status message to error message
- convert tabs to spaces

FichteFoll:
Running `str` on the input you receive from `show_input_panel` will convert the string to a non-unicode one on ST2 (but not on ST3). I suggest you remove that "cast".

PR #6073, opened 2017-01-13
Added Syntax History plugin

malexer:
Syntax History is a plugin for Sublime Text 3 which stores the history of syntax used for files and apply it when files are reopened.

 - [Code repository](https://github.com/malexer/SyntaxHistory)
 - [Tags page](https://github.com/malexer/SyntaxHistory/tags)

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -4625,6 +4625,17 @@
 			]
 		},
 		{
+			"name": "Syntax History",
+			"details": "https://github.com/malexer/SyntaxHistory",
+			"labels": ["syntax", "syntax highlight"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Syntax Matcher",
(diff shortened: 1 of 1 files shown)


math2001:
There's an error with your messages: you're pointing to an unexisting message file: `messages/1.1.0.txt` Otherwise, your package looks good

malexer:
Fixed

FichteFoll:
I wonder if the `set_file_type` command is also executed when the syntax is changed via API (`View.set_syntax_file`). If it isn't, you'll have to resort to listening to changes to the view's settings object, where the syntax is saved under `'syntax'`.

malexer:
@FichteFoll Thanks for your suggestions. I will include them into the next releases.

malexer:
JFYI `set_file_type` command is not executed when `View.set_syntax_file` is called. I am not sure if I will add listening of this change because the original idea of this plugin was to track manual changes of the syntax.

FichteFoll:
Fair decision.

PR #6074, opened 2017-01-14
add All Black color scheme

estevanmaito:
Repository: https://github.com/estevanmaito/allblack
Tags: https://github.com/estevanmaito/allblack/tags

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -439,6 +439,17 @@
 			]
 		},
 		{
+			"name": "All Black Color Scheme",
+			"details": "https://github.com/estevanmaito/allblack",
+			"labels": ["color scheme"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Alphpetize",
 			"details": "https://github.com/foxxyz/sublime_alphpetize",
 			"releases": [


math2001:
What about using bold and italic? Here is a [topic about this](https://forum.sublimetext.com/t/suggestion-b-w-color-scheme/25380/2). If you don't want to, it's perfectly fine, it's just in case you didn't thought about using emphasis...

math2001:
Otherwise, you package looks good.

estevanmaito:
Hey, @math2001
Thanks for your suggestions. I didn't find the "Color Scheme - Vintage Terminal" when I researched themes like the one I developed (maybe a name with black or white could have helped).
It is very similar to the one I'm proposing, despite the color tones.
I'm not planning on adding bold/italic, 'cause I feel it can be a bit distracting, and I'm trying to see what happens when you develop something like reading some normal text.
BUT, I'll try the Color Scheme - Vintage Terminal tomorrow to see if my package is really needed, and tell you if this must go on.

Thanks for your attention.

FichteFoll:
## Report for allblack ################################

### Repository checks ###

Reporting 1 failures:
- No semantic version tags found (semantic versions consist of exactly three numeric parts)

No warnings

### Package checks ###

No failures

No warnings

estevanmaito:
I'm closing this PR as there is no need for another package that does the same thing (as Vintage Terminal).

PR #6080, opened 2017-01-16
add BetterSnippetManager

math2001:
Repository: https://github.com/math2001/BetterSnippetManager/
Tags: https://github.com/math2001/BetterSnippetManager/releases

- [x] ChannelRepositoryTools' tests
- [x] `package_reviewer`

# What it does, real quick

Creates snippets, and allow to find really quickly any user snippet (quick panel, all of them listed in the same one, no need to navigate)

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -506,6 +506,17 @@
 			]
 		},
 		{
+			"name": "BetterSnippetManager",
+			"details": "https://github.com/math2001/BetterSnippetManager",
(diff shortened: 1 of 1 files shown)


FichteFoll:
I discourage saving sublime-text relevant files in a hidden folder (`.sublime`) because these files matter (like any other resource file).

Eagerly awaiting "MoreBetterSnippetManager".

PR #6082, opened 2017-01-16
Added Invert Current Color Scheme to i.json

CJEdgerton:

 - Link to your code repository: https://github.com/CJEdgerton/sublime-invert-current-color-scheme
https://github.com/CJEdgerton/sublime-invert-current-color-scheme/tree/1.0.0
Also make sure you:

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -745,6 +745,16 @@
 			]
 		},
 		{
+			"name": "Invert Current Color Scheme",
+			"details": "https://github.com/CJEdgerton/sublime-invert-current-color-scheme",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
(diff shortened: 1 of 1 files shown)


math2001:
Your tag isn't valid. Have a look here: http://semver.org/

CJEdgerton:
Sorry about that, updated just now.

FichteFoll:
Instead of using `os.path.join().replace()` you could just directly invoke `"/".join()`.

You might also want to store inverted schemes in a subfolder in the User package instead of a "dummy" package (and mention this in your readme, for curious users).

CJEdgerton:
@FichteFoll 

Thanks. I've updated the readme to show where to find the inverted schemes.

PR #6083, opened 2017-01-16
Add FASTBuild plugin

Manuzor:
This is a plugin that adds basic syntax support for [FASTBuild](http://fastbuild.org) scripts to Sublime Text 3 (haven't tested it with ST2).

Code repo: https://github.com/Manuzor/FASTBuild-Sublime
Tags page: https://github.com/Manuzor/FASTBuild-Sublime/releases

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -191,7 +191,7 @@
 			"details": "https://github.com/Manuzor/FASTBuild-Sublime",
 			"releases": [
 				{
-					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


math2001:
You need to ST3 (build 3103 at least) because you're using `.sublime-syntax`. Could you add a screenshot of a file that has the syntax `FASTBuild` activated in your `README`?

Manuzor:
Good points there. I've addressed both of them. Hope it's alright now!

PR #6085, opened 2017-01-17
Add "TestCafe Test Runner" plugin

churkin:

 - Link to your code repository: https://github.com/churkin/testcafe-sublimetext

Also make sure you:

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -499,6 +499,19 @@
 			]
 		},
 		{
+			"name": "TestCafe Test Runner",
+			"details": "https://github.com/churkin/testcafe-sublimetext",
+			"readme": "https://github.com/churkin/testcafe-sublimetext/blob/master/README.md",
+			"issues": "https://github.com/churkin/testcafe-sublimetext/issues",
+			"labels": ["testing", "testcafe", "auto-testing", "test"],
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


math2001:
- The tests didn't pass. Please make sure that they do. (it might be because you need to alphabetically sort the list of package)
- You need to create at least one tag

churkin:
@math2001 I've added 0.1.0 tag for my repo

FichteFoll:
Your package will not work properly in ST3 because you are trying to write to a folder that does not exist. Package Control installs packages as `.sublime-package` zip archives, which means `__file__` will likely point to something you don't expect, and more specifically the package folder isn't created on installation.

You can either explicitly tell PC to unextract your package by adding a `.no-sublime-package` file (and prevent users from being able to override files of your package) or try to add workarounds that work even when the package is archived (the package name is in `__package__`; you can create the Packages folder if it does not exist).

Also, you should use `os.path.join` instead of hardcoding back- or forwardslashes for joining paths.

churkin:
@FichteFoll Thanks a lot!
I've fixed your remarks: https://github.com/churkin/testcafe-sublimetext/commit/1ceab8cfbc05cd116866f1e3feb4b01585bccf9f

PR #6086, opened 2017-01-18
Add NPM runner

szastupov:
- Link to the code repository: https://github.com/szastupov/npm_runner
 - Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/szastupov/npm_runner/tags

Other checks: 
 1. Uses `"tags": true` and not `"branch": "master"` ([versioning docs](https://packagecontrol.io/docs/submitting_a_package#Step_4))

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -705,6 +705,17 @@
 			]
 		},
 		{
+			"name": "npm_runner",
+			"details": "https://github.com/szastupov/npm_runner",
+			"labels": ["npm", "node", "javascript", "task", "runner"],
+			"releases": [
+				{
+					"sublime_text": ">3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "NPMInfo",
 			"details": "https://github.com/dsteinbach/npm-info",
 			"releases": [


FichteFoll:
How's this different from https://packagecontrol.io/packages/npm?

szastupov:
This one is focused on scripts and supports multiple projects.

szastupov:
To be honest, differences are subtle, and everyone's tools are always a bit opinionated so why not give user some choice?

FichteFoll:
Here's some copypasta:

1. If the feature provided is trivial, the user is prompted with different packages all achieving practically the same thing and confused about which they should use.
2. If the difference is trivial, the user is again confused about which they should use, but the package authors or maintainers are also effectively doubling their work (of maintenance). If they worked together on the same package (either by collaborating or by proposing pull requests), the community would be.
3. If one of the authors or maintainers can not keep maintaining his version of a package, it is abandoned. With two or more authors collaborating the chance of this happening is reduced.

There are exceptions to this rule in cases where developers simply can't or don't want to work together due to certain reasons, or the packages themselves are different enough in that they provide different use cases and would attract different users. However, there might also still be the option of combining these kind of packages by including functionality of one in the other.

Either way, there is no definitive objective criteria here.

szastupov:
I see your point.  Well, let's see if anyone interested upvotes this PR.

FichteFoll:
I wouldn't count on that if I were you. Besides a select few people, I doubt anyone looks at these pull requests because they are quite numerous and there usually isn't anything of interest in them. Unless you want to help with the package approval process, you  most likely do not care about what happens here at all.

I'm sure @PixnBits would gladly accept any PRs to his package that add missing functionality or improve its behavior in some way.

PR #6091, opened 2017-01-19
Add Unity Shader

waqiju:
Repo Link
https://github.com/waqiju/UnityShaderST3

Tag Link
https://github.com/waqiju/UnityShaderST3/releases

Thanks

### DIFF
--- a/repository/u.json
+++ b/repository/u.json
@@ -286,6 +286,17 @@
 			]
 		},
 		{
+			"name": "Unity Shader",
+			"details": "https://github.com/waqiju/UnityShaderST3",
+			"labels": ["language syntax", "auto-complete", "shader"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"details": "https://github.com/UnicornForest/Unity3D",
 			"releases": [
 				{
--- a/repository/u.json
+++ b/repository/u.json
@@ -287,11 +287,11 @@
 		},
 		{
 			"name": "Unity Shader",
-			"details": "https://github.com/waqiju/UnityShaderST3",
+			"details": "https://github.com/waqiju/unity_shader_st3",
 			"labels": ["language syntax", "auto-complete", "shader"],
 			"releases": [
 				{
-					"sublime_text": ">=3000",
+					"sublime_text": ">=3092",
 					"tags": true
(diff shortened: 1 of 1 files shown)


FichteFoll:
## Report for UnityShaderST3 ##########################

### Repository checks ###

No failures

No warnings

### Package checks ###

Reporting 1 failures:
- Invalid JSON (with comments)
    File: global.sublime-completions
    Exception: Invalid control character at: line 54 column 72 (char 3384)

Reporting 3 warnings:
- '.sublime-syntax' support has been added in build 3092 and there is no '.tmLanguage' fallback file
    File: UnityShader-CG.sublime-syntax
- '.sublime-syntax' support has been added in build 3092 and there is no '.tmLanguage' fallback file
    File: UnityShader.sublime-syntax

---

If you are only using `.no-sublime-package` for reading files in your package, use `sublime.load_resource` instead, which will work regardless of whether the package is archived or not.

waqiju:
Hi, FichteFoll. 
The error in package has fixed. Plz review it.
When unity version was setting to another value, the package need the permissions to rewrite symbol file and completes file. So it seems necessary to using .no-sublime-package.
And I have no plan to support older build of sublime.
Thanks for your advises.

FichteFoll:
Generating different resource files (completions, mainly) is a valid uses case for `.no-sublime-package`, yes.

However, you are using the `add_on_change` function incorrectly [here](https://github.com/waqiju/UnityShaderST3/blob/92dff0a4fc5642c4e23be83a8f8be0a68fed9cd0/UnityShader.py#L15).

1. You never call `clear_on_change` when the plugin is unloaded
2. The first parameter should be an identifier for the listener, which you later use to clear callbacks, instead of a setting name you want to listen for.
3. As a result, you'll get updates for any changes happening to the file and sometimes even when there weren't any changes to that file at all but to other settings files.

I don't specifically want you to add support for older versions, but to tell package control that a build newer or equal to 3092 is required (and not 3000).

waqiju:
Hi, @FichteFoll .
Thank you for pointing out my mistake. I certainly misunderstand the usage of `add_on_change ` and have fixed the error by following your advice. :-)
And the requirement of sublime version have improve to 3092.
BTW, I can't wait to make it done when receiving your advice. But due to away from workplace, I have to delay doing this.

PR #6092, opened 2017-01-19
Add IndentToParenthesis package

rchl:

 - Link to your code repository: https://github.com/rchl/IndentToParenthesis
 - Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/rchl/IndentToParenthesis/tags

Also make sure you:

 1. Used `"tags": true` and not `"branch": "master"` ([versioning docs](https://packagecontrol.io/docs/submitting_a_package#Step_4)) - DONE
 2. Ran the tests ([tests docs](https://packagecontrol.io/docs/submitting_a_package#Step_7)) - DONE

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -415,6 +415,17 @@
 			]
 		},
 		{
+			"name": "Indent To Parenthesis",
+			"details": "https://github.com/rchl/IndentToParenthesis",
+			"labels": ["indent", "align", "parenthesis"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Indent XML",
 			"details": "https://github.com/alek-sys/sublimetext_indentxml",
 			"releases": [
--- a/repository/i.json
+++ b/repository/i.json
@@ -417,7 +417,7 @@
 		{
 			"name": "Indent To Parenthesis",
 			"details": "https://github.com/rchl/IndentToParenthesis",
-			"labels": ["indent", "align", "parenthesis"],
+			"labels": ["indent", "align", "parenthesis", "bracket"],
 			"releases": [
 				{
 					"sublime_text": "*",


math2001:
Here's a few things you could improve

- in your readme, use a .gif: [screentogif](http://www.screentogif.com/) (I use it, it's open-source and really good)
- in python, the convention is to indent your code with 4 spaces.

Otherwise, your package looks good! I'm going to try it.

math2001:
Ironically, your package doesn't work for me on ST3, while the default option (that I wasn't aware of, so thanks), does... Are you using ST2 or ST3?

rchl:
> Ironically, your package doesn't work for me on ST3, while the default option (that I wasn't aware of, so thanks), does... Are you using ST2 or ST3?

I'm using ST3 and it works there, but it only works when `translate_tabs_to_spaces` setting is set to `true`. That is, you are using spaces for indentation. I guess I could make it clearer in documentation.
I don't think this plugin would make sense when using tabs for indentation. It would have to mix tabs and spaces then and that seems wrong to me.

> in your readme, use a .gif: screentogif (I use it, it's open-source and really good)

I don't see why screenshot would be better than just a plain text. Screenshot is not accessible for one.

> in python, the convention is to indent your code with 4 spaces.

I'm not going to change that. I'm using 2 spaces for indentation in all my code (I follow Google style).

rchl:
Found one case now when it's is better:
With code like:
```
    function(arg1, arg2) {}
```
If I decide to break the arguments into two lines (because line is too long, for example), I place the cursor after the comma and press enter. My plugin will align it properly:
```
    function(arg1,
             arg2) {}
```
while built-in functionality will mess it up like this:
```
    function(arg1,
     arg2) {}
```
If I place the cursor after the space that is following the comma, then it (built-in function) will align things properly but will leave trailing space on the line. My plugin handles both of those cases correctly.

rchl:
Updated documentation fixing mentioned issues.

deathaxe:
I confirm the bug in Sublime Text's `insert {characters: '\n'}` command rchl describes. If it is called with a whitespace after the cursor, it won't indent the arguments to the opening parenthesis as expected. 

But I would prefere to see a fix of the default `insert {characters: '\n'}`  command instead of adding another package introducing a hack to indention handling. There are so many outside right now promising improved indentions. But most (all) of them seem buggy, interfere with each other and make things even worse.

The idea and intention of this package are very good, but it's more a personal fix of a Sublime Text bug. IMO it is incomplete and thus not ready for general use.

1. It doesn't work with multiple selections as the code seems to intend to. Only the first selected argument is put onto the next line and intended correctly.
2. It doesn't work with 'translate_tabs_to_spaces': false, but it should do for general use. (Fixing this would cause trouble with TabNukerNuker)
3. It dosn't respect 'indent_to_bracket' setting.

FichteFoll:
@deathaxe could you reference a few packages that try to patch (re-)indentation behavior? Searching for "indent" I don't see any.

Imo it's not necessarily a "bug" in ST but rather a shortcoming of a less smart algorithm. Trying to detect indentation can be anything but trivial with nested (or not) brackets/braces/parens and optionally literal sequences such as strings. It surely would be possible, but it also becomes increasingly harder and I suspect that Jon didn't consider these situations to be worth his effort.

TabNukerNuker is a hack around a bug or wonky behavior in ST and I wouldn't expect plugins to be able to work with it, nor require them to.

I generall agree with your other points.

---

I, myself, intended to work on a better bracket indentation algorithm that would follow my preferred indentation style, but I never got beyond drafting some test situations/cases. I think I posted this information on the forum as well.

rchl:
> But I would prefere to see a fix of the default insert {characters: '\n'} command instead of adding another package introducing a hack to indention handling.

@deathaxe Built-in functionality is not fixable from the outside. It's either implemented in compiled in python or implemented natively in c++. With such a slow development of ST, I would't expect it to be fixed anytime soon.

> 1. It doesn't work with multiple selections as the code seems to intend to. Only the first selected argument is put onto the next line and intended correctly.

@deathaxe It works for me in this case for example:

```
function Function1(arg1, arg2, |arg3) {
  // body...
}

function Function2(arg1,| arg2, arg3) {
  // body...
}
```
(Notice two cursor positions. One after arg2 in Function1 and on after arg1 in Function2. Arguments are indented properly in this case.

> 2. It doesn't work with 'translate_tabs_to_spaces': false, but it should do for general use.

I would myself call it a missing feature.

> 3. It dosn't respect 'indent_to_bracket' setting.

It's not meant to. It's a separate package and I would expect that someone who installs it expects it to work immediately and not depend on some built-in setting.

math2001:
> > in python, the convention is to indent your code with 4 spaces.

> I'm not going to change that. I'm using 2 spaces for indentation in all my code (I follow Google style).

It's up to you to choose with how many spaces you want to indent your code, but https://google.github.io/styleguide/pyguide.html#Indentation

PR #6097, opened 2017-01-23
InSales liquid snippets

VladimirIvanin:


math2001:
I'm not sure that 2 snippets are worth doing a package. I don't know what `InSales liquid` is, but please try to add a bit more. If you want to share those 2 snippets, I'd say just create a readme that explains how to create them.

I'm quite new, so I'm not sure if those kind of package are accepted...

FichteFoll:
@math2001, there is no set policy on this, so it depends on my discretion, I guess.

PR #6098, opened 2017-01-23
SubVal package for validating subtitle timings

verrev:
https://github.com/verrev/SubVal
https://github.com/verrev/SubVal/tree/0.1.0

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -4024,6 +4024,16 @@
 			]
 		},
 		{
+			"name": "SubVal",
+			"details": "https://github.com/verrev/SubVal",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "subversion_for_sublime_txt3",
 			"details": "https://github.com/xiewandongqq/subversion_for_sublime_txt3",
 			"releases": [


math2001:
Could you add a demo in your readme please?

I looked quickly at the code, and here's what I advice you:

Your function should be pure. This mean that they use only variables that are given as parameter, and do NOT change outside variable. We can't really talk of impure function, since they're method. But, you could improve your code readability:

```python
# wrong, you have to look at the code of the function to know what's happening 
# (and IMO, I'd call it get_file_contents()
self.set_file_contents()

self.file_contents = self.get_file_contents()
```

With the second solution, it's almost a sentence, and you don't need to look at the code of the function, you just **know** what it's doing.

But, it's good to make functions (in your case methods), it brings structure to your code, and it limit the usage of comments. So 👍

verrev:
I refactored the code thinking carefully about which variables belong to the class and which have nothing to do with the state of my object. It should now be more cohesive and decoupled as well as being easier to read. Thanks for your input. I've also added a demo screenshot of the plugin.

I'm not sure how the workflow is set up - I've updated my repo with the changes, is there anything I need to do in addition?

Thanks

FichteFoll:
You need to push a new tag for package control to pick up the latest changes as a release.

PR #6099, opened 2017-01-24
fundamentals snippets

thespacedoctor:
Snippets for the python [fundamentals package](http://fundamentals.readthedocs.io/en/latest/).

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -1503,6 +1503,16 @@
 			]
 		},
 		{
+			"name": "fundamentals Snippets",
+			"details": "https://github.com/thespacedoctor/fundamentals-Sublime-Snippets",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Fuse",
 			"details": "https://github.com/fusetools/Fuse.SublimePlugin",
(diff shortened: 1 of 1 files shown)


math2001:
What's the use of this file: `snippet_settings.yaml`?

in your snippets:

```xml
 <!-- Optional: Set a tabTrigger to define how to trigger the snippet -->
    <tabTrigger>xfundamentals-create-console-logger</tabTrigger>
    <!-- Required: Base Phrase to generate tab-triggers -->
    <triggerBase>create-console-logger</triggerBase> # what's the use of this one?
    <!-- Optional: Set a scope to limit where the snippet will trigger -->
    <scope>source.python</scope>
    <!-- Optional: Description to show in the menu -->
    <description>snip</description>
    <!-- Optional: Tags to help with tab completion -->
    <tags></tags>
```

thespacedoctor:
The `snippet_settings.yaml` is used by a script I have to homogenise the names of my snippets. I can easily add it to my .gitignore if preferred. 

The `triggerBase` is used by the same script to lift the snippets and order them in the `tag.sublime-completions` file. Any snippets with tags are duplicated and prefaced with the tag name which makes it much easier to find the snippet when typing.

PR #6100, opened 2017-01-24
FewColors ColorScheme

r03:
New ColorScheme named FewColors:

Repo: https://github.com/r03/fewcolors-scheme
Tags page: https://github.com/r03/fewcolors-scheme/tags

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -356,6 +356,17 @@
 			]
 		},
 		{
+			"name": "FewColors Color Scheme",
+			"details": "https://github.com/r03/fewcolors-scheme",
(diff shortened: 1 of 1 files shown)


math2001:
What do you think about contributing to [this package](https://packagecontrol.io/packages/Color%20Scheme%20-%20Vintage%20Terminal)? I'm not sure if you should or creating an other package is a good idea...

r03:
Ok. I'll try that first

r03:
Just for reference: Here is the pull request: https://github.com/tonylegrone/terminal-sublime/pull/1

PR #6108, opened 2017-01-30
Updated k.json - added new KarmaRunner plugin

knee-cola:

 - Link to your code repository: https://github.com/knee-cola/KarmaRunner
 - Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/knee-cola/KarmaRunner/tags

Also make sure you:

CHECK!
CHECK!

### DIFF
--- a/repository/k.json
+++ b/repository/k.json
@@ -59,6 +59,17 @@
 			]
 		},
 		{
+			"name": "KarmaRunner",
+			"details": "https://github.com/knee-cola/KarmaRunner",
+			"labels": ["Karma", "Jasmine"],
+			"releases": [
+				{
+					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


FichteFoll:
You are executing the `build` command twice, for some reason.

It would also be better not to save the karma_running state in persistent storage such as `Preferences.sublime-settings` because the process handles to subprocesses launched i build systems are not persisted upon ST restarts.

knee-cola:
I have pushed a new version, in which double execution of `build` is removed, along with some other fixes (hotkeys were interfering with how Escape button interacted with outer parts of UI (i.e. multiple cursors, search bar etc)).

I hope that the new version is "production ready". Thanks!

PR #6110, opened 2017-01-30
Fix repo urls and author names

ihodev:
✔️️ Used `"tags": true`
✔️️ Ran the tests

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -1038,8 +1038,8 @@
 		{
 			"name": "Boxy Theme",
 			"previous_names": ["Theme - Otto"],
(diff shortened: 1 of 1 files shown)


FichteFoll:
Note that all your author fields are redundant because PC defaults to that from the URL already.

ihodev:
@FichteFoll Thanks, I'll remove them later :wink:

FichteFoll:
Don't bother. The will be removed eventually with an automated script, hopefully (https://github.com/wbond/package_control_channel/issues/4310). Until then, it's not worth spending time on such unimportant things.

PR #6111, opened 2017-01-30
Merge GitGutter-Edge and GitGutter

r-stein:
[GitGutter](https://github.com/jisaacks/GitGutter) and GitGutter-Edge are the same package with different release strategies. GitGutter-Edge is branch based, which is deprecated. We have discussed to merge them in https://github.com/jisaacks/GitGutter/pull/359.
If I understand the [Renaming a Package](https://packagecontrol.io/docs/renaming_a_package) section correctly we just need to remove the entry and add it to the `previous_names`. Please confirm that this is the case.

Ping @jisaacks

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -481,21 +481,11 @@
 		},
 		{
 			"details": "https://github.com/jisaacks/GitGutter",
-			"previous_names": ["Git Gutter"],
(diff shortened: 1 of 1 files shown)


jisaacks:
👍

FichteFoll:
>Please confirm that this is the case.

It is.

PR #6114, opened 2017-02-03
add  hex to int plugin

unknownuser88:

 - Link to your code repository: https://github.com/unknownuser88/hex2int
 - Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/unknownuser88/hex2int/releases/tag/v1.0.0

Also make sure you:

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -337,10 +337,10 @@
 			]
 		},
 		{
-			"name": "Hex to Int",
+			"name": "Hex to Int Preview",
 			"details": "https://github.com/unknownuser88/hex2int",
 			"author": "David Bekoyan",
-			"labels": ["hexadecimal", "converter"],
+			"labels": ["hexadecimal"],
 			"releases": [
 				{
 					"sublime_text": "*",
--- a/repository/h.json
+++ b/repository/h.json
@@ -337,6 +337,18 @@
 			]
 		},
 		{
+			"name": "Hex to Int",
+			"details": "https://github.com/unknownuser88/hex2int",
+			"author": "David Bekoyan",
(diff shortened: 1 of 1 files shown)


FichteFoll:
- Since your package only previews the decimal value of a hexadecimal, you probably want to add "status"or "preview" to its name.
- Similarly, the "convert" tag seems inappropriate.
- Your method of setting `self.view` instead of calling run with a `view` parameter is very bad practice and the problem immediately manifests in your `on_close` hook where you operate on `self.view` instead of the `view` you're given in the parameters. This means that you errornously erase the status preview of the currently active view when a view is closed in the background. Also, removing the status from an already closing view is redundant.
- You set the view's status display regardless of what you do in your loop, so erasing it *sometimes* and potentially *multiple times* when you re-set it anyway is totally redundant and might even cause flickering.
- I think you're trying to interpret the 2-complement in your `str_to_int` function, but no language I know uses 24 bit for an integer (most commonly 32) and it's quite ridiculous to just assume that all integers would be of the same size, since 64-bit integers (and 16, and 8) also exist.

unknownuser88:
I'm going to 

1. rename it to Hex to Int Preview
2. remove convert tag
3. give view to run function as parameter (like this `self.run(view)`)
4. remove on_close hook
5. erase status once before set status outside of for loop

Can you please give a little more explanation about last paragraph ? I want to fix it too

FichteFoll:
In my last point I was specifically referring to https://github.com/unknownuser88/hex2int/blob/15c3497f332920220f0c8b3f1005a4bafb95b4d7/htoint.py#L35-L36.

PR #6117, opened 2017-02-07
Add PyTest package

kaste:

 - Link to your code repository: https://github.com/kaste/PyTest
 - Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/kaste/PyTest/releases

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -2347,6 +2347,17 @@
 			]
 		},
 		{
+			"name": "PyTest",
+			"details": "https://github.com/kaste/PyTest",
+			"labels": ["python", "pytest", "testing"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Pytest Snippets",
 			"details": "https://github.com/sloria/sublime-pytest-snippets",
 			"labels": ["snippets"],


math2001:
- You don't need the `__init__.py` at the root
- why do you need the `.no-sublime-package`?
- since you're only supporting ST3, you could use the `edit_settings` command to open your package's settings.
- Good readme! 
    - I really like the theme tweak, you should ask for support from different theme! Something that checks for the settings `status_bar_error` for example. (@ihodev would, I guess, find it pretty cool!). I didn't get it to work with your solution though: you have to add it to `User/<your-theme-name>.sublime-theme`, no? Or do you have a solution to make it apply to *every* theme (which would be weird, because the color would have to be changed).
     - For the install part, have a look [here](https://github.com/math2001/JSONComma/blob/f03f416380563501dea733ce1052b37f5308a653/README.md#using-package-control) for a nicer way to install a package which isn't listed in the default package, with package control.

kaste:
I need the `__init__.py` here to get the tests pass. From the tests I need to import the 'implementation' code.

I want to be able to import this plugin from another one, that's why I added the `.no-sublime-package`

The theme tweaks must be added **at the end** of your theme file. Because the styles are applied top-down technically they must come after the `status_bar` styles your theme defines. (Updated the README to make that clearer) I have a similar status bar flashing as an addon to SublimeLinter, so that my status bar indicates if I have lint errors or warnings. Maybe I should release this functionality as a plugin.  Could this plugin then depend on the other one, t.i. auto install it?

Applied the `edit_settings` tip. Will look into the install part later.

math2001:
> The theme tweaks must be added at the end of your theme file

That's not the best way to tweak a theme. Have a look [here](https://github.com/buymeasoda/soda-theme/wiki/Theme-customisation). Like this, your change stays when you update the theme.

> I have a similar status bar flashing as an addon to SublimeLinter, so that my status bar indicates if I have lint errors or warnings. Maybe I should release this functionality as a plugin. Could this plugin then depend on the other one, t.i. auto install it?

As I said earlier, you should ask for the *theme* support. So, they'd add a rule to their theme themselves, so that it'd *fit* for their theme.

kaste:
Updated README to make the theme tweaking clearer. Updated install instructions as well.

kaste:
Removed `.no-sublime-package` in version `0.4.2`.

FichteFoll:
The theme idea for status bar coloring is quite interesting, imo. I'd generalize the setting names a bit so that different plugins could make use of this feature and themes could support these natively, without requiring user moficiations/tweaks.

PR #6119, opened 2017-02-08
add SVG Icon Snippets

idleberg:

Also make sure you:

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -4343,9 +4343,9 @@
 			]
 		},
 		{
-			"name": "SVG Icons",
-			"details": "https://github.com/s10wen/Sublime-Text-SVG-Icon-Snippets",
-			"labels": ["svg", "icon", "snippets"],
+			"name": "SVG Icon Snippets",
+			"details": "https://github.com/idleberg/sublime-svg-icons",
+			"labels": ["snippets", "auto-complete", "svg", "svg icons"],
 			"releases": [
 				{
 					"sublime_text": "*",
@@ -4354,9 +4354,9 @@
 			]
 		},
 		{
-			"name": "SVG Icon Snippets",
-			"details": "https://github.com/idleberg/sublime-svg-icons",
-			"labels": ["snippets", "auto-complete", "svg", "svg icons"],
+			"name": "SVG Icons",
+			"details": "https://github.com/s10wen/Sublime-Text-SVG-Icon-Snippets",
+			"labels": ["svg", "icon", "snippets"],
 			"releases": [
 				{
 					"sublime_text": "*",
--- a/repository/s.json
+++ b/repository/s.json
@@ -4354,6 +4354,17 @@
 			]
 		},
 		{
+			"name": "SVG Icon Snippets",
+			"details": "https://github.com/idleberg/sublime-svg-icons",
+			"labels": ["snippets", "auto-complete", "svg", "svg icons"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "SVG to JSX",
 			"details": "https://github.com/tylergould/sublime-svg-to-jsx",
 			"labels": ["svg", "jsx", "react"],


math2001:
This package already exists: https://packagecontrol.io/packages/SVG%20Icons, no?

idleberg:
Not quite, my package supports 1269 icons, the other only 16. Since web-developers are slowly moving away from icon fonts to SVG icons, I expect there will be plenty more icons to add in the future. I maintained the [Icon Fonts](https://github.com/idleberg/sublime-icon-fonts) package for several years now.

math2001:
Unless SVG Icon is dead (no modification since 1 year now, and the author is not really active, I guess the solution to tell is to PR), you should consider contributing to the original repo.

idleberg:
SInce when is it policy to encourage single packages? There currently are at least two packages for Bootstrap 3 snippets, four packages for CoffeeScript highlighting, countless snippets and highlighting packages for JavaScript, not to mention the Better <something> packages. Following your argument, there would be one of each only.

Unlike the previously mentioned packages, there isn't even an intersection of same snippets. I recall an instance, where a package of mine was rejected for having a generalized name. I wonder why this did not apply to “SVG Icons” when it's just a pack containing a hand full of social media logos.

My snippets are usually available for three platforms: Atom, Sublime Text and Visual Studio Code. The creation is fully automated and I aim at providing consistent behaviour on all platforms, making it easy for everyone switching editors or having to work with multiple editors. Therefore, I need to be in control of my snippets.

Lastly, it's the commit history of the author of “SVG Icons” that makes me doubt any PRs would be merged as quickly as I prefer to update my own packages.

math2001:
> Since when is it policy to encourage single packages?

https://packagecontrol.io/docs/submitting_a_package#Step_1

> My snippets are usually available for three platforms: Atom, Sublime Text and Visual Studio Code. The creation is fully automated and I aim at providing consistent behaviour on all platforms, making it easy for everyone switching editors or having to work with multiple editors. Therefore, I need to be in control of my snippets.

That'd be a good reason to submit your package, I guess.

As I'm not an "old" guy here, I'll won't say if this package should be merged in or not.

FichteFoll:
>SInce when is it policy to encourage single packages?

For at least as long as I do the reviews. Note that policy != requirement. There are several reasons why multiple packages with the same or similar functionality exist, including myself not realizing it's a duplicate or authors not willing to collaborate or the packages differing enough that two separate versions are warranted.

That said, since your package is much bigger than @s10wen's, I'd suggest integrating the icons provided by that package into @idleberg's and replacing the current package.

@s10wen, what do you think of this?

Otherwise, I would like to assign a different name to the current "SVG Icons" package, since that has a small and specific set of icons only.

s10wen:
I'm happy to go with whatever people think is best.

I'm not that familiar with how other repos / packages deal with multiple platforms. Is it normally the case that they all live together in one repo, or separate for each platform?

@idleberg how about I add you as a collaborator and you can add in the extra icons?

idleberg:
> Otherwise, I would like to assign a different name to the current "SVG Icons" package, since that has a small and specific set of icons only.

IMHO, “SVG Logos” would make for an appropriate name

> That said, since your package is much bigger than @s10wen's, I'd suggest integrating the icons provided by that package into @idleberg's and replacing the current package.

Aside from the reasons given in a previous post, I'd have a hard time integrating for legal reasons. My snippets are support popular SVG icons and only insert tags and class-names, I don't have to worry about the actual icons or whether they might violate usage terms. In contrast to that, the logos in the SVG Icons package are full paths. And lastly, Font Awesome v5 will be released in May as SVG icons, and it already includes all of these logos.

FichteFoll:
@idleberg, so, what does that mean for us?

idleberg:
It seems like this whole discussion only took off because of @math2001's concerns about the name. I'm open to renaming it, but I'm not willing to change the automated process that produces the completions. I thought somebody else might appreciate this package, just like the 15k users of my rather similar Icon Fonts package. Unless you can think of a better compromise, feel free to close the PR and I'll distribute this package otherwise.

wbond:
The purpose of the review process is to try and prevent confusion in names and help identify problems with submissions or the packages themselves.

If we are pushing people to distribute via other means, that would suggest to me we may be being a little too picky. Suggestions on the package contents should be suggestions only, unless security is the issue.

For naming, the only real reason I could see blocking would be an exceedingly confusing name, such as just adding a single `s` at the end.

Considering how long @idleberg has been a part of the community and the number of contributions he has made, I would be sad if we pushed him to distribute via other means. We aren't trying to make the submission process so hard that people give up, mostly just trying to help make packages more likely to be successful and useful without being harmful to PC users.

FichteFoll:
> If we are pushing people to distribute via other means, that would suggest to me we may be being a little too picky. Suggestions on the package contents should be suggestions only, unless security is the issue.

This is definitely true. I do ultimately reserve the right to reject packages for reasons (clear  and easy duplicate, malicious intend, etc.), though none of these would apply to the package at hand. My question in my latest comment was merely in order to ask for clarification, because I could not derive a cause of action from your last comment, @idleberg. Notably the Font Awesome mention. In reality, I'd much rather have this package take over the existing "SVG Icons" name because its purpose is in fact more general, but regardless of whether @s10wen agrees with this or not, this could turn out to be troublesome for users because Package Control isn't built for this particular kind of situations.

Anyway, since the packages are clearly incompatible due to copyright issues and due to the other reasons mentioned above, I'll merge this as-is.

PR #6121, opened 2017-02-08
Add uroboroSQL Formatter package

shout-star:

 - Link to your code repository: https://github.com/future-architect/Sublime-uroboroSQL-formatter
 - Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/future-architect/Sublime-uroboroSQL-formatter/releases

Also make sure you:

### DIFF
--- a/repository/u.json
+++ b/repository/u.json
@@ -454,6 +454,17 @@
 			]
 		},
 		{
+			"name": "uroboroSQL Formatter",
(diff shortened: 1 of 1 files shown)


FichteFoll:
I would suggest to refrain from modifying `sys.path` and instead use relative imports, but your bundled dependencies all make rigerous usage of absolute imports, which would make this a hassle.

PR #6122, opened 2017-02-09
Added CodeMap package

oleg-shilo:
Please find the new CodeMap plugin at GitHub:
- Code repository: https://github.com/oleg-shilo/sublime-codemap
- Tags:  https://github.com/oleg-shilo/sublime-codemap/tags

The plugin description can be found [here](https://github.com/oleg-shilo/sublime-codemap/blob/master/README.md)

Thank you,
Oleg

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1815,6 +1815,16 @@
 			]
 		},
 		{
+			"name": "CodeMap",
+			"details": "https://github.com/oleg-shilo/sublime-codemap",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Codepad.org Paste Plugin",
 			"details": "https://github.com/dangelov/Paste-Codepad.org",
 			"releases": [


FichteFoll:
## Report for sublime-codemap #########################

### Repository checks ###

No failures

No warnings

### Package checks ###

Reporting 1 failures:
- unable to load `messages.json`
    File: messages.json
    Exception: Expecting property name enclosed in double quotes: line 5 column 1 (char 111)

Reporting 1 warnings:

---

It appears you are storing data in your package location. I suggest using a sub directory in `sublime.cache_path()` for this purpose, which you can clear periodically or on `plugin_unloaded`.

oleg-shilo:
Thank you, the latest release (tag: 1.0.3) has the problems/concerns addressed:
* _unable to load messages.json_
Fixed by removing trailing commas. 
* _I suggest using a sub directory in sublime.cache_path()_
The suggested approach has been implemented
* _'.no-sublime-package' is defined. Please verify that it is really necessary_
Unfortunately it is the way the package needs to be deployed. The plugin file system (package folder) is required for storing user defined custom mappers as well as the mapper sample/template.

FichteFoll:
It appears your cache_path change didn't go through, since I can't find it in your code.

Regarding the 'plugin file system', I'm condifent that the current method is suboptimal because it suggests to place custom mappers inside the package's folder. The entire folder will, however, be wiped (moved to Backup) on each package upgrade. As such, it would be much better to either suggest specifying these mappers elsewhere, or to distribute the package as a normal `.sublime-package` and then import bundled plugins directly or loading by their contents via `sublime.load_resource`.

oleg-shilo:
- My apologies, indeed the `sublime.cache_path()` change didn't get through.
Done now. Line 21: https://github.com/oleg-shilo/sublime-codemap/blob/master/code_map.py

- I also reviewed the way the custom mappers get loaded and it is already fully decoupled from the package's folder. Meaning the there is no reason for not allowing `.sublime-package` package mode. Thus I have removed '.no-sublime-package' file. Tested via Repository Chanel - seems OK.

One question though. I keep creating new tags when I want to let you test/review my adjustments assuming you cannot do this without a new tag. Am I right doing so? Or I can avoid multiplying tags until the package completely passes the validation?  

Thank you.

oleg-shilo:
Hi there,

The two concerns of FichteFoll have been fully addressed now.

And I am just wondering if there anything else that still prevents the package from going through to the plugins repository?
 
Thank you

FichteFoll:
>I keep creating new tags when I want to let you test/review my adjustments assuming you cannot do this without a new tag. Am I right doing so? Or I can avoid multiplying tags until the package completely passes the validation?

Well, it certainly doesn't hurt to have more tags. Package Control will index all tags on the repo the moment it is added to the channel (i.e. this PR is accepted) and select the one of highest precedence for installation for each client (in case of multiple release channels for different ST versions or platforms).

PR #6124, opened 2017-02-11
Added url-utils package.

ulasozguler:
- Link to my code repository: https://github.com/ulasozguler/sublime-url-utils
 - Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/ulasozguler/sublime-url-utils/tags

### DIFF
--- a/repository/u.json
+++ b/repository/u.json
@@ -445,6 +445,16 @@
 			]
 		},
 		{
+			"name": "url-utils",
+			"details": "https://github.com/ulasozguler/url-utils",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": "st3-"
+				}
+			]
+		},
+		{
 			"details": "https://github.com/mastahyeti/URLEncode",
 			"releases": [
 				{


math2001:
Since all your keymaps are the same, you can just create one with no platform specified: `Default.sublime-keymap`

ulasozguler:
@math2001 I didn't realize that was possible. Updated my code. Thanks!

vovkkk:
Generally speaking you should avoid using <kbd>super</kbd> on Windows because most certainly your choice will conflict with global shortcuts (so this is the same as not provide any default keybindings at all).

For example, <kbd>super+u</kbd> opens Utility manager (accessibility apps), <kbd>super+e</kbd> opens new Explorer window, <kbd>super+d</kbd> shows Desktop (i.e. minimises all windows).

ulasozguler:
I've reassigned them to `ctrl+alt+e` and `ctrl+alt+d` on Windows. @vovkkk 
Thanks for the warning!

FichteFoll:
Your `ReplaceCommand` overrides the default `replace` command. I suggest using this class as a mixin and not subclassing `sublime_plugin.TextCommand`.

Also, the control and alt modifier keys are usually equivalent to the AltGr (the right alt key) in Windows and used for special characters. `ctrl+alt+e` inserts the Euro sign (`€`) on a German keyboard, for example.

So, this package is obviously very similar to https://packagecontrol.io/packages/URLEncode (and a few others), but it also provides additional functionality that wouldn't fit either of these packages' scope.

ulasozguler:
I've changed `ReplaceCommand` to `ReplaceCommandBase`, so it doesn't get recognized as a command now.

Shortcuts were fine on my test but I guess my keyboard layout is different.I've updated them to `["ctrl+shift+y", "ctrl+shift+e"]` and `["ctrl+shift+y", "ctrl+shift+d"]`.

FichteFoll:
It will still get picked up as `replace_command_base`, but at least that's not overriding anything anymore. To be fair, a lot of people are using this pattern without realizing how ST picks up plugin subclasses, so I'm not going to force you to do it "the correct way". As long as it doesn't break other stuff (like previously), this should be good enough.

PR #6125, opened 2017-02-12
Added package "PlotGraph", that allows plotting selected number columns

tibsel:
with matplotlib.

 - Link to your code repository: https://github.com/tibsel/PlotGraph
 - Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/tibsel/PlotGraph/tags

Also make sure you:

[Running_Default_Channel_Tests_Results.txt](https://github.com/wbond/package_control_channel/files/769626/Running_Default_Channel_Tests_Results.txt)

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -1435,6 +1435,16 @@
 			]
 		},
 		{
+			"name": "PlotGraph",
+			"details": "https://github.com/tibsel/PlotGraph",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "PM2-Snippets",
 			"details": "https://github.com/riyadhalnur/PM2-Snippets",
 			"releases": [


tibsel:
PlotGraph has been updated to version 0.1.1
Respective tag has been set: [https://github.com/tibsel/PlotGraph/tags](https://github.com/tibsel/PlotGraph/tags)

math2001:
Your plugin requires `python` installed, with `matplotlib`. You should mention that in your readme... 

I've never used `matplotlib` before, but couldn't you export your graph as an image, and then show it in a phantom, under the selection? Or am I missing something...

If you want to use the second solution, you'll have to create a dependency, and then use it for your package.

tibsel:
Oh boy.. you are totally right! 
I added that info to the settings file as a comment, but it should definitely also be in the Readme. 

I'll update that. 

Thanks for the feedback!

tibsel:
@math2001 
Regarding your second point: 
That also sound like a good option.
The plot window created by mathplotlib has the advantage that it is dynamic. So I can zoom in/out, pan etc. in case a specific detail of the plot is more interesting to me. 
Also, the mathplotlib plot window allows for saving the gained view as an image, thus enabling the use to export and share a quick curve comparison.

tibsel:
@math2001 I have updated the README in [0.1.2](https://github.com/tibsel/PlotGraph/tags). 
Tag for new release version has been set.

FichteFoll:
## Report for PlotGraph ###############################

### Repository checks ###

No failures

No warnings

### Package checks ###

Reporting 1 failures:
- The binding ['f3'] unconditionally overrides a default binding
    File: Default.sublime-keymap

No warnings

---

You're [executing a Python file bundled in your package](https://github.com/tibsel/PlotGraph/blob/d5996d50f0558ca449f7c846aa7ff1efc80cd6fc/plotgraph.py#L96). However, you must consider the following:

1. You need to specify a `.no-sublime-package` file because otherwise your package will be installed as a `.sublime-package` archive and the file you want to point python to won't exist.
2. You need to take care of potential whitespaces in the path to the Python file.
3. Windows doesn't like single quotes.

tibsel:
@FichteFoll

tibsel:
Haha.. well of course the double quotes around the path worked to be able to have white spaces in it... _(note to my self)_ you just have to make no typos when adding the white-spaces-path to the `User/PlotGraph.sublime-settings` file...

New release: [0.1.4](https://github.com/tibsel/PlotGraph/tags) 
Tested in Linux (GNOME-Ubuntu) and Windows 10.

Thanks again @FichteFoll for the hint. 

Best regards.

tibsel:
Thanks for the merge :+1: :smile:

PR #6126, opened 2017-02-13
Add ChromiumTodoView-sublime to available packages

dougt:

https://github.com/dougt/ChromiumTodoView-sublime

https://github.com/dougt/ChromiumTodoView-sublime/tags

Also make sure you:

Done.

Done.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -945,6 +945,16 @@
 			]
 		},
 		{
+			"name": "ChromiumTodoView",
+			"details": "https://github.com/dougt/ChromiumTodoView-sublime",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Chuby Ninja Color Scheme",
 			"details": "https://github.com/jturcotte/SublimeChubyNinja",
 			"labels": ["color scheme"],


math2001:
You need a `readme.md` file

dougt:
> You need a readme.md file

Done.

> why don't you just import your script get_chromium_issue_data.py instead of running it as a subprocess?

The python version that used in Sublime Text is different and incompatible with the version required for depot_tools.  See the comment here: https://github.com/dougt/ChromiumTodoView-sublime/blob/master/ChromiumTodoView.py#L53

math2001:
But, a user that doesn't have python installed (the right version, btw) on its system *cannot* use this plugin?

dougt:
> But, a user that doesn't have python installed (the right version, btw) on its system cannot use this plugin?

Yes, that is correct.  To use this plugin, you need to have depot installed and configure the settings to point to it.  Note, that in order to build chromium, people use a set of tools called depot.  (You can read about them here: https://dev.chromium.org/developers/how-tos/install-depot-tools).  These tools require Python 2.7+.  So, a user may have newer versions of Python installed on their system, but they typically will use 2.7 for chromium development.

> BTW, this doesn't work on my machine. Do we need to install them?

Please make sure you have set up a chromium build, verified the settings for this plugin are correct.  If it still doesn't work, please send me a console log.  Happy to see what's going on.

> What is incompatible?

A ton! Someone probably could go update all of the chromium tooling, but I don't think anyone has had time/desire.  For example, the depot tooling still uses 2.7-isms like "print a" instead of "print(a)".

FichteFoll:
Since I just mentioned this in a different PR, I'll just reference that here: https://github.com/wbond/package_control_channel/pull/6125#issuecomment-280937670
It's the second part that is relevant.

FichteFoll:
@dougt ping

dougt:
merge conflict resolved.

As for protecting windows developers from path woes, I sympathize.  I will add better logging to console if I can't find any of the paths in settings.  That's a good suggestion.

FichteFoll:
It seems you misunderstood. What I was going at was the need for a `.no-sublime-package` file.

dougt:
Ah. Got it.  Fixed.

PR #6131, opened 2017-02-15
New package added: CodeforcesUtil

shankhs:
- Link to code repository: https://github.com/shankhs/CodeforcesUtil
- Link to the tags page: https://github.com/shankhs/CodeforcesUtil/tags
- Link to the release page: https://github.com/shankhs/CodeforcesUtil/releases

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1729,7 +1729,7 @@
 			"author": "shankhs",
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": ">=3000",
 					"platforms": ["osx", "linux", "windows"],
 					"tags": true
 				}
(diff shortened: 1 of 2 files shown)


FichteFoll:
The `readme` url must point to markup text directly. Just remove it, as Package Control knows what to pick by default.

It's quite unconventional to add a top menu entry for a plugin (and if all plugins did that, it would become unusable rather quickly), so I strongly suggest not doing that. A Command Palette entry would be more important imo.

You're leaking a file handle here: https://github.com/shankhs/CodeforcesUtil/blob/2c2ed2da6e756af93d28479baab3d19e5598b679/codeforces_problem_parser.py#L81
Windows' doesn't like that too much.

shankhs:
@FichteFoll Thanks! I am still learning python :) I have modified the code to implement your suggestion.

FichteFoll:
Now you only need to create and push a new tag with your changes since the last one and we're ready to go.

shankhs:
@FichteFoll Done. Thanks!

PR #6134, opened 2017-02-17
Add dependency: package_setting_context

math2001:
This package is a *dependency*, but not a regular one: it adds a listener, instead of providing a library.

It allows you to do this:

```json
[
    {
        "keys": ["ctrl+l"],
        "command": "my_super_command",
        "context": [
            { "key": "plugin_setting.YourPackage.enabled_shortcut_my_super_command" }
        ]
    }
]
```

It looks in the *package* setting, and enable/disable the key binding depending on the value. It supports nested keys ([it can look into `dict` and `list`](https://github.com/math2001/PackageSettingContext#further-infos))

Repository: https://github.com/math2001/package_setting_context
Tags: https://github.com/math2001/package_setting_context/releases

The ChannelRepositoryTools' tests passed. 

I need this dependency in [FileManager](https://github.com/math2001/FileManager/issues/9), so if you think that's wrong, but you don't have time to look deeper, don't hesitate to let me know! 😄

### DIFF
--- a/repository/dependencies.json
+++ b/repository/dependencies.json
@@ -364,7 +364,7 @@
 			]
 		},
 		{
-			"name": "PackageSettingContext",
+			"name": "package_setting_context",
 			"load_order": "01",
 			"description": "Allow final user to disable key bindings with very few changes needed from the dev",
 			"author": "math2001",
--- a/repository/dependencies.json
+++ b/repository/dependencies.json
@@ -364,6 +364,21 @@
 			]
 		},
 		{
+            "name": "PackageSettingContext",
+            "load_order": "01",
+            "description": "Allow final user to disable key bindings with very few changes needed from the dev",
+            "author": "math2001",
+            "issues": "https://github.com/math2001/PackageSettingContext/issues",
+            "releases": [
+                {
+                    "base": "https://github.com/math2001/PackageSettingContext",
+                    "sublime_text": "*",
+                    "platforms": ["*"],
+                    "tags": true
+                }
+            ]
+        },
+		{
 			"name": "paramiko",
 			"load_order": "51",
 			"description": "Python implementation of the SSHv2 protocol - http://paramiko-www.readthedocs.org/en/latest/index.html",
--- a/repository/dependencies.json
+++ b/repository/dependencies.json
@@ -364,20 +364,20 @@
 			]
 		},
 		{
-            "name": "PackageSettingContext",
-            "load_order": "01",
-            "description": "Allow final user to disable key bindings with very few changes needed from the dev",
-            "author": "math2001",
-            "issues": "https://github.com/math2001/PackageSettingContext/issues",
-            "releases": [
-                {
-                    "base": "https://github.com/math2001/PackageSettingContext",
-                    "sublime_text": "*",
-                    "platforms": ["*"],
-                    "tags": true
-                }
-            ]
-        },
+			"name": "PackageSettingContext",
+			"load_order": "01",
+			"description": "Allow final user to disable key bindings with very few changes needed from the dev",
+			"author": "math2001",
+			"issues": "https://github.com/math2001/PackageSettingContext/issues",
+			"releases": [
+				{
+					"base": "https://github.com/math2001/PackageSettingContext",
+					"sublime_text": "*",
+					"platforms": ["*"],
+					"tags": true
+				}
+			]
+		},
 		{
 			"name": "paramiko",
 			"load_order": "51",


FichteFoll:
Dependencies interfacing directly with Sublime Text is only possible because of an implementation detail (that is subject to change!) and thus not officially supported. No other dependency behaves like this.

You may only provide custom `loader.py` code, which is guaranteed to be loaded by Sublime Text. Please be aware of how dependency loading and upgrade happens and what the potential consequences are. Notably, you should keep `loader.py` code to a minimum and only use it to delegate somewhere else *if it isn't loaded or done already*.

Also, please follow the implicit naming convention of lowercase letters.

math2001:
Ok... So, if I understand right, the `loader.py` *load* should be the only one loaded and therefore load the actual Listener...

> Please be aware of how dependency loading and upgrade happens and what the potential consequences are.

Could you explain this to me really quickly (or is it mentioned in the docs? I've read the `dependencies` page again, and didn't find what you could mean)

> Also, please follow the implicit naming convention of lowercase letters.

Should it be something like `package_setting_context` then?

FichteFoll:
>Should it be something like package_setting_context then?

Yes.

>Could you explain this to me really quickly

Well, it's not that trivial, but I can try. It's not mention in any docs either, as far as I'm aware.

Dependencies get an implicit `loader.py` code, that constructs a few paths based on the dependency name and ST's packages path and also on the ST version and platforms, checks whether the folders exist and whether they are already in `sys.path`. If they exist but are not in `sys.path`, they are appended. Note that this specifically means that this doesn't hapen more than once *even if the code is executed again*. This loader code is put into the artificial `Installed Packages/0_package_control_loader.sublime-package` package.

The second part is important because *for each dependency removal or installation*, including updates (which are both), this loader package has to be ignored, replaced with a modified file and then re-enabled, which means all of the loader code is executed again.

math2001:
Thank you very much for this explanation! 

So, I should find a way to create a `loader.py` which loads this plugin only when st starts, not when it is removed & installed? Is it possible?

Or should I add the code in a file called `loader.py`?

FichteFoll:
You should add a `loader.py` file to your repo and place all other (Python) files in a subdirectory. Then, in this file, you need to make sure that certain actions aren't executed twice even when the file is reloaded.

math2001:
(sorry for the delay 😕 )

I pushed a new commit. It should be a tiny bit better (`tests.py` can only be loaded manually). i just `from ... import` the event listener. 

There is no action exectuted twice, it just defines functions and add a listener. I tried to look if the listener was defined, but I didn't find the list of "normal" `EventListener` in `sublime_plugin`. 

Is it good? Am I missing something still?

FichteFoll:
I think the code this is fine now. You didn't rename the dependency to `package_setting_context` though.
Also, just FYI, it's "invalid" not "unvalid".

math2001:
> You didn't rename the dependency to package_setting_context though

Sorry, I forgot. I'll do this

> Also, just FYI, it's "invalid" not "unvalid".

Thanks 🙂 

> You might also want to implement the other context operators.

Yep, I'll do this too.

math2001:
- [x] Renamed dependency to `package_setting_context` 
- [ ] Support other operator

math2001:
@FichteFoll?

repository/dependencies.json
```diff
@@ -364,6 +364,21 @@
 			]
 		},
 		{
+			"name": "PackageSettingContext",
```
FichteFoll: This still needs to be changed.

math2001:
@FichteFoll Sorry, I forgot about this one too. Done now.

FichteFoll:
All right. Unfortunatly I (just now) noticed that you haven't pushed a new tag for all the recent changes. Please do that and I'll gladly (finally) merge.

By the way, you should also clarify that the context key should be constructed as `plugin_setting.<SettingFileName>.<setting_name>` instead of the Package's name, since they may differ.

math2001:
- [x] Tag changes
- [x] update readme (BTW, I left a mistake in there: the prefix had been changed to `package_setting.`

Sorry for forgetting so many things... 😖

FichteFoll:
There doesn't appear to be a new tag pushed on the repo.

math2001:
Done.

math2001:
The `loader.py` isn't included when installing with `Satisfy Dependencies`... Is this a bug?

FichteFoll:
It should be added to `0_package_control_loader.sublime-package`. I'll investigate later whether anything isn't working as intended.

math2001:
> It should be added to 0_package_control_loader.sublime-package

It is, sorry. I didn't realized this part. Anyway, I pushed a new tag and it seems to load properly now.

> I'll investigate later whether anything isn't working as intended.

So don't worry about this.

PR #6142, opened 2017-02-18
added Nova Template Highlighter

daveismyname:
Link to code repo https://github.com/nova-framework/template-highlighter-sublime-text
Link to tags page https://github.com/nova-framework/template-highlighter-sublime-text/tags

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -666,7 +666,7 @@
 			"labels": ["nova", "syntax highlighter"],
 			"releases": [
 				{
-					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


math2001:
Your package is only available on ST3 since you use `.sublime-syntax`.

daveismyname:
Ok that makes sense I've added the release for ST3, please do let me know if I've done this incorrectly.

FichteFoll:
More specifically, `.sublime-syntax` is available starting with build 3092, so you should specify that as the minimum requirement.

PR #6145, opened 2017-02-19
Add InsertTimestamp package

fricklerhandwerk:
This package allows to insert a timestamp at cursor position. You can optionally specify User key bindings and pass a `strftime` format string as argument. I admit it is similar to `InsertDate`, but a lot simpler.

Repo: https://github.com/fricklerhandwerk/InsertTimestamp
Tags: https://github.com/fricklerhandwerk/InsertTimestamp/tags

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -647,6 +647,16 @@
 			]
 		},
 		{
+			"name": "InsertTimestamp",
+			"details": "https://github.com/fricklerhandwerk/InsertTimestamp",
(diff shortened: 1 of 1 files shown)


FichteFoll:
>I admit it is similar to InsertDate, but a lot simpler.

If it was simpler to use, then I could agree, but InsertDate does pretty much the same thing and more, so I don't see any merit in this package.

PR #6147, opened 2017-02-20
Added Electric Imp Developer package (#2)

ppetrosh:
Requested information:

 - Link to your code repository: https://github.com/electricimp/ElectricImp-Sublime
 - Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/electricimp/ElectricImp-Sublime/tags

All tests from [tests docs](https://packagecontrol.io/docs/submitting_a_package#Step_7) passed on the branch.

@FichteFoll, this pull request addresses your comments/concerns discussed in https://github.com/wbond/package_control_channel/pull/5949

Thank you!

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -384,16 +384,16 @@
 				}
 			]
 		},
-                {
-                        "name": "Electric Imp Developer",
-                        "details": "https://github.com/electricimp/ElectricImp-Sublime",
-                        "releases": [
-                                {
-                                        "sublime_text": ">=3000",
-                                        "tags": true
-                                }
-                        ]
-                },
+		{
+			"name": "Electric Imp Developer",
+			"details": "https://github.com/electricimp/ElectricImp-Sublime",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
 		{
 			"name": "Electron Color Scheme",
 			"details": "https://github.com/kcmr/electron-theme",
--- a/repository/e.json
+++ b/repository/e.json
@@ -351,6 +351,16 @@
 			]
 		},
 		{
+			"name": "Electric Imp Developer",
+			"details": "https://github.com/electricimp/ElectricImp-Sublime",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Elementor",
 			"details": "https://github.com/andresdominguez/elementor-sublime",
 			"releases": [
--- a/repository/e.json
+++ b/repository/e.json
@@ -384,6 +384,16 @@
 				}
 			]
 		},
+                {
+                        "name": "Electric Imp Developer",
+                        "details": "https://github.com/electricimp/ElectricImp-Sublime",
+                        "releases": [
+                                {
+                                        "sublime_text": ">=3000",
+                                        "tags": true
+                                }
+                        ]
+                },
 		{
 			"name": "Electron Color Scheme",
 			"details": "https://github.com/kcmr/electron-theme",
@@ -394,16 +404,6 @@
 					"tags": true
 				}
 			]
-		},
-		{
-                        "name": "Electric Imp Developer",
-                        "details": "https://github.com/electricimp/ElectricImp-Sublime",
-                        "releases": [
-                                {
-                                        "sublime_text": ">=3000",
-                                        "tags": true
-                                }
-                        ]
 		},
 		{
 			"name": "Elementor",


FichteFoll:
## Report for ElectricImp-Sublime #####################

### Repository checks ###

No failures

No warnings

### Package checks ###

Reporting 2 failures:
- The binding ['ctrl+shift+l'] unconditionally overrides a default binding
    File: Default.sublime-keymap
- '.sublime-workspace' files contain session data and should never be submitted to version control
    File: project-template/electric-imp.sublime-workspace

Reporting 3 warnings:
- The binding ['tab'] is also defined in default bindings but is masked with a 'context'
    File: modules/Sublime_AdvancedNewFile_1_0_0/Default (Windows).sublime-keymap
- The binding ['tab'] is also defined in default bindings but is masked with a 'context'
    File: modules/Sublime_AdvancedNewFile_1_0_0/Default (Linux).sublime-keymap
- The binding ['tab'] is also defined in default bindings but is masked with a 'context'
    File: modules/Sublime_AdvancedNewFile_1_0_0/Default (OSX).sublime-keymap

---

As the report indicates, you include the entire AdvancedNewFile package inside your package. This is very bad practive because of potential conflicts. Either suggest users to install this package manually in your readme, or only include relevant parts of it, i.e. its plugin code. I highly suggest the first method, however.

Additionally, `requests` is available as a [Package Control dependency](https://github.com/packagecontrol/requests).

ppetrosh:
@FichteFoll, thank you for the review!

Regarding the failures

1. I don't see the key binding ['ctrl+shift+l'] in the documentation http://docs.sublimetext.info/en/latest/reference/keyboard_shortcuts_osx.html, http://docs.sublimetext.info/en/latest/reference/keyboard_shortcuts_win.html. Where can I find this information?
2. The file project-template/electric-imp.sublime-workspace contains the template workspace for the Sublime projects that are generated by the plugin. I need specific window layout that opens by default. Which I don't have a better idea how to implement in Sublime. Do you have any better suggestions?

As for the warnings, I'm reusing the code of Sublime_AdvancedNewFile 1.0.0, which I don't want to modify. It's key bindings are not used.

Thanks!

FichteFoll:
1. The list you are referencing is not exhaustive. The actual key bindings are defined in *Default ($platform).sublime-keymap` files in the Default package. You can inspect them with the PackageResourceViewer package.
2. That file is only for you to work on and it contains private session information that you are not supposed to share with anyone. It's usually not leaking any important data, the there is a possibility. You can set layouts with the `set_layout` command, which has a couple examples in the default menu and key binding files as well as [on this page](http://docs.sublimetext.info/en/latest/reference/api.html#sublime.Window.set_layout).
3. Sublime Text sees and parses them regardless of whether they are "used" or not. Please either suggest your users to install the package themselves (preferred) or remove all non-plugin files.

ppetrosh:
Hi @FichteFoll,

1. FIXED. That's good to know. Thank you. I wish I could find it in the Sublime documentation somewhere. Anyway this item is fixed, thank you!

2. I have a challenge here. The plugin creates a new Sublime Text 3 project and opens it in the Sublime Text (in the same or in a new window depending on whether this one has anything open already). And in the new window I want to a) apply the layout b) open one file on the left and another one on the right buffer in a two column layout. If I had access to the window object I would be able to do of of that programmatically. So what I was looking for was an API to open a new project that would return a Window object for me, but I couldn't find one. So what I ended up was launching an external process with Sublime binary and the generated *.sublime-project as the argument, which did what I need. I supply this with the *.sublime-workspace file that does what I need with the initial layout and buffers. I find it a very hacky way myself, which I'm not happy with.

3. FIXED: removed the keymap and commands files from the included sources.

Thank you!

FichteFoll:
2) Ah, thanks for the explaination. I didn't investigate closely enough to realize that this was meant to be a template file for your new project. In that case, the workaround is fine.

   However, this opened my eyes for another issue: Since packages are installed in an archived state as `.sublime-package` files, you won't have much luck with [this line](https://github.com/electricimp/ElectricImp-Sublime/blob/81e806c1ca6d9134ae82841e58605317eb39a1b3/imp_developer.py#L996). You have one out of two options (again):

  1. Use `sublime.load_resource` to load the contents of each of the files and write them manually to the target folder from memory. (preferred)
  2. Specify a `.no-sublime-package` file so that Package Control knows to unextract your package manually. This is disadvantegeous because it makes override packages impossible (see more [here](http://docs.sublimetext.info/en/latest/extensibility/packages.html#customizing-or-overriding-packages).

Also, make sure to push a new tag after your changes because otherwise they won't be picked up by Package Control.

ppetrosh:
Got it. Thanks for catching this issue! That's a very valid point, I will take care of what you pointed out.

BTW do you know how to use a colorful text area (self.window.get_output_panel("textarea")) in Sublime? I need to highlight some substrings (errors, warnings) with different colors there. But I didn't find a nice solution for that.

FichteFoll:
The normal procedure is to create a syntax definition for the text to be contained in the output panel and then set that panel's `syntax` setting to your syntax. Colors are sourced from the default color scheme, unless you also set the `color_scheme` setting.

You can provide a `syntax` argument to the `exec` command, which does exactly this (for the exec panel).

ppetrosh:
Wow, thanks for the hint! It's good to know.

ppetrosh:
@FichteFoll I'm working on the update now...

ppetrosh:
@FichteFoll, I fixed loading resources with `sublime.load_resource` as you suggested.

This made me looking at the code that I wrote a half year ago and to be honest, it terrified me a lot ) I'm going to rewrite most parts of it in the next major version (along with moving to the new APIs of the Electric Imp platform).

Anyway I appreciate all your comments and suggestions. They were very useful.

ppetrosh:
@FichteFoll is there a chance that the PR can be merged anytime soon? Our customers are asking for a simpler way to install the plugin through the Package Control. 
Please let me know if anything's missing from our side.
Thanks!

FichteFoll:
Added a small comment to some commit.

PR #6149, opened 2017-02-23
add: Escape HTML - a plugin to escape HTML entities in selected code

vtsvang:

repository: [vtsvang/Sublime-HTMLEscape](https://github.com/vtsvang/Sublime-HTMLEscape)
tag: [0.0.1](https://github.com/vtsvang/Sublime-HTMLEscape/tags)

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -886,6 +886,16 @@
 			]
 		},
 		{
+			"name": "Escape HTML",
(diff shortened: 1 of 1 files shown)


FichteFoll:
https://packagecontrol.io/search/entities

Your key bindings are "valid" but only because ST is lax on its rules. `ctrl+k+e` gets parsed as `ctrl+e`.

PR #6151, opened 2017-02-23
add ActualVim

lunixbochs:

 - Link to your code repository: https://github.com/lunixbochs/actualvim
 - Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/lunixbochs/actualvim/releases/tag/0.9.0

Also make sure you:

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -177,6 +177,17 @@
 			]
 		},
 		{
+			"name": "ActualVim",
+			"details": "https://github.com/lunixbochs/actualvim",
+			"labels": ["input", "vim"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Are you sure that you need `.no-sublime-package`? I never had to deal with compiled Python binaries, so I don't know if they can be imported from within `.sublime-package` or not.

lunixbochs:
Straight from the docs: https://packagecontrol.io/docs/submitting_a_package#Step_5

> If including executables or shared libraries, add a .no-sublime-package file.

FichteFoll:
Well, if @wbond says so, I'm going to believe him. I also know that he said this earlier, but I was interested in whether that is actually the case.

PR #6155, opened 2017-02-24
Add SelectiveUppercase

guokeke:

Also make sure you:

---
This change is [Reviewable](https://reviewable.io/reviews/wbond/package_control_channel/6155)

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -797,18 +797,6 @@
 			]
 		},
 		{
-			"name": "SelectiveUppercase",
-			"details": "https://github.com/guokeke/selective-uppercase",
-			"author": "guokeke",
-			"labels": ["text manipulation"],
-			"releases": [
-				{
-					"sublime_text": "*",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "Selection Evaluator",
 			"details": "https://github.com/lengstrom/MathEvaluator",
 			"labels": ["text manipulation"],
@@ -837,6 +825,18 @@
 				{
 					"sublime_text": "<3000",
 					"branch": "master"
+				}
+			]
+		},
+		{
+			"name": "SelectiveUppercase",
+			"details": "https://github.com/guokeke/selective-uppercase",
+			"author": "guokeke",
+			"labels": ["text manipulation"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
 				}
 			]
 		},
--- a/repository/s.json
+++ b/repository/s.json
@@ -797,6 +797,18 @@
 			]
 		},
 		{
+			"name": "SelectiveUppercase",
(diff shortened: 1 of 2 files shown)


FichteFoll:
How is this different from the "Convert Case: Upper Case" command in the command palette?

guokeke:
you can custom keywords in the keywords.json, and this will only uppercase word in the json file.

I use it to write SQL. 

" select * form blablaba where blaballaba"  ( select all and press alt+u) => "SELECT * FORM blablaba WHERE balbalbal ";

FichteFoll:
Maybe rename it to "SelectiveUppercase" and improve the readme?

I also think many people will be confused with the notion of creating `keywords.json` in an override package. Maybe you can aid them by using ST's setting storage API instead and providing a menu and command palette entry to open the user settings file?

guokeke:
Thank you for your suggestion.

In face i had add a button in ST's Tools menu.
I will rename it to "SelectiveUppercase" and improve my readme.

guokeke:
Hi @FichteFoll 
I add a gif to my readme. 
https://github.com/guokeke/selective-uppercase

math2001:
Hum... Ok... Anyway, you should use settings instead of having a `.no-sublime-package` and allowing the user to update the `keywords.json`. The main advantage is that the user's keywords will be saved into the `User` folder (as every other customization), which makes it much easier to back-up.

Look in the API page for `settings`.

guokeke:
Thank you for your suggestion. @math2001

guokeke:
Hi @math2001 
I've moved `keywords.json` to `User` folder, and remove the `.no-sublime-package`

math2001:
> I've moved keywords.json to User folder, and remove the .no-sublime-package

Sorry, I haven't been clear enough... That's the right way to do it. 

- In your package, create a file called `SelectiveUppercase.sublime-settings`.
- set it's content to something this:

```json
{
    "keywords": []
}
```

- now, from your *plugin*, you can load those settings by doing `sublime.load_settings('SelectiveUppercase.sublime-settings)` (and get the `keywords` by calling `.get('keywords')` upon it)

The advantage is that the `SelectiveUppercase.sublime-settings` in the `User` folder (that isn't created by default) will override this setting if it is specified.

A good way to allow the user to easily create this file in the `User` directory is simply to let him edit the settings using the command `edit_settings`. Here is [an example](https://github.com/math2001/FileManager/blob/0e968de8a26a21a6b7b0822f210f3c8de737966c/.sublime/FileManager.sublime-commands#L77-L84)

If there is anything you don't understand, please let me know, I'll help you out 🙂

guokeke:
@math2001 
Good suggestion! Thanks a lot!
I've created `settings file` in my folder, and removed `keywords.json`.

PR #6165, opened 2017-03-02
Added BackgroundColorEdit repo info

gsboylan:
https://github.com/gsboylan/Sublime-BackgroundColorEdit
https://github.com/gsboylan/Sublime-BackgroundColorEdit/tags

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -76,6 +76,16 @@
 			]
 		},
 		{
+			"name": "BackgroundColorEdit",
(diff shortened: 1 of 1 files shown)


FichteFoll:
You should use the term "color scheme" instead of "theme" in your readme.

You also want to display some sort of error message when the user attempts to edit a color scheme that does *not* exist in raw form on the file system instead of generating a traceback on the console.

gsboylan:
Thanks for the feedback, I'll fix these and push an update.

PR #6172, opened 2017-03-06
Plugin to generate Apex code from JSON sample or schema

nchursin:

 - Link to your code repository: https://github.com/nchursin/json2apex
 - Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/nchursin/json2apex/tags

Also make sure you:

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -948,6 +948,23 @@
 			]
 		},
 		{
+			"name": "JSON2Apex",
+			"details": "https://github.com/nchursin/json2apex",
+			"labels": [
+				"JSON",
+				"Apex",
+				"Salesforce",
+				"code generation",
+				"JSON to Apex"
+			],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "JSONComma",
 			"details": "https://github.com/math2001/JSONComma",
 			"labels": ["JSON", "comma", "text manipulation", "formatting", "formatter", "trailing"],


FichteFoll:
`View.begin_edit` has been removed in ST3. I'm also quite confident that an edit object only works for a certain view and not one that was created in a ViewCommand. You can use the `insert` command instead, like `view.run_command('insert', {"characters": "text be here"})`.

Your entire loader code (which is kind of all over the place and seems to be from [this thread](https://forum.sublimetext.com/t/module-module-loader-reload-modules-dependencies-without-restarting-sublime-text/18781)) will most definitely not work in archived state. Please either remove the code and load modules the traditional way, or add a `.no-sublime-package` file to your repo (usually disouraged because it prevents override packages).

nchursin:
Thanks for the review. Actually, I didn't use that `edit` object after all, so I just removed those lines.
I will make a more traditional loader ASAP and make a new release.

nchursin:
@FichteFoll implemented other loader. Got the code from one of the packages, I hope this is the one you've meant.

FichteFoll:
Yes, that loader codelooks better.

Regarding the second part of the edit object concern, I was referring to this part specifically: https://github.com/nchursin/json2apex/blob/1ec4f44292529ecb2f574c2445d654121aba99bc/json2apex_main.py#L39-L41
If you can confirm that this does indeed work as intended, I'll merge.

nchursin:
@FichteFoll yes, that works as intended. Double checked now it just in case

PR #6173, opened 2017-03-06
adding Python-ContinuousTestRunner to channel

SteveCarratt:
This package is very basic and runs nose2 command on save of a .py file.

https://github.com/SteveCarratt/Sublime3-PythonContinuousTestRunner

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -2599,6 +2599,17 @@
 			]
 		},
 		{
+			"name": "Python-ContinuousTestRunner",
+			"details": "https://github.com/SteveCarratt/Sublime3-PythonContinuousTestRunner",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": ["windows"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Python2UnicodeFixer",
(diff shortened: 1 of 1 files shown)


FichteFoll:
There seem to be [a](https://packagecontrol.io/packages/CommandOnSave) [couple](https://packagecontrol.io/packages/On%20Save) [ways](https://packagecontrol.io/packages/SublimeOnSaveBuild) to do this generically within ST already (not to mention external tools). [Some](https://packagecontrol.io/packages/PythonTestRunner) (flagged for ST2 but should work on 3 too) even for Python. There's also one for pytest specifically, but that doesn't count here.

Considering how inflexible this package is, I'd normally suggest to just use a (project-specific) build system and either build periodically or use the last package I referenced, so I'm not too sure what to think about it. If you think this packages provides functionality that warrants it being available for installation alongside all the others I mentioned, I'll merge.

SteveCarratt:
Thanks, I'll have a look at the packages you said and have a think about it.

SteveCarratt:
I think [this](https://packagecontrol.io/packages/On%20Save) is the best way to do this. I didn't think the other packages were as easy to use. Thanks for pointing me in the right direction.

PR #6174, opened 2017-03-07
add some labels to existing plugin

unknownuser88:
add some labels to existing plugin (Console Wrap for js)


FichteFoll:
When renaming, you need to add the old name to a `"previous_names"` array.

PR #6175, opened 2017-03-09
create JSON Faster Plugin

CourageSuo:

 - Link to your code repository: https://github.com/CourageSuo/INTASublime.git

Also make sure you:

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -902,6 +902,16 @@
 			]
 		},
 		{
+			"name": "JSON Faster",
+			"details": "https://github.com/CourageSuo/INTASublime.git",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "JSON L18n Resources Editor",
 			"details": "https://github.com/alex18881/JsonL18nResourcesEditor",
 			"releases": [
--- a/repository/j.json
+++ b/repository/j.json
@@ -903,7 +903,7 @@
 		},
 		{
 			"name": "JSON Faster",
-			"details": "https://github.com/CourageSuo/INTASublime.git",
+			"details": "https://github.com/CourageSuo/JSONFaster.git",
 			"releases": [
 				{
 					"sublime_text": "*",


math2001:
You need to type your readme in english.

Also, your command need to be more "specific". The convention is to have this kind of name `MyCommandNameCommand` (PascalCase)

CourageSuo:
change the readme in English and Specify the command

math2001:
I'm probably not a reference at English, but we say `faster` and not `more faster`

### Command Names

Again, you need to change your commands name: `ExampleCommand` doesn't make any sense, `tabCommand` could be used by another plugin (so, you could get overridden, or override an other's plugin's command),

What you can do is prefix your command with `JSONFaster`, or do an internal system of command: you create only one command which takes an argument `action` for example, and this command runs the appropriate function.

### Repo Name

Why isn't your repo's name `JSONFaster`?

### `Default.sublime-keymap`

If all your platform specific `.sublime-keymap` are the same, you can just create a file called `Default.sublime-keymap` and will work on every platform.

CourageSuo:
Thanks your patient and detailed explanation and so shamed about my poor English😁😁😁
Now I rename my command name, as you suggest I add prefix ahead of my command.  of course I combined three .sublime-keymap to one Default.sublime-keymap.
The Repo name INTASbulime which is my website www.intalesson.com theme. Is that affect the pull request?
thanks

math2001:
> The Repo name INTASbulime which is my website www.intalesson.com theme. Is that affect the pull request?

No, it doesn't. But it'll be hard(er) to find your package's repo (and there is no reason to put your websites *in* the repo's name. You can mention it in your repo description, if you want to). I don't recommend it though.

CourageSuo:
so I change the repo name, but it still one failing check?

math2001:
First, you don't need the `.git` at the end: `https://github.com/CourageSuo/JSONFaster`.

Second, the `details` value is the URL to the *repo*. So, if you rename it here, you have to rename your repo   accordingly (which, again I recommend).

FichteFoll:
As @math2001 mentioned, you must remove the `.git` in your details url.

Please note that you don't have any tags pushed to your repo (they must be [semantic versions](http://semver.org) and that the `alt+-` binding is already bound by default on some platform, causing your to override it. What you want to do instead is define the key bindings in a way that they are only active on JSON files. See also the [docs](http://docs.sublimetext.info/en/latest/reference/key_bindings.html).

PR #6184, opened 2017-03-11
Added DotClean

kjellberg:

Also make sure you:

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -1062,6 +1062,18 @@
 			]
 		},
 		{
+			"name": "DotClean (dot_clean on save)",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Do not use `time.sleep`, not even in async callbacks because async callbacks share the same thread. Use a callback/lambda and pass that to `sublime.set_timeout{,_async}`. You might also want to use `subprocess.Popen` in that case, since that does not wait for the subprocess to finish.

Please also add a readme that at least explains what the package does.
Also, I'm quite sure this doesn't make a lot of sense to use on Linux.

PR #6192, opened 2017-03-14
json faster for sublime text plugin

CourageSuo:

 - Link to your code repository: https://github.com/CourageSuo/JSONFaster.git

Also make sure you:

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -902,6 +902,16 @@
 			]
 		},
 		{
+			"name": "JSON Faster",
+			"details": "https://github.com/CourageSuo/JSONFaster",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "JSON L18n Resources Editor",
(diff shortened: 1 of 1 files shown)


math2001:
You didn't need to create a new PR, but anyway.

Your repo's much better. But I don't understand why you have a command named `JsonFasterExample`... Why don't you rename it to actually match what it's doing?

Also, another little tip: when you want to display a keystroke, you can wrap with a `kbd` tag: `<kbd>alt+-</kbd>` gives <kbd>alt+-</kbd>. It's HTML, and it works in with GitHub Flavoured Markdown (so it's displayed properly on GitHub).

Once you'll have correct this naming problem, your package will be ready to go!

FichteFoll:
Quoting myself from #6175:

> Please note that you don't have any tags pushed to your repo (they must be [semantic versions](http://semver.org) and that the `alt+-` binding is already bound by default on some platform, causing your to override it. What you want to do instead is define the key bindings in a way that they are only active on JSON files. See also the [docs](http://docs.sublimetext.info/en/latest/reference/key_bindings.html).

PR #6195, opened 2017-03-16
Add Simple JSX package

ccampbell:
This is a simple JSX package that adds JSX syntax highlighting on top of the Sublime Text built in JavaScript grammars. 

Relevant discussion in sublimehq/Packages#133

Repo: https://github.com/ccampbell/sublime-jsx
Tags: https://github.com/ccampbell/sublime-jsx/tags

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1550,6 +1550,17 @@
 			]
 		},
 		{
+			"name": "Simple JSX",
+			"details": "https://github.com/ccampbell/sublime-jsx",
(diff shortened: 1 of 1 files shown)


FichteFoll:
See also https://github.com/sublimehq/Packages/issues/860, which should be interesting to you. Because that will most likely take a while until it is done and accepted (and a new (beta) build gets released with it), I'll merge this regardless since it's a band-aid until official JSX support is added.

PR #6196, opened 2017-03-17
add Six package

guillermooo:
Please add the Six package.

- https://github.com/guillermooo/Six
- https://github.com/guillermooo/Six/releases

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1505,6 +1505,22 @@
 			]
 		},
 		{
+			"name": "Six",
+			"description": "Advanced Vim emulation.",
+			"author": "Guillermo López-Anglada",
+			"homepage": "http://www.sublimesix.com",
+			"buy": "http://www.sublimesix.com/buy.html",
+			"details": "https://github.com/guillermooo/six",
+			"readme": "https://raw.githubusercontent.com/guillermooo/Six/master/README.md",
+			"labels": ["vim", "vi", "vintage", "vintageous", "emulation", "emulator", "editor emulation"],
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
You should add a description and a url to your github repo, which will then carry over to PC's metadata. This allows you to remove the explicit keys here and dynamically change their values yourself without the need of a PR.

Since you offer a paid package, you'll probably want to make sure your site is served over https to protect the users' privacy. I can't see how exactly your payments are processed (via paypal). Currently, your site/hoster offers a certificate only for *.webfaction.com.

Also, your homepage's code page seems to be messed up. Since the file itself already carries `?` question marks, it probably happened somewhere on the server storage side.

guillermooo:
> You should add a description and a url to your github repo, which will then carry over to PC's metadata. This allows you to remove the explicit keys here and dynamically change their values yourself without the need of a PR.

GH hosting is only a temporary solution until I get a CDN set up, but I don't want to wait on that before releasing. That's why I'm avoiding tying the project to GH in the first place.

> Since you offer a paid package, you'll probably want to make sure your site is served over https to protect the users' privacy. I can't see how exactly your payments are processed (via paypal). Currently, your site/hoster offers a certificate only for *.webfaction.com.

Definitely! It's on my todo list, but not for release. I'll apply for a Letsencrypt cert soon.

> Also, your homepage's code page seems to be messed up. Since the file itself already carries ? question marks, it probably happened somewhere on the server storage side.

Also on my todo list. I fixed encoding for other pages, but the Buy page slipped through.

Thanks!

PR #6202, opened 2017-03-23
Add Window Mode

rwols:
- Link to code repository: https://github.com/rwols/WindowMode
- Link to the tags pag: https://github.com/rwols/WindowMode/releases

Add "Window Mode" plugin to the channel

### DIFF
--- a/repository/w.json
+++ b/repository/w.json
@@ -423,6 +423,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
As a tip, you could also set `on_clone = on_load = on_activated = on_new` inside the class scope. No guarantees.

PR #6203, opened 2017-03-24
AddLibrary plugin to easily download libs

caiofsouza:

 - Link to your code repository:  https://github.com/caiofsouza/AddLibrary
 - Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/caiofsouza/AddLibrary/releases/tag/v1.0.0

Also make sure you:

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -252,6 +252,18 @@
 			]
 		},
 		{
+			"name": "AddLibrary",
+			"author": "Caio Fernandes",
+			"details": "https://github.com/caiofsouza/AddLibrary",
+			"labels": ["javascript", "css", "library", "dev"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name" : "ADOdb library snippets for PHP",
 			"details": "https://github.com/madhs/ADOdb_snippets",
 			"labels": ["snippets", "php", "adodb", "database"],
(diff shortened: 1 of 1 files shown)


FichteFoll:
On ST3, use `json.loads(sublime.load_resource("Packages/AddLibrary/Libraries.json"))` instead of directly reading the file, since that will fail when the package is installed in archived state, which is the default. This method doesn't exist in ST2, so you either have to add a fallback or remove ST2-support.

You don't import `Request` in the ST2 import code.

`from .Libraries import Libraries` as a relative import will only function in ST3. For ST2, you need to use `from Libraries import Libraries` for root-level modules (yes, this is not ideal).

My suggestion is to remove ST2 support, since only about 95% of ST users are using it.

math2001:
> My suggestion is to remove ST2 support, since only about 95% of ST users are using it.

You probably meant the opposite no? 95% are using st**3**

Is there any reason not to use `sublime.decode_value` instead `json.loads`? The first one accepts trailing commas...

FichteFoll:
@math2001 thanks for spotting that.

`sublime.decode_value` can parse trailing commas and C-style comments, so it is better feature-wise. It is however not available in ST2 and I never benchmarked it on how it compares to `json.loads`.

caiofsouza:
Hello @FichteFoll , 
Thanks for the suggestions. This is my first plugin and I'm learning how to build that and the methods of sublime API. And, of course, the `sublime.load_resource` seems me better. About support to ST2, I'll change the support to only ST3 :D

caiofsouza:
@math2001 I did not use `sublime.decode_value` because I did not known this method, nothing special...

PR #6204, opened 2017-03-24
Requesting addition of Simple Word Count plugin.

BrettHansen:

Also make sure you:

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1594,6 +1594,16 @@
 			]
 		},
 		{
+			"name": "Simple Word Count",
+			"details": "https://github.com/BrettHansen/SimpleWordCount",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "master"
+				}
+			]
+		},
+		{
 			"name": "SimpleClone",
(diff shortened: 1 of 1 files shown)


FichteFoll:
In Python, namespacing should be done through modules and not through classes (re your `Helper` class). Classes are only needed when you actually use OOP paradigms. In your particular case, you can just remove the class entirely and store the functions in the module directly.

Could you explain what the differences to https://packagecontrol.io/packages/WordCount are?

BrettHansen:
I didn't come across that plugin before creating my own. I did search for one though, and it appears that [titoBouzout's plugin](https://packagecontrol.io/packages/WordCount) may not have been searchable until recently since installs spiked on March 30th and have maintained that level.

My plugin is not significantly different so I'm closing the request.

FichteFoll:
Yes, the plugin was uninstallable for a while because the author pulled all his packages from the list a while ago. Someone else re-added his (unmodified) fork of it a while ago. 

You would probably have found it if I had reviewed it faster, but that didn't happen. I apologize if I was wasting your time, but hopefully you had a learning experience with this.

PR #6206, opened 2017-03-24
Add package: Prevent Duplicate Windows

aristidesfl:


FichteFoll:
https://github.com/packagecontrol/package_reviewer/wiki/Repository-checks#no-semantic-version-tags-found

Your package design is not compatible with the context of using multiple workspaces for the same project. See [the docs on workspaces](http://docs.sublimetext.info/en/latest/file_management/projects.html#workspaces) for what they do.

Unfortunately I'm not aware of a way to determine the currently used `.sublime-workspace` file, so you might want to restrict this functionality only to windows that do not have a project file (and can subsequently have multiple workspaces) associated with them. You can check out the source of https://github.com/rwols/WindowMode, which I just reviewed. Alternatively, just mention this limitation in the readme.

aristidesfl:
The semantic tag was fixed to 0.1.1.

Do you mean that multiple windows can be opened with the same project but different work spaces? Aren't people using the `Quick Switch Project` panel to switch between workspaces in the same window?

aristidesfl:
@FichteFoll need to know if I understood the problem correctly.

FichteFoll:
>Do you mean that multiple windows can be opened with the same project but different work spaces? Aren't people using the Quick Switch Project panel to switch between workspaces in the same window?

Yes and yes. You can have multiple windows with different workspaces for the same project open at a time. "Quick Switch Project" will list workspace files that you opened in ST explicitly. Unfortunately, it does not indicate any of this in its API as far as I can tell. `window.extract_variables()` certainly doesn't have any `sublime-workspace` mention.

aristidesfl:
I will add a note about the caveat brought up by @FichteFoll in the weekend. Can also have a look if there is a way around the problem.

aristidesfl:
Didn't find a way to distinguish between workspaces. 
Added a to readme: https://github.com/aristidesfl/sublime-prevent-duplicate-windows#caveats

Should be good to merge?

PR #6207, opened 2017-03-25
SublimeHaskell: drop ST2 support

mvoidex:
Also switching to tags release strategy


FichteFoll:
FYI, your repo description still reads "A Sublime Text 2/3".

mvoidex:
> FYI, your repo description still reads "A Sublime Text 2/3".

Thanks, fixed

PR #6211, opened 2017-03-28
Add csso

1000ch:
Add csso plugin for Sublime.

- https://github.com/1000ch/Sublime-csso
- https://github.com/1000ch/Sublime-csso/releases/tag/v1.0.0

sublime-csso

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3642,6 +3642,16 @@
 			]
 		},
 		{
+			"name": "CSSO",
+			"details": "https://github.com/1000ch/Sublime-csso",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
`env['PATH'] += os.path.expanduser('~/n/bin')` this seems a little arbitrary, but shouldn't do anything bad.

However, please remove the monkeypatching of `sublime.Region` ([here](https://github.com/1000ch/Sublime-csso/blob/270fba1ce8258e7ac2a67f61c8d9c1acbbadfab5/Csso.py#L12-L13)). You probably copied this from the source origin and aren't even using it. Let me tell you that monkeypatching is in many situations *not* the right thing to do and if you really *do* need to do it, you will know.

1000ch:
@FichteFoll Fixed them, thank you for checking.

PR #6220, opened 2017-04-01
Add AmnesiaIO Package

jolyongray:

 - Link to your code repository: https://github.com/jolyongray/amnesia-sublime
 - Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/jolyongray/amnesia-sublime/releases

Also make sure you:

 1. Used `"tags": true` and not `"branch": "master"` ([versioning docs](https://packagecontrol.io/docs/submitting_a_package#Step_4)) ✓
 2. Ran the tests ([tests docs](https://packagecontrol.io/docs/submitting_a_package#Step_7)) ✓

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -601,6 +601,16 @@
 			]
 		},
 		{
+			"name": "AmnesiaIO",
+			"details": "https://github.com/jolyongray/amnesia-sublime",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
`sublime.Region.empty` returns whether a region is empty or not (regarding your `getSelections`). You could also just check whether `trimmedSelections` is empty; that is, if you don't directly use it on the filter result. Example: `trimmedSelections = [sel for sel in selections if not sel.empty()]; if not trimmedSelections: return`

PR #6221, opened 2017-04-01
extract lines plugin

poma:

 - Link to your code repository: https://github.com/poma/sublime-extractlines
 - Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/poma/sublime-extractlines/tags

This is my first experience with Python and Sublime API so it could contain bugs. But I tested it and it looks fine.

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -1413,6 +1413,16 @@
 			]
 		},
 		{
+			"name": "Extract Lines",
+			"details": "https://github.com/poma/sublime-extractlines",
+			"releases": [
+				{
+					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


FichteFoll:
A few suggestions:

1. Extracting the entire view's contents and then running `str.splitlines` on it should be faster than extracting the regions returned from `View.lines`. Newlines are always normalized by ST internally, so you'll always want to use just `"\n"` (addressing your comment in `get_all_lines`).
2. You can use set comprehensions in `find_duplicates`, since you convert it to a set in line 30 anyway. Just use curly braces instead of brackets.

PR #6226, opened 2017-04-02
Add ClickableRequires package

hajnalben:

 - Link to your code repository: https://github.com/hajnalben/ClickableRequires
 - Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/hajnalben/ClickableRequires/tags

Also make sure you:

 1. Used `"tags": true` and not `"branch": "master"` ([versioning docs](https://packagecontrol.io/docs/submitting_a_package#Step_4)) :+1:
 2. Ran the tests ([tests docs](https://packagecontrol.io/docs/submitting_a_package#Step_7)) :+1:

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1236,6 +1236,17 @@
 			]
 		},
 		{
+			"name": "ClickableRequires",
+			"details": "https://github.com/hajnalben/ClickableRequires",
+			"labels": ["require", "node", "javascript", "js"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"details": "https://github.com/trentrichardson/Clientside",
 			"releases": [
 				{


FichteFoll:
## Report for ClickableRequires #######################

### Repository checks ###

No failures

No warnings

### Package checks ###

No failures

Reporting 3 warnings:
- It is not advised to specify mousemaps because of how limited the configuration options are. Instead, suggest the user to add specific bindings to their User package.
    File: Default (Linux).sublime-mousemap
- It is not advised to specify mousemaps because of how limited the configuration options are. Instead, suggest the user to add specific bindings to their User package.
    File: Default (OSX).sublime-mousemap
- It is not advised to specify mousemaps because of how limited the configuration options are. Instead, suggest the user to add specific bindings to their User package.
    File: Default (Windows).sublime-mousemap

---

Notably, ctrl+leftclick (Linux, Windows) is used to create and add a caret to the selection and thus is already used in the default bindings.

If you used the regular expression `'require\s*\(\s*[\'"].+?[\'"]\s*\)' you could match more require calls since that is just a function and can have spaces pretty much everywhere. You'll then need to extract the actual module name differently, e.g. with `module = re.match(r"""['"](.+?)['"]""", statement).group(1)`.

You should also use `os.path.join` instead of hard-coding forward slashes. Your current code will choke on Windows because of your use of `.split('/')` at some point. Use `os.path.dirname` there instead.

Windows also doesn't like leaking file handles. To properly close them, wrap the `open` call in a `with` block.

hajnalben:
Thanks for the response! I will fix them soon. Can you suggest any proper solution for for key combination? The IDEs often use this ctrl/cmd + left click mapping.

FichteFoll:
>Instead, suggest the user to add specific bindings to their User package.

You can provide an example configuration in your readme with ctrl+left-click, but the users will have to add these bindings to their setup themselves. This will help them realize why the default functionality does't work anymore and allows them to adjust the bindings should they already be using this modifier for a different command or should they prefer a different binding.

I'd probably go for something unused in my setup, like ctrl+alt.

FichteFoll:
Alternatively, you could also use the `on_hover` hook to display a popup with a link that users can click to trigger the desired action. That way they can even know beforehand whether they are taken to the file on their system or a website with documentation is opened and it's almost guaranteed to be collision-free.

hajnalben:
I have created a new tag with the changes you suggested:
- removed mousemaps (documented in the configuration in readme)
- file `open` now wrapped with `with`
- implemented popup on hover with in-app and browser links
- improved regex
- removed hardcoded path concatenation

PR #6245, opened 2017-04-10
added AutoFoldCode and MatchSelection

callodacity:
- Repo: https://github.com/callodacity/AutoFoldCode
 - Tags: https://github.com/callodacity/AutoFoldCode/tags

This package makes Sublime Text remember folded code regions when files are closed. There is a similarly named plugin (https://packagecontrol.io/packages/Auto%20Fold) but that serves a different purpose to this one (this saves folded regions, that one automatically folds them based on a pre-defined option).

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -1914,6 +1914,17 @@
 			]
 		},
 		{
+			"name": "AutoFoldCode",
+			"details": "https://github.com/callodacity/AutoFoldCode",
+			"labels": ["foldcode", "fold", "region", "persistent"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "autogit",
 			"details": "https://github.com/chrisparnin/autogit-st",
 			"releases": [
--- a/repository/m.json
+++ b/repository/m.json
@@ -604,17 +604,6 @@
 			]
 		},
 		{
-			"name": "Match Selection",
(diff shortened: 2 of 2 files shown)


callodacity:
Also pushed another package: `MatchSelection`.

It extends Sublime Text's default `match_selection` options.

* Repo: https://github.com/callodacity/MatchSelection
* Tags: https://github.com/callodacity/MatchSelected/releases/tag/1.0.0

FichteFoll:
`Region.totuple` is not an official API and was monkey-patched by some other package (there is a certain "runs code in node.js" package family that does this). You must not rely on this, and I highly discourage adding the patching code to your package as well exactly because of this.

"MatchSelection" seems very similar to https://packagecontrol.io/packages/WordHighlight. Did you know about this package? Can you point out some notable differences?

Also, in the future, please create a new PR for each package you intend to add.

callodacity:
@FichteFoll thanks for pointing that out - I've removed `Region.totuple` in the `AutoFoldCode` package.

I didn't know about https://packagecontrol.io/packages/WordHighlight - there are some minor differences, but I'll try a PR on that package. (better than having two similar packages).

> Also, in the future, please create a new PR for each package you intend to add.

Okay, will do 👍

FichteFoll:
You could also just use a simple list comprehension:

```py
regions_to_save = [(r.a, r.b) for r in view.folded_regions()]
```

PR #6248, opened 2017-04-10
Update a.json

alienhard:
Added label to existing package

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -490,6 +490,7 @@
 		{
 			"name": "All Autocomplete",
 			"details": "https://github.com/alienhard/SublimeAllAutocomplete",
+			"labels": ["auto-complete"],
 			"releases": [
 				{
 					"sublime_text": "*",


FichteFoll:
By convention, the package doesn't actually provide auto-completions because those would be context-aware (and smart) and only offer relevant completions such as function names or available identifiers/variables.

The package has been named like this for quite a while, but I try to keep at least the "auto-complete" tag clean.

alienhard:
@FichteFoll – no problem if you prefer not to tag it as auto-complete. To me this sounded like the right tag, since all this package does is to extend Sublime's native auto complete feature [1]. I guess that most people who find Sublime's completion behavior limiting would look for a package tagged auto-complete... But in any case, it appears they still find the package even without the tag, as it has been quite popular for a long time.

[1] https://www.sublimetext.com/docs/3/auto_complete.html

FichteFoll:
Interesting. I forgot that this page existed, which quite clearly declares the general completion popup menu as the "auto-complete" popup. That's quite unfortunate because, to me, the two concepts I mentioned before are similar in functionality, yet fundamentally different in their quality, and having a label to distinguish them would be quite useful. That said, I wonder how useful labels are in general.

I'll leave this up to @wbond, as I'm -0 on this.

wbond:
I'm not particularly opinionated about this, so I'm going to merge. The labels can be useful for exploring, but I imagine most people just search.

PR #6253, opened 2017-04-14
Updated MultiLang Color Schemes Repo Location

ProjectCleverWeb:
Package was renamed forever ago but repo URL was never updated in package control.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -2136,7 +2136,7 @@
 		},
 		{
 			"name": "MultiLang Color Scheme",
-			"details": "https://github.com/ProjectCleverWeb/MultiLang",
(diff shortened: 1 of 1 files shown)


ProjectCleverWeb:
Yes, but younger and less experienced me didn't know about that. So back then I just copied the files and created a new repo from them, and started modifying that. Which means a the correct process for renaming won't work any more.

gerardroche:
Cool, add a link to the new repository from the old repository and it's good.

ProjectCleverWeb:
Done. ([see the old one's Github page header](https://github.com/ProjectCleverWeb/MultiLang))

ProjectCleverWeb:
Awesome, does that mean this pull request is ready to be merged? I have been waiting a while for this.

PR #6256, opened 2017-04-15
re-added Invisible Tabs Spaces Switcher with changes

isar0se:

 - Link to your code repository: https://github.com/isar0se/InvisibleTabsSpacesSwitcher
 - Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/isar0se/InvisibleTabsSpacesSwitcher/tags

Also make sure you:

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -788,6 +788,17 @@
 			]
 		},
 		{
+			"name": "Invisible Tabs Spaces Switcher",
+			"details": "https://github.com/isar0se/InvisibleTabsSpacesSwitcher",
+			"labels": ["formatting", "text manipulation"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Io Language",
 			"details": "https://github.com/yangsu/sublime-io",
 			"labels": ["language syntax"],


isar0se:
notes from last pull request:

> This package violates a few conventions and has numerous error-prone situations because the conversion happens both on loading and on closing files instead of when saving them. I highly suggest hooking both the on_pre_save and on_post_save hooks and doing replacements there (if desired) instead of only when closing.
> The close event is not fired when an entire project is closed (afaik) and definitely not when ST is closed entirely. Furthermore, the file changing contents significantly upon closing the view as opposed to when saving it just moments before will not match user expectations.

the whole point of the plugin is to hook into the on_load & on_pre_close hooks instead of the save hooks, so that the conversion is invisible to the user. it's true i could convert on_pre_save & convert back on_post_save—i tried that in a previous iteration—but then every time you save a file, your file shows as having unsaved changes (because of the conversion which happens immediately after you save), which is a more troubling & obvious mismatch with user expectations imo...

the close event actually _is_ fired when a project is closed, as well as when ST is closed entirely—i've tested the plugin in both those situations & it works fine

curious about what conventions the plugin violates—this is my first ST plugin so i'm definitely interested in picking up best practices

thanks!

PR #6257, opened 2017-04-15
Add Theme - Monokai Pro

Monokai:

 - Link to your code repository: **I'm using manual hosting, final package at https://packages.monokai.pro/theme-monokai-pro/1.0.0.zip**
 - Link to the tags page with at least one [semver](http://semver.org) tag: **I'm using manual hosting**

Also make sure you:

**I'm using manual hosting, so don't use either**

**Check, test passed OK.**


wbond:
It seems that the zip in your repository contains a single file - a `.sublime-package` file. Instead of zipping that package file, reference it directly from the repository JSON.

Monokai:
@wbond Thanks. I've updated `package.json` and the referenced `sublime-package` file on the server.

PR #6265, opened 2017-04-22
Add NeoVintageous

gerardroche:
- GitHub: https://github.com/NeoVintageous/NeoVintageous
- Tags: v1.0.0 - https://github.com/NeoVintageous/NeoVintageous/releases/tag/1.0.0
- NeoVintageous Organisation Members: https://github.com/orgs/NeoVintageous/people

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -213,6 +213,18 @@
 			]
 		},
 		{
+			"name": "NeoVintageous",
+			"description": "Vim emulation, a Vintageous fork.",
+			"details": "https://github.com/NeoVintageous/NeoVintageous",
+			"labels": ["vim", "vi", "vintage", "vintageous", "neovim", "nvim", "emulation", "emulator", "editor emulation"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "NESASM",
 			"details": "https://github.com/klaussilveira/SublimeNESASM",
 			"releases": [


gerardroche:
Sorry I don't really understand the package reviewer failures and warnings.

There are essentially 2 failures:

A syntax issue:

   - Set "sublime_text": ">=3092" in your "releases" since the package contains a .sublime-syntax file
       File: res/Command-line mode.sublime-syntax

I don't think that should be a failure because the package provides an old style syntax as fallback for older versions.

Invalid keys:

    - Invalid key 'A'
       File: Default.sublime-keymap
       Binding: {"args": {"key": "A"}, "command": "press_key", "context": [{"key": "vi_command_mode_aware"}], "keys": ["A"]}
   
Is this because the plugin os overriding the A key? The package is a fork of Vintageous which is a vim emulation plugin, it overrides many of the keys to provide vim emulation.
 
There's also a warning:

   - The binding ['ctrl+f5'] is also defined in default bindings but is masked with a 'context'
       File: Default.sublime-keymap

The same as above unless the keymaps I'm missing something.

wbond:
The reason it didn't pick up the fallback is because it was named `.hidden-tmLanguage` instead of `.tmLanguage`. @FichteFoll can provide more info about the keybinding issues.

gerardroche:
> The reason it didn't pick up the fallback is because it was named .hidden-tmLanguage instead of .tmLanguage.

Is that a bug in the package reviewer? Adding `hidden-` to the file in older versions is how you make the syntax hidden. In newer versions the hidden flag is indicated in the file itself `hidden: true`.

gerardroche:
@FichteFoll Any chance I get some feedback on what these keymap issue are. I'm assuming they're just notices that a package is overriding core key bindings.

wbond:
I am guessing the invalid key issue is because they are uppercase? If you wanted it to be uppercase, it should be shift+lowercase, right @FichteFoll?

wbond:
(Just testing the enhanced PR automated review system)

FichteFoll:
Aparently, upper-case characters are only invalid with modifiers, not if they are the sole character of a chord, in which case they represent any actual symbol produced by the keyboard.

Fixed in packagecontrol/st_package_reviewer@b1be13c6f3174051bfb1c16563c7cd6a497cf68e.

PR #6299, opened 2017-05-12
add package: CopyAsHtml

liulex:

 - Link to your code repository: https://github.com/liulex/CopyAsHtml
 - Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/liulex/CopyAsHtml/tags

Also make sure you:

- [x] Used `"tags": true` and not `"branch": "master"` ([versioning docs](https://packagecontrol.io/docs/submitting_a_package#Step_4))
- [x] Ran the tests ([tests docs](https://packagecontrol.io/docs/submitting_a_package#Step_7))

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2904,6 +2904,18 @@
 			]
 		},
 		{
+			"name": "Copy as HTML",
+			"details": "https://github.com/liulex/CopyAsHtml",
+			"labels": ["copy", "html", "clipboard"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"platforms": ["windows"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Copy Block",
 			"details": "https://github.com/xsleonard/sublime-CopyBlock",
 			"releases": [


facelessuser:
Why didn't you just offer a pull request to ExportHtml to add the functionality you needed?  The license allows you to do whatever you want, but in the interest of the community, it makes more sense to consolidate efforts.

liulex:
@facelessuser 
Hi, Isaac, I really appreciate your work and I did consider sending you a pull request to your plugin ExportHtml instead of creating a new one. Here are the reasons why I changed my mind:
1. The html document generated by ExportHtml contains js scripts and the toolbar, which is unnecessary for clipboard purpose. And ExportHtml hides the gutter by setting it to `display: none`, which may not be recognized by some applications, e.g., Qt's QTextBrowser. QTextBrowser will always show the toolbar and the gutter. To make it work, I need to remove the code of toolbar and gutter completely from the html document, which would require huge modifications to your current code structure...
2. I have always wanted this "Copy as html" feature, which is supported by most code editors. I searched the web for a long time and eventually found ExportHtml and SublimeHighlight. SublimeHighlight cannot use the current color scheme. ExportHtml is nearly what I want, but it needs to open the html in browser before I can copy the format into the clipboard. (The option "clipboard_copy" only copies plain html text to the clipboard.) I think it reasonbale to have this feature be an independent plugin if it's not natively supported by Sublime.

facelessuser:
I think a true optional gutter wouldn't be too big a deal. I also don't think excluding the toolbar when copying formatted HTML is a big deal either.

Anyways, I was just curious, I'm not against people forking it. At some point I will clean up ExportHtml. It may also gain this functionality as well.  It's always been something on my list to do.

PR #6304, opened 2017-05-15
Added Element Finder (ST3)

thinkv:

 - Link to your code repository: https://github.com/thinkv/sublime3-elfinder
 - Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/thinkv/sublime3-elfinder/tags

Notes:
Please note that this is an update and continuation of the following package (https://github.com/keeganstreet/sublime-elfinder) as I will be taking over from ST3 and onwards.

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -396,6 +396,16 @@
 			]
 		},
 		{
+			"name": "Element Finder (ST3)",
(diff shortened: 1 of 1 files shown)


thinkv:
Going to modify Element Finder and take over instead.

PR #6308, opened 2017-05-16
Transferring ownership of the Element Finder package to thinkv

keeganstreet:
@thinkv has updated the package for Sublime Text 3.


thinkv:
Might need to move it up in the order @keeganstreet 
`Repositories must be sorted alphabetically`

keeganstreet:
Thanks @thinkv, fixed alphabetisation now.

thinkv:
Hi @wbond,

My packages.json has a hybrid of URL releases and then the main channel with tagged releases, is that not possible?

If not, I have made the change to point to the version number instead.

wbond:
The current run of the tests is:

```
2 Warnings
 - Creating a readme for your package will help users understand what it does and how to use it
```

You should probably add a `"readme"` entry, or use `"details"` instead of `"homepage"`, in which case it will pick up the existing readme and issues from the GitHub repo.

thinkv:
Thanks @wbond, have made the changes now.

PR #6320, opened 2017-05-22
Added Ionic 2 Package

Bernardoow:

 - https://github.com/Bernardoow/Ionic2Package
 - https://github.com/Bernardoow/Ionic2Package/tags 

This package is for new Ionic 2. New html tags and I will adding auto-completion and others things.

@FichteFoll 
Maybe don't remember, but I learned with you. Heheh For This package I created a branch to do pr hehehe.

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -810,6 +810,17 @@
 			]
 		},
 		{
+			"name": "Ionic 2 Package",
+			"details": "https://github.com/Bernardoow/Ionic2Package",
+			"labels": ["snippets",  "auto-complete", "completion", "markup", "ionic"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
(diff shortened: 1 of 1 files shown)


repository/i.json
```diff
@@ -810,6 +810,17 @@
 			]
 		},
 		{
+			"name": "Ionic 2 Package",
```
wbond: This should be changed to something like `Ionic 2 Snippets`. The word `Package` should definitely not be there.

Bernardoow: Hi @wbond ,

I will improve this package with more things. Like completions. Because this I put this name with ionic 2 package.

wbond: You'll at least need to remove the word `Package`.

Bernardoow:
@wbond I changed name.

PR #6324, opened 2017-05-24
Add Better Less

radium-v:

 - Link to your code repository: https://github.com/radium-v/Better-Less
 - Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/radium-v/Better-Less/tags

Also make sure you:

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -494,6 +494,17 @@
 			]
 		},
 		{
+			"name": "Better Less",
+			"details": "https://github.com/radium-v/better-less",
+			"labels": ["less", "language syntax", "css"],
+			"releases": [
+				{
+					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


FichteFoll:
How is this different from https://packagecontrol.io/packages/LESS and why couldn't you contribute to that package instead?

radium-v:
The LESS package hasn't been updated in three years, and is based on the original [Textmate bundle](https://github.com/creationix/LESS.tmbundle) which itself is 5 years behind.

Another package, [LessImproved](https://github.com/braver/LessImproved), was not in active development when I started this project, but has since been getting some new commits in the last month. Regardless, since it uses the `sublime-syntax` format, it's only compatible with Sublime Text 3.

PR #6326, opened 2017-05-26
Add vue formate plugin

baixuexiyang:
sublime plugin at https://github.com/baixuexiyang/VUEFormatter

tag v1.0.2 listed at https://github.com/baixuexiyang/VUEFormatter/tags

used "tags": true

run tests and everything ok


wbond:
Please verify that you require a `.no-sublime-package` file, since it prevents users from utilizing package overrides.

PR #6330, opened 2017-05-28
Added the Amxx Pawn syntax repository/a.json

evandrocoan:

 - Link to your code repository: https://github.com/evandrocoan/SublimeAmxxPawn
 - Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/evandrocoan/SublimeAmxxPawn/releases

Ran the tests ([tests docs](https://packagecontrol.io/docs/submitting_a_package#Step_7))
```
Running Default Channel Tests

  .............................................
  Ran 7634 tests in 2.469s
  
  OK
```

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -665,6 +665,17 @@
 			]
 		},
 		{
+			"name": "Amxx Pawn",
+			"details": "https://github.com/evandrocoan/SublimeAmxxPawn",
+			"labels": ["build system", "language syntax", "amx mod x", "amxx pawn"],
+			"releases": [
+				{
+					"sublime_text": ">=3114",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Anaconda",
 			"details": "https://github.com/DamnWidget/anaconda",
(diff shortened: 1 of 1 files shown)


repository/a.json
```diff
@@ -665,6 +665,17 @@
 			]
 		},
 		{
+			"name": "Amxx Pawn",
+			"details": "https://github.com/evandrocoan/SublimeAmxxPawn",
+			"labels": ["language syntax", "amx mod x", "amxx pawn"],
```
wbond: It would probably make sense to remove the last two labels, especially the one that repeats the package name.

evandrocoan: My bad, it seems pretty redundant. I removed it on 2833e30

But the amxmodx I think it is worth because when you look for `amxmodx` on the search you find absolutely nothing about it. Then after adding this package, you would find a syntax for its language:

1. https://packagecontrol.io/search/amx%20mod%20x

Amxmodx is a virtual machine which executes code written in pawn with the amx mod x API:

1. https://www.amxmodx.org/api/

PR #6351, opened 2017-06-06
Added jsonschema Python package to dependencies.json

kylebebak:
From this package: <https://github.com/Julian/jsonschema>

 - <https://github.com/kylebebak/sublime-jsonschema>
 - <https://github.com/kylebebak/sublime-jsonschema/releases>

This is the first time I've tried adding a dependency. I think everything is ready. I didn't know what to make of the `load_order` property dependencies, so I used 55, which at this time is equal to the largest value of `load_order`.

### DIFF
--- a/repository/dependencies.json
+++ b/repository/dependencies.json
@@ -221,6 +221,20 @@
 			]
 		},
 		{
+			"name": "jsonschema",
+			"load_order": "55",
+			"description": "An(other) implementation of JSON Schema for Python",
+			"author": "kylebebak",
+			"issues": "https://github.com/kylebebak/sublime-jsonschema/issues",
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
The jsonschema readme mentions support for Python 2.7+, however ST2 uses 2.6. I believe this won't work on ST2, so you should probably restrict installation to ST3.

FichteFoll:
If your dependency does not rely on other dependencies loading first, `load_order` is of no significance for you. In fact there are so few situations where they would matter, I believe it should be made optional and default to 50 instead.

kylebebak:
@FichteFoll 

Thank you for clarifying, I'll restrict the dependency to ST3 only. I'll leave the `load_order` property alone, given that in the case of `jsonschema` it has no effect.

PR #6293, opened 2017-05-08
add Term package

percevalw:
Plugin for emulating a terminal inside a tab

 - Link to your code repository: https://github.com/percevalw/SublimeTerm
 - Link to the tags page: https://github.com/percevalw/SublimeTerm/tags

Used "tags": true => Done
Ran the tests => Done

PR #6359, opened 2017-06-09
Add TerminalView package

Wramberg:

https://github.com/Wramberg/TerminalView

https://github.com/Wramberg/TerminalView/releases

Also make sure you:

Check!
Check

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -435,6 +435,18 @@
 			]
 		},
 		{
+			"details": "https://github.com/Wramberg/TerminalView",
+			"labels": ["terminal", "console", "shell", "bash"],
+			"name": "TerminalView",
+			"releases": [
+				{
+					"platforms": ["osx", "linux"],
+					"sublime_text": ">=3092",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "termX",
 			"previous_names": ["MacTerminal"],
 			"homepage": "http://malinowski.be/termX",


Wramberg:
Note that it shadows a lot of ST3 keybindings on purpose (only when in the TerminalView of course).

There are a few plugins that look similar to this one but none of them uses a pseudo-terminal (that I know of) so this plugin offers a different approach that provide additional features,

FichteFoll:
The [Term](https://github.com/percevalw/SublimeTerm) package (PR pending https://github.com/wbond/package_control_channel/pull/6293) also emulates a terminal. Yours already seems to have color support (and I generally think the keybinding overrides are a good idea).

This is kind of an odd situation because we usually try to keep packages doing the same thing at a minimum, but you were both working on your packages at the same time and both of them are also in the queue right now. Ideally we'd only have one packages, but I can see that you both put a lot of effort into it. Do you have any suggestions about this?

Also pinging @percecalw.

FichteFoll:
(pinging @percevalw again because I made a typo earlier and had to edit it)

percevalw:
Indeed, you did a pretty good job with this package. It looks more reactive and uses pyte to communicate with the shell, which is a good idea I think. Moreover, I won't have time to take care of mine in the near future, so go ahead !

FichteFoll:
Thanks for understading and taking a look, @percevalw. I'll close your PR then. You'll probably want to enhance @Wramberg's package in the future, in case you find a mising feature 😉

rwols:
@percevalw I think your runloop implementation is better than TerminalView's, so consider turning that idea into a pull request for TerminalView.

Wramberg:
Hi gents,

After looking into the term package I see that they are indeed quite similar. I second @rwols observation that SublimeTerm probably has a better approach in terms of the run loop since it blocks while waiting for input thus using less CPU.

Moreover it seems that more research has gone into the terminfo definitions etc. in the SublimeTerm package so thats probably something I would like to look into to improve TerminalView compatibility with other shells (than bash).

I appreciate your decision to step down @percevalw - feel free to test out the TerminalView plugin and start some discussions in the gitrepo if you have any ideas. PRs are welcome too of course but some times they conflict with some of my development ideas so it's probably best to discuss first.

PR #6360, opened 2017-06-10
Add Breadcrumbs

qbolec:
https://github.com/qbolec/Breadcrumbs
https://github.com/qbolec/Breadcrumbs/tags
I havent added  `"branch": "master"` as I'm using semver on master.

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -1193,6 +1193,16 @@
 			]
 		},
 		{
+			"name": "Breadcrumbs",
+			"details": "https://github.com/qbolec/Breadcrumbs",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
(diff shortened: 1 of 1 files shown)


AlexKvazos:
screen shot 2017-06-10 at 5 10 59 pm

The package is giving the following errors in Sublime Text 3134. Please address them so we can merge your package.

qbolec:
Do I understand correctly that I have to write my plugin in a common subset of Python 2 and Python 3, to make it work on both Sublime Text 2 and Sublime Text 3?

qbolec:
I've released bugfix, tagged 1.0.1, which seems to work on both Sublime Text 2, and Sublime Text 3 portable.

AlexKvazos:
Works great, thanks!

PR #6361, opened 2017-06-11
Add NotmuchAddress Command.

yeled:
Import NotmuchAddress insertion plugin:

https://github.com/yeled/sublime-notmuch-address
https://github.com/yeled/sublime-notmuch-address/releases/tag/1.0.0

This is if you are using ST3 as an external editor for the `notmuch` email system

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -698,7 +698,6 @@
 			"releases": [
 				{
 					"sublime_text": ">3000",
-					"base": "https://github.com/yeled/sublime-notmuch-address",
 					"platforms": ["osx"],
 					"tags": true
 				}
--- a/repository/n.json
+++ b/repository/n.json
@@ -693,6 +693,18 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


AlexKvazos:
@yeled Can you please use a valid semver tag? X.X.X

See http://semver.org/

yeled:
Sorry, I was wondering why.. I expected the PATCH to be implied

This is done now.

repository/n.json
```diff
@@ -693,6 +693,18 @@
 			]
 		},
 		{
+			"name": "NotmuchAddress",
+			"details": "https://github.com/yeled/sublime-notmuch-address",
+			"releases": [
+				{
+					"sublime_text": ">3000",
+					"base": "https://github.com/yeled/sublime-notmuch-address",
```
wbond: This line can be removed since it is inferred from the `details` URL.

yeled: done!

PR #6364, opened 2017-06-13
Added YAML Macros package.

Thom1729:
Repository: https://github.com/Thom1729/YAML-Macros/tags
Tags: https://github.com/Thom1729/YAML-Macros/tags

### DIFF
--- a/repository/y.json
+++ b/repository/y.json
@@ -12,6 +12,16 @@
 			]
 		},
 		{
+			"name": "YAML Macros",
+			"details": "https://github.com/Thom1729/YAML-Macros",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "YAML Nav",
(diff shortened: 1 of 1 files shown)


Thom1729:
The package should be compatible with older versions. The `sublime-syntax` is an example file; it doesn't need to be loaded by Sublime. The only thing that older versions can't do is run the tests for that example file.

wbond:
Just a heads up @Thom1729, if you release it with the existing git tags, no one will even see it unless they opt in to prereleases for this package name.

Thom1729:
It's still undergoing rapid changes today. When I release "v1.0.0", it should be generally available, right?

wbond:
Yes, any tag with an optional leading `v` following by three dotted numbers will be available to all users.

FichteFoll:
You rely on the `pyyaml` dependency but don't have that specified in a `dependencies.json` file.

Thom1729:
I've fixed the dependencies.

PR #6366, opened 2017-06-15
repository: add UbuntuPaste for sublime-text-3

3v1n0:

 - Link to your code repository: https://github.com/3v1n0/UbuntuPaste
 - Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/3v1n0/UbuntuPaste/tree/3.1.1

Also make sure you:

 1. Used `"tags": true` and not `"branch": "master"` ([versioning docs](https://packagecontrol.io/docs/submitting_a_package#Step_4)): yes
 2. Ran the tests ([tests docs](https://packagecontrol.io/docs/submitting_a_package#Step_7)): done

```
Ran 7630 tests in 0.282s
  
  OK
```

This is a fork of a package that is already available for Sublime-2 (https://github.com/frankban/UbuntuPaste/), I've made a [PR against it](https://github.com/frankban/UbuntuPaste/pull/1) (long time ago) and asked the author to review it multiple times, but he has no time for it. So, I guess we can safely use this version for sublime-3 while keeping the other for sublime-2.

### DIFF
--- a/repository/u.json
+++ b/repository/u.json
@@ -26,7 +26,7 @@
 			"releases": [
 				{
 					"sublime_text": "<3000",
-					"tags": "2."
+					"tags": "st2-"
 				},
 				{
 					"sublime_text": ">=3000",
--- a/repository/u.json
+++ b/repository/u.json
@@ -22,17 +22,12 @@
 			]
 		},
 		{
-			"details": "https://github.com/frankban/UbuntuPaste",
+			"details": "https://github.com/3v1n0/UbuntuPaste",
 			"releases": [
 				{
 					"sublime_text": "<3000",
-					"branch": "master"
-				}
-			]
-		},
-		{
-			"details": "https://github.com/3v1n0/UbuntuPaste",
-			"releases": [
+					"tags": "2."
+				},
 				{
 					"sublime_text": ">=3000",
 					"tags": true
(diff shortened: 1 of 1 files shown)


AlexKvazos:
We can bring @frankban into the conversation so he can have a say.

repository/u.json
```diff
@@ -22,11 +22,15 @@
 			]
 		},
 		{
-			"details": "https://github.com/frankban/UbuntuPaste",
+			"details": "https://github.com/3v1n0/UbuntuPaste",
 			"releases": [
 				{
 					"sublime_text": "<3000",
-					"branch": "master"
+					"tags": "2."
```
FichteFoll: This tag setup won't work because after stripping the prefix specified here (`2.`), the remaining tag name is not a valid semver anymore (only two parts).

3v1n0: Mh, so how to do that?
I read this in the docs...

wbond: You need to add a prefix. Most users who utilize this feature have a prefix of something like `st2-`. Then their tags would be `st2-2.0.0`, `st2-2.0.1`, etc. Users will see version `2.0.0` installed, as the prefix is only used to figured out which tags are relevant for the channel.

FichteFoll:
@3v1n0 ping

3v1n0:
@FichteFoll ok i've updated the tags to be st{2,3}-version and the file accordingly.

PR #6369, opened 2017-06-16
Add BetterFindNext Package

notbaab:
[Repo](https://github.com/notbaab/BetterFindNext)
[Tag](https://github.com/notbaab/BetterFindNext/releases/tag/1.0.0)

Also make sure you:

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -527,6 +527,16 @@
 			]
 		},
 		{
+			"name": "BetterFindNext",
+			"details": "https://github.com/notbaab/BetterFindNext",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "BetterSnippetManager",
 			"details": "https://github.com/math2001/BetterSnippetManager",
 			"labels": ["snippets", "utilities", "file open", "code navigation", "utils", "file creation"],


notbaab:
Rebound to unused binding with [this](https://github.com/notbaab/BetterFindNext/releases/tag/v1.0.1) release.

Actually ran it against st_package_reviewer this time.

notbaab:
Does it need to pass with no warnings? 

Also, I'm trying to provide similar functionality to the find_next (ctrl+d) command, so having escape clear the added regions makes sense, like so. 

add_next

AlexKvazos:
Yeah, you can't override default keybindings as people don't find that cool. Try with a third modifier key or two-step (prefix) key.

AlexKvazos:
Actually, now that I see it is "escape", I think it is no problem, let me pass this to someone else though.

cc @wbond @keith-hall

keith-hall:
I suspect it is fine, as it has a specific context - user feedback will dictate whether closing autocomplete popups etc. should take precedence over the clearing find regions or not :)

AlexKvazos:
Alright then!

r-stein:
Maybe worth to say, that you can just give it a lower precedence by taking the context in the default keybindings and inverting them:

```js
    {
        "keys": ["escape"], "command": "clear_better_find_selection", "context":
        [
            { "key": "has_next_field", "operator": "not_equal", "operand": true },
            { "key": "has_prev_field", "operator": "not_equal", "operand": true },
            { "key": "panel_visible", "operator": "not_equal", "operand": true },
            { "key": "overlay_visible", "operator": "not_equal", "operand": true },
            { "key": "popup_visible", "operator": "not_equal", "operand": true },
            { "key": "auto_complete_visible", "operator": "not_equal", "operand": true },
            { "key": "has_region", "operand": "find_region" }
        ]
    },
```

notbaab:
That is a great suggestion, done! 

This might be covered in the documentation in which case I apologize but does the auto tester test every release? Or intermittently to make sure developers haven't violated the rules?

r-stein:
The auto reviewer is only to help the reviewers before merging the package to package control and won't be executed afterwards.
However you can manually download and execute it: https://pypi.python.org/pypi/st-package-reviewer

notbaab:
Awesome, thanks!

PR #6376, opened 2017-06-21
Add "FoldFunctions" package

math2001:
- Repo: https://github.com/math2001/FoldFunctions
- Tags: https://github.com/math2001/FoldFunctions/releases
- [x] used the tag
- [x] ran the tests

A simple package to fold functions in different languages. For now, only Python and JavaScript are supported.

Note: this package uses the `view.indentation_level` method, which is (still) not documented. Should I create a kind of polyfill for this, or it's likely to stay?

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -1174,6 +1174,16 @@
 			]
 		},
 		{
+			"name": "Fold Functions",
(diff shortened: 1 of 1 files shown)


AlexKvazos:
Thank you for your contribution!

PR #6388, opened 2017-06-23
Added Yandex.Weather plugin

bolknote:
https://github.com/bolknote/PogodaStatusBar
https://github.com/bolknote/PogodaStatusBar/tags

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -1516,6 +1516,18 @@
 			]
 		},
 		{
+			"name": "PogodaStatusBar",
+			"details": "https://github.com/bolknote/PogodaStatusBar",
+			"labels": ["status bar", "weather", "info", "utilities"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": ["osx"],
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


keith-hall:
It looks like you want to exclude the readme and screenshot from the `.sublime-package` file, judging by https://github.com/bolknote/PogodaStatusBar/commit/9823b70dbd24732a0b3ad15687cd339d02d2ae22.

This is not currently possible - everything in the repo will be included in the package - [if a `.no-sublime-package` file is present, it causes the whole package to be unpacked as opposed to stored in a zip file, which makes it impossible for users to have custom overrides.](https://packagecontrol.io/docs/submitting_a_package#Step_5) Indeed, the file isn't read by Package Control, it is normally empty.

Also, most users actually want the readme to be part of the package, so they can open it without going to the package's website/repo. Therefore, please consider removing the `.no-sublime-package` file.

FichteFoll:
There actually is a way to exclude files from zip-/tarballs, but that works with `.gitattributes` and I don't really recommend it. Either way, the current usage of `.no-sublime-package` is wrong and harmful.

bolknote:
I've fixed it.

PR #6393, opened 2017-06-27
renamed "Blade Spacer" to "Laravel Blade Spacer" for better visibility

austenc:
I think a lot of people might miss this as they search for "Laravel" and the package doesn't have that in its name. Thanks!

 - Code repository: https://github.com/austenc/blade-spacer
 - Tags page: https://github.com/austenc/blade-spacer/tags

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -768,16 +768,6 @@
 			]
 		},
 		{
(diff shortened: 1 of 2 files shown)


austenc:
Thanks @AlexKvazos !

PR #6398, opened 2017-06-28
Add package: `Polymer Syntax`

jaychsu:
package hosting: https://github.com/jaychsu/sublime-polymer-syntax

This package is about syntax highlighting for [Polymer](https://www.polymer-project.org/) and [Web Component](https://developer.mozilla.org/en-US/docs/Web/Web_Components) in Sublime-Text.

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -1575,7 +1575,7 @@
 			"labels": ["syntax", "polymer", "web component"],
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": ">=3092",
 					"tags": true
 				}
 			]
--- a/repository/p.json
+++ b/repository/p.json
@@ -1570,6 +1570,17 @@
 			]
 		},
 		{
+			"name": "Polymer Syntax",
+			"details": "https://github.com/jaychsu/sublime-polymer-syntax",
+			"labels": ["syntax", "polymer", "web component"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Pomodoro",
 			"details": "https://github.com/Neway6655/Sublime-Pomodoro",
 			"labels": ["timer"],


jaychsu:
I've added `polymer.tmLanguage` in [version 1.0.1](https://github.com/jaychsu/sublime-polymer-syntax/commit/2470e84dd5c97efe6ce2e30ea8cc2ac9f668ed40).

FichteFoll:
`.tmLanguage` and `.sublime-syntax` are not compatible formats. `.sublime-syntax` can be converted *to* from `.tmLanguage` because it is a superset, but it is structured differently.

You should remove the invalid fallback file and instead restrict installation to builds >=3092.

I'm also unsure about whether sublime-syntax works with string keys in captures mappings.

jaychsu:
> You should remove the invalid fallback file and instead restrict installation to builds >=3092.

OK, my following commit will do this.

> I'm also unsure about whether sublime-syntax works with string keys in captures mappings.

These captures work on my local(build 3126, macOS). Should I pass any test?

FichteFoll:
If they work for you locally, then that is fine. I never saw or used them, which is why I got curious.

ChannelRepositoryTools does *not* test your package's contents - that's what [st_package_reviewer](https://github.com/packagecontrol/st_package_reviewer) does.

jaychsu:
Thanks a lot! I'll try to use this tool.

jaychsu:
@FichteFoll 
I want to check that: have you seen [polymer.tmLanguage](https://github.com/jaychsu/sublime-polymer-syntax/blob/master/polymer.tmLanguage)? 

I knew `.tmLanguage` and `.sublime-syntax` are not compatible formats. So I use [PackageDev](https://github.com/SublimeText/PackageDev) to convert [polymer.sublime-syntax](https://github.com/jaychsu/sublime-polymer-syntax/blob/master/polymer.sublime-syntax) to [polymer.tmLanguage](https://github.com/jaychsu/sublime-polymer-syntax/blob/master/polymer.tmLanguage).

FichteFoll:
Yes, I have seen it. You can trust me that I know enough about syntax definitions to see that this `tmLanguage` file is invalid. Try removing the sublime-syntax file (or change its extension) and then open a file with your syntax.

PackageDev only does a plain conversion of Property List data into YAML or JSON (and vice versa) but it does translate into different formats. If you'd look at a proper tmLanguage file, such as [this one](https://github.com/FichteFoll/CSScheme/blob/6575b53d2c40a64839f86e624af33bf86f6b8e34/Package/CSScheme.tmLanguage), you'll see that it is vastly different in structure. 

There is a command to translate a tmLanguage to sublime-syntax, provided by ST itself, but you can't translate a sublime-syntax to tmLanguage, because the former is more powerful.

jaychsu:
Thank you @FichteFoll. You're right. I've removed the invalid `polymer.tmLanguage` and restrict installation to builds >=3092.

jaychsu:
I'll add it back when the valid one is done :)

FichteFoll:
You don't really need to worry about supporting builds older than 3092. Statistics from Package Control have shown that only 5% or less users are still on version below that, most of which are still on ST2.

jaychsu:
OK, thx. lol

PR #6399, opened 2017-06-30
Update r.json with React-CSS-Converter

davidfufu:
This is a small plugin for Sublime Text 3 that converts CSS notation to and from its normal syntax to the DOM notation that is accessed by React. This special notation is also called Dom notation. I made this to speed up React development due to the fact that CSS style objects can be created and stored within React components, but rewriting the CSS into React friendly, DOM notation was a hassle. This is important because it would be very tedious to convert regular css to react ready CSS.There currently does not exist a plugin to do so.  MDN link about the differences in notation: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Properties_Reference

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -492,6 +492,17 @@
 			]
 		},
 		{
+			"name": "React-CSS-Converter",
+			"details": "https://github.com/davidfufu/React-CSS-Notation-Converter",
+			"labels": ["css", "react", "conversion", "notation", "styling"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "react-native-snippets",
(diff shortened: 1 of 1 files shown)


FichteFoll:
You should take a look at [string formatting in Python](https://docs.python.org/3/library/stdtypes.html?highlight=str.format#str.format).

davidfufu:
Thanks, will take a look.

davidfufu:
When I have made improvements, do I make another pull request?

FichteFoll:
No, just push a new tag to your repository and it will get picked up after a while (up to 1 hour).

davidfufu:
So, 1. I do a new commit with changes and 2. make a new semantic versioning
tag like 1.1.0 and it'll be fine? Thanks.

FichteFoll:
Exactly.

PR #6400, opened 2017-06-30
Add UrlConverter package.

gh640:
I'd like to add a package which provides a function to convert urls into html/markdown links.

In general, make sure you:

 1. Used `"tags": true` and not `"branch": "master"` 
    ([versioning docs](https://packagecontrol.io/docs/submitting_a_package#Step_4)) ... Yes.
 2. Added a readme to your repository so that users (and reviewers) 
    can understand what your package provides. ... Added.
 
You may proceed with a short description of what the package does and, 
in case a similar package already exists, 
why you believe it is needed
below this line. -->

This package provides a command to convert selected urls into html or markdown links. I made this because I couldn't find any package with this function.

### DIFF
--- a/repository/u.json
+++ b/repository/u.json
@@ -454,6 +454,16 @@
 			]
 		},
 		{
+			"name": "UrlConverter",
+			"details": "https://github.com/gh640/SublimeUrlConverter",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"details": "https://github.com/mastahyeti/URLEncode",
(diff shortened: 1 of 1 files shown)


gh640:
The repository page is `https://github.com/gh640/SublimeUrlConverter`. Please tell me if did this request wrongly. Thank you in advance.

FichteFoll:
I probably would have named in something along the lines of "Linkificator", but UrlConverter works too.

gh640:
Thank you for handling the request :)

As with the name, I see! they look better. I'll try changing the name later. Thanks.

PR #6401, opened 2017-06-30
Vtex-tags package

Rodrigonavarro23:
This package is an completion feature for Vtex platform tags, there is a vtex snnipet but this is up to date.

### DIFF
--- a/repository/v.json
+++ b/repository/v.json
@@ -608,7 +608,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"branch": "master"
+					"tags": true
 				}
 			]
 		},
--- a/repository/v.json
+++ b/repository/v.json
@@ -603,6 +603,16 @@
 			]
 		},
 		{
+			"name": "Vtex Tags",
+			"details": "https://github.com/Rodrigonavarro23/vtex-tags",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "master"
+				}
+			]
+		},
+		{
 			"name": "Vue Syntax Highlight",
 			"details": "https://github.com/vuejs/vue-syntax-highlight",
 			"releases": [


FichteFoll:
It doesn't look like your plugin code does anything that a `.sublime-completions` file couldn't do. See [here])(http://docs.sublimetext.info/en/latest/reference/completions.html) for docs.

There's also https://github.com/ricardodantas/sublime-vtex-snippets, but that seems to provide less "completion" and they are actually snippets, which behave slightly differently.

Rodrigonavarro23:
Hi, thank you.

What you recommend me? Chase it by a completion file?

Cheers.

FichteFoll:
Completion files have less overhead than plugins, so yes, I suggest providing these completions in a completions file. You can just generate the file from the completions you already have in Python code (using `json.dump`) and pick an appropriate selector. Probably along the lines of `source.js.jsx | text.html`. I don't know what kind of file the `.hbs` extension indicates.

Rodrigonavarro23:
Ok, thanks. Today I will change to that. Has files are Handlebars files, I should remove it.

FichteFoll:
Via the [Handlebars package](https://github.com/daaain/Handlebars/blob/fa350d393dc04cf67e238258be154625f7388bc6/grammars/Handlebars.tmLanguage#L1319), the scope would be `text.html.handlebars`, which is already matched by `text.html`, so we don't need to add it to the selector.

Rodrigonavarro23:
Thank you, I will change it today.

Rodrigonavarro23:
Now it is fixed, but never get open completions with open tag, just ctrl+space.

Cheers.

FichteFoll:
>but never get open completions with open tag, just ctrl+space.

Can you elaborate on that?

Also, remember that you'll need to push a new tag with the latest changes.

Rodrigonavarro23:
Vete tags repo it is on the last version now, the 1.0.3.

Cheers.

PR #6405, opened 2017-07-01
Added ruamel dependency.

Thom1729:
The more well-known PyYaml library does not roundtrip gracefully. As a result, the output of my YAML Macros package is quite ugly and hard to debug. Using ruamel.yaml, the input and output look almost identical except for the results of the macros themselves.

### DIFF
--- a/repository/dependencies.json
+++ b/repository/dependencies.json
@@ -607,6 +607,20 @@
 			]
 		},
 		{
+			"name": "ruamel",
+			"load_order": "01",
+			"description": "Python ruamel.yaml module",
+			"author": "Thomas Smith",
+			"issues": "https://github.com/Thom1729/sublime-ruamel/issues",
+			"releases": [
+				{
+					"base": "https://github.com/Thom1729/sublime-ruamel",
+					"tags": true,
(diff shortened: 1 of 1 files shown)


FichteFoll:
I only noticed after merging, but there is a reason why this dependency is called `ruamel.yaml` and not `ruamel`; compare https://pypi.python.org/pypi?:action=search&term=ruamel&submit=search. I suggest adding `yaml` somewhere in the dependency name, possibly just using the same name as on ppyi.

Thom1729:
Good catch — I wasn't aware that there was anything else under the `ruamel` namespace.

I had thought that the dependency name should not have a period in it. If I was mistaken about this, then the best name would be `ruamel.yaml`. Otherwise, perhaps `ruamel-yaml`.

FichteFoll:
Dots shouldn't matter for dependencies, because the directory name itself isn't relevant for Python packaging. Only the `st3` will be added to `sys.path` and there are no plugins for ST being loaded. Either way, `ruamel-yaml` would be the safer option and I suggest we just use that. It also aligns with the current dependency naming conventions (that was never specified but generally seems to work well).

PR #6406, opened 2017-07-02
Add FlowType package.

Pegase745:

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -1099,6 +1099,17 @@
 			]
 		},
 		{
+			"name": "FlowType",
+			"details": "https://github.com/Pegase745/sublime-flowtype",
+			"labels": ["formatting", "auto-complete", "linting", "coverage", "flow", "javascript"],
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


Pegase745:
Anyone for this error please ? I changed the binding, but the error still appears cc @wbond @FichteFoll

wbond:
`super` and `command` are the same thing. I don't think it is necessary to correct that warning, it is just telling you that in some situations your key binding won't work since the default key bindings include the same key binding, but in a more specific situation where a context is added.

Also, your PR seems to add the package twice.

Pegase745:
@wbond thanks, it seems fixed now. can it be merged ?

FichteFoll:
FYI, your error message in `plugin_loaded` will never be seen on ST2 because it will already fail on the imports. Not considering that `plugin_loaded` isn't called on ST2 to begin with.

Another aside, a "warning" is simply a prompt for potential action, i.e. something a human reviewer should look at.

Pegase745:
Thanks for the info on ST2.

PR #6408, opened 2017-07-03
Add package: SublimeLinter-contrib-yaml-lint

witsec:
I created a YAML linter which uses the yaml-lint Ruby gem.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -3940,6 +3940,16 @@
 			]
 		},
 		{
+			"name": "SublimeLinter-contrib-yaml-lint",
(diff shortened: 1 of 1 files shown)


witsec:
New packages aren't allowed to use branch based releases, so will change this to tags.

FichteFoll:
Please follow the steps for SublimeLinter plugins outlined here: http://www.sublimelinter.com/en/latest/creating_a_linter.html#preparing-for-publication

witsec:
I see... Really didn't know that, so will look into that.

PR #6414, opened 2017-07-04
Rename “YAML Macros” to “YAMLMacros”

Thom1729:
To make it easier for users to reference libraries inside the package.


FichteFoll:
You need to rearrange the package data,t oo.

PR #6419, opened 2017-07-05
Add quick google anno by meloalright

meloalright:

### DIFF
--- a/repository/q.json
+++ b/repository/q.json
@@ -174,6 +174,16 @@
 			]
 		},
 		{
+			"name": "Quick Google Anno",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Seems a little over the top to use a plugin for a simple task like this, especially when the only dynamic part about it are the comment puncutation marks 

Though I have to say that DoxyDoxygen in general does this a lot better by automatically parsing the function signature and supporting more languages and more styles: http://20tauri.free.fr/DoxyDoxygen/v2/page_overview.php
I'm not sure how I feel about this. I mean, it is lighter, but it's technically also just a snippet that anyone could create.

PR #6422, opened 2017-07-07
add simple_ConTeXt

equiva1ence:

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1664,6 +1664,16 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
Please refrain from adding your package's path to `sys.path` (especially doing it once per plugin). Instead, use relative imports like `from .scripts import common`.

equiva1ence:
@FichteFoll I've gone ahead and made those changes, and made a new release with them in it.

PR #6424, opened 2017-07-08
Add plugin to only go back to the last edit position and locate the cursor 

diqye2014:
https://github.com/Wuage-FE/GotoLast/blob/master/README.md

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -1464,7 +1464,7 @@
 			"labels": ["only last modified", "file navigation"],
 			"releases": [
 				{
-					"sublime_text": "<3000",
+					"sublime_text": ">3000",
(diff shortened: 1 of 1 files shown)


diqye2014:
The version already exists at https://github.com/Wuage-FE/GotoLast/tags

FichteFoll:
This is quite similar to the built-in "Jump back" functionality, available in the *Goto* menu, except that also records regions where you didn't make any changes, but is more robust in handling actual or multiple selections.

diqye2014:
Thank you for letting me know the built-in feature, which is very nice. one thing is that after I've browsed a lot of code, it's not sure where it's going back.

PR #6428, opened 2017-07-09
add react native autocomplete

gebeto:
This plugin add react native autocompletes, I wrote parsers that parse react native official documentation and create .sublime-completion files from that data, now it have small amount of completions, and have some defect in parser of StyleSheet but it small, now styles autocomletes parse very good and autocomplete looks great:)

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -499,6 +499,7 @@
 				{
 					"version": 0.46,
 					"sublime_text": "*",
+					"tags": true
 				}
 			]
 		},
--- a/repository/r.json
+++ b/repository/r.json
@@ -492,7 +492,7 @@
 			]
 		},
 		{
-			"name": "React Native Completions",
+			"name": "React Native Autocomplete",
 			"details": "https://github.com/gebeto/react-native-autocomplete",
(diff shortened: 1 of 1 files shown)


repository/r.json
```diff
@@ -499,6 +499,7 @@
 				{
 					"version": 0.46,
```
FichteFoll: You must remove this line

FichteFoll:
You have to push a [semantic version](http://semver.org) tag to your repo. Note that semvers consist of exactly three numeric parts.

You should also move your python files to build the completions to a subdirectory, because currently ST would try to load them as plugins, which will very likely fail as they're not designed for that (external dependencies, directly accessing files in the package dir, uses ST2 syntax).

FichteFoll:
Please fix the package order: https://travis-ci.org/wbond/package_control_channel/builds/253703343#L220

After that I can merge. The @packagecontrol-bot comment is irrelevant here; maybe we should make that a warning instead of an error.

gebeto:
Thanks for the hint! I really thought for a long time what for the mistake

PR #6429, opened 2017-07-09
Add python 2to3 converter

jsn5:
Made a plugin to convert python code from version 2 to 3 on the go.

### DIFF
--- a/repository/0-9.json
+++ b/repository/0-9.json
@@ -35,6 +35,17 @@
 			]
 		},
 		{
+			"name": "2to3",
+			"details": "https://github.com/jsn5/2to3_sublime_plugin",
+			"labels": ["python", "2to3"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "3024 Color Scheme",
 			"details": "https://github.com/idleberg/3024.tmTheme",
 			"labels": ["color scheme"],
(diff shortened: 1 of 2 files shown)


AlexKvazos:
Can you please add a single package per PR?

jsn5:
@AlexKvazos my bad, 2nd one got added by mistake, changed it.

FichteFoll:
You shouldn't wait for the process to finish if you don't actually need to do anything with it. It will just make the UI unresponsive.

Besides, your method of running `2to3.py` will fail, because ST stores packages in archived `.sublime-package` files, which aren't accessible by default. It also depends on `python` being installed on the system and on PATH, which is not the default for Windows.

An easier way would be to just import `lib2to3` directly and call `lib2to3.main.main` directly. It would require ST3 to be used, but basically nobody uses ST2 anyway.

jsn5:
@FichteFoll  I tried importing `lib2to3.main` and call `main` directly, but gives the error `NotADirectoryError: [Errno 20] Not a directory: '/opt/sublime_text/python3.3.zip/lib2to3/Grammar.txt'`, I am using ST3.

FichteFoll:
Oh, I see. That's unfortunate but not really something that we can fix.

Anyway, you can also execute 2to3 from the command line, if Python is installed on the system, using `python -m lib2to3`, so you don't need your bundled script (which has a shebang line).

PR #6430, opened 2017-07-09
Add the Rune.js package to channel

mphstudios:
A package of code completions and snippets for the Rune.js library.


mphstudios:
Invalid control character, \escape characters, and missing delimeter fixed in package repository.

PR #6434, opened 2017-07-10
Adding RamdaDocs package

bijoythomas:
Pull ramda library function documentation from within Sublime Text. You can either enter a Ramda function or highlight a function in the current buffer to pull its documentation

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -365,6 +365,18 @@
 			]
 		},
 		{
+			"name": "JavaScript Scratch Pad",
+			"author": "Bijoy Thomas",
+			"details": "https://github.com/bijoythomas/sublime-jsscratchpad",
+			"labels": ["javascript", "scratch pad"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "JavaScript Snippets",
 			"details": "https://github.com/jprichardson/sublime-js-snippets",
 			"labels": ["snippets"],
--- a/repository/r.json
+++ b/repository/r.json
@@ -279,7 +279,6 @@
 			"releases": [
 				{
 					"sublime_text": ">=3000",
-					"platforms": ["osx", "linux", "windows"],
 					"tags": true
 				}
 			]
--- a/repository/r.json
+++ b/repository/r.json
@@ -279,7 +279,7 @@
 			"releases": [
 				{
 					"sublime_text": ">=3000",
-					"platforms": ["osx", "linux"],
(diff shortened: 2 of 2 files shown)


rwols:
@bijoythomas You can add a context to the keybinding so that it only triggers when the base scope is `source.js`, something like this:

    { "keys": ["alt+r"], "command": "ramda_docs", "context":
      [
        { "key": "selector", "operator": "equal", "operand": "source.js", "match_all": true }
      ]
    }
I'm not sure if `"operator": "equal"` is the right thing to use, you'll have to play around with it. [Docs about keybindings and contexts here](http://docs.sublimetext.info/en/latest/reference/key_bindings.html).

FichteFoll:
Please open a new PR for each package in the future.

`alt+r` is used to toggle the regex operation mode in the find panel (Linux and Windows). Masking your binding with a context as suggested by @rwols will prevent one binding from completely overriding the other, as they are only active when their context matches.

bijoythomas:
@rwols thank you for your suggestion. It looks like it helped reduced the error to a warning now and based on the review wiki docs, I assume someone will have to manually eyeball this.

@FichteFoll My apologies for including both packages in the same PR. In the future, I will refrain from doing so. 

I have a question on triggering the PR builds when only changes are made to the package repo. For example, when I change something only in a package and say tag the package repo, do I always have to make some change in the package control repository file as well to trigger a PR build? In order to trigger the PR builds, I just made some cosmetic edits in the r.json file. Is this the usual workflow?

rwols:
You can just `pip install st_package_reviewer` too 👍 https://pypi.python.org/pypi/st-package-reviewer

wbond:
You can paste the JSON for your package at https://packagecontrol.io/test_repo to get the exact response that packagecontrol-bot will post here.

bijoythomas:
Thanks! I guess it would be a good idea to add this as an extra step in [https://packagecontrol.io/docs/submitting_a_package](url)

wbond:
@bijoythomas I am planning on revising those docs at some point in the future, however I wanted to make sure the bot was working reasonably well before I did so.

bijoythomas:
Thanks @rwols and @wbond. I guess I had the whole package addition/review check workflow inverted where I was making changes to the package and then triggering PR builds over here to make sure things were passing. 

Is it fair to say that the right way of going about this is

1. Create/host the package
2. Run `st-package-reviewer` or the online repo test against the package
3. Iterate/refine the package based on the review output
4. Add the package JSON to the package control repository file
5. Run the Channel Repository Tests
6. Submit PR

PR #6436, opened 2017-07-11
Added Undo close tab repository

scriptburn:


FichteFoll:
https://packagecontrol.io/packages/File%20History also provides this functionality and more.

Besides, there is a stack structure in the stdlib as `collections.deque` (and your custom print function doesn't do anything more than `print`).

PR #6437, opened 2017-07-11
Adding VHDL Mode package

Remillard:

### DIFF
--- a/repository/v.json
+++ b/repository/v.json
@@ -277,10 +277,10 @@
 		{
 			"name": "VHDL Mode",
 			"details": "https://github.com/Remillard/VHDL-Mode",
-			"labels": ["language syntax snippets VHDL"],
+			"labels": ["language syntax", "snippets", "VHDL"],
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": ">=3092",
 					"tags": true
 				}
 			]
--- a/repository/v.json
+++ b/repository/v.json
@@ -275,6 +275,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


repository/v.json
```diff
@@ -275,6 +275,17 @@
 			]
 		},
 		{
+			"name": "VHDL Mode",
+			"details": "https://github.com/Remillard/VHDL-Mode",
+			"labels": ["language syntax snippets VHDL"],
+			"releases": [
+				{
+					"sublime_text": "*",
```
AlexKvazos: Please change to >=3092

AlexKvazos: And also please add commas between tags.

Remillard:
Sorry about that.  I should have studied more examples.  I made the requested changes to the VHDL Mode entry.

rwols:
It looks like a lot of work went into this. Congratulations, and thanks for this package.

FichteFoll:
You might want to add an image to your readme. From my experience, an image showing some highlighting (or other things) is worth a lot for people that are discovering your package.

Remillard:
Thanks for the note.  I'm not entirely sure how people do those little animated gifs but I'll dig around and see if there's a straightforward way to capture that.

FichteFoll:
On Windows, I preferred using ScreenToGif. I haven't created a gif on Linux yet, but would probably just resort to recording it with teiler (which interfaces ffmpeg) and then use ffmpeg again to convert to a .gif, because Github *still* doesn't allow `<video>` embeds.

PR #6441, opened 2017-07-15
Add Goto With Common Suffix

ivantsepp:
This plugin allows you to do file searches with common suffixes. For example, If I opened up my project, I want to be able to paste `https://github.com/ivantsepp/goto-with-common-suffix/blob/master/goto_with_common_suffix.py` and it will open up that file.

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -1371,6 +1371,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
Interesting idea.

ivantsepp:
Thanks!

PR #6442, opened 2017-07-15
Combiner plugin for Sublime Text.

bite-your-idols:
Combine a list of CSS and/or Javascript files into a single one with Sublime Text using nodejs concat module.
There was a CombineAndMinify plugin (just for JS files) but it it not working in ST3 and it seems it it not currently maintained.


repository/c.json
```diff
@@ -2583,6 +2583,16 @@
 				}
 			]
 		},
+				{
```
FichteFoll: That's a bit too much indentation on this line.

bite-your-idols:
key binding changed: https://github.com/bite-your-idols/Combiner/commit/7612bcad0792efff71d5245cfb36c2ae41eb0d18

bite-your-idols:
Sorry, I don't know if it is a matter of time or if I need to make anything else, but change requested by pacakgecontrol-bot has been made.

PR #6444, opened 2017-07-17
Adding QuickEdit

gamcoh:
I am creating a package that allows to edit css code directly into html code using the tag class names, like QuickEdit on the Brackets IDE.

### DIFF
--- a/repository/q.json
+++ b/repository/q.json
@@ -183,6 +183,18 @@
 			]
 		},
 		{
+			"name": "QuickEdit",
+			"details": "https://github.com/gamcoh/QuickEdit",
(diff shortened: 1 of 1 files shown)


gamcoh:
i just added my first tag release, it should be good now

gamcoh:
But i don't understand, i change the key binding, is there some kind of cache system or anything ?

FichteFoll:
Instead of using `re.findall` and then only fetching the first element you might as well just use `re.search`.

Also, you should probably be using the first entry in `window.folders()` and combine paths with `os.path.join` instead of manually adding a platform-specific forward slash.

gamcoh:
Thanks, i just fixed the bugs

PR #6448, opened 2017-07-17
Add "Goto Selection"

fgb:
Package that pre-fills the Goto Anything overlay with the selected text.

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -1360,6 +1360,16 @@
 			]
 		},
 		{
+			"name": "Goto Selection",
+			"details": "https://github.com/fgb/goto_selection",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Goto Usage",
 			"details": "https://github.com/syko/SublimeGotoUsage",
 			"labels": ["javascript", "coffeescript"],


fgb:
https://github.com/fgb/goto_selection/commit/1835a5ff1e0cde6d20b0856d4218fe561c2caf28 fixes the issue pointed out by @packagecontrol-bot (removed the keybinding that clashed with the defaults on win/lin).

FichteFoll:
Looks like https://packagecontrol.io/packages/QuickGotoAnything already does this.

fgb:
I'm familiar with it, but I've had to uninstall it due to it unconditionally overriding the default <kbd>SUPER</kbd>/<kbd>CTRL</kbd>+<kbd>P</kbd> keybinding. In essence, it wouldn't have passed @packagecontrol-bot automated testing that flagged the problem I had with <kbd>CTRL</kbd>+<kbd>;</kbd> above.

_Goto Selection_ adds an argument for choosing whether or not to use the word under the cursor, which fixes the main issue I had with _QuickGotoAnything_, since it usually copied an empty string or a stray character (e.g. a semicolon or bracket) instead of showing me an empty _Goto Anything..._ overlay with all the project files displayed below when I didn't select a thing.

I could submit a fix to _QuickGotoAnything_ that adds this argument, but I doubt I would have much luck changing the default keybindings with the installed base it already has.

PR #6460, opened 2017-07-21
Rebasing Flask Starter

Joepriesto:
After repeated requests to the package maintainer with no response I believe this package is no longer maintained. I would like to re base it to my fork and take over support.

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -972,7 +972,7 @@
 		},
 		{
 			"name": "Flask Starter",
-			"details": "https://github.com/geekpradd/Sublime-Flask-Starter",
+			"details": "https://github.com/JoePriesto/Sublime-Flask-Starter",
 			"releases": [
 				{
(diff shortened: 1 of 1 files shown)


Joepriesto:
Hi is there any reason why this isn't moving forward?

FichteFoll:
I've been on vacation last week and am currently dealing with exams, which is why I haven't been able to set aside time for reviewing. Don.t worry, your package will eventually be looked at.

Joepriesto:
Apologies, I thought there was a bit of a bigger team reviewing and merging!

Obviously, take your time, there's really no rush. I just thought I might have made an error.

FichteFoll:
In theory there are up to three people reviewing, but in practice it's just me.

FichteFoll:
@geekpradd, if you are still alive and object to this change of ownership, please respond here within the next two weeks. If you don't, I will merge after that.

PR #6463, opened 2017-07-24
change the Solo Snippets to a new Github repository

zeusLeeJh:
The old repository is not using any more

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -2388,7 +2388,7 @@
 		},
 		{
 			"name": "Solo Snippets",
-			"details": "https://github.com/zeusLeeJh/solo-snippets",
+			"details": "https://github.com/landray/code-snippets/tree/solo-sublime",
 			"releases": [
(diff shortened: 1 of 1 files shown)


repository/s.json
```diff
@@ -2388,7 +2388,7 @@
 		},
 		{
 			"name": "Solo Snippets",
-			"details": "https://github.com/zeusLeeJh/solo-snippets",
+			"details": "https://github.com/landray/code-snippets/tree/solo-sublime",
```
FichteFoll: I'm not sure whether this will work properly. You should add a `readme` key pointing to the raw readme from that branch, probably `homepage` too, and have `details` point to the repo directly.

This kind of package architecture isn't exactly recommended. It's usually a better idea to have a single code base (with small variations) in a single repository rather than using branches.

FichteFoll:
@zeusLeeJh ping

PR #6466, opened 2017-07-27
Add BrightScript Booster package

lukewang1024:
To provide a complete BrightScript language support with common snippets.

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -1240,6 +1240,17 @@
 			]
 		},
 		{
+			"name": "BrightScript Booster",
+			"details": "https://github.com/lukewang1024/sublime-brightscript",
+			"labels": ["language syntax", "snippets"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Broadcast Color Scheme",
 			"details": "https://github.com/vinkla/broadcast-theme",
 			"labels": ["color scheme"],


FichteFoll:
How is this different from the https://packagecontrol.io/packages/BrightScript package?

lukewang1024:
This package adopts the language definition from [VBScript](https://github.com/SublimeText/VBScript), which provides better language support than the existing `BrightScript` package. Also it has many snippets that could make coding faster.

FichteFoll:
@hastinbe what do you think of incorporating these improvements into your packages or instead have @lukewang1024's package replace yours in the list?

hastinbe:
@FichteFoll this is a copy of the VBScript package with the file type coverage changed. While BrightScript and VBScript are similar they are also very different. For example, this package uses the keywords for VBScript, most of which don't exist in BrightScript, and consequentially also doesn't have keywords that are unique to BrightScript. In addition it contains highlighting mismatches because of nuances in syntax. While I welcome others to improve upon BrightScript, this package does not really provide it. Instead I would encourage @lukewang1024 to fork and incorporate what he/she considers improvements from the VBScript package and re-implement them to work properly for BrightScript and create a PR.

FichteFoll:
@hastinbe thanks for your insights. Your proposal sounds very good to me. @lukewang1024, what do you think about this?

lukewang1024:
Thanks for the insights, @hastinbe. I have not used VBScript myself and was not aware of the difference between them. I think I can start by creating a PR with the snippets. As for the language definition, I am not familiar with how it works and how BrightScript and VBScript differs at the moment so may not tough that in near future.
Closing this one. Thanks @FichteFoll.

PR #6470, opened 2017-07-30
Add Open Files In List plugin

Wargazm:
Hi,
This is a simple plugin that is pretty self explanatory.  I've found it quite useful for my workflow. I didn't find anything similar in Package Control so I figured I'd share with the community. 

Thanks!


AlexKvazos:
Thank you!

PR #6471, opened 2017-07-31
Added TwoColumns

bfelder:
You may proceed with a short description of what the package does and, 
in case a similar package already exists, 
why you believe it is needed
below this line. -->

Described in readme already. :-)


bfelder:
Coming back later.

PR #6475, opened 2017-08-02
Added TwoColumns and ShellVE

pyking2:
Please see Readme.md for a description of what TwoColumns does and why it is a necessary alternative to Origami.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1318,6 +1318,20 @@
 				}
 			]
 		},
+				{
+			"name": "ShellVE",
+			"details": "https://github.com/bfelder/ShellVE",
+			"readme": "https://github.com/bfelder/ShellVE/blob/master/README.md",
+			"author": "bfelder",
+			"labels": ["terminal", "python", "Virtualenv"],
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"platforms": ["linux"],
+					"tags": true
+				}
+			]
+		},
 		{
 			"name": "Shen",
 			"details": "https://github.com/rkoeninger/sublime-shen",


bfelder:
The PR was not merged at the time of my second commit.
I added a new plugin called ShellVE.
It automatically starts a terminal view with a ready python virtual environment, if you open a project whose project file contains a path to that virtual environment.
To the best of my knowledge, there is no similar plugin currently available on PackageControl.

FichteFoll:
In the future, please create a pull request for each package. That way one failed review won't hold back the other from being merged and it's generally more organized.

Your plugin architecture in `TwoColumns/TwoColumns.py` is a bit odd because you explicitly instantiate your EventListener class to run some code on it that doesn't actually depend on any state hold within that class. `_set_two_columns` should probably be an external function and just invoked by both `plugin_loaded` and the hooks in the listener.

Package Control's dependencies are a separate concept from packages. It does not install other packages. You'll need to prompt the user to install `TerminalView` manually if it can't be found on the user's machine (and make that clear in the readme). Besides, to my knowledge TerminalView is also available on macOS.

bfelder:
@FichteFoll Thanks for being strict and doing such thorough code reviews. Indeed this is vital to prevent the Sublime ecosystem to overflow with very bad and even damaging plugins like it is/was with Atom.

I am going to fix the points required.

bfelder:
@FichteFoll I have just overhauled and released new versions of both packages.
I have addressed all the points you mentioned.
1. TwoColumns.py underwent some refactoring. The method of question is now found outside of the event listener class.

2. ShellVE will now give out a status bar message if the dependency TerminalView is not found. It's README.md is changed accordingly as well.

Both packages now feature menu entries to access their key bindings.

FichteFoll:
Your check for whether TerminalView is installed or not is error prone in the occasion that it *is* installed but disabled or when it was installed manually/via git. You should check for `"TerminalView" is in sys.modules` to be sure.

I'll merge it for now anyway since this is only a minor issue.

pyking2:
@FichteFoll New dependency check implemented and released.

PR #6481, opened 2017-08-04
Add VueJS formatter for es6/es7/pug template/less/sass

luozhihua:


luozhihua:
@wbond Could you merge this pull request? If do not merge, please tell me why.

wbond:
@luozhihua Unfortunately I haven't had time to go over this, which is why it hasn't been merged. Most of my open source time recently has been tied up dealing with people trying to launch phishing attacks against the ST community pretending they are Sublime HQ launching an official plugins repository, and the fallout of the privacy issues related to Kite using SideBarEnhancements to track users.

luozhihua:
Thank you, Sublime Text is an awesome text editor，I love it and I'm very glad to  contribute something, If you has free time latter, please have a look for my plugin, thank again

FichteFoll:
What are the differences to https://packagecontrol.io/packages/VueFormatter?

You must also remove `package-control.cleanup` from the repo.

luozhihua:
[VueFormatter](https://packagecontrol.io/packages/VueFormatter) only supported js(es5)/css/html，If the file's syntax using [es6](http://www.ecma-international.org/ecma-262/6.0/)/[es7](http://www.ecma-international.org/ecma-262/7.0/)、[Pug](https://github.com/pugjs/pug) and [Less](http://lesscss.org/)/[Sass](http://sass-lang.com/)，VueFormatter can not format it perfact, I try to contribute for it, but has a lot of changes if make it works well with these syntax, so I wrote this new plugin.

luozhihua:
I removed `package-control.cleanup` and added it to `.gitignore` now.

FichteFoll:
On ST3, the API isn't always available at module load time, which means [this](https://github.com/baixuexiyang/VUEFormatter/blob/master/VUEFormatter.py#L12) call to `sublime.packages_path` may fail and prevent your plugin from being loaded. You'll have to move the construction of this path elsewhere, either to where it is used or into a `plugin_loaded` function.

FichteFoll:
@luozhihua ping

luozhihua:
@FichteFoll I moved `sublime.packages_path` into a function where it is used now, https://github.com/luozhihua/sublime-vue-formatter/blob/master/vue-next-formatter.py

FichteFoll:
Trying to determine the package folder using `__file__` won't work when the package is stored as a `.sublime-package`.
For storing temporary files, you should use the `tempfiles` module. For more persistent temporary storare (i.e. a cache), use `sublime.cache_path()`.
For allowing other processes to access files of the package (i.e. allow `node` to run a javascript file bundled in the package), add a `.no-sublime-package` file to your repository, which will tell Package Control to unextract the package on installation. You could also unextract the file to `sublime.cache_path()` and run it there instead. Use `sublime.load_binary_resource` to load files regardless of them being packed or extracted (ST3 only).
Also note that calls to `window.open_file` with a path based on `__file__` would not work within a `.sublime-package`. Instead, you could use `window.run_command("open_file", {"file": "/".join("${packages}", __package__, ".vueformatterrc")`. Note that `__package__` is only supported in ST3. The `open_file` command also supports the `${platform}` variable.

FichteFoll:
@luozhihua ping

PR #6482, opened 2017-08-06
Remove CSS3_Syntax package

y0ssar1an:
This is an ancient ST2 package of mine that has been deprecated since 2014. It has been replaced by [the CSS3 package](https://github.com/y0ssar1an/CSS3). I'm deleting [the CSS3_Syntax repo](https://github.com/y0ssar1an/CSS3_Syntax).

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3733,16 +3733,6 @@
 			]
 		},
 		{
-			"name": "CSS3_Syntax",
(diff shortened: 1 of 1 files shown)


y0ssar1an:
I don't understand why Travis rejected this. The build was cancelled before it started.

PR #6486, opened 2017-08-07
Added pywinauto dependency

randy3k:
provides the [pywinauto](https://github.com/pywinauto/pywinauto) module.


randy3k:
well, originally, I wanted to replace the need of `pywin32` by `pywinauto`, but it turns out `pywinauto` requires `pywin32` to work. So, it is basically not useful at all.

PR #6487, opened 2017-08-09
add webhare package

unilynx:
This package adds a lot of utility for WebHare developers - whether they have a local installation or only have network/webdav access to such a system.

We understand this market/target group is currently small, but having this package in PackageControl instead of having to manually distribute configurations would make it a lot easier for the sublime/webhare fans to use this integration

### DIFF
--- a/repository/w.json
+++ b/repository/w.json
@@ -207,6 +207,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


unilynx:
Updated the version tags

PR #6490, opened 2017-08-10
Added Meson language syntax

Salamandar:
This is the language syntax for the Meson Build system.
It's my first ST package, so please feel free to criticize ;)


Salamandar:
SemVer tag added.

PR #6491, opened 2017-08-10
MediaMonks Temple plugin

mockingbird87:
The MediaMonks Temple package is used throughout our company to assist with our day to day development workflow. The package is specific to our infrastructure and thus can only be used internally. 
The package facilitates running our Grunt tasks through Sublime Text.


mockingbird87:
Changes have been implemented and package updated on URL

FichteFoll:
Your current version selector only selects a specific build, 3000. It should select a range with `>=3000`.

You should also remove the custom update code from your package if you intend to distribute it via Package Control. It will only create a mess. Besides, forcefully closing the sublime_text process is almost never a good idea.

You should probably unexctract the openssl binaries into `sublime.cache_path()` (or a temporary directory since you seem to overwrite the files every ST restart anyway).

You need to define `global _popupData` in the `showPopUp` function because you have an assignment statement in the function body, which will make this a local variable even if the assignment occurs later.

Instead of `threading.Timer` you can also use `sublime.set_timeout` or `sublime.set_timeout_async`.

FichteFoll:
@mockingbird87 ping

PR #6492, opened 2017-08-10
Replaced Wrapper with WrapIt

pyking2:
Original author @lattespirit offered to remove Wrapper from the channel in favor of a new strongly enhanced version.
_See here: https://github.com/lattespirit/SublimeWrapper/issues/7_

**WrapIt** shares the core function of Wrapper but is a _complete rewrite_. It adds important features which makes it much more powerful than its predecessor:
- the support for custom wrappers
- a new more  maintainable JSON structure
- very easy templating of custom code blocks
- custom key bindings.

### DIFF
--- a/repository/w.json
+++ b/repository/w.json
@@ -858,12 +858,15 @@
 					"branch": "master"
 				}
 			]
-		},
+		},		
 		{
-			"name": "Wrapper",
+			"name": "WrapIt",
 			"description": "Wrap Contents With Specific Code",
-			"details": "https://github.com/lattespirit/SublimeWrapper",
-			"author": "Lattespirit",
+			"details": "https://github.com/bfelder/WrapIt",
+			"readme": "https://github.com/bfelder/WrapIt/blob/master/README.md",
+			"author": "bfelder",
+			"previous_names": ["Wrapper"],
+			"labels": ["text manipulation"],
 			"releases": [
 				{
 					"sublime_text": ">=3000",
--- a/repository/w.json
+++ b/repository/w.json
@@ -862,8 +862,7 @@
 		{
 			"name": "WrapIt",
 			"description": "Wrap Contents With Specific Code",
-			"details": "https://github.com/bfelder/WrapIt",
-			"readme": "https://github.com/bfelder/WrapIt/blob/master/README.md",
+			"details": "https://github.com/bfelder/WrapIt",	
 			"author": "bfelder",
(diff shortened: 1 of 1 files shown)


FichteFoll:
@lattespirit it would be nice if you could explicitly acknowledge this move. You offered to delete the pacakge but didn't necessarily agree on replacing it (although this is the preferred action).

@bfelder you should prepare an update message informing about the changes for the users that have this pacakge installed already.
Besides, `readme` should point to a raw file, which your URL does not. You can just remove the line as PC will pick up the readme from your repo's master branch by default.
Finally, you need to escape some characters from the selection if you use it with the `insert_snippet` command. To be precise, replace `$` with `\$`.

bfelder:
@FichteFoll Great point escaping in `$` in the selections. It escaped my attention.

Where should I place this UpdateMessage? In the readme?

FichteFoll:
See https://packagecontrol.io/docs/messaging.

bfelder:
@FichteFoll @lattespirit 
I have just released a new version of WrapIt.
From its release note:
- added install message to inform about WrapIt to take over from SublimeWrapper
- added menu entry for settings and key bindings
- added escaping mechanism for text selection to prevent interference from '$' containing selections

FichteFoll:
@lattespirit any words from you on this?

@pyking2 you need to adjust the package URL to your rename.

pyking2:
@FichteFoll I'll do in an instant together with the other recent merges.

FichteFoll:
Don't worry about the others, I just spotted that and fixed it myself in fd4c070645a779bc726e3c2c044a397e4083a14d.

pyking2:
Should be done by now.

lattespirit:
@FichteFoll @pyking2 Sorry for the late reply, I agree with transfering this package and i hope more people  
 enjoy the benifit from this new package.

FichteFoll:
@pyking2 I just checked your "update message" and noticed you show it on installation. That won't do since this is *only* shown on installation and any user already using Wrapper will not see the message, while it is instead shown to any new users installing "WrapIt". You need to add a message for your current latest release tag with a message you want previous users to see.

pyking2:
@lattespirit Thank you for your trust. This is very honorable. As you may have noted I have included your alias under contributors, as a special mention. If you like to have your full name state just tell me.

@FichteFoll I have made a new release. The install message should now be shown to any users installing the latest release.

PR #6495, opened 2017-08-11
add LSCheckstyle

kauboy26:
This package helps programmers write Java code that conforms to CS1331's coding style.

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -1391,6 +1391,17 @@
 			]
 		},
 		{
+			"name": "LSCheckstyle",
+			"details": "https://github.com/kauboy26/LimeStyle",
+			"readme": "https://github.com/kauboy26/LimeStyle/blob/master/README.md",
+			"releases": [
+				{
+					"sublime_text": ">3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "LSL",
 			"details": "https://github.com/Makopo/sublime-text-lsl",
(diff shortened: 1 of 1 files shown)


kauboy26:
'.no-sublime-package' is defined since the plugin uses the subprocess module to run the Java .jar bundled with the plugin.

FichteFoll:
`ctrl+l` is a default binding (on Linux and Windows at least) and ST doesn't like overriding these with multi-chord bindings too much. The package reviewer doesn't check for these currently. I suggest using a different binding or have the user specify it themselves. A `.sublime-command` file would certainly be useful, however.

kauboy26:
@FichteFoll I changed the binding the plugin was using, and added .sublime-commands file.

kauboy26:
Keybindings have been removed, users need to set them up according to their liking.

repository/l.json
```diff
@@ -1391,6 +1391,17 @@
 			]
 		},
 		{
+			"name": "LSCheckstyle",
+			"details": "https://github.com/kauboy26/LimeStyle",
+			"readme": "https://github.com/kauboy26/LimeStyle/blob/master/README.md",
```
FichteFoll: The URL to the readme must be direct. You can just remove this line and PC will pick up the correct URL regardless.

kauboy26:
Keybindings have been removed.
The .sublime-package is necessary.
"l.json" has been updated (readme line removed).

PR #6499, opened 2017-08-12
CollaBroText plugin: Writing Collaborative code with commenting feature.

shantanu3637:

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2152,6 +2152,17 @@
 			]
 		},
 		{
+			"name": "CollaBroText",
+			"details": "https://github.com/shantanu3637/CollaBroText",
+			"releases": [
+				{
+					"sublime_text": ">=3084",
+					"platforms": ["linux"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Collider Snippets",
 			"details": "https://github.com/simonsinclair/collider-snippets",
 			"releases": [


FichteFoll:
Platform-specific settings (`collabro (Linux).sublime-settings`) only work inside the default package. This must be named `collabro.sublime-settings`. (Or does it work for you?)

There's a hard-coded path [here](https://github.com/shantanu3637/CollaBroText/blob/e930f7b38d62fa37dd72efa83c659e726a09cc85/CollaBroText.py#L36)

You should store comment data in `sublime.cach_path()` instead of inside your package's directory.

shantanu3637:
Regarding point 1, the settings only configure theme of the comments file, and it works for us as tested on Linux Ubuntu machines, hence not changing.

Regarding point 2 and 3 we have followed what you recommended and made changes accordingly in the latest commits. Would we need to create new version number so as to reflect the changes?

FichteFoll:
You need to create the `sublime.cache_path() + '/CollaBroText'` folder before you try writing to a file within it.

Yes, you'll have to make a new release by pushing a tag for Package Control to pick up the changes.

shantanu3637:
we've fixed the folder creation part in the latest commit. If there are no more issues from your side then we'll create a new tag.

FichteFoll:
Looks good. You only need to push a new tags with the latest changes now.

shantanu3637:
We've pushed a new tag v0.1.1 and incorporated the new changes in it. 👍

PR #6500, opened 2017-08-13
Add ABAP language support plugin

flaiker:
This is just a simple plugin to provide ABAP language support / syntax highlighting. ABAP is the main language in use by [SAP](https://www.sap.com).
The .tmLanguage file for this is taken from the repository that GitHub's linguist also uses for highlighting. You can find some more details in the README.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -24,6 +24,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


flaiker:
I created a release / tag.

PR #6503, opened 2017-08-14
Title updated to improve the way to search the package

gepd:
Just minor changes in the title and labels

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -493,7 +493,7 @@
 			"name": "Deviot (Arduino IDE)",
 			"details": "https://github.com/gepd/Deviot",
 			"donate": "https://gratipay.com/~gepd/",
-			"labels": ["build system", "ide"],
+			"labels": ["build system", "ide", "arduino", "iot", "platformio"],
 			"releases": [
 				{
 					"sublime_text": ">=3000",
--- a/repository/d.json
+++ b/repository/d.json
@@ -490,10 +490,10 @@
 			]
 		},
 		{
-			"name": "Deviot",
+			"name": "Deviot (Arduino IDE)",
 			"details": "https://github.com/gepd/Deviot",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Why do you remove the labels?

gepd:
I don't know if they are necessaries, I've looked into the labels sections in the website, and they aren't listed. Do you think I should keep them?

FichteFoll:
I think labels are only listed if they occur three or more times. It seems like no other package besides yours defines the labels.

I'm okay with removing the labels since it's your package, but I was curious regardless. Tags should probably have a higher impact on search results (or rather have an impact in the first place) so that package descriptions in parentheses aren't needed to improve search accuracy.

Adding the term "Arduino" to your repo's description would also include it in search results for "Arduino" though.

gepd:
Do you mean add it into the Readme file?

My confusion started because, at the beggining, I though in labels like tags, I mean they will be considered in the search results, but if I search the term `Arduino` in the web or in ST, my plugin isn't listed. The same happens with `iot` and `platformio` this last one is in the readme more than one time, so I guess it isn't working as I expected

FichteFoll:
No, I mean this description:
img

The description is fetched from the repository and is searchable on the packagecontrol.io site.

gepd:
Thanks!, I've just updated the description to test it, I'll wait at least a day to see  if there is any change in packagecontrol

FichteFoll:
Changes *should* be discovered within an hour.

gepd:
Ok, I'll close this PR because make the changes you suggested, did what I was looking for.

Thanks!

gepd:
I'll reopen this PR, because I keep having a doubt, I just check that when I search for "arduino" or "platformio" from ST it does not show any result related with my plugin. However if I search `deviot` I can see new the description, is this a bug or is the way that PC works in ST?

PR #6504, opened 2017-08-15
Add Erl-AutoCompletion plugin.

lintingbin2009:

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -919,6 +919,17 @@
 			]
 		},
 		{
+			"name": "Erl-AutoCompletion",
+			"details": "https://github.com/lintingbin2009/Erl-AutoCompletion",
+			"labels": ["erlang autocompletion"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"details": "https://github.com/minimedj/ErlDoc",
 			"releases": [
 				{


lintingbin2009:
The contents of the Default.sublime-mousemap file have been deleted.

FichteFoll:
Using `__init__.py` as the entry point for your package's plugins is suboptimal because the file gets loaded twice by ST (for internal reasons, once by Python's standard importing system and once by the code that explicitly checks for classes interfacing with ST). It works and probably isn't harmful because there isn't a lot to compile for that file (everything is imported), but just as a heads up.

You shouldn't be using a subfolder inside your package as a cache folder. Your package's contents should be considered immutable, because it will be installed as a `.sublime-package` zip archive and thus not directly accessible on the file system. There is a `sublime.load_resource` API to *read* package resource files, but they are not writable. Instead, use a subfolder in `sublime.cache_dir()`.

FWIW, uppercase constants are usually not to be modified, so something like `CACHE` is usually a Bad Idea™. That's just a convention, however.

lintingbin2009:
@FichteFoll  Thanks for your advice.
I've modified the code as you suggested it. But I have a question: when my plug-in is uninstalled by user, the corresponding cache will be deleted together?

FichteFoll:
No, it won't, but it wouldn't have been previously either, iirc. There is an events API for Package Control which you can read up on here: https://packagecontrol.io/docs/events

You could, however, also just remove the cache when `plugin_unloaded` is called, which happens when the package is deactivated (and updated or uninstalled). Depending on what your cache stores, this would be easier.

lintingbin2009:
@FichteFoll Thanks for your answer. I've added the code to remove the cache when my plugin is uninstalled. Can you merge my plugin into the master now?

FichteFoll:
Also note that your plugin will not work on ST2 because of the relative import `from . import utils`. You need to use just `import utils` on ST2, while `from . import utils` is the only way to do it on ST3.

Also also, you can retrieve the actual package's name from the `__package__` global variable, but that is only available from ST3 onwards.

And finally, your unload handler won't be called with the remove event set when it is disabled first and uninstalled later. You *might* want to consider this as well.

lintingbin2009:
Problem 1:
I test my plugin in ST2, that has some problems. ST2 has no plugin_loaded function, so when I open ST2 call sublime.active_window (), it returns None. The code is as follows:
if sys.version_info < (3,): 
    plugin_loaded()
    unload_handler = plugin_unloaded
If it can't be solved, I can remove the support for ST2.
Problem 2: 
It isn't a problem.
Problem 3:
I can't find the disable event of the Package Control event api, so I think that I can put the cache dir in package dir of my plugin, when my plugin is remvoed, the cache dir also be removed.

FichteFoll:
1. You probably shouldn't do window-specific initialization in `plugin_loaded`. Instead, do it in `on_activated`. As a safety net, you might want to check for `sublime.windows()` in plugin_loaded for when the plugin is (re-)loaded while ST is running (when installing, re-enabling the package or modifying the plugin file).

3) Correct, it appears there is no such event. The general rule is that your package is *most likely* being disabled when the `plugin_unloaded` handler is called, but that could also be happening when a single plugin file is reloaded. It depends on whether you are okay with re-creating the cache in each of these occurances, or whether you just want to remove the cache when the package is removed.
I cannot say anything about the reliability of PC removing the package folder if it is an override package. In case you want PC to remove the entire package directory (where you placed its cache into a subfolder), you might have to specify a `.no-sublime-package` file, which has other implications.

I cannot make this decision for you, but I can explain some of the options. I will merge the pacakge for now, since it is in a working state and anything else can be added later.

PR #6506, opened 2017-08-15
Added the SpotifyWeb plugin.

agilesteel:

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -2868,6 +2868,18 @@
 			]
 		},
 		{
+			"name": "SpotifyWeb",
+			"details": "https://github.com/DevInsideYou/SpotifyWeb",
+			"labels": ["spotify", "music"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"platforms": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Sprak Syntax & Completions for else Heartbreak()",
 			"details": "https://github.com/Eforen/sublime-syntax-sprak",
 			"author": "Eforen (Ariel Lothlorien)",


agilesteel:
Hi,

I'm using an external library that reads and writes from/to files. Therefore I cannot use the sublime.load_resource API. I hope this pull request will be accepted. It's the first time I'm submitting a package into packagecontrol.

Cheers
Vlad

FichteFoll:
spotipy uses relative imports internally, so you can just import it relatively as well instead of modifying `sys.path`. 

Just move `lib/spotipy-2.4.4/spotipy` to `lib/spotipy` (and delete everything else except for the license) and then write `from ...lib import spotipy`. The number of dots depends on how deeply nested you are within `src` of course.

agilesteel:
Cool, good to know!

I just did it here https://github.com/DevInsideYou/SpotifyWeb/commit/2bf3743d0a4a0581268c4e251744b9053c43b610

I'm relatively new to python. Does this mean that I don't need the .no-sublime-package file anymore?

Let me know and I'll remove it!

Thx for the help!

FichteFoll:
>Does this mean that I don't need the .no-sublime-package file anymore?

Yes, exactly.

agilesteel:
Awesome!

Done https://github.com/DevInsideYou/SpotifyWeb/releases/tag/v0.0.6

agilesteel:
Hi again,

thanks for accepting my package!

The removal of the .no-sublime-package file seems to prevent me from opening the settings file for my plugin. Would you be so kind to look at line 38 of https://github.com/DevInsideYou/SpotifyWeb/blob/master/src/sublime/SettingsManager.py#L38 and tell me if there is some packagecontrol API that would help me out with that? As of right now I'm trying to open the Packages/SpotifyWeb/SpotifyWeb.sublime-settings file but the SpotifyWeb folder doesn't exist since the plugin is packaged by packagecontrol.

Sorry to bother...

FichteFoll:
You can use the `edit_settings` command, which has been added in a dev build around 3100. It provides the same functionality, basically. Examples are in the default menu (and commands) file. 

Other than that, you can tell ST to load files from the packages using the open_file command and specifying the path as $packages/SpotifyWeb/... This is, to my knowledge, the only way to make st open virtualized resource files.

PR #6508, opened 2017-08-16
Adding open_in_browser

vicke4:
Hi, 

This plugin enables users to open links/files in browsers simply by hovering over it with the help of show_popup API. And this is my first open source contribution. I'm open to suggestions. So, please let me know if anything can be improved. 

Thanks

### DIFF
--- a/repository/o.json
+++ b/repository/o.json
@@ -417,6 +417,16 @@
 			]
 		},
 		{
+			"name": "Open In Browser",
+			"details": "https://github.com/vicke4/open_in_browser",
+			"releases": [
+				{
+					"sublime_text": ">=3006",
+					"tags": true
+				}
+			]
+		},		
+		{
 			"name": "Open in Default Application",
 			"details": "https://github.com/SublimeText/OpenDefaultApplication",
 			"releases": [
--- a/repository/o.json
+++ b/repository/o.json
@@ -431,11 +431,11 @@
 			"details": "https://github.com/vicke4/open_in_browser",
 			"releases": [
 				{
-					"sublime_text": ">=3006",
+					"sublime_text": ">=3118",
 					"tags": true
 				}
 			]
-		},		
+		},
 		{
 			"name": "Open in Default Application",
 			"details": "https://github.com/SublimeText/OpenDefaultApplication",


FichteFoll:
Related: https://packagecontrol.io/packages/Clickable%20URLs

Maybe you want to add this functionality to that package instead?

I also have plans to add an inline phantom for this functionality and render a small icon next to the url, where clicking that would open the page. If you're interested, this is the icon in base64:
`url(data:image/svg+xml,%3C%3Fxml%20version%3D%221.0%22%20encoding%3D%22UTF-8%22%20standalone%3D%22no%22%3F%3E%0A%3Csvg%20xmlns%3Ardf%3D%22http%3A%2F%2Fwww.w3.org%2F1999%2F02%2F22-rdf-syntax-ns%23%22%20xmlns%3D%22http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%22%20height%3D%2213%22%20width%3D%2213%22%20version%3D%221.1%22%20xmlns%3Acc%3D%22http%3A%2F%2Fcreativecommons.org%2Fns%23%22%20xmlns%3Adc%3D%22http%3A%2F%2Fpurl.org%2Fdc%2Felements%2F1.1%2F%22%3E%0A%3Cg%20transform%3D%22translate%28-36.467808%2C-258.39005%29%22%3E%0A%3Cpath%20style%3D%22opacity%3A1%3Bfill%3A%233366bb%3B%22%20d%3D%22m41.47%2C259.4%2C7%2C0%2C0%2C7-2-2-3%2C2%2C0-1%2C3-2.25%2C1%2C1%2C0-3.75-3.75%2C0%2C1%2C1-2.25%2C3-1%2C0%2C2-3z%22%2F%3E%0A%3Cpath%20style%3D%22opacity%3A1%3Bfill%3A%233366bb%3B%22%20d%3D%22m43.47%2C261.4-5%2C0%2C0%2C8%2C8%2C0%2C0-5-1%2C0%2C0%2C4-6%2C0%2C0-6%2C4%2C0z%22%2F%3E%0A%3Cpath%20style%3D%22opacity%3A1%3Bfill%3A%2315a5ea%3B%22%20d%3D%22m40.55%2C263.9c0-0.6213%2C0.6213-0.6213%2C0.6213-0.6213%2C1.864%2C0.6213%2C3.107%2C1.864%2C3.728%2C3.728%2C0%2C0%2C0%2C0.6213-0.6213%2C0.6213-1.243-1.864-1.864-2.485-3.728-3.728z%22%2F%3E%0A%3C%2Fg%3E%0A%3C%2Fsvg%3E%0A)`
though ST most likely doesn't support svg and it needs to be converted to a png first.

vicke4:
Yeah, I also stumbled upon ClickableUrls. But, I don't think it's maintained. Also, the way they're detecting URLs is totally different from mine. I will have to change the whole code if I have to add this functionality there. 

I checked the phantom way. It seemed cool. So, I've added it. Kindly check.

FichteFoll:
The way you're reading the image won't work when the package is inside a `.sublime-package` archive. You need to use `sublime.load_binary_resource` for this, which is not available at module load time since it's a sublime API function.

Other than that, I actually think that having the phantom after the URL and not next to the current cursor position looks better. You might want to try that, too. And finally, since you're using the Phantom API, you'll need to update the version restriction to something higher than 3006. I don't know exactly when this was added, but you just use the latest (or second latest) beta release number.

vicke4:
Okay I'll look for an alternate way to read the image. Also, having the phantom after the URL will make people to move the cursor to URL end which I didn't want. Having it next to current position will save some milli seconds. Don't you think??

FichteFoll:
> Having it next to current position will save some milli seconds. Don't you think??

There's a tradeoff to be paid here. Either the phantom obscures the URL by being inserted in the middle, or the user needs to move the mouse futher, taking more time to click it.

I suppose with that in mind, there are two solutions for this:

1. Display the link in a popup instead of a phantom. That way the URL isn't split and layout doesn't change on hover. This would be very similar to the current behavior.
2. *Always* display links for all URLs in phantoms behind them. This slightly modifies the layout for text after the URL (talking about alignment, mostly), but users know immediately where they need to click and what is recognized as a URL.

Maybe add a setting for the user to decide which they prefer? (Initially, I had 2. in mind when I proposed this.)

vicke4:
> 1. Display the link in a popup instead of a phantom. That way the URL isn't split and layout doesn't change on hover. This would be very similar to the current behavior.

That's what I had implemented at first...:p

> 2. Always display links for all URLs in phantoms behind them. This slightly modifies the layout for text after the URL (talking about alignment, mostly), but users know immediately where they need to click and what is recognized as a URL.

I'm trying to implement this. The plan is to remove the matched URL from the buffer and replacing it with the `<a></a>` phantom on hover. When the cursor is away phantom will be replaced by the text it had before. Is it a proper way?

Also, I'm facing a problem while implementing this. I'm getting `Parse Error` when the URL has `&`

FichteFoll:
>The plan is to remove the matched URL from the buffer and replacing it with the <a></a> phantom on hover.

That's not what I proposed, though. I mean, you are free to implement it however you like, but my idea was to have the link to click on visible *permanently* immediately after the URL. Kind of like some websites do, when a link leaves the current domain. Unfortunately, I don't have an example for this right now.

>Also, I'm facing a problem while implementing this. I'm getting Parse Error when the URL has &

Pipe the URL through `html.escape` before inserting it.

vicke4:
I totally get you. 

https://en.wikipedia.org/wiki/Game_of_Thrones_(season_7)#References - Like this??

Yeah, we can have it as a setting in future, I'll work on it.  Bare minimum how this plugin must behave? This discussion is going on for so long...:)

FichteFoll:
Yeah, exactly like on wikipedia. (I was actually thinking of using that as an example but then too lazy to verify whether they actually do this.) Sounds good.

FichteFoll:
I just noticed, but phantoms were added in build 3118, so you should update the version requirement either to that or to the next beta release (which was 3126).

FichteFoll:
@vicke4 ping

vicke4:
@FichteFoll Extremely sorry for the delay,
I have made the suggested changes. Could you please check once?

PR #6515, opened 2017-08-16
Add XPX language syntax, auto-complete, snippets and theme

pmilon:
Add XPX language syntax, auto-complete, snippets and propose a syntax coloring theme to use with syntax.

### DIFF
--- a/repository/x.json
+++ b/repository/x.json
@@ -162,6 +162,16 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
Please see what I wrote regarding `__init__.py` here: https://github.com/wbond/package_control_channel/pull/6504#issuecomment-323375845d

PR #6516, opened 2017-08-16
Gopher Sauce syntax support and completion.

cheikhshift:
Add [Gopher sauce](https://github.com/cheikhshift/Gopher-Sauce) syntax and auto-complete


FichteFoll:
By convention, **auto-** completions are context-aware and provide complitions based on previously written text. Since your package only contains static completions, it cannot be automatic. Please adjust the package name accordingly.

PR #6525, opened 2017-08-19
added Python Debugger

wvlia5:
Link to your code repository: https://github.com/wvlia5/SublimeDebugger
Link to the tags page with at least one semver tag: https://github.com/wvlia5/SublimeDebugger/releases

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -2474,6 +2474,16 @@
 			]
 		},
 		{
+			"name": "Python Debugger",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Your imports will fail because you hardcoded them as `from SublimeDebugger.backends import dbPython3` but the package will be installed as "Python Debugger". Instead, use relative imports like `from . import dbPython3`.

wvlia5:
Corrected relative imports.

PR #6530, opened 2017-08-21
added DXMate package

jtowers:

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -1447,6 +1447,17 @@
 				}
 			]
 		},
+				{
+			"name": "DXMate",
+			"details": "https://github.com/jtowers/dxmate",
+			"labels": ["salesforce", "dx", "salesforce1", "force.com", "ide", "apex", "visualforce"],
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


jtowers:
.no-sublime-package is used because the package uses the subprocess module to call the .jar bundled with the plugin.

FichteFoll:
I suspect that your imports will fail because you wrote the package name in lower case but the package is installed as "DXMate" and Python's imports are case-sensitive.

Just use relative imports instead, e.g. `from .lib.printers import PanelPrinter`.

`sublime.packages_path()` may not available at the time classes are evaluated, so [this](https://github.com/jtowers/dxmate/blob/ff363b45492746a8cec5cee8150979c5ecd94f1d/dxmate.py#L24) call may result in an exception and will prevent the entire plugin from being loaded. See also the end of this page: https://www.sublimetext.com/docs/3/porting_guide.html

jtowers:
Thanks. I've fixed the import statements and put the sublime.packages_path() call in a method that's only called after the plugin loads.

PR #6532, opened 2017-08-21
Add Purescript IDE

b123400:
This plugin integrates Purescript IDE server with Sublime Text 3, with functions like showing type info when cursor hovers identifiers, autocomplete, rebuild, error report and etc.


FichteFoll:
Note that the ViewEventListener class has been added somewhere around build 3100 (rather than 3000).

PR #6544, opened 2017-08-29
Adding a new package (Yet Another Launcher) to PackageControl  

dahanbn:
Yet Another Launcher - a Sublime Text 3 launcher plugin that is an easy launcher for local files, directories or urls.

### DIFF
--- a/repository/y.json
+++ b/repository/y.json
@@ -143,6 +143,7 @@
 		{
 			"name": "Yet Another Launcher",
 			"details": "https://github.com/dahanbn/Yet-Another-Launcher",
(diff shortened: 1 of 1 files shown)


dahanbn:
I will do the PR again after adding the semver tag.

FichteFoll:
You don't have to create a new PR to re-run the tests. They are also run when a PR is re-opened and can be triggered manually.

dahanbn:
Didn't know that. I waited 24 h as in the forum suggested and then I closed it and resubmitted. Thanks for merging my first PR. :-)

PR #6551, opened 2017-09-02
Commit Clipboard History for Sublime Text 3 

gdiamantg:

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1321,6 +1321,16 @@
 			]
 		},
 		{
+			"name": "Clipboard History (Sublime Text 3)",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Since the old repo doesn't seem to be maintained anymore and I find the trailing `(Sublime Text 3)` part to be unreasonably ugly, I suggest replacing the original package with this fork. It should work with ST2 just fine since only `xrange` references were removed or adjusted.

@kemayo, if you're still alive and object to @gdiamantg taking over your package's spot here, please speak up within 2 weeks.

FichteFoll:
@gdiamantg you adjust your PR to replace the existing package.

evandrocoan:
@gdiamantg ping.

PR #6555, opened 2017-09-03
Add package: ocp-indent

bluepichu:
ocp-indent is a tool for OCaml that auto-indents files.  This package ports ocp-indent to Sublime Text by modifying inserted text as it is written and by providing commands to indent a set of lines or the whole file.

### DIFF
--- a/repository/o.json
+++ b/repository/o.json
@@ -132,6 +132,16 @@
 			]
 		},
 		{
+			"name": "ocp-indent",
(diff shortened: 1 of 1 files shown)


bluepichu:
Context on the warnings:

- The binding for enter indents the next line with `ocp-indent` in addition to inserting a newline.
- The binding for tab reindents the selected line(s) (rather than indenting them one step further).

PR #6563, opened 2017-09-04
added assorted_gutter_themes package

TruJared:


TruJared:
Not sure what I can do about this error? I'm assuming that possibly you don't push themes like this through package control anymore as I haven't seen any added for over 2 years. Please advise. Thanks!

FichteFoll:
The package reviewer just wasn't designed for situations like these. I, as a human, can still merge the package if I determine that it is indeed a useful addition, which it is since it directly interfaces with another package that we list.

I'll revisit the PR once you reopened it.

TruJared:
Thanks so much. This is my first pull request, so I'm 57% clueless on how this works.

PR #6565, opened 2017-09-05
Add pyFlakesLw to the repository

ibensw:


FichteFoll:
Please change or remove the default binding, since it overrides a default binding.

PR #6566, opened 2017-09-05
add a new plugin luatool

piecepaper:

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -1513,6 +1513,17 @@
 			]
 		},
 		{
+			"name": "LuaTool",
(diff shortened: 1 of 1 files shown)


FichteFoll:
You should look into using `sublime.cache_path()` for your cache data storage. Note that this only exists in ST3, but since that's out now there shouldn't be an issue with dropping support for ST3.
However, you need to consider that the API is not always available at module load time and may return `None`. You should probably delay loading the file until `plugin_loaded` anyway.

FichteFoll:
@piecepaper ping

PR #6570, opened 2017-09-08
Add bson

idleberg:


FichteFoll:
You should discourage usage of the `bson.patch_socket` function since all plugin code in Sublime Text shares the same environment and monkey patching in unknown environments is always dangerous.
I hope you weren't planning on using it either.

The other part of this package is fine, obviously.

PR #6571, opened 2017-09-08
HL7 Redux

cjohnson496:
This is a change to the existing HL7 syntax to allow for more message types. More will be added as needs or requests are met.
Thank you,
Charlie

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -662,6 +662,18 @@
 			"releases": [
 				{
 					"sublime_text": "*",
+					"branch": "master",
+					"tags": true
+				}
+			]
+		},
+		{
+			"name": "HL7 Redux",
+			"details": "https://github.com/cjohnson496/HL7-Syntax-Highlighting",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
 					"tags": true
 				}
 			]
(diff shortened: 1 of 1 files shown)


FichteFoll:
What are the differences to @craighurley's https://github.com/craighurley/sublime-hl7-syntax? Have you tried reaching out to the author to submit a pull request?

If @craighurley doesn't maintain their package anymore, we can replace the entry here on package control with your fork, unless they object within the next two weeks.

I'll have to investigate into why packagecontrol-bot fails to check the package.

craighurley:
I'm more than happy to receive PRs at https://github.com/craighurley/sublime-hl7-syntax.  Right now I don't see any real difference in this PR.

craighurley:
I've released `1.0.1` (https://github.com/craighurley/sublime-hl7-syntax/releases/tag/1.0.1) which contains the extra message type, SIU, that @cjohnson496 sought to add.  I've confirmed `1.0.1` is available via package update in sublime.

@cjohnson496 if you need more message types added, send a PR my way, having 2 packages do the exact same thing is not optimal.

:)

FichteFoll:
I determined the reviewer failure to be caused by a malformatted property list, where the first tag within `<plist>` is `<key>` and not `<dict>`.

Due to the inactivity of this PR despite @craighurley's offer to accept contributions, I'll consider this matter to be solved for the future by pull requests to the existing repo.

PR #6572, opened 2017-09-09
add TestExUnit repository

Dania02525:

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -564,6 +564,19 @@
 			]
 		},
 		{
+			"name": "TestExUnit",
+			"details": "https://github.com/Dania02525/TestExUnit",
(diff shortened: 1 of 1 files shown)


Dania02525:
I see this has a warning, but the keymap is context specific to elixir files, and the shortcut matches that found in the [TestRspec](https://github.com/astrauka/TestRSpec) library- which makes it extremely helpful for those who jump back and forth between ruby and elixir source.

FichteFoll:
The bindings probably only affect OSX since they use `super`, but thanks for the heads up. I don't know exactly what these bindings do by default, but I can say that I would be annoyed if a package overrode the binding to re-open a closed tab whenever I am using the Python syntax, for example. Please do this investigation yourself.

PR #6578, opened 2017-09-12
Add Nginx Log Highlighter package

jakebathman:
This PR will add a new package, which defines a syntax for NGINX access.log files.

Please let me know if there are any questions or suggestions.

Thanks!

*Resubmitted with semver tag on the repo*


jakebathman:
Updated the package to support version >= `3092` to support the .sublime-syntax spec

PR #6586, opened 2017-09-13
Add Monokai++

dcasella:


FichteFoll:
People sure love this name … https://packagecontrol.io/search/monokai

Imo yours doesn't even look similar to the "original", so I wonder why you didn't just choose a new name that would distinguish your color scheme from the competition. Or is this just hopping onto a bandwagon?

Anyway, you can still rename your package later if you want to. I don't care about all these monokai packages anymore. To be fair, I also have my own color scheme, so it's not like I'm your target audience.

dcasella:
I decided to go with a Monokai name just because most of the colors are either from the Monokai palette or similar to it; I may add or work on some colors in the future, and I'll be sure to take your advice into consideration.
Thanks for your reply and for merging.

PR #6591, opened 2017-09-14
UWaterloo SE212 - Sublime plugin for George (language) to test program correctness

shividhar:

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -254,6 +254,17 @@
 			]
 		},
 		{
+			"name":"Georgify",
+			"details":"https://github.com/shividhar/Georgify",
+			"labels":["correctness proofs", "uwaterloo"],
+			"releases":[
+				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
You should to mention in your readme that the results are provided by a web service and all the code in the current view is sent to a remote.

We haven't written down this requirement, but it has been agreed upon from the past fiasco with SideBarEnhancements. See here for details: https://forum.sublimetext.com/t/rfc-default-package-control-channel-and-package-telemetry/30157

shividhar:
@FichteFoll I added the disclaimer. I hope it's alright I paraphrased what you said. Is there a standardised disclaimer I can put up instead?

FichteFoll:
No, there isn't a standardised disclaimer since this is the first time I had to note it, but yours passes imo.

PR #6592, opened 2017-09-14
added theme - classic

IceTimux:
Added the classic theme

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -577,6 +577,18 @@
 			]
 		},
 		{
+			"name": "Theme - Classic",
+			"details": "https://github.com/IceTimux/Sublime-Text-3-Classic-Theme",
+			"labels": ["theme"],
+			"previous_names": ["Theme - Classic"],
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


wbond:
You didn't actually test installing this, did you?

wbond:
Also, how is this different than https://packagecontrol.io/packages/Theme%20-%20Retina?

IceTimux:
@wbond Not sure what you mean, but its tested. If you take a loot at the Retina theme you can see that it's broken in the latest build, there are missing assets

FichteFoll:
There was an issue with the retina theme being temporarily removed from the package control channel. Updates weren't picked up anymore but the package was still cached and installable.

This has since been fixed and the theme was moved to the SublimeText org: https://github.com/SublimeText/LegacyTheme. Because the theme isn't maintained anymore, issues have been disabled but pull requests can still be submitted. Is this more to your liking, @IceTimux?

IceTimux:
@FichteFoll Yeah that's really good!

PR #6596, opened 2017-09-15
Add Scalafmt package

fedragon:
Hi, 
I know about other such plugins (`ScalaFormat`, `Scalariform`) and have tried them out before developing this plugin: the difference is in the formatting engine used (`scalafmt` in this, `scalariform` in the others). Seeing that those plugins are slim wrappers on top of `scalariform`, it didn't seem to me a viable option to generalize them to support `scalafmt` as well. 
I hope this can save you some time :)

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -272,18 +272,6 @@
 			]
 		},
 		{
-			"name": "ScalaFormatter",
-			"details": "https://github.com/fedragon/scalaformatter",
-			"labels": ["scala", "formatting"],
-			"releases": [
-				{
-					"sublime_text": ">=3000",
-					"platforms": "osx",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "Scala Snippets",
 			"details": "https://github.com/iDev0urer/sublime-scala-snippets",
 			"labels": ["snippets", "scala"],
@@ -301,6 +289,18 @@
 			"releases": [
 				{
 					"sublime_text": ">=3084",
+					"tags": true
+				}
+			]
+		},
+		{
+			"name": "ScalaFormatter",
+			"details": "https://github.com/fedragon/scalaformatter",
+			"labels": ["scala", "formatting"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"platforms": "osx",
 					"tags": true
 				}
 			]
--- a/repository/s.json
+++ b/repository/s.json
@@ -294,13 +294,12 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
Since the unique property of your package seems to be that it uses `scalafmt`, why didn't you name your package like that as well? Seems to make more sense to me than "ScaFor", which says absolutely nothing by itself.

You should also update your installation instructions for Package Control, as they don't make a lot of sense when it is installed with PC.

fedragon:
Hi @FichteFoll, 
thank you for the feedback. I didn't name it like that because I'm not one of the `scalafmt` authors (nor affiliated with them), I'm just using it to do what I want to do: format Scala code (and I'm using other tools such as `nailgun` to get the job done). 
Since `ScalaFormat` is already taken, I just abbreviated both words... other alternatives, from the top of my head: `FormatScala`, `ScalaFormatter`... if you think one of them would be better than the current name, please let me know and I will make the appropriate changes!

ps: as for installation instructions, I just updated them to include Package Control (my plan was to update them as soon as this pull request got approved)

fedragon:
Plugin renamed & installation instructions updated!

fedragon:
Hi @FichteFoll,
is there any update on the status of this PR? I think I addressed your comments in my commit from last week

FichteFoll:
I don't have a set schedule for reviewing pull requests, so I just do them when I have enough time and feel like it. If you haven't heard from me, I haven't looked at it yet.

To me personally, using the name `scalafmt` just makes the most sense as a package name. Using names that are associated with external programs is a common way to express something being a bridge between Sublime Text and that program, or (web) service. Unless you fear getting sued for a trademark violation (which I doubt the scalafmt guys have filled), there is no issue with using that name. 

You are still free to follow your own guidelines, however in my opinion the names `ScalaFormat` and `ScalaFormatter` are *really* similar and just by the names themselves, I could absolutely not point out the difference. If one of them was named `scalafmt` however and I was knowledgeable in this area, I could tell that `scalafmt` *probably* uses the external tool `scalafmt` (whereas I still wouldn't know anything about the internals of ScalaFormat).

For searchability, I suggest including the word "format" in your repo's description, which you already do in "formatter". ("scalafmt" is included as well, so that's nice.)

fedragon:
Hello again,
I apologize for putting pressure on you, I was honestly just concerned that there was still something wrong in my changes.

Thank you for your feedback! I renamed the package as you suggested and updated the repository topics.

Keep up the good work!

PR #6599, opened 2017-09-16
add SmartIM: reset input method to us when enter normal_mode, and restore previous input method when you enter insert_mode

icymind:

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -2179,6 +2179,20 @@
 			]
 		},
 		{
+			"name": "SmartIM",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Due to how importing works in ST2, using `__file__` anywhere other than in the global module scope is unreliable and your code will not work properly because of it. I suggest just dropping ST2 support since ST3 is officially released now.

icymind:
@FichteFoll I have replaced `__file__` with `sublime.packages_path()`, could you do me a favor by review it again? thanks

PR #6601, opened 2017-09-17
+Libretto

denis020611:

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -704,7 +704,7 @@
 			"labels": ["language syntax"],
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": ">3092",
 					"tags": true
 				}
 			]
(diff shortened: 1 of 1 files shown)


FichteFoll:
In addition to the warnings emitted by the bot, you should choose a distinct name for your `XML.sublime-syntax` file because syntax-specific settings are applied based on that file name, not based on the display name. How is that different from the one in the default packages anyway?

denis020611:
We can use XML in Libretto and also Libretto in XML (see test.ltt), so default XML packages need some changes. Now I put XML syntax into Libretto.sublime-syntax file.

FichteFoll:
Depending on how this embedding works (I didn't investigate closer), you might be interested in using `with_prototype` as explained here: https://www.sublimetext.com/docs/3/syntax.html#include-syntax

Be sure to test this thorougly if you end up using it however, since I cannot evaluate whether this will work properly for you or not.

PR #6603, opened 2017-09-17
add syntax support for various 65816 assemblers.

ksherlock:


FichteFoll:
Actually, since you have tmLanguage files for all but Macrosss.sublime-syntax, I'd say it is okay to release this package for ST2 as well, but since less than 5% are using this, it probably doesn't really matter.

PR #6604, opened 2017-09-18
add language syntax for M/MUMPS/Cache Objectscript.

ksherlock:


FichteFoll:
As mentioned in the other PR, since you have a tmLanguage file for your onl .sublime-syntax file, it's okay to release this package for ST2 as well, but since less than 5% are using this, it probably doesn't really matter.

PR #6608, opened 2017-09-18
rename "Theme - Retina" to "Theme - Legacy"

FichteFoll:
PR to be merged together with the potential update of the Retina theme to rename it's main theme file to `Legacy.sublime-theme` in accordance to the rename.


FichteFoll:
cc @braver

PR #6609, opened 2017-09-19
Add JBuilder support for sublime

zwade:
Adds a JBuilder-based build system for sublime. This is still relatively new, and only supports target selection and building. However, it makes it significantly easier to use `merlin` and `OCaml` in general by making it easy to build projects from within sublime.

As mentioned, its still very basic, but there are several people whom I know will use this functionality.

Thanks,
Zach (@zwade)

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -443,6 +443,19 @@
 			]
 		},
 		{
+			"name": "JBuilder",
+			"details": "https://github.com/zwade/sublime_jbuilder",
+			"author": "Zachary Wade",
+			"labels": ["ocaml", "build-system", "jbuilder", "jbuild"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"platforms": ["osx", "linux"],
+					"tags": true
+				}
+			] 
+		},
+		{
 			"name": "JCommander",
 			"details": "https://github.com/zanuka/jcommander",
 			"labels": ["snippets","auto-complete","utilities"],


FichteFoll:
You should avoid modyfing `sys.path` since it is shared with all other plugins. You can import modules in your package with relative imports on ST3, i.e. `from .sexp import parse_sexp`.

You're leaking file handles. This is especially problematic on Windows, so you should be using `open` as a context manager in a `with` statement.

I'm not sure what you're doing with the path you build for `find_targets_exe`, but it doesn't seem to point to a file that exists and seems to be built "live". That said, the path points to something relative to your plugin file (and within the package), which will be an issue when this package is installed as a `.sublime-package` archive. 
You can either build the file in `sublime.cache_path()` (and put the required files there by reading them with `sublime.load_binary_resource`) or add a `.no-sublime-package` file to your repo, which tells PC to unextract your package into the user's Packages dir.

zwade:
Thanks for the feedback! I'll get on this ASAP

zwade:
How does version `0.2.0` look. I think I addressed all of the concerns you mentioned, and tested it as a `.sublime-package` file.

FichteFoll:
That looks a lot better (and kudos for not going the easy route with `.no-sublime-package`), but unfortunately `sublime.cache_path()` may be inaccessible at the time the module is loaded, so you need to wrap that code and any other code that depends on it into a `plugin_loaded` function, which will be run by ST once the API is ready.

Another thing is that I'm not exactly sure when window commands are instantiated (and this is not documented either, to my knowledge, so it may change at any time), which is why you shouldn't set `self.view = window.active_view()` inside your `__init__` methods. Just use `view = self.window.acive_view()` in the run methods. It will be shorter in the long one anyway.

FichteFoll:
@zwade ping

PR #6610, opened 2017-09-19
Add a simple Web Search package for Sublime 3

asalahli:
Hi,

This is a package to perform a web search (currently only supports DuckDuckGo) from Sublime Text 3.

It includes two commands:
1. **Web Search** prompts the user for a query and performs a search with that query
2. **Web Search Selected** performs a search with the selected text. If there are multiple regions selected, it will display a list of first 10 selection and ask user to pick one.

When I added my repo to the channel, I realized that an identical (and more feature rich) package exists for Sublime Text 2. In the future, I might try to merge these to packages together, but that involves contacting the author, doing some refactoring etc, which is more work, so I think merging this as is would still be valuable until then.

Let me know if there's any additional work I need to do for this PR to get merged.

Thanks,
Azad

### DIFF
--- a/repository/w.json
+++ b/repository/w.json
@@ -257,6 +257,17 @@
 				}
 			]
 		},
+        {
+            "name": "Web Search",
+            "details": "https://github.com/asalahli/sublime-web-search",
+            "labels": ["browser integration", "search", "web"],
+            "releases": [
+                {
+                    "sublime_text": ">=3000",
+                    "tags": true
+                }
+            ]
+        },
 		{
 			"name": "WebSuite",
 			"details": "https://github.com/bthorben/WebSuite",
--- a/repository/w.json
+++ b/repository/w.json
@@ -134,6 +134,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


asalahli:
^ I fixed this. Not sure how to trigger a re-run.

FichteFoll:
The names "Web Search" and "WebSearch" (the existing one) are too similar. I suggest to do exactly as you mentioned earlier and try to contact the author of the existing package for either patching in ST3 compatibility or asking them to have their package be replaced with yours. I expect less than 2% to be using ST2 anyway, so compatability shouldn't be of high concern.

cc @catalinc

Besides, there are also https://packagecontrol.io/packages/OnlineSearch and https://packagecontrol.io/packages/Search%20Anywhere, which are compatible with ST3 and allow adding custom search engines.

asalahli:
Thanks for the feedback. I wasn't aware of OnlineSearch package. It works for all my use cases, and I don't see any need for a yet another search package.

PR #6612, opened 2017-09-20
Add tmux_open (later 'tmux') package, rename Tmux to 'Tmux Syntax Highlight'

huntie:
New package providing commands to open [tmux](https://github.com/tmux/tmux#readme) windows at the current file or project folder. Looking to provide more functionality in future, hence the wide-reaching name (another package `Tmux` already exists and provides syntax highlighting). Lower case `tmux` would be the ideal package name, if it's allowed to be that similar – the description may be adequately differentiating.

Many thanks,
Alex

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -4209,18 +4209,6 @@
 			]
 		},
 		{
-			"name": "SublimeTmux",
-			"details": "https://github.com/huntie/sublime-tmux",
-			"labels": ["tmux", "terminal"],
-			"releases": [
-				{
-					"sublime_text": ">=3000",
-					"platforms": ["linux", "osx"],
-					"tags": true
-				}
-			]
-		},
-		{
 			"details": "https://github.com/R3AL/SublimeTransporter",
 			"releases": [
 				{
--- a/repository/s.json
+++ b/repository/s.json
@@ -4209,6 +4209,18 @@
 			]
 		},
 		{
+			"name": "SublimeTmux",
+			"details": "https://github.com/huntie/sublime-tmux",
+			"labels": ["tmux", "terminal"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"platforms": ["linux", "osx"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"details": "https://github.com/R3AL/SublimeTransporter",
 			"releases": [
 				{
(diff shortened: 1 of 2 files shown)


FichteFoll:
The [existing "Tmux" package](https://github.com/kei-q/sublime-tmux-syntax-highlight) seems to be about highlighting its config files only, which is a much more specific scope than the name "tmux" would imply. Thus, I'm inclined to rename the existing package to "Tmux Config Syntax" or similar and allow this package to take the "tmux" name instead. This would have to occur over a let's say 4 week period to have existing installations of the "Tmux" package properly renamed by Package Control instead of downloading this new "tmux" package.

Either way, prefixing your package name with "Sublime" is disallowed, because any package on PackageControl is for "Sublime Text", unless in very specific exceptions, so "SublimeTmux" is a no-go. You might want to add your package as "Open Tmux" until the plain "tmux" name is freed. Due to case-insensitive file systems (NTFS on Windows), packages names need to be case-insensitively unique as well.

@kei-q, do you object to having your package renamed in favor of this one?

FichteFoll:
Since @kei-q didn't reply, you may adjust the PR to do what I suggested earlier.

huntie:
Thanks @FichteFoll. I've added two further commits to clearly illustrate the changes (although the total file changes are resultingly minimal), and this should be ready to merge.

Since the existing package project repo is called [sublime-tmux-syntax-highlight](https://github.com/kei-q/sublime-tmux-syntax-highlight), I've renamed their entry to "Tmux Syntax Highlight", preserving the uppercase T used in their README and as originally submitted.

Then after some more deliberation over case formatting, I've entered my package under the lowercase name "tmux", as this is consistently how the tmux project is referred to. PR in my repo ready to action this. https://github.com/huntie/sublime-tmux/pull/4

FichteFoll:
>This would have to occur over a let's say 4 week period to have existing installations of the "Tmux" package properly renamed by Package Control instead of downloading this new "tmux" package.

So, you'll have to add `previous_names` to the old tmux package,use a different name for your own for now and in ~4 weeks you can submit another PR that removes the `previous_names` entry for the old package and instead renames yours.

huntie:
@FichteFoll Got you, updated. I've added my package as "tmux_open" for the time being.

FichteFoll:
Thanks for following through with my requests. Please include a reference to this PR when you create the follow-up.

PR #6617, opened 2017-09-22
add jenkinsfile syntax highlighter

bjarneo:
Automatically sets syntax highlighting for the Jenkinsfile when opened.

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -551,6 +551,15 @@
 					"tags": true
 				}
 			]
+		},		{
+			"name": "Jenkinsfile syntax",
+			"details": "https://github.com/bjarneo/jenkinsfile-syntax",
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
Your package will be installed as "Jenkinsfile syntax", but in your plugin you hardcode the path `'Packages/jenkinsfile-syntax/Groovy.tmLanguage'` (note the hyphen). Besides, you are splitting on forward slashes where Windows uses backslashes. Instead, you should be using `os.path.basename`.

That aside, you don't need a plugin for this at all. Just add (or set) `jenkinsfile` in the syntax definition's `fileTypes` array. Or, even better, if a `jenkinsfile` is always supposed to be using the Groovy syntax, you might as well do this for the syntax in the default packages by submitting a PR to the repo: https://github.com/sublimehq/Packages.

PR #6618, opened 2017-09-22
added weoml syntax package

damky:
Adds syntax highlighting for proprietary language, as well as extra completions for weo keys and keyboard shortcuts for weo brackets.  This is my first ever pull request, please let me know how I did. 👍 

Cheers,
Doug

### DIFF
--- a/repository/w.json
+++ b/repository/w.json
@@ -278,6 +278,17 @@
 			]
 		},
 		{
+			"name": "WEOML Syntax",
+			"details": "https://github.com/damky/WEOML",
+			"labels": ["weo", "keys", "syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "WeSnipp Add",
 			"details": "https://github.com/osternaudClem/Sublime-WeSnipp-Save",
 			"labels": ["wesnipp", "snippet"],
--- a/repository/w.json
+++ b/repository/w.json
@@ -283,7 +283,7 @@
 			"labels": ["weo", "keys", "syntax"],
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": ">=3092",
 					"tags": true
 				}
 			]


damky:
v0.1.0-alpha tag added. https://github.com/damky/WEOML/releases/tag/v0.1.0-alpha

FichteFoll:
I'm not sure why you are including `html_completions.py` in your package. It looks to be unmodified compared to the one in the HTML package. Completions should be applied to your file anyway, since the base scope is also matched by `text.html`, which I think is intended.
The same goes for `Miscellaneous.tmPreferences`, `Symbol List - ID.tmPreferences`, `Comments.tmPreferences`, and `encode_html_entities.py`.

Also, apparently `control` is an alias for `ctrl`, so you can ignore the errors regarding that. The warning about 3092 is still valid, however.

damky:
Thanks! I wasn't sure about the duplicate files. I've taken care of those files except for the `Comments.tmPreferences` since I have custom comment code. Or does that even matter? 

Regarding the warning about build 3092, can that be fixed by changing `"sublime_text": "*"` to `"sublime_text": ">=3092"`?

FichteFoll:
>Regarding the warning about build 3092, can that be fixed by changing "sublime_text": "*" to "sublime_text": ">=3092"?

Yes, exactly.

Your `Comments.tmPreferences` is different indeed, however it defines the same keys twice. You should be appending `_2` to the second set of comment markers ([docs](http://docs.sublimetext.info/en/latest/reference/comments.html#shellvariables-subelements)) and adjust the `text.hml` selector to only match for your syntax, i.e. `text.html.weo`.

damky:
updated the repo file as well as made the changes you suggested for the comments file.
Thanks!

FichteFoll:
All right, now you just need to push a new tag to the repo with all the recent commits.

damky:
I think I've done it... https://github.com/damky/WEOML/releases/tag/v0.1.3-alpha
(I'm not sure if I understand your instruction.) Pushing a tag is the same as creating a release version? I don't push a tag for the package control repo, but for my repo, right?

FichteFoll:
You push a tag to your package's repo. Package Control will periodically scan your repo for tags and pick those up as releases.

Note that pre-releases aren't offered for installation by default and will have to be enabled by a user in order to have PC offer those to them (refer to PC's settings).

PR #6620, opened 2017-09-24
added pane navigation package for st3

logileifs:
Adds icon themes to Sublime Linter's gutter themes. This is my first ever pull request, please let me know if I did anything incorrectly -- apologies in advance if I did :) 

Thanks!!
Jared

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -207,6 +207,16 @@
 			]
 		},
 		{
+			"name": "Pane Navigation ST3",
+			"details": "https://github.com/logileifs/SublimePaneNavigation",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"branch": "master"
+				}
+			]
+		},
+		{
 			"name": "Pane Resizer",
 			"details": "https://github.com/markalfred/sublime-pane-resizer",
 			"releases": [
--- a/repository/p.json
+++ b/repository/p.json
@@ -212,7 +212,8 @@
 			"releases": [
 				{
 					"sublime_text": ">=3000",
-					"branch": "master"
+					"branch": "master",
+					"tags": true
 				}
 			]
 		},


FichteFoll:
So, how is this different from https://packagecontrol.io/packages/Pane%20Navigation? It doesn't have any new commits.

It also doesn't match your comment in the original post at all.

logileifs:
the original one does not work with ST3, is abandoned and the original maintainer has not responded to any requests from other developers to transfer ownership to someone else. therefore I forked it, and wanted to make it available for ST3. of course ideally I would update the original one to work for ST3 but unfortunately I have no access to that code

logileifs:
Will it not be merged while there are conflicts with the default key bindings? Because that's kind of the point with this package

FichteFoll:
Overriding default bindings is okay if a package is clearly intended as a replacement of the default functionality (and this is indicated in the readme).

I would rather prefer to replace the current package for ST2 with one that works on ST3, since ST2 is deprecated, however your repository *still* does not contain any commit that the fork origin does not. I don't see any point in adding two exactly identical packages or replacing one with the other.

Besides, you need to remove the `"branch": "master"` line from this PR and push a semantic version tag to your repo in order to make a release (once you pushed some commits).

logileifs:
You are absolutely right, my fork does not contain anything different than the original repository, but like I tried to explain before, this package can not be installed using ST3 and the original one has been abandoned. I'm not going to submit a package to Package Control that references an abandoned github repository. I am completely committed to maintain this package and fix any issues that come up but first I have to make it available in Package Control

logileifs:
For further clarification: https://github.com/borist/SublimePaneNavigation/issues/4

FichteFoll:
Ah, I see. So the package does in fact support ST3 but it's not available for installation there because it was never flagged as such.

In that case I would rather either flag the original one as compatible (just a matter of replacing `"sublime_text": "<3000",` with `"sublime_text": "*",`) or have you replace the entire package since it is seemingly abandoned. I'm fine with either. The latter would require conset from the original author or no response for 14 days, after which we could proceed with the takeover.

FichteFoll:
@logileifs could you decide on either of the two options I provided? Otherwise I'll go ahead and make the change for option 1 myself.

logileifs:
you could change the original one so it is available for ST3 but I was offering to maintain it so users wouldn't run into problems installing an unmaintained package. So I guess I would prefer option 2

FichteFoll:
Fine with me. Please go ahead and make the changes in this PR.

logileifs:
@FichteFoll which changes would that be exactly?

FichteFoll:
1. Remove your addition.
1. Replace the `details` url of the original with your url.
1. Replace `<=3000` with `*`.
1. Replace `"branch": "master"` with `"tags": true`.
1. Push a tag named `v1.0.0` to your repo.

evandrocoan:
@logileifs ping.

FichteFoll:
Note to self to flag original package as st3-comaptible on timeout.

PR #6624, opened 2017-09-27
Add SetWindowTitle plugin

gwenzek:
This plugin allows you to configure the title of your Sublime Text windows. Every time you switch view, save a view, start editing a view, it will update the window title.

Only works for Linux for now.

https://github.com/gwenzek/SublimeSetWindowTitle

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1177,6 +1177,7 @@
 			"author": "gwenzek",
 			"releases": [
 				{
+					"platforms": "linux",
 					"sublime_text": ">=3000",
 					"tags": true
 				}
--- a/repository/s.json
+++ b/repository/s.json
@@ -1171,6 +1171,18 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
You can determine the package's installed name via the module-global `__package__` attribute (or similarly `__name__`).

FichteFoll:
Besides, the package needs `.no-sublime-package` because it is supplying an explicit path to a file within the package to `sh`. (Alternatively, mirror the script in `sublime.cache_path()`.)

gwenzek:
Thanks for the tips.

I choose to add a `.no-sublime-package` file so the `.sh` script is visible to users and they can check that it's not doing something spurious.

FichteFoll:
Oh, and since the package only works on Linux you should limit platform availability.

gwenzek:
Like that ?

PR #6628, opened 2017-09-30
Add pqmarkup package

alextretyak:


alextretyak:
Unfortunately I can not easily get rid of this 2 warnings as Ctrl+9 and Ctrl+0 also already defined in default bindings, and Win+9/Win+0 overriden by Windows (this keys switch task to corresponding window in the taskbar).
Can you allow this 2 key bindings overrides just for .pq files?
(Anyway, I suppose, that commands "select_by_index", "args": { "index": 8 } and "select_by_index", "args": { "index": 9 } are quite rarely used.)

FichteFoll:
I am okay with conditionally overriding alt+9 and alt+0 for .pq files when it's documented in the readme like it is now.

After all, warnings are not errors 😉

PR #6633, opened 2017-09-30
Add GitHub 3 Color Theme

sergeymakinen:
Hi!

I would like to share my GitHub like color theme. I use it all the time and it's crafted with love. :)

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -545,6 +545,16 @@
 			]
 		},
 		{
+			"name": "GitHub 3 Color Theme",
+			"details": "https://github.com/sergeymakinen/sublime-github-theme",
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
Could you outline how this is different from https://packagecontrol.io/packages/Github%20Color%20Theme (and https://packagecontrol.io/packages/Inspired%20GitHub%20Color%20Scheme also)?

Besides, this is a color scheme and not a theme in Sublime Text terminology, so a name change to reflect that would be apprecitated. And finally, what does the 3 in "Github 3" refer to?

sergeymakinen:
nvm, the theme/scheme can be installed from a TextMate's theme repo and that's okay for now.

PR #6634, opened 2017-10-01
Added ST3 syntax for Active4D

aparajita:
There's an ancient Active4D syntax for ST2, this one is totally rewritten for ST3.

### DIFF
--- a/channel.json
+++ b/channel.json
@@ -13,7 +13,6 @@
 		"https://raw.githubusercontent.com/ajryan/CSharpreter/master/packages.json",
 		"https://raw.githubusercontent.com/amazedkoumei/SublimeKnifeSolo/master/packages.json",
 		"https://raw.githubusercontent.com/Andr3as/Sublime-SurroundWith/master/packages.json",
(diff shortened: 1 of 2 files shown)


FichteFoll:
How about providing two different channels for this package, one targeting ST3 and the other targeting older versions, notably ST2? The ST2 package is yours anyway. You can specify a different base url for a different release with the `base` key. See here:

https://github.com/wbond/package_control/blob/ccae5b79d9ef60a645b1bab2988072095f21faee/example-repository.json#L322-L338

Note that the ST2 version is specified in `channel.json` and should be removed there if you end up choosing to combine the two repos to one "package".

 https://github.com/wbond/package_control_channel/blob/810565dcd85a27ee43667bc28f2fbe435ad96645/channel.json#L16

aparajita:
@FichteFoll Is the new version what you had in mind?

FichteFoll:
Yes, that looks good. You should, however, push a version higher than `1.0.2` to the new repo since that is the old repo's latest version.

aparajita:
Done, thanks.

PR #6638, opened 2017-10-03
Add Translate-CN plugin

mjd507:
A translate Plugin for Sublime Text 3, for Chinese


FichteFoll:
Maybe you should mention what service you are using in the package name or at least the description (which is fetched from your repository's).

mjd507:
Thanks for reminding, I have changed my repository's description

PR #6641, opened 2017-10-04
Aded new package PHPUnit Linux Lite

ainme:
Hi,

I have seen a lot of phpunit packages here and that works pretty well. But additionally, this package will allow users to be able to run tests at the method level, uses the default terminal and shortcuts are ordered in a way that can be found in one place. I have been using this package for some time and also has shared with some of my colleagues to install it manually. The [original package](https://github.com/adamwathan/sublime-phpunit) which is being used by multiple users supports Mac only, so I thought of making a fork of it available for Linux platforms.
I hope it would be a great help for the Linux laravel community, which is a lot.
from original package by adam: "Convenient Sublime Text commands for running your PHPUnit tests. Scans up the directory tree to find the closest phpunit.xml file and runs phpunit from there. If it can't find one, it just runs phpunit from /."

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -1311,6 +1311,17 @@
 			]
 		},
 		{
+			"name": "PHPUnit Linux Lite",
+			"details": "https://github.com/ainme/sublime-phpunit-ubuntu",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": ["linux"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "PHPUnit Snippets",
 			"details": "https://github.com/florianeckerstorfer/fe-sublime-phpunit",
 			"labels": ["snippets"],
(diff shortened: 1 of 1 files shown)


ainme:
So? will it merged? :)

FichteFoll:
Have you considered submitting your additions as a pull request to the original? I'm sure the author(s) would appreciate that, seeing that the package is still maintained. If they don't, we can still add it as a separate package.

cc @davidhemphill, @adamwathan

FichteFoll:
@ainme ping

gerardroche:
There's also mu plugin PHPUnitKit: 

https://github.com/gerardroche/sublime-phpunit

I may be willing to implement whatever features are missing. 

It looks like the only feature missing is the ability to run the tests in the terminal rather the current functionality, which is to run the tests within Sublime Text in an output panel?

ainme:
@gerardroche cool :)

PR #6644, opened 2017-10-05
Add CopyFilenameWithLineNumbers plugin

maxx27:

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3146,6 +3146,17 @@
 			]
 		},
 		{
+			"name": "Copy Filename with Line Numbers",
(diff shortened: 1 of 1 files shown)


maxx27:
I have added the tag v1.0.0

FichteFoll:
You mention quite a few similar packages in your readme, out of which two are already installable via Package Control. Could you outline the motivation behind this package and how it is different to these other packages?

FichteFoll:
@maxx27 ping

maxx27:
> You mention quite a few similar packages in your readme, out of which two are already installable via Package Control. Could you outline the motivation behind this package and how it is different to these other packages?

@FichteFoll : Yes, I will update README with differences with them.
Sorry for delay.

PR #6645, opened 2017-10-05
added EnvSwitcher package info

SaschaMzH:


SaschaMzH:
I added now a tag. Thanks

PR #6651, opened 2017-10-07
Added Sheller package

bantya:
This is a simple package for working with the shell from within Sublime.
Some similar packages exist out there, but this is different in a way that except keybinding it requires almost zero configuration. It also provides facility to open current file directory or project root directory. Also if required it also opens the shell window at specified locations.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1307,6 +1307,19 @@
 			]
 		},
 		{
+			"name": "Sheller",
+			"description": "A brand new package to work with Command Prompt without using Command Prompt.",
+			"details": "https://github.com/bantya/Sheller",
+			"issues": "https://github.com/bantya/Sheller/issues",
+			"labels": ["shell", "cmd", "directory"],
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
You have some hard-coded backslashes in the code base, which imply Windows-specifity. Additionally some `start cmd` command line stuff that is only valid on Windows as well. Either the package is for Windows only, which you should reflect in the pltaform restrictions, or it should be patched to work for all platforms.

PS: `if not os.path.isfile("%s" % file_name):` is quite interesting 😉

FichteFoll:
@bantya ping

bantya:
> You have some hard-coded backslashes in the code base, which imply Windows-specifity. Additionally some `start cmd` command line stuff that is only valid on Windows as well. Either the package is for Windows only, which you should reflect in the platform restrictions, or it should be patched to work for all platforms.
> 
> PS: `if not os.path.isfile("%s" % file_name):` is quite interesting 😉

Was just getting started with Python. 😄

PR #6652, opened 2017-10-08
Update GDL language package

runxel:
Update the package so users get appropiate packages according to their version of ST.


FichteFoll:
You cannot downgrade from tag-based to branch-based releases. You must use tag prefixes instead, i.e. specify `"tags": "st2-"` and push a tag named `st2-1.3.3` (something that's higher than the latest) from that st2 branch.

runxel:
Appreciate your help, thanks alot!  
Committed a correction and made a new tag on the st2 branch.

PR #6654, opened 2017-10-09
Combine Purescript and Purescript IDE

b123400:
Hello. It has come to a conclusion that the author of `Purescript` would like me to takeover the package, combining my `Purescript IDE` package.

This PR combines them.

There is one thing I am not sure though, the original `Purescript` package is not using tag for versioning, it has a version of `2015.06.16.21.00.00` on package control's page. The new combined package (if following semvar) should be `2.0.0`. Will that work?

If not then I have to make it something like `2017.10.09.21.00.00`?

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -2266,19 +2266,9 @@
 		},
 		{
 			"name": "PureScript",
-			"details": "https://github.com/joneshf/sublime-purescript",
-			"labels": ["language syntax"],
-			"releases": [
-				{
-					"sublime_text": "*",
-					"branch": "master"
-				}
-			]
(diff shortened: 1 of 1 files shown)


joneshf:
I approve this move, 100%!

FichteFoll:
Any tag-based version (other than `0.0.0`, probably) will be higher than the date-based version from a commit, so this is fine.

You'll probably want to add an update message for the users that are being migrated, which according to current Package Control stats will be the overwhelming majority.

b123400:
Thx, I just added messages for 2.0.0

PR #6659, opened 2017-10-11
Added Python pexpect dependency

varp:


varp:
@wbond @FichteFoll hey guys. Are there some problems with merging this?

varp:
@wbond please review this

FichteFoll:
https://pypi.python.org/pypi/pexpect doesn't mention support for Python 2.6, which is used by ST2. Did you check whether it works on ST2? If not, just restrict the dependency to ST3 only (and rename the `all` folder to `st3`).

varp:
You were right. The pexpect could not work with ST2. I've made required change in both repos.

PR #6662, opened 2017-10-13
Add snippet FlaskHello

tanrax:
Flask modern snippet for Sublime Text 3

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -971,6 +971,17 @@
 			]
 		},
 		{
+			"name": "Flask Hello",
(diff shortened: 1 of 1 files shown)


FichteFoll:
In addition to fixing the lack of a semantic version tag, you'll also have to remove the `.sublime-package` file from the repo, since it doesn't belong there. Package Control downloads a snapshot of the repository's source files, based on the latest versioned tag it finds.

tanrax:
I think it is ready.

PR #6663, opened 2017-10-13
Include Fuse Snippets package, handy snippets for fusetools

igorls:

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -1615,6 +1615,16 @@
 			]
 		},
 		{
+			"name": "Fuse Snippets",
+			"details": "https://github.com/igorls/fuse-snippets",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Maybe you want to add this to the [Fuse package](https://github.com/fusetools/Fuse.SublimePlugin) directly?

cc @knatten, @Tapped

knatten:
Thanks for the suggestion @FichteFoll! However, merging that package into ours will require us to start maintaining these snippets for the future, which we unfortunately can't prioritise.

igorls:
@FichteFoll I also agree with that, because not every user wants/need to have a lot of snippets.. Anyway that's up to the Fuse team to decide, I'm committed to maintaining compatibility to the latest fuse version ;)

FichteFoll:
Fine with me. You can always merge packages when you change your mind(s).

PR #6664, opened 2017-10-13
Add Email on Acid Extension

flatlevel:
Tests passing

This sublime extension adds sublime support for Email on Acid email development and testing.


FichteFoll:
This is not how the `details` key works. What you have there currently should go into `description`. Additionally, you should definitely provide a readme in the `readme` key. Any raw markdown (or restructured Text, probably more) will do. Within that readme, please describe what your package actually does and how users can interact with it.

Note that you need to disclose what kind of data you send to which services, if you do so. I haven't actually looked *into* the package yet.

flatlevel:
Hey @FichteFoll - We've added a readme and updated `packages.json`.

PR #6665, opened 2017-10-14
Monokai Dark added

thatal:
This is Monokai Dark color schema, and it is suitable with seti_ui user and I search for a color which provide me a dark background with Monokai original syntax higlights. then i came up with my own Monokai color schema.


FichteFoll:
Please follow the steps outlined at https://packagecontrol.io/docs/submitting_a_package and add your package to the existing `repository/m.json` file.

PR #6669, opened 2017-10-17
HighlightDuplicates repo ownership change

LordBrom:
The plugin is the same, but it has been transferred to a new owner.
Here is the pull request which discusses the ownership change: https://github.com/LordBrom/HighlightDuplicates/pull/4

Also, adding name attribute.

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -554,7 +554,7 @@
 			"previous_names": ["HighlightDuplicates"],
 			"releases": [
 				{
-					"sublime_text": "<3000",
+					"sublime_text": "*",
 					"tags": true
 				}
 			]
--- a/repository/h.json
+++ b/repository/h.json
@@ -570,7 +570,8 @@
 			]
 		},
 		{
-			"details": "https://github.com/qur2/HighlightDuplicates",
+			"name": "Highlight Duplicates",
+			"details": "https://github.com/LordBrom/HighlightDuplicates",
 			"releases": [
 				{
 					"sublime_text": "<3000",
--- a/repository/h.json
+++ b/repository/h.json
@@ -549,6 +549,16 @@
 			]
 		},
 		{
+			"name": "Highlight Duplicates",
+			"details": "https://github.com/LordBrom/HighlightDuplicates",
+			"releases": [
+				{
+					"sublime_text": "<3000",
+					"branch": "master"
+				}
+			]
+		},
+		{
 			"name": "Highlight Trailing Whitespace",
 			"details": "https://github.com/p3lim/sublime-highlight-trailing-whitespace",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Does this still only work for ST2? ST2 is pretty old, you know.

Also, may I encourage you to switch over to tag-based releases? Details can be found [here](https://github.com/wbond/package_control/blob/master/example-repository.json#L22-L36). You just need to change to `"tags": true` and push a tag like `v1.0.0` to your repo.

Finally, since you renamed the package you have to add its old name to a `previous_names` array.

LordBrom:
I have updated this plugin to work in ST3.

repository/h.json
```diff
@@ -551,10 +551,12 @@
 		{
 			"name": "Highlight Duplicates",
 			"details": "https://github.com/LordBrom/HighlightDuplicates",
+			"previous_names": ["HighlightDuplicates"],
 			"releases": [
 				{
 					"sublime_text": "<3000",
-					"branch": "master"
+					"branch": "master",
```
FichteFoll: You must remove this line.

FichteFoll:
You may not use `sublime.load_settings` in the module root because it is not guaranteed to be available at that time. See the bottom of this page: https://www.sublimetext.com/docs/3/porting_guide.html. Additionally, you should consider reading the settings repeatedly and not just once since they may get updated by the user during runtime and not having to restart ST to make your plugin pick up the new settings is always nice. On the other hand, it doesn't make a whole lot of sense to have an "enabled" switch in the settings since the user might as well ignore (and effectively disable) the entire package instead, in which case the plugin would be unloaded and a `plugin_unloaded` function in your module would be called.

LordBrom:
I have removed the usage of the sublime module from the root. 
The settings get loaded each time they are checked. 
And using ToggleHighlightDuplicatesCommand will actually update the users setting.

Thank you for your feedback, and patience during this transfer/update.

LordBrom:
Hey, 

I was wondering if there was anything further I needed to do to get this merged. It seems to be working fine in ST3, and I have addressed all the notes made here.

I don't mean to be annoying, this is my first attempt to get something in the package control and I'm just making sure this hasn't fallen through the cracks.

LordBrom:
I think this has fallen through the cracks. :'C

FichteFoll:
You must remove `package-metadata.json` from the repo. It's a file that is auto-generated by Package Control and not supposed to be provided by hand.

Also remember to push a new tag with the latest commit to signal a release.

LordBrom:
Ah, I see. That makes sense.
I have removed the file, and created a new release with all the updates.

PR #6670, opened 2017-10-18
Replace fish-shell

Phidica:
This package provides syntax highlighting for the friendly interactive shell (fish)

The [existing package for fish](https://packagecontrol.io/packages/fish-shell) by toru-hamaguchi is abandonware, and though my package was originally forked from that repo I have kept only the snippets and completely rewritten the syntax in an entirely new (hence "improved") way. I did initially try to have my ideas merged into toru-hamaguchi's repo, but never received a response

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -895,6 +895,17 @@
 			]
 		},
 		{
+			"name": "FishShellImproved",
+			"details": "https://github.com/Phidica/sublime-fish-shell-improved",
+			"labels": ["language syntax", "snippets"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "FishTank",
 			"details": "https://github.com/ahuff44/sublime-fish-tank",
 			"releases": [
--- a/repository/f.json
+++ b/repository/f.json
@@ -885,18 +885,9 @@
 			]
 		},
 		{
-			"name": "fish-shell",
-			"details": "https://github.com/toru-hamaguchi/sublime-fish-shell",
-			"releases": [
-				{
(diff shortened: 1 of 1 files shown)


Phidica:
Whoops, I expected to have a little longer to write the tag :/ Sorry, I have one now

rwols:
I vote to *replace* the old fish package with this one, it's *way* better (and not abandoned).

Phidica:
I'm open to this suggestion :) If that were the case, could the package also be renamed to simply "fish"? This is obviously more in alignment with how other language packages are named. The redundant acronym of "fish shell" has bothered me since I forked the original repo (though I am aware that the fish repo is called "fish-shell", as is the website, for reasons I can't fathom)

FichteFoll:
Replacing and renaming the old package is perfectly fine in general, unless the original owner objects.

@toru-hamaguchi, if you want to object to @Phidica replacing your "fish-shell" package on Package Control, please comment within the next 14 days.

Phidica:
Hey @FichteFoll is this all okay?

FichteFoll:
Looks good to me.

You may want to provide an update message to inform users of the current fish-shell package of the probably huge changes and owner change. Refer to https://packagecontrol.io/docs/messaging for help.

Phidica:
That's a good idea, I'll look into it. I presume that all update messages would trigger in the event of an upgraded package switching from branch releases to tag releases?

FichteFoll:
Kind of. PC will show messages for all versions between the source and the target version of the update, so you only want to define one update message for the latest version, which is then shown to all previous users on update.

Phidica:
I've prepared an update message, so I am happy for this to be merged now

PR #6674, opened 2017-10-19
add repo simpleifdef

rumly111:

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1659,6 +1659,16 @@
 			]
 		},
 		{
+			"name": "Simple Ifdef",
+			"details": "https://github.com/rumly111/simpleifdef",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Simple Import",
 			"details": "https://github.com/vini175pa/sublime-simple-import",
 			"releases": [


rumly111:
added tag 0.1.0

rwols:
Could you add a .gitattributes file to your repo to exclude the .gif from export to package control? [Look at how GitGutter does it](https://github.com/jisaacks/GitGutter/blob/master/.gitattributes).

rumly111:
added .gitattributes
also finally learned how to add tags. Turns out I need git push --tags
Should I use v0.1.1 or 0.1.1 tag ? For now I made both :-/

rwols:
Just leave out the `v` and use `<major_number>.<minor_number>.<patch_number>` for the tag. I know it's confusing, I had that same question. The `v` prefix works fine though, it's just that it is not necessary.

I took a closer look at your plugin and I noticed you attempt to find a preprocessor token. Have you considered using something like this?

```
regions = view.find_by_selector("meta.preprocessor keyword.control.import")
for region in regions:
    candidate = view.substr(region)
    if candidate.endswith("if"):
        print(region, candidate)
```

rumly111:
It's the first ST plugin I wrote, so I didn't know about find_by_selector . I was reading docs from here https://www.sublimetext.com/docs/3/api_reference.html , but IMO the documentation isn't satisfying. For example, what are correct values for selectors? Also, in add_regions(...) I used "comment" scope, but what other types can I use? I couldn't find much info. 

Anyway, I might switch to find_by_selector in v0.2.0. For now it works fast enough, there is only 1 regex, but even that can be replaced with string comparison.

rwols:
> It's the first ST plugin I wrote

That's okay, we're all here to learn 👍 Also, I have some interest in this plugin so I'm trying to guide it on the right path.

> what are correct values for selectors?

When you click on Tools -> Developer -> Show Scope Name (or use the hotkey), ST will show the current "scope" at the cursor. A scope is just a list of strings, separated by whitespace. It is these strings that you can use for selectors, or parts of them.

> in add_regions(...) I used "comment" scope, but what other types can I use?

I think anything listed [here](https://www.sublimetext.com/docs/3/scope_naming.html) should work, but I'm not sure about that.

> I couldn't find much info.

Yes, some parts of the documentation could be improved I agree.

rumly111:
I'm working on 0.2.0 now, using find_by_selector. Will release soon.
About the plugin yes, I saw this functionality in Lazarus, and wanted the same in ST.

rumly111:
Version 0.2.0 is out. 

I needed to use some workarounds, because Region class doesn't have __hash__ method (which is strange), so I couldn't use Region as dictionary key etc. I am extracting r.a and r.b for that purpose.

rumly111:
0.3.0 adds error checking

FichteFoll:
Regions not being hashable is most likely just an oversight but understandable from the implementation. Ideally it would subclass `tuple` though. Anyway, extracting the .a and .b points to construct a tuple on your own is the proper way to go around this.

PR #6679, opened 2017-10-20
Add Z

externl:
The Z plugin adds z directory jumping to Sublime Text; quickly open directories that you frequent on the command line.

See
* https://github.com/externl/sublime-z
* https://github.com/rupa/z


FichteFoll:
Looks interesting. I'll use that.

externl:
Thanks @FichteFoll! Any feedback/comments would be great.

PR #6681, opened 2017-10-21
Add Language-Spanish to channel.json

CAERMALO:
User can use dictionary in Sublime Text for spell check.


CAERMALO:
Fixed the semver tags

PR #6689, opened 2017-10-23
add appcloud plugin

laoyao:


laoyao:
why do my commit has not approved?

PR #6692, opened 2017-10-25
Add Edge Theme

tricinel:
This is a new theme for Sublime Text 3 using colors inspired from nature, real photos taken by myself. It comes in three variants:

* **Night Sky** using colors from a series of photos of a clear night sky
* **Ocean** using colors from from photos taken by the river Rhine in Dusseldorf, Germany
* **Serene** using colors from from photos taken from parks in autumn

All photos have been taken in Dusseldorf, Germany, spanning about 2 weeks of work.

The theme features custom-designed icons for all three variants and a couple of sensible settings for fonts and icons, as well as easy install and configuration using the Command Palette.

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -234,6 +234,18 @@
 			]
 		},
 		{
+			"name": "Edge Theme",
+			"details": "https://github.com/tricinel/edge-theme-sublime-text-3",
+			"author": "Bogdan Lazar",
+			"labels": ["theme", "color scheme"],
+			"releases": [
+				{
+					"sublime_text": ">=3103",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Note that you use the `mdpopups` dependency but don't specify it as being required. It's not a hard dependency, so I'll merge either way, but you should either catch the case when mdpopups is not installed or request it to be installed yourself.

tricinel:
@FichteFoll thanks for the heads up! I will add it to the dependencies file (which I'm missing at the moment).

Also thank you for the support and merge!

PR #6696, opened 2017-10-27
Added YouCompleteMe plugin.

OmgLag:
A YouCompleteMe client for ST3.

### DIFF
--- a/repository/y.json
+++ b/repository/y.json
@@ -207,6 +207,17 @@
 			]
 		},
 		{
+			"name": "YouCompleteMe",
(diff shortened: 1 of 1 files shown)


jrappen:
@OmgLag Is there a reason for the `.no-sublime-package` file?

OmgLag:
@jrappen It was copied over from similar projects. I've removed it, as it should not be necessary.

FichteFoll:
You don't really want your test file to be loaded by ST since it doesn't interface with it at all. I suggest moving it to a `tests` subfolder.

PR #6699, opened 2017-10-29
Add type language syntax highlighting

goodmind:

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -3070,6 +3070,17 @@
 			]
 		},
 		{
+			"name": "Type Language Syntax",
+			"details": "https://github.com/goodmind/language-typelanguage-sublime",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "TypeScript",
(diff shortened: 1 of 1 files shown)


goodmind:
Anyone here?

jrappen:
@goodmind If you're targeting ST3, maybe you could bump the requirements to ST3092+ (and update the requested changes here) and provide a `*.sublime-syntax` file? I see you already have a YAML file for PackageDev you're working with. Compare <https://www.sublimetext.com/docs/3/syntax.html>. I also recommend adding test files, see the "testing" section on that page for more.

goodmind:
@jrappen can i convert TextMate to `*.sublime-syntax` automatically?

jrappen:
@goodmind Try opening the syntax file, then choose `Tools > Developer > New .... syntax` from the menu or whatever it was called. Can't check right now, but should be there. Alternatively, you can open a file and set that syntax, then the same from the menu.

FichteFoll:
Unintuitively, you need to be *using* the syntax file you want to convert instead of having it open with that command.

I'll wait for you to switch to the new format and update ST build requirements before merging.

goodmind:
@FichteFoll updated package

PR #6700, opened 2017-10-30
Update CodeFormatter config

akalongman:
Allow separate development of package for ST2 and ST3


FichteFoll:
You cannot switch from tag-based releases to branch-based. You must instead push a prefixed tab, e.g. `st2-1.2.3` from your st2 branch to the repo and specify this here as `"tags": "st2-"`.

akalongman:
@FichteFoll is it OK now?

PR #6701, opened 2017-10-30
Add TM1 Planning Analytics Developer Tools

ajmyers:
https://github.com/ajmyers/Sublime-TM1


repository/dependencies.json
```diff
@@ -804,6 +804,20 @@
 			]
 		},
 		{
+			"name": "TM1py",
+			"load_order": "99",
+			"description": "Python module for interfacing with IBM TM1 Planning Analytics",
+			"author": "MariusWirtz / ajmyers",
```
jrappen: You can write `"author": ["MariusWirtz", "ajmyers"],` here.

FichteFoll:
1. You can use relative imports to reach the [prettytable](https://github.com/ajmyers/Sublime-TM1/blob/master/include/prettytable/prettytable.py) module, which doesn't require messing with `sys.path`. `__file__` is also unreliable for this inside a .sublime-package archive.
2. Why do you intregrate https://github.com/bordaigorl/sublime-project-completions as well? That already exists as a separate package on Package Control.
3. You don't list your custom dependency in your `dependencies.json` file. You also don't support ST2, so defining dependencies for that seems redundant as well.

ajmyers:
Thanks for the feedback. I'll update and resubmit.

ajmyers:
@FichteFoll I have made the relevant updates. Please let me know if there is anything else you think should be changed.

PR #6704, opened 2017-10-31
Add SublimeBump.

yavorsky:
A package allows to preview and manage latest versions of npm/yarn packages easily. Inspired by the vscode feature, but was rethought and enhanced.
For now, it allows:

-. Show latest released package version.
-. Show next (alpha, beta, etc) package version.
-. Compare semver versions and detect whether or not current version should be updated.
-. Format to latest or next version from the context menu.
-. Use custom registries.
-. Customize tooltip for your needs and taste.

screen shot 2017-10-31 at 18 34 07


yavorsky:
Tag added.

jrappen:
@yavorsky Nice package. I can recommend the [`mdpopups`](http://facelessuser.github.io/sublime-markdown-popups/) (ST 3124+) dependency when providing tooltips.

yavorsky:
@jrappen Oh, thanks so much!

yavorsky:
@jrappen Do you know average time for PR to be approved? :)

jrappen:
As soon as @FichteFoll finds some time.

FichteFoll:
1. Your package name violates our naming guidelines. Please remove the "Sublime" part.
2. It is usually a good idea to bundle third-party libraries that you depend on in a subfolder instead of in the package root because ST will scan every .py file in the root for commands or event listeners, which these don't provide (e.g. `semver.py`). This also makes it more clear for developers which files are actually entry points for ST.
3. Have you considered using phantoms to display the "popup" for all libraries at once instead of only on hover?
4. Using `mdpopups` is overkill for a small task like this, imho.

yavorsky:
@FichteFoll Thanks so much! Going to make some fixes according to the notes.

yavorsky:
> Have you considered using phantoms to display the "popup" for all libraries at once instead of only on hover

I think it's worth to implement this in the next update after some discussion with users (if someone will be interested).
Other issues were fixed.

FichteFoll:
`ctrl+j` is a default binding (and one that I use very frequently, even). You also need to push a new tag in order to make a new release.

yavorsky:
Thanks! Updated to be `ctrl + b` and published tag 0.0.2. 👍

FichteFoll:
`ctrl+b` is even *worse* since that is used for building. Do you even check the default bindings before making changes?

If you can't find a reasonable default binding, it might make sense to not provide any at all and have the user add them themselves when they need to. You already provide a commands file, so the commands should be discoverable.

yavorsky:
@FichteFoll I'm sorry for incomprehension, it's my first plugin and I'm not very familiar with sublime's plugin system. And thanks for your help.
I've tried FindKeyConflicts plugin and just mixed up super with ctrl keys for windows/linux.

For now, I've found good and not reserved combination and published an update under the 0.0.3 tag.
`alt+shift+n`(next) and `alt+shift+l`(latest).

Also, checked with FindKeyConflicts and no Bump's mentions there.

yavorsky:
@FichteFoll Thanks so much for your work!

PR #6705, opened 2017-10-31
Add NaNoWriMo Word Count Updater

Nanofus:
This package adds a way to update the user's word count for National Novel Writing Month (https://nanowrimo.org/) from inside Sublime Text.

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -34,6 +34,16 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


jrappen:
@Nanofus Your requested changes here make the package available for all Sublime Text versions, but your README says it's a Sublime Text 3 package. If the README is correct, please update your requested changes here.

PR #6707, opened 2017-11-02
Added cel plugin https://github.com/cel-lang/sublime-cel

michael-michalski:

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -404,6 +404,18 @@
 			]
 		},
 		{
+			"name": "Cel",
+			"details": "https://github.com/cel-lang/sublime-cel",
+			"description": "The official sublime plugin for the Cel language",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Center Comment",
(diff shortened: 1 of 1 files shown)


jrappen:
@michael-michalski 

* Build files have the extension `*.sublime-build`
* You need to [`git tag`](https://git-scm.com/docs/git-tag) your commits because in the changes requested here you set `"tags": true`
* `*.tmLanguage`:
  * Sublime Text ignores the doc-type line
  * As you're not accessing the file via the UUID from the Python file, you can safely remove that, too.

michael-michalski:
I have fixed the changes you requested. :)

FichteFoll:
You should remove your `ExampleCommand` as it does nothing (and remember that all plugins share the same namespace for commands). Actually, your whole `cel.py` is a mess and does practically nothing besides print some stuff on the console and generate an exception.

Additionally, semantic versions (for your tags) consist of three numeric identifiers and I have no idea why travis ci didn't run on this. 😕

michael-michalski:
I know that the functionality is boilerplate currently, thanks for noticing. I wanted it setup, so i could have rolling releases, but okay i will recreate a pull request later. 
I guess you could have said it in a friendlier way, have a nice day

PR #6708, opened 2017-11-03
Add RevealCommand

marinintim:
This is a simple plugin that adds two commands to Command Palette.

One minor point is that `Reveal: in Finder` despite it's name should
work everywhere, because it uses `open_dir` under the hood.

Package: https://github.com/marinintim/RevealCommand

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -1257,6 +1257,16 @@
 			]
 		},
 		{
+			"name": "RevealCommand",
+			"details": "https://github.com/marinintim/RevealCommand",
+			"releases": [
+				{
+					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


marinintim:
@packagecontrol-bot: could you recheck?

FichteFoll:
Instead of going through your own plugin's commands, you could just run the commands from the commands file directly since you don't do anything else in that plugin.

FichteFoll:
Again, I suggest you to specify the commands to run in the `.sublime-commands` file directly and removing the plugin as it is entirely redundant. The package works as-is, but it'd be better if it didn't unnecessarily slow down plugin loading.

PR #6709, opened 2017-11-04
Add old release version of JSX for Sublime Text 2

allanhortle:
I'm screwed up and broke this package for Sublime Text 2 users:
https://github.com/allanhortle/JSX/issues/21

Will this add a fixed older release for users who are still using ST2, while still giving new versions for ST3?

I couldn't find anything in the docs that talked about specific tags.

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -1149,7 +1149,7 @@
 			"releases": [
 				{
 					"sublime_text": "<3000",
-					"tags": "v2.0.2"
+					"tags": "st2-v"
 				},
 				{
 					"sublime_text": ">3000",
--- a/repository/j.json
+++ b/repository/j.json
@@ -1148,7 +1148,11 @@
 			"labels": ["language syntax"],
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": "<3000",
+					"tags": "v2.0.2"
+				},
+				{
+					"sublime_text": ">3000",
 					"tags": true
 				}
 			]


jrappen:
@allanhortle You can use tag prefixes, random example via `BracketHighlighter` below:

https://github.com/wbond/package_control_channel/blob/414c035a16fb40d4baa657d685f9ef756a7d3fa4/repository/b.json#L1153-L1165

For comparison:

https://github.com/facelessuser/BracketHighlighter/tags

allanhortle:
@jrappen is it possible to just link to a single tag? I hadn't planned to keep publishing updates for ST2 as the original .tm-theme had some rather fundamental issues.

FichteFoll:
It is possible in theory, but not here in the default channel because we don't allow sticking versions here. Just push a tag named `st2-v2.0.2` and specify `"tags": "st2-v"` for the release channel.

Also, always suggest ST users to switch to 3.0 because it's awesome 👍

allanhortle:
@FichteFoll I've made that change. Happy to merge if you are.

jrappen:
* You can test language files with Sublime Text directly:
  * add a `// SYNTAX TEST "Packages/JSX/JSX.sublime-syntax"` line at the top of your syntax test file
  * see <https://www.sublimetext.com/docs/3/syntax.html#testing>
  * compare test files on [sublimehq/Packages](https://github.com/sublimehq/Packages)
  * there are different ways to test your language file, for a list see `Tools > Build with ... > Syntax Tests ...` with your language file open
* metadata files (`*.tmPreferences`)
  * Sublime ignores the doctype lines
  * you can remove the UUIDs as your not using them (via Python to access files or whatever)
* you could add a `json` tag to the README code block

allanhortle:
Thanks @jrappen. I've fixed the bottom two and made an issue for the lack of tests. That's fore sure something I haven't had time to get to.

FichteFoll:
Remember that you have to push a `st2-v3.0.2` tag (higher than the latest for both so it gets installed) for the commit that you want to be installed for ST2.

PR #6710, opened 2017-11-04
Add GitIgnorer

ThinkChaos:
I wrote this plugin because the [existing similar plugin](https://github.com/ExplodingCabbage/sublime-gitignorer) uses a timer to check for ignored files.
This plugin uses events, and a couple tricks, to be more efficient.

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -662,6 +662,17 @@
 			]
 		},
 		{
+			"name": "Gitignorer",
+			"details": "https://github.com/ThinkChaos/sublime-gitignorer",
+			"labels": ["project", "utilities"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "GitlabIntegrate",
 			"details": "https://github.com/SnoringFrog/GitlabIntegrate",
 			"labels": ["gitlab", "issues"],


FichteFoll:
Instead of using a list comprehension and then turning the resulting list into a set, you can just use a set comprehension (talking about `find_git_repos`).

The ST API may not be available at the time your module is loaded. To ensure that the API calls work, rename your `start` function to `plugin_loaded` and remove the call at the bottom of the file to have ST notify you when it's ready.

FichteFoll:
Is there any reason why you can't add this to the existing package with a pull request?

Besides, you might want to mention that the project settings are only updated when the `.gitignore` file is edites with sublime text itself and not from the outside.

ThinkChaos:
As far as I can tell, the last activity from the owner was over a year ago.
It seems like it was abandoned to me, so I didn't want to take the time to write a PR just for it to be ignored.

Good point, I pushed a commit to improve the [README](https://github.com/ThinkChaos/sublime-gitignorer/blob/master/README.md).

ExplodingCabbage:
It's abandoned-ish. There are some serious problems with my existing plugin - performance being one of them - that cannot easily be fixed without fundamentally rearchitecting it.

One problem with updating on events instead of on a timer is that you won't catch updates that happen outside Sublime - notably including ones triggered by changing Git branch. That doesn't necessarily mean it's the worse tradeoff than my approach - especially given that I've seen my plugin completely cripple Sublime when running in a project with a huge `node_modules` folder - but it still *is* a tradeoff.

I remember needing some logic to massage paths on Windows into a Sublime-friendly format; in particular, I remember that Sublime (or perhaps only Sublime 2) would not understand an exclude pattern like `C:\path\to\file` and needed this to be written as `C/path/to/file`. This path massaging happens at https://github.com/ExplodingCabbage/sublime-gitignorer/blob/master/gitignore_plugin.py#L50-L59. I haven't gone over your code too carefully, but I don't spot anything similar in your plugin; are you sure that it works properly on Windows, and doesn't just populate the `file_exclude_patterns` with a bunch of misformatted patterns that Sublime doesn't respect?

While I guess having this behaviour as an alternative to that of my plugin is kind of a good thing - it gives users a choice between two solutions with different performance versus thoroughness tradeoffs - I feel like both of our solutions are a bit ugly. It's an indictment of my own plugin that, even though I wrote it to solve a real problem I faced, it annoyed me enough that I uninstalled it after a couple of days. Although your plugin reduces the rate at which `git clean` gets spammed, it's still gonna get called a lot, and in large projects (or more precisely, ones with massive amounts of dependencies inside the project, like in a `node_modules` folder), each call is still going to cause a noticeable lag.

I think that the ideal outcome for users would be for us to completely fix the performance issues and to have patterns update both on events *and* on a timer (so that events are responded to immediately but changes outside Sublime also get caught and handled after a few seconds) and that the only way to do that is going to be to parse the `.gitignore` files instead of relying on `git clean` to give us a full list of files to exclude. I haven't given much thought to how to translate `.gitignore` patterns to globs in the format Sublime understands, but I'm hoping that such a thing is at least possible.

If you could do *that*, I'd be happy that what you'd made was unambiguously superior to what I've currently got published, and I'd be more than willing to transfer my repo to you and let your name be listed on Package Control. There's negative value to users in keeping a competing package around if there's an unambiguously superior alternative available... but I'm not convinced that what you're offering is unambiguously superior, *yet*.

ExplodingCabbage:
> I haven't given much thought to how to translate `.gitignore` patterns to globs in the format Sublime understands

Hmm. After a little reading around and experimentation I realise that this isn't possible, at least if we want to support !-negations in `.gitignore`. Sublime offers no way to un-exclude files, nor does its glob syntax support negation within patterns of the sort described by https://stackoverflow.com/a/26648205/1709587.

It's frustrating, because having to defer to `git clean` sucks and we could avoid it if Sublime had just a *slightly* more flexible system for including/excluding files. But after confirming that my suggestion above is impossible, I don't see a better approach.

FichteFoll:
FYI, I did some experimentation with exclude patterns in this issue: https://github.com/SublimeTextIssues/Core/issues/1895#issuecomment-328544813

ThinkChaos:
> are you sure that it works properly on Windows, and doesn't just populate the file_exclude_patterns with a bunch of misformatted patterns that Sublime doesn't respect?

Honestly when I tested Windows, I didn't even check the sidebar !
I made sure there was no command prompt opening up, and that the project settings were filled.
I don't have access to a Windows PC right now, and my VM has an issue, I'll check that it actually works another time.

> each call is still going to cause a noticeable lag.

I used async events not to block the UI, so it shouldn't be too noticeable.

> I think that the ideal outcome for users would be for us to completely fix the performance issues and to have patterns update both on events and on a timer

Yeah I thought about adding a timer to my plugin, but wasn't sure if it's a good idea. That's why I added the command to trigger the plugin.  
What I'll do is add a setting to change how often the plugin runs, and anything `<= 0` disables the timer run. I think that's the best I can do.
Another thing would be FS events for `.gitignore` files, but still there's no way to catch every update as there might be parent repos (in the case of submodules), or event the global ignore file.

> having to defer to git clean sucks

Yeah I had investigated translating the git patterns into ST, but came to the same conclusions.

Anyways, we're here to discuss the package situation.  
I'd rather have my own package published, but if I have to, I guess I could make a PR on @ExplodingCabbage's repo. Only thing is, I don't want to rewrite what I already did so it would be a 2.0 release with no code copied over; which kind of defeats the purpose.  
What are your thoughts?

ThinkChaos:
I pushed [a commit](https://github.com/ThinkChaos/sublime-gitignorer/commit/45f9e2c0080a49d43ef7992dc0380755d6bf92e0) to add the timer functionality.

FichteFoll:
>Only thing is, I don't want to rewrite what I already did so it would be a 2.0 release with no code copied over; which kind of defeats the purpose.
What are your thoughts?

I don't have an issue with that. My goal is to keep the number of packages to a minimum, so that we have fewer packages with high quality rather than multiple packages doing very similar things but each with their own flaws. This situation here would be an ideal fit for a takeover of some sort. Since apparently @ExplodingCabbage's package is "abandon-ish-ware", I'd vote for your package to *replace* the current entry in the channel and potentially inform users about the update. Functionality should be largely the same, however.

@ExplodingCabbage do you agree with this?

ExplodingCabbage:
Yes.

Given that, I suggest that the following things ought to happen:

- @ThinkChaos tests on Windows in both Sublime 2 and Sublime 3 and verify that paths are being formed properly (given the recollection I mentioned above that correctly-formed paths like `C:\somefile` don't work, and need to be written as `C/somefile` instead)
- @ThinkChaos updates the README to reflect the fact that the plugin can now run on a timer like mine does (the code is already made, but the docs are behind)
- I review @ThinkChaos's version and try to poke more holes in it if I can
- @ThinkChaos updates this PR so that my package is replaced instead of the two packages coexisting
- I edit the README of my version to indicate that it is deprecated and point users towards @ThinkChaos's version
- @FichteFoll merge's this PR
- I edit my Stack Overflow answer at https://stackoverflow.com/a/26372261/1709587 (which has 10k views and probably accounts for most of the traffic to my package currently) to recommend @ThinkChaos's version, and to reflect its superior capabilities.

Does that sound about right to both of you?

ThinkChaos:
Here's what I was able to get done:

- test on Windows in Sublime 3
- fix paths on Windows ([889f5b2](https://github.com/ThinkChaos/sublime-gitignorer/commit/889f5b2c9bce1b44c17e23fa7979a3524fe5518c)), thanks for the tips
- update the README ([274ff43](https://github.com/ThinkChaos/sublime-gitignorer/commit/274ff4332275af029991e33bd51351c34f179220))

I still need to test in Sublime 2.  
I'll also want to write an update note to notify current plugin users about the change.  
After that, and @ExplodingCabbage reviews the plugin, I'll update this PR.

FichteFoll:
Thanks for keeping us updated.

As a side node, I wouldn't exactly bother with ST2 support. A month ago, less than 2% of users were still on ST2.

ThinkChaos:
> As a side node, I wouldn't exactly bother with ST2 support. A month ago, less than 2% of Package Control users were still on ST2.

Good to know, I don't think it's worth the effort then.

FichteFoll:
I'll be closing this for now to clear up the open PR list. Please comment once you have updates to share and I will re-open.

PR #6711, opened 2017-11-05
Restrict ST2 to older versions of Elm Language Support

sentience:
New versions of the Elm Language Support package include a .sublime-syntax file and therefore require ST3.

All ST2-compatible versions of Elm Language Support have been tagged with the "st2-" prefix, so that ST2 users can continue to install the older, compatible version of the package.

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -489,7 +489,11 @@
 			"previous_names": ["Elm Syntax Highlighting"],
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": "<3092",
+					"tags": "st2-"
+				},
+				{
+					"sublime_text": ">=3092",
 					"tags": true
 				}
 			]


jrappen:
@sentience `*.sublime-syntax` files require at least Build 3084 as per the changelog.

sentience:
Thanks, @jrappen. Updated.

FichteFoll:
You might want to review the necessity of `	.no-sublime-package`.

stoivo:
@FichteFoll, 
is there a simple way we can build Elm Lang Support and as a sublime-package and test if it works?

FichteFoll:
Package Control provides a command in the Command Palette to create a package of a command. Alternatively, `.sublime-package` files are just zip archives of the package files.

stoivo:
Thanks. We have tested and removed that file for the next release. :)

PR #6713, opened 2017-11-05
bundled metadata of packages by jrappen

jrappen:


FichteFoll:
1. Your package name may not start with (or contain) the word "Sublime".

2. I wouldn't store files of a Sublime Text package that directly interface with Sublime Text in a hidden folder (`.sublime`), since that's entirely not transparent, but you're obviously free to do that.

3. You should add a sentence or two regarding the motivation behind that package. Currently it says absolutely nothing about what it is.

jrappen:
@FichteFoll Package was named "Sublime Color Schemes" as it contained `*.sublime-color-schemes`. I was under the impression that it doesn't matter how I name the folders, as I expect developers to not have those hidden and it's in a zip file anyway for end users. Motivation was implied by the name of the package. Anyhow, removed it now as I'm not planning to maintain it as a public package.

FichteFoll:
Sure, how you name subfolders doesn't matter for ST itself. I was just speaking about general conventions regarding any kind of repository, with respect to potential contributors.

I understand why the package was named that way, but that's not enough reason to warrant starting with "Sublime" imho. See also https://packagecontrol.io/docs/submitting_a_package#Step_2.

PR #6715, opened 2017-11-06
Add Titan Edge Highlighter.

ErdemAkin:
It is a package developed for the php framework "Titan". "Titan" uses a template structure named "edge" in view parts. With this package, it has been tried to increase the readability of files with edge structure. The tests in step 7 were performed and no errors were found.

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -2074,7 +2074,7 @@
 		},
 		{
 			"name": "Titan Edge Highlighter",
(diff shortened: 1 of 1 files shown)


jrappen:
@ErdemAkin The changes you proposed make the package available for Sublime Text 2 as well. The `*.sublime-syntax` file format requires at least Build 3084. Also you need to [`git tag`](https://git-scm.com/docs/git-tag) your commits.

ErdemAkin:
Update messages.json

PR #6720, opened 2017-11-09
Add fast translator with minimum settings. Only for Russia.

mkdk:
This package for fast translate from "some_lang" to ru, with minimum settings.


FichteFoll:
What does this do different from https://packagecontrol.io/packages/YandexTranslate? Maybe you want to submit a pull request to that package instead?

FichteFoll:
@mkdk ping

PR #6721, opened 2017-11-09
Add "Import Cost"

julianburr:


FichteFoll:
1. You cannot use `sublime.plugin_path()` in the module scope because it may not be available at the time. You can use `__file__` instead, so long as you fix 2.
2. Because you call an external process with a file you ship in the package (`node`), you must provide a `.no-sublime-package` file to tell PC to unextract the package archive. Packages are installed as zip archives by default, which allows for overrides.
3. Your `Popen` call does not consider the Windows platform, notably regarding `env['PATH']` and that on windows you must include the extension of the program you want to execute.
4. You should note that the package requires `node` to be on the system's PATH.
5. The installed plugin's name will be available under the `__package__` variable.

julianburr:
Hey, thanks for the quick response 😊 

I changed the path issues, added a `.no-sublime-package` file and added the path to the node binary to the settings. I also specified `osx` as the platform here, cause atm I don't have any machines with the other os' to test on.

Let me know if there is anything else that needs to be changed

FichteFoll:
Looks like you forgot to remove this line:

https://github.com/julianburr/sublime-import-cost/blob/master/import-cost.py#L12

Semi-colons are optional in Python, btw (and basically never used since a newline terminates a statement).

julianburr:
Fixed in tag v0.0.3 😊

PR #6723, opened 2017-11-10
add VisualChanger plugin

kivS:
VisualChanger is a simple tool that allows you to add profiles that affect settings of the editor and its plugins. You can then change between profiles using the command palette or key bindings.

### DIFF
--- a/repository/v.json
+++ b/repository/v.json
@@ -511,7 +511,7 @@
 			"author": "Vik Borges",
 			"releases": [
 				{
-					"sublime_text": ">= 3000",
+					"sublime_text": ">=3000",
 					"tags": true
 				}
 			]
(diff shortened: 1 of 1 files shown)


FichteFoll:
Your readme states that the package is for ST3 and you're using API that isn't available on ST2, too. Please update the ST build requirements to exclude ST2.

Is there any particular reason why you are not using `sublime.load_settings` and `sublime.save_settings` and write to the User preferences file directly? I mean, you *do* use them at another point.

Instead of calling `.__contains__` directly, just use the `in` operator.

It might be sensible to be explicit about what the keys in the `plugins` mapping actually mean. Specifically, that they are not about the plugins' names but the settings files' they use. To add to that, one could use the package for syntax-specific settings as well, which the mapping's name (`plugins`) doesn't reflect.

kivS:
@FichteFoll Thank you, I'll get to it

kivS:
@FichteFoll done!

PR #6724, opened 2017-11-10
Added MapTool Syntax, a syntax highlighting for RPTools' MapTool macros.

wwmoraes:
MapTool Syntax, a syntax highlighting for RPTools' MapTool macros (.rpmacro files).


wwmoraes:
I'll fix the support for build 3092.

PR #6726, opened 2017-11-11
Add Test Switcher

davidmreed:
This package provides functionality not currently exposed by any similar package: it is cross-platform and is not specific to any language, development environment, or testing framework, but includes customizable settings for any test/source setup.

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -543,6 +543,17 @@
 			]
 		},
 		{
+			"name": "Test Switcher",
+			"details": "https://github.com/davidmreed/test_switcher",
+			"labels": ["testing"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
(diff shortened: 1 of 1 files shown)


FichteFoll:
It should be trivial to add cross-platform support to https://packagecontrol.io/packages/Test%20Jumper, if that isn't already the case, instead of duplicating this functionality across two packages.

cc @tomcharles

davidmreed:
Test Jumper does not have some of the behaviors I am interested in (walking open file trees to locate test files) and uses hard-coded POSIX path manipulation, but I am happy to contribute there as well if @tomcharles is interested in contributions or pull requests.

FichteFoll:
Test Jumper doesn't seem to be maintained anymore (no commits since its creation, basically) and @tomcharles didn't reply either. As unfortunate as it is, I think adding this package is the best option currently.

PR #6730, opened 2017-11-14
Add laravel-blade-autocomplete package

ahmedash95:
Laravel Blade autocomplete is a simple package to autocomplete blade files when using laravel blade engine

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -139,6 +139,16 @@
 			]
 		},
 		{
+			"name": "Laravel Blade AutoComplete",
+			"details": "https://github.com/ahmedash95/sublime-laravel-blade-autocomplete",
+			"releases": [
+				{
+					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


ahmedash95:
Tag added .. check again

FichteFoll:
Note that you hardcode forward slashes in your code. You should look into using `os.path.join` and replace `os.path.sep` with `"."`.

ahmedash95:
Hi @FichteFoll thanks for advice .. I did some enhancements and I think PR is ready to merge now 

Thanks

FichteFoll:
You don't need to close files that you open in a context manager (i.e. with `with`), but it doesn't hurt in this case.

Remember that you have to push a new tag with your changes to make a new release.

ahmedash95:
I drafted a new release, Thanks a lot

PR #2524, opened 2013-12-22
Test against package names containing periods

icio:
The test rejects any package name containing a period because they will error in Sublime Text 3 if they define python modules.

When the packages are compiled into sublime-package bundles any filenames containing a period confuse the Sublime Text plugin loader. This is on account of python (sub)modules being separated by a period. It attempts to, for example, `importlib.import_module('Module.app Menu')`. This maps to looking for module "Module" and "app Menu" within for which the sublime-package would be Module.sublime-package but is not where our plugin is packaged to by Package Control.

I'm not sure whether this is the right way to solve this. Arguably Package Control or Sublime should handle the scenario I'm testing against here. Some examples of troublesome packages (you'll see an ImportError in the console when you load Sublime Text 3):
- Marked.app Menu (rename PR: #2525)
- Mou.app Menu

Currently failing tests, which do not all include python modules and therefore do not experience this error:

```
$ python tests/test.py 2>&1 | grep AssertionError
AssertionError: True is not false : Package names should not contain .: 'Apiary.io Blueprint'
AssertionError: True is not false : Package names should not contain .: 'Asp.Net File Switch'
AssertionError: True is not false : Package names should not contain .: 'Backbone.Marionette'
AssertionError: True is not false : Package names should not contain .: 'Backbone.js'
AssertionError: True is not false : Package names should not contain .: 'Chaplin.js'
AssertionError: True is not false : Package names should not contain .: 'Codepad.org Paste Plugin'
AssertionError: True is not false : Package names should not contain .: 'Conky.tmLanguage'
AssertionError: True is not false : Package names should not contain .: 'Dust.js'
AssertionError: True is not false : Package names should not contain .: 'Ember.js Snippets'
AssertionError: True is not false : Package names should not contain .: 'FakeImg.pl Image Placeholder Snippet'
AssertionError: True is not false : Package names should not contain .: 'jQuery Mobile 1.4 Snippets'
AssertionError: True is not false : Package names should not contain .: 'Kohana 2.x Snippets'
AssertionError: True is not false : Package names should not contain .: 'Lazy Backbone.js'
AssertionError: True is not false : Package names should not contain .: 'Marked.app Menu'
AssertionError: True is not false : Package names should not contain .: 'Mou.app Markdown'
AssertionError: True is not false : Package names should not contain .: 'objc .strings syntax language'
AssertionError: True is not false : Package names should not contain .: 'PEG.js'
AssertionError: True is not false : Package names should not contain .: 'Paste to friendpaste.com'
[...]

PR #6734, opened 2017-11-16
Added Toggle .only package

CurtisHumphrey:
Toggles ".only" on and off of describe and it for test files in javascript

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -2257,17 +2257,6 @@
 			]
 		},
 		{
-			"name": "Toggle only for Tests",
-			"details": "https://github.com/CurtisHumphrey/sublime_toggle_only",
-			"labels": ["toggle", "test", "mocha", "javascript"],
-			"releases": [
-				{
-					"sublime_text": "*",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "Toggle Css Format",
 			"details": "https://github.com/speilberg0/SublimeToggleCssFormat",
 			"releases": [
@@ -2290,6 +2279,17 @@
 		{
 			"name": "Toggle File",
 			"details": "https://github.com/Harurow/sublime_togglefile",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
+			"name": "Toggle only for Tests",
+			"details": "https://github.com/CurtisHumphrey/sublime_toggle_only",
+			"labels": ["toggle", "test", "mocha", "javascript"],
 			"releases": [
 				{
 					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


FichteFoll:
You cannot use dots in the package's name due to how Python importing works.

CurtisHumphrey:
Are you sure @FichteFoll as I have this installed via add github source and everything has been working fine.  I think the name field was just a pretty print thing.

FichteFoll:
Yes, I am sure. #2524 covers this.

Package Control installs the package under the name specified, i.e. it either calls your package `Toggle .only.sublime-package` or extracts it to a folder named `Toggle .only` (if `.no-sublime-package` is provided, which isn't necessary here). You can try this yourself, as importing the Python plugin will fail.

CurtisHumphrey:
hmm, I see what you are getting at now.  They way I installed it, it use the repo name. Fixing it in just a moment.

FichteFoll:
You'll have to fix the package order after the name change: https://travis-ci.org/wbond/package_control_channel/builds/308559511#L675

Thom1729:
I read through the documentation and I still can't quite tell what this package is for. Some specific unit-testing framework? What framework? Is there an existing package for that framework that may provide similar functionality, or to which this functionality could be added?

You're creating a global keybinding that applies in every scope. If possible, you should add `context` conditions to the binding so that it's only available when you're using the relevant testing framework.

You're only performing the modification on the first selection. You should probably do it on every selection. (This would also handle the case where there is no selection.)

FichteFoll:
I agree with these.

>You're creating a global keybinding that applies in every scope. If possible, you should add context conditions to the binding so that it's only available when you're using the relevant testing framework.

>You're only performing the modification on the first selection. You should probably do it on every selection. (This would also handle the case where there is no selection.)

Please address this feedback before we proceed with the merge.
I don't care which testing framework this is for, but I don't recall having seen a package like this before.

CurtisHumphrey:
So If I understand correctly you would like:
- [ ] Better readme file that says what framework this is for
- [ ] Somehow add context conditions to restrict it to only certain type of files?
- [ ] Allow modification on more than one sections

I did quite a bit of looking but the other packages (ones related to Mocha or Jasmine - both JS testing) seem to all be related to either running the tests or generating templates.  This one doesn't really fit with either as it more a helper for toggle tests on and off that are auto-running in the background somewhere on a file watcher.  I'm not sure how to do the context but I can look it up.

FichteFoll:
>- Somehow add context conditions to restrict it to only certain type of files?
>- Allow modification on more than one sections

These are the important parts here. If you need any help with that, I can guide you towards documentation or an implementation.

FichteFoll:
@CurtisHumphrey ping

PR #6735, opened 2017-11-16
Add "=BB= LSL" package

buildersbrewery:
This package bundles the functionality of the packages `LSL`, `TooltipLSL` and `SublimeLinter-contrib-lslint`, differs from them however in a few key points:

* This package specifically targets the latest Sublime Text builds and their abilities.
* The binaries for building and linting are included as most end-users will have problems installing binaries in their path and/or keeping them up-to-date.
* This package's grammar file uses the new `sublime-syntax` format and I've added many test files to ensure it is accurate and up-to-date.
  * `LSL` has frequent updates, so this is necessary
* The tooltips from this package however:
  * have completely been redone to show much more information closer to what `LSP` displays for other languages
  * show code usage examples, as `LSL` in general has **a lot of** "features" that require a work-around
* Most popular indentation styles are supported by completions and snippets.

---

Package name was chosen due to how the Builder's Brewery labels its free class material and products. The Brewery is one of the biggest groups with well more than 40k members, a well respected educational institution and a top brand in Second Life.

This package is meant to support students learning to program using the `LSL` at our online school.

To differentiate from the `LSL` package, we've chosen `=BB= LSL` as the package name.

### DIFF
--- a/channel.json
+++ b/channel.json
@@ -18,6 +18,7 @@
 		"https://raw.githubusercontent.com/blachniet/sublime-mimosa/master/packages.json",
 		"https://raw.githubusercontent.com/blueplanet/sublime-text-2-octopress/master/packages.json",
 		"https://raw.githubusercontent.com/bradfeehan/SublimePHPCoverage/master/packages.json",
+		"https://raw.githubusercontent.com/buildersbrewery/sublime-lsl/master/package_control_channel.json",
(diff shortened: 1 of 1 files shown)


buildersbrewery:
`.no-sublime-package` is necessary because the package includes binaries.

FichteFoll:
As reported, you should remove `main.pyc`, though it's not significant since this is the only file. For obvious reasons, I cannot review the rest of the plugin code.

PR #6736, opened 2017-11-16
Added package DBpedia Lookup

rvolz:
This plugin was made to enable quick word lookups using the DBpedia Lookup service while checking a text, for example while perusing a Markdown or Asciidoctor document. Lookup here means looking for context, not spell checking.

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -254,6 +254,17 @@
 			]
 		},
 		{
+			"name": "DBpedia Lookup",
+			"details": "https://github.com/rvolz/sublime_dbpedia_lookup",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Do not use `logging.basicConfig`. It will affect *all* packages, since it installs a root logger. Instead, add a handler (and formatter) to your own logger that you retrieve with `logging.getLogger`.

You should ensure that the `name` you add to the GET query is properly url-escaped.

You can get infromation regarding the word you right-clicked on by requesting the entire event information on a rightclick. See `Default/symbols.py` for an example.

rvolz:
Thanks for the comments, the changes are now pushed to tag 1.0.1.

PR #6737, opened 2017-11-17
Add VEX and HScript packages

teared:
Hm, It seems like both commits got joined into a single PR, though I tried to commit and made PRs separately. Hope it is not a problem.

## VEX
Houdini add-on, VEX part of it. Main focus of this add-on shifted towards writing and building VEX Expressions, which is a special syntax on top of VEX language, popular among community this days. "Pure" VEX is also fully supported, though. Also, integrates HScript part, ~which I'll PR shortly~. Relies on new features in syntax and popups, so, does not support old Sublimes.

## HScript
HScript and HScript Expressions part of the Houdini add-on. Defines couple of another syntaxes (inside one). I decided to split it from VEX add-on, so, they can be installed separately. I develop them as a single project [here].

[here]: https://github.com/teared/packages-dev

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -909,6 +909,17 @@
 			]
 		},
 		{
+			"name": "HScript",
+			"details": "https://github.com/teared/HScript",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": ">=3126",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "HSP",
 			"details": "https://github.com/potato4d/sublime-HSP",
 			"releases": [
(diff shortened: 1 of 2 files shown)


FichteFoll:
In order to create a PR, you have to create a new branch (from upstream/master) before committing a change. Any commits in that branch not existing on the target branch (upstream/master again) are offered to be pulled. If you want to make two PRs at the same time with different contents, you need to create two branches and make commits on the respective branch. However, it wasn't a problem here since the packages are quite related.

Also, you might want to look into ensuring that the "term" you want to open in the browser is properly url-escaped.

PR #6740, opened 2017-11-19
add ABNF syntax

achun:

 - Link to your code repository: https://github.com/ZxxLang/ABNF-sublime-syntax
 - Link to the tags page with at least one [semver](http://semver.org) tag: https://github.com/ZxxLang/ABNF-sublime-syntax/releases/latest

Also make sure you:

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -2371,6 +2371,17 @@
 					"tags": true
 				}
 			]
+		},
+		{
+			"name": "ABNF-sublime-syntax",
+			"details": "https://github.com/ZxxLang/ABNF-sublime-syntax",
+			"labels": ["abnf", "language syntax"],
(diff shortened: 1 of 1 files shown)


FichteFoll:
I would use [`support.type.rule.abnf`](https://www.sublimetext.com/docs/3/scope_naming.html#support) for the commonly used upper-case rules like `CRLF` and `VCHAR` (when they are *used*, not defined) and [`entity.name.rule`](https://www.sublimetext.com/docs/3/scope_naming.html#entity) for the target of a rule definition (i.e. the variable on the left of `=`).

FichteFoll:
I'd also name the package just "ABNF", but it's too late for that now. You can however submit another PR for changing this.

PR #6749, opened 2017-11-20
add filter code (fixed conflict key)

happyqingye:

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -724,6 +724,16 @@
 			]
 		},
 		{
+			"name": "Filter Code",
+			"details": "https://github.com/happyqingye/FilterCode",
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
You should remove the mousemap file. It will override the default double-click action and annoy users.
If they want this, they can add it to their User mousemap file per instructions that you could include in the readme.

Or even better you should set the settings explained here: http://docs.sublimetext.info/en/latest/reference/build_systems/configuration.html#capturing-build-system-results. That's what the Find Results panel does internally as well.

I don't really get your gifs, but that might just be me.

happyqingye:
@FichteFoll 
Appreciate your advice. I removed the mousemap files and add the instructions in readme.md. 
Please check it again.

PR #6752, opened 2017-11-22
Add Postfixer package - postfix completions plugin for Sublime Text 3

mbnuqw:

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -1687,6 +1687,19 @@
 			]
 		},
 		{
+			"name": "Postfixer",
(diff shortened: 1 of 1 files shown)


FichteFoll:
You use Pyyaml but don't specify it as a dependency. https://github.com/packagecontrol/pyyaml/

I also suggest using relative imports when possible.

mbnuqw:
Hi, I've added dependency and changed import. Could you check it again. https://github.com/mbnuqw/sublime-postfixer

Thanks

PR #6754, opened 2017-11-24
Rename "Hide Sidebar When Not Focussed" to SideBarHider

akrabat:
This is to match the recent project rename.

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -494,16 +494,6 @@
 			]
 		},
 		{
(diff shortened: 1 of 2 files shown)


FichteFoll:
You must add the old name to a `previous_names` array of the new package entry's metadata.

akrabat:
> You must add the old name to a previous_names array of the new package entry's metadata.

Thanks. I've updated.

PR #6756, opened 2017-11-26
Elm Language Support repository change url

stoivo:
As soon as this PR is merged we will change the URL.

ref: https://github.com/elm-community/Elm.tmLanguage/issues/35

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -484,7 +484,7 @@
 		},
 		{
 			"name": "Elm Language Support",
-			"details": "https://github.com/elm-community/Elm.tmLanguage",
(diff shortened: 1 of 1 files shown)


sentience:
Actually, as the maintainer of this package, I'm going to close this PR for now. The correct order is for us to rename our repository first (at which point GitHub will redirect requests for the old name), and then update Package Control. Doing it in the opposite order will cause downtime. Will reopen this once we've renamed the repository.

sentience:
The repository has now been renamed. We would now like to update the URL in wbond/package_control_channel.

PR #6759, opened 2017-11-27
Add Monokai Dark Color Sheme

thatal:


FichteFoll:
Semantic Versions consist of exactly three numeric parts.

You also added your package to the wrong file. Rememver to insert it in order as well.

PR #6761, opened 2017-11-27
Added Particle Reverse Polish Language

KubeRoot:
A simple package containing completions, indentation rules and syntax highlighting for the PRPL language used in Particle Fleet: Emergence.

The package is similar to the Creeper Reverse Polish Language due to language similarities (both made by same developer) - however, since the CRPL package has been merged a while ago, I decided to simply PR it separately.

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -363,6 +363,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


KubeRoot:
(Probably?) Fixed tags.

PR #6768, opened 2017-11-29
Add Se7en Color Scheme

csknklc:

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -702,6 +702,17 @@
 			]
 		},
 		{
+			"name": "Se7en",
+			"details": "https://github.com/csknklc/se7en",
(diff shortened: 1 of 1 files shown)


jrappen:
@csknklc you can use [PackageDev](https://github.com/sublimetext/PackageDev) to convert the color scheme(s) to the **new** [color scheme format for ST3149+](https://www.sublimetext.com/docs/3/color_schemes.html)

On another note, the term `themes` as a subfolder in your source code and the github topics applied to your repo is wrong, as these are "color schemes" in Sublime terms.

Compare the docs:

* <https://www.sublimetext.com/docs/3/color_schemes.html>
* <https://www.sublimetext.com/docs/3/themes.html>

PR #6778, opened 2017-12-06
Add SublimeTextAPICompletions

gerardroche:
Provides Sublime Text API completion.

Other packages provide similar things, but there's none I can find that provides basic completions. This is just an initial version I'll expand on as I go.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -4340,6 +4340,17 @@
 			]
 		},
 		{
+			"name": "SublimeTextAPICompletions",
+			"details": "https://github.com/gerardroche/sublime-api-completions",
+			"labels": ["completions"],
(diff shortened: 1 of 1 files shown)


FichteFoll:
PackageDev used to provide these, but they were "disabled" by default. I found them to be quite annoying since I also use Python for non-ST purposes and a proper auto-completion package with the ST folder added to its lookup path proved to be much more convenient.

This package *does* include "SublimeText" in its name, but I don't really see a way around it. "APICompletions" is too general and it needs to be stated somehwere that this is indeed for Sublime Text's API.

For reference, a similar package would be https://packagecontrol.io/packages/Sublime%20Text%20API%20Helper.

PR #6781, opened 2017-12-07
bad charactor kicker relaxing programmer

shikaiwen:

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -86,6 +86,16 @@
 			]
 		},
 		{
+			"name": "badcharkicker",
+			"details": "https://github.com/shikaiwen/bad_char_kicker",
+			"releases": [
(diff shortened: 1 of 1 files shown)


ianharper:
The semantic versioning for your package needs to be v1.0.0 to pass this. Made the same correction for a pull request of mine.

FichteFoll:
https://packagecontrol.io/packages/Find%20Non%20ASCII%20Characters is a quite similar package, except that it supports multiple regions and you can do more with the characters than just delete them (although you have to press an additional key for that), since it selects the characters.

I'd say that is better user experience and thus an overall better package than this.

FichteFoll:
@shikaiwen do you have to say anything about my earlier comment?

PR #6782, opened 2017-12-07
add file-downloader

fcannizzaro:
`file-downloader` allow to download, browse or open in sublime url (**files**).


FichteFoll:
Is there any particular reason to use `_thread` directly instead of `threading`?

PR #6784, opened 2017-12-10
Add YardGenPlus to repository/

nimylian:
Yardgen, the package this is based on, isn't as customizable and the package is not represented on Github.

### DIFF
--- a/repository/y.json
+++ b/repository/y.json
@@ -78,6 +78,16 @@
 			]
 		},
 		{
+			"name": "YardGenPlus",
+			"details": "https://github.com/nimylian/YardGenPlus",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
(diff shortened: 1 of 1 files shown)


FichteFoll:
Your keymap change has not been included in a tag yet. You also haven't updated the OSX bindings.

I suggest using a context and checking against `selector` instead of defining a global hotkey that will only do something inside a function definition.

What are the particular differences to [the original](https://packagecontrol.io/packages/yardgen)? Why can't you create a pull request against that instead of forking? cc @fappelman (hoping he sees this)

fappelman:
I don't know about this version of Yardgen. If at some time there would be a pull request I would be happen to include the changes.

FichteFoll:
Any update on this?

nimylian:
I'll close this. I don't have the time to fight with the system.

PR #6785, opened 2017-12-10
Add VimModelines to Package Control

pestilence669:
This plugin applies basic Vim commands (tab width, auto-indent, word wrapping, line endings, ...) set with modeline annotations in the headers and footers.

There is an existing plugin for Emacs modelines and another for vim-inspired Sublime modelines, but I couldn't find anything that parses Vim syntax natively.

I have a lot of source code that has the annotation in the header and it's always been ignored. Switching between PEP-8 and non-PEP-8 Python files, was a motivator as was extending functionality that one already has using Vim with CLI edits over SSH.

### DIFF
--- a/repository/v.json
+++ b/repository/v.json
@@ -339,6 +339,18 @@
 			]
 		},
 		{
+			"name": "VimModelines",
+			"details": "https://github.com/pestilence669/VimModelines",
+			"author": "Paul Chandler",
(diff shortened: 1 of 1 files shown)


FichteFoll:
`from __future__ import print_function` is redundant if you only target Python 3.

PR #6786, opened 2017-12-10
Add Discord Rich Presence

Snazzah:
This package basically utilizes rich presence in a application called Discord. It's pretty nice to look at for everyone else using Discord.

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -728,6 +728,16 @@
 			]
 		},
 		{
+			"name": "Discord Rich Presence",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Looking good. I am curious where the icons are sourced from, however. I suppose you have to upload them to discord directly, after which they become available under the key they were registered with?

Snazzah:
@FichteFoll Yeah that's pretty much how I do it.

PR #6788, opened 2017-12-10
Changed the name.

csknklc:


FichteFoll:
It has to be an array and named `previous_names`.

PR #6792, opened 2017-12-11
Add monokai 2

dimaeromin97:
This is a new color scheme that is backward and readable

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -2031,6 +2031,17 @@
 			]
 		},
 		{
+			"name": "Monokai 2",
+			"details": "https://github.com/dimaeromin97/Monokai-2",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": ["windows", "osx", "linux"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Moo",
 			"details": "https://github.com/maltize/sublime-text-2-moo",
 			"releases": [
--- a/repository/m.json
+++ b/repository/m.json
@@ -1929,7 +1929,6 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"platforms": "*",
 					"tags": true
 				}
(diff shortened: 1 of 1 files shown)


dimaeromin97:
ок

FichteFoll:
Quite bold to claim the successor of the popular "Monokai" scheme for yourself when you aren't the author of the original one. We're not required to keep the name free of course, considering how many monokai schemes there are already, but that's just something I would consider here.

Will merge if you say you are "okay" with this.

FichteFoll:
@dimaeromin97 ping

tajmone:
> We're not required to keep the name free of course, considering how many monokai schemes there are already, but that's just something I would consider here.

I agree with you @FichteFoll, the "2" in "Monokai 2" is confusing because [original Monokai] author Wimer Hazenberg has recently created "Monokai Pro", which is also often referred to as "Monokai 2" informally:

- https://www.monokai.pro/

I actually ended up in this pull request because it popped up while I was looking for information on Monokai Pro. Of course, as you said, the name "Monokai" is now popular and has been used (and misused) in many different forms — which is generally ok, as long as Wimer Hazenberg is credited as the author of the original color scheme (even when deviating slightly from it).

Many of us have benefited from this great color scheme. I think that Hazenberg choice to make a commercial follow-up in his extremely successful color scheme is a well earned priviledge, and I wish him the best and that he might monitize on it for all the good that came from the original free version. At the same time, I think that avoiding confusion with his new Monokai successer ("Pro", or "Monokai v2" as many call it) would be courteous toward him; especially since "Monokai Pro" is also available as a Sublime Text commercial Package.

After all, we are free to use just "Monokai" in any variation of it. But as you said, there is a long history of reusing the Monokai name in various projects, so it's more a courtesy than a legal question.

[original Monokai]: https://www.monokai.nl/

PR #6794, opened 2017-12-11
Atlantis VIP added

dziaineka:
This paskage adds supporting of Atlantis VIP language

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -1842,6 +1842,17 @@
 			]
 		},
 		{
+			"name": "Atlantis VIP Language",
+			"details": "https://github.com/dziaineka/Atlantis-VIP-sublime",
+			"labels": ["Atlantis", "VIP", "language syntax", "snippets", "galaktika"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Atomic Color Scheme",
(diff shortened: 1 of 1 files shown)


dziaineka:
all done

FichteFoll:
I don't know whether this language only has a Russian target audience, but I can't read Russion and the completion hints wouldn't help me at all.
It also doesn't strike me as a good idea to define syntax-specific `fallback_encoding` in a package, if that works in the first place. Not even considering all the other settings that you define.

You should realize that syntax-specific settings override the User settings and you're messing with the user's environment. Things like `auto_complete_commit_on_tab` and `drag_text` do not belong into a package's syntax-specific settings file.
You can argue about indentation settings for languages like YAML, where tabs are invalid by the spec and 2 spaces highly encouraged, but I know nothing about this language.

dziaineka:
It's an internal language of Russian corporation "Galaktika". Some of it's software products has been developed with this language (e. g. [https://www.galaktika.ru/en/erp.html](https://www.galaktika.ru/en/erp.html)). Company is focused on Russian speaking segment of business and Russian legislation, therefore all developers are Russian speaking.

The thing is all Atlantis VIP source code has Windows 866 encoding. That's why I've defined `fallback_encoding` in a package.

I've deleted user specific settings.

2 spaces indentation settings resulted from language specification, but specification is only in Russian.

Thank you for your help!

FichteFoll:
I see, thanks for the explanations.

PR #6795, opened 2017-12-12
added simplesyntax

budRich:
syntax highlighting for config files, with symbols and comments. There are probably similar packages, but i don't think any uses the symbols in the way I do here, and it's also a good general package that covers different types of config syntax. It only highlights and enables comments (`#`,`!`,`//`,`;`)


budRich:
I added a release with the tag 1.0.0 now.

PR #6799, opened 2017-12-15
Update Vue Syntax Highlight release tags

yyx990803:
We are now releasing a rewritten version for build > 3153.

### DIFF
--- a/repository/v.json
+++ b/repository/v.json
@@ -662,8 +662,12 @@
 			"details": "https://github.com/vuejs/vue-syntax-highlight",
 			"releases": [
 				{
-					"sublime_text": "*",
-					"tags": true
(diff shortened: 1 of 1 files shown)


FichteFoll:
Because the highest `newsyntax-` tag is `1.0.1` and the last normal tag is `3.2.3`, people would be upgraded to it. You'll have to push something higher than `3.2.3`.

yyx990803:
@FichteFoll thanks, /cc @skyronic

skyronic:
Will take care of it today

skyronic:
The tags are created and the latest version is being picked up now.

image

yyx990803:
👍

PR #6802, opened 2017-12-15
bioSyntax: Syntax highlighting for Computational Biology

alyeffy:
Enables syntax highlighting for computational biology formats e.g. fasta, sam, vcf, pdb etc. in Sublime across platforms.

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -691,7 +691,6 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"platforms": ["linux", "osx", "windows"],
 					"tags": true
 				}
 			]
--- a/repository/b.json
+++ b/repository/b.json
@@ -686,6 +686,18 @@
 			]
 		},
 		{
+			"name": "BioSyntax"
+			"details": "https://github.com/bioSyntax/bioSyntax-sublime",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": ["linux", "osx", "windows"],
+					"branch": "master",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Bison",
 			"details": "https://bitbucket.org/artyom_smirnov/sublime-text-bison-highlighter",
 			"labels": ["language syntax"],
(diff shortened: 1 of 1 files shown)


alyeffy:
I have removed the Color Scheme sublime package file causing this error a while ago. Why is this still not passing?

alyeffy:
Hi, I have removed the Color Scheme sublime package file causing this error. Please review this when you get the chance. Thanks!

wbond:
@alyeffy Have you created a new tag since you removed the file? If not, that is most likely the issue.

In regards to your email:

>However, in order for our package to work, it would actually be essential for that file to be included, but in a separate directory from the other syntax files in our package. We were wondering if there was a way to bypass this. Essentially, our package enables syntax highlighting for computational biology formats e.g. sam, vcf, fasta, etc. and as such, there are very specific structures in these files that do not exist in typical language files used with standard themes, and vice versa. Anyway, for our package to work, our *.sublime-syntax files would need to be located in the usual sublime package folders but our color scheme needs to be located in the same path as the default color schemes. Any assistance would be greatly appreciated; thank you for your time!

If you want to include a color scheme, simply place the `.tmTheme` file in your repository, and it will be available for users to select using the `UI - Select Color Scheme` command palette entry. You can't put a `.sublime-package` in another `.sublime-package` because Sublime Text does not recursively load packages. There is also no way to have one package cause another to be installed automatically. You could show a user an install message telling them about the color scheme package, or as I mentioned above, you can include the color scheme in this package and everything should work fine.

In terms of using different scope names, it may be in your best interest to use the closest approximation from http://www.sublimetext.com/docs/3/scope_naming.html. This way users won't need to switch color schemes to your custom option. Typically the first two labels in each scope name are the most influential in color schemes, and you could always add more suffix labels to be more specific with your structure.

alyeffy:
Alright, thanks so much @wbond!
@ababaian

evandrocoan:
For reference, duplicated of: https://github.com/wbond/package_control_channel/pull/6833

PR #6804, opened 2017-12-15
Added City Lights Theme

noudadrichem:
Added City Lights Theme sublime version see:
[citylights.xyz/](http://citylights.xyz/)


noudadrichem:
Updated release tags to `1.0.0`

PR #6805, opened 2017-12-15
Add OpenBazelBUILDFile

macrael:
Open Bazel Build File opens the correct BUILD file for the current open file.

### DIFF
--- a/repository/o.json
+++ b/repository/o.json
@@ -394,17 +394,6 @@
 			]
 		},
 		{
-			"name": "OpenBazelBUILDFile",
-			"details": "https://github.com/macrael/OpenBazelBUILDFile",
-			"labels": ["file navigation", "bazel"],
-			"releases": [
-				{
-					"sublime_text": "*",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "Open File in Current Folder",
 			"details": "https://github.com/SublimeText/OpenFileInCurrentFolder",
 			"labels": ["file navigation", "current folder"],
@@ -606,6 +595,17 @@
 				{
 					"sublime_text": ">=3000",
 					"platforms": ["osx"],
+					"tags": true
+				}
+			]
+		},
+		{
+			"name": "OpenBazelBUILDFile",
+			"details": "https://github.com/macrael/OpenBazelBUILDFile",
+			"labels": ["file navigation", "bazel"],
+			"releases": [
+				{
+					"sublime_text": "*",
 					"tags": true
 				}
 			]
--- a/repository/o.json
+++ b/repository/o.json
@@ -394,6 +394,17 @@
 			]
 		},
 		{
+			"name": "OpenBazelBUILDFile",
+			"details": "https://github.com/macrael/OpenBazelBUILDFile",
+			"labels": ["file navigation", "bazel"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Open File in Current Folder",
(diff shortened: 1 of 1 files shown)


macrael:
The only other feature I'm considering adding to this is something that automatically sets BUILD and WORKSPACE files to be considered python files. What's the right part of the sublime plugin api to look at for that?

Thom1729:
Some thoughts:

- You don't need the shebang (`#!/usr/bin/env python3`).
- On the other hand, if this requires Python 3 then it also requires Sublime 3 and should be marked accordingly.
- The keybinding should be limited only to relevant file types.
- The command should have a more specific name to avoid conflicts; e.g., "open_bazel_build_file".
- You can avoid testing for an active view by making `OpenBuildFileCommand` a `ViewCommand` instead of a `WindowCommand`. (A view should always have a window, but the opposite is not true.)
- `sublime.status_message` is undocumented. Use the `status_message` method of `Window` instead.
- It's redundant to test `file_path and len(file_path) > 0`. For one thing, the empty string is falsy; for another, `View.file_name()` should return a file name or `None`, never an empty string (unless there's some quirk I'm not aware of).
- Will `elif check_directory == os.path.sep` work on Windows? `ntpath.sep` is `\`, but `ntpath.dirname('c:')` is `c:`. I'm not sure if there's a standard idiom, but what about `os.path.dirname(check_directory) == check_directory`?
- You're accepting arbitrary `**kwargs`, but you're not using them.

FichteFoll:
In addition to all the things that @Thom1729 said:

You only return a result in `_active_file_path` if the path is *valid*, yet your code doesn't consider the case other when the function returns `None`.

Overall, I find this package quite questionable since the only thing it does is open a `BUILD` file in the directory of the file currently being edited. You could also use Goto Anything or the normal ctrl+o for this, but I guess everyone wants a little automization sometimes.

>The only other feature I'm considering adding to this is something that automatically sets BUILD and WORKSPACE files to be considered python files. 

I don't understand what you mean with this.

Oh, and I would rather describe `sublime.status_message` as deprecated. The window-specific one should be used instead, as mentioned.

macrael:
Thank you both for your comments, this is my first extension and y’all have been very helpful. I wrote this precisely because using open anything was very frustrating. A Bazel project has filed named BUILD in many directories so it was a pain to get the right one. This extension finds the only one that is associated with the current file. 

BUILD files are written in Bazel’s language, which is a subset of python but since they don’t have an extension it’s difficult to get sublime text to recognize them as such. I’m not sure how an extension can affect that, can they add preferences to the preferences file on behalf of the user?

Cheers,

—MacRae

FichteFoll:
When you open a file with `window.open_file`, the created view is returned. You can then run `view.set_syntax_file("Packages/Python/Python.tmLanguage")` on that to activate the default Python syntax. (Iirc it will automatically recognize and prioritize a `.sublime-syntax` file, if it exists.)

macrael:
I assume you meant TextCommand, that seems to have worked.

macrael:
I updated my project based on your comments, let me know if it looks good to you. Thank you again for your help.

macrael:
@FichteFoll Any chance this could be merged? I'm not sure who else to ask.

FichteFoll:
Remember that you need to push a new tags to make your changes get picked up by Package Control. After that I can merge. Sorry for not visiting earlier, but I've been busy with other stuff.

`TextCommand` is the right class name. It just gets confused quickly.

And just a final hint, you could easily make `_find_build_file` iterative, which is more efficient due to less function calls, but that's really minor.

macrael:
Tag pushed! Should be ready to go.

PR #6810, opened 2017-12-19
Add JS Custom package.

Thom1729:
This package provides a framework (powered by YAML Macros) for extending the core JavaScript syntax with new features, along with a basic set of extensions including:

- JSX support
- Flow typing support
- Custom tagged templates (e.g. ```style`div { color: red; }` ```)

The goal is to allow devs to use nonstandard extensions such as the above in any combination and without introducing a brand-new syntax that would inevitably diverge from core.

This is an ~alpha~ beta release; some work remains to be done on the engine and extensions. However, the system does work and I'm using it to write production code. Some of the remaining work involves Package Control integration (for example, installing YAML Macros), so it would be very helpful to have the package properly available in Package Control before the initial public release.

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -849,7 +849,7 @@
 			"labels": ["javascript", "syntax"],
 			"releases": [
 				{
-					"sublime_text": ">=3000",
+					"sublime_text": ">=3092",
 					"tags": true
 				}
 			]
--- a/repository/j.json
+++ b/repository/j.json
@@ -844,6 +844,17 @@
 			]
 		},
 		{
+			"name": "JSCustom",
+			"details": "https://github.com/Thom1729/Sublime-JS-Custom",
+			"labels": ["javascript", "syntax"],
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
Since the package creates `.sublime-syntax` files, it should be installable to only 3092 and up. Probably even limit it to the latest stable release since I imagine it relies on the updated JavaScript syntax quite a lot.

If it uses the `embed` action (didn't check), it would require 3153 even, iirc.

Thom1729:
I've marked it for >=3092.

It bundles its own copy of the latest version of the core syntax. This is a bit clumsy, but otherwise changes to the core syntax could break the extensions. Consequently, this version should work fine on any version of Sublime that supports `.sublime-syntax` files (and should continue to work in the future).

There's one exception: one of the extensions does use `embed`. As a result, this feature will not work on older builds, although the other features will. I'll make a note of this in the documentation. If I get questions or complaints about it, it might be worth adding internal error handling for this case.

PR #6811, opened 2017-12-19
Add Rainglow color themes.

daylerees:
This package provides 100+ of my Color themes for Sublime (also available on vscode and jetbrains repos).

Thanks for maintaining Package Control! :)

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -2,16 +2,6 @@
 	"schema_version": "3.0.0",
 	"packages": [
 		{
-			"name": "Rainglow",
-			"details": "https://github.com/daylerees/rainglow",
-			"releases": [
-				{
-					"sublime_text": "*",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "R Tools",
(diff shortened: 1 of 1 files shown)


daylerees:
Fixed the location. Didn't account for single letter package names. :)

jrappen:
Why not use the new `sublime-color-scheme` format based on JSON?

---

For examples, do the following in `Build 3149` or later:

* Tools > Command palette > `View package file` > `sublime-color-scheme`

FichteFoll:
An easy answer would be that it's only available for dev builds.

daylerees:
@jrappen @FichteFoll thanks for the replies and the merge! I didn't know about the new format. Maybe I'll convert them when it's more widely available. :)

jrappen:
@daylerees just a heads-up: you can now use [PackageDev](https://github.com/sublimetext/PackageDev) to convert the color scheme(s) to the **new** [color scheme format for ST3149+](https://www.sublimetext.com/docs/3/color_schemes.html)

PR #6815, opened 2017-12-22
Add Smart Indent

YKZDY:
The package allows user separately control Indent Size and Tab Size in Sublime Text.


YKZDY:
Created Release Package

PR #6817, opened 2017-12-22
Adds `CDNUpdates`

HorlogeSkynet:
Hi there,

I've recently started to develop a plugin to automatic check for CDN updates (because it's pretty boring to do manually).
Some providers are currently well handled, and I'll add some more in the future.

This is my first time here, so I'll accept each one of your comments (maybe about its name ?) !

Thanks,
Bye 👋

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -394,6 +394,17 @@
 			]
 		},
 		{
+			"name": "CDNUpdates",
+			"details": "https://github.com/HorlogeSkynet/CDNUpdates",
+			"labels": ["CDN", "updates", "web development"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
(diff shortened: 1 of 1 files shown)


FichteFoll:
I think I reviewed or at least saw a similar package more or less recently, but I can't seem to find it.

You should prefer using relative imports over absolute ones. Also, your package will only work on ST3 so you should adjust the version requirements in this PR.

You also might want to urlescape the name for your GET requests.

HorlogeSkynet:
Hi, thanks for the review. I've looked up before starting it, and didn't found any result 🤔

I'll fix what you point out this WE. Bye 👋

HorlogeSkynet:
Hello @FichteFoll, I hope I fixed what you've highlighted. Just keep me posted 👌

Bye

PR #6818, opened 2017-12-23
Added the PythonDebugTools on repository/dependencies.json

evandrocoan:
Fixes https://github.com/evandrocoan/SublimeAMXX_Editor/issues/2

The package https://packagecontrol.io/packages/amxmodx now requires this as dependency, but it is missing on the default channel.

### DIFF
--- a/repository/dependencies.json
+++ b/repository/dependencies.json
@@ -567,7 +567,10 @@
 		},
 		{
 			"name": "PythonDebugTools",
-			"details": "https://github.com/evandrocoan/PythonDebugTools",
+			"author": "evandrocoan",
+			"load_order": "10",
+			"description": "Alternate simplified logging support",
+			"issues": "https://github.com/evandrocoan/PythonDebugTools/issues",
 			"releases": [
 				{
 					"sublime_text": ">=3114",
--- a/repository/dependencies.json
+++ b/repository/dependencies.json
@@ -573,6 +573,7 @@
 			"issues": "https://github.com/evandrocoan/PythonDebugTools/issues",
 			"releases": [
 				{
+					"base": "https://github.com/evandrocoan/PythonDebugTools",
 					"sublime_text": ">=3114",
 					"tags": true
 				}
--- a/repository/dependencies.json
+++ b/repository/dependencies.json
@@ -566,6 +566,16 @@
 			]
 		},
 		{
+			"name": "PythonDebugTools",
+			"details": "https://github.com/evandrocoan/PythonDebugTools",
+			"releases": [
+				{
+					"sublime_text": ">=3114",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "pytz",
 			"load_order": "50",
 			"description": "Python pytz module",


FichteFoll:
You're missing the `base` key here. Wonder why the tests didn't catch that.

FichteFoll:
In your readme you say to import `debug_tools`, but the folder with the pacakge is named `python_debug_tools`.

Don't run `logging.basicConfig`. It will install a root logger, which is global, and affect *all packages*, whether they want or not. It seems like you piggyback the logging module for basic output but don't actually know how to use it.

You should create a local logger (as you do already) and then attach one or multiple handlers to it, each with their own formatters. Then you can call logging functions on that instance without affecting global state. Consult a guide on Python logging or the docs in the stdlib for more details.

---

I generally prefer interfacing with the logging module directly as it functions pretty well as-is already. You can see example usages in PackageDev:

https://github.com/SublimeText/PackageDev/blob/master/_logging.py
https://github.com/SublimeText/PackageDev/blob/5f4c6ec058d7daa6bb806a774a17fbd678c3d41c/plugins_/color_scheme_dev.py#L15

evandrocoan:
> You're missing the base key here. Wonder why the tests didn't catch that.

Opps. Perhaps open a issue for this?


___ 

> In your readme you say to import debug_tools, but the folder with the pacakge is named python_debug_tools.

I had moved it some days ago and forgot to fully update it.


___ 
> Don't run logging.basicConfig. It will install a root logger, which is global, and affect all packages, whether they want or not

At the time, I did not understand the `logging` and for what seems a miracle, after adding that it started working. Now I looked into the `logging.basicConfig` implementations and figured out why it was working.

___ 
> I generally prefer interfacing with the logging module directly as it functions pretty well as-is already. You can see example usages in PackageDev.

Thanks for the examples! I solved this limitation by making this extend the `logging` module. Now this is partially independent of some aspects from the logging module.

I wrote this because when I was looking for a logger module I saw it does not supported to use bit-mask to do logging filtering. Now I extended the logging module adding support to this.

Now you instantiate this keeping the full potencial of the logging module. I would like to use this because I added some features and it still working exactly as the current logging module. But I also created some setups which can easily change same stuff, for example:
```python
# To run this file, run on the Sublime Text console:
# import sublime_plugin; sublime_plugin.reload_plugin( "PythonDebugTools.tests.manual_tests" )

# Import and reload the debugger
from python_debug_tools.logger import getLogger

import sublime_plugin
sublime_plugin.reload_plugin( "python_debug_tools.logger" )

log = getLogger( 127, __name__ )

log( 1, "My logging" )
log.warn( "A warning" )
log.debug( "A debugging" )

def function_name():
    log( 1, "Bitwise" )
    log( 8, "Bitwise" )
    log.warn( "Warn" )
    log.info( "Info" )
    log.debug( "Debug" )

log.setup_logger( function=False, level=True )
log.insert_empty_line()
function_name()

log.setup_logger( date=True )
log.insert_empty_line()
function_name()
```

```
reloading plugin PythonDebugTools.tests.manual_tests
reloading plugin python_debug_tools.logger
[PythonDebugTools.tests.manual_tests] 16:31:26:638.928890 2.19e-04 <module>:13 My logging
[PythonDebugTools.tests.manual_tests] 16:31:26:639.429092 4.90e-04 <module>:14 A warning
[PythonDebugTools.tests.manual_tests] 16:31:26:639.930010 4.77e-04 <module>:15 A debugging

[PythonDebugTools.tests.manual_tests] 16:31:26:640.930891 3.87e-04 DEBUG(1) Bitwise
[PythonDebugTools.tests.manual_tests] 16:31:26:641.431093 4.62e-04 DEBUG(8) Bitwise
[PythonDebugTools.tests.manual_tests] 16:31:26:641.431093 3.04e-04 WARNING Warn
[PythonDebugTools.tests.manual_tests] 16:31:26:641.931057 2.96e-04 INFO Info
[PythonDebugTools.tests.manual_tests] 16:31:26:642.431974 2.80e-04 DEBUG Debug

[PythonDebugTools.tests.manual_tests] 2018-01-13 16:31:26:642.931938 2.47e-04 function_name:18 Bitwise
[PythonDebugTools.tests.manual_tests] 2018-01-13 16:31:26:642.931938 2.88e-04 function_name:19 Bitwise
[PythonDebugTools.tests.manual_tests] 2018-01-13 16:31:26:643.431902 3.04e-04 function_name:20 Warn
[PythonDebugTools.tests.manual_tests] 2018-01-13 16:31:26:643.431902 2.85e-04 function_name:21 Info
[PythonDebugTools.tests.manual_tests] 2018-01-13 16:31:26:643.933058 2.99e-04 function_name:22 Debug
```

FichteFoll:
You shoudln't be using `sublime_plugin.reload_plugin` on something that isn't even a plugin. If you want to reload the module, use `imp.reload` directly.

PR #6822, opened 2017-12-25
Add LinesMultisets

heldev:

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -885,6 +885,22 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


evandrocoan:
Nice package, thanks for sharing!

heldev:
Thank you :)

FichteFoll:
Instead of defining lambdas for simple operators like `lambda a, b: a + b`, you can also use `operator.add` and other functions in the `operator` module.

PR #6826, opened 2017-12-28
Add timestamp converter

Zlitus:


Thom1729:
- What does this package do? The readme is unenlightening.
- Rather than having separate "real" commands, why not use the new `TextInputHandler` mechanism?
- Did you mean to leave in those `print` statements? If you really do want the commands to print output to the console, you should probably prefix it with the name of your package so the user knows where it's coming from.
- On failed encoding/decoding, you're replacing the selection with an error message. This is sort of a rude failure mode. Instead, you might want to use a phantom to display an error.
- Instead of using positional arguments to `show_input_panel` and commenting them, why not use keyword arguments?
- Is `%Y/%m/%d %H:%M:%S` a common format? (Ideally, the default would be customizable.)

Zlitus:
Thank you @Thom1729 , I'll see and work on what you suggest. Is it blocking for the release of the plugin? I will not have time to do it this week.

Thom1729:
It's not me doing the approvals, but I'd suggest improving the description and removing the unneeded print statements. The other stuff is less important.

FichteFoll:
I have a similar feature lined up for InsertDate (https://github.com/FichteFoll/InsertDate/issues/29), but I don't see myself finding the time to implement it properly in the near future.

A more appropriate method to show the error would be a `window.status_message` call, or maybe even `.error_message`.

The package generally works, so I'm gonna merge it. You should still address @Thom1729's remaining comments though (except for maybe the input handler). It would increase the overall quality of the package.

PR #6827, opened 2017-12-29
Add SwitchPanel

stoivo:


repository/s.json
```diff
@@ -5041,6 +5041,19 @@
 				}
 			]
 		},
+		{
+			"name": "SwitchPanel",
+			"description": "Switch which output to open and keep open",
```
gwenzek: suggestion:
"Open output panels from the command palette"

What do you mean by "keep open" ?

gwenzek:
Good plugin idea ! Clean and simple implementation.

Nits:
* you have a typo in your function name: `inclide_panel`
* the plugin description reads a bit awkward to me

Feature requests for future version of this plugin: 
* add a command to cycle through the panels without opening the panel selection menu.
* display a nicer names for the panels, to match the displayed name  (not sure if possible)

Thom1729:
In addition, you should use `Window.status_message` instead of `sublime.status_message`.

stoivo:
> Good plugin idea! Clean and simple implementation.

Thanks

>Nits:
>
>I don't think you need to call self.window.run_command("ps_output_panel", {"idx": idx}) inside on_hightlight

That seems correct. First I tried to open all panels, including console, find, replace, etc. Those would take the focus away. And that's who I got the focus back. But since I excluded those views completely, I can remove it.

>you have a typo in your function name: inclide_panel

Thanks, fixed.

>the plugin description reads a bit awkward to me

It was clear in my mind. Something went wrong on the export

>Feature requests for future version of this plugin:
>
>add a command to cycle through the panels without opening the panel selection menu.

Yes, implemented.

>display a nicer names for the panels, to match the displayed name (not sure if possible)

Now if the panel either is a file or has a name it will be displayed.

stoivo:
I also create a new tag for the release

stoivo:
> In addition, you should use Window.status_message instead of sublime.status_message.

I will change this for some other package as well. 
Thanks, this is fixed too.

evandrocoan:
I found this on the top of your README.md:

> With this package you can switch with panel should be displayed

But I do not understand this. Can you:

1. Elaborate more on this description?
1. Why I should use this package? i.e., for what it is good for?

stoivo:
I have a screenshot in the readme. You should look at that one. It shows what the plugin does.

I think it explains it quite good. It is short and concrete.

FichteFoll:
The key bindings shown in your readme are invalid. The `+`s should not be excluded from the string.

Also note the typo in the sentence quoted by @evandrocoan, `with` vs `which`.

Good plugin though. In fact, I believe this to be so substantial that it should be included in ST by default.

stoivo:
> The key bindings shown in your readme are invalid. The +s should not be excluded from the string.

Thanks, I do not know how I got the wrong thing there. I copied it from somewhere else. 

> Also note the typo in the sentence quoted by @evandrocoan, with vs which.

The second with should be a `which`. I didn't see it when @evandrocoan because I stopped looking after the first `with`.

> Good plugin though. In fact, I believe this to be so substantial that it should be included in ST by default.
 
Wow, it would be an owner to have it a part of the ST.

> Edit: Oh, and some comments on the code:
> Your Mixin class should actually just be two functions, where output_panels had the signature list_panels(window) and include_panel should_include_panel(name). I wonder why you exclude the find results panel though.

After starting sublime and running git graph with GitSavvy I have these panels. 
```
['console', 'find', 'find_in_files', 'output.GitSavvy', 'output.find_results', 'output.show_commit_info', 'replace']
```


console => The python console *
find => The find for current buffer *
find_in_files => The find for open folders *
output.GitSavvy => GitSavvy log of git command in development
output.find_results => In `find_in_files` there is an option to use buffer. I am using find to use buffer instead of output panel. Therefore `output.find_results` is always empty. Since I don't use it I excluded it. I could check if it empty before excluding it. *
output.show_commit_info => GitSavvy, commit details
replace => The find and replace for current buffer *

'*' Created by ST at boot

PR #6814, opened 2017-12-21
Fixes latest version breaks syntax highlighting

evandrocoan:
Fixes evandrocoan/SublimeNotepadPlusPlusTheme#1 latest version breaks syntax highlighting

PR #6829, opened 2017-12-29
PostCSS Sorting dropped support for Sublime Text 2

hudochenkov:
In the new version, I use new settings split view, which is introduced in Sublime Text 3 while ago.

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -1682,7 +1682,7 @@
 			"details": "https://github.com/hudochenkov/sublime-postcss-sorting",
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": ">=3000",
 					"tags": true
 				}
 			]
--- a/repository/p.json
+++ b/repository/p.json
@@ -1684,6 +1684,10 @@
 				{
 					"sublime_text": ">=3000",
 					"tags": true
+				},
+				{
+					"sublime_text": "<3000",
+					"tags": "st2-"
 				}
 			]
 		},


evandrocoan:
@hudochenkov, you could allow older users with Sublime Text 2 to use the latest version which worked on Sublime Text 2 by creating a tag like `st2-2.3.4` and configuring package control to ship only the tags with prefix `st2-` to Sublime Text 2 and keep the current unprefixed tags to the latest version of Sublime Text.

To do this:

1. You can set the channel file like this:
    ```json
            {
                "name": "PostCSS Sorting",
                "details": "https://github.com/hudochenkov/sublime-postcss-sorting",
                "releases": [
                    {
                        "sublime_text": ">=3000",
                        "tags": true
                    },
                    {
                        "sublime_text": "<3000",
                        "tags": "st2-"
                    }
                ]
            },
    ```
1. And create the tag `st2-x.x.x` on the latest commit support Sublime Text 2.

I had did this to my package which had breaked on Sublime Text build 3144. You can see the commit here: 

1. https://github.com/wbond/package_control_channel/pull/6814/files
1. https://github.com/evandrocoan/NotepadPlusPlusColorScheme/releases/tag/build3144support-1.2.4

hudochenkov:
@evandrocoan thanks! Added this change.

FichteFoll:
Also note that any ST2 users that have upgraded to the 3.0.0 version already will not be downgraded to 2.1.1 again automatically. You'd need to push a version higher than 3.0.0 for the st2 channel to ensure that everyone receives the proper version of the package.

PR #6831, opened 2017-12-30
Adding Markdown Complement Package

junShimoji:
This plugin provides functions to complement markdown a list item making/changing or indent up/down.
Users need to set Key Bindings only, doesn't need to care extensions(.md,.mdown,.txt or like that).
Further more, although other similar plugins conflicts with Japanese Input method that selected characters are disappearing, this plugin doesn't occur such errors.
And you can change list items manually not connected with indent levels.

I tested with package_control_panel and results are followings:

```
$ python test.py
....
Ran 8213 tests in 0.992s

OK
````

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -470,6 +470,16 @@
 			]
 		},
 		{
+			"name": "MarkdownCompletion",
+			"details": "https://github.com/junShimoji/markdowncomplements",
+			"releases": [
+				{
+					"sublime_text": "<3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "MarkdownEditing",
 			"details": "https://github.com/SublimeText-Markdown/MarkdownEditing",
 			"labels": ["markdown", "github markdown", "gfm", "multimarkdown", "color scheme"],
--- a/repository/m.json
+++ b/repository/m.json
@@ -470,8 +470,8 @@
 			]
 		},
 		{
-			"name": "MarkdownCompletion",
-			"details": "https://github.com/junShimoji/markdowncomplements",
+			"name": "MarkdownComplements",
+			"details": "https://github.com/junShimoji/MarkdownComplements",
 			"releases": [
 				{
 					"sublime_text": "<3000",

### CODE REVIEW

**Comment by gwenzek** [comment]
Note that you need to tag your releases like this "0.1.0".

You talk about "similar plugins", did you try to submit PRs to those plugins instead ?

About your plugin:

* you have a while loop to find the position of the previous '\n'. Can't you just use something like:
`view.line(pos[0]).begin() - 1` ?
* all your commands read the setting `tab_size`, few of them use it.
* you should suggest the user to limit the keybings to Markdown files:

```json
{
  "keys": ["ctrl+enter"], "command": "markdown_indent_down",
  "context": [{ "key": "selector", "operator": "equal", "operand": "text.html.markdown" }]
}
```

* If you call your command on a non list line, it does nothing. I feel it would be more useful to call the default sublime command. This way, you could map `enter` to `MarkdownNewLineCommand` instead of `ctrl+enter`.
* The repo name doesn't seem to match the guidelines (PascalCase or snake_case or is that rule just for the name on PackageControl ?)
* You use `pos[0].a` through out your code, be aware that's not always equal to  `pos[0].begin()` ([doc](http://www.sublimetext.com/docs/3/api_reference.html#sublime.Region))
* I feel like the code in `MarkdownRotateListItem` could be simplified by capturing the number of spaces and the decoration char in your regex.

junShimoji:
> Note that you need to tag your releases like this "0.1.0".

Replaced 0.1 with 0.1.0 and incremented it to 0.2.0 .

> You talk about "similar plugins", did you try to submit PRs to those plugins instead ?

Added at README.

> you have a while loop to find the position of the previous '\n'. Can't you just use something like:
> view.line(pos[0]).begin() - 1 ?

I substituted it for while loop. 

> all your commands read the setting tab_size, few of them use it.

The part which tab_size is not used are deleted.

> you should suggest the user to limit the keybings to Markdown files:
> {
>   "keys": ["ctrl+enter"], "command": "markdown_indent_down",
>   "context": [{ "key": "selector", "operator": "equal", "operand": "text.html.markdown" }]
> }

I added it at README.

> If you call your command on a non list line, it does nothing. I feel it would be more useful to call the default sublime command. 
> This way, you could map enter to MarkdownNewLineCommand instead of ctrl+enter.

I understand it. But I let it stand.

> The repo name doesn't seem to match the guidelines (PascalCase or snake_case or is that rule just for the name on PackageControl ?)

Changed to PascalCase.

> You use pos[0].a through out your code, be aware that's not always equal to pos[0].begin() (doc)

pos[0].a was changed other expressions as long as I can.

> I feel like the code in MarkdownRotateListItem could be simplified by capturing the number of spaces and the decoration char in your regex.

Simplified it.

FichteFoll:
Just for the record, repository names do not matter.

PR #6832, opened 2017-12-30
Add Graphvizer Package

hao-lee:
Signed-off-by: Hao Lee <haolee.swjtu@gmail.com>

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -1777,6 +1777,18 @@
 			]
 		},
 		{
+			"name": "Graphvizer",
+			"details": "https://github.com/hao-lee/Graphvizer",
+			"author": "Hao Lee",
+			"labels": ["graphviz", "dot", "visualization"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "GraphvizPreview",
 			"details": "https://github.com/munro/SublimeGraphvizPreview",
 			"releases": [


FichteFoll:
It's encouraged to use relative imports over mutating `sys.path` where possible. It seems that the vendored packages `antlr4` and `lexerparser` use absolute imports internally though, so you'd have to adjust all the imports of that and I can understand how that could turn out to be annoying for updating them eventually.

You must be aware that the entire interpreter state (including all imported modules) is shared among all plugins, though.

A way to circumvent that would be to provide your third-party modules as Package Control dependencies, though for really specialized packages that doesn't make sense most of the time. It would for `enum` and `typing` though and I in fact already have plans to add them. `pathlib` also. You can find documentation for that here: https://packagecontrol.io/docs/dependencies, and a list of most (not all) currently available dependencies here: https://packagecontrol.github.io/.

Finally, you should bundle your vendored packages in a `lib` subfolder and only add that to `sys.path` for now. Currently, other package would end up with your `panel.py` if they `import panel` and that's not something we need to happen.

hao-lee:
Hi, thanks for your reply.
I still have some questions about how to deal with the `enum` and `typing` backport libraries. These libraries are not in the Package Control dependencies. It hard to change the source code of `antlr4` to import `enum` and `typing` relatively. Could you give me some hints? Thanks!

FichteFoll:
I just added the `enum` and `typing` dependencies: https://github.com/wbond/package_control_channel/commit/808f6390189edd51ec3b642ebd498862917c34e2. You can now use them by specifying them in a `dependencies.json`, as described in the readmes on the respective repos (the `base` url). The modules will then be available on `sys.path` and importable as usual.

hao-lee:
@FichteFoll Thanks for your hard work! `dependencies` is very convenient.

hao-lee:
Hi, I have change all `absolute import` to `relative import` and use `dependencies` instead of providing libraries by myself :smiley:

FichteFoll:
Good work.

PR #6838, opened 2018-01-01
Add RulerColumn

gerardroche:


gwenzek:
Few comments about the plugin:

* It doesn't seem to work on new views. You have to toggle word wrap on and off for the plugin to start highlighting the view.
* `settings.get('ruler_column')` doesn't provide a default. This makes the plugin crash if the user didn't set it up.
  you should either provide a reasonable default and/or use the settings "rulers" from the Default settings, this way you're consistent with the user settings.
* about perf, you could make the plugin only run on file saved and view activated (like a linter) and make the running on every edit optional.

gerardroche:
> It doesn't seem to work on new views. You have to toggle word wrap on and off for the plugin to start highlighting the view.

I can't reproduce this. Can you show me some steps of how to reproduce it e.g. what do you mean by new view? How are your creating it? etc.

> settings.get('ruler_column') doesn't provide a default. This makes the plugin crash if the user didn't set it up.
you should either provide a reasonable default and/or use the settings "rulers" from the Default settings, this way you're consistent with the user settings.

Good catch! I fixed it so that we just don't create a ruler column if `ruler_column` is not set.

> about perf, you could make the plugin only run on file saved and view activated (like a linter) and make the running on every edit optional.

I kinda wanted to do this, but it there are situations like when you wrap a long line the ruler column gets wrapped with the text onto the next line.

Example

Say you have a line where `| |` is the ruler column. Now put your cursor on `x` and press enter to break the line.

```
........................x..| |....
``` 

If we don't update the ruler column on_modified then we end up with the ruler column on the next line:

```
........................
..| |....
``` 


There is also still an issue where enabling/disabling word wrap doesn't update the ruler column. 

It reasons like this that I'm suggesting this feature be added to the core, because though these issues can be solved, I don't think the performance impact is all that trivial.

See https://github.com/SublimeTextIssues/Core/issues/2135

PR #6839, opened 2018-01-02
Update j.json - JavaScript Enhancements plugin

pichillilorenzo:
Added [JavaScript Enhancements](https://github.com/pichillilorenzo/JavaScriptEnhancements) plugin. It offers not only a smart javascript autocomplete but also a lot of features about creating, developing and managing javascript projects. You could use it also in existing projects (see the Wiki).
It turns Sublime Text into a JavaScript IDE like!

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -325,7 +325,7 @@
 		{
 			"name": "JavaScript Enhancements",
 			"details": "https://github.com/pichillilorenzo/JavaScriptEnhancements",
-			"labels": ["auto-complete", "linting", "autocomplete", "completions", "javascript", "javascript completions", "flow", "flow-typed", "javascript tools", "project management"],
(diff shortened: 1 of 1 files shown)


pichillilorenzo:
'.no-sublime-package' is defined because the package runs scripts included in the package using  node.js.

Thom1729:
Why create a new package rather than add features to your other package, JavaScript Completions?

pichillilorenzo:
@Thom1729 because it has a different approach. It works using Flow and it is not just only for javascript completions, but also to manage different type of projects. Furthermore there will be a lot of features. Instead JavaScript Completions plugin is just a list of static completions with other simple small features.

FichteFoll:
I suspect the reason why you put everything in a single file (that is generated from several different files) is to allow proper reloading of the file whenever you change something small.
I suggest checking out the https://packagecontrol.io/packages/AutomaticPackageReloader package and instead structuring your code modularly like any other Python project.

The rest of the code is too complex and too much for me to review, so we'll have to live with that.

(Also, there's no package like `TerminalView` for Windows so far.)

PR #6840, opened 2018-01-02
Add EasyWorkspace Package

jmooney:
Hello! Thanks for taking the time to review this package. 

This package aims to provide simpler and easier to use workspace management into sublime text. Particularly, we provide for saving open files and folders to a workspace file, and reopening as needed. Per-file state such as selection and visible region are saved as well. 

Additionally, all the added sublime commands are designed to work with the command line. In this way, we've provided the ability to integrate with Git, such that git aliases can be used to automatically open/save sublime's workspace to match a git repo and branch name.

Probably the most similar plugin currently available would be something like FileBinder (https://github.com/JeroenVdb/FileBinder). EasyWorkspace tries to be a bit more ambitious in mimicking sublime workspaces instead of a set of useful files. We also think the git integration design feature is a nice plus.

Thanks again for your review. Any feedback is welcomed :).

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -154,6 +154,17 @@
 			]
 		},
 		{
+			"name": "EasyWorkspace",
+			"details": "https://github.com/jmooney/EasyWorkspace-sublime",
+			"labels": ["workspace", "files", "project"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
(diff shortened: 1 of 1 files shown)


jmooney:
Thanks for the feedback. I'll make the change here in the next day or two. Anything else you might suggest as improvements?

PR #6842, opened 2018-01-03
Add Newton Theme

bertolinimarco:

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -1394,7 +1394,6 @@
 			"releases": [
 				{
 					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


jrappen:
Why not use the new `sublime-color-scheme` format based on JSON?

---

For examples, do the following in `Build 3149` or later:

* Tools > Command palette > `View package file` > `sublime-color-scheme`

FichteFoll:
According to Sublime Text terminology, this is a color scheme and not a theme. Please adjust the package name and labels.

PR #6850, opened 2018-01-06
Adding Markdown Complement Package

junShimoji:
>  1. Used `"tags": true` and not `"branch": "master"` 
>     ([versioning docs](https://packagecontrol.io/docs/submitting_a_package#Step_4))

Done.

>  2. Added a readme to your repository so that users (and reviewers) 

Done. README exists at https://github.com/junShimoji/MarkdownComplements

> You may proceed with a short description of what the package does and, 
> in case a similar package already exists, 

* This plugin provides functions to complement a markdown list item making/changing or indent up/down.
* Need to set Key Bindings only.
* No need to care extensions(.md,.mdown,.txt or something like that).
* Can change list items manually not connected with indent levels.
* Can prevent conflicting with Japanese Input method that selected characters are disappearing.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -470,6 +470,16 @@
 			]
 		},
 		{
+			"name": "MarkdownComplements",
+			"details": "https://github.com/junShimoji/MarkdownComplements",
+			"releases": [
+				{
+					"sublime_text": "<3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "MarkdownEditing",
 			"details": "https://github.com/SublimeText-Markdown/MarkdownEditing",
 			"labels": ["markdown", "github markdown", "gfm", "multimarkdown", "color scheme"],


junShimoji:
It is "RE" pull requested one from #6831 because of not cleared travis tests after revised. Follows are answers of your previous comments.

> Note that you need to tag your releases like this "0.1.0".

Replaced.

> You talk about "similar plugins", did you try to submit PRs to those plugins instead ?

Added at README.

> you have a while loop to find the position of the previous '\n'. Can't you just use something like:
> view.line(pos[0]).begin() - 1 ?

I substituted it for while loop. 

> all your commands read the setting tab_size, few of them use it.

The part which tab_size is not used are deleted.

> you should suggest the user to limit the keybings to Markdown files:
> {
>   "keys": ["ctrl+enter"], "command": "markdown_indent_down",
>   "context": [{ "key": "selector", "operator": "equal", "operand": "text.html.markdown" }]
> }

I added it at README.

> If you call your command on a non list line, it does nothing. I feel it would be more useful to call the default sublime command. 
> This way, you could map enter to MarkdownNewLineCommand instead of ctrl+enter.

Understood it. But let it stand.

> The repo name doesn't seem to match the guidelines (PascalCase or snake_case or is that rule just for the name on PackageControl ?)

Changed to PascalCase.

> You use pos[0].a through out your code, be aware that's not always equal to pos[0].begin() (doc)

pos[0].a was changed other expressions as long as I can.

> I feel like the code in MarkdownRotateListItem could be simplified by capturing the number of spaces and the decoration char in your regex.

Simplified it.

gwenzek:
Thanks for the changes, my only question that you didn't answer is:
Did you try to put these functionalities in one of the many Markdown plugins ?

Also could you add a 1 line description to your plugin, like: 'Easy indenting of Markdown lists."

junShimoji:
> Did you try to put these functionalities in one of the many Markdown plugins ?

I use MarkdownEditing or SmartMarkdown those are very useful.
But for especially Japanese, <kbd>enter</kbd> or <kbd>Tab</kbd> conflicts Japanese ```Kana``` Input method.
I'd like to have a workaround other than changing the setting of those plugins.
Remaining small amount of functions, we can use the functions provided by present Markdown's plugins.

FichteFoll:
Your package is not aware of multiple selections. You should iterate over `view.sel()` and perform the requested action on each selection (note that this collection can also be empty).

junShimoji:
> Your package is not aware of multiple selections. 

I added functions at v0.4.0.

> Imo this package can co-exist along other markdown packages as a stripped down version only supplying the specified functionality. 

As I said at previous comment, one of the purpose of applying this package is that to avoid the conflicts Japanese IME(input method) or like that.

> I don't even know if any of the packages already include this.

Some popular packages don't maneuver effectively.

PR #6852, opened 2018-01-06
Add the Make Automatic Variables package

roperzh:
Hello, this package provides a series of snippets that I use to remember make automatic variables.


FichteFoll:
Good idea to reduce the mental capacity required for those.

roperzh:
wohooo! thank you :)

PR #6862, opened 2018-01-08
Add Swift autocomplete

Dan2552:
Provides autocomplete and mouse-hover documentation for Swift, in a similar manner to Xcode or other IDEs by using SourceKitten.

A similar package to https://github.com/johncsnyder/SwiftKitten, which doesn't really work anymore, and isn't maintained (I originally started with a pull request into this project before I went with a complete rewrite). My plugin, at least in my opinion, has surpassed the functionality provided by SwiftKitten.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -4907,6 +4907,18 @@
 			]
 		},
 		{
+			"name": "Swift Autocomplete",
+			"details": "https://github.com/Dan2552/SublimeTextSwiftAutocomplete",
+			"labels": ["auto-complete", "completions", "documentation", "swift"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": ["osx", "linux"],
+					"tags": true
+				}
+			]
+		},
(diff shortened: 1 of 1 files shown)


FichteFoll:
You should avoid modifying `sys.path` as it will affect the entire runtime and all plugins. Especially for your own code, you can just use relative imports instead of absolute ones. For the `ijson` dependency, you would ideally just replace the few absolute imports that it uses internally (for whatever reason, \*sigh\*) with relative ones.

Here's an example PR where the author did simlar things recently: https://github.com/wbond/package_control_channel/pull/6832

Dan2552:
Thanks for the suggestions of improvement. This is the only thing I've really done in Python, so I'm not really sure on what kinds of things are best practice.

- I've removed the `ijson` dependency. I believe I originally had it for performance reasons. Didn't really see a notable difference this time though... 🤷‍♀️ 
- No longer modifying `sys.path`

PR #6865, opened 2018-01-10
add rainbow_csv to package control

mechatroner:
Hello Everyone!
This is my first sublime plugin, it highlights column in *.csv and *.tsv files in different rainbow colors and provides some info about current column on mouse hover. This is an initial version, later I plan to add more features, e.g. ability to run SQL-like queries against the csv files, like it is done in vim version of this plugin ( https://github.com/mechatroner/rainbow_csv )

I am aware that sublime text already has CSV plugins, most popular is "Advanced CSV", but my plugin uses totally different syntax highlighting, and in the future will provide a different way to manipulate table data (using SQL-like language as I mentioned above).

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -218,6 +218,17 @@
 			]
 		},
 		{
+			"name": "rainbow_csv",
+			"details": "https://github.com/mechatroner/sublime_rainbow_csv",
+			"labels": ["csv", "tsv", "highlight"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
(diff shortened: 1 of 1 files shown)


FichteFoll:
You should name your syntax file "CSV (Rainbow)" or similar to differentiate it from potential other CSV syntax files (and for more clarity towards the user). Don't forget to rename the settings file too. And the TSV variant.

mechatroner:
I've renamed the syntaxes, see the linked commit. Also I've created a new tag "0.0.2" in my repo.

PR #6866, opened 2018-01-10
K Framework Syntax Highlighter

TeodorPloae:

### DIFF
--- a/repository/k.json
+++ b/repository/k.json
@@ -13,6 +13,17 @@
 			]
 		},
 		{
+			"name": "K Framework Syntax Highlighter TeodorPloae",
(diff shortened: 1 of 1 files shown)


FichteFoll:
You should stick to the scope naming guidelines laid out here: http://www.sublimetext.com/docs/3/scope_naming.html

Notably, that means `comment.line.double-slash.k` instead of `comment` or probably some keyword-based and variable-based scope names for `mov`, `rosu` and `rosu_inchis`, for example. Currently, your syntax definition is basically useless for any color scheme other than the bundled one.

FichteFoll:
@TeodorPloae ping

PR #6870, opened 2018-01-11
add meld compare plugin

scott1028:

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -1038,7 +1038,8 @@
 			"releases": [
 				{
 					"sublime_text": ">=3000",
-					"branch": "master"
+					"branch": "master",
+					"tags": true
 				}
 			]
 		},
--- a/repository/m.json
+++ b/repository/m.json
@@ -1033,6 +1033,16 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


scott1028:
add tag 1.0.0

scott1028:
update message.json

FichteFoll:
You could replace the https://packagecontrol.io/packages/Meld%20diff package, which seemingly has been deleted a while ago. Remove the corresponding entry from `channel.json` where this was referenced and then either take the name of that package or alternatively add a `previous_names` array to get users of that package to switch over to yours, if there are any still.

In your `plugin_loaded`, you should check for `os.path.exists(settings().get('meld_compare_path'))` in the first branch as you don't want to override user configuration on every restart. Actually, you could just use the manual search for the executable as a backup when no valid exeutable has been specified instead of saving the setting on disk. That way users don't need to have the settings file when they use the default location anyway. Also, I believe that you should prioritize the 64 bit version.

`is_windows` computes a static value that never changes, so you could just assign that to a variable directly instead of wrapping it inside a function.

FichteFoll:
@scott1028 ping

PR #6872, opened 2018-01-11
Added LC2200 Assembly syntax highlighting

driazati:
Similar idea to LC3 Assembly but with different formatting/definitions

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -488,6 +488,16 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
FYI, a new syntax definition format has been added to Sublime Text which is much more readable (and also supported by PackageDev): http://www.sublimetext.com/docs/3/syntax.html Just note that it's only available for ST builds 3092 and onwards.

PR #6873, opened 2018-01-12
Add MacDictionary package.

gh640:
I tested the change in my local environment and it passed.

> 1. Used `"tags": true` and not `"branch": "master"` 

Yes.
 
> 2. Added a readme to your repository so that users (and reviewers) 
 
Yes.

[There're already some dictionary packages](https://packagecontrol.io/search/dictionary) which use web dictionary services. But there's no package which uses the MacOS Dictionary.app and I'd like to add this package.

As for the change around the line 1957,  there was trailing spaces and they were automatically removed when I edited the file with [EditorConfig](https://packagecontrol.io/packages/EditorConfig) enabled. If it's better to keep the spaces, I'll adjust the commit and please let me know.

Thank you in advance.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -34,6 +34,19 @@
 			]
 		},
 		{
+			"name": "MacDictionary",
+			"details": "https://github.com/gh640/SublimeMacDictionary",
+			"author": "Goto Hayato",
+			"labels": ["mac", "dictionary"],
+			"releases": [
+				{
+					"sublime_text": ">=3124",
+					"platforms": ["osx"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "MacDown App Menu",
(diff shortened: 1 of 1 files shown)


gh640:
> Processing package "MacDictionary"
>   - WARNING: '.no-sublime-package' is defined. Please verify that it is *really* necessary

Please let me explain this. This package requires `.no-sublime-package` since the script [`scripts/ConsultMacDictionary.py`](https://github.com/gh640/SublimeMacDictionary/blob/master/scripts/ConsultMacDictionary.py) needs to be run with Python 2.7 provided by MacOS to use the Dictionary.app service.

FichteFoll:
Whops, looks like I forgot to merge this last time around.

gh640:
@FichteFoll thank you very much!

PR #6878, opened 2018-01-14
Add package: Default File Type 3

diegotf30:
This package is a fork of https://github.com/spadgos/sublime-DefaultFileType, since it only works for Sublime 2, I made a version compatible only with Sublime 3. It allows to set a default syntax type to new windows.

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -332,7 +332,7 @@
 			"releases": [
 				{
 					"sublime_text": ">=3000",
-					"branch": "master"
+					"tags": true
 				}
 			]
 		},
--- a/repository/d.json
+++ b/repository/d.json
@@ -321,7 +321,7 @@
 			"details": "https://github.com/spadgos/sublime-DefaultFileType",
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": "<3000",
 					"branch": "master"
 				}
 			]
--- a/repository/d.json
+++ b/repository/d.json
@@ -327,6 +327,16 @@
 			]
 		},
 		{
+			"name": "Default File Type 3",
+			"details": "https://github.com/diegotf30/sublime-DefaultFileType",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"branch": "master"
+				}
+			]
+		},
+		{
 			"name": "DelayedSaveAll",
 			"details": "https://github.com/shagabutdinov/sublime-delayed-save-all",
 			"donate": "https://github.com/shagabutdinov/sublime-enhanced/blob/master/readme-donations.md",


evandrocoan:
Hi @diegotf30, 

There are a few things which can be improved:

1. Do not duplicate all the syntaxes files into the folder `Packages/DefaultFileType/Syntaxes/` because:

   1. Duplicate files on the file system is bad because when the syntax updates, your copy becomes out-of-date automatically.
   1. Increases the files system storage usage.
   1. If the user do not use one of the syntaxes you ship, the user has to modify your package to use their new syntax.

1. Do not hard code the package name, use the python variable `__package__` which has the updated package name, then even if you change the package name, your package is not broken or need changes.

1. Do not duplicate the `Default File Type` package name entry on the package control channel.

   1. Ask @spadgos if he would like to have your fixes as pull requests on his repository or if he would allow you to keep maintaining his repository, then you just need to update the original repository URL to point to your fork.
   1. Then, if this repository is not compatible with Sublime Text 2 anymore, create 2 different releases:

      1. One for Sublime Text 2
      1. Other for Sublime Text 3
      
      You could allow older users with Sublime Text 2 to use the latest version which worked on Sublime Text 2 by creating a tag like `st2-1.0.2` and configuring package control to ship only the tags with prefix `st2-` to Sublime Text 2 and keep the current unprefixed tags to the latest version of Sublime Text.

      To do this:

      1. You can set the channel file like this:
          ```json
                  {
                      "name": "Default File Type",
                      "details": "https://github.com/diegotf30/sublime-DefaultFileType",
                      "releases": [
                          {
                              "sublime_text": ">=3000",
                              "tags": true
                          },
                          {
                              "sublime_text": "<3000",
                              "tags": "st2-"
                          }
                      ]
                  },
          ```
      1. And create the tag `"st2-1.0.2"` on the latest commit support Sublime Text 2. See this link for more information on how to create git tags:

         1. https://stackoverflow.com/questions/18216991/create-a-tag-in-github-repository

      I had did this to my package which had broke on Sublime Text build 3144. You can see the commit here:

      1. https://github.com/wbond/package_control_channel/pull/6814/files
      1. https://github.com/evandrocoan/NotepadPlusPlusColorScheme/releases/tag/build3144support-1.2.4

1. On this case, do not replace the user command for creating new files with the `Default.sublime-keymap`, etc. You can create event listener for the command `new_file` and then do your package things. For example:
   ```python
   import sublime
   import sublime_plugin

   class Listener(sublime_plugin.EventListener):

       def on_new( self, view ):
           ... do your stuff
           settings = sublime.load_settings("default_file_type.sublime-settings")

           if settings.get('use_current_syntax') and view:
               syntax = view.settings().get('syntax')
           else:
               syntax = settings.get('default_new_syntax')

           # sublime.error_message("%s" % syntax)
           view.set_syntax_file(syntax)
   ```

   You can consult the Sublime Text API on https://www.sublimetext.com/docs/3/api_reference.html

diegotf30:
Hi @evandrocoan,

First of all, thank you so much for all your help! Especially on the first point - I thought you needed explicit access to the `.sublime-syntax` file, since the Sublime API is a little outdated only shows an example with a `.tmLanguage`.

I've made most of the changes you've recommended, and contacted @spadgos to merge my fork.

PR #6879, opened 2018-01-14
Added Package 'Better Whitespace'

voltavidTony:
'Better Whitespace' package added to repository

When going through the settings for Sublime, I noticed the trim whitespace and ensure trailing newline settings. However, some of them didn't work in the way I thought they would, so I made my own plugin to handle these tasks instead. It is the first plugin I created, and its simplicity made it easy and fun to make, while also allowing me to self-teach the use of Sublime's API.
The features of this plugin are: Remove trailing whitespace at EOL, ensure a singular empty newline at EOF, two view-specific settings to toggle their activation on saving a file, an additional setting to ignore lines that the cursor or selection is on, an additional command to remove empty or whitespace only lines, and an override snippet. By default, the first two whitespace commands are enabled to run on save. The defaults for the settings can be set in the user preferences file.

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -540,6 +540,17 @@
 			]
 		},
 		{
+			"name": "Better Whitespace",
+			"details": "https://github.com/voltavidTony/better-whitespace",
+			"description": "Plugin to cleanup whitespace in your file",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "master"
+				}
+			]
+		},
+		{
 			"name": "BetterFindBuffer",
 			"details": "https://github.com/aziz/BetterFindBuffer",
 			"labels": ["find results", "search"],
--- a/repository/b.json
+++ b/repository/b.json
@@ -546,7 +546,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"branch": "master"
+					"tags": true
 				}
 			]
 		},


evandrocoan:
Hi @voltavidTony,

1. After loading the user settings, you are setting they back and saving the file:
    ```python
        s.set("better_whitespace_empty_line_at_eof", eofsetting)
        s.set("better_whitespace_remove_whitespace_at_eol", eolsetting)
        s.set("better_whitespace_ignore_active_lines", ialsetting)
        ...
        sublime.save_settings("Preferences.sublime-settings")
    ```
    https://github.com/voltavidTony/better-whitespace/blob/master/Better%20Whitespace.py#L15-L22

    This is forcing these setting to be put on the user setting file, even if they are not there (bad). But if you just remove these lines out:
    ```python
        s.set("better_whitespace_empty_line_at_eof", eofsetting)
        s.set("better_whitespace_remove_whitespace_at_eol", eolsetting)
        s.set("better_whitespace_ignore_active_lines", ialsetting)
    ```
    Then, this behavior would not happen.

1. Also, it is required to disable the user settings `ensure_newline_at_eof_on_save` and `trim_trailing_white_space_on_save` in order to use your package?

1. Did you searched for existent packages before writing this pacakge? Perhaps you could contribute to some of these packages instead of submmiting a new one.

1. How much does it differ from these packages already existent on Package Control when seaching for https://packagecontrol.io/search/whitespace on Package Control:
   i. https://packagecontrol.io/packages/Whitespace
   ii. https://packagecontrol.io/packages/Trimmer

1. What is this snipped for?
   i. https://github.com/voltavidTony/better-whitespace/blob/master/Better%20Whitespace.sublime-snippet

voltavidTony:
Hi @evandrocoan,
1. I am doing this on purpose. It is so that the user can go into his preferences file and change the default values. If I didn't do this, then the user would have to know that in order to disable the empty line setting he would have to know how to write "better_whitespace_empty_newline_at_eof". This way the setting is already present in the settings file. (Toggling the setting from the menu changes the settings for each view individually.)
2. In my opinion, this plugin would be useless if it didn't "override" the built-in behavior. You wouldn't still use an old lawnmower to mow the lawn and then mow it again with your new one, you'd use just the new one.
3. I was actually expecting that. I wanted to make my own plugin as a practice most of all. Since I didn't know how to use the Sublime API (since this is my first Sublime plugin), I looked at other plugins on how to find and edit parts of the text, and so I found "Whitespace" by randy3k, which brings me to point 4.
4. The features are in fact inspired by an already existing package "Whitespace" by randy3k, so I guess you can say I just contributed to an existing project.
5. Like the readme states, this snippet is the file specific disabling of the on save execution of the commands. I used it so that the plugin wouldn't accidentally edit itself while I was developing the patterns. I left it in so that the user can disable the automatic execution and not have to worry about turning it off every time he loads a file.

Also, the plugin file is 69 lines long.

voltavidTony:
So final edit:
1. The default settings are written to the preferences file only if they are not present. Otherwise, the file is not modified, opened or closed.
2. The built-in features are no longer disabled, instead, a comment recommends to disable these.
3. I will treat this plugin as a contribution to randy3k's Whitespace.

PR #6884, opened 2018-01-16
Add Extract Text to Laravel Blade View

MohannadNaj:
from the README:

This plugin for Sublimee Text 3 allows you to extract the selected text into a new or existing laravel blade views.

## Usage

Select some text, open the Command Palette, and look for "Extract: Move selection to blade view ..."

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -1589,6 +1589,16 @@
 			]
 		},
 		{
+			"name": "Extract Text to Laravel Blade View",
+			"details": "https://github.com/MohannadNaj/sublime-extract-to-blade",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
You should use `os.path.join` and similar functions to modify file paths in a cross-platform way.

MohannadNaj:
Thanks @FichteFoll .. yes I will.

If I may ask, what will happen if I rebased the same current latest tag version to a new updated forced commit?

After some refactoring, I preferred to start over and start with a new 1.0.0.

FichteFoll:
Users would not receive the update, but new users would get the newer version. Once you pushed a new version and Package Control has picked it up (which usually happens within an hour), you should not re-upload the tag anymore. You may remove it to prevent further users from downloading a broken package or just push a new fixed version, but you should not replace it.

You can either push 1.1.0 or 2.0.0 once your rewrite is done.

PR #6885, opened 2018-01-16
commit Plugin

xiaozzjl:
Using this plug-in makes it easier to require a lua file with a deeper path, or to create a specifies format code template.

### DIFF
--- a/repository/x.json
+++ b/repository/x.json
@@ -66,6 +66,16 @@
 			]
 		},
 		{
+			"name": "XiaozLuaPlugin",
+			"details": "https://github.com/xiaozzjl/XiaozLuaPlugin",
+			"releases": [
+				{
+					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


FichteFoll:
You should explain in your readme how to use the package and what it actually does. Currently it only contains a summary. Surely your potential users don't want to read the source code to understand that.

You should look into using `sublime.cache_path()` for your cache path. Note that this method is not available at module load time and can only be used after `plugin_loaded` was called. As is `sublime.packages_path()`, which you use already. You'll want to remove your cache path when the package is uninstalled (or disabled) as well, so you should clean up in `plugin_unloaded` (`unload_handler` on ST2 iirc). You should probably also remove the debug prints for these paths.

FichteFoll:
The repository doesn't exist anymore. Closing as timeout.

PR #6888, opened 2018-01-18
Add RemoveLineBreaks plugin.

fsaad:
New package hosted at https://github.com/fsaad/Sublime-RemoveLineBreaks/

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -1035,6 +1035,16 @@
 			]
 		},
 		{
+			"name": "RemoveLineBreaks",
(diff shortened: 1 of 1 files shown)


FichteFoll:
A view can have no selection, in which case `view.sel()[0]` errors.

You should always use `\n` as the newline character for text to be extracted and added to ST views. ST unifies them to `\n` internally and writes them back according to the detected original line endings or user configuration.

You also need to fix the invalid JSON file.

fsaad:
@FichteFoll Thanks for the feedback, I have addressed the three items you mentioned and pushed a new release with tag 1.0.1.

PR #6889, opened 2018-01-19
Add CWL syntax highlight

manabuishii:


FichteFoll:
You must limit installation to a build higher than or equal to 3092.

PR #6891, opened 2018-01-20
Add natsort package to dependencies.

alimony:


FichteFoll:
Please refer to some example dependencies and the docs on how a dependency needs to be structured: https://packagecontrol.io/docs/dependencies

alimony:
I've updated this PR to point at a properly structured repo.

PR #6892, opened 2018-01-21
Added Package Startup Files

voltavidTony:
I use text files to write down my to-do lists, project ideas, etc., and in order to not have to search them every time I run Sublime, I have created this plugin to take care of that.
What this plugin does: even if you close a file, it still opens when you run Sublime again if it is specified.
You can specify these files manually in the preferences file or run the command to set the files from the context menu (refer to the readme).

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -3301,6 +3301,18 @@
 			]
 		},
 		{
+			"name": "Startup Files",
+			"details": "https://github.com/voltavidTony/Startup-Files",
+			"labels": ["file", "open", "startup"],
+			"releases":
+			[
+				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
Do you not use projects or the `hot_exit` setting (which is `true` by default)?

voltavidTony:
Yes, I do. The purpose of this plugin is to ensure the specified files are always opened when sublime starts. So if I close Sublime with a file open, the `hot_exit` will reopen it, but if I closed the file, then `hot_exit` won't reopen it, but the plugin will.
It's like setting the default tabs on chrome.

PR #6893, opened 2018-01-22
Added syncrow package info

Afzal7:
I created this plugin to ease the process of creating snippets and syncing them into my various machines. Details are in the repo readme file. Thanks.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -5179,7 +5179,7 @@
 		},
 		{
 			"name": "Syncrow",
-			"description": "Create and Sync Snippets",
+			"description": "Create and Sync Snippets.",
 			"details": "https://github.com/Afzal7/syncrow",
 			"issues": "https://github.com/Afzal7/syncrow/issues",
 			"homepage": "https://syncrow.herokuapp.com",
--- a/repository/s.json
+++ b/repository/s.json
@@ -5178,6 +5178,21 @@
 			]
 		},
 		{
+			"name": "Syncrow",
+			"description": "Create and sync snippets",
+			"details": "https://github.com/Afzal7/syncrow",
+			"issues": "https://github.com/Afzal7/syncrow/issues",
+			"homepage": "https://syncrow.herokuapp.com",
+			"author": "Afzal7",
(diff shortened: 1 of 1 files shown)


Afzal7:
I have made the necessary changes, how do i rerun the bot?

FichteFoll:
1. It's kind of obvious, but you should mention in your readme that the snippet's contents are synced using a third-party remote (i.e. your server) and the contents can be read there.

2. You should use `os.path.join` when combining paths instead of hardcoding forward slashes.

3. I strongly suggest storing your snippets inside a subfolder of the User package instead of directly in the Packages folder. That one is supposed to contain whole packages (read: folders) and not individual resource files. A subfolder in `sublime.cache_path()` might work as well, but that needs to be tested.

4. `sublime.status_message` has been deprecated. Use `Window.status_message` instead.

5. You can infer a view's window object through its `.window()` method instead of using `sublime.active_window()`.

6. The `sublime.platform` print in your `plugin_loaded` can be removed.

7. I also suggest supplying a `.sublime-commands` file instead of a weird key binding (ctrl+shift+x seems quite arbitrary) because those are basically conflict-free.

8. Two menu items in the same menu and the same file should never have the same id.

9. I suggest splitting your command into two separate commands because they do fundamentally different things depending on a single argument.

FichteFoll:
@Afzal7 ping

Afzal7:
Hi @FichteFoll 

1. Readme updated.
2. `os.path.join` added.
3. Created a new folder _Syncrow Snippets_ for storing the snippets.
4. Using window.status_message
5. Still using `sublime.active_window()`, will update over time.
6. Cleaned `plugin_loaded` function
7. Left the shortcuts as `ctrl+shift+x `. I will update these in some time.
8. Given separate ids for both the menu options.

I will keep on updating the plugin over time. I hope, for now, everything is good to go. 

Thanks.

FichteFoll:
Now you just need to push a new tag with the latest changes.

Afzal7:
@FichteFoll  Done!

Afzal7:
@FichteFoll Thanks for merging. Appreciate it.

PR #6894, opened 2018-01-23
Add QNav package

ta-tikoma:

### DIFF
--- a/repository/q.json
+++ b/repository/q.json
@@ -69,7 +69,7 @@
 		{
 			"name": "QNav",
 			"details": "https://github.com/ta-tikoma/QNav",
-			"labels": [ "navigation", "files navigation"],
+			"labels": ["navigation", "files navigation"],
 			"releases": [
 				{
 					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Your code uses a weird mixture of foward and backward slashes for paths in various places. You should unify them at some point to the platform's default (preferrably directly after user input) and then use `os.path` methods to join and split them.

ta-tikoma:
Thanks, I unify slash. See:
https://github.com/ta-tikoma/QNav/blob/8f35092df0ed77d360e5d4229296eabae228b920/QNav.py#L9
and
https://github.com/ta-tikoma/QNav/blob/8f35092df0ed77d360e5d4229296eabae228b920/QNav.py#L53

FichteFoll:
This is not at all what I said.

You can retrieve the platform's pathsep via `os.path.sep`, though you'll rarely need it. Instead, use other methods of the `os.path` module like `os.path.join` or `os.path.split`.

ta-tikoma:
Oh, Thanks again.
I updated my code.
This is really better.

PR #6898, opened 2018-01-27
Add Zig Language

jfo:
Hi! I'd like to submit this syntax highlighting package for [Zig](http://ziglang.org/). In the course of submitting a PR to add this syntax to the github/linguist project, it made sense to make it available as a sublime package, as well. The package is [here](https://github.com/zig-lang/sublime-zig-language).

I haven't tagged a release yet. Does package control pick up on the most recent semver tag and serve that? 

Thanks!

Edit: I just went ahead and tagged this 1.0.0.

### DIFF
--- a/repository/z.json
+++ b/repository/z.json
@@ -131,6 +131,17 @@
 			]
 		},
 		{
+			"name": "Zig Language",
+			"details": "https://github.com/zig-lang/sublime-zig-language",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Zimpl",
 			"details": "https://github.com/joneshf/sublime-zimpl",
 			"labels": ["language syntax"],


jfo:
Thanks for the feedback, I will address it when I have some time, this pr is not abandoned.

FichteFoll:
@jfo any update?

jfo:
@FichteFoll Sure, I should have a little time to address your comments this weekend. Sorry it feel through.

jfo:
Hi, I've incorporated your comments [here](https://github.com/zig-lang/sublime-zig-language/commit/e379bc4c92cd26f651b5c8c9a5ca3d390ea957a3)

But I don't understand what you mean by:

> entity.name scopes should be used for the name that is created, not the keyword that denotes it. For keywords, a keyword scope should be used.

as [this](https://www.sublimetext.com/docs/3/scope_naming.html#entity) seems to indicate using these in this precise way:

```
The names of data structures will use one of the following scopes, or a new sub-scope of entity.name ... 
...
entity.name.struct
entity.name.enum
entity.name.union
...
```

could you elaborate? 

Sorry it took me a while to get back to this.

FichteFoll:
`entity.name` scopes are supposed to be used for the *definition* of an identifier, e.g. a function name. Usages of an identifier that is more or less known to be a function (mostly in a function call) should then be `variable.function` if it is a self-defined name, or `support.function` if it is provided by the language or its standard library.

In your [`structure` context](https://github.com/zig-lang/sublime-zig-language/blob/3ae4049070a544ea0da7adf810f1002050b508bd/Zig.YAML-tmLanguage#L38-L45), you assign `entity.name` to keywords like `union` which should rather get a `storage.type` scope, as far as I can tell.
Similarly, the [`keyword` context](https://github.com/zig-lang/sublime-zig-language/blob/master/Zig.YAML-tmLanguage#L67-L69) seems to match provided functions, which should get `support.function`.

jfo:
Ah, I understand, thanks.

Updates to your suggestions have been made. Should I reset the tag to the current head or release a patch?

PR #6900, opened 2018-01-28
Add ElixirFormatter package

karolsluszniak:
Adds automatic formatting on save via Elixir formatter recently released in Elixir 1.6.

This is my first package release for Sublime and I'm not a regular Python developer so looking forward to any comments and change requests. Thank you.

In particular, I hope the way I've used threading goes ok with the way Sublime executes Python. It's threaded in order not to block editor during the formatting which takes ~0.5s after save, and yet to wait for formatting command result in order to log errors to Sublime console.

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -463,6 +463,16 @@
 			]
 		},
 		{
+			"name": "ElixirFormatter",
+			"details": "https://github.com/karolsluszniak/elixir-formatter-sublime",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "ElixirPlayground",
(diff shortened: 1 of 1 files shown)


karolsluszniak:
Updated PR description with some notes that may be helpful in the review. Thanks.

FichteFoll:
Using `os.getcwd()` to determine the project folder is a bad idea because that will usually not change once ST has been launched, and ST can have multiple projects with different folders or might have been launched in the home dir. The code segment is effectively dead because you always pass a parameter, but I suggest removing it anyway before you start getting weird ideas later.
`Window.folders()` might also yield interesting results, though going through a file path's parents will be safer. (I also recommed doing it iteratively, although you won't go through that many iterations.)

PR #6902, opened 2018-01-28
Add the 'choo snippets' package

kareniel:


FichteFoll:
You should always use tabs for snippets. Using tabs allows ST to convert them to the user's preference when inserting.

kareniel:
FichteFoll: Got it! Thank you :)

PR #6908, opened 2018-01-31
CSS Grid Snippets port for Sublime Text

antenando:

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3732,7 +3732,7 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
You don't need that "Sublime" in your package name there. While it's not as severe as having the name start with it, it is entirely redundant in a package index specifically *for* Sublime Text. You may keep it in your repo, but please remove it from the `name` key in this PR.

antenando:
Done.

PR #6913, opened 2018-02-03
Added the package FixCommandPalette

evandrocoan:
Package: https://github.com/evandrocoan/FixCommandPalette

Force the Sublime Text command palette to remember the input text,
even after closing it after pressing the `escape` key, fixing the issue:

* https://github.com/SublimeTextIssues/Core/issues/1814 Command palette cleans the last command on escape key press

Also forces the command palette to make the initial text selected, fixing the regression created on
the Sublime Text build 3156:

* https://github.com/SublimeTextIssues/Core/issues/2083 Command palette text not selected between commands

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -964,6 +964,16 @@
 			]
 		},
 		{
+			"name": "FixCommandPalette",
+			"details": "https://github.com/evandrocoan/FixCommandPalette",
+			"releases": [
+				{
+					"sublime_text": ">=3126",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "FixMyJS",
 			"details": "https://github.com/addyosmani/sublime-fixmyjs",
 			"labels": ["formatting"],
--- a/repository/f.json
+++ b/repository/f.json
@@ -964,16 +964,6 @@
 			]
 		},
 		{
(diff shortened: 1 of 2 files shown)


evandrocoan:
> ```
> Processing package "FixCommandPalette"
>   - WARNING: The binding ['escape'] is also defined in default bindings but is masked with a 'context'
>     - File: Default.sublime-keymap
>   - WARNING: The binding ['ctrl+shift+p'] is also defined in default bindings but is masked with a 'context'
>     - File: Default.sublime-keymap
> ```

I need to override those to control the command palette closing and opening and be able to select and save the inputs.

FichteFoll:
I think you might be able to implement that with `on_(post_)window_command` hooks too, but you'd have to do some experimentation.

FichteFoll:
I personally find the package name too general. Considering that https://github.com/SublimeTextIssues/Core/issues/2083 has been fixed in the latest build, maybe you can come up with a name that better describes what it does regarding the other issue? E.g. "RememberCommandPaletteInput".

evandrocoan:
I cannot, they are not logged by Sublime Text:

1. https://github.com/SublimeTextIssues/Core/issues/2198 The command show_overlay and hide_overlay is not logged by on_window_command

PR #6915, opened 2018-02-04
Add VSCodeColumnSelection

acheronfail:
This package adds a mouse command that makes column selection similar to VSCode.
This will make it easier for those transitioning from that editor, or for those who regularly develop with both editors and sometimes find themselves getting confused between Sublime's column selection and VSCode's.

Basically, click once at "anchor" point, and then <kbd>shift</kbd> + <kbd>alt</kbd> + <kbd>click</kbd> at the other side of the column selection.

Example gif from package's [README](https://github.com/acheronfail/VSCodeColumnSelection):

example gif

### DIFF
--- a/repository/v.json
+++ b/repository/v.json
@@ -638,6 +638,16 @@
 			]
 		},
 		{
+			"name": "VSCodeColumnSelection",
+			"details": "https://github.com/acheronfail/VSCodeColumnSelection",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Vtex Snippets",
 			"details": "https://github.com/ricardodantas/Vtex-Snippets",
 			"releases": [


acheronfail:
I get the error `"It is not advised to specify mousemaps..."`, and I understand where it's coming from, but this is primarily a mouse-command package. 

Is it alright to leave them in there, or should I rename them/add it to the readme?
Keep in mind that if I do so, then this package won't work out-of-the-box for users, they'll have to perform another configuration step before being able to use this package.

deathaxe:
I guess the warning points to the identical content of all of the OS specific mouse-maps. But if I recall correct, OSX doesn't know about the `alt` key. You need to use `option` instead.

acheronfail:
> I guess the warning points to the identical content of all of the OS specific mouse-maps. But if I recall correct, OSX doesn't know about the alt key. You need to use option instead.

I actually wrote this on macOS, and Sublime handles the `alt` key fine (on my machine at least). I'll change it anyway just in case there are older versions of Sublime which don't handle the `alt` key.

FichteFoll:
Imo this is all right, since the primary focus of the package is to provide a mouse binding. The binding itself is strongly advertised in the readme, so this gets a pass from me.

Users will have to deal with potential issues themselves if they install multiple packages that override shift+alt+click.

`alt` and `option` are just two aliases for the same key. On all platforms. The only keys with platform-dependant mappings are `super` and `primary`.

Note that there are already default bindings for this, e.g. on Linux:
```js
        "button": "button1", "modifiers": ["shift", "alt"],
        "press_command": "drag_select",
        "press_args": {"subtractive": true, "extend": true}
    },
```
however, they don't seem to be useful for anything als `alt+left` drag is much superior and predictable, so this gets a pass from me also.

This is a special occasion.

By the way, the same is also supported when holding shift and rightclick-dragging.

PR #6916, opened 2018-02-04
Renamed the dependency PythonDebugTools to DebugTools

evandrocoan:
https://pypi.python.org/pypi/debug_tools

### DIFF
--- a/repository/dependencies.json
+++ b/repository/dependencies.json
@@ -594,14 +594,14 @@
 			]
 		},
 		{
-			"name": "PythonDebugTools",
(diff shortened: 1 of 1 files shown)


FichteFoll:
I presume you're aware of the effects this will have on all your packages using this (and potentially any others, though I highly doubt anyone besides you is using it).

evandrocoan:
My packages were updated to use the new name since before I did this pull request. I have my own channel where I put all my packages and their forks.

Indeed, it would be a problem for anyone else using this, but I also do not think anyone else is using it.

PR #6917, opened 2018-02-04
Add Syntax Package: SkoolkitZ80

mrcook:
Z80 Assembly language syntax highlighting that is focused on the [Skoolkit](https://github.com/skoolkid/skoolkit) development environment.

The [z80asm](https://github.com/psbhlw/sublime-text-z80asm/tree/master/z80asm) package does handle Z80 assembly, but it is quite a heavy package that requires build tools of very specific 3rd party applications.

`SkoolkitZ80`'s intention is to remain lightweight, only providing syntax highlighting for Skoolkit files -- no build tools, and no custom menu items.

Many thanks for your consideration.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -2045,6 +2045,17 @@
 			]
 		},
 		{
+			"name": "SkoolkitZ80",
+			"details": "https://github.com/mrcook/SkoolkitZ80",
+			"labels": ["language syntax", "z80", "assembly"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "SLAB",
 			"details": "https://github.com/increpare/slab-markup-sublime",
 			"labels": ["language syntax"],
--- a/repository/s.json
+++ b/repository/s.json
@@ -2050,7 +2050,7 @@
 			"labels": ["language syntax", "z80", "assembly"],
 			"releases": [
 				{
-					"sublime_text": ">=3000",
+					"sublime_text": ">=3092",
 					"tags": true
 				}
 			]


FichteFoll:
I strongly advice against forcing users to use your bundled color scheme for your syntax. If you follow the scope naming guidelines, as you seemingly do, you shouldn't need to bother about color schemes as they will all have their individual interpretation of the scopes you used.

Look-behind patterns are inefficient because the non-backtracking sregex engine can't be used for them and instead Oniguruma is used. It is possible to express all patterns that use look behinds with proper context management, however I don't think this is even necessary for your case.

mrcook:
Thanks for your advice @FichteFoll, I'll take a look at these points over the weekend.

mrcook:
I refactored the regexp and made the change as you suggested.

I've struggled with the theme though. Following the scope naming guidelines made sense, but there's just too many subtleties needed to make the code easy to read without a custom colour scheme. so I've decided to leave that as-is.

I will however look into creating a light-theme for those who prefer that.

Thanks. Michael.

FichteFoll:
A compromise would be to include instructions on how to enable your bundled color scheme rather than using it by default. That way the user can opt-in to your scheme rather than opt-out. This will be my last attempt at persuading you.

The only occurances where bundling a scheme is pretty much a neccessity is for very distinct syntax involving background coloring etc, such as with https://packagecontrol.io/packages/BetterFindBuffer and potentially with https://packagecontrol.io/packages/PlainTasks (for the green and red marks, though those can be done with a recent *dev* build feature now as well).

mrcook:
Hi @FichteFoll, many thanks for your feedback and time on this - I do appreciate it.

It is not a case of needing to persuade me - I really would prefer not to rely on a custom colour scheme. However, the skool config is quite distinct and all the scheme I've tried make reading the text confusing.

Were the colouring to become opt-in then the initial experience would be terrible, which is the opposite of what I wanted to give people. In the light of this, it is perhaps better to close this PR.

If in the future I learn more about SublimeText development, I might find a solution to appease both us :)

Again, thanks for your time.

Cheers,
Michael.

FichteFoll:
By saying

>This will be my last attempt at persuading you.

I didn't mean that I would reject the PR.

mrcook:
Sorry, my misunderstanding. I'm happy to have the PR re-opened :)

Please let me know how you want to proceed.

mrcook:
Thanks @FichteFoll, I've removed the settings as suggested - these were added early on after using other plugins as reference.

PR #6919, opened 2018-02-05
Added the dependency concurrent-log-handler to dependencies.json

evandrocoan:
1. https://github.com/evandroforks/concurrent-log-handler 

Is a recent dependency required by:

1. https://github.com/wbond/package_control_channel/pull/6818 Added the PythonDebugTools on repository/dependencies.json

### DIFF
--- a/repository/dependencies.json
+++ b/repository/dependencies.json
@@ -89,6 +89,20 @@
 			]
 		},
 		{
+			"name": "ConcurrentLogHandler",
+			"author": "evandroforks",
+			"load_order": "09",
+			"description": "An additional concurrent file log handler for Python's standard logging package",
+			"issues": "https://github.com/evandroforks/ConcurrentLogHandler/issues",
+			"releases": [
+				{
+					"base": "https://github.com/evandroforks/ConcurrentLogHandler",
+					"sublime_text": ">=3114",
+					"tags": true
+				}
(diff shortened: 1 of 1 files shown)


FichteFoll:
I'm not sure what you are going to use this for as a dependency? Do you want specific packages to have their log messages archived or all?

By itself, this isn't harmful, but I do wonder a little about the impact on the end user having various log files created for them they didn't ask for.

evandrocoan:
This allows the debug tools dependency to not halt logging when the file rotation maximum size is reached. For example, on Windows with the standard logging module rotation file handler, the logging module will stop logging and start throwing a lot of errors when the maximum size is reached. But with this, the logging will continue as usual. Then, soon as there are no other processes locking the file, the rotation should happen.

By default, the user will not have logging written to the file system, but for example, I added this as a option on my fork of [LSP](https://github.com/evandroforks/LSP/blob/d12751e0f5b00a56318872eb46665641f3b7b6d0/LSP.sublime-settings#L255-L258) package, which has embed a maximum 50MB of log size before start rotating:
```js
  // Send the all logging statements to the following file instead of the
  // Sublime Text console. If `log_file` is not an absolute file path, it will
  // create the log file relatively to the User directory.
  "log_file": "",
```

PR #6920, opened 2018-02-05
Add Carbon package

stianmartinsen:
A simple package that sends the currently selected editor content to carbon.now.sh


stianmartinsen:
Closing this as we don't need two identical packages -> #6947

PR #6925, opened 2018-02-07
add GoDebug package

dishmaev:

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -1031,7 +1031,7 @@
 		{
 			"name": "GoDebug",
 			"details": "https://github.com/dishmaev/GoDebug",
-			"labels": ["go", "debug"],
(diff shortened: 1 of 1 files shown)


dishmaev:
Hi
Can i use a 'context' in the Default.sublime-mousemap similar to the Default.sublime-keymap?
How do I get rid of warnings?

FichteFoll:
No, you cannot specify a context for mousemaps unfortunately. I suggest adding instructions for the user in the readme. (Or using `edit_settings` to open your defaults with a different extension next to the User file.)

You should use relative imports because those are independent of the package's installed name.

FichteFoll:
@dishmaev ping

dishmaev:
Hi
I will make the necessary changes to GoDebug tomorrow

PR #6929, opened 2018-02-07
删除 ToElement，提交 ToggleReadEditMode

Sir0xb:

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -2452,6 +2452,16 @@
 			]
 		},
 		{
+			"name": "ToggleReadEditMode",
+			"details": "https://github.com/Sir0xb/ToggleReadEditMode",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "ToggleSettings",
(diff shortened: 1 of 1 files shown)


FichteFoll:
You don't even need a plugin for this. Just use a snippet.

```json
[
    {
      "caption": "Word to element",
      "command": "insert_snippet",
      "args": {
         "contents": "<$SELECTION>$0</$SELECTION>"
       }
    }
]
```

It's also quite similar to the <kbd>Edit > Tag > Wrap selection with tag…</kbd> command, although that wraps the selection and offers to type a tag for the surrounding instead. So I guess this *is* different, although it's just a simple snippet.

FichteFoll:
For the second package, there is https://packagecontrol.io/packages/ReadonlyMode already. Please open a new PR for each package in the future.

I'd also suggest adding English readmes so that a wider audience can understand what your packages do. Fortunately I can read source code, but I'm not sure everyone can.

FichteFoll:
As mentioned, https://packagecontrol.io/packages/ReadonlyMode appears to provide this exact functionality already. What value does your package provide over that one?

PR #6931, opened 2018-02-07
add syntax highlighting package for charmci files

matthiasdiener:
This package provides syntax highlighting for Charm++'s .ci files.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -666,6 +666,17 @@
 			]
 		},
 		{
+			"name": "Charmci",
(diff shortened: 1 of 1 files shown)


matthiasdiener:
Bump.

FichteFoll:
You probably want your base scope to be `source.c++.charmci` (and also apply that to your custom keywords, but that's not as important).

FYI, bumping does not help your package to be reviewed faster.

matthiasdiener:
Noted, thanks.

PR #6933, opened 2018-02-08
Sublime 3 Plugin helps to search change records of the Drupal hooks (core and contrib modules).

thamanameexusa:

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -1356,6 +1356,18 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


thamanameexusa:
Team,

Looks like my package "DrupalSearch" is added to the repository. When this package will be listed in https://packagecontrol.io 

Is there any steps to add it. Please guide here

Thanks,
Thaman

PR #6937, opened 2018-02-09
Add SublimeFcitx

ubuntugege:
Integrate Chinese/Japanese/Korean/Vietnamese input via fcitx on linux platform.

Genrate command line with fcitx
Create desktop entry with fcitx
Test pass on ubuntu and Fedora, also works well on other linux platform

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -332,6 +332,18 @@
 			]
 		},
 		{
+			"name": "FcitxInput",
+			"details": "https://github.com/ubuntugege/FcitxInput",
+			"labels": ["input", "fcitx", "chinese"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": "linux",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Feature Presentation",
 			"details": "https://github.com/ctruett/FeaturePresentation",
 			"releases": [
--- a/repository/s.json
+++ b/repository/s.json
@@ -4118,18 +4118,6 @@
 			]
 		},
 		{
-			"name": "SublimeFcitx",
-			"details": "https://github.com/ubuntugege/SublimeFcitx",
-			"labels": ["input", "fcitx", "chinese"],
-			"releases": [
-				{
-					"sublime_text": "*",
-					"platforms": "linux",
-					"tags": true
-				}
-			]
-		},
-		{
 			"details": "https://github.com/quarnster/SublimeGDB",
 			"releases": [
 				{
--- a/repository/s.json
+++ b/repository/s.json
@@ -4118,6 +4118,18 @@
 			]
 		},
 		{
+			"name": "SublimeFcitx",
+			"details": "https://github.com/ubuntugege/SublimeFcitx",
+			"labels": ["input", "fcitx", "chinese"],
+			"releases": [
+				{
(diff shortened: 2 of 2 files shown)


FichteFoll:
Please remove the "Sublime" from your package name.

ubuntugege:
> Please remove the "Sublime" from your package name.
I found many packages with "Subilme" prefix, [https://packagecontrol.io/search/sublime](https://packagecontrol.io/search/sublime)

> Also, you're leaking file handles.
Could you please show more details?

FichteFoll:
>I found many packages with "Subilme" prefix, packagecontrol.io/search/sublime

Very rare exceptions are allowed. However, most of these were added many years ago before we had this policy.

>Could you please show more details?

Any `open` call needs a `close` call on the returned file object, or you use it as a context manager with `with`. It's not as severe on UNIX systems as it is on Windows, but you should pay attention to it regardless.

An example is here: https://github.com/ubuntugege/SublimeFcitx/blob/d0f2db33ab2df9639e5d55d4caded8f8af62d60b/sublimefcitx.py#L63

ubuntugege:
@FichteFoll 
Thanks for your explanations and suggestions. 
The problems you pointed out have been corrected.
Please review again.

Thom1729:
FYI, if you're using `with`, you don't need an explicit `close`. For example:

```python
def file_get_contents(filename):
	with open(filename, "r") as f:
		return f.read()
```

In `run_create`, you're checking the platform and returning if it isn't Linux. You should probably raise an exception in that situation.

It looks like you have a lot of code specifically supporting ST2. ST2 is pretty ancient at this point; do you expect that many potential users will still be using it? Are you specifically testing on ST2?

You're using `file_get_contents` in several places to get the text of a file within your package. Wouldn't it be simpler to use (e.g.) `sublime.load_resource('Packages/FcitxInput/sublime_text_fcitx.sh')`?

You're modifying a lot of files inside your package (for example, in `make_launch_shell`). I wonder if it would be better to place the modified files inside `Packages/User/FcitxInput` instead? That way, you wouldn't need to use `.no-sublime-package`. I'm not sure that there's a perfect solution without downsides.

ubuntugege:
> FYI, if you're using with, you don't need an explicit close.

Fixed.

>In run_create, you're checking the platform and returning if it isn't Linux. You should probably raise an exception in that situation.

Fixed.

>It looks like you have a lot of code specifically supporting ST2. ST2 is pretty ancient at this point; do you expect that many potential users will still be using it? Are you specifically testing on ST2?

I like ST2 very much, and pleased to make compatibility for ST2. I has made full testing on ST2.

>You're using file_get_contents in several places to get the text of a file within your package. Wouldn't it be simpler to use (e.g.) sublime.load_resource('Packages/FcitxInput/sublime_text_fcitx.sh')?

`sublime.load_resource()` not work for ST2.

>You're modifying a lot of files inside your package (for example, in make_launch_shell). I wonder if it would be better to place the modified files inside Packages/User/FcitxInput instead? That way, you wouldn't need to use .no-sublime-package. I'm not sure that there's a perfect solution without downsides.

Shortcut need load `sublime_fcitx.x32.so` or `sublime_fcitx.x64.so` to work. It should be stay on `.no-sublime-package`.

PR #6941, opened 2018-02-10
Add File Icons

braver:
Adds a simpler alternative to [A File Icon](https://packagecontrol.io/packages/A%20File%20Icon):

- Doesn't introduce "alias" languages like "Javascript (Gulpfile)"
- No runtime code: no manipulation of settings or themes, no restarting, no "zzz" folder shenanigans
- No configuration: just uses existing theme override behaviour
- Easier to understand and maintain

It's not that A File Icon is bad or unmaintained (well, it isn't maintained really, but whatever). IMO it's severely over engineered and does things that are really unwise. It messes with user space stuff like settings for other packages and it's super hard to contribute to. A lot of people probably won't care, but I do and sometimes the community is better if it has some choice.

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -589,6 +589,17 @@
 			]
 		},
 		{
+			"name": "FileIcons",
+			"details": "https://github.com/braver/FileIcons",
+			"labels": ["file", "icons"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "FileList",
 			"details": "https://github.com/shagabutdinov/sublime-file-list",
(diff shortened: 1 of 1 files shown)


evandrocoan:
Is there some unadvantage of using `File Icon`  instead of `A File Icon`?

braver:
A disadvantage you mean? 

- A File Icon lets you configure monochrome icons of any color, this only has multi color icons. 
- Other themes can be made to be compatible with A File Icon, this only supports ST’s default themes. 
- A File Icon has different icons for e.g. Gruntfile.js and gulpfile.js, with this they’re just JavaScript files so they get a js icon.

PR #6942, opened 2018-02-11
Add Lebab ES6 Transform

uniibu:
This package transforms js codes to ES6 using Lebab.

It also bundles the Lebab module using `zeit/pkg`, therefor there is no longer a need to install nodejs or lebab it self. 

It also has additional features such as `format_on_save` and `show_warnings` aside from `transforms`

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -575,6 +575,17 @@
 			]
 		},
 		{
+			"name": "Lebab ES6 Transform",
+			"details": "https://github.com/uniibu/Lebab-ES6-Transform",
+			"labels": ["lebab", "es5", "es6", "convert", "transform"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Ledger syntax highlighting",
 			"details": "https://github.com/moeffju/sublime-ledger-syntax",
 			"releases": [


FichteFoll:
Because you run exectuables bundled in your package, you need to provide a `.no-sublime-package` file that tells PC to unextract the package on installation. Otherwise you will not be able to access them.

If you get handed a `view` object in a callback, you should use it and not pick your own view from `sublime.active_window().active_view()` (in `isJavascript`).

You can use `traceback.print_exc()` to print a traceback in exception handlers.

FichteFoll:
@uniibu ping

uniibu:
Ah, sorry i missed the notification. Will update asap. Thanks

uniibu:
@FichteFoll Added your suggestions. Also added a new release 0.0.2 [Lebab-ES6-Transform - releases](https://github.com/uniibu/Lebab-ES6-Transform/releases/latest)
```
Lebab-ES6-Transform 0.0.2 changelog

CHANGES/FIXES

+ added .no-sublime-package
+ fix source code, reuse view instead of calling sublime.active_view
+ use traceback.print_exc() to print traceback on sublime console
+ fix readme.md's settings example
```
- I've also updated my fork and merged the upstream master.
- Added `platform` and specific sublime build range 994c250
- Revert and remove the `platform` as suggested since the package is compatible to all platform. 
- Accidentally changed a different `lebab` package cdb6720 while doing the above.
- Reverted that accident and edit the correct package 11c9f75

Thom1729:
To determine whether a given view contains JavaScript, why not simply check `view.match_selector(0, 'source.js')`? That's the most common way of doing it, and it might be less surprising unless you have a specific reason for doing it differently.

There are a lot of print statements. Are they all supposed to be there in the release version?

Are you sure you want to show an error popup if the command fails? You're already printing it to the console. (Or, if you removed the try-except entirely, it will still print the exception to the console.)

uniibu:
@Thom1729 Thanks for the suggestions. As a NodeJs dev, I only have basic knowledge in regards to developing sublime plugins (I just basically use this personally, and just wanted to share it), tho I try my best to read up and make it work, so pardon me for any 'unconventional' way of doing things, and I really appreciate the suggestions. 

- For the Javascript check, you are right, I just tested `view.match_selector(0, 'source.js')` locally and it works perfectly. Thanks again for this (I learned something new).
-  For print statements. It actually only prints 1 message which is `Lebab executable is: /path/to/executable`
- For error popup, thanks again for this. I forgot to remove `sublime.error_message(str(sys.exc_info()[1]))` which was supposed to be just for development.

I'll push some updates today. Thanks again!

Thom1729:
This is perhaps a nitpick, but in `execute`, what's the point of catching `OSError` only to wrap it in `Exception`?

PR #6944, opened 2018-02-12
Renamed 'Theme - Hsiang' Refined to 'Theme - Pure'

xianghongai:
Renamed 'Theme - Hsiang' Refined to 'Theme - Pure'。

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -1207,18 +1207,6 @@
 			]
 		},
 		{
-			"name": "Theme - Hsiang",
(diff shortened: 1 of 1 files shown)


jrappen:
@xianghongai you can use [PackageDev](https://github.com/sublimetext/PackageDev) to convert the color scheme(s) to the **new** [color scheme format for ST3149+](https://www.sublimetext.com/docs/3/color_schemes.html)

PR #6945, opened 2018-02-13
Add CoNLL-U.tmLanguage to packages

stephsamson:
Syntax highlighting for [CoNLL-U files](http://universaldependencies.org/format.html).

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2988,6 +2988,16 @@
 			]
 		},
 		{
+			"name": "CoNLL-U.tmLanguage",
+			"details": "https://github.com/stephsamson/CoNLL-U.tmLanguage",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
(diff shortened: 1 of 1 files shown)


repository/c.json
```diff
@@ -2987,6 +2987,17 @@
 				}
 			]
 		},
+		{
+			"name": "CoNLL-U.tmLanguage",
```
wbond: This should probably just be `CoNLL-U`.

repository/c.json
```diff
@@ -2987,6 +2987,17 @@
...
+			"details": "https://github.com/stephsamson/CoNLL-U.tmLanguage",
+			"labels": ["conllu", "ud", "universaldependencies", "language syntax"],
```
wbond: The labels are for finding related packages. There probably won't be more `conllu` packages. There also don't seem to be any `ud` or `universaldependencies`, and it wasn't clear to me what these were or how they were related.

PR #6948, opened 2018-02-17
Add Polygen-Sublime Package

tajmone:
New ST3 package for Polygen syntax.

Passed all tests (local and remote) with ChannelRepositoryTools.

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -1605,6 +1605,19 @@
 			]
 		},
 		{
+			"name": "Polygen",
+			"author": "Tristano Ajmone",
+			"details": "https://github.com/tajmone/sublime-polygen",
+			"issues": "https://github.com/tajmone/sublime-polygen/issues",
+			"labels": [ "polygen", "language syntax" ],
+			"releases": [
+				{
+					"sublime_text": ">=3149",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "polyglot Snippets",
 			"details": "https://github.com/thespacedoctor/polyglot-Sublime-Snippets",
 			"releases": [


FichteFoll:
Please see my comment regarding forced default color schemes: https://github.com/wbond/package_control_channel/pull/6917#issuecomment-373226322. This also applies to modifying a user's settings in general, i.e. `match_brackets`. Unless they are the only thing that makes sense (e.g. a syntax prohibiting usage of tabs like YAML), don't mess with a user's settings. Not even for your syntax.

Specifying the `extensions` setting is redundant if it's also in the syntax's `file_extensions` list.

tajmone:
Thanks for the advise, this is my very first package I've created so I wasn't aware of the guidelines you mentioned regard settings (I've worked my way through by reading the docs and looking at examples).

I'll revise the project and remove all settings which are not strictly necessary, and the redundant extension settings. 

As for the color scheme, the reason why I imposed the custom scheme by default is because this syntax uses some uncommon scopes to cover BNF2-like elements (eg: `non-terminal`). I've looked at other BNF/EBNF extensions (TextMate and Sublime Text) and I've seen that some extensions do use as scopes the actual BNF terms like terminal, non-terminal, etc.; so I thought it would be best to use meaningful scope names instead of falling back on some more common but not-specific scope — some BNF syntaxes opted to use scopes like `variable` or `constant` for BNF symbols, just in order to have them covered by most color schemes. I'd rather use scopes which are meaningful for the syntax.

The problem with this is that these scopes are not covered by ST default templates, so unless I created a custom scheme for this syntax, its core elements wouldn't be highlighted (and there aren't many elements to this syntax after all); and this is the reason why I've created two specific color schemes and set one of them as default settings. 

So, at the end of the day, the user will have to manually associate this syntax with one of the dedicate schemes that ship with it in order to enjoy basic highlighting of its code. Wouldn't these be good reasons to make an exception for the scheme settings in this case? For example, the markdown editing package also automatically sets a default scheme for markdown documents, which is required in order to correctly display text stylings.

Or should I just advise the user, in the package documentation, to manually associate the provided color scheme to the syntax?

tajmone:
Sorry, I'll bother you with a couple of more questions:

- Can I just comment out the current settings which have to be removed, so that users can use them as references to copy in their user settings? I'd like to specify that they are the advised settings for working with this syntax (from my experience with the syntax, different settings would make the editing experience unpleasant with this kind of syntax).

- Can the tabs/indentation settings be left as they are? they are not mandatory, but different settings cause editing problems. Ie, this block of settings:

```json
    "tab_size": 2,
    "translate_tabs_to_spaces": true,
    "detect_indentation": false,
    "use_tab_stops": true,
    "auto_indent": true,
    "smart_indent": true,
    "trim_automatic_white_space": true,
    "indent_to_bracket": true,

    "auto_complete_commit_on_tab": true
```

FichteFoll:
> this is my very first package I've created so I wasn't aware of the guidelines you mentioned regard settings

Yeah, I don't think they are mentioned there. At least they are not in the unofficial documentation, although they should. My time is limited, unfortunately.
Fortunately for you, however, all packages are reviewed (by me) before they are available for download, so that we can fix these misconceptions or other problems before users run into it.
Unfortunately, I'm the only person doing these reviews and as you can probably tell, they take quite some time, which is why the current backlog is quite huge.

>some BNF syntaxes opted to use scopes like variable or constant for BNF symbols

Sounds sensible to me. Non-terminals can be though of as variables, right?

Rule definitions (of non-terminals) get a variant of `entity.name`, usages of such rules get `variable`. Terminals should probably get `string.unquoted`.

By variant I mean you can specify more segments after the initial segments so that they are sure to be highlighted in some way by a color scheme. For example, `entity.name.constant.non-terminal.polygen`. It doesn't really make sense to include "non-terminal" anywhere in there because, afaics, Polygen doesn't have another form of "variable", but if other BNF syntaxes were to use the same scope until there, e.g. `entity.name.constant.non-terminal.enbf`, then color schemes could target both of these with `entity.name.constant.non-terminal.enbf` without matching a programming language's constant definitions.

> I'd rather use scopes which are meaningful for the syntax.

The entire point of scope names is to convey *semantic* meaning so that the syntax elements of various different languages can be colored in similar fashions by a few standardized selectors without issues. You can always get more specific, which is why syntaxes usually add a final segment with their name to scopes so they can be targeted *very specifically*, but the decision to do so is left up for the color scheme.

>Can I just comment out the current settings which have to be removed, so that users can use them as references to copy in their user settings?

Yes, that is fine.

>Can the tabs/indentation settings be left as they are? they are not mandatory, but different settings cause editing problems. Ie, this block of settings:

- Usually, `detect_indentation` does the right thing most of the time.
- I don't see a reason for forcing a tab-width of 2, nor do I think messing with the user's choice to disable `smart_indent` is good. It's enabled by default, so they probably disabled it for a reason.
- The same goes for other settings like `auto_indent` and `indent_to_bracket`. 
- `trim_automatic_white_space` could be argued about, but I don't see a reason to treat it any different from other languages. I don't want trailing whitespace neither in Python, nor in some BNF, but I'm sure there are people who do (for whatever reason). 
- `auto_complete_commit_on_tab` is a no-go.

Changing the user's settings without a very good reason is just not something you should consider. Personally, I would complain about that vocally.

tajmone:
@FichteFoll thanks for the detailed explanations

I've added 2 commits, the 1st one deals with the settings (all settings are now commented out and indicated as a reference for the user, and removed the redundant file extensions), the 2nd one adds a new snippet (I though I just as well take the chance and add a new feature I had been working on).

I've also created a new release (`1.1.0`):

https://github.com/tajmone/sublime-polygen/releases/tag/1.1.0

I just wanted to make sure I understood correctly the scopes discussion above (as I haven't changed scopes in the syntax since the initial release).

Currently, the syntax relies on these scopes (they are not many in total):

- __Labels__: they always get the `variable.other.label.polygen`, and labels definitions get an additional `entity.name.label.polygen` (labels in this context are not like Goto labels in Basic, or off-set placeholder like in Assembly, here they have a use of their own which I can't pin-down to any of the scopes listed in documentation)
- __Non-Terminals__: `variable.other.non-terminal.polygen`, and in definition rules they get an additional `entity.name.non-terminal.polygen`
- __Strings__ (just normal strings, with some custom escape rules)
- __Operators__: nothing special here, except that some operators get an extra scope added (eg: `keyword.operator.probability.polygen`)

I haven't provided any scope for __Terminal__ symbols, as there doesn't really seem to be a need for it — strings are just treated as strings, and operators as operators, and basically if it's not a non-terminal then it's a terminal.

You mentioned:

> It doesn't really make sense to include "non-terminal" anywhere in there because, afaics, Polygen doesn't have another form of "variable"

you're right, but if it doesn't harm I'd like to keep it for potential use in plugins that might be developed in the future. If I udnerstood corerctly, color schemes should be happy with the `variable` part of the scoping, and ignore the rest.

I'm happy with current scopes definitions, my original worry was the fact that the syntax might not look too good with other color schemes, this being the reason I had originally enforced the color scheme that ships with it. But it doesn't look too bad either, and most likely users will end up using one of the custom schemes.

So, if the scopes I've provided are good, and the other changes now meet the requirements, I'm happy to go with the v1.1.0 release as the entry point for this package in Package Control. If you advise me to do some more changes, then we can keep it on hold until I polish it further.

I don't think this package will be undergoing many updates in the future. Polygen's author is working on an update of the original Polygen application, right now, but I think upcoming chances won't affect this package, except maybe for Builds (in case new options are introduced).

Thanks again.

PR #6959, opened 2018-02-21
rename package MBXTools -> PreTeXtual

daverosoff:

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -957,18 +957,6 @@
 			]
 		},
 		{
-			"name": "MBXTools",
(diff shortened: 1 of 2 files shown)


FichteFoll:
Please address the `.sublime-syntax` warning.

daverosoff:
Do I need to create a `.tmLanguage` equivalent of the `.sublime-syntax`? This package has never had a `.tmLanguage` file and it wasn't a problem before. Is it enough to change the package so that it is ST3-specific? Thanks.

PR #6960, opened 2018-02-22
adding lazy unittester package entry

connor-philip:
### Lazy Unittester
This package creates python unittest stubs from a file containing your functions.
More details and example can be found [here](https://github.com/connor-philip/lazy_unit_tester).

I haven't updated the readme of the repo to include plugin details as of yet, as it's not officially an available plugin in package control. I plan to update the readme if this package is accepted.

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -466,6 +466,17 @@
 			]
 		},
 		{
+			"name": "Lazy Unittester",
+			"details": "https://github.com/connor-philip/lazy_unit_tester",
+			"labels": ["Python", "Unittesting"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "LazyTimeTracker",
 			"details": "https://github.com/Bwata/LazyTimeTracker",
 			"labels": ["Time", "tracking"],
--- a/repository/l.json
+++ b/repository/l.json
@@ -468,9 +468,8 @@
 		{
 			"name": "Lazy Unittester",
 			"author": "Connor Philip",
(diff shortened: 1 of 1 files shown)


repository/l.json
```diff
@@ -465,6 +465,19 @@
 				}
 			]
 		},
+		{
+			"name": "Lazy Unittester",
+			"author": "Connor Philip",
+			"description": "Creates Python unittest stubs",
```
wbond: You can remove this line and it will pull the repo description from your GitHub repo. This makes it easier to tweak the description in the future.

repository/l.json
```diff
@@ -465,6 +465,19 @@
...
+			"details": "https://github.com/connor-philip/lazy_unit_tester",
+			"labels": ["Python", "Unittesting"],
```
wbond: You'll want to change the label to `python` lowercase, since all but 1 package is under that label. You probably want `Unittesting` to be `testing`: https://packagecontrol.io/browse/labels/testing.

FichteFoll:
Your imports need to be relative because the package will be installed as "Lazy Unittester" and importing from `lazy_unit_tester` won't work.

Thom1729:
I don't understand what `setup.py` is for. It looks like it's interacting with some other package system. Is this repository designed to work with another package system in addition to Sublime?

connor-philip:
Sorry for the late reply, I was on holiday. The setup.py file was for the TravisCI tests.
The repo is designed to work with itself stand alone and also as a sublime plugin. I realise it's not great having both in one but I didn't think it'd be a problem. However, now I feel like I'd rather just close this PR, and create a seperate repo for the work at a later date.

PR #6961, opened 2018-02-22
Add package 'XUL and XBL'

thomcc:
This package adds syntax highlighting for XUL and XBL files. It's an admittedly niche use case.

### DIFF
--- a/repository/x.json
+++ b/repository/x.json
@@ -254,6 +254,17 @@
 					"tags": true
 				}
 			]
+		},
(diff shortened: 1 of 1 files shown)


thomcc:
The warning here is no longer necessary. The [most recent tag](https://github.com/thomcc/sublime-xul-xbl/tree/0.2.0) of the package doesn't have the warned against file.

Please let me know if there's anything else needed.

PR #6962, opened 2018-02-23
Add package color covert.

zhouyuexie:
Better and more convenient to use color conversion


zhouyuexie:
Who can explain why it can't be downloaded....

zhouyuexie:
@wbond Can you give me a hand? Thank you very much.

PR #6963, opened 2018-02-23
add Diagram plugin

jvantuyl:
This package generates diagrams from excerpts of source code.  This package was removed a few years ago because it didn't make the move to tag-based releases.

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -579,6 +579,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


jvantuyl:
Ummmm... I don't think this is something I can fix...

jvantuyl:
Hmmmm, bot is broken.  Going to re-issue pull request.

PR #6967, opened 2018-02-25
added enfold media to repository list

AlexanderRueffer:
Signed-off-by: Alexander Rüffer <alexanderruffer@AlexandersMBP2.fritz.box>

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -808,6 +808,17 @@
 			]
 		},
 		{
+			"name": "Enfold Media Queries",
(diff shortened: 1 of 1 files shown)


wbond:
You'll need to add a semver tag for this to be installable, and for the bot to test it

FichteFoll:
I can tell you already that you should remove the `Enfold Media.sublime-package` from the repo. The repo itself will be installed as a `.sublime-package`, so it doesn't make any sense to include that in the repo.

Also, these are snippets and not "auto completion" (regarding your repo description).

PR #6970, opened 2018-02-26
add Diagram plugin

jvantuyl:
This package generates diagrams from excerpts of source code. This package was removed a few years ago because it didn't make the move to tag-based releases.

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -579,6 +579,17 @@
 			]
 		},
 		{
+			"name": "Diagram",
+			"details": "https://github.com/jvantuyl/sublime_diagram_plugin",
+			"labels": ["diagrams", "plantuml"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
Please see https://github.com/wbond/package_control_channel/pull/6946.

FichteFoll:
I assume the fallback for `DiagramEx.sublime-syntax` is `diagram.tmLanguage`. I also assume this syntax is not the primary functionality of this package and it functions reasonably fine without it, especially considering that the readme explicitly claims support for ST 2 and 3.

bar4mi:
This package has a problem with finding its location. I installed it using Package Controller, but I can't use its example and syntax. I found that ST 3 on macOSX has wrong location definition. The Package Controller installed this package to ${packages}/Diagram, but the package defines its location to ${pacakges}/sublime_diagram_plugin in Main.sublime-menu. Please modify the location to use it correctly.

FichteFoll:
@bar4mi you should open this as an issue directly on the package's repository.

PR #6971, opened 2018-02-26
Adding Sf4Helpers to repo list

HeyIAmBob:
Symfony4 Helpers let you use Symfony4 main command line directly from Sublime Text via the command palette. All the commands available are listed in the github readme

As I know, it's the first plugin for symfony4, older version aren't compatible.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1249,6 +1249,20 @@
 			]
 		},
 		{
+			"name": "Sf4Helpers",
+			"description": "Symfony 4 command line helpers for Sublime",
+			"details": "https://github.com/HeyIAmBob/sublime-sf4helpers",
+			"issues": "https://github.com/HeyIAmBob/sublime-sf4helpers/issues",
+			"labels": ["Symfony", "Symfony4"],
+			"releases": [
+				{
+					"platforms": ["osx"],
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "SfCommand",
 			"details": "https://github.com/jtwebb/sfcommand",
 			"releases": [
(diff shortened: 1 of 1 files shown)


repository/s.json
```diff
@@ -1248,6 +1248,20 @@
 				}
 			]
 		},
+		{
+			"name": "Sf4Helpers",
+			"description": "Symfony 4 command line helpers for Sublime",
+			"details": "https://github.com/HeyIAmBob/sublime-sf4helpers",
+			"issues": "https://github.com/HeyIAmBob/sublime-sf4helpers/issues",
```
wbond: You can remove this line since it is automatically detected from `details`.

repository/s.json
```diff
@@ -1248,6 +1248,20 @@
 				}
 			]
 		},
+		{
+			"name": "Sf4Helpers",
+			"description": "Symfony 4 command line helpers for Sublime",
```
wbond: You can remove `description` and it will pull the description from your GitHub repo

repository/s.json
```diff
@@ -1248,6 +1248,20 @@
...
+			"details": "https://github.com/HeyIAmBob/sublime-sf4helpers",
+			"issues": "https://github.com/HeyIAmBob/sublime-sf4helpers/issues",
+			"labels": ["Symfony", "Symfony4"],
```
wbond: These labels have no other packages. I'd recommend adding `php` instead.

FichteFoll:
Looks similar to https://packagecontrol.io/packages/SfCommand, but since that package has been missing for quite a while, I don't see a problem.

HeyIAmBob:
@wbond : I update the json description following your recommendations.  Let me know if it's OK or not.

@FichteFoll : SfCommand is no longer maintened and only compatible with symfony 3.

Thom1729:
There is probably no benefit to tagging the package with `Symfony` and `Symfony4`, because no other package uses them.

PR #6977, opened 2018-03-01
added oracle commerce cloud design code utility plugin

byk0t:

### DIFF
--- a/repository/o.json
+++ b/repository/o.json
@@ -772,6 +772,17 @@
 			]
 		},
 		{
+			"name": "Oracle Commerce Cloud Design Code Utility Putter",
+			"details": "https://github.com/byk0t/sublime-occ-putter",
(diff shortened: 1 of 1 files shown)


repository/o.json
```diff
@@ -771,6 +771,17 @@
 				}
 			]
 		},
+		{
+			"name": "Oracle Commerce Cloud Design Code Utility Putter",
+			"details": "https://github.com/byk0t/sublime-occ-putter",
+			"labels": ["occ", "oracle commerce cloude", "design code utility"],
```
wbond: These labels can probably be removed since it is unlikely there will be multiple plugins with them

FichteFoll:
What @Thom1729 meant was to use `os.path`, not `sys.path`. But still, you'll definitely have to address the first issue as the plugin will just fail to load in most circumstances because of it.

PR #6979, opened 2018-03-03
Add Niji color scheme

umazi:


jrappen:
@umazi you can use [PackageDev](https://github.com/sublimetext/PackageDev) to convert the color scheme(s) to the **new** [color scheme format for ST3149+](https://www.sublimetext.com/docs/3/color_schemes.html)

PR #6982, opened 2018-03-06
Move OverrideAudit to new package repository file

OdatNurd:
This adds a new organization specific package repository file to the
channel.json file and moves the existing OverrideAudit package entry
from the o.json file in this repository to the new repository file.

The OverrideAudit entry has been augmented with a specific README URL,
home page entry and some labels.

Channel Repository Tools has been used to verify that the new
repository file in the STealthy-and-haSTy organization is valid as
well as the changed made to the channel.json and o.json in the
default channel.


FichteFoll:
May I ask what the motiviation behind this change was?

OdatNurd:
Basically I wanted the ability to modify the meta information for the package (minimum version, dependencies, etc) without having to add to the list of PR's on the channel (and the associated delay) since the package is already approved. 

In discussing possibilities with @keith-hall he mentioned he was interested in that as well, so we decided that having a single one in the organization was less clutter-y than doing the same thing twice

PR #6983, opened 2018-03-06
Rename phpunitkit -> PHPUnitKit

gerardroche:

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -1346,9 +1346,10 @@
 			]
 		},
 		{
-			"name": "phpunitkit",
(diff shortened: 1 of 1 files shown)


wbond:
I forget it this will work on Windows or Mac since it is a case change. I'll have to look just so things don't get broken if I merge.

gerardroche:
I think this maybe be merged. I took a quick a look at the relevant Package Control and it looks like the case insensitive issue is anticipated and handled.

FichteFoll:
(That's what I thought as well, but it's been a longer while since I looked at the code.)

PR #6986, opened 2018-03-08
Add "Tandem" Package

rageandqq:
Tandem is a solution that allows for decentralized, cross-editor collaborative editing!
This is the Sublime plugin for Tandem.

Thanks in advance for reviewing!

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -242,6 +242,18 @@
 			]
 		},
 		{
+			"name": "Tandem",
+			"details": "https://github.com/typeintandem/sublime",
+			"labels": ["Tandem", "collaborative", "cross-editor", "real-time"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"platforms": ["osx"],
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


repository/t.json
```diff
@@ -241,6 +241,18 @@
 				}
 			]
 		},
+		{
+			"name": "Tandem",
+			"details": "https://github.com/typeintandem/sublime",
+			"labels": ["Tandem", "collaborative", "cross-editor", "real-time"],
```
wbond: The `Tandem` label isn't necessary since that is the name of the package.

FichteFoll:
You import from `tandem` but your package will be installed as `Tandem`. Python is case-sensitive (and many file systems are as well).

>Does this invoke shell scripts or load shared libs, so as to need the .no-sublime-package file?

It runs a python daemon.

In your readme, you mention `view.run_command("tandem_connect", "abcd-1234")`. This will not work, since the second argument to run_command must be a mapping (i.e. a `dict`).

FichteFoll:
@rageandqq ping

PR #6987, opened 2018-03-09
Add package Git_Status

deanwang539:
GitStatus is a Sublime Text 2/3 plugin that shows git status information such as current branch/state in status bar and file changes in message popup.

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -492,7 +492,7 @@
 			]
 		},
 		{
-			"name": "Git Show Status bar",
+			"name": "Git Show StatusBar",
 			"details": "https://github.com/deanwang539/Git_Status",
 			"releases": [
 				{
--- a/repository/g.json
+++ b/repository/g.json
@@ -492,21 +492,21 @@
 			]
 		},
 		{
-			"name": "Git Status Files",
-			"details": "https://github.com/mkraft/git-status-files",
+			"name": "Git Show Status bar",
+			"details": "https://github.com/deanwang539/Git_Status",
 			"releases": [
 				{
-					"sublime_text": "<3000",
+					"sublime_text": "*",
 					"tags": true
 				}
 			]
 		},
 		{
-			"name": "Git_Status",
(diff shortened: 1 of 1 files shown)


deathaxe:
1. Why another GitStatus? (https://packagecontrol.io/packages/GitStatus)

2. Looks like there's an issue in certain situations. Got a couple of tracebacks.

```
  File "C:\Apps\Sublime\sublime_plugin.py", line 538, in on_activated_async
    callback.on_activated_async(v)
  File "C:\Apps\Sublime\Data\Packages\Git_Status\plugin.py", line 51, in on_activated_async
    self.check()
  File "C:\Apps\Sublime\Data\Packages\Git_Status\plugin.py", line 47, in check
    GitStatusCommand().run()
  File "C:\Apps\Sublime\Data\Packages\Git_Status\plugin.py", line 24, in __init__
    self.is_clean = status_manipulate(self.my_repo.get_git_status())[1]
AttributeError: 'NoneType' object has no attribute 'get_git_status'
```

deanwang539:
Hi,

We are aware that naming is similar but our features are different, which shows current branch the project is on and pop up window shows messages from command git status.

Thank you for the error message we will looking into that. Could you let us know how to reproduce it?

Best,
Dean

FichteFoll:
You may not use the `alt+p` binding because that is an accelerator for the "Project" menu.

~~You don't need to modify `sys.path` when you use a relative import later anyway. Just remove the `sys.path` line.~~ Unfortunately, your vendored Python packages use absolute imports internally. I hate Python packages that do this.

In your `on_activated_async` event hook you get handed a `view` object. You should pass that along to the code that does something with it (`GitStatusCommand.run`) instead of fetching a potentially different view via `sublime.active_window().active_view()`. This entire architecture with `GitStatusCommand` is pretty weird to begin with. Just use a module-level function.

`"info"` isn't really a unique key to display your status bar message. You should include something specific to your package, such as your package's name.

PR #6988, opened 2018-03-09
Added package Teacode-Sublime-Helper

r-zawislak:

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -328,7 +328,7 @@
 			]
 		},
 		{
-			"name": "TeaCode-Sublime-Helper",
+			"name": "TeaCode Helper",
 			"details": "https://github.com/Apptorium/TeaCode-Sublime-Helper",
 			"labels": ["teacode"],
 			"releases": [
--- a/repository/t.json
+++ b/repository/t.json
@@ -333,6 +333,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
+					"tags": true,
 					"platforms": ["osx"]
 				}
 			]
--- a/repository/t.json
+++ b/repository/t.json
@@ -330,11 +330,9 @@
 		{
 			"name": "TeaCode Helper",
 			"details": "https://github.com/Apptorium/TeaCode-Sublime-Helper",
-			"labels": ["teacode"],
 			"releases": [
 				{
 					"sublime_text": "*",
-					"tags": true,
 					"platforms": ["osx"]
 				}
 			]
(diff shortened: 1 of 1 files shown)


repository/t.json
```diff
@@ -327,6 +327,18 @@
 				}
 			]
 		},
+		{
+			"name": "TeaCode Helper",
+			"details": "https://github.com/Apptorium/TeaCode-Sublime-Helper",
+			"labels": ["teacode"],
```
wbond: Since this label is in the package name, and there is unlikely to be another teacode package,  it can be removed

Thom1729:
The documentation is still very vague. The description is meaningless to anyone who isn't already a TeaCode user.

You should choose a more specific name for your command than `expand`, such as `expand_teacode`. You should probably also provide a `TeaCode Helper.sublime-commands` file so that the command is accessible from the command palette.

r-zawislak:
@Thom1729 Done.

Thom1729:
The commands file refers to the command as `expand_with_teacode`, but the name of the `TextCommand` is just `ExpandCommand`.

Also, now that I look at it, the way you're building the command seems problematic:

```python
"Application('TeaCode').expandAsJson('" + text + "', { 'extension': '" + extension + "' })"
```

Won't this break if `text` or `extension` contains special characters?

r-zawislak:
@Thom1729 Thanks. TeaCode notified that expression was wrong when there were some special characters. Now they're escaped, so there is no more problem with it.

Thom1729:
If you're sending JSON-formatted text, you should really just use a JSON formatter rather than trying to escape it manually. Try `sublime.encode_value`. (You can also use `sublime.decode_value` in place of the `json` library in `handleJson`, if you want.)

Thom1729:
You should use proper JSON encoding on this line:

```py
command = ["osascript", "-l", "JavaScript", "-e", "Application('TeaCode').expandAsJson('" + re.escape(text) + "', { 'extension': '" + re.escape(extension) + "' })"]
```

Please fix this so that I can merge the package.

r-zawislak:
@Thom1729 Hope everything is alright now.

Thom1729:
Something like this:

```python
script = 'Application("TeaCode").expandAsJson({text}, { "extension": {extension} })'.format(
    text=sublime.encode_value(text),
    extension=sublime.encode_value(extension)
)
command = ["osascript", "-l", "JavaScript", "-e", script]
```

JSON is a subset of JavaScript (minus some irrelevant corner cases), so encoding the values as JSON should do the trick.

r-zawislak:
@Thom1729 Done.

PR #6989, opened 2018-03-09
Add new official Pipenv plugin to package list

kennethreitz:
✨🍰✨

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -1388,6 +1388,17 @@
 			]
 		},
 		{
+			"name": "Pipenv — Python Dev Workflow for Humans",
+			"details": "https://github.com/kennethreitz/pipenv-sublime",
+			"labels": ["python", "pip", "virtualenv", "pipenv", "Pipfile"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "PixarUSD",
 			"details": "https://github.com/davidlatwe/PixarUSD-Sublime",
 			"releases": [


kennethreitz:
\o/

repository/p.json
```diff
@@ -1387,6 +1387,17 @@
 				}
 			]
 		},
+		{
+			"name": "Pipenv — Python Dev Workflow for Humans",
```
wbond: The ` - Python Dev Workflow for Humans` should be in your repo description, not the package name.

repository/p.json
```diff
@@ -1387,6 +1387,17 @@
...
+			"details": "https://github.com/kennethreitz/pipenv-sublime",
+			"labels": ["python", "pip", "virtualenv", "pipenv", "Pipfile"],
```
wbond: `pipenv` and `Pipfile` should probably be removed since it is unlikely there will be more packages with those tags

repository/p.json
FichteFoll: (It's also using the EM DASH character and not the ASCII hyphen.)

repository/p.json
FichteFoll: Imo those can stay, as some other package might provide something to work with pipenv or with Pipfiles.

FichteFoll:
1. InputHandlers were added in build 3154, so you should restrict installation to that or higher builds.

2. Not a huge deal, but since you only define WindowCommands, your `is_enabled` hook in your mixin can use `self.window` over `sublime.active_window()`. Same for your commands to open files (and `sublime.active_window().active_view().window()` is pretty weird).

3. There is no need to override methods when you just defer to `super`.

4. ```

FichteFoll:
Accidentally submitted early.

4) This block will never be run, because `__name__` won't be `main` and your plugin will fail to load on any version below 3154.

  ```py
  if __name__ == '__main__':
      if sublime.version() < '3000':
          plugin_loaded()
  ```

5) You should pass absolute files to `Window.open_file` because you don't (want to) modify the working dir, which is the dir that ST was opened in. The fact that it worked for you is only because you opened ST from within your project folder, but it will break as soon as you start having multiple projects open or launch ST in any other way.

6) Instead of opening (anc closing) the console and `print`ing debug messages, look into using an output panel. They are normal views but attach to the bottom of the editor. All important API is being used [here](https://github.com/SublimeText/sublime_lib/blob/master/st3/sublime_lib/output_panel.py).

7) Since commands are run synchronously, the UI will most likely lock while pipenv is installing packages. To prevent this, either schedule a callback in the async thread using `sublime.set_timeout_async` or spawn a custom thread.

8) `requests` is available [as a dependency](https://github.com/packagecontrol/requests), so you don't need to vendor it. `parse`, however, is not. You can view a more or less pretty listing of currently available dependencies here: https://packagecontrol.github.io/

repository/p.json
FichteFoll: (And there we go, a package like this has already been submitted: https://github.com/wbond/package_control_channel/pull/7007/files#diff-a19631c4b751bdd943200931c7e6b81cR2480)

FichteFoll:
@kennethreitz ping

PR #6990, opened 2018-03-09
theme based on Armory blue (Armory.io) instead of the yellow from the original Cobalt2

JackReis:

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -1585,6 +1585,18 @@
 			]
 		},
 		{
+			"name": "ArmoryBlue",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Surely adding a screenshot of *your* theme instead of just saying "imagine these with blue" would help people more, don't you think?

jrappen:
@JackReis you can use [PackageDev](https://github.com/sublimetext/PackageDev) to convert the color scheme(s) to the **new** [color scheme format for ST3149+](https://www.sublimetext.com/docs/3/color_schemes.html)

PR #6991, opened 2018-03-10
added sublime_zk Zettelkasten package

renerocksai:
I created a package for people implementing Luhmann's Zettelkasten method via Markdown note archives (more in the README). I has gained a little fan base, especially over at https://zettelkasten.de, and is quite mature. People using it are loving it. Up until now I had perceived going package control would be too involved, but it turns out, it's not that hard after all 😄 
Please include this package.
Rene

### DIFF
--- a/repository/z.json
+++ b/repository/z.json
@@ -120,6 +120,19 @@
 			]
 		},
 		{
+			"name": "Zettelkasten - sublime_zk",
+			"details": "https://github.com/renerocksai/sublime_zk",
+			"description": "Markdown Zettelkasten support for Sublime Text 3",
+			"author": "Rene (https://github.com/renerocksai)",
+			"labels": ["Markdown", "Zettelkasten"],
+			"releases": [
+				{
+					"sublime_text": ">2999",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Zeus",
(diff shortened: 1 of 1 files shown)


FichteFoll:
I don't think you need the secondary `sublime_zk` name - "Zettelkasten" should be enough.

You have a `.no-sublime-package` file in your repo. What do you need it for? I can't spot anything. If you don't know, just remove it.

You might also want to look into modularizing your plugin, because it's quite long. You can use https://github.com/randy3k/AutomaticPackageReloader to reload the package more easily then because ST only reloads top-level plugins and not their dependencies in subfolders. I didn't take a close look at the code, but the quality seems high in general, so I presume you know what you are doing.

Instead of looping and searching by scope multiple times, you can just apply your modification in reverse order so you don't end up modifying the offsets of earlier regions.

Great readme though. Shows a lot of the features without being too detailed. I would probably reduce the TOC level though.

renerocksai:
Thanks, that's great feedback. I will consider all your suggestions. Great tip about the AutomaticPackageReloader BTW. And thanks for pointing out the looping can be sped up 👍 
This will take a bit as I am currently travelling and don't have as much time as I would like to have.

FichteFoll:
@renerocksai ping

PR #6992, opened 2018-03-11
Add JapaneseWordSeparator to repository/j.json

ASHIJANKEN:
This is a plugin for reinforcing word separator for Japanese.
With this plugin, we can separate Japanese words better than default word separator does by double-clicking and dragging.
I want to separate Japanese words in the same way as English words, but there aren't any plugins providing such functionality.

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -109,6 +109,17 @@
 			]
 		},
 		{
+			"name": "JapaneseWordSeparator",
+			"details": "https://github.com/ASHIJANKEN/JapaneseWordSeparator",
+			"labels": ["Japanese", "separator"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Jasmin JVM Assembler",
 			"details": "https://github.com/atmarksharp/jasmin-sublime",
 			"releases": [


FichteFoll:
You should be using relative imports instead of modifying `sys.path`.

You might also want to override `ctrl+right`, `ctrl+shift+right` and left variants for word-wise navigation, but mask them with a context (e.g. by checking whether the following or preceding characters are japanese).

ASHIJANKEN:
I stopped using `sys.path`, and I'm now using explicit relative import style instead.
I implemented keyboard version of the same system, as you suggested.
Your suggestion was helpful for me,. Thank you!

Thom1729:
You should probably limit the keybindings like so:

```json
{
    "keys": ["ctrl+left"],
    "command": "key_select_jp",
    "args": {"key": "left"},
    "context": [
        { "key": "preceding_text", "operator": "regex_contains", "operand": "\\p{Jpan}$" }
    ]
}
```

ASHIJANKEN:
I'm trying to mask like you said, but it doesn't seems to work...
```
"context": [
  { "key": "preceding_text", "operator": "regex_contains", "operand": "[\\p{Katakana}\\p{Hiragana}\\p{Han}。、，．！？　・「」：”（）ーﾟﾞ]$" }
]
```
I want it to be executed when preceding text is Japanese multibyte word, but always this plugin is executed.

Do you have any idea for this problem?
I also posted a question about this problem to Japanese StackOverFlow, so I'm now waiting for someone to answer.

Thom1729:
It looks like the regexps used in keymaps do not support unicode character properties. You can probably emulate them by specifying character ranges instead

FichteFoll:
I didn't mention this earlier, but I believe the usage of mousemap overrides to be fine in this case, since it replecates built-in behavior but adds special behavior for a not-so-unusual case.

Restricting the key maps to only japanese characters would be useful, I agree, but not a requirement to get it merged, so I'll go ahead and do that.

PR #6995, opened 2018-03-13
Add the 'With Python' package

nneonneo:
This package adds a set of tools which allow you to transform text using arbitrary snippets of Python code. Three tools are provided: sorting lines using a Python expression as a sort key, sorting selections using a Python sort key, and replacing text selections using a Python statement. Regex search, multiselect, and Python code replacement combine together to produce a very powerful and flexible text manipulation tool.

### DIFF
--- a/repository/w.json
+++ b/repository/w.json
@@ -538,6 +538,17 @@
 			]
 		},
 		{
+			"name": "With Python",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Some of this functionality (replacing the selection through Python-evaluated code) is available in other packages, but sorting by expressions is quite useful. Good idea.

PR #7007, opened 2018-03-19
Add PypiPackageInfo package.

gh640:
> 1. Used `"tags": true` and not `"branch": "master"` 
>     ([versioning docs](https://packagecontrol.io/docs/submitting_a_package#Step_4))

Yes.

>  2. Added a readme to your repository so that users (and reviewers) 
 
Yes, I did. The readme can be read here: https://github.com/gh640/SublimePypiPackageInfo .

Please let me know if I did anything wrong with this. Thank you in advance :)

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -2474,6 +2474,18 @@
 			]
 		},
 		{
+			"name": "PypiPackageInfo",
+			"details": "https://github.com/gh640/SublimePypiPackageInfo",
+			"author": "Goto Hayato",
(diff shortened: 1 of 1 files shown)


FichteFoll:
I don't think overriding the default `BaseException` is a good idea, but that won't matter for end users.

gh640:
@FichteFoll thank you for handling these PRs :)

You're right. The name `BaseException` conflicts with the default one and I'll change it.

PR #7009, opened 2018-03-20
add ACF Programmatic Fields Package

timlogemann:
Although similar to the existing ACF Snippets package, as they both add snippets for the WordPress plugin "Advanced Custom Fields", the difference is is in the snippets added, and their use cases.

The existing ACF snippets package allows the user to retreive data from fields. My package adds snippets to allow the user to create fields.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -124,8 +124,8 @@
 			]
 		},
 		{
-			"name": "ACF Snippets",
-			"details": "https://github.com/iamhexcoder/acf_snippets",
+			"name": "ACF Programmatic Fields",
+			"details": "https://github.com/timlogemann/sublime-acf-snippets",
 			"releases": [
 				{
 					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


timlogemann:
Sorry for the spam, I did not know it would check again. Thought I had to create a new Pull Request

FichteFoll:
Your PR uses mixed indentation. Please use tabs.

You also have a couple `.DS_Store` files in your repo. I suggest adding a pattern to your [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) for these.

timlogemann:
@FichteFoll thanks. I've updated the indentation, didn't find any .DS_store files though.. Added the pattern to my global gitignore nonetheless.

Everything should be correct now.

FichteFoll:
Well, they are right here: https://github.com/timlogemann/sublime-acf-snippets/blob/master/.DS_Store https://github.com/timlogemann/sublime-acf-snippets/blob/master/snippets/.DS_Store. You probbaly have to enable display of hidden files in your file manager (or use `-ls -a`).

timlogemann:
Ah right! I misunderstood and thought we were still talking about the package_control_channel fork. Thanks, going to remove them right away.

PR #7014, opened 2018-03-25
Fixed STProjectMaker's entry

pykong:
The repository was recently renamed to `ProjectMaker` to give that plugin a cleaner name.
Subsequently it got missing from package control.
Unfortunately packagecontrol then fetched the unreleased, now backwards incompatible, master branch.
Until we release `ProjectMaker2.0`, I like packagecontrol to only fetch the last fully working commit.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -3491,10 +3491,12 @@
 			]
 		},
 		{
-			"details": "https://github.com/bit101/STProjectMaker",
+			"name": "STProjectMaker",
+			"labels": ["project", "template", "jinja2"],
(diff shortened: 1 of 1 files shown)


FichteFoll:
PC will still fetch the `master` branch like this.

FichteFoll:
@pykong feel free to hit me up on discord if you have question, but this PR will do practically nothing (if it even ends up working at all, but it might).

pykong:
@FichteFoll Thanks for your reply. I am going to have time to fix things in July.

PR #7015, opened 2018-03-25
Add Folder Aliaser

tmelot2:
Allows you to alias your top-level project folders quickly and easily! 

There exists a similar project (FolderAlias by rablador, see Thanks section in readme) that is no longer maintained. The goal of this project was to have something simpler from both user and developer perspectives; and to have something that is maintained should issues arise. This project is currently more singular in focus and code easier to understand.

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -1295,6 +1295,16 @@
 			]
 		},
 		{
+			"name": "Folder Aliaser",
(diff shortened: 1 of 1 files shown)


FichteFoll:
I wonder, does the older unmaintained package need to be updated? It doesn't seem like this feature requires a lot of maintenance. Most of my packages haven't been touched in years and they do much more complex stuff.

PR #7016, opened 2018-03-27
Update CommandOnSave

klaascuvelier:
Updating the package CommandOnSave:
- Remove <3000 version and point to the newest repo
- Use tags instead


FichteFoll:
You could also provide the ST2 version of the package as a separate release, target versions <3000 and set the `base` key to the url of the old repo. Refer to the example repositories file for help.

FichteFoll:
@klaascuvelier ping

PR #7017, opened 2018-03-27
added OAT syntax highlighting package

dylanmann:
This is a simple syntax highlighting file for OAT code, which is written in the undergraduate compilers class at Penn [(website here)](https://www.cis.upenn.edu/~cis341/current/).

### DIFF
--- a/repository/o.json
+++ b/repository/o.json
@@ -13,6 +13,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


dylanmann:
Forgot to add tags, should work now, should I resubmit the pull request?

PR #7020, opened 2018-03-28
Software.com sublime text 3 plugin publish pull request

xavluiz:


xavluiz:
I've added a semver tags for the package https://github.com/swdotcom/swdc-sublime/tags

PR #7023, opened 2018-03-30
Added Commander package

TheDoctor0:

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2751,6 +2751,16 @@
 			]
 		},
 		{
+			"name": "Commander",
+			"details": "https://github.com/TheDoctor0/Commander",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Commando",
 			"details": "https://github.com/ericpridham/sublime-commando",
 			"releases": [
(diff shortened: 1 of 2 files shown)


FichteFoll:
How does this differ to [all the other "command" packages](https://packagecontrol.io/search/command)?

TheDoctor0:
Most of them are not in active development for years now, additionally, some of them work only on UNIX based systems.
For now, it aims to make Composer/Laravel based web applications development in Sublime easier with the bundle of ready to use commands, still leaving an option to add any other commands you want to use.
In future, I want to add commands for npm, yarn, vuejs, django, git and more to cover the full spectrum of web development tools.

FichteFoll:
Does "not in active development" mean they haven't been updated in a while or they are unmaintained and have open/unresponded issues?

I see that your package has a more application-specific focus and is less general, so I guess it qualifies at least as a specialization of the "external command" package type.

TheDoctor0:
@FichteFoll, yes, there are many open issues on repositories for packages with kinda similar functionality:
1. https://github.com/markbirbeck/sublime-text-shell-command/issues
2. https://github.com/gbaptista/sublime-3-shell-exec/issues
3. https://github.com/misfo/Shell-Turtlestein/issues

Also, I assume you are right about the name.
I changed it to WebDevShell, it's now probably more appropriate.

Thom1729:
Is the `WebdevshellCommand` constructor doing anything?

TheDoctor0:
@Thom1729, I added it at the start because for some reason command arguments were not passed correctly - yeah, it was a dirty hack.
I checked now and it works fine without it, so I removed it permanently.

PR #7024, opened 2018-03-30
Add Cup package for CoffeeScript integration

alin23:

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -4194,7 +4194,7 @@
 			"details": "https://github.com/alin23/Cup",
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": ">=3092",
 					"tags": true
 				}
 			]
--- a/repository/c.json
+++ b/repository/c.json
@@ -4190,6 +4190,16 @@
 			]
 		},
 		{
+			"name": "Cup",
+			"details": "https://github.com/alin23/Cup",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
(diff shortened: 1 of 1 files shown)


Thom1729:
- Why is `.no-sublime-package` set?
- The package is declared to be compatible with all versions of Sublime. You probably want to declare compatibility with 3092 and above.
- Conversely, why is the literate syntax a `tmLanguage`? If the package requires `sublime-syntax` anyway, you might as well convert it.
- What in the world is `sort.js` doing?

alin23:
Hi @Thom1729, thank you taking the time to look into this ^_^

* I set `.no-sublime-package` because I'm executing `sort.js` using `node` and from what I understood I can't run scripts like that if the package is archived
* I fixed the compatibility declaration
* I got the literate syntax directly from the Better CoffeeScript package but I didn't make any changes to it so I forgot to convert it. I just did it.
* `sort.js` is a webpack bundled code that reads CoffeeScript code from stdin and outputs the code with the imports sorted. I uploaded the code here: [cup-import-sort](https://github.com/alin23/cup-import-sort)

Thom1729:
Any luck with factoring out `sort.js`?

Thom1729:
Is the `sort.js` file in the repo no longer needed?

alin23:
Oh right, I forgot to delete it

Thom1729:
Also, you should probably add a license.

alin23:
Okay, I added a license. Is there anything more I should do before drafting a new release?

Thom1729:
Looks good to me.

PR #7025, opened 2018-03-31
Add FuzzyImports package

11bit:
SublimeText 3 plugin for browsing file system from quick panel and adding javascript module imports. Similar to FuzzyFileNav but for adding javascript import statements.

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -1752,6 +1752,17 @@
 					"branch": "master"
 				}
 			]
+		},
+		{
+			"name": "FuzzyImports",
+			"details": "https://github.com/11bit/fuzzy-imports",
+			"labels": ["javascript", "js", "imports"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
 		}
 	]
 }


Thom1729:
I see that you're modifying `sys.path`. This should generally be avoided, because you're sharing the path with other packages. What is that being used for?

11bit:
@Thom1729 I am using it to import a third party package. It's a parser library "Lark". I saw this approach in a couple of other packages. Is it a wrong way?

Thom1729:
You shouldn't generally do that. But I don't see where you're bundling Lark in the first place.

If you need a third-party library, you should probably create a [dependency](https://packagecontrol.io/docs/dependencies) for it. Dependencies are automatically added to the path in a sane manner.

11bit:
@Thom1729 thanks for explanation. 

> You shouldn't generally do that. But I don't see where you're bundling Lark in the first place.

I am adding `./lib` folder to sys.path here:
https://github.com/11bit/fuzzy-imports/blob/master/FuzzyImports.py#L21

and importing `lark` package from that folder afterwards:
https://github.com/11bit/fuzzy-imports/blob/master/tools/exports_parser.py#L4-L5

> If you need a third-party library, you should probably create a dependency for it. Dependencies are automatically added to the path in a sane manner.

Yes, I see. But this is an early version of my package. I am just experimenting with idea (if it will be useful for anybody except me, and what features will be really needed). Currently I am not sure about this particular library and don't want to dive into a whole publishing process for it right now. 

So is it possible to add my package with this dirty way of importing lark and create a dependency later?

Thom1729:
I don't see where `lark` is actually coming from. It's not a built-in library and I don't see it anywhere in the repo. I did see a reference in `.build` to `lib/lark`, a directory that does not exist. Is it possible that that directory was not committed?

The best way to provide the `lark` package would be as a dependency. But you can also bundle the package inside your own and simply refer to it as `FuzzyImports.lark`. Munging `sys.path` is not recommended.

Thom1729:
> Is it correct that I don't need '.no-sublime-package' anymore (i used it only import lark via sys.path)?

Correct.

11bit:
Removed. Is there something else to tweak?

Thom1729:
You should include the `lark` license in that directory.

What are `.build`, `makefile`, and `requirements-dev.txt` for?

For testing, you might want to look at the [UnitTesting](https://github.com/SublimeText/UnitTesting/) package. The way you're doing it is probably fine.

11bit:
> You should include the lark license in that directory.

Done

> What are .build, makefile, and requirements-dev.txt for?

It is only for dev env: 
.build is a autogenerated helper file for packagereloader plugin that helps to reload a package after changes without restarting sublime
makefile, requirements-dev.txt - I am using nosetests framework for testing and these files will remind me what I am using and how I am running it

Thom1729:
One more thing. You have three seemingly identical keymaps. Is there any reason not to instead have one `Default.sublime-keymap`?

11bit:
No, there is no reason for now. So I deleted separate keymaps. Thanks!

PR #7026, opened 2018-03-31
Add JavascriptExtractFunction package

tricinel:
Thank you so much for reviewing!

I needed a way to refactor Javascript code. I found myself quite often cutting and pasting a piece of code from one place into another, basically to extract some code into its own function that I could reuse afterwards. I hated that I had to keep track of where I got the code from, so that after I pasted it, formatted it and named my new function, I could go back to where it was initially and insert the function call.

This package adds a quick and easy way just automate that process. It will let users select one or more lines of code, choose the type of Javascript function they want to create (ES6 arrow fn, regular fn or an anonymous fn) and extract that code easily into its own code block. It will ask for a function name and automatically insert the function call where the original code was.

Nice and sweet with no overhead.

Thanks again!

Cheers,
Bogdan

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -432,6 +432,17 @@
 			]
 		},
 		{
+			"name": "JavaScriptExtractFunction",
+			"details": "https://github.com/tricinel/JavascriptExtractFunction",
+			"labels": ["javascript", "javascript refactor", "javascript extract", "javascript tools"],
+			"releases": [
+				{
+					"sublime_text": ">=3103",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "JavaScriptFunctionDefinition",
 			"details": "https://github.com/asdf23/JavaScriptFunctionDefinition",
 			"labels": ["javascript", "autocomplete"],


Thom1729:
Why is `.no-sublime-package` declared?

tricinel:
@Thom1729 thank you so much for reviewing this! I must admit I wasn't sure about it. I switched between having it and not having it.

What do you recommend?

Thanks again! I really appreciate the work that's being put into this!

Cheers,
Bogdan

Thom1729:
If you don't have a specific need to use `.no-sublime-package`, then you shouldn't use it.

Other thoughts:

- You have three identical platform-specific keymaps. You can simplify this by having one `Default.sublime-keymap`.
- You're not going for ST2 compatibility, so I'm not sure what the point is of using that edit library.

tricinel:
Hi @Thom1729! Thanks again for the feedback. I've made the suggested changes. Thank you for the help!

Regarding the use of the Edit library. This is actually my first real encounter with Python (I'm learning). As I was developing this package, I found out that I couldn't use the edit buffer outside of my run function. So I took a look at how other people have done this - manipulating the user selection and replacing content - and came across this Edit "hack" if I may call it that.

Is there a better way to do it?

I appreciate all your effort on this, of course!

Cheers,
Bogdan

Thom1729:
The second parameter to the `run` method is a `sublime.Edit` object. (In your code, you name that parameter `_edit`.) In order to use that object elsewhere, you would pass it along to the other methods. You could probably refactor some of the code to make it more concise, and that would make it easier to pass `edit` along.

FichteFoll:
Well, the code is kind of all over the place, but asking you to improve it would be unreasonable. The usage of the `Edit.py` library is certainly unnecessary, but it's not critical in that it affects other functionality.

You cannot use `alt+f` as binding, however, because that's the accellerator for the File menu.

tricinel:
Hi @FichteFoll! Thank you for the reply. I would love to be able to refactor this to match Sublime's standards. I'd appreciate it immensely if you could point me in the right direction...perhaps a Sublime package that uses good standards and I can learn from, as I'm a newbie with Python.

The `Edit.py` is something that I couldn't figure out how to do without - passing the `_edit` object outside of the `run` didn't work for me - but again, I'm sure I'm doing something wrong.

Any help is immensely appreciated and I understand it's not your priority to help others release packages on Sublime :)

Cheers,
Bogdan

Thom1729:
Sorry about the delay! As FichteFoll says, this could be merged as-is, except that the alt-f keybinding should be changed.

tricinel:
Hi @Thom1729! Thank you for the reply! I had that key binding as `alt+m` initially (m for method). Is there a place where I can see if that conflicts with anything? I've been using FindKeyConflicts to see if there are any conflicts, but apparently it doesn't catch everything (at least it didn't catch the `alt+f`).

Thank you for your help!

I'm taking a look at the plugins both you and @FichteFoll have and will try to learn from them and refactor my own.

Thom1729:
In this case, the binding interferes with the menu mnemonic. This is defined in `Packages/Default/Main.sublime-menu` under the `mnemonic` key. I suspect that FindKeyConflicts may not look at `sublime-menu` files.

At any rate, it doesn't look like any of the top-level menus has a mnemonic of `m`, so `alt-m` should be fine.

tricinel:
Hey! Thanks again for the heads up! Yep, I assume the FindKeyConflicts looks at plugin/user defined key bindings.

I've changed that as suggested to `alt+m` and I'm going to refactor the code - see how I can get rid of the Edit as well.

Cheers!
Bogdan

PR #7027, opened 2018-04-02
Add StoryScript Highlight package

swwind:
language syntax.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -3487,7 +3487,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"branch": "master"
+					"tags": true
 				}
 			]
 		},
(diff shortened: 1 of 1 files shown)


swwind:
Shall I use `"sublime_text": ">=3092"`?

Thom1729:
The syntax definition is a bit rough.

- Negative lookbehinds slow down syntax highlighting. They cause Sublime to use the older, slower regexp engine. Therefore, it's better to avoid them wherever possible.
- You might need more `\b` assertions here and there, such as on line 100.
- The numbers regexp is terrifying. Can you factor parts of it out into variables? It would also be a good idea to use a multiline string and the `(?x)` regexp flag for readability's sake.

PR #7028, opened 2018-04-02
Now LaTeX Accents is compatible also with ST3

kpym:
I have made small corrections to the code to make the package 
ST3 compatible.


Thom1729:
Just to clarify, the package now is both ST2- and ST3-compatible?

kpym:
@Thom1729 Yes, it is.

PR #7030, opened 2018-04-03
add fewest moves plugin

oyyq99999:
This is a plugin for Rubik's Cube Fewest Moves practising using an editor, more powerful functions are upcoming.

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -428,6 +428,16 @@
 			]
 		},
 		{
+			"name": "Fewest Moves",
+			"details": "https://github.com/oyyq99999/sublime_fewest_moves",
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Fex",
 			"details": "https://github.com/hwangzhiming/sublime-text-fex",
 			"releases": [


Thom1729:
The 'notation.fm' scope is not standard. What about `constant.other.fm`? And is there really a need for an accompanying color scheme?

If you're not supporting Sublime Text 2, there's not much point in supporting builds before 3092, because the 3.0 final release was build 3143. Rather than adding the `.tmLanguage`, I'd just declare support for `"sublime_text": ">=3092"`.

Thom1729:
It looks like `FewestMovesEventListener` is running the `count_moves` command every single time any view is modified. This should only apply to views using the Fewest Moves syntax.

oyyq99999:
I have set it to run only when the syntax is fewest moves, see [here](https://github.com/oyyq99999/sublime_fewest_moves/blob/0.2.3/fewest_moves.py#L102-L103)

Thom1729:
Fair enough. It might be clearer to explicitly check for that in the listener, although I'm not aware of concrete advantages to doing so.

Another thing I noticed: the keymap doesn't check the scope either. Even if the command is not enabled, I think that the keymap will eat the keystroke, possibly interfering with other syntax-specific bindings.

oyyq99999:
Added syntax checks for keymap and eventlistener

PR #7032, opened 2018-04-03
Add Perlgem syntax highlighting package

dom111:


Thom1729:
I don't quite understand what this is for. Is Perlgem some kind of perl-based package system, and is this highlighting for a config file?

dom111:
This can be closed, we've switched to a private channel. Thanks for reviewing though!

PR #7036, opened 2018-04-03
Add MySFTP Package

icjmaa:

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -2535,6 +2535,16 @@
 			]
 		},
 		{
+			"name": "MySFTP",
+			"details": "https://github.com/icjmaa/MySFTP",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "master"
+				}
+			]
+		},
+		{
 			"name": "MySQL Snippets",
 			"details": "https://github.com/ancor-dev/sublime-sql-snippets",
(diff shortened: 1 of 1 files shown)


Thom1729:
- It's difficult for me to review this package because my Spanish is rusty and lacks technical vocabulary.
- What does this package do? How does it differ from other FTP packages?
- Why are you extracting files from the package?
- The package claims to be compatible with all platforms, and it even includes a Linux keymap (though oddly not a Mac keymap). But the `/bin` directory is full of `.dll` and `.exe` files. This seems problematic.
- Why are you bundling the Monokai theme?

icjmaa:
@Thom1729 

- My package make connections of type sftp to a remote server for list the files and open file selected in a new tab to be able to edit and modify for after update file in the server remote, how diferecce the package is free and warn who this modifying the file for other people

- Need extract files, to be able to access Monokai.tmTheme, MySFTP.tmLanguage and binary resource to be able to make the conexion to server in the case of Windows System or if there is another  form for to access a the package files

- The package was develope mainly for Linux and Windows, and for Windows i require the files .exe and .dll

- The file Monokai.theme is for customize the output console.

FichteFoll:
Instead of manually creating a JSON string (`str_json_config`), you could just use `json.dumps`.

PR #7042, opened 2018-04-08
Add ChromeREPL Package

acarabott:
This package wraps the [Chrome DevTools Protocol](https://chromedevtools.github.io/devtools-protocol/) to execute code in Sublime Text in a Chrome Tab. This allows for live coding and interactive debugging of webapps.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -998,6 +998,16 @@
 			]
 		},
 		{
+			"name": "Chrome REPL",
+			"details": "https://github.com/acarabott/ChromeREPL",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Chrome Snippets",
 			"details": "https://github.com/ismnoiet/ChromeSnippets",
 			"releases": [


acarabott:
Will this still be accepted, or should the key bindings be changed?

Thom1729:
Before I get to the nitpicks: Awesome plugin! I'm looking forward to trying it out.

- You're modifying `sys.path`, which is a bad thing because `sys.path` is shared among all plugins. I believe the only place you're really using this is for importing `six` from `websockets`. The simple way to correct this would be to patch `websockets` to use a different import path. The complicated way would be to create a [dependency](https://packagecontrol.io/docs/dependencies) for `six`. The absolute best way might be to port `websockets` to pure Python 3 and then publish `websockets` as a dependency.
- You should include the licenses for the libraries you're bundling.
- What's the point of connecting to Chrome when the plugin loads? That will happen every time Sublime is opened, regardless of what the user is actually doing. Overall, I'm a bit worried that this plugin might be doing a lot of work when the user might not want or need it.
- Does this maintain one shared chrome tab for all of Sublime? Would it make more sense to maintain one Chrome tab per Sublime tab that uses the REPL?
- Tip: `tab['url'].find('chrome-devtools://') != -1` &rarr; `tab['url'].startswith('chrome-devtools://')`.

acarabott:
This is great feedback! Making relevant changes...

acarabott:
Changes done [tag 1.2.0](https://github.com/acarabott/ChromeREPL/commit/3629e56b1a31c87bd9cf775c8c5c166d38645528)

1. I took the simple way, and created a script to patch the import paths for six and websockets. TBH I'd rather not use any Package Control dependencies, as they rely on contributers repos continuing to exist, but including requests is such a pain so I've kept that one.
2. Licenses included
3. This was left over from development, removed
4. Done. I wasn't sure if you could have multiple simultaneous connections,  but it works! Much better, thanks for the suggestion.
5. Done

Thom1729:
What I meant for (1) is to change the import statements in the `websockets` code, e.g.:

```python
import six # before
from .. import six # after
```

Thom1729:
@acarabott Is it possible to modify the imports to remove `sys.path`? That's the only blocker on merging this issue.

acarabott:
Yep, will have some time this weekend to fix it up. Cheers!

acarabott:
Right, I've pushed [1.3.1](https://github.com/acarabott/ChromeREPL/releases/tag/1.3.1)

There are no more modifications to `sys.path` 

I'd prefer to keep the dependency imports as `import ChromeREPL.libs.six as six` as this makes it a lot easier to automate patching of multiple modules, especially with nested directories which would require a mix of `from .. import` and `from ../.. import` etc. 

Does that sound reasonable to you? 🤔

Thom1729:
Awesome!

acarabott:
I've had to change the package name from "Chrome REPL" to "ChromeREPL" as imports break. Have submitted a pull request here: https://github.com/wbond/package_control_channel/pull/7113

PR #7045, opened 2018-04-09
Add Reason package

chenglou:
ReasonML is a new language. This is the official plugin support for it.

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -700,7 +700,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"tags": "true"
+					"tags": true
 				}
 			]
 		},
--- a/repository/r.json
+++ b/repository/r.json
@@ -695,6 +695,16 @@
 			]
 		},
 		{
+			"name": "Reason",
+			"details": "https://github.com/reasonml-editor/sublime-reason",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": "true"
+				}
+			]
+		},
+		{
 			"name": "ReaSyntax",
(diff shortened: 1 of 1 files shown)


Thom1729:
- For a complex syntax like this, I'd recommend using the [`.sublime-syntax` format](https://www.sublimetext.com/docs/3/syntax.html). This would limit compatibility to ST3, but virtually no one uses ST2 anyway. Up to you.
- What's the purpose of `OCaml.sublime-settings`? It doesn't seem to apply to the Reason language's scope.

chenglou:
- We're reworking the syntax a bit later on. If it's ok, we'd like to keep this as-is?
- Reason's a dialect of OCaml, and we're trying to unify the package with OCaml, with that as a small step. Unless you'd like me to remove it?

Thom1729:
I would say that because that setting file isn't related to the rest of the package (i.e. it won't be used in Reason files because the scopes don't match), it probably shouldn't be part of the package at this time. However, if those `auto_complete_triggers` would be useful and appropriate for general OCaml development, then it would be worth [opening a PR to add them to the core OCaml package](https://github.com/sublimehq/Packages/).

Also, it seems to me that this package provides complete, self-contained syntax highlighting and such, but that other features are provided by [a separate language server](https://github.com/freebroccolo/ocaml-language-server). The documentation seems to imply that the language server is a necessary part of this package. It might be a good idea to clarify this distinction: that this package provides Reason syntax highlighting, and that the recommended language server provides additional functionality in combination with the LSP package.

chenglou:
K, I've removed the ocaml setting and updated the README. Thanks for the tips!

PR #7047, opened 2018-04-10
support ux syntax

yupeng528:

### DIFF
--- a/repository/u.json
+++ b/repository/u.json
@@ -522,6 +522,16 @@
 			]
 		},
 		{
+			"name": "Ux Syntax Highlight",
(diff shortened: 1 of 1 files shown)


jrappen:
@yupeng528 You can use the new [`*.sublime-syntax` format for ST3092+](https://www.sublimetext.com/docs/3/syntax.html) based upon YAML. Also see the [`testing`](https://www.sublimetext.com/docs/3/syntax.html#testing) section there for automated testing, as well as the [scope naming guide](https://www.sublimetext.com/docs/3/scope_naming.html).

PR #7049, opened 2018-04-10
add BlackOcean tools

jesus-aldrete:
v1.0.0

captura de pantalla 2018-04-15 a la s 17 03 41

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -797,6 +797,20 @@
 			]
 		},
 		{
+			"name": "BlackOcean",
+			"details": "https://github.com/jesus-aldrete/SB3P_BlackOcean",
+			"readme": "https://raw.githubusercontent.com/jesus-aldrete/SB3P_BlackOcean/master/README.md",
+			"homepage": "http://blackocean.run",
+			"author": "Jesus Aldrete <jesus@aldrete.run>",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": "st3-"
+				}
+			]
+		},
+		{
 			"name": "BlackRain",
 			"details": "https://github.com/ginfuru/Sublime-BlackRain",
 			"labels": ["color scheme"],
--- a/repository/b.json
+++ b/repository/b.json
@@ -799,14 +799,11 @@
 		{
 			"name": "BlackOcean",
 			"details": "https://github.com/jesus-aldrete/SB3P_BlackOcean",
-			"readme": "https://raw.githubusercontent.com/jesus-aldrete/SB3P_BlackOcean/master/README.md",
-			"homepage": "http://blackocean.run",
-			"author": "Jesus Aldrete <jesus@aldrete.run>",
 			"labels": ["language syntax"],
 			"releases": [
 				{
 					"sublime_text": ">=3000",
-					"tags": "st3-"
+					"tags": true
 				}
 			]
 		},
--- a/repository/b.json
+++ b/repository/b.json
@@ -802,7 +802,7 @@
 			"labels": ["language syntax"],
 			"releases": [
 				{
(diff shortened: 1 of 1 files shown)


Thom1729:
- I have no idea what this package is for. Some kind of templating system?
- There's no readme.
- The scopes are inconsistent. For example, sometimes scopes end in `.css.blackoceancss` and sometimes just in `.blackoceancss`. The former is preferred.
- Syntax tests would be nice, though they're not mandatory.
- You're using regexp lookbehinds in the syntaxes. These cause Sublime to use the old, slower regexp engine.
- Were the syntax definitions derived from another source? If not, they probably should be. If so, you should probably credit the original authors.
- In particular, the modified JS syntax seems to be based on an old version.

jesus-aldrete:
- In effect, it is a system of web templates in which I am working.

- Unfortunately, I am still working on documenting other parts of the code and the documentation does not mention that readme was mandatory. Do I have to do it before the authorization of the package?

- An apology, I am new to this and I will carefully check that there are no problems with the scopes, I will also change them to the first recommendation. Once corrected, do I have to do another PR or just wait for a new revision?

- The project is something big and if you allow me not to have the syntax tests (for the moment), I would appreciate it.

- Please, I could elaborate this point more. I took as an example the sublime syntax and the online documentation and I did not read anything about "regexp lookbehinds", you can tell me what it means to correct this.

- The only syntax that derives from a different source is "BlackOceanJS.sublime-syntax" (the others are written from 0), this being a derivative of the original definition of sublime (I could not inherit it because that definition includes other definitions and I was not enough to inherit it), being so, as a reference to the original author? By means of a comment in the head of the definition taking as author the author of subl 3?

Thom1729:
(Just to clarify, I'm not one of the channel maintainers and I have no power to approve or reject — merely to recommend.)

> I am still working on documenting other parts of the code and the documentation does not mention that readme was mandatory. Do I have to do it before the authorization of the package?

I would recommend having at least a brief statement of what the package is, what it's for, and what functionality it provides (i.e. syntax highlighting and completions). A few sentences would suffice. Without that, it can't be very useful to anyone else.

> An apology, I am new to this and I will carefully check that there are no problems with the scopes, I will also change them to the first recommendation. Once corrected, do I have to do another PR or just wait for a new revision?

I doubt that the scoping quirks would hold up the package. Either way, it isn't necessary to change the PR here.

> The project is something big and if you allow me not to have the syntax tests (for the moment), I would appreciate it.

It's up to you. I've written a good many syntax definitions, and I've always found that having tests made my own life a lot easier.

> Please, I could elaborate this point more. I took as an example the sublime syntax and the online documentation and I did not read anything about "regexp lookbehinds", you can tell me what it means to correct this.

Sublime has two regexp engines: a generic engine (Oniguruma) and a specialized custom engine. The custom engine is substantially faster *because* it does not support several features: backreferences, lookbehind, subexpressions, atomic groups, and greedy quantifiers. It's better to avoid these features whenever possible to avoid using the slower generic engine. Sublime comes with a “Syntax Tests — Regex Compatibility” build system that will tell you if you are triggering the old engine.

> The only syntax that derives from a different source is "BlackOceanJS.sublime-syntax" (the others are written from 0), this being a derivative of the original definition of sublime (I could not inherit it because that definition includes other definitions and I was not enough to inherit it), being so, as a reference to the original author? By means of a comment in the head of the definition taking as author the author of subl 3?

Yeah, I'd put a comment at the top of the file and in the readme. No need for anything elaborate.

On a semi-related note, I wrote a package called [YAML Macros](https://github.com/Thom1729/YAML-Macros) that makes it much easier to extend an existing syntax with new features. It may or may not be useful to you.

jesus-aldrete:
Thank you very much for your recommendations. I'll start the corrections now.

PR #7051, opened 2018-04-10
Added sublime_lib dependency.

Thom1729:
Utility library for frequently used functionality in Sublime Text library and convenience functions or classes

Created in collaboration with @FichteFoll.

Needs documentation and more tests before the existing features are ready for public consumption, but having it in Package Control will make it easier to use it in pre-release versions of other packages, which will in turn provide invaluable feedback for sublime_lib's development.

### DIFF
--- a/repository/dependencies.json
+++ b/repository/dependencies.json
@@ -901,6 +901,20 @@
 			]
 		},
 		{
+			"name": "sublime_lib",
+			"load_order": "01",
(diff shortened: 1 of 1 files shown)


Thom1729:
@FichteFoll I think that we can merge this, but I'd rather have a second human take a look at it if you have a minute.

FichteFoll:
Sorry, but with the number of notifications from this repo, mentions just get lost. If you need urgent feedback from me on a pr, ping me on discord~

Anyway, this looks fine.

PR #7052, opened 2018-04-10
Add Constellation project-/workspace-management plugin

abathur:
This plugin is a slightly different take on managing projects and workspaces in ST3. While there are already some project managers that attempt to provide a replacement for the core project-manager, Constellation is just a small, simple layer that enables you to group existing projects (into a named constellation) and load/close all of the projects in a constellation at once.

Projects can be in more than one constellation, so you can create slightly-different groupings for different tasks.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3057,6 +3057,15 @@
 			]
 		},
 		{
+			"details": "https://github.com/abathur/constellation",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Contao Front-End Snippets",
 			"details": "https://github.com/marcobiedermann/sublime-contao-front-end-snippets",
 			"labels": ["cms", "contao", "snippets", "css", "less", "sass", "scss", "stylus"],
@@ -4299,7 +4308,7 @@
 					"tags": true
 				}
 			]
-		},		
+		},
 		{
 			"name": "Cycle Setting",
 			"details": "https://github.com/jmm/Sublime-Text-Cycle-Setting/tree/package-control",


abathur:
I've addressed the auto-review suggestion in the package's repository, but it's not clear to me if I need to make and push an additional trivial change to trigger a fresh check, or if will automatically re-check?

Thom1729:
If you haven't tested this on Windows, and some features are known not to work on Windows, then I wouldn't recommend marking the package as Windows-compatible.

abathur:
@Thom1729 FWIW, I've pushed some changes that disable the two commands in question on the Windows platform, and replace them (for Windows users) with a nudge to help contribute them.

Thom1729:
It might be a good idea to factor out the logging and provide a setting to turn it off. There are a lot of `print` statements laying around.

I'm not quite sure why it's necessary to use the `subl` command.

Other than that, it looks good.

abathur:
@Thom1729 Good point. I don't have time to clean up the logging just this moment, but I've pushed a quick release that comments out the print statements until I have a little time to do it right.

Re: subl command, I'm a little disappointed to need it myself. I didn't think to make notes on this decision, but AFAIR: the problem is that there's no (documented?) feature in the API that enables opening a project, and all of the commands that ST3 dogfoods for its own project menu are either narrowly focused on selecting a recent project, or open up file browsers for the user to search. After struggling against these for a bit, I found and lifted the current idiom from randy3k's ProjectManager plugin ([subl definition](https://github.com/randy3k/ProjectManager/blob/1ded38f4b9069af49dda9c41853a78143a892e2c/project_manager.py#L12), [subl used to open project](https://github.com/randy3k/ProjectManager/blob/1ded38f4b9069af49dda9c41853a78143a892e2c/project_manager.py#L374)).

Hope that helps. Thanks for taking a look.

Thom1729:
Well, that's a a baffling API omission. ([Issue](https://github.com/SublimeTextIssues/Core/issues/134).)

abathur:
Indeed. I was a little flabbergasted. Thanks again for your help with this.

PR #7054, opened 2018-04-11
MicroPython package added.

bisguzar:

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -1321,7 +1321,6 @@
 			"releases": [
 				{
 					"sublime_text": ">=3000",
-					"platforms": ["windows", "osx", "linux"],
 					"tags": true
 				}
 			]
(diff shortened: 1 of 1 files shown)


Thom1729:
- You're manually patching `sys.path`. This is generally a bad idea, because you're sharing that path with - other packages. Is there some reason you absolutely must do this?
- You're missing a LICENSE file. Is this deliberate or an oversight?
- Is the code under `modules` your work, or is it existing software? If the latter, you should attribute it and (if necessary) include that license.

bisguzar:
@Thom1729 thanks for notices. I added module's LICENSE and created a LICENSE for my package. 
And I'm not adding a path to sys.path manually anymore. 

Is there any lack except these?

Thom1729:
I don't see the package LICENSE.

bisguzar:
@Thom1729 you are right, I not need it anymore. I created a LICENSE file and removed REAL_PATH veriable

PR #7055, opened 2018-04-12
Adding a theme- RetroSoft

ryansisco:
I created it a while back, and recently changed it to have less contrast in order to be better in the dark.


jrappen:
Your package naming seems wrong as you're confusing "Color Schemes" with "Themes".

Compare the docs:

* <https://www.sublimetext.com/docs/3/color_schemes.html>
* <https://www.sublimetext.com/docs/3/themes.html>

PR #7056, opened 2018-04-13
From now on I take charge of the plugin WebSearch, now I am the owner.

kallookoo:


kallookoo:
Is there a problem? I see that it did not update well on the web.

PR #7057, opened 2018-04-13
Added new package (QuickReplaceAtCursor)

jlaurenssen:

### DIFF
--- a/repository/q.json
+++ b/repository/q.json
@@ -287,7 +287,7 @@
 		},
 		{
 			"name": "QuickReplaceAtCursor",
-			"details": "https://bitbucket.org/JorisL/quickreplaceatcursor,
+			"details": "https://bitbucket.org/JorisL/quickreplaceatcursor",
 			"releases": [
 				{
 					"sublime_text": ">=3000",
--- a/repository/q.json
+++ b/repository/q.json
@@ -294,7 +294,7 @@
 					"tags": true
 				}
 			]
-		}
+		},
 		{
 			"name": "QuickSearchEnhanced",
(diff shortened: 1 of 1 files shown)


Thom1729:
You might want to precompute the sorted list of replacement keys.

Thom1729:
@jlaurenssen What about moving some of the logic out of the looks, as follows:

```python
import sublime
import sublime_plugin


class QuickReplaceAtCursorCommand(sublime_plugin.TextCommand):
    def run(self, edit):
        settings = sublime.load_settings('QuickReplaceAtCursor.sublime-settings')
        replacements = settings.get('replacements')

        keys = list(sorted(replacements.keys(), key=len, reverse=True))

        # Get the length of the largest defined replacement key. This limits the selection that has to be stored
        # and checked for a possible match.
        max_key_length = len(keys[0])

        for selection in self.view.sel():
            # Get the text before the cursor, with max_key_length as maximum string length
            cursor_pos = selection.begin()
            text_upto_cursor = self.view.substr(sublime.Region(max(cursor_pos - max_key_length, 0), cursor_pos))

            # Loop over each possible key in the user list of possible replacements, from large to small
            # By looping from large to small, it is possible to have defined replacements that are also a subset of
            # another defined replacement, for example tm->™ in combination with t->τ
            for key in keys:
                # If the current key matches the selection, replace it with the replacement value for that key
                if text_upto_cursor.endswith(key):
                    key_region = sublime.Region(cursor_pos - len(key), cursor_pos)
                    self.view.replace(edit, key_region, replacements[key])
                    break
```

PR #7059, opened 2018-04-14
Added mIRC Scripting Language

eneerge:
This is the first package that provides support for mIRC scripting language syntax highlighting and command autocompletion.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -1486,6 +1486,18 @@
 			]
 		},
 		{
+			"name": "mIRC Scripting Language (Highlighting and Autocomplete)",
+			"details": "https://github.com/eneerge/mIRC-Scripting-Language-for-Sublime-Text",
+			"labels": ["language syntax", "auto-complete"],
+			"author": "Evan Greene",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Mirodark Color Scheme",
 			"details": "https://github.com/djjcast/mirodark-st2",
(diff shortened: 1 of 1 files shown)


Thom1729:
It looks like you're not consistently using the `.msl` suffix for scopes.

eneerge:
I have added the missing .msl to the scopes.

Thom1729:
After reviewing the syntax more thoroughly:

- You have a lot of complex, repeated snippets, like giant lists of words. Putting these in variables would make the syntax much, much easier to read.
- You're scoping a lot of whitespace. I'm not sure that's correct.
- You're doing most of the work in giant regexp captures. This is likely to make the syntax difficult to extend, and it often leads to subtle bugs.
- You have several lookbehinds. These are very bad for performance, because they trigger the older, slower regexp engine.
- You're using capture groups in a lot of places where you're not using the result. This is likely to lead to bugs, especially in variables, and it's not great for performance, either.

FichteFoll:
You should definitely try to avoid lookbehinds as they will be detrimental in performance (relative to the other points pointed out by @Thom1729). You *may* get the package merged in its current state since it's pretty much just performance that's lacking, but I want you to acknowledge this and decide on whether you want to improve on it first.

eneerge:
Hello.  At this time, I'm not sure  how I would implement the regex without the lookbehinds.  While testing, I didn't experience any slow downs on my 2600k system.  I will review other highlighters to see if there's an alternative solution.  Since this is the only mIRC syntax highlighter available for Sublime, can we go ahead and merge?

PR #7060, opened 2018-04-16
Add Find-PHP-Vulnerabilities

WangYihang:
Short Desc:
```
A plug-in of sublime 2/3 which is able to find PHP vulnerabilities
```
Test Case: 
image

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -809,7 +809,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"tags": true,
+					"tags": true
 				}
 			]
 		},
--- a/repository/f.json
+++ b/repository/f.json
@@ -809,7 +809,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"tags": true,
+					"branch": "master",
 					"platforms": ["osx", "linux", "windows"]
 				}
 			]
--- a/repository/f.json
+++ b/repository/f.json
@@ -810,7 +810,6 @@
 				{
 					"sublime_text": "*",
 					"tags": true,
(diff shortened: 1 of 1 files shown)


WangYihang:
@packagecontrol-bot Key map override bug fixed, please recheck, thank you

WangYihang:
WTF ?????

WangYihang:
How to solve this shit error!!! Fuck

Thom1729:
How do you decide what a vulnerability is? Is there a standard for this? From the readme, it is totally unclear what kinds of vulnerabilities this will detect. In addition, the regexp-based approach does not look very reliable, and there are no tests. Are there existing PHP linters that may provide a superset of this functionality?

I see that all of the output is in (I think) Chinese. I'm not sure whether there's an actual rule about language, but as far as I know the channel maintainers don't read Chinese.

Is inserting comments really the way to go here? The simple alternative would be printing to an output panel. A more complicated approach would be to use phantoms.

WangYihang:
@FichteFoll Hi, Fichte~ I got a problem, just now I updated the documentation of this plugin, and run `git tag` command, then push the new commit to GitHub...
But I tried to search in sublime plug repo, but can not find this repo...
Should I open a new pull request everytime I updated my plugins?

FichteFoll:
It may take up to 1 hour for your new tag to get picked up by the crawler.

WangYihang:
got it, the problem is solved after waiting for a  few minutes, thank you
bro

FichteFoll <notifications@github.com>于2018年6月14日周四 上午8:39写道：

PR #7061, opened 2018-04-16
sublack

jgirardet:
Black integration for SublimeText.

Black is The Uncompromising Code Formatter : https://github.com/ambv/black

This Plugin provide usage to this formatter directly in sublimetext 3

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -3767,6 +3767,7 @@
 		{
 			"name": "sublack",
 			"details": "https://github.com/jgirardet/sublack",
+			"author": "jgirardet",
 			"releases": [
 				{
 					"sublime_text": ">=3000",
--- a/repository/s.json
+++ b/repository/s.json
@@ -3765,6 +3765,16 @@
 			]
 		},
 		{
+			"name": "sublack",
+			"details": "https://github.com/jgirardet/sublack",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"details": "https://github.com/yrammos/SubLilyPond",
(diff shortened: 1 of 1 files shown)


jgirardet:
removed workspace

jgirardet:
@abathur I tried to talk with the other one. it seems he doesn't want to talk.
Maybe can we go on ?

Thom1729:
- You're checking Sublime's version, but you're not using that information and the package is only compatible with ST3 anyway.
- In a number of places, you're invoking methods on the active window or the active window's active view, where you already have access to the pertinent view or window. This could cause the wrong view or window to be used. Examples:
  - In `get_command_line`, `sublime.active_window().extract_variables()` should be `self.view.window().extract_variables()`
  - In `get_settings`, `sublime.active_window().active_view().settings()` should be `view.settings()`.
  - In `do_diff`, `window = sublime.active_window()` should be `window = self.view.window()`.
- In `is_python`, `view.score_selector(0, "source.python") > 0` could be `view. match_selector("source.python")`.
- You might want to use `sublime.platform()` over `sys.platform`. Then, you wouldn't need `sys` at all. Same with `platform.system()`.
- Personally, I'd add a verb to the name of the `black_file` command, such as `black_format`.
- You're using the `on_save` settings key, which is fine in the package's own config but could conflict with another package in the view settings. I'd use something more specific, like `black_format_on_save`.
- In your event listener, would it work to use `on_pre_save_async` instead of the synchronous callback?

jgirardet:
@Thom1729 I finally apply all your changes (except the black_file). thanks.

I hope @abathur  will be able to add it since there is no other "black" project anymore.

PR #7062, opened 2018-04-17
Add Threejs Autocomplete package

wrbing728:

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -735,7 +735,7 @@
 					"sublime_text": ">=3000",
 					"tags": true
 				}
-			]   
+			]
 		},
 		{
 			"name": "Theme - Argonaut",
@@ -2006,6 +2006,16 @@
 				{
 					"sublime_text": "*",
 					"branch": "master"
+				}
+			]
+		},
+		{
+			"name": "Threejs Autocomplete",
+			"details": "https://github.com/wrbing728/threejs-sublime",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
 				}
 			]
 		},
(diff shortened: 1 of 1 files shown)


wrbing728:
how long will it take to handle the pr on average please?

Thom1729:
This is a fork of an existing package that's already listed in Package Control. Have you reached out to the original developer about adding features to that package?

wrbing728:
This exist package(blackjk3/threejs-sublime) pushed the last commit 3 years ago, so i fork from it and try to upgrade some new ThreeJS properties

Thom1729:
Have you reached out to the original developer about adding features to that package?

wrbing728:
Actually nope.  I just created an issue to ask if he can continue to work for his package，but have not receive a reply.

Thom1729:
You might even submit a PR with the changes you've made.

wrbing728:
Yeah, maybe I should and I just submited a PR.   
By the way, what if there is still no reply？ How long should i keep waiting?

wrbing728:
It seems like this package was not maintained any longer.   https://github.com/blackjk3/threejs-sublime/pull/6

FichteFoll:
If the author is unresponsive, you may take over maintenance of the package entry on Package Control. Please submit a new PR or adjust this PR to change ownership of the existing package to yourself (and change the `details` url). While doing so, please also rename the package to not contain a period because those are known to cause issues and we'd like to make periods illegal in package names (#2524). Once you did that, we will give the current maintainer a response period of 2 weeks until which they can object to the change.

The changes in your PR are minimal, so users probably won't notice a change anyway.

wrbing728:
I have change this PR as you said(the details url and package name),  is it all right?

PR #7063, opened 2018-04-18
Update repository.json

devil-may-cry-era:

### DIFF
--- a/repository.json
+++ b/repository.json
@@ -31,5 +31,6 @@
 		"./repository/y.json",
 		"./repository/z.json",
 		"./repository/dependencies.json"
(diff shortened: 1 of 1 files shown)


Thom1729:
The documentation is extremely sparse, and does not mention that the command *alters* the clipboard contents.

Why is it necessary to case out the line ending normalization? Sublime will normalize the line endings (logically to '\n') when you paste, anyway.

FichteFoll:
> does not mention that the command *alters* the clipboard contents.

This is a critical concern. Please address this.

Regarding line ending normalization, you certainly don't have to consider the current view's setting for this. Just consider all possibly valid line ending combinations and normalize them to `\n` before pasting, as ST will always use that internally.

PR #7069, opened 2018-04-21
added a Chaussette package to highlight chaussette file

lanathlor:


Thom1729:
Why add a repository instead of simply adding the new package to the existing repository?

FichteFoll:
@lanathlor please respond. Unless you use specific features that we don't support in the main channel, it's highly suggested to add the package directly.

You should also be targetting `>=3092`, not just `3092` (which I'm not even sure is a valid specifier).

PR #7070, opened 2018-04-21
ACPI Source Language syntax highlighting for Sublime Text

Fansaly:


Thom1729:
Recommended changes:

- `punctuation.math.asl` should be `keyword.operator.arithmetic.asl`. Likewise for other operators.
- Opening and closing quotes should have `punctuation.definition.string`.

General suggestions:

- It's always safer to wrap variables in a non-capturing group, e.g. `function_methods: (?:AccessAs|Acquire|Add)`.
- It's preferable to use non-capturing groups instead of capturing groups when you're not using the capture value.

Thom1729:
This PR can be merged, but I do recommend making the above changes so that the package will work properly with typical color schemes.

PR #7071, opened 2018-04-22
Add RemoveDuplicateLines Plugin

ilyakam:
**RemoveDuplicateLines** is a plugin for Sublime Text that allows you to remove duplicate lines from the file. [Please see the plugin's repo for details.](https://github.com/ilyakam/RemoveDuplicateLines)

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -1056,6 +1056,17 @@
 			]
 		},
 		{
+			"name": "RemoveDuplicateLines",
+			"details": "https://github.com/ilyakam/RemoveDuplicateLines",
+			"labels": ["remove duplicates", "delete duplicates", "dedup"],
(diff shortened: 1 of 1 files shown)


Thom1729:
- There's a `View.erase` method, so you don't have to replace with an empty string.
- Does this keep the *first* occurrence? The documentation should probably specify.
- You should add a `sublime-commands` file so that the command will show up in the command palette.

ilyakam:
Hi @Thom1729,

All of your concerns have been fixed and released in [version 1.1.0](https://github.com/ilyakam/RemoveDuplicateLines/releases/tag/1.1.0).
Please let me know if there's anything else that I need to address.

As always, thank you for all of the feedback!

PR #7079, opened 2018-04-23
added Romanize plugin

GCTsalamagkakis:

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -1543,6 +1543,17 @@
 			]
 		},
 		{
+			"name": "Romanize",
+			"details": "https://github.com/GCTsalamagkakis/Romanize",
+			"labels": ["numbers"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Ros Snippets",
 			"details": "https://github.com/ckrooss/ros_snippets",
 			"labels": ["snippets", "build system", "language syntax"],


Thom1729:
There is no need to submit a new PR when you make a change; simply update the old one. As this is the current PR, the others should be closed.

Thom1729:
- The name "Romanize" is cute, but "Convert to roman numerals" might be clearer. Up to you.
- You should add a `sublime-commands` file so that the command will appear in the command palette.
- The context menu item will be active and invokable even if it won't do anything. It would be better to check for this by implementing the `is_enabled` method.
- Will `4` be converted to `IV` or `IIII`? The documentation should specify.

GCTsalamagkakis:
I've added the sublime-commands file to the repository, you can check it if
you want
4 will be converted to iv, i dont see the reason for it to be converted to
iiii

Thank you

Thom1729:
It looks like the algorithm has a bug. 128 is converted to CXIXVIV, but it should be CXXVIII. Try this:

```python
special_chars = [
    (1000, 'M'),
    (900, 'CM'),
    (500, 'D'),
    (400, 'CD'),
    (100, 'C'),
    (90, 'XC'),
    (50, 'L'),
    (40, 'XL'),
    (10, 'X'),
    (9, 'IX'),
    (5, 'V'),
    (4, 'IV'),
    (1, 'I'),
]

def to_romans(decimal):
    roman = ""
    for number, symbol in special_chars:
        while decimal >= number:
            roman += symbol
            decimal -= number
    return roman

print(to_romans(128))
```

It would be a good idea to write some unit tests for this.

GCTsalamagkakis:
Is this not the same number written differently?

Στις 15 Μαΐ 2018 4:42 μ.μ., ο χρήστης "Thomas Smith" <
notifications@github.com> έγραψε:

Thom1729:
CXIXVIV = 100 + 10 + 9 + 5 + 4. This sums to 128, but I have never seen Roman numerals written that way. I think that most users would expect it to be broken out 128 = 100 + 10 + 10 + 5 + 3. That is, the algorithm should always use the largest possible symbol. Here's another example. Your algorithm converts 300 to CXCLXLXIXI. I would expect CCC instead.

On an unrelated note, you should probably have some handling for large numbers. If someone tries to convert, say, 10000000000, I don't think that the package will handle it very gracefully.

GCTsalamagkakis:
I "fixed" it, seems to be working as expected now. As far as the large
numbers are concerned, i can check if the number exceeds a certain int and
if it does, just pass, i tried 70 trillion a few days ago and sublime
crushed, though i would not expect anyone to use it on such a number for
any reason other than curiosity

Thom1729:
I would definitely add such a check. Running the package should not crush Sublime. :-)

That's the only blocker before merging, but I have a few more suggestions:

- The screenshots in the readme should be updated.
- The outer `while` loop on line 22 is unnecessary.
- If the selection is just a cursor, it might make sense to see if the cursor is inside a number, and if so convert that number.
- You might implement the `is_enabled` method of the command, checking whether any numbers are selected. This would gray out the menu item when it wouldn't do anything.
- It might be useful to provide a reverse conversion.

GCTsalamagkakis:
I will probably have it all solved by tomorrow night, I just first need you
to explain your third bullet, thanks

Στις 16 Μαΐ 2018 12:06 π.μ., ο χρήστης "Thomas Smith" <
notifications@github.com> έγραψε:

GCTsalamagkakis:
As long as a number is selected, the cursor can be anywhere, so this could
lead to unintentionally converting numbers. Also, just like bolding text or
underlining text, I think it is more intuitive if it is selected

Στις 16 Μαΐ 2018 12:22 π.μ., ο χρήστης "Thomas Smith" <
notifications@github.com> έγραψε:

GCTsalamagkakis:
Hello,

i've made some changes on the files, i think everything is ok now but maybe
you should check it too first

Thank you

Thom1729:
You should use the same filtering logic in `run` and `is_enabled`. That way, the command will be enabled if and only if it would do something when run. For example:

```python
LARGEST_NUMBER_TO_CONVERT = 1000000000

class RomanizeCommand(sublime_plugin.TextCommand):
    def get_number_selections(self)
        for region in self.view.sel()
            if region.empty(): break

            text = self.view.substr(region)
            if not re.match(r'[0-9]+', text): break

            number = int(text)
            if text > LARGEST_NUMBER_TO_CONVERT: break
            
            yield (region, int(text))

    def is_enabled(self):
        return any(self.get_number_selections())
    
    def run(self, edit):
        for region, number in self.get_number_selections():
            self.view.replace(edit, region, to_romans(number))
```

GCTsalamagkakis:
it's done, i'll update readme

Thom1729:
The size check in `to_romans` is now redundant.

GCTsalamagkakis:
fixed

GCTsalamagkakis:
hello, can you try installing and using the package and then tell me if
anything is wrong? Thanks

PR #7080, opened 2018-04-23
Add `industrial-coding-theme` sublime text theme

zaizupro:
it is sublime text color theme

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -548,6 +548,17 @@
 			]
 		},
 		{
+			"name": "industrial coding theme",
+			"details": "https://github.com/zaizupro/industrial-coding-theme",
+			"labels": ["theme", "color scheme"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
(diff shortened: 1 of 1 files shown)


Thom1729:
Command names should be UpperCamelCased. `ictActivateCommand` should be `IctActivateCommand` or the like.

Actually, what's the point of that command? There are already commands to select themes and color schemes.

Thom1729:
@zaizupro ping

zaizupro:
The point of that command is setting Theme and Color Scheme in one action,
also i didn't  know about "UI: Select Color Scheme" and "UI: Select Theme"
commands till that moment, and i copied that functional from :
github.com/dempfi/ayu


I changed command to UpperCamelCased and also i can remove dat command from
command palette.

jrappen:
@zaizupro you can use [PackageDev](https://github.com/sublimetext/PackageDev) to convert the color scheme to the **new** [color scheme format for ST3149+](https://www.sublimetext.com/docs/3/color_schemes.html)

PR #7082, opened 2018-04-24
Utility for running programs (i.e. starting a Node server) when switching Sublime Text project

andrew4699:
When you switch projects, assuming you've set this up in your project file, it will try to start the program you tell it to. Intended for use with node (npm start).

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -2147,6 +2147,17 @@
 			]
 		},
 		{
+			"name": "Project Runner",
+			"details": "https://github.com/andrew4699/sublime-project-runnner",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": ["windows"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Project Specific Syntax Settings",
 			"details": "https://github.com/reywood/sublime-project-specific-syntax",
 			"labels": ["language syntax", "project"],


Thom1729:
`ProjectRunner` is an application-global listener, but it looks like you're putting window-specific data in it. Because there's no `WindowEventListener` class, your best bet is keeping a `dict` mapping `window.id()` to a window-specific settings object.

Also, I wouldn't give `StartHandler` a reference to the `ProjectRunner`. Let `StartHandler` keep track of its own state.

You're not actually running the user's command. You have it hardcoded: `os.system("npm start")`.

`show_quick_panel` is probably the wrong way to ask for user confirmation. `sublime.ok_cancel_dialog()` would be better. You could use that same dialog to show the user what command would be run.

andrew4699:
Thanks for the suggestions. I've gone ahead and implemented them and squashed some other bugs in the process.

Thom1729:
You should take out the commented-out code and the print statements.

I see that you have switched from using `subprocess` to using `os.system`. Using `os.chdir` is not a good idea, because you're sharing the working directory with the entire runtime and any other package could interfere with it.

andrew4699:
Went back to `subprocess` and did it properly. I used `os.system` before because I couldn't get it to display the command prompt.

I also took out the commented code/print statements.

PR #7083, opened 2018-04-24
Add GBKToUTF8

kryond:


Thom1729:
Why on earth is this making network calls?

Thom1729:
@kryond ping

PR #7084, opened 2018-04-24
Whitespace Removal plugin

Melodatron:
Removes all the whitespace characters in a selection.

### DIFF
--- a/repository/w.json
+++ b/repository/w.json
@@ -360,6 +360,16 @@
 			]
 		},
 		{
+			"name": "Whitespace Removal",
+			"details": "https://github.com/Melodatron/WhitespaceRemoval-Sublime",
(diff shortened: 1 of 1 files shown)


Thom1729:
You should have a `sublime-commands` file so that the command is accessible from the command palette.

The Mac OS keybind in the keymap doesn't match the documentation.

Thom1729:
@Melodatron ping

Melodatron:
@Thom1729 Updated as per feedback and added a new release to the repo

Thom1729:
That's because that's the backspace key. It's only interpreted as the delete key when used with fn. If you press ctrl+command+delete on an ordinary keyboard on the Mac OS, it will not trigger your keybinding.

PR #7085, opened 2018-04-24
Add svelte syntax definition package

corneliusio:
This package adds syntax highlighting for components built with [Svelte](https://svelte.technology/).  
No such package exists yet.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -4871,6 +4871,16 @@
 			]
 		},
 		{
+			"name": "Svelte",
+			"details": "https://github.com/corneliusio/svelte-sublime",
+			"releases": [
+				{
+					"sublime_text": ">=3153",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "SVG Icon Snippets",
 			"details": "https://github.com/idleberg/sublime-svg-icons",
 			"labels": ["snippets", "auto-complete", "svg", "svg icons"],


corneliusio:
Sorry, my tag just hadn't pushed with the original commit. Has been fixed now.

corneliusio:
Hmm, I actually updated any reference to the core HTML syntax about 3 days ago when making some other updates, so shouldn't be referencing anything out of date unless I'm just totally spaced on something.

I've also been using this since build 3163 and am up to date now (3175 on the dev branch) without any issues. (I believe 3.1 was 3170, yeah?) If there's something that sticks out to you that I'm not seeing just let me know and I'd be happy to smooth it out, as it stands I've not run into any bugs.

I actually referenced Vue syntax definitions quite a bit for this (and actually have my own Vue syntax def that I use for some a little more specificity in definitions). There are a few key differences between Vue SFCs and Svelte components that cause the implementation to be a bit different (e.g. no `<template>` scope, control flow being outside tag attribute scope). But, I'm really open to suggestions on quality improvements as I've not actually published a syntax def before—was trying to keep this iteration as minimal as possible and take advantage of as much built-in HTML definitions as possible.

Thom1729:
What I was thinking of in particular is `script-embeds`. The HTML syntax is using a much more robust approach to identifying languages. The Vue syntax uses macros to extend this system in a graceful manner. I think it might be a cleaner solution -- but then, I'm hardly unbiased.

corneliusio:
Thanks for the feedback, I’ll take a closer look at those and see about cleaning it up. The script and style embeds are actually some of the oldest parts of the definition and we’re actually pulled from Vue’s definitions quite a long while ago, just didn’t revisit them when I updated this to Svelte 2.0 and decided to publish.

PR #7089, opened 2018-04-29
Add NumberFormatter Plugin

ilyakam:
# NumberFormatter - A Sublime Text Plugin

This [Sublime Text](http://www.sublimetext.com/) plugin formats plain numbers by adding a thousands separator to them. It also unformats them in reverse. See the [Features](#features) section below for details.

## Features:

As long as the number is partially or fully selected, or while the cursor is on the number, this plugin will format or unformat it in the following way:

| Before      | After       |
| ----------- | ----------- |
| `12345`     | `12,345`    |
| `12,345`    | `12345`     |
| `12345.67`  | `12,345.67` |
| `12,345.67` | `12345.67`  |

This plugin supports multiple cursors and multiple selections. It can process both formatted and unformatted numbers in one go.

---

Please see [the plugin's homepage](https://github.com/ilyakam/NumberFormatter) for mode details.

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -1020,6 +1020,17 @@
 			]
 		},
 		{
+			"name": "NumberFormatter",
+			"details": "https://github.com/ilyakam/NumberFormatter",
+			"labels": ["format numbers", "unformat numbers", "thousands separator", "decimal separator"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Numix Theme",
 			"details": "https://github.com/nauzethc/sublime-text-numix",
 			"labels": ["numix", "theme"],


ilyakam:
Hey @tmelot2, thanks for listing out the detailed findings! I was able to fix the first and last issues in [version 1.0.1](https://github.com/ilyakam/NumberFormatter/releases/tag/1.0.1). As for the second issue, it's an interesting bug that definitely needs to be addressed. I've spent about an hour doing some research and finding potential solutions for it. For the time being, I have filed it as a new issue [here](https://github.com/ilyakam/NumberFormatter/issues/5). I'll be sure to resolve it soon.

Please let me know if there's anything else that I need to do in order for you to add this plugin to the Package Control Channel.

tmelot2:
I did a pull, running 4a3e74bb now. Noticed an error in another case that might be common: multiple selection with blank line in the middle:

```
2222

22222
```

with a selection on each line yields:

```
  File "/Applications/Sublime Text.app/Contents/MacOS/sublime_plugin.py", line 1072, in run_
    return self.run(edit)
  File "~/Library/Application Support/Sublime Text 3/Packages/NumberFormatter/number_formatter.py", line 91, in run
    if self.is_number(self.view.substr(expanded_region)):
  File "/Applications/Sublime Text.app/Contents/MacOS/sublime.py", line 847, in substr
    s = sublime_api.view_cached_substr(self.view_id, x, x + 1)
TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'
```

Regarding the rounding bug, I think making a note in the readme will be good while you look into potential solutions.

ilyakam:
Hi again @tmelot2,

All of the issues that were brought up in this PR have been addressed. Please [check out release 1.0.2](https://github.com/ilyakam/NumberFormatter/releases/tag/1.0.2) for the latest and greatest this plugin has to offer.

Thank you so much for helping me to QA this plugin! I think that it's important to maintain at least some level of quality in the plugins that are provided by Package Control and I really appreciate you taking the time to personally check each submission.

Please let me know if there's anything else that you'd like me to do before you can merge this PR.

Thom1729:
You're fetching all of the settings once, outside the command. This means that they won't be reloaded when the user changes them. Also, you're specifying default values both in the `settings.get` call and in the default settings file, which isn't a big deal but is redundant.

Also, have you thought about using native Python string formatting for the implementation?

ilyakam:
> You're fetching all of the settings once, outside the command. This means that they won't be reloaded when the user changes them.

Agreed. This has been fixed in https://github.com/ilyakam/NumberFormatter/pull/15 and deployed in [version 1.0.3](https://github.com/ilyakam/NumberFormatter/releases/tag/1.0.3).

> Also, you're specifying default values both in the `settings.get` call and in the default settings file, which isn't a big deal but is redundant.

I figured it's nice to have a fallback, just in case.

> Also, have you thought about using native Python string formatting for the implementation?

I have actually. Prior to https://github.com/ilyakam/NumberFormatter/pull/8/, I was using the built-in thousands formatter `"{:,}".format(number)`. The problem was that it required `int`egers or `float`s to be passed to it and as @tmelot2 pointed out in this thread, when `float()`ing a large number like `9999999999.9999999999`, it would automatically round up to `10,000,000,000.0`.

Unless you're referring to [Python's custom format handling](https://pyformat.info/#custom_1), which I honestly don't have much experience with. Please let me know if you feel like [my solution](https://github.com/ilyakam/NumberFormatter/blob/develop/number_formatter.py#L124-L134) is too amateurish and I'll revisit it.

Thank you for the feedback, @Thom1729!

Thom1729:
It makes sense now that you've pointed out the floating point thing.

tmelot2:
Hey thanks for getting this wrapped up Thom!

PR #7093, opened 2018-05-05
Split up ST2 and ST3 releases for Jedi

srusskih:


Thom1729:
You should have `>=3000` instead of `>3000`.

srusskih:
@Thom1729 thanks! updated

PR #7098, opened 2018-05-05
include PythonNiceToHaveFeatures in the list

monobot:


Thom1729:
All of your commands should end in `Command` (see `RegionModifier`). You might also want to use slightly more detailed names (e.g. `RelativePathCommand` to `CopyRelativePathCommand`).

In `is_enabled`, should the commands be available only for Python files?

In `get_path`, instead of `sublime.active_window()` you should use `self.view.window()`.

PR #7099, opened 2018-05-06
Liquid was transferred and upgraded

braver:
The Liquid package was broken and unmaintained. It has been transferred to me and upgraded to ST3 features and tagged releases.


Thom1729:
I assume that you've talked to the original maintainer, or attempted to?

PR #7102, opened 2018-05-08
Add DocumentSearch package

dave-ellis:
Searches all documents in a project using tf-idf and displays the results in scratch view.  Future enhancements will improve on this initial rough implementation and will add a page rank implementation for personal wikis.

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -1070,6 +1070,16 @@
 			]
 		},
 		{
+			"name": "Document Search",
+			"details": "https://github.com/dave-ellis/DocumentSearch",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Dog Ears",
 			"details": "https://github.com/skyronic/DogEars",
 			"releases": [


Thom1729:
What is the relationship between this package and the [FindInProject](https://github.com/Wramberg/FindInProject) package?

dave-ellis:
This package used the FindInProject package source as a starting point as it had very similar UI handling that I wanted, however it has a different search and ranking mechanism.  In particular, it uses tf-idf and the next version will combine these results with page ranking on results to improve results ordering (in final testing).

dave-ellis:
@Wramberg What are your thoughts on having these changes merged back into your package?

Wramberg:
@dave-ellis I haven't looked into the specifics of your package but it sounds like we could find a way to extend FindInProject with your features. If we can do this without removing any of the original functionality that would be great. Maybe the different behaviors can be chosen by the user through settings.

I don't have much time to spend on packages these days so I've added you as collaborator on FindInProject.

dave-ellis:
Okay, that makes sense.  I will merge DocumentSearch back into FindInProject instead of adding a new package.

PR #7103, opened 2018-05-08
Add Mika markup language syntax highlighting

kyoto-shift:
A syntax highlighting package for the Mika markup language.

Thank you for your consideration.


Thom1729:
You need to change `"sublime_text": ">=3000"` to `"sublime_text": ">=3092".

PR #7105, opened 2018-05-10
Add Grammalecte

procsynth:
This package enable the usage of Grammalecte, a grammatical checker for french, inside Sublime Text.
This is the only one doing that at the moment.

screenshot from 2018-05-10 21-26-11

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -1754,6 +1754,18 @@
 			]
 		},
 		{
+			"name": "Grammalecte",
+			"details": "https://github.com/procsynth/GrammalecteST3",
+			"readme": "https://raw.githubusercontent.com/procsynth/GrammalecteST3/master/README.md",
+			"labels": ["gramma", "french", "spelling", "text"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Grammatical Framework",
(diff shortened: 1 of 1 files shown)


Thom1729:
A `is_enabled` that always returns `True` is a no-op.

Rather than use an `EventListener` that always runs, it would be better to use a `ViewEventListener` that only runs for plain text views (or somesuch). This will be especially important when the `on_modified_async` is implemented.

procsynth:
Hi,
I've done the modifications you asked for. Those modifications enabled me to implement some more features : https://github.com/procsynth/GrammalecteST3/releases/tag/0.0.3

procsynth:
Thank you for taking the time to review my code.

- In the future, the best will be to work with the scope indeed. But I think that change will not be quick. I'd rather prefer for now to listen for events from every view. For now, it will return `True` if the plug-in is running for the current view.
- I believe the default config file is here : https://github.com/procsynth/GrammalecteST3/blob/master/Grammalecte.sublime-settings
- Fixed
- I was not familiar with `enumerate`, fixed, thanks
- Also fixed 
- Got rid of the temp file

With some work, and modification to the libs, I'm now able to use the libs without requiring a shell. So I don't need to run unpacked anymore. I tested, it seems to work out of the shelf.

Thanks for your input.

PR #7106, opened 2018-05-10
added OLED Color Scheme

pavelrevak:

### DIFF
--- a/repository/o.json
+++ b/repository/o.json
@@ -239,6 +239,17 @@
 			]
 		},
 		{
+			"name": "OLED Color Scheme",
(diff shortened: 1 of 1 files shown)


pavelrevak:
What else I need? there is file ending with `.sublime-color-scheme` and this file is working with latest sublime text.
https://github.com/pavelrevak/OLED-color-scheme
Thanks.

pavelrevak:
https://github.com/packagecontrol/st_package_reviewer/pull/20

pavelrevak:
"ERROR: with the package does not define any file that interfaces with Sublime Text" **is not an error**,
the package contain correct file, but automated tests don't know the file `.sublime-color-scheme`.

PR #7111, opened 2018-05-14
Added shorn-mariana colour scheme

shornuk:

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1465,6 +1465,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


Thom1729:
You can ignore this error: `ERROR: The package does not define any file that interfaces with Sublime Text`.

The others should be addressed. You should also add a license.

Thom1729:
@shornuk ping

jrappen:
I'm not even sure whether or not the original Mariana color scheme was released with an open source license, to begin with. There is no license information in @wbond 's original file.

PR #7116, opened 2018-05-18
Add c2console

meloalright:
Add a simple sublime plugin called [c2console](https://github.com/meloalright/c2console) to query console automatically when type letter c

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -120,6 +120,17 @@
 			]
 		},
 		{
+			"name": "C2console",
+			"details": "https://github.com/meloalright/c2console",
+			"labels": ["javascript", "console", "log"],
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


Thom1729:
Is there any reason not to just use a `sublime-completions` file instead of writing a completions plugin?

meloalright:
Yeah.. however  a `sublime-completions` maybe like this 
```
   "completions":
   [
      { "trigger": "c", "contents": "console.log($0)" },
      { "trigger": "c", "contents": "console.table($0)" },
      { "trigger": "c", "contents": "console.err($0)" },
      { "trigger": "c", "contents": "console.info($0)" }
   ]
``` 
could generate completions by constantly pressing the tab button but not create such a completions list like the console of chrome

PR #7120, opened 2018-05-19
add eslint fix

kufii:
runs eslint --fix on the current file

This differs from the other ESLint plugins in the repository because it runs the fix on the current buffer, instead of the last saved instance. I tried creating PRs to other plugins before creating this plugin but they've been ignored for months.

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -1113,6 +1113,16 @@
 			]
 		},
 		{
+			"name": "ESLint Fix",
+			"details": "https://github.com/kufii/Sublime-ESLint-Fix",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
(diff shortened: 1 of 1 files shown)


kufii:
Hey, any update on this?

Thom1729:
Sorry, I made a mental note to review this and then lost track of it.

You're modifying `sys.path`, which is bad because that's shared among all packages. Instead, you should use relative imports.

kufii:
Thanks for the feedback! I've switched to relative imports and created a new release. Any other comments?

kufii:
Could you give me an update on the status of this when you get a chance? thanks!

FichteFoll:
Sorry for the delay.

kufii:
No problem at all! thanks for merging

PR #7121, opened 2018-05-21
Added new package: Lux (syntax highlighting and basic functionality for the Lux language)

jokaah:


jokaah:
Added tag

Thom1729:
I'd recommend updating the syntax to the `sublime-syntax` format. There's no reason to use `tmLanguage` unless you need backward compatibility.

PR #7123, opened 2018-05-22
add entry for dm-macro

truemoid:


truemoid:
I have changed the version tag. Hope it is in the right format now.

PR #7125, opened 2018-05-22
Add entry for dm-macro package

truemoid:

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -939,6 +939,16 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


truemoid:
The '.sublime-syntax' format is supported in the stable releases of Sublime Text 3. The older specification should not need to be supported. Do I need to provide a '.tmLanguage' file to progress this package?

Thom1729:
I'd recommend factoring some of the repeated regexp snippets out into variables.

PR #7126, opened 2018-05-23
Remove Grunt package

tvooo:
This PR removes the Grunt package (https://github.com/tvooo/sublime-grunt), which is no longer maintained. If someone is willing to pick up maintenance, I'm happy to hand it over; it can then be re-added to Package Control.

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -2009,16 +2009,6 @@
 			]
 		},
 		{
-			"name": "Grunt",
-			"details": "https://github.com/tvooo/sublime-grunt",
-			"releases": [
-				{
-					"sublime_text": "*",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "Grunt Snippets",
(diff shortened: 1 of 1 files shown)


Thom1729:
Is it necessary to actually remove the package?

tvooo:
It seems to be working for some people still, but I get a bug report every once in a while that I'm not intending to address... along with comments from people complaining about that very fact.

Thom1729:
If you really want to remove it, I'll do that. But I don't think it's doing any harm just sitting there. I'd recommend opening an issue declaring that the package is no longer maintained and asking if anyone wants to take it over.

Thom1729:
@tvooo Do you still want to remove this package?

tvooo:
Thanks, let's leave it in for now.

FichteFoll:
When the amount of issues (for some reason) becomes unbearable, you can set your repository to read-only. That should be clear enough to everyone.

PR #7129, opened 2018-05-23
Add chester-atom theme to package list

gborges0727:


Thom1729:
You should include the original copyright and license information.

gborges0727:
I've updated the repository with the original license included

PR #7131, opened 2018-05-25
Add Mojom syntax definition URL to channel.json

sdefresne:
This package provides syntax highlighting for [mojo binding definition files](https://chromium.googlesource.com/chromium/src/+/master/mojo/README.md).

### DIFF
--- a/channel.json
+++ b/channel.json
@@ -95,6 +95,7 @@
 		"https://raw.githubusercontent.com/royisme/SublimeOctopressTool/master/packages.json",
 		"https://raw.githubusercontent.com/rpowers/sublime_stata/master/packages.json",
 		"https://raw.githubusercontent.com/rse/sublime-scheme-rse/master/packages.json",
+		"https://raw.githubusercontent.com/sdefresne/sublime-mojom/master/packages.json",
 		"https://raw.githubusercontent.com/sdolard/sublime-jsrevival/master/packages.json",
 		"https://raw.githubusercontent.com/seanliang/ConvertToUTF8/master/packages.json",
 		"https://raw.githubusercontent.com/sentience/DokuWiki/master/packages.json",


sdefresne:
Added missing "date" key...

Thom1729:
What is the purpose of adding a new channel rather than adding a new package to the default channel?

sdefresne:
@Thom1729:

I wrote a package defining syntax highlighting for mojom file. As I don't like writing plist (which is the syntax used by Sublime Text for its tmLanguage file), my repository does not contains the tmLanguage file directly but instead contains some file that need to be preprocessed. So I cannot directly add it as a github repository and instead need to distribute it as a sublime-package file.

The documentation on how to distribute sublime-package appears to be sparse on https://packagecontrol.io/docs/submitting_a_package but seems to point to adding a channel.

Here is the relevant part of the documentation:

> You‘ll need to pick one of the following two hosting options:
> 
> 1. Use a public GitHub or BitBucket repository. Only include one package per repository and be sure the root of the package is the root of the repository.
> 2. Host .sublime-package files and a packages.json on a web server with SSL. For each release you‘ll need to create and upload a new package file and update the packages.json information. Hosting of the files must be done on a secure server to help ensure the security of users‘ machines. Unsecure URLs could lead to malicious code being loaded and run. This is rarely used – see example-repository.json for documentation.

So I looked at how other sublime-package have recently be added in this repository. They appear to be doing that by adding a new channel. This is the reason I tried to do it this way. Maybe this is incorrect. If so, I'd be more than happy if you could point me to information on how I can add my package to package control. Thanks.

Thom1729:
The methods I'd recommend, in order, are:

1. Write a [`sublime-syntax` definition](https://www.sublimetext.com/docs/3/syntax.html) instead of a `tmLanguage`. The new format is more powerful and easier to use. It's compatible with Sublime Text 3 (build >= 3043, which includes 3.0 and above) but not generally with other editors.
2. If you need compatibility with Sublime Text 2 or with other editors, write a [`yaml-tmLanguage`](http://docs.sublimetext.info/en/latest/extensibility/syntaxdefs.html) definition and compile it with [PackageDev](https://github.com/SublimeText/PackageDev/). Then, your package would contain both the source `yaml-tmLanguage` file (which editors will ignore) and the compiled `tmLanguage` file.

sdefresne:
Thank you for the possible alternatives. I'll try to go with option 1. Should I close this pull request and reopen it when I've converted my package (and updated the pull request to add a package definition)?

Thom1729:
Yes, in this case I would close this PR and open a new one.

FichteFoll:
Quick note that the required build for sublime-syntax is 3092 (and the review bot will remind you of that if you specify a lower build).

PR #7132, opened 2018-05-25
add new package 'HTML Syntax in Script'

cmcleese:
Add syntax highlighting for all mime types in script tags that are beginning with `text/` .

**Ex:** `<script type="text/template"></script>`

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -1062,6 +1062,17 @@
 			]
 		},
 		{
+			"name": "HTML Syntax in Script",
+			"details": "https://github.com/cmcleese/sublime-html-script-all",
+			"labels": ["syntax", "html", "script", "highlighting"],
+			"releases": [
+				{
+					"sublime_text": ">=3170",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


Thom1729:
```
(?i)(?=text/html(?!{{unquoted_attribute_value}})|'text/'*|"text/"*)
```

This regexp will match the empty string. I don't think that's what you intended.

cmcleese:
This regex has now been updated to include the unquoted text just incase:
`(?i)(?=text/(?!{{unquoted_attribute_value}})|'text/'*|"text/"*)`
It matches anything beginning with `text/` .

Thom1729:
I think that what you want is:

```
(?i)(?=text/(?:{{unquoted_attribute_value}})?|'text/[^']*'|"text/[^"]*")
```

cmcleese:
Thanks, this seems to work better.
I updated the branch.

Thom1729:
You should probably add a license. Sorry I missed this before.

cmcleese:
Thanks for the suggestion. I added a license to the repo.

PR #7133, opened 2018-05-27
adding theme to N's

argyleink:

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -565,9 +565,9 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


Thom1729:
You should probably add a license.

argyleink:
good call.
added a license, created a new release

jrappen:
@argyleink you can use [PackageDev](https://github.com/sublimetext/PackageDev) to convert the color scheme(s) to the **new** [color scheme format for ST3149+](https://www.sublimetext.com/docs/3/color_schemes.html)

PR #7136, opened 2018-05-30
Added STVenvWrapper Package to s.json. 

nvanwitt:
STVenvWrapper is a wrapper script for the SublimeREPL package allowing easier access to Virtual Environment REPLs. The current Virtual Environment REPL from SublimeREPL is outdated and the code is relatively difficult to navigate for those just familiarizing themselves with the package. This wrapper package simply allows a more streamlined approach to Venv access in a user-friendly manner.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -3688,6 +3688,16 @@
 			]
 		},
 		{
+			"name": "STVenvWrapper",
+			"details": "https://github.com/nvanwitt/STVenvWrapper",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Stylefmt",
 			"previous_names": ["CSSfmt"],
 			"details": "https://github.com/dmnsgn/sublime-stylefmt",


nvanwitt:
Removed .no-sublime package as it indeed was unnecessary.

nvanwitt:
Is the Rate Limit on my end or through the control-bot?

nvanwitt:
It does not seem others have had this issue in the past so I've looked into this on my end and I can't seem to find anything that would throw this error.
I've run curl -i "https://api.github.com/rate_limit" and I get the following response:
```
X-RateLimit-Limit: 60
X-RateLimit-Remaining: 58
```

From what I understand, I am not exceeding my api calls.  I'll try one more time in a few hours to close/repoen and see if this issue is resolved.

Any advice/explanation would be greatly appreciated.

nvanwitt:
Attempting to resolve Rate Limit Issue

nvanwitt:
Hi, just wanted to know if there is anything more to review for the package? Any advice would be appreciated.

Thom1729:
I haven't had the chance yet to dig into the code, and I'm not familiar with SublimeREPL. It might help if you gave a brief summary of the shortcomings of SublimeREPL that this addresses and how this package interacts with SublimeREPL.

nvanwitt:
Hey Thom,

Sure thing!

The command ```SublimeREP: Python - virtualenv``` does not work by default. 
The error seems to be in the obfuscated ```lang_integration.py``` file in which the ```PythonVirtualenvRepl``` sublime command is defined.

The first issue I saw was that the code relies upon an older version of Virtual Envs where the activation file is a python file (```activate_this.py```) as opposed to current models which generate a shell script ```activate``. 
Not only that, but when following the code it seems that the only relevance for setting the environment variable ```"SUBLIMEREPL_ACTIVATE_THIS"``` is for this bit of code in ```ipy_repl.py```: 
```if activate_this:
    with open(activate_this, "r") as f:
        exec(f.read(), {"__file__": activate_this})
```
Which beyond reading the python file has no use in actually generating the REPL.

The second issue was that REPL generation for the virtual environment was dependent on the user also having IPython.

So I did try to do some basic edits on the file but in the face of many exceptions, I decided it was easier to just circumvent the entire thing and just launch the REPL based of the intended Virtual Environments python executable (which is really all that is needed).

I also added more user interface features to make it simpler to access (using Sublimes menu/context/keyboard shortcuts). By pressing ```cmd/ctrl + shift + y``` a user can launch the quick panel which shows all designated virtual environment paths (as provided by the user in the settings file) and, where possible, it will put the venv for the current project path on top of the list.

Sorry if that's a lot, please let me know if there's anything else I should add/elaborate on.

nvanwitt:
Hi Thom, 

Did you get a chance to look over the code/SublimeREPL?

nvanwitt:
Hey Thom,

I went through and made changes to reflect your review. 

- I changed ```SublimeVenvWrapper``` to ```RunVenvSelectorCommand``` and updated it to a WindowCommand. 
- I removed all debugging statements.
- Changed iter syntax
- The package will now detect SublimeREPL on launch and give users a pop-up.
- I changed the layout of the ```venv_list``` creation to something a little less cluttered and should also handle more use cases (I assume when you say ```a lot of errors``` that you were referring to finding/handling of venv paths).

Could I have your thoughts on the above and if I missed anything?

nvanwitt:
@Thom1729 Have you had the chance to look over the changes I made?

Thom1729:
Not yet, but probably will this weekend.

Thom1729:
- You're checking for SublimeREPL in the filesystem, but if it's installed as a zip you'll miss it. I recommend importing something from SublimeREPL or using Package Control's API.
- You're calling `exit()`. It looks like you should simply `return`.
- `if 'activate' and 'python' in os.listdir(venvpath):` I don't think that this does what you want it to.

nvanwitt:
Hey Thom,

- Unless I'm mistaken, SublimeREPL should not be zipped due to the .no-sublime-package. Do you think it's worthwhile to use import/PC API in the off chance that is altered?

- You're right on the return, I changed that.

- For this bit, I wrote that with the thought that it would be quite rare to have files with the names 'activate' and 'python' by chance so it would be a decent filter. However, considering you highlighted this, I changed this section to ```if shutil.which(os.path.join(venvpath, 'python')) and os.path.isfile(os.path.join(venvpath, 'activate'))```. This should verify the existence of both the python executable and activate file, respectively.

Also,  just wanted to say thank you for continuing to look over the code, I realize this has taken up some of your time.

Thom1729:
> Do you think it's worthwhile to use import/PC API in the off chance that is altered?

Probably. Not absolutely necessary, though.

PR #7137, opened 2018-05-31
Adds websocket-client to dependencies

jcoc611:
A compliant websocket client library for packages that want to stay up to date with a server state.

### DIFF
--- a/repository/dependencies.json
+++ b/repository/dependencies.json
@@ -1014,6 +1014,20 @@
 			]
 		},
 		{
+			"name": "websocket-client",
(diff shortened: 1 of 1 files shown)


Thom1729:
If it isn't too much hassle, I think it would make more sense to factor out `six` as a separate dependency. I know there are plenty of packages that would use it.

jcoc611:
@Thom1729 Absolutely, I've gone ahead and added the six package as `python-six`, and updated the repo for `websocket-client` by removing its copy of `six`.

PR #7138, opened 2018-05-31
Update d.json

alchemi5t:
This package tries to highlight all the words(written in devanagari script) in a view which do not conform to the context free grammer which dictates the construct of a word. for e.g.,( ूूूूूू) consecutive occurences of Matras(ूूू -> this is a matra, which i a class of vowel modifier) are not allowed. And also, there aren't any other grammar checkers for devanagari as far as i know.

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -480,7 +480,7 @@
 			]
 		},
 		{
-			"name": "devanagari_invalid_grammar_detect",
+			"name": "Detect Invalid Devangari",
 			"details": "https://github.com/alchemi5t/devanagari_invalid_grammar_detect",
 			"releases": [
 				{
--- a/repository/d.json
+++ b/repository/d.json
@@ -480,6 +480,16 @@
 			]
 		},
 		{
+			"name": "devanagari_invalid_grammar_detect",
+			"details": "https://github.com/alchemi5t/devanagari_invalid_grammar_detect",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Devastate",
 			"details": "https://github.com/vlakarados/devastate",
 			"labels": ["theme", "color scheme"],


alchemi5t:
fixed tag

alchemi5t:
any reviews? newbie here. any instructions would be helpful. yet to upload a plugin for making devanagari text readable, waiting on this one's reviews so that the next process would go smoother. please help :D

Thom1729:
Everything I know about the Devangari script I learned in the last ten minutes on Wikipedia, but from the list of rules at the end of the plugin, it looks like a regular expression could do the matching. Is a context-free grammar really necessary? It seems to add an awful lot of code.

alchemi5t:
Trying to see if a regex can handle it. will respond as soon as done.

alchemi5t:
Have uploaded the regex version. Anything else?

Thom1729:
That's much better, and it's a lot less code for me to review.

You can make that regex easier to work with by using a multiline string with the `(?x)` flag and by using `format` rather than string concatenation. Example (that I haven't double-checked):

```python
reg = r'''
(?x)
(
    (
        (({C}{N}?{H})?({C}{N}?{H})?({C}{N}?{H}))?
        ({C}{N}?){M}?{D}?
    )
    |{V}{D}?
)+
(
    ({C}{N}?{H})?
    ({C}{N}?{H})?
    ({C}{N}?{H})
)?
$
'''.format(C=C, N=N, H=H, D=D, V=V, M=M)
```

In any case, because this expression does not change, it would be better to define it once, outside the event listener, rather than recreating it every time it runs.

Here:

```python
words = re.split("\n| ",content)
		for idx,word in enumerate(words):
			if len(word)==0:
				del words[idx]
```

Instead, you should probably use `re.split(r'\s+')` to avoid empty words; then you don't have to filter them.

To create the list of valid words, it would be simpler to just use a list comprehension:

```python
valid_words = [ word for word in words if re.match(reg, word) ]
```

Also, if all the plugin does is highlight valid words, you can just do `regions = view.find_all(reg)`, where `reg` is the main expression that recognizes valid words.

alchemi5t:
Multi line regex works!, deleted removing empty strings because it was unnecessary, list comprehension used! is everything else ok?

Thom1729:
FYI, you don't have to close and reopen the issue. Just post a comment and I will see it.

Some more comments:

- The name of the package should be human-friendly. I suggest "Detect Invalid Devangari" instead of "devanagari_invalid_grammar_detect".
- Now that you're excluding non-Devangari characters, it looks like you could save several steps by searching the file for sequences of Devangari characters. There's no need to read the whole view. First, define `Charset` to include (rather than exclude) all Devangari characters. Then, your main function can be as simple as:

```python
    def dothings(self, view):
        valid_regions = [
            region
            for region in view.find_all(r'{Charset}+'.format(Charset=Charset))
            if re.match(reg, view.substr(region))
        ]
        if valid_regions:
            view.add_regions('valid', valid_regions, 'valid')
```

- The class `isbetter` should be named something more descriptive, like `DetectInvalidDevangariListener`.
- The error handling in the callbacks is very bad. It would be better to not catch at all and let the error show in the console.
- I would factor out `dothings` into a `TextCommand` and call `view.run_command` in the listener. Then, you can create a `sublime-commands` file to expose the command in the command palette.
- It would also be useful to let the user turn the listener on and off via settings. To do this, you would add a `Detect Invalid Devangari.sublime-settings` file with default settings, and check those settings in the listener.

Putting all of this together, that would be something like this:

```python
import sublime
import sublime_plugin
import re

consonants=["ॿ","ॾ","ॼ","ॻ","क","ख","ग","घ","ङ","च","छ","ज","झ","ञ","ट","ठ","ड","ढ","ण","त","थ","द","ध","न","प","फ","ब","भ","म","य","र","ऱ","ल","ळ","व","श","ष","स","ह"]
vowels=["ॷ","ॶ","ॵ","ॴ","ॳ","ॲ","अ","आ","इ","ई","उ","ऊ","ऋ","ऍ","ऎ","ए","ऐ","ऑ","ऒ","ओ","औ","ॠ"]
matras=["ा","ि","ी","ु","ू","ृ","ॅ","ॆ","े","ै","ॉ","ॊ","ो","ौ","ऻ","ऺ","ॄ","ॆ","ॏ","ॗ","ॖ"]
vowelMod=["ँ","ं","ः"]
halant=["्"]
nukta=["़"]

def character_class(characters):
    return '[' + ''.join(characters) + ']'

Charset=character_class(consonants+vowels+matras+vowelMod+halant+nukta)
C=character_class(consonants)
N=character_class(nukta)
H=character_class(halant)
D=character_class(vowelMod)
M=character_class(matras)
V=character_class(vowels)

reg = r'''
    (?x)
    (
        (
            (({C}{N}?{H})?({C}{N}?{H})?({C}{N}?{H}))?
            ({C}{N}?){M}?{D}?
        )
        |{V}{D}?
    )+
    (
        ({C}{N}?{H})?
        ({C}{N}?{H})?
        ({C}{N}?{H})
    )?
    $
'''.format(C=C, N=N, H=H, D=D, V=V, M=M)

class DetectInvalidDevangariListener(sublime_plugin.EventListener):
    settings = sublime.load_settings("Detect Invalid Devangari")

    def on_load_async(self, view):
        if self.settings.get("run_on_load", False):
            view.run_command('detect_invalid_devangari')

    def on_post_save_async(self, view):
        if self.settings.get("run_on_save", False):
            view.run_command('detect_invalid_devangari')

class DetectInvalidDevangariCommand(sublime_plugin.TextCommand):
    def run(self, edit):
        valid_regions = [
            region
            for region in self.view.find_all(r'{Charset}+'.format(Charset=Charset))
            if re.match(reg, self.view.substr(region))
        ]
        if valid_regions:
            self.view.add_regions('valid', valid_regions, 'valid')
```

Thom1729:
Oh, and something I overlooked before -- you should have a license file.

alchemi5t:
License done, code edited(Thank you for editing the whole thing :D , sorry for the trouble), command and setting file added. Changed the package name too. Should i change/add/remove anything else?

Thom1729:
Only one thing. The description of the package seems to imply that it highlights invalid words. However, the current functionality is to highlight valid words. Personally, highlighting invalid words makes more sense to me, but if you think that it should highlight valid words instead, then it would be a good idea to rename things accordingly.

alchemi5t:
I had some issues using view.findall() to highlight invalid regions(highlighted some parts of some words which should not have been highlighted and vice-versa), but it worked just fine with the valid words, so left it as valid words. have renamed all of the "...invalid..." to detect or validator so there would be no confusion. changed readme accordingly. Is there is any specific file you want me to rename or edit?

Thom1729:
Have you tried that approach again after the recent revisions?

alchemi5t:
I had not tried it post the revisions. Just did and it works just fine now!! edited the code for highlighting the invalid words only.

alchemi5t:
I think the view.substr() did the trick!!

alchemi5t:
Thank you for helping me through this! T'was great working with you!! will finish a few more plugins.!

PR #7139, opened 2018-06-01
Added package Cacher

jookyboi:
[Cacher](https://www.cacher.io/) is the code snippet organizer for pro developers. It is a cross-platform, cloud-based app used to curate a snippet library.

This package for Sublime Text gives Cacher users the ability to perform popular actions on their personal and team snippet libraries.

Commands include:
- Insert Snippet
- Create Snippet
- Open Snippet in App
- Open Snippet Page
- Refresh Snippets

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -142,6 +142,17 @@
 			]
 		},
 		{
+			"name": "Cacher",
+			"details": "https://github.com/CacherApp/cacher-sublime",
+			"labels": ["snippets", "gists", "code library"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Caddyfile Syntax",
 			"details": "https://github.com/caddyserver/sublimetext",
 			"labels": ["language syntax", "syntax highlighting"],


Thom1729:
In `lib/util.py@load_config`:

- You're loading configuration data from a YAML file inside the package (`config.yml`). This will not work when the package is installed, because installed packages are zipped. Instead, you should put this data in a `Cacher.sublime-settings` file and load it with `sublime.load_settings('Cacher')`.

- You rely on the user's Cacher settings being located under `~/.cacher`. Is this reliable across every platform? Is this user-configurable? If that file isn't a normal part of Cacher, and you're only using it in this package, then you should manage this information with Sublime's settings API rather than creating files elsewhere.

- You're manually checking that the credentials file exists, then opening it. This leaves you open to race conditions. Instead of checking, simply try to open the file and catch the error. In addition, on an error you're returning `None`, which will cause `request_headers` to break. If you can't open the credentials file, you should probably raise an error and handle it up the stack with an error message.

- The capitalization of the headers is inconsistent.

In `lib/store.py`:

- What's the point to defining two methods to get and set from a dictionary? Why shouldn't client scripts simply use `store[key]` instead of `store.get_val(key)`? And should this whole thing also just use Sublime settings?

In `lib/snippets.py`:

- `description = snippet["description"] or ""` looks like it should be `description = snippet.get("description", "")`. In general, a lack of a value should be represented by a missing key rather than a falsy value.

In `lib/filetypes.py`:

- Why are you using `ntpath`? Isn't that going to cause platform compatibility problems?

In `cacher_setup.py`:

- `CacherSetupCommand` is a `TextCommand`, but you don't have any use for the current view or window. Should it be an `ApplicationCommand`?
- Why is it necessary to dedupe calls here? How does this happen?
- In your error handler, you're calling `json.loads` but not catching `JSONDecodeError`.

In `cacher_refresh.py`:

- Should this be an `ApplicationCommand` rather than a `WindowCommand`?

In `cacher_create_snippet.py`:

- You're using `ntpath` here again.


General:

- In several places, you're building up lists by initializing them to `[]`, iterating over another collection, and appending an item at each iteration. Generally, list comprehensions are a better way to initialize lists.
- Method names with double underscores are weird and look like builtins.
- Instead of this:
```python
group = None
if "group" in self.args:
    group = self.args["group"]
```

Do this:

```python
group = self.args.get("group", None)
```
- In several places, you're calling `webbrowser.open` with the same URL and some query parameters. This would probably be a good thing to factor out, and use `urllib.parse` to make sure that the URL is built correctly.

jookyboi:
@Thom1729 Thanks so much for the thorough code review! I'm coming from Javascript/Ruby land so it's taking me some time to ramp up on idiomatic Python. Your tips have definitely helped.

You can see all my latest changes at: https://github.com/CacherApp/cacher-sublime

Here are my responses:

In `lib/util.py@load_config`:

- YAML file: I've switched things to use `.sublime-settings` instead for the hostname constants.
- `.cacher` is used across several Cacher apps, including the CLI, VSCode and Atom extensions. It is meant to be a centralized place for settings and preferences. I'd prefer to keep it consistent for Sublime as well.
- I've changed the code to catch exceptions up the stack while making HTTP requests.
- Thanks for noticing the headers typo. That's been fixed.

In `lib/store.py`:

- Changed to use `.sublime-settings` as well.

In `lib/snippets.py`:

- Thanks for the tip. I've changed my code to use dictionary `.get()` with default values.

In `lib/filetypes.py`:

- The reason why I'm using `ntpath` is due to the top answer on SO: https://stackoverflow.com/questions/8384737/extract-file-name-from-path-no-matter-what-the-os-path-format. If this isn't correct, could you suggest another cross-platform way of fetching filenames from paths?

In `cacher_setup.py`:

- Changed `CacherSetupCommand` to be `ApplicationCommand`.
- Removed the de-dupe calls. There was some bug I was working around but it's not happening now.
- My version of Python doesn't have `JSONDecodeError` so I'm handling `ValueError` instead when the JSON decoding fails.

In `cacher_refresh.py`:

- Changed to be a `ApplicationCommand`.

In `cacher_create_snippet.py`:

- Same comment as above about cross-platform filenames.

General:

- I've switched most of my iteration blocks to use list comprehensions.
- Removed the `__` from method names.
- Switched to using dictionary `.get()` with default values.
- Added a `util` method for checking URLs before calling `webbrowser.open(...)`

Thom1729:
> Changed [lib/store.py] to use .sublime-settings as well.

I'd call it `Cacher Store.sublime-settings` rather than simply `Store.sublime-settings`, because the namespace is shared.

> The reason why I'm using ntpath is due to the top answer on SO: https://stackoverflow.com/questions/8384737/extract-file-name-from-path-no-matter-what-the-os-path-format. If this isn't correct, could you suggest another cross-platform way of fetching filenames from paths?

That makes sense if you expect that you may receive Windows paths even when running in a Linux environment. Is that the case?

> Added a util method for checking URLs before calling webbrowser.open(...)

Myself, I'd even go a step further:

```python
def open_apphost_url(path, **kwargs):
    result = settings().get("appHost") + path + '?' + urllib.parse.urlencode(kwargs)
    # Validation, etc.
    webbrowser.open(result)
```

jookyboi:
@Thom1729 I've made all the changes you requested in your last comment. I am running into one issue though. The following piece of code should run when the plugin is loaded:
```
def plugin_loaded():
    if not util.credentials_exist():
        util.prompt_user_setup()

# util.py
def prompt_user_setup():
    if sublime.ok_cancel_dialog("Cacher needs to be setup before use", "Start Setup"):
        sublime.run_command("cacher_setup")
```

The user does get the prompt asking them to start setup. But `sublime.run_command("cacher_setup")` doesn't seem to execute. I tried calling this outside of the `plugin_loaded()` function and it does trigger the command. Any ideas?

jookyboi:
Hi @Thom1729. Have you had a chance to look over the latest code?

Thom1729:
Are you still having the double-execution issue?

jookyboi:
@Thom1729 Yes. I'm using `sublime.set_timeout_async(util.prompt_user_setup)` to launch the setup command now but it still has the double-execution issue. Unless you have another suggestion, I may just put the de-duping logic back in.

jookyboi:
@Thom1729 I've updated master with the latest code (with the de-dupe workaround). Please let me know if there's anything else.

PR #7140, opened 2018-06-02
Add VintageColor package

acarabott:
This package allows you to set different color schemes for the different Vintage modes: `Insert`, `Command` and `Visual`/`Visual Line`, making it easier to determine the current mode, and reducing errors.

### DIFF
--- a/repository/v.json
+++ b/repository/v.json
@@ -414,7 +414,7 @@
 		},
 		{
 			"name": "VintageColor",
-			"details": "https://github.com/acarabott/package_control_channel",
+			"details": "https://github.com/acarabott/VintageColor",
 			"releases": [
 				{
 					"sublime_text": ">=3000",
--- a/repository/v.json
+++ b/repository/v.json
@@ -413,6 +413,16 @@
 			]
 		},
 		{
+			"name": "VintageColor",
+			"details": "https://github.com/acarabott/package_control_channel",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "VintageES",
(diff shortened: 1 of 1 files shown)


Thom1729:
It looks like you've gone through a couple of iterations of the mode logic. How about something like the following:

```python
mode_commands = {
    'enter_insert_mode': 'insert',
    'exit_insert_mode': 'command',
    'enter_visual_mode': 'visual',
    'enter_visual_line_mode': 'visual_line',
    'exit_visual_mode': : 'command',
}

class VintageColor(sublime_plugin.EventListener):
    def on_post_text_command(self, view, command_name, args):
        if command_name not in mode_commands:
            return

        mode = mode_commands[command_name]
        theme = plugin_settings.get('{}_theme'.format(mode)):

        if theme:
            view.settings().set('color_scheme', theme)
        else:
            view.settings().erase('color_scheme')
```

If the view's `color_scheme` setting is not set, then it will default to the global setting.

acarabott:
Thanks Thom.

I've pushed an update (https://github.com/acarabott/VintageColor/commit/3e6336a371a516c4c18963a14223028975a7c1d9) 
Your approach definitely cleans up the logic, and I hadn't realised you could just erase the current colour scheme.

I had to make an additional change to the insert mode logic, as it is possible to enter insert mode *without* getting an `enter_insert_mode` command, e.g. by pressing the `c` key with a selection.

jrappen:
@acarabott you can use [PackageDev](https://github.com/sublimetext/PackageDev) to convert the color scheme(s) to the **new** [color scheme format for ST3149+](https://www.sublimetext.com/docs/3/color_schemes.html)

PR #7142, opened 2018-06-03
show variable type in statusbar

mohammad6006:


Thom1729:
@mohammad6006 Please address the error.

FichteFoll:
Next time, please only add a single package per pull request.

You're using mixed indentation in your RTL Python file, btw.

PR #7143, opened 2018-06-04
Added UDF Format

SpyceTewan:
I added my simple file format syntax highlighter to the package control repository. 
I passed all of the tests so everything should work fine

### DIFF
--- a/repository/u.json
+++ b/repository/u.json
@@ -46,6 +46,16 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


Thom1729:
There should basically never be a `sublime-package` file inside a package. Sublime will automatically zip the contents of your package; you shouldn't do it yourself (and it won't work if you do).

Thom1729:
@SpyceTewan ping.

PR #7147, opened 2018-06-08
Add Text Marker package

Kronuz:
Text Marker (Highlighter) highlights words in Sublime Text

Although there are similar packages (Word​Highlight comes to mind) this one is much more flexible and improves a lot of things. I've used it for years and now I'm releasing it, hoping someone else can find ut useful too.


Thom1729:
The build failed due to alphabetization.

Kronuz:
@Thom1729, fixed alphabetization issue.

Thom1729:
It looks like there's a lot of code in common between this and the Color Highlight package. Specifically, it seems that they're both implementing a general system for adding rules to color schemes. Would it make sense to factor this out into a dependency?

Kronuz:
Yes, maybe... but I don’t have a clear idea how those dependencies work. I maintain common code separated in modules like such (colorize and settings are exactly the same) ...and settings is also used in SublimeCodeIntel. Maybe we can add those as dependencies later

Thom1729:
- Instead of your `regex_escape` function, would `re.escape` work?
- Several of the `TextCommand`s do not refer to the current window or view. In addition, the `block` parameter is frequently unused.
- As with the other package, it would be better to have an explicit LICENSE file with the text of the license rather than merely referring to the name of the license in the README.

Kronuz:
@Thom1729, fixed issues you suggested:

* I added comments about the need of `regex_escape()` and updated to use the same code as `WordHighlight` for it.
* Removed unused block argument
* Added LICENSE

Regarding `TextCommand`s not using the window or view (text_marker_reset, text_marker_restore and text_marker_clear), how should I handle those then? as those are called from menus, command palette and keybindings.

Thom1729:
> Regarding TextCommands not using the window or view (text_marker_reset, text_marker_restore and text_marker_clear), how should I handle those then? as those are called from menus, command palette and keybindings.

I would use `ApplicationCommand`.

Kronuz:
@Thom1729, I've changed it so it now uses `ApplicationCommand`, you think we can publish the plugin now?

Kronuz:
Thank you @Thom1729 👍

FichteFoll:
This is actually not a bad idea. Good job, thinking about this enhancement for WordHighlight.

Kronuz:
@FichteFoll, yes, I use this package all the time, and recently (from my trip to VS Code world and back) I helped this other guy improving his extension (also by the name Text Marker) and then I copied some usability bits I liked from there as well, making mine even better than it was. Only thing is I always have live word highlight disabled (kind of annoys me) but it’s on by default so people know it exists.

PR #7148, opened 2018-06-08
Add Color Highlight package

Kronuz:
Color Highlight colorizes Sublime Text

https://github.com/Kronuz/ColorHighlight

There is a similar packages (ColorHighlighter) this one is much more lightweight and has small features that are different from the other one, which I feel bloated. I've also used it for years and now I'm releasing it, hoping someone else can find it useful too.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2329,6 +2329,17 @@
 			]
 		},
 		{
+			"name": "Color Highlight",
+			"labels": ["color", "highlight", "highlighter", "hex", "rgb", "hsl"],
+			"details": "https://github.com/Kronuz/ColorHighlight",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Color Highlighter",
 			"details": "https://github.com/Monnoroch/ColorHighlighter",
 			"releases": [


Thom1729:
- Rather than having one `highlight` setting with many values, would it make more sense to have several settings, e.g. `highlight_on_load`, `highlight_on_save`, `highlight_on_edit`, etc?
- Similarly, are the options for format highlighting comprehensive?
- Should the delays be user-configurable?
- You're building the highlighting XML via string formatting. I would use `plistlib` instead to avoid any conceivably escaping issues.
- Does this work with `sublime-color-scheme`s?
- In a few places, you're opening and then closing files. Instead, you should use the `with` statement to ensure that file handles are cleaned up:

```python
# Before
f = open(rf, 'w')
f.write(s)
f.close()

# After
with open(rf, 'w') as f:
    f.write(s)
```

- Conditioning on a path existing (or not) introduces a race condition. The preferred idiom is to try the operation and catch the exception:

```python
# Before
if not os.path.exists(dn):
    os.makedirs(dn)

# After (probably make a function for this)
try:
    os.makedirs(dn)
except OSError as e:
    if e.errno != errno.EEXIST:
        raise
```

- `ColorHighlightCommand` is a `WindowCommand`, but it operates on a view. Should it be a `TextCommand`?

Kronuz:
@Thom1729, After your observations and to answer some of your inquiries:

- I'm now using `plistlib`, I do believe it looks cleaner now too.
- The package does work with `sublime-color-scheme`.
- I'm now using `with` statement where appropriate.
- Fixed the potential race condition during `makedirs()`
- I'm not sure about WindowCommand, I can't remember why it was this way.
- About the settings, I'll tweak those in time.

Thom1729:
You have license information in the README, but it would be better to have an explicit LICENSE file with the text of the license.

Kronuz:
@Thom1729 I added the license now

PR #7149, opened 2018-06-09
add format-apex-log package

jdrbc:
Salesforce Apex logs as they are generated from the Salesforce service are difficult to read without additional formatting. In particular, the almost-JSON format of serialized objects is hard to read without formatting. Before making this plugin I would often paste logs into jsonformatter dot curiousconcept dot com to make logs more readable. This plugin uses tags reliably found in Apex logs to generate a much more readable file that any non-Salesforce specific plugin could. 

I work with a few dozen Salesforce developers in Vancouver, and we all have run into this problem. It's not a matter of configuring Salesforce correctly to pretty-print logs. Thanks for your time!

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -1439,6 +1439,17 @@
 			]
 		},
 		{
+			"name": "Format Apex Log",
+			"details": "https://github.com/jdrbc/format-apex-log",
+			"labels": ["apex", "salesforce", "formatting"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Format JSDoc @params",
 			"details": "https://bitbucket.org/finitewisdom/sublime-jsdocparam",
 			"labels": ["jsdoc", "javascript", "formatting"],


Thom1729:
You should have a license file. Other than that, it looks fine. Is there a syntax definition for these files?

jdrbc:
Added licence. This is as close to a formal definition I could find for Salesforce log files. https://developer.salesforce.com/docs/atlas.en-us.apexcode.meta/apexcode/apex_debugging_debug_log.htm

Thom1729:
When you open these files in Sublime, is there any syntax highlighting? I ask because ideally, the key bindings should only apply when looking at an Apex log. If there's highlighting, you can use that to restrict the bindings; if there's no highlighting, then creating a Sublime syntax definition would probably be helpful to you (though it's certainly not necessary).

jdrbc:
There is no highlighting when I open apex logs in sublime. I'll look into syntax definitions. Thanks for the suggestion!

jdrbc:
I did some reading on syntax definitions. If I add more highlighting to the log file I'll use a syntax file. It also looks like I could use syntax scopes to find regions for formatting? Limiting the keybinding to run only on .log files might be inconvenient because sometimes log contents are pasted into empty files.

PR #7150, opened 2018-06-10
Add package: SasSubmit

sjiangDA:
This plugin sends SAS command from Sublime to SAS programs, including SAS. SAS studio and SAS studio university edition.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -238,6 +238,17 @@
 			]
 		},
 		{
+			"name": "SasSubmit",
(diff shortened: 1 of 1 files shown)


sjiangDA:
* "F3" was used because SAS programmers use F3 to submit commands.
* ".no-sublime-package" is indeed needed. For this plugin to run properly, python package "selenium" is required but cannot be installed using Sublime. The "command_sender/dist/command_sender.exe" was built from "command_sender/command_sender.py" using "pyinstaller". Besides, this package also uses chromedriver.

PR #7151, opened 2018-06-10
Add new package "Git Blame Status Bar"

rodrigobdz:
Shows Git blame author for the currently selected line.

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -433,7 +433,7 @@
 			"name": "Git Blame Status Bar",
 			"author": "rodrigobdz",
 			"details": "https://github.com/rodrigobdz/subl-gitblame-statusbar",
-			"labels": ["git", "blame", "statusbar", "simple"],
+			"labels": ["git", "blame", "statusbar"],
 			"releases": [
 				{
 					"sublime_text": "*",
--- a/repository/g.json
+++ b/repository/g.json
@@ -430,6 +430,18 @@
 			]
 		},
 		{
+			"name": "Git Blame Status Bar",
+			"author": "rodrigobdz",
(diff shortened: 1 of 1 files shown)


Thom1729:
I see that this is partly derived from the [Git blame](https://github.com/psykzz/st3-gitblame) package. Would it make sense to add this feature to that package rather than creating a new package?

rodrigobdz:
@Thom1729 I thought about it but decided to publish a new package instead to offer a minimalistic solution to a common problem.
[Git Blame](https://github.com/psykzz/st3-gitblame) sure offers more functionality but IMO it has a different goal than [Git Blame Status Bar](https://github.com/rodrigobdz/subl-gitblame-statusbar).

rodrigobdz:
[Git Blame Status Bar](https://github.com/rodrigobdz/subl-gitblame-statusbar) has an approach comparable to the one of the VSCode Extension [Active File In Status Bar](https://github.com/RoscoP/ActiveFileInStatusBar).
Minimal solution to a simple problem.

Thom1729:
- You should explicitly handle the case where the file is not in a git repo.
- Generally, ignoring errors is not ideal.
- You should throttle multiple activations to avoid performance problems.

What I would do myself is separate the code into a `TextCommand` that does all of the real work and a `ViewEventListener` that invokes that command. I'd also add some sort of timeout to prevent the listener from activating the command too frequently.

Thom1729:
@rodrigobdz ping.

rodrigobdz:
@Thom1729 I haven't forgotten this open PR. I haven't got around to it yet.

PR #7152, opened 2018-06-11
move to SublimeText repo

randy3k:
two things.

1. specified UnitTesint information here rather than https://github.com/randy3k/sublime_packages/blob/master/packages.json#L190-L204

2. moved `sublime-pyzmq` and `sublime-pywin32` to ~SublimeText~ packagecontrol as discussed below

### DIFF
--- a/repository/dependencies.json
+++ b/repository/dependencies.json
@@ -667,10 +667,10 @@
 			"author": "randy3k",
 			"load_order": "50",
 			"description": "Pywin32 module",
-			"issues": "https://github.com/SublimeText/sublime-pywin32/issues",
+			"issues": "https://github.com/packagecontrol/sublime-pywin32/issues",
 			"releases": [
 				{
-					"base": "https://github.com/SublimeText/sublime-pywin32",
+					"base": "https://github.com/packagecontrol/sublime-pywin32",
 					"platforms": ["windows"],
 					"tags": true,
 					"sublime_text": ">=3000"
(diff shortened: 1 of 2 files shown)


randy3k:
once it is merged, I will remove the reference in https://github.com/randy3k/sublime_packages/blob/master/packages.json#L190-L204

FichteFoll:
You could also move your dependencies to https://github.com/packagecontrol/ if they don't really need maintenance. I can add you to that. That's certainly not a requirement, just an idea to group PC-specific dependencies in a separate repo.

randy3k:
@FichteFoll I moved [pyzmq](https://github.com/SublimeText/sublime-pyzmq) a
and [pywin32](https://github.com/SublimeText/sublime-pyzmq) to SublimText. Could you move them to packagecontrol too? (I have lost the admin rights of them)

FichteFoll:
Right, I forgot to add them to you when I received the "new repo" notification (I have to grant the rights manually). Will do.

FichteFoll:
Okay, deps are moved and you have admin access on them again.

However, you cannot switch from explicit versioning to branch-based. You must use tag prefixes for your st2 release channel.

randy3k:
Do I need to tag a different prefix for st3? How PC determine the latest release?

FichteFoll:
>Do I need to tag a different prefix for st3? 

No, if you set it to `true`, PC will look for no prefix or a `v` prefix, which is then stripped automatically.

PR #7155, opened 2018-06-15
Added ServiceNow Sync package

salcosta:

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1137,6 +1137,18 @@
 			]
 		},
 		{
+			"name": "ServiceNow Sync",
+			"details": "https://github.com/salcosta/servicenow-sync",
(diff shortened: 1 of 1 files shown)


Thom1729:
- You shouldn't modify `sys.path`, because that is shared between all packages.
- You have an empty `__init__.py`.
- Your readme doesn't describe what the package does or who would use it.

salcosta:
@Thom1729 Appending the current path to the `sys.path` seems to be the standard across other packages.  Is there an alternative you are wanting to use now?

Regarding the other 2 changes I will resolve those.

Thom1729:
Instead of modifying `sys.path`, use relative imports: `from .Commands.create_connection import *`.

Thom1729:
@salcosta ping.

PR #7156, opened 2018-06-15
Add Vuejs Snippets Airbnb

kavalcante:

### DIFF
--- a/repository/v.json
+++ b/repository/v.json
@@ -728,7 +728,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"branch": "master"
+					"tags": true
 				}
 			]
 		},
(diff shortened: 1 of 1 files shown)


Thom1729:
I see that this is a fork of an existing package. Why is this necessary? Would it make sense to integrate any improvements into the existing package instead?

kavalcante:
@Thom1729 The package is a bunch of snippets for Vue.js. The original package uses Standard Styleguide. My fork, uses Airbnb Styleguide. To integrate properly would be necessary create a "config system" to know what styleguide's to use. In my opinion is too much for a simple package of snippets. Because that, I submit a new package instead making that kind of improvement in original folder.

kavalcante:
@Thom1729 done!

PR #7162, opened 2018-06-24
Add package: FosterNeko

shellbotx:
This is a color scheme, based off of "neko" which was abandoned. Currently only syntax highlighting has been done, but UI will be added in the future.

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -1502,6 +1502,16 @@
 			]
 		},
 		{
+			"name": "FosterNeko",
+			"details": "https://github.com/shellbotx/foster-neko",
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


Thom1729:
>   - ERROR: The package does not define any file that interfaces with Sublime Text

This error can be ignored. The bot doesn't yet know about `sublime-color-scheme` files. However, these schemes only work on build 3149 and above, so you should update the `releases` key accordingly.

You should link to the original. (Is it “neko” or “neka”?)

shellbotx:
@Thom1729 Thanks! I'll update the `releases` key. And it's "neko" based off of "neka", sometimes I get them mixed up a bit. I'll add the original to the README

PR #7163, opened 2018-06-24
add Mattermost Post

rubberneck:
This package will post selected text to a [Mattermost Server](https://mattermost.com/). If you set post_fileinfo to true it will also post the filename and the line numbers of the selected text.

It will make 2 network requests to the url in the settings every time it is run. One to get the channel id and the other to post the message.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -871,6 +871,16 @@
 			]
 		},
 		{
+			"name": "Mattermost Post",
+			"details": "https://github.com/rubberneck/sublime-mattermost-post",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Maude Syntax Highlighting",
 			"details": "https://github.com/jpsikorra/maude_syntax_highlight",
 			"labels": ["maude", "language syntax"],


rubberneck:
@packagecontrol-bot

Thom1729:
```
conn.request("GET", "/api/v4/teams/name/" + team + "/channels/name/" + channel, "", headers)
```

Would it make sense here to do some encoding? Alternatively, you might check to make sure that `team` and `channel` are sane.

Rather than manually calling `conn.close()`, is it possible to use `with` to avoid leaking connections?

Overall, you might want to use `format` instead of so much concatenation.

rubberneck:
I am now checking `url`, `team`, and `channel` to make sure they only contain lowercase letters, numbers, dots and dashes. Nothing that needs to be encoded. I just used re.match for this. I am all for a better idea.
```python
if re.match("^[a-z0-9-.]*$", url) is None:
    sublime.error_message("Mattermost Post: Setting url may contain only lowercase letters, numbers, dots and dashes.")
    return
```
https://github.com/rubberneck/sublime-mattermost-post/commit/9696a7c5924a8e3144905521f115bcdcf0a0e72c


For the `with` I had to use closing from the contextlib as `HTTPSConnection` doesn't have a `__exit__` method. Is this the correct way to do this? This does seem like a much better idea as I won't miss a .close() on the connection anywhere which I think I was when looking at the code again. 
```python
from contextlib import closing

with closing(http.client.HTTPSConnection(url)) as conn:
```
https://github.com/rubberneck/sublime-mattermost-post/commit/67810547fe1614ea452f9fcd133782f4fa1a3dfa


I changed to using .format instead of string concatenation on the single line stuff.
https://github.com/rubberneck/sublime-mattermost-post/commit/652f5b5798033076ce5b7c900ab0c78a34cf4b73

Thanks for the suggestions @Thom1729 If you have any more please let me know.

Thom1729:
You might want to take a look at the [requests](https://github.com/packagecontrol/requests) dependency.

PR #7164, opened 2018-06-25
add new package 'AddToFile'

nchauhan890:
This package allows user to quickly copy their selection and have it added to the end of a chosen open file (or they can select to add to a entirely new file). Eases copying text between files by using a simple key binding and not switching the view focus to paste text.
[Demonstration picture](https://github.com/nchauhan890/AddToFile/blob/master/assets/demonstration-1.png)

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -322,6 +322,17 @@
 			]
 		},
 		{
+			"name": "AddToFile",
+			"details": "https://github.com/nchauhan890/AddToFile",
+			"labels": ["text manipulation"],
+			"releases": [
+				{
+					"sublime_text": ">=3114",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Adianti Snippets for PHP",
 			"details": "https://github.com/dmonteirosouza/adianti-snippets",
 			"releases": [


nchauhan890:
Thanks!   
  
  

  
I'm not sure about how this works completely yet, but could you say when this package will be available to install via Package Control inside Sublime Text? I'm very new and I don't have any idea of what happens now that this pull request has been approved.   
  
  

  
  

  
  
>

Thom1729:
The approval means that the package has passes a series of automated checks. The package must still be manually reviewed, probably by me. It is likely (but not guaranteed) that I will be able to review it within the next few days.

Thom1729:
```python
return [view for view in sublime.active_window().views()
                # if view.file_name()
                if view != sublime.active_window().active_view()]
```
Should this be `return [view for view in self.view.window().views() if view != self.view]`? (This may apply generally wherever you use `sublime.active_window()` in a `TextCommand`.)

nchauhan890:
Would there be any difference?

Thom1729:
Yes. The command could be run on a view that is not the active view, or on the active view in another window.

Plus, `self.view` is just simpler than `sublime.active_window().active_view()`.

nchauhan890:
Fair enough, I'll try and update it as soon as I can get back on

nchauhan890:
Sorry for the wait, changes implemented, but "conflicted" with a new commit I just made. Still works as intended with a new addition. Documentation update in progress

Thom1729:
You still have a bunch of unnecessary `sublime.active_window()` calls that should be replaced by `self.view.window()`.

nchauhan890:
Whoops! Looks like I missed a few. Updated now with a bit of cleanup and proper <80 line lengths

PR #7166, opened 2018-06-29
Added PaletteFile plugin

vshotarov:
The PaletteFile plugin creates an easier way for navigating the project hierarchy when in need of creating a new directory or file from the command palette.

I find it necessary, considering I am working on a very deep project hierarchy on a daily basis. Opening files is really easy with **Ctrl + P**, so I wanted to have the same for creating new files/folders as well.

A quick search on package control didn't produce any results with similar functionality.  The closest I have found is [AdvancedNewFile](https://github.com/skuroda/Sublime-AdvancedNewFile), as it alleviates some of the problems, but I believe PaletteFile is sufficiently different considering it is giving the ability to navigate the folder structure instead of relying on it being memorized.

I have successfully ran the tests from the ChannelRepositoryTools.

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -147,6 +147,17 @@
 			]
 		},
 		{
+			"name": "PaletteFile",
+			"details": "https://github.com/vshotarov/PaletteFile",
+			"labels": ["file creation", "file navigation", "directory creation", "directory navigation"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Panda Syntax Sublime",
 			"details": "https://github.com/siamak/panda-syntax-sublime",
 			"labels": ["color scheme"],


Thom1729:
When using the filesystem in Python, it's generally better to try an operation that could fail and catch errors than to check beforehand. This avoids race conditions that could otherwise occur. For instance, when running `_create_path` to create a file, you should simply `open(full_path, "x")`, catching `FileExistsError`, `PermissionError`, and finally `OSError`. Then there is no need to explicitly check `os.path.exists` or `os.access`. Something like this:

```python
def _create_path(name, pathType, chosen_directory, window):
    '''Creates a file or folder {pathType} with the specified {name} in the
    {chosen_directory}.
    Args:
        name: The last portion of the path - the name of the file or folder
        pathType: The type of path being created - file or directory
        chosen_directory: The parent directory of the file or folders
        window: The sublime Window object from the WindowCommand
    '''
    full_path = os.path.join(chosen_directory, name)

    try:
        if pathType == "file":
            with open(full_path, "x"):
                pass
            window.open_file(full_path)
        else:
            os.makedirs(full_path, exist_ok=False)

        window.status_message("Created " + full_path)
    except FileExistsError:
        sublime.error_message("%s %s already exists." % (
            pathType.capitalize(), full_path.capitalize()))
    except PermissionError:
        sublime.error_message("Cannot write to " +
                              full_path + ". Check access permissions.")
    except OSError:
        sublime.error_message("Cannot write to " + full_path + "." +
                              "The path may be invalid.")
```

vshotarov:
Thanks, that's really helpful! 

It makes way more sense to catch an error and also is getting rid of that horrible *failed* variable I was using previously.

I have rolled out an update - [ Changed the implementation of _create_path to use a try and a sequence of except statements instead of doing preliminary checks for whether the file exists, the path is valid or the current user has permissions.](https://github.com/vshotarov/PaletteFile/commit/f123eec975ee03d11d9bb317a4e44a0977b10a12)

PR #7168, opened 2018-07-01
submission for software.com sublime text package to allow users to view kpm and session time data on the status bar and in their dashboard

xavluiz:

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -2501,6 +2501,17 @@
 			]
 		},
 		{
+			"name": "Software.com",
+			"details": "https://github.com/swdotcom/swdc-sublime",
+			"labels": ["software", "insights", "code", "work life balance"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Solarized Color Scheme",
 			"details": "https://github.com/braver/Solarized",
 			"labels": ["color scheme"],
--- a/repository/s.json
+++ b/repository/s.json
@@ -2501,9 +2501,9 @@
 			]
 		},
 		{
-			"name": "Software.com",
+			"name": "Software",
 			"details": "https://github.com/swdotcom/swdc-sublime",
-			"labels": ["software", "software.com", "code time", "work life balance", "productivity"],
+			"labels": ["utilities", "status bar", "javascript", "google", "productivity"],
 			"releases": [
 				{
 					"sublime_text": "*",
--- a/repository/s.json
+++ b/repository/s.json
@@ -2503,7 +2503,7 @@
 		{
 			"name": "Software.com",
 			"details": "https://github.com/swdotcom/swdc-sublime",
-			"labels": ["software", "insights", "code", "work life balance"],
+			"labels": ["software", "software.com", "code time", "work life balance", "productivity"],
 			"releases": [
 				{
 					"sublime_text": "*",


Thom1729:
The license for this would seem to prohibit Package Control from redistributing it.

Other issues at first glance:

- The package seems to have its own internal update system. Why is this?
- There doesn't seem to be any way of temporarily disabling telemetry without removing the package.
- The `EventListener` implements the synchronous methods, rather than the async ones.
- You're manually checking the OS when you could be using generic functions from `os.path`.
- The script is nearly a thousand lines of code. It would probably be a good idea to break it up.

xavluiz:
Thank you so much for the review. We've updated all of the requests and cleaned up the code much more.

- removed unused modules, plugin manager download and update logic (internal update system), and constants
- broke up the 1 file into their specific manager modules
- using available async event listener methods
- added "Tools" menu items to pause and enable kpm updates

Thom1729:
Those changes are very helpful; thanks.

SoftwareUtil.py:

- Instead of checking the OS, you should use the generic functions from `os.path`, e.g. `softwareDataDir = os.path.join(getHomeDir(), '.software')`.
- In particular, you should use `os.path.expanduser` to get the user's home directory instead of relying on an environment variable.
- `isMac` seems to be unused.
- Rather then checking if softwareDataDir exists, you should eagerly create it using `exist_ok=False` and catch `FileExistsError`. This avoids race conditions.
- `getSoftwareSessionAsJson` leaks a file handle. You should use `with`. The whole thing should look something like this:

```python
def getSoftwareSessionAsJson():
    try:
        with open(getSoftwareSessionFile()) as sessionFile:
            return json.loads(sessionFile.read())
    except FileNotFoundError:
        return dict()
```

SoftwareHttp.py:

- `LOGGING` and `TELEMETRY` should use the Settings API, not global variables.
- The API endpoint should probably also be a setting rather than hardcoded and cased out.

SoftwareSession.py:

- The URLs could easily be settings.
- `fetchingUserFromToken` seems to be unused.
- In `sendOfflineData`, you should handle `FileNotFound` rather than checking.

Software.py:

- The commands shouldn't take unused `**kwargs`. You want them to error if invalid arguments are passed.
- As discussed before, `Pause`/`EnableKpmUpdatesCommand` should use settings.

FichteFoll:
Also note that your package name must not contain periods as those will cause Python to fail importing any plugins within that. See #2524 for more information.

Another suggestion regarding the snippet that @Thom1729 mentions:

```py
def getSoftwareSessionAsJson():
    try:
        with open(getSoftwareSessionFile()) as sessionFile:
            return json.load(sessionFile)
    except FileNotFoundError:
        return dict()  # or just {}
```

(I did not review the package itself.)

xavluiz:
Thank you again for the insightful review. We're very impressed with the level of detail and knowledge about what to look for.

I've updated everything you've requested. We now have settings for the url, api, telemetry, and logging. Updated the file handling based on your suggestions.  Thank you again.

Thom1729:
The period in the package name should be fixed. Everything else is minor.

Instead of:

```python
if (isWindows()):
    file += "\\session.json"
else:
    file += "/session.json"
```

Use `os.path`:

```python
file = os.path.join(file, 'session.json')
```

You can apply this change to all three functions in `SoftwareUtil` and remove `isWindows`.

Instead of:

```python
if (sublime_settings.get("software_logging_on", True) is False):
    return
print(message)
```

Why not:

```python
if (sublime_settings.get("software_logging_on", True)):
    print(message)
```

You should print error messages even if logging is disabled. Otherwise, operations will fail silently. Example:

```python
    except (http.client.HTTPException, http.client.CannotSendHeader, ConnectionError, Exception) as ex:
        log("Software.com: " + api + " Network error: %s" % ex)
        return None
```

It's also perfectly fine not to catch an unexpected exception and let Sublime print it to the console. If you can identify it as a network error, that's one thing, but catching `Exception` might be a bit overzealous.

xavluiz:
Thank you so much for the latest review in trimming down how we build the filenames and a cleaner way to check for boolean.

We've pushed those changes and the package control s.json file without a dot in the name.  We appreciate the feedback.

PR #7171, opened 2018-07-03
Add regex escape package

Jonnymcc:
Adds a command that will replace selected text with regex escaped text.


Thom1729:
You should add a license.

PR #7173, opened 2018-07-03
Added High-Sienna theme to the channel

davidbwaters:

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -512,6 +512,16 @@
 			]
 		},
 		{
+			"name": "High-Sienna",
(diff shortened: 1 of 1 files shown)


jrappen:
@davidbwaters you can use [PackageDev](https://github.com/sublimetext/PackageDev) to convert the color scheme to the [new color scheme format](https://www.sublimetext.com/docs/3/color_schemes.html)

davidbwaters:
OK, I'll work on it.  I just modified a current theme and didn't realize I wasn't on the new format.

PR #7180, opened 2018-07-10
Add new package "Mellanox Syntax Highlighter"

KeyboardInterrupt:
This Package adds Syntax highlighting for Mellanox router/switch/firewall configuration files.

Also Thank you very much for package control, I could not imagine Sublime Text without it!

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -1057,6 +1057,16 @@
 			]
 		},
 		{
+			"name": "Melanox Syntax Highlighter",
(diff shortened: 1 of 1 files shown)


KeyboardInterrupt:
I added a Tag now, Missed that step somehow.

Thom1729:
You should add a license.

KeyboardInterrupt:
I added a LICENSE.md

Thom1729:
The channel entry appears to be sorted incorrectly.

PR #7181, opened 2018-07-10
Added BreakLines package details

Moojuiceman:

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -1272,6 +1272,18 @@
 			]
 		},
 		{
+			"name": "BreakLines",
+			"details": "https://github.com/Moojuiceman/BreakLines",
+			"issues": "https://github.com/Moojuiceman/BreakLines/issues",
+			"labels": ["break", "group", "split", "separate", "lines"],
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


Thom1729:
- You should add a license.
- It would be better to use a `TextCommand` for `PromptBreakLinesCommand` so that you can use `self.view` instead of `self.window.active_view()`. You can access the current window via `self.view.window()`.
- In `BreakLinesCommand`, is it necessary to handle invalid `break_gap`? Nothing should be passing in invalid values, and if something does, it's probably better to let the command fail and an error appear in the console.
- In `break_lines`, you can probably simplify the logic by iterating over `range(first_line, last_line + 1, break_gap)[1:]` in reverse.

Moojuiceman:
@Thom1729 
- Added MIT License
- Updated as recommended
- The menu/hotkey commands were passing the variable as a string. These have been changed and try/catch removed
- Updated as recommended, but with first_line + break_gap otherwise an extra break is added before the first line

If you have any other recommendations let me know, otherwise I'll re-tag so the changes are included in the package control build

PR #7182, opened 2018-07-11
Added AccuTermClient

ianharper:
AccuTermClient allows Sublime to connect to the AccuTerm terminal emulator which is used to connect to MultiValue databases (a legacy NoSQL database).

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -113,6 +113,17 @@
 			]
 		},
 		{
+			"name": "AccuTermClient",
+			"details": "https://github.com/ianharper/AccuTermClient",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true,
+					"platforms": ["windows"]
+				}
+			]
+		},
+		{
 			"name": "AceJump",
 			"details": "https://github.com/ice9js/ace-jump-sublime",
 			"labels": ["code navigation", "text navigation"],


Thom1729:
- You should add a license.
- Rather than requiring the PyWin32 *package*, you should require the newer [python-pywin32 *dependency*](https://github.com/packagecontrol/sublime-pywin32). Package Control will automatically ensure that dependencies are satisfied.
- Instead of using `sublime.active_window()`, you should pass a window into the appropriate functions. That way, the correct window will be used even if the user has switched windows.

ianharper:
@Thom1729 thanks for the input! I've added the MIT license, pywin32 dependency, and removed calls to active_window()/view().

Thom1729:
`AccuTermCompileCommand` looks like it should be a `TextCommand`, being view-specific.

Does `ConvDataInputHandler.__init__` do anything?

I'll look over it in more detail later tonight.

ianharper:
`AccuTermCompileCommand` is used in the build systems which require a window command.

`ConvDataInputHandler.__init__` sets the view which is used to extract selected text. I patterned this after arithmetic.py in the default package.

ianharper:
@Thom1729 Initially I had made the compile command a text command but using it as the target for a custom build system requires a window command so I changed it.

PR #7184, opened 2018-07-11
Add PureScript Syntax

bosyi:


Thom1729:
You should avoid regexp lookbehinds (e.g. in `numbers`), because the cause Sublime to use the old, slower regexp engine.

PR #7187, opened 2018-07-14
Sublime package of D-Laravel, runing composer and php artisan command in the fpm container. 

DevinY:

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -2,6 +2,17 @@
 	"schema_version": "3.0.0",
 	"packages": [
 		{
+			"name": "D-Laravel alias",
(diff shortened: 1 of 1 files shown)


Thom1729:
- You should add a license.
- Why have unimplemented commands?
- You generally shouldn't use `sublime.active_window()` in a command where you have access to the command's window.
- What is the point of `DlaravelCommand`?

DevinY:
Hi,
I am beginner of python, I hope everything is correct now. 

License is added.
unimplemented command is removed.
no sublime.active_window().
DlaravelCommand is also removed.

Thanks,
Devin.

PR #7189, opened 2018-07-15
Updated RTagsComplete with new repo and name.

tillt:
Sublime Text 3 C/C++ code completion, navigation plugin. It is based on RTags.

Fork and slight rename of the original "sublime-rtags", "RtagsComplete". The original author  abandoned the project for good and I am offering to take it over; see an exchange with that author and me at  https://github.com/rampage644/sublime-rtags/issues/39.

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -1792,8 +1792,11 @@
 			]
 		},
 		{
-			"name": "RtagsComplete",
-			"details": "https://github.com/rampage644/sublime-rtags",
+			"name": "RTagsComplete",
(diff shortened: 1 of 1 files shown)


Thom1729:
Because this package supports all platforms, it isn't necessary to list them.

Given that the conversation on moving the repo happened most of a year ago, it would be best to give @rampage644 a couple of days to reply if he chooses. Other than that, there's no reason we can't move it here.

rampage644:
@tillt :+1:

PR #7190, opened 2018-07-15
Add syntax highlighter for the eno notation language

simonrepp:
This adds a package that offers syntax highlighting support for the eno notation language (https://eno-lang.org)

I'm deeply impressed that all of this is being reviewed manually. :bowing_man:
Thanks so much, you rock!

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -909,6 +909,16 @@
 			]
 		},
 		{
+			"name": "eno",
+			"details": "https://github.com/eno-lang/sublime-eno",
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Ensime",
 			"author": "The Ensime committers",
 			"details": "https://github.com/ensime/ensime-sublime",


simonrepp:
Update: Fixed the CI warning :+1:

Thom1729:
Providing color schemes is fine, but is it really necessary to override the user's color scheme settings? Do the builtin color schemes not highlight this syntax adequately?

simonrepp:
@Thom1729 I can't say if the built in highlighters do an *adequate* job because that's too subjective for me to judge; what I am dead sure about however is that in comparism they are absolutely inferior and flaky. My plan was to un-default the setting override as soon as I have really good video documentation for users that communicates the existence and the how-to for the better custom schemes (which will get more soon), until then I opted for the "good user experience by default" route (see also reasoning here: https://atom.io/packages/language-eno under (*) Custom color scheme). If you think it's a showstopper, I'll of course change it, but it was definitely not without consideration that I went that way. ;)

simonrepp:
Thanks @Thom1729! :balloon:

PR #7191, opened 2018-07-15
add Persistent Console

idleberg:

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -19,6 +19,7 @@
 				{
 					"sublime_text": ">=3000",
 					"tags": true
+				}
 				
 			]
 		},
--- a/repository/p.json
+++ b/repository/p.json
@@ -20,7 +20,6 @@
 					"sublime_text": ">=3000",
 					"tags": true
 				}
-				
 			]
 		},
 		{
--- a/repository/p.json
+++ b/repository/p.json
@@ -19,7 +19,7 @@
 				{
 					"sublime_text": ">=3000",
 					"tags": true
-				}
+				
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


Thom1729:
> This borrows heavily from the BinaryPlist package.

What is borrowed, exactly?

Would `on_new_async` work just as well as `on_new`?

> settings = sublime.load_settings("Persistent Console.sublime-settings") or True

Can the `or True` ever run?

Rather than unconditionally running `ConditionalShowConsoleCommand`, would it make more sense to conditionally run `ShowConsoleCommand`?

idleberg:
> What is borrowed, exactly?

Whoops, it's a remainder from another package of mine which I used as boilerplate

Thanks for the input, I made all changes except for conditionally running `ShowConsoleCommand`. Will get to that a bit later, I had no idea I can pass arguments to `run_command`. So I guess it makes sense 🙂

Thom1729:
Why not just check the setting in the listener and only run the command at all if it's enabled?

idleberg:
How can I get `window` in the EventListener?

Thom1729:
Something like this?

```python
class ConditionalShowConsoleListener(EventListener):
    """Adds listener events."""

    def on_new_async(self, view):
        """Listen for new windows."""
        enabled = sublime.load_settings("Persistent Console.sublime-settings").get("enable", False)
        if enabled:
            view.window().run_command("show_console")
```

idleberg:
Thanks. I tried that before and it works, but it also prints an error for the line where it's used

> NameError: global name 'window' is not defined

Thom1729:
Does it still do that? I don't see where the unqualified name `window` is used anywhere in the plugin.

idleberg:
I restarted Sublime Text, now it's gone

Thom1729:
On an unrelated note, you need to prefix tags with `v`, e.g. `v0.1.0`, due to the way that Package Control works.

idleberg:
> On an unrelated note, you need to prefix tags with `v`, e.g. `v0.1.0`

I don't think that's true, as none of my other packages are tagged that way. I *usually* prefix all my projects with a `v`, Sublime Text being an exception. Well, but since were at 0.1, I can still <del>fix</del> change that.

PR #7193, opened 2018-07-16
Renamed Amxmodx package to AmxxEditor

evandrocoan:
The original package I forked this is named `amxmodx` and it not available on package control, however they can install it externally: https://forums.alliedmods.net/showthread.php?p=2603984#post2603984

And it is causing the original package to be replaced with this. Then, I am renaming it to `AmxxEditor` and after a month this is merged, I am opening another pull request removing the previous names `Amxmodx`.

For now, I am orienting the users of the original package to add `Amxmodx` package to the `auto_upgrade_ignore` setting, until this is completely fixed.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -766,10 +766,9 @@
 			]
 		},
 		{
-			"name": "Amxmodx",
-			"details": "https://github.com/evandrocoan/AmxxEditor",
-			"labels": ["autocompletion", "auto-complete", "amxmodx", "amx mod x", "amxx", "pawn", "build system"],
-			"previous_names": ["amxmodx"],
+			"name": "Amxx Pawn",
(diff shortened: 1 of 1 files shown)


Thom1729:
Shouldn't `amxmodx` still be in `previous_names`?

evandrocoan:
I had put it there, but the unit tests throwed an error about duplicated packages.

Anyways, there is quite some time that I renamed the package from `amxmodx` to `Amxmodx`, so by now, everybody should be renamed to `Amxmodx`.  Actually, it is renaming even more as there is the bug: 

1. https://github.com/wbond/package_control/issues/1370 Package Control is renaming packages on auto_upgrade_ignore setting

Even if the users had added `amxmodx` to `auto_upgrade_ignore` the package was renamed from `amxmodx` to `Amxmodx` anyways.

PR #7196, opened 2018-07-18
A Sublime Text 3 (windows) plugin "MarkdownWriter"

vanleo2001:
The plugin "MarkdownWriter"  is a Sublime Text 3 (windows) plugin that make markdown text writing more easy, especially when you need take note from html to markdown file.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -564,8 +564,7 @@
 				{
 					"sublime_text": ">=3118",
 					"platforms": ["windows"],
-					"tags": true,
-					"dependencies": ["python-pywin32"]
+					"tags": true
 				}
 			]
 		},
--- a/repository/m.json
+++ b/repository/m.json
@@ -557,6 +557,18 @@
 			]
 		},
 		{
+			"name": "MarkdownWriter",
+			"details": "https://github.com/vanleo2001/MarkdownWriter",
+			"releases": [
+				{
+					"sublime_text": ">=3118",
+					"platforms": ["windows"],
+					"tags": true,
+					"dependencies": ["python-pywin32"]
+				}
+			]
+		},
+		{
 			"name": "Marked App Menu",
 			"details": "https://github.com/icio/sublime-text-marked",
 			"labels": ["markdown", "preview", "build", "marked.app"],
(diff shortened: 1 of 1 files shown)


vanleo2001:
I modify the key binding and delete the file 'package-metadata.json'. Please check again

Thom1729:
You should add a license.

Is it really necessary to use win32clipboard? Will the Sublime API `get_clipboard` and `set_clipboard` functions not suffice?

vanleo2001:
I have added the license file. 
For distinguishing the clipboard format, I think it is necessary to use win32clipboard.

vanleo2001:
@Thom1729, I replace the library `pywin32` with `ctypes` to reduce the dependency library.

Thom1729:
It looks like you have a lot of extra print statements in there. If you want to have verbose output for debugging, you should define a logging function that checks a setting before printing.

You should not modify `sys.path`. Use relative imports instead (e.g. `from .lib.PIL import ImageGrab`).

What's the advantage of the Markdown GFM syntax over the built-in Markdown syntax?

It might be a good idea to name the context menu command something more explicit, like “paste HTML as Markdown”. Also, you should implement `is_enabled` and/or `is_visible` to keep it out of the way when it's not relevant. Likewise, the keymap should use contexts.

I'm not a Windows dev, but the function `GlobalLock` looks kind of scary. What does that do? Does it need a `finally` to ensure that the lock is released in case of error?

vanleo2001:
Thank you @Thom1729 for your good advice.

Now I remove the unnecessary print statements, use relative imports and the built-in Markdown syntax, rename the context menu command as“paste HTML as Markdown", and surround the function `GlobalLock` with `try...finally`.

I wonder when should I implement is_enabled and/or is_visible for the context menu command “paste HTML as Markdown? When open the context menu and find clipboard empty? Can you give me more advice? Thanks.

Thom1729:
> I wonder when should I implement is_enabled and/or is_visible for the context menu command “paste HTML as Markdown"? Disable it when opening the context menu and finding clipboard empty? Or just use the shortcut? Can you give me more advice? Thanks.

I would say that it should be *visible* whenever it could reasonably be relevant (perhaps only for Markdown documents?) and *enabled* only when it can actually be used (there is HTML data in the clipboard).

vanleo2001:
`is_enabled` and `is_visible` added to make the plugin working only in Markdown file.

PR #7197, opened 2018-07-18
Add lzma dependency

ehamiter:
This adds the Python 3 module `lzma` as a dependency that can be used for compression.

### DIFF
--- a/repository/dependencies.json
+++ b/repository/dependencies.json
@@ -335,6 +335,21 @@
 			]
 		},
 		{
+			"name": "lzma",
+			"load_order": "50",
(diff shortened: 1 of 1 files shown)


Thom1729:
You should include the original license.

Is it possible to support more than just OS X?

ehamiter:
Good catch @Thom1729 -- I added the license. I would love to support more than OS X, but that's all I have, so I'm not sure how to compile the others-- but I would welcome any pull request for any additional libraries.

dvhh:
The name is confusing, as it seems to be for `p7z`, I haven't inspected the library.

PR #7198, opened 2018-07-19
Add Hover Locales package.

alvesjtiago:
Link to code repo: https://github.com/alvesjtiago/hover-locales
Link to tags page: https://github.com/alvesjtiago/hover-locales/tags

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -897,6 +897,16 @@
 			]
 		},
 		{
+			"name": "Hover Locales",
+			"details": "https://github.com/alvesjtiago/hover-locales",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "How Do I Code Search",
 			"details": "https://github.com/jlangston/SublimeHowDoI",
 			"releases": [


Thom1729:
- The following imports seem to be unused: `struct`, `base64`, `imghdr`.
- Instead of bundling a YAML library, you should declare [`pyyaml`](https://github.com/packagecontrol/pyyaml) as a dependency.
- The event listener should immediately check whether the file is actually a rails file before doing any more work.
- You're not closing file handles: `raw_yaml = yaml.load(open(file_name, "rb").read())`. You should use `with` to ensure that they are closed correctly.

alvesjtiago:
Thank you for the review @Thom1729.

Points 1, 2 and 4 are fixed. Unfortunately i18n can be used not only with rails but also with other frameworks (such as Sinatra) and it is impossible to specify a file type (.rb, .haml, .erb, etc).

Please let me know if any other changes are required. Thank you!

Thom1729:
> Unfortunately i18n can be used not only with rails but also with other frameworks (such as Sinatra) and it is impossible to specify a file type (.rb, .haml, .erb, etc).

In that case, can you just check that the scope under the cursor matches `source.ruby`?

PR #7199, opened 2018-07-20
added Datalife Engine snippets

jyarali:

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -211,6 +211,16 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


jyarali:
added semver tag

Thom1729:
You should add a license.

Also, this is totally subjective, but I'd put the snippets in a directory rather than at the root level of the package. Sublime will still find them.

jyarali:
license added and moved the snippets to directory.

PR #7200, opened 2018-07-20
Add GraphQL Pro Syntax Highlighting Package for Sublime Text 3

thecodechef:


thecodechef:
Okay took out the .no-sublime-package and its still saying i need to take it out please help @wbond

Thom1729:
The latest tag still has `.no-sublime-package`.

Also, you should add a license.

thecodechef:
Ok Added a LICENSE and pushed a new tag to reflect changes everything should be all good now! @Thom1729

FichteFoll:
- You should avoid using look-behind patterns. They will cause the old Oniguruma enging to be used, which is much slower than the sregex engine.
- Unnecessary capture groups slow down parsing slightly as well.
- Commas are to be used in scope *selectors*. In syntax definitions, you only stack scopes and separate them with spaces.
- Instead of `scope: keyword.declaration.reserved-words.graphql, keyword.declaration.reserved-words.directive.graphql`, just use the more specific one, i.e. `scope: keyword.declaration.reserved-words.directive.graphql`.
- `GraphQL Meta.sublime-syntax` has a lot of seemingly useless scope stacking.

thecodechef:
ok @FichteFoll pushed a new tag with some of the changes you requested but im keeping the look behind patterns cuz im still getting used to ruby regex until im more comfortable with it then i will does that sound good to you

FichteFoll:
- In Comments.sublime-syntax, you should not match the entire remainder of the line with `.*\n?` but instead only push and let that context take care of pushing at the end of the line. Also, your `^\s*` match will almost never be useful since the other match should capture all cases already.
- Are you sure you want to set this? https://github.com/thecodechef/sublime-graphql-pro/blob/master/Syntaxes/hidden_syntaxes/GraphQL%20Braces.sublime-syntax#L34 Doesn't strike me as a good idea.
- In variables, you should always use inline flags instead of globally changing the regex mode. I.e. `(?x: extended regex syntax here)` instead of `(?x) extended regex syntax here and for everything following the variable usage`. ([via](https://github.com/thecodechef/sublime-graphql-pro/blob/master/Syntaxes/hidden_syntaxes/GraphQL%20Types.sublime-syntax#L10)).
- You don't need to escape forward slashes ([via](https://github.com/thecodechef/sublime-graphql-pro/blob/master/Syntaxes/GraphQL%20Config.sublime-syntax#L10))
- If you can avoid usage of backreferences in pop patterns, performance should also increase ([via](https://github.com/thecodechef/sublime-graphql-pro/blob/master/Syntaxes/GraphQL%20Config.sublime-syntax#L36))


This is of course not all mandatory but just some tips for working on syntaxes now and in the future.

PR #7203, opened 2018-07-23
New Crontab Syntax Highlighter

nayyarv:
Hi, the current Crontab Syntax Highlighter in Package Control, https://github.com/clarkewd/SublimeCrontab hasn't been touched in 3 years and was written 5 years ago. When I installed it a short while back, it didn't work and I couldn't work out why.

I wrote my own https://github.com/nayyarv/CrontabSublime and as the screenshot shows, is a bit more advanced than the old version and works for the default molokai theme quite nicely. I've been using it for a few weeks and haven't had any issues as of yet and think it might be worth sharing on Package Control for anyone else who'd like to use it. 

Given the lack of development or maintenance of the old package I didn't want to submit my work as a patch on the old one. Furthermore it's been entirely rewritten using the new `.sublime-syntax` style and is significantly different except for one regex (which I intend to replace in a future release). I'm not sure what the current protocol is for abandoned packages.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3728,6 +3728,17 @@
 			]
 		},
 		{
+			"name": "Crontab",
+			"details": "https://github.com/nayyarv/CrontabSublime",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


Thom1729:
I would suggest reaching out to the developer of the old package and seeing if they are still interested in maintaining it. In the meantime, this can be merged as a new package.

PR #7204, opened 2018-07-25
Adding idyll lang snippets

fhightower:
This package provides snippets for the [idyll language](https://idyll-lang.org/).


fhightower:
I've added a semver tag.

PR #7205, opened 2018-07-25
Added package for syntax highlighting for new Rockstar programming language

paxromana96:
This package performs syntax highlighting for the newly-created [Rockstar programming language](https://github.com/dylanbeattie/rockstar). As far as I can find, no one else has created a similar package.


Thom1729:
If you don't intend to provide a `.tmLanguage` fallback (which is fine), then please update the version support declaration.

PR #7206, opened 2018-07-26
Removed "previous_names": ["BBCode"], from "BBCode Syntax"

evandrocoan:
This renaming is quite annoying because I use a private channel where I have this package named `BBCode`. However, as I also have the standard channel, Package Control keeps renaming my package from `BBCode` to `BBCode Syntax`.

There is a long time this package was renamed. So, no one nowadays should be using it with his old name.

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -218,7 +218,6 @@
 			"name": "BBCode Syntax",
 			"details": "https://github.com/chipotle/BBCode",
 			"labels": ["language syntax"],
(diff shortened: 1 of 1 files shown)


FichteFoll:
In fact, the rename happened 4 years ago, before the schema was changed to 3.0: https://github.com/evandroforks/package_control_channel/commit/def67defb3ba2ca383d949c4fb5e2bfce66e6edd

PR #7211, opened 2018-07-28
Added StatusBarSymbols plugin to package control repo channel

OdinTech3:
The plugin shows the symbol name in sublime's status bar based on cursor position. So essentially it smartly displays the symbol of the method or class that your cursor is positioned at in the sublime status bar. 

At the moment it only works for python source code, however I plan to expand it to javascript, php, ruby etc.

There is another plugin that does a similar thing [PythonStautsBarShowSymbol](https://github.com/scriptkitz/PythonStautsBarShowSymbol)

However it has some problems with different python syntaxes or `tmLanguage` definitions. It only only supports python and has been abandoned.

So this new plugin should be a better version that will benefit many sublime text users. 😄

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -3504,8 +3504,7 @@
 			"releases": [
 				{
 					"sublime_text": ">=3000",
-					"tags": true,
-					"platforms": ["linux", "osx", "windows"]
+					"tags": true
 				}
 			]
 		},
--- a/repository/s.json
+++ b/repository/s.json
@@ -3511,6 +3511,17 @@
 			]
 		},
 		{
+			"name": "StatusBarSymbols",
+			"details": "https://github.com/OdinTech3/StatusBarSymbols.git",
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


Thom1729:
You should add a license.

Also, on a subjective note, a unicode arrow character might be better than `->`. Your mileage may vary.

Thom1729:
@OdinTech3 ping.

FichteFoll:
For a general approach (which would probably be quite useful), I suggest looking at the symbol list and then the indentation of the line where the symbol is defined to automatically determine a hierarchy.

That other package is pretty bad, though.

OdinTech3:
Yes what you've proposed is essentially what I want to achieve. The symbol list however depends on the sublime syntax file. For example, the symbol list that appears while using the default Python syntax is different from the one that appears when using MagicPython syntax.

I am trying to figure out how to handle symbol list based on the different syntax definitions.

PR #7213, opened 2018-07-29
Add LinkToSource

jkhulme:
Add an item to the context menu to get a link to the current file in the repository.  This makes it much easier to share links with collaborators.

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -978,6 +978,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


Thom1729:
`__inity__.py` is unnecessary.

Rather than using string manipulation on file paths and URLs, you should use the built-in functionality from `os.path` and `urllib`

Thom1729:
@jkhulme ping.

PR #7216, opened 2018-08-01
Added CodeShark Plugin

JacobSamro:
Helps to bring code into the editor by searching. Users have to save the programs at https://app.codeshark.live to use it within the editor.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2131,7 +2131,7 @@
 			"releases": [
 				{
 					"sublime_text": ">=3000",
-					"branch": "master"
+					"tags": true
 				}
 			]
 		},
--- a/repository/c.json
+++ b/repository/c.json
@@ -2126,6 +2126,16 @@
 			]
 		},
 		{
+			"name": "CodeShark",
+			"details": "https://github.com/aroliant/codeshark-sublime",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"branch": "master"
+				}
+			]
+		},
+		{
 			"name": "CodeStats",
 			"details": "https://github.com/code-stats/code-stats-sublime",
 			"labels": ["analytics", "statistics"],


JacobSamro:
I've made a release tag, please publish the extension

JacobSamro:
I've made all the necessary changes. Will you please merge this pull request. thank you

Thom1729:
- You should have a readme.
- `API_KEY` should probably be a setting rather than a global variable.
- `pickedID` should definitely be a command argument rather than a global variable.
- You don't need to pass no-op `on_change` and `on_cancel` callbacks to `show_input_panel`.
- There doesn't seem to be any error handling if an API request fails.

JacobSamro:
Noted with thanks. Will give an update soon.

JacobSamro:
Okay I'll work on it. Will you please help us to improve the plugin. Please fork the repo and contribute. Pull requests are always welcome!

JacobSamro:
Changes has been made
1. Both of the opened file handlers are closed properly
2. We won't reuse the request that's why we haven't moved it to a function. We'll do it if there is a need in future
3. Now the code will be inserted on all the cursor positions
4. Yes the path is looks UNIX-specific. We have tested it on various platforms and it is working fine.
5. Exceptions are handled properly.

JacobSamro:
@Thom1729 Will you please take a look. Thank you.

FichteFoll:
1. Unless the `~/.codeshark` file is shared among multiple applications, it makes more sense to store the api key inside an ST settings file using
`settings = sublime.load_settings("CodeShark.sublime-settings"); settings.set("api_key", key); sublime.save_settings("CodeShark.sublime-settings")`.

1. Instead of overriding `Request.get_method`, just specify it in the constructor.

Imo catching `Exception` here is fine.

PR #7217, opened 2018-08-02
Added new CommentFold repository

mapsiter:


mapsiter:
Missing release tag was added, please check!

mapsiter:
Thanks for your comments.
Should all be fixed. Please check!

FichteFoll:
When removing folds, you could also just unfold the whole view, but this works as well.

PR #7218, opened 2018-08-02
added GoUnit package

hexdigest:

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -1406,17 +1406,6 @@
 			]
 		},
 		{
-			"name": "GoUnit",
-			"details": "https://github.com/hexdigest/gounit-sublime",
-			"labels": ["go","testing"],
-			"releases": [
-				{
-					"sublime_text": "*",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "Gotham Theme",
 			"details": "https://github.com/9sevendesign/Gotham-Theme",
 			"labels": ["theme", "color scheme", "dark"],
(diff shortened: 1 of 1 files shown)


Thom1729:
Rather than modifying the environment, it would be cleaner to pass a modified environment to the subprocess.

It might be easier to leverage the syntax highlighting for finding the function name. YMMV.

hexdigest:
@Thom1729 can you please point to where I'm modifying the environment? 

I would also apreciate if you point me to the right direction regarding syntax highliting for finding the function name. As far as I understand to leverage sytax higliting I have to require some other plugins to be installed that provide the highliting for Go and this is something I don't want to do, correct me if I'm wrong.

FichteFoll:
How does this differ from the fork origin https://packagecontrol.io/packages/GoTests? Have you tried extending that package instead?

You are modifying the environment [here](https://github.com/hexdigest/GoUnit-Sublime/blob/431d55c9df07df7354077c4d1a4241eafd6a08a5/GoUnit.py#L12-L13). Instead pass a dictionary with env overrides to your `Popen` call. Also, if the user already has `GOPATH` defined, maybe you don't want to override it?

Regarding using the syntax lexer, you can retrieve all symbols of the current file with `view.symbols()`, which returns a list of (region, symbol name) tuples. (Except that for Go it also includes the `func` keyword and return type, but that's workable.)
You can then get the signatures with `view.find_by_selector('meta.function.parameters')`.
Sublime Text ships with a Go syntax definition, so no third-party package is required.

Thom1729:
@hexdigest ping

hexdigest:
@Thom1729 I'm not a Sublime user and honestly I don't have time to fix the PR according all your comments.
This is just a copy/paste of something that is already in your repository (gotests). I just wanted to make life of people who switched from GoTests to GoUnit a bit easier.

PR #7219, opened 2018-08-02
Added the repository

yathartha1:

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -2393,6 +2393,16 @@
 			]
 		},
 		{
+			"name": "Multi Search",
+			"details": "https://github.com/yathartha1/multi_search",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Multi Select Alphabetizer",
 			"details": "https://github.com/jasongornall/sublime-alphabetizer",
 			"releases": [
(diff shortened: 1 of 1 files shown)


yathartha1:
Made the changes

Thom1729:
The global variables are unnecessary. The entire plugin could be something like:

```python
import webbrowser
import sublime
import sublime_plugin


settings = sublime.load_settings("multi_search.sublime-settings")


class MultiSearchCommand(sublime_plugin.TextCommand):
    def run(self, edit):
        for region in self.view.sel():
            if region.empty():
                query_string = self.view.substr(self.view.word(region))
            else:
                query_string = self.view.substr(region)

        if query_string != '':
            def on_done(self, index):
                if index != -1:
                    url = settings.get('domain_urls')[index] + "/search?q=" + query_string
                    webbrowser.open(url)

            self.view.show_popup_menu(settings.get('domains'), on_done)

        else:
            sublime.status_message("No Selection to Search")
            print("StackSearch Plugin: No Selection to Search")
```

Also, it doesn't look like multiple selections are handled.

FichteFoll:
These two packages already do pretty much the same thing:

- https://packagecontrol.io/packages/OnlineSearch (specifically this one)
- https://packagecontrol.io/packages/Search%20Anywhere

Also, the configuration seems pretty [inflexible](https://github.com/yathartha1/multi_search/blob/ca6fc9c0bb9353fb5a532d3bbb641078490f4227/multi_search.py#L18).

PR #7220, opened 2018-08-03
add SCSS and Sass package

braver:
This publishes a new package for [SCSS and Sass](https://sass-lang.com). 

There are currently several packages that provide syntax  definitions and completions for this language, none of which are actively maintained. 

- [SCSS](https://packagecontrol.io/packages/SCSS) is actually an old TextMate bundle, which given the current state of TM is of course RIP. I don't think it makes sense to replace this package, although it's a shame it's so popular because [it doesn't actually work](https://forum.sublimetext.com/t/best-setup-for-scss-editing/38091).
- I've been in touch with the maintainer of the [Sass](https://packagecontrol.io/packages/Sass) package, but he prefers to not change the highlighting (even though it's completely out of line with other CSS-like languages). So, we're not going to merge and replace that, even though chances of that package seeing an update are pretty much zilch.

So, yeah, I'd rather not add another package especially since ancient packages will forever rule the search results. But that's where we're at.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -629,7 +629,7 @@
 			"labels": ["language syntax", "scss", "sass", "completions"],
 			"releases": [
 				{
-					"sublime_text": ">3000",
+					"sublime_text": ">3092",
 					"tags": true
 				}
 			]
--- a/repository/s.json
+++ b/repository/s.json
@@ -160,9 +160,8 @@
 			]
 		},
 		{
-			"name": "Sass",
+			"name": "Syntax Highlighting for Sass",
 			"details": "https://github.com/P233/Syntax-highlighting-for-Sass",
-			"previous_names": ["Syntax Highlighting for Sass"],
 			"labels": ["language syntax"],
 			"releases": [
 				{
@@ -624,7 +623,7 @@
 			]
 		},
 		{
-			"name": "SCSS and Sass",
+			"name": "Sass",
 			"details": "https://github.com/braver/SublimeSass",
 			"labels": ["language syntax", "scss", "sass", "completions", "snippets"],
 			"releases": [
--- a/repository/s.json
+++ b/repository/s.json
@@ -624,6 +624,17 @@
 			]
 		},
 		{
+			"name": "SCSS and Sass",
+			"details": "https://github.com/braver/SublimeSass",
+			"labels": ["language syntax", "scss", "sass", "completions"],
+			"releases": [
+				{
+					"sublime_text": ">3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "SCSS Compiler",
 			"details": "https://github.com/mattarnster/scsscompiler",
 			"releases": [
--- a/repository/s.json
+++ b/repository/s.json
@@ -626,7 +626,7 @@
 		{
 			"name": "SCSS and Sass",
 			"details": "https://github.com/braver/SublimeSass",
-			"labels": ["language syntax", "scss", "sass", "completions"],
+			"labels": ["language syntax", "scss", "sass", "completions", "snippets"],
 			"releases": [
 				{
 					"sublime_text": ">3092",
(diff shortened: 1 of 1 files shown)


Thom1729:
Commented on https://github.com/P233/Syntax-highlighting-for-Sass/pull/96.

braver:
@Thom1729  I already contacted @P233 and he wants to maintain his package and take a different approach at some time in the future. I don’t see it happening, but let’s give him the benefit of the doubt here.

braver:
@FichteFoll @Thom1729 Without disclosing what was essentially a private conversation, the points made against merging my changes:

> It seems it contains too many break changes.
> I personally would like to remain the current highlighting results and focus on bug fixing. But the remaining bugs are a bit tricky, and I no longer use Sublime Text for a very long time, so I am a bit lake (sic) of motivation. But I still want to make this package better and I wii do once I have a bunch of free time.

Perhaps superfluously, breaking changes are necessary, since the Sass/SCSS package in its current state is completely different from CSS even though SCSS should be a superset. I personally think his approach is not a good idea. I think it would be better if the primary packages for CSS pre-processors are in line with the CSS syntax. I'm currently maintaining LESS and my Sass/SCSS fork in parallel and it's super efficient. I've half a mind to merge the repos and simply publish them as a single package in place of the LESS package. I'd love to see all the separate approaches merge so that people can contribute to an actively maintained package, in stead of every time creating a new fork of a dead project to create yet another new dead project.

FichteFoll:
Thanks for the clarifications.

Honestly, I'm not convinced by this. Considering they aren't using ST anymore is usually a red flag for abandonware. (The PR has been opened in March.)

I believe it is in our best interest to replace the current Sass and SCSS packages. We can re-add the current Sass package under its old name "Syntax Highlighting for Sass" and add a notice for that in an update message so that users who prefer the old version (because muh changes or whatever) can still get it.

I don't particularly mind about merging LESS as well, however you might want to consider naming/marketing here. What should the package's name become? "Sass, SCSS and LESS"? "CSS Precompilers"? All of these have downsides. Unless you would benefit from a merge by sharing contexts, I suggest to keep them separate.

I may be a bit biased here, but from my experience with your work I am confident  to believe that the work you did on this has been made with careful consideration and is of good quality overall.

@Thom1729, what do you think?

@P233, would you be okay with this? I don't usually want to overrule package authors' decisions regarding where there packages should go, but when packages become immensly popular we need to be a bit stricter with moderation.

braver:
> the work you did on this has been made with careful consideration and is of good quality overall.

Well quality isn't guaranteed, but basically the same code has been in use by a couple thousand LESS users for a few months already. No big complains there yet. ;) Thanks for the confidence though!

I think @P233 has his notifications silenced for GitHub, I was only able to reach him via direct e-mail.

FichteFoll:
Sent a mail asking for his opinion on this PR.

braver:
@FichteFoll Get any response yet?

FichteFoll:
No, nothing.

Thom1729:
Given the lack of response from the original author, I think that we should move ahead with transitioning the Sass name over to braver's fork.

FichteFoll:
@braver you need to reorder the entries.

Also, what about the current SCSS package? The maintainer hasn't replied either, so it's save to replace that one as well with a previous_names entry. Should consider that in the update message, too.

braver:
@FichteFoll oh, that's right. GitHub and travis weren't available most of yesterday so I didn't look into it more. I need to update READMEs and messages to let people know what happened. Hopefully all of that will happen this week, I'll ping you when I'm ready to merge this.

braver:
@FichteFoll @Thom1729 

I've added the old version of the Sass package under it's previous name. The new package has a README and install, update messages to inform users about the change and the possibility of installing the old version. The latest version of the package has been tagged as 2.0.3, I assume PC will pick that up as a newer semver than v1.2.4. I think that means it's all set to go!

PR #7223, opened 2018-08-05
Add OpenInGitRepository package

arnellebalane:


Thom1729:
@arnellebalane Please address the above error.

FichteFoll:
We have something similar for Github already, but this package is general purpose and attempts to be compatible with any repository, which is :ok_hand: .

PR #7224, opened 2018-08-06
Add ColumnMassage

yangshuairocks:


FichteFoll:
Normally, just providing a Command Palette entry (and maybe a menu item) is enough. Ideally, the user should add key bindings by themselves as there are only so many bindings available, but the number of packages (with bindings) keeps growing.

PR #7225, opened 2018-08-06
added CPrettify package

Gopinath001:

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3583,6 +3583,18 @@
 			]
 		},
 		{
+			"name": "CPrettify",
+			"details": "https://github.com/Gopinath001/CPrettify-Sublime-text-3",
+			"labels": ["c", "prettify", "cprettify"],
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
1. Normally, just providing Command Palette entries (and maybe menu items) is enough. Ideally, the user should add key bindings by themselves as there are only so many bindings available, but the number of packages (with bindings) keeps growing.

2. Your `only_selection` command should also receive the `cprettify` prefix to prevent name collisions. As well as `restore_config`.

3. Since you ship executables, you need to add a `.no-sublime-package` file to have Package Control unextract the package instead of keeping it as a `.sublime-package` zip archive.

4. When accessing the user's global settings, using `view.settings()` allows you to look into project-specific settings as well and should always be preferred over `sublime.load_settings("Preferences.sublime-settings")`.

Gopinath001:
removed the comments from "Default (Windows).sublime-keymap" to fix invalid json

PR #7226, opened 2018-08-06
Stylelint Auto fix with global stylelint command

paleite:

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -3733,18 +3733,14 @@
 		},
 		{
 			"name": "StylelintAutoFix",
-			"description": "Stylelint Auto fix with global stylelint command.",
-			"author": "paleite",
-			"homepage": "https://github.com/paleite",
-			"last_modified": "2018-8-03",
-			"platforms": {
-				"*": [
-					{
-						"version": "1.0.3",
-						"url": "https://github.com/paleite/stylelintAutoFix.git"
-					}
-				]
-			}
+			"details": "https://github.com/paleite/stylelintAutoFix",
+			"labels": ["style", "stylelint", "autoFix", "formatter"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
 		},
 		{
 			"name": "StyleSorter",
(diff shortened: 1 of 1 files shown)


paleite:
The package runs a script included in the package using `node`.

Thom1729:
Why fork the existing package rather than improving it?

Also, I don't see any node scripts in the repo.

paleite:
Hi, and thanks for the feedback!

The existing package was for running eslint (linter for js), but this package is for running stylelint (linter for css, scss, less, etc), which I don’t reckon have the same use cases, hence why I forked it, rather than improving the existing package. What are your thoughts on this?

The package doesn’t run a node-script. I think I may have misunderstood in that case; it runs a script included in the package, which in turn uses a global node script. (I.e. the same procedure as the existing eslintAutoFix-package). Can I remove the .no-sublime-package file?

Thanks!

Thom1729:
> The existing package was for running eslint (linter for js), but this package is for running stylelint (linter for css, scss, less, etc), which I don’t reckon have the same use cases, hence why I forked it, rather than improving the existing package. What are your thoughts on this?

That makes sense. You should clarify in the readme that it does something substantially different from the source; it's not just an improved or modified version.

> The package doesn’t run a node-script. I think I may have misunderstood in that case; it runs a script included in the package, which in turn uses a global node script. (I.e. the same procedure as the existing eslintAutoFix-package). Can I remove the .no-sublime-package file?

Yes, you can (and should).

Other thoughts:

- You should add a license.
- It's probably better to include a default sublime-settings file than to use hardcoded defaults in the script.
- It looks like `Stylelint_auto_fixCommand` should be a `TextCommand` rather than a `WindowCommand`. That way, you can invoke it on a particular view rather than getting the active view.
- Instead of `platform.system()`, you can use `sublime.platform()`.
- I don't really see the point of the `Preferences` class, but YMMV.

FichteFoll:
The auto-fixer operates on the saved file, so perhaps you should verify that the file being edited is indeed unchanged compared to the saved version `view.is_dirty()`. Besides, `view.file_name()` may be `None` in case the file is unsaved.

You can just pre

Thom1729:
@paleite ping

PR #7228, opened 2018-08-08
Add bear app plugin

malexer:
Added Bear App plugin.

This plugin helps to integrate Sublime Text with [Bear App](http://www.bear-writer.com) for taking notes.

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -256,6 +256,18 @@
 			]
 		},
 		{
+			"name": "Bear App",
+			"details": "https://github.com/malexer/SublimeBear",
+			"labels": ["bear", "bear app", "bear-writer", "note", "notes"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"platforms": ["linux"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Beau",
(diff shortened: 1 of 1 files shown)


FichteFoll:
@Thom1729 the package uses a custom URI scheme, not a HTTP request, so POST is out of the question.

Have you tried whether using `webbrowser.open` instead calling `osascript` also works? It does for me, on Linux.

Normally, just providing Command Palette entries (and maybe menu items) is enough. Ideally, the user should add key bindings by themselves as there are only so many bindings available, but the number of packages (with bindings) keeps growing.

malexer:
@Thom1729 as it was already mentioned, I am using [Custom URL Scheme](https://developer.apple.com/documentation/uikit/core_app/allowing_apps_and_websites_to_link_to_your_content/defining_a_custom_url_scheme_for_your_app) to create a Note in Bear App.

I have tested the last version (webbrowser module) with 1 MB text - it worked, but Bear App was very slow consuming that data. In fact: Bear was also slow just to show that note.

@FichteFoll :
1. Thanks, forgot about that module. Fixed.
2. Absolutely valid, removed default key bindings.

Mentioned fixes were pushed to the repo as 1.0.1 version.

PR #7229, opened 2018-08-09
Updates versioning strategy of Open URL from branches to tags

kylebebak:
I'm now the maintainer of [Open URL](https://github.com/noahcoad/open-url), and I created this PR to update the versioning strategy of the package from branches to tags.

It also officially drops support for ST2, which hasn't seen a commit since 2013. ST2 users that want to install the plugin can still go to [the ST2 branch](https://github.com/noahcoad/open-url/tree/st2) and download the repo into their packages directory.


kylebebak:
No, there probably isn't. It just didn't occur to me to change the versioning strategy for only ST3. I'll modify this PR!

PR #7230, opened 2018-08-09
added STexercism

thosgood:

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -3542,6 +3542,16 @@
 			]
 		},
 		{
+			"name": "STexercism",
+			"details": "https://github.com/thosgood/STexercism",
+			"releases": [
+				{
+					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


Thom1729:
You should pick more specific names for the commands than "submit" and "open".

thosgood:
ok, I've changed them

FichteFoll:
I suggest prefixing your commands with `exercism_` to prevent any chances of name collisions.
You also need to push a new tag with the latest changes.

FichteFoll:
It just eluded my mind for some reason that this package uses "ST" as the prefix, which not technically but  violates our naming guidelines (by including a name reference to Sublime Text). You don't need that on a package repository for Sublime Text.

Please re-add the package as just "exercism" and add a `previous_names` entry to that while removing the one I just merged. Otherwise I will do it in a few days most likely.

PR #7231, opened 2018-08-09
Added Latest Migration Plugin

zaidakram:
Added The "Latest Migration" plug-in, which opens the latest migration in Ruby on Rails and Amber applications. Support for Phoenix apps is also in the pipeline. 

This plug-in is very much like the "Rails Latest Migration" But it supports multiple web frameworks. And a could support a lot more web frameworks down the line.

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -358,6 +358,16 @@
 			]
 		},
 		{
+			"name": "Latest Migration",
+			"details": "https://github.com/zaidakram/latest-migration",
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


zaidakram:
Phoenix Framework is also supported now!

FichteFoll:
You have two slashes in your phoenix config (`'folder':'priv/repo//migrations',`).

zaidakram:
Thanks for catching that. Will fix it ASAP.

PR #7234, opened 2018-08-13
adds AMP Snippets package

wethrift:
No existing packages contain AMP (Active Mobile Pages) HTML Snippets


wethrift:
Fixed invalid XML

PR #7237, opened 2018-08-18
Add FileTasks plugin

arthur-s:
https://github.com/arthur-s/FileTasks

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -509,16 +509,6 @@
 			]
 		},
 		{
-			"name": "FileTasks",
-			"details": "https://github.com/arthur-s/FileTasks",
-			"releases": [
-				{
-					"sublime_text": ">=3000",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "File Switcher",
 			"details": "https://bitbucket.org/rbadmaev/sublime_file_switcher",
 			"releases": [
@@ -732,6 +722,16 @@
 			"releases": [
 				{
 					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
+			"name": "FileTasks",
+			"details": "https://github.com/arthur-s/FileTasks",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
 					"tags": true
 				}
 			]
--- a/repository/f.json
+++ b/repository/f.json
@@ -509,6 +509,16 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
This kind of "task" is not unusual, which is why we have [several packages for this already](https://packagecontrol.io/search/command). Do you believe your package provides something overe these?

arthur-s:
@FichteFoll I've found one plugin which is close to mine - it is https://packagecontrol.io/packages/WrapCommand
But this plugin is for building systems, mine is for running commands like copying files from 
`~/work/project/.../view.js` to `remote:/var/www/web/project/.../view.js`
Mine also can be used for building systems, but noy using Tools -> Build System
Also as I mentioned, such feature is presented in VS code.

FichteFoll:
I fail to see what benifit your package provides over normal build systems. It displays a selection of "commands" to execute on-demand. Build system can do that as well, except they have auto-detection for when they should be active, have more variables to be replaced, more features (like changing environment variables) and can be extended by other third-party packages along with an acompanying syntax definition, if desired. If you don't know about what I just talked about, I suggest you read https://www.sublimetext.com/docs/3/build_systems.html. VS Code's "Tasks" were most likely inspired by it.

Packages like https://packagecontrol.io/packages/Commando provide related functionality through providing possibilities for doing something with a command's output and building pipes of them, like showing output in a new view (and only then piping it to a different tool).

Thom1729:
@arthur-s ping

arthur-s:
Give me 1-2 days, I want to learn how to use Build Systems. If build system works, then there's no need in my tool.
BTW I use my tool every day for copying files to remote server)

FichteFoll:
Okay, so I guess the only relevant difference is that build systems don't have a variable for the relative file path. Which is [a significant missing feature](https://github.com/SublimeTextIssues/Core/issues/970), so that gets a pass. It's just that we try to maintain a minimum standard in the package listing and to prevent duplicates where possible so that we don't get 15 packages doing the same thing but with very slight differences and then 14/15 are abandoned, leaving all the users of these empty-handed and having to go through the other 14 ones to find the single one that's still being updated.

As such, I would instead suggest to add this one missing variable to https://packagecontrol.io/packages/ShellCommand (still maintained), which is extremely similar to your package. If you're not interesed in this, then I'll give in and add the package as-is, but currently I find it hard to reason about.

cc @markbirbeck

PR #7239, opened 2018-08-21
add `Dhall` package

SQbQxeKd3JHD8:


FichteFoll:
Dhall looks interesting.

SQbQxeKd3JHD8:
@FichteFoll Indeed, I think so. I only recently stumbled across it, and felt that it deserved some love vis-a-vis editor support.

PR #7240, opened 2018-08-22
add isort-plugin update from thijsdezoete/sublime-text-isort-plugin 

DiegoGallegos4:
Original plugin was discontinued since 2015, updating isort version and maintining it.

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -1120,6 +1120,16 @@
 			]
 		},
 		{
+			"name": "isort-plugin",
+			"details": "https://github.com/DiegoGallegos4/sublime-text-isort-plugin",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "master"
+				}
+			]
+		},
+		{
 			"name": "ITDCHelper",
 			"details": "https://github.com/akalongman/sublimetext-itdchelper",
 			"author": "Avtandil Kikabidze",


FichteFoll:
[Quite unmaintained indeed](https://github.com/thijsdezoete/sublime-text-isort-plugin/pull/25). This is so old and @thijsdezoete hasn't made a single commit in the last year, so I'll skip the 2 week waiting period entirely, that we usually give package maintainers. I don't think this will harm anyone.

FichteFoll:
That said, I propose that you *replace* the existing isort package instead of adding a clone under a different name and that you also use tags for your releases, as @packagecontrol-bot mentioned.

If you drop support for ST2, you can also remove argparse, ordereddict. If you were to change the isort imports to relative ones, you could also remove the `sys.path` modification and `.no-sublime-package`, as it will be able to find the other packages through relative imports. I don't see how this can work in the current form in the first place, since the import of isort happens before the path is modified :thinking:

FichteFoll:
Actually, why don't you just require `isort` to be installed by the user? That way it can be updated independently of the package and you don't need to bother about vendoring. The [project readme](https://github.com/timothycrosley/isort) also mentions that it supports Python 2.7-3.7, so no mention of 2.6, which ST2 uses.

Thom1729:
@DiegoGallegos4 ping

PR #7241, opened 2018-08-23
Add `UseStrict` package

phts:


Thom1729:
Not to be a downer, but would this actually save any time over typing "use strict" manually?

phts:
Yup! It saves a little bit.

PR #7242, opened 2018-08-23
Add `JavaScriptTemplateLiteralEnhancements` package

phts:


Thom1729:
You should add a license.

For what it's worth, the core JavaScript syntax supports the first two features already, although that improvement isn't in the stable release yet.

phts:
Added.

PR #7245, opened 2018-08-27
Magically simple PHP templates

vovayatsyuk:
MagicTemplates plugin automatically inserts autogenerated content for all newly created files, so you don't need to write the same boilerplate code over and over.

And it's not just a static snippet, but a magically generated working code powered by simple and extendable template engine system.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -155,6 +155,17 @@
 			]
 		},
 		{
+			"name": "MagicTemplates",
+			"details": "https://github.com/vovayatsyuk/sublime-magic-templates",
+			"labels": ["magento", "php templates", "boilerplate"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Magma",
 			"details": "https://github.com/chrisdoris/sublime-magma",
 			"labels": ["language syntax", "magma"],


Thom1729:
I'm not getting a clear picture from the documentation of exactly what this package does. Does it automatically insert code into new files? Under what circumstances? How does the user customize that code and those circumstances?

vovayatsyuk:
Yes. Plugin automatically inserts code template into the new file if its name was matched against one of the rules of [detected project type](https://github.com/vovayatsyuk/sublime-magic-templates/blob/master/mt/project.py#L43). (Detection is based on composer.json file contents)

Rules are the set of regex expressions per each project type:

 - [Magento2](https://github.com/vovayatsyuk/sublime-magic-templates/blob/master/templates/magento2/rules.json)
 - [PHP](https://github.com/vovayatsyuk/sublime-magic-templates/blob/master/templates/php/rules.json)

Code templates have its own template processing engine which allows using dynamically calculated variables like `{namespace}`, `{parent_folder|camelcase}` and so on ([Template example](https://github.com/vovayatsyuk/sublime-magic-templates/blob/master/templates/magento2/files/Model/ResourceModel/Entity.php.txt)). 

Most of the expressions are build around composer.json file found in the root of yout project and current filename. That means that plugin automatically detects wich namespace and classname should be used for current file. Specific templates may use [other variables](https://github.com/vovayatsyuk/sublime-magic-templates/blob/master/mt/placeholders.py#L41-L71)].

> How does the user customize that code and those circumstances?

This is not possible. All logic is invoked automatically.

Thom1729:
At the very least, you should make it clear from the readme how and when templates will be inserted.

vovayatsyuk:
Okay, thanks. Will let you know about the update.

vovayatsyuk:
Hi,

hope it's not too long: https://github.com/vovayatsyuk/sublime-magic-templates#how-it-works

I'm also planning to add section about file names recommendations per project. While they are pretty same as MagentoCore use, it may be useful for some people.

p.s. I'm talking about Magento, because, it's the only [well-supported](https://github.com/vovayatsyuk/sublime-magic-templates/tree/master/templates) framework at the moment.

PR #7250, opened 2018-08-30
Add TailwindCSSAutocomplete

bradlc:
This package adds autocompletion for [Tailwind CSS](https://tailwindcss.com/) class names in HTML and CSS files, based on the user’s Tailwind configuration file.

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -220,6 +220,18 @@
 			]
 		},
 		{
+			"name": "TailwindCSSAutocomplete",
+			"details": "https://github.com/bradlc/sublime-tailwindcss",
+			"labels": ["tailwind", "auto-complete"],
+			"author": "Brad Cornes",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Takana",
 			"details": "https://github.com/mechio/takana-sublime",
 			"releases": [
@@ -769,7 +781,7 @@
 					"sublime_text": ">=3000",
 					"tags": true
 				}
-			]   
+			]
 		},
 		{
 			"name": "Theme - Argonaut",


Thom1729:
I don't think this will work when the package is zipped. Generally, you can't send paths to package resources to external programs like node because Package Control will install the package in a zip file. In this case, the best way around this is probably to feed node the text of your script instead of a file path to it.

Is there any reason to compile your script into a bundle? The bundle is a lot bigger and I don't see where you're using any features that node couldn't handle directly.

bradlc:
Hey @Thom1729, thanks for your feedback.

That makes sense. I have updated the plugin to pass the script as a string. Hopefully it will work fine now?

Regarding the JS bundle, the script has large dependencies including [PostCSS](https://github.com/postcss/postcss) which are required. This isn’t immediately obvious because these aren’t direct dependencies, they are dependencies of the `tailwind-class-names` module.

Thanks,
Brad

bradlc:
Do you think this is okay now @Thom1729?

bradlc:
Would it be better to revert the change I made and add a `.no-sublime-package` file?

Thom1729:
Instead of this:

```python
script_path = os.path.join(packages, 'TailwindCSSAutocomplete', 'dist', 'bundle.js')
with open(script_path, 'r', encoding = 'utf-8') as f:
	script = 'var sublime={config:"' + tw + '",plugin:"' + tw_plugin + '"};' + f.read() + '\n'
```

You need to do this:

```python
script = stuff + sublime.load_resource('Packages/TailwindCSSAutocomplete/dist/bundle.js') + '\n'
```

Thom1729:
@bradlc ping

FichteFoll:
Now you just need to push a new tag with the changes and we are good to merge.

bradlc:
@Thom1729 Thanks for merging! What is the process for making this available on Package Control? I don’t see it.

PR #7251, opened 2018-08-31
Add Better Logos Syntax Highlighting

ptcong:
Better Logos Syntax for Sublimetext (supports C,C++, Objective-C, Objective-C++)

Supports all syntax of Logos http://iphonedevwiki.net/index.php/Logos

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -2,6 +2,16 @@
 	"schema_version": "3.0.0",
 	"packages": [
 		{
+			"name": "Better Logos Syntax Highlighting",
+			"details": "https://github.com/ptcong/sublime-logos",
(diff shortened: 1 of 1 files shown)


Thom1729:
You have to alphabetize your package in the channel JSON file.

Acecool:
The current error: Expecting property enclosed in double quotes means you likely incorrectly formatting your JSON... In these files, you can't have a trailing comma... For instance a = { a = true, b = true, } would be bad because b = true has a comma... Also x, if there is nothing after the table, shouldn't have a comma after either.

Thom1729:
It looks like you may have to add `"tags": true` to the channel entry (like other entries have). Also, it definitely needs to be alphabetized.

Thom1729:
@Acecool ping

PR #7253, opened 2018-09-01
Update g.json

DisposaBoy:

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -1380,7 +1380,7 @@
 			"releases": [
 				{
 					"sublime_text": ">=3000",
-					"tags": true
+					"branch": "legacy"
 				}
 			]
 		},
--- a/repository/g.json
+++ b/repository/g.json
@@ -1380,7 +1380,7 @@
 			"releases": [
 				{
 					"sublime_text": ">=3000",
-					"branch": "legacy"
+					"tags": "unsupported-"
 				}
 			]
 		},


Thom1729:
What's the story here?

Thom1729:
Branch-based releases are deprecated. Instead, you should use `"tags": "myPrefix"`. Then, PC will only find tags beginning with `myPrefix`.

DisposaBoy:
PR updated.

billinghamj:
> I don't want it to just vanish without users knowing what happened

It kinda seems like it just vanished? lol

How can users install the old version with PC?

FichteFoll:
It doesn't look like there is a tag with an `unsupported-` prefix on https://github.com/DisposaBoy/GoSublime/tags. @DisposaBoy, you mentioned you want PC users to install the stale branch/release, but currently there isn't any release at all for GoSublime that PC could install.

When doing so, you might also want to include a changelog mentioning this change in your policy so that users that *do* want to receive updates can opt to do so by cloning it manually, or similar. See https://packagecontrol.io/docs/messaging for details.

DisposaBoy:
Thanks for the tips @FichteFoll. This has fallen down the priority list but I'll keep it in mind when I get back to it.

DisposaBoy:
@billinghamj You can't - at least not currently - but I don't know why you'd even want to do that.

If you still want to install it, you can follow the instructions https://github.com/DisposaBoy/GoSublime/tree/legacy#installation and then checkout the `legacy` branch.

billinghamj:
I want to because I have a setup and config which works, and I have no interest in redoing it from scratch when what I have now works absolutely fine.

PR #7254, opened 2018-09-02
add html2scss package

hxss:
The package parse scss rules from html.


Thom1729:
Because these packages are unrelated, it would be better to split this into two PRs so they can be reviewed separately.

You should add a license to each.

hxss:
@Thom1729 License added. "Active view" removed from this pr.

PR #7257, opened 2018-09-05
Added Markdown Code Blocks

molnarmark:
Markdown Code Blocks is a simple plugin that inserts a code block with the given syntax name to your cursor's position, as a snippet.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -389,6 +389,17 @@
 			]
 		},
 		{
+			"name": "Markdown Code Blocks",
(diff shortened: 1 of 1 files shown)


Thom1729:
You should provide a name for your `.sublime-commands` file.

Instead of using `sublime.active_window()` in a `TextCommand`, use `self.view.window()`.

molnarmark:
Thank you for the tips. Those have been added.

PR #7259, opened 2018-09-06
Added PL1 Syntax Plugin

guija:
- PL1: Package to enable basic PL1, PL/I syntax support. Will be improved in the future.
- Pseudo Block Mode: Enables block mode so that cursors are still aligned in same column when jumping to end of line with multiple cursors and when overwriting text.

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -2332,6 +2332,16 @@
 			]
 		},
 		{
+			"name": "Pseudo Block Mode",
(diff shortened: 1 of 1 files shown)


Thom1729:
Both packages should have licenses.

Is the syntax for some specific dialect of PL/I? It would be helpful if the readme would clarify.

The readme for the block mode package should describe how to activate it and exactly what it does.

PR #7260, opened 2018-09-07
Added Rave to package control repo

chen36:
Package for Rave syntax and Rave evaluator.

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -397,6 +397,17 @@
 			]
 		},
 		{
+			"name": "Rave",
+			"details": "https://github.com/chen36/Rave",
+			"labels": ["Rave", "language syntax"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"details": "https://github.com/facelessuser/RawLineEdit",
 			"releases": [
 				{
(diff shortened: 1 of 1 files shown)


Thom1729:
Please fix the above error.

Thom1729:
Instead of explicitly calling `read_settings`, I'd recommend loading the settings on plugin load:

```python
def plugin_loaded():
	global settings
    
    settings = sublime.load_settings(setting_file)
```

I would also recommend making `Evaluator` a `TextCommand` and then running `FetchRave` on that view, not on the view of the output panel. This avoids `active_view = self.view.window().active_view()`, which can be problematic if the user changes views.

As a totally optional suggestion, you might want to try [sublime_lib](https://github.com/SublimeText/sublime_lib)'s [OutputPanel](https://sublimetext.github.io/sublime_lib/modules/sublime_lib.output_panel.html) class:

```python
from sublime_lib import OutputPanel

panel = OutputPanel.create(window, "my_panel")
panel.show()
panel.print("Hello, World!")
```

Thom1729:
@chen36 ping

chen36:
Thank you so much for the recommendation, but unfortunately I will not be available for the improvement now. I will close the request first and reopen it once it is done.

PR #7261, opened 2018-09-07
new package name esoftplay

esoftplay:


esoftplay:
I already added a new tag

PR #7262, opened 2018-09-07
new package name esoftplay

esoftplay:


esoftplay:
What should I do, I already added a tag in my repository

PR #7266, opened 2018-09-09
Add Brittany plugin

eddiejessup:
The package calls brittany, a program to format Haskell source code, and replaces the contents of a file or selection with the program's output. I don't think there's an existing package to do this.

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -783,6 +783,18 @@
 			]
 		},
 		{
+			"name": "Brittany",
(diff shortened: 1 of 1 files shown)


Thom1729:
The `__init__.py` is unnecessary and should probably be removed.

eddiejessup:
@Thom1729 thanks for the merge :)

I've removed the `__init__.py` file, thanks for the hint.

PR #7267, opened 2018-09-09
Add TabNine package.

zxqfl:
Link to code (for the Sublime Text frontend, the backend is closed source): https://github.com/zxqfl/tabnine-sublime/
Link to main website: https://tabnine.com

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -60,6 +60,18 @@
 			]
 		},
 		{
+			"name": "TabNine",
+			"details": "https://github.com/zxqfl/tabnine-sublime",
+			"author": "Jacob Jackson",
+			"labels": ["auto-complete"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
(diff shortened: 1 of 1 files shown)


zxqfl:
Added a tag 0.1.0 to https://github.com/zxqfl/tabnine-sublime/tags.

Thom1729:
It's not clear to me that Package Control can/should redistribute the binaries without an explicit license. @FichteFoll?

zxqfl:
@Thom1729 I tried to address your concerns by adding a note saying that the binaries are covered by https://tabnine.com/eula, which explicitly permits redistribution of unmodified copies.

As the copyright holder I would not want to file any kind of claim against Package Control in any case, since I want my software to be distributed as widely as possible. But I understand why you would want an explicit guarantee.

zxqfl:
Added a new release 0.2.0 with bug fixes and usability improvements.

PR #7270, opened 2018-09-10
add Kung Fury retro theme

argyleink:

### DIFF
--- a/repository/k.json
+++ b/repository/k.json
@@ -459,7 +459,7 @@
 			"releases": [
 				{
 					"sublime_text": ">=3000",
-					"branch": "release"
+					"tags": true
 				}
 			]
 		},
--- a/repository/k.json
+++ b/repository/k.json
@@ -453,6 +453,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


Thom1729:
A lot of functionality is hardcoded to work with particular languages. This may produce poor results when used with other languages.

Also, you probably shouldn't be putting copyrighted movie gifs in the repo.

argyleink:
hi @Thom1729! Thanks for the feedback! 
I started the theme without all the language specifics, a solid theme base is present for other languages. I have another package with the same base (without all the special language cases) that was previously approved here: https://packagecontrol.io/packages/Nintendo%20Switch%20Dark%20Scheme. Diligence has been done to ensure the theme is good, i estimate more than others in many cases. 

regarding copyrighted movie gifs, i think that's a healthy recommendation from you, i appreciate it, but am willing to take the risk that the kung fury team wont come after me for using some of giphys content as a pop reference. though i could be wrong!

argyleink:
pwease

PR #7282, opened 2018-09-13
added MagicTime package

nivin-studio:
This is a format conversion plug-in between time and date format and timestamp format.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -155,6 +155,16 @@
 			]
 		},
 		{
+			"name": "MagicTime",
+			"details": "https://github.com/nivin-studio/MagicTime",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Magma",
 			"details": "https://github.com/chrisdoris/sublime-magma",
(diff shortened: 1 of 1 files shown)


Thom1729:
I seem to recall that there may be other packages that do similar things. Have you looked at any of them?

The formats should probably be configurable via settings.

It would be very helpful if the readme contained at least a brief summary in English.

Thom1729:
@nivin-studio ping

FichteFoll:
The only other package I found that seems to handle timestamp *conversion* is https://packagecontrol.io/packages/Timestamp%20converter, which has quite a small readme and no examples. From the code, you need to specify the format you want to parse manually, which is unfortunate, imho. It does allow specifying a custom target format, however.

Instead of trying to match the different formats, you might as well just check if `strptime` returns a result or throws an exception.

I don't know why you're running the `cut` command there to modify the user's clipboard. I wouldn't want that, personally. If you need to replace a region (selection), use `View.replace` instead of `View.insert`.

There is a merge conflict here. Please resolve it.

FichteFoll:
Also, since this PR timed out, please submit a new PR when you addressed our comments.

PR #7285, opened 2018-09-14
add esoftplay package

esoftplay:

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -1166,6 +1166,17 @@
 			]
 		},
 		{
+			"name": "esoftplay",
+			"details": "https://github.com/esoftplay/sublimetext-snippets",
+			"labels":["snippets", "esoftplay inc", "language syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "ESOLua",
 			"details": "https://bitbucket.org/Cpsgames/sublime-3-esolua",
 			"releases": [


esoftplay:
dear admin, how long its normally take for you to merge a pull request? I saw the other pull request has been posted 5 years ago and it still wait

Thom1729:
- What is the purpose of the snippets.php file?
- The package includes a color scheme, but the readme does not mention it.

esoftplay:
it contains all the independent functions inside the framework. I tought I
don't need to explain every single file is used for.. too much information
in readme only make people lazy to read. do I need to explain that file in
readme.md too?

Thom1729:
If the package provides a color scheme, then the readme should mention it.

The `snippets.php` file is another matter. I'm trying to understand what it does and why it exists as part of reviewing the package. I doubt that there would be any reason to mention it in the readme. Is it a source file that you're using to build the snippet files?

esoftplay:
you're right snippets.php is the source file. my bad

esoftplay:
do you want me to delete snippets.php? maybe I can keep it somewhere else

Thom1729:
It's fine to have the source file there. Personally, I'd suggest at least putting the generated snippets in a subdirectory so that handwritten and generated files aren't intermingled in the same place.

You should mention the color scheme in the readme, though.

esoftplay:
I use that color scheme only to match android studio. which is at that time
we didn't find any theme that close to android studio (dracula).
but now we've found one, so I will delete that color scheme. it's no longer
needed

PR #7288, opened 2018-09-15
Added SublimeLinter-contrib-eslint_d-with-lightscript-support

Darkle:
This is basically the same as https://github.com/roadhump/SublimeLinter-contrib-eslint_d but with support added for the LightScript language. I would have liked to have added this to the original package, but it seems it's no longer maintained or accepting Pull Requests.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -4360,6 +4360,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
Linter plugins should be filed against https://github.com/SublimeLinter/package_control_channel.

Darkle:
Thanks, will do.

PR #7289, opened 2018-09-19
Added gamY

MartinBonde:


Thom1729:
You should add a license.

The keybinds should only apply when you have a gamY file open.

There's a stray print statements in one of the scripts.

MartinBonde:
Thank you for the comments, they are all implemented in the latest tag.

PR #7293, opened 2018-09-21
Add SEPath package

eudorokhov:


Thom1729:
What does this package do? The readme is not very enlightening.

Thom1729:
@eudorokhov ping

PR #7295, opened 2018-09-22
add Simple Haskell package

friedbrice:

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1738,6 +1738,17 @@
 			]
 		},
 		{
+			"name": "Simple Haskell",
+			"details": "https://github.com/friedbrice/simple-haskell",
+			"labels": ["haskell"],
+			"releases": [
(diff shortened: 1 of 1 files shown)


Thom1729:
Is there a reason to have a separate Haskell syntax in this package rather than improving the core syntax?

friedbrice:
@Thom1729 My changes break some scopes in the core haskell syntax. The existing syntax tests can be modified to match, but ultimately the behavior is different. I'm not interested in keeping the old behavior; however, I'd very much like to contribute my changes to the core syntax (instead of publishing them as a separate package) if I had some guidelines on exactly what the policies are on breaking changes.

friedbrice:
@Thom1729 Thanks very much for explaining the process. I'm going to close this pull request and open an issue/pull request to improve the core syntax.

Thanks again.

PR #7297, opened 2018-09-25
add Active View plugin

hxss:
The plugin switch view settings on activate/deactivate.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -220,6 +220,16 @@
 			]
 		},
 		{
+			"name": "Active View",
+			"details": "https://bitbucket.org/hxss/active-view",
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


Thom1729:
When you call `sublime.load_settings`, the `Settings` object you get is live. When `ActiveView.sublime-settings` changes, that `Settings` object will automatically provide the current settings. It's not necessary to manually reload the settings.

Additionally, `window.open_file(sublime.packages_path() + "/User/ActiveView.sublime-settings")` won't work if the package is installed via Package Control.

hxss:
@Thom1729 sorry, I dont have time now for subl plugins. But I remember that it was problematic to use settings object without manual update.

PR #7298, opened 2018-09-26
add toml libray in dependencies

jgirardet:


repository/dependencies.json
```diff
@@ -1073,6 +1073,20 @@
 				}
 			]
 		},
+		{
+			"name": "toml library",
```
FichteFoll: This should have been `"name": "toml"`, at least according to your readme.

jgirardet: there is already a package named toml so I had to find a new name.
I will change this on my repo.
That said adding `toml library` in dependencies today fails. I do not know if it's due to the space in the name or if it only take a bit time after merge.

FichteFoll: I don't think spaces have usage implications on the curretn setup, although I would strongly recommend avoiding them.

Do you have a link for the `toml` dependency and where it's defined? It's not in `dependencies.json` on this repo, which means it's hard to find as dependencies aren't displayed on the website.

jgirardet: https://packagecontrol.io/packages/TOML
You mean windows is case Insensitive ?
At the end whats you proposal for the name : sublime_toml, toml_library, toml_lib ? should I repush a PR with a new name ?

FichteFoll: I suppose `python-toml`, like we already have `python-markdown` and `python-jinja2`, is the best solution.

jgirardet: done in  #7343

PR #7299, opened 2018-09-27
Added Hide Mini Map

owenbyrne:
A simple plugin that will ensure the mini map in Sublime Text 3 is always hidden.


Thom1729:
Silly question, but is this functionality not available as a built-in setting?

PR #7303, opened 2018-09-29
add Create Init plugin

Boris-code:

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3700,6 +3700,16 @@
 			]
 		},
 		{
+			"name": "Create Init",
+			"details": "https://github.com/Boris-code/CreateInit",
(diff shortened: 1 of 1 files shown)


Boris-code:
fixed shortcut key conflict

Boris-code:
Hi, I haven't seen the create init plugin in package contral. what can i do ?

Thom1729:
The keybind is unconditional; it will run regardless of whether it would make sense to do so. That should probably be changed. In addition, you should have a `sublime-commands` file so that the command is accessible from the command palette.

It also like like the command will silently overwrite an `__init__.py` file if one already exists.

Thom1729:
@Boris-code ping.

PR #7307, opened 2018-10-03
adding Dank Neon color scheme

jpetrucciani:
Dank Neon is a nice neon theme that is loosely based on Miami Vice coloring

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -82,7 +82,7 @@
 		},
 		{
 			"name": "Dank Neon",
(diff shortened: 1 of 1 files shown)


Thom1729:
Do all of those files have to be there? Is there some way of organizing them by editor?

jpetrucciani:
The latest update to this PR links to [this repo](https://github.com/DankNeon/sublime) which is only the sublime theme - we ended up splitting them into repos per editor/terminal

PR #7308, opened 2018-10-03
Added Cosmos Scope package

KristianHolsheimer:
This package is a simple syntax highlighter for Cosmos Scope, which is a big data tool used at Microsoft.


KristianHolsheimer:
Will open a new PR for which the automatic tester won't complain.

PR #7312, opened 2018-10-03
Add MDX package

jonsuh:
Syntax highlighting support for [MDX](https://github.com/mdx-js/mdx) (`.mdx`) files.

Package: https://github.com/jonsuh/mdx-sublime

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -1037,6 +1037,19 @@
 			]
 		},
 		{
+			"name": "MDX Syntax Highlighting",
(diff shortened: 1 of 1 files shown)


Thom1729:
I'd recommend a somewhat different implementation for the syntax. You should be able to use `embed` to highlight the JavaScript/JSX rather than duplicating the logic. I'd recommend using [JS Custom](https://packagecontrol.io/packages/JSCustom) (which I wrote) as the baseline, because babel-sublime is quite old and buggy.

PR #7314, opened 2018-10-06
Created the DiffMatchPatch dependency

evandrocoan:
https://github.com/evandroforks/DiffMatchPatch

### DIFF
--- a/repository/dependencies.json
+++ b/repository/dependencies.json
@@ -219,6 +219,20 @@
 			]
 		},
 		{
+			"name": "DiffMatchPatch",
+			"author": "evandrocoan",
+			"load_order": "09",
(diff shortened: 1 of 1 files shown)


Thom1729:
I don't think the `.no-sublime-package` file does anything for dependencies.

Is all of the code necessary? If only the python3 code is used, perhaps the rest should be removed.

evandrocoan:
The .no-sublime-package file does nothing. 

The other code is there because I just forked the repository and added the loader code for Sublime Text. I just removed it.

Thom1729:
Could you remove the `.no-sublime-package` file too, if it doesn't do anything?

evandrocoan:
DOne

PR #7317, opened 2018-10-09
Add tagml-sublime-syntax

brambg:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 .DS_Store
 *.pyc
 *.sublime-workspace
+.idea
+*.iml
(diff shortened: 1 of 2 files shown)


Thom1729:
What is the `tagml-sublime-syntax.iml` file?

It's fine to bundle a theme with the syntax, but is it actually necessary or could the syntax definition be made more specific to work with many themes?

brambg:
The  tagml-sublime-syntax.iml is a system file used bij the Intellij ide.
It shouldn't have been pushed to the git repo, and I've removed it.
Do I need to make a new release for this?

The syntax works without the theme, but not all scopes will be highlighted.
The enclosed theme will highlight all the scopes defined in the syntax.

PR #7318, opened 2018-10-10
Package Name: Acecool's Code Mapping System ( Using folder AcecoolCodeMappingSystem )

Acecool:
…1.0.0 ( which will have a lot more features and controllability but I wanted to go ahead and get this process tarted )

In short: This package takes a look at source-code and converts tens of thousands of lines into a concise overview of that code while creating clickable lines ( jump-to links ) to code in the source file for all entries... Entries include anything from notes, tasks, errors, function definitions, class definitions, callbacks, accessor-func creation, optimization issues ( print statements, and more ), and a lot more and outputs it in an easily configurable and easy to read panel.

It makes coding more efficient because you no longer need to scroll through the file, tens of thousands of lines in some cases, just to find out which segment of the file a function was defined within a class. The output is, by default, organized alphabetically so you scroll to the class or global function list category, then click on the function and you're there in the file. It can take seconds to 10s of seconds to scroll through a large file to find something but with Acecool Code Mapping System, it takes a few seconds at most to locate and click on the thing you're looking for.

With all of the callbacks, you define how the output looks and feels - whether you want it to look exactly as it does in the code, or to modify the function arguments to show the data-types they expect as a pseudo-live-wiki style output, you can do it.. You can add dynamically generated functions to the appropriate categories in the output panel too.. You can even design checks, some languages have them coded in by default and I plan on working on something more automatic, which locate hard to spot errors which could be easily missed by the Human eye because of how we process text and read. For instance, it can be hard to notice a missing : in Python just like it can be easy to confuse != with := or other characters and if you're like me and switch between dozens of languages it can be just as easy to forget Lua vanilla uses ~ and not instead of ! and doesn't have continue ( you'd use an inverted logic statement, to the continue statement, encapsulating everything after where the continue would've gone ) so it can find these hard to detect issues and show you a viable solution.

More info:

This package started as a modification to CodeMap by oleg-shiro, he said it was everything he envisioned his addon to be but wasn't interested in working together or even introducing the category system to prevent wasted space or fix the issues relating to circular imports so I started my own addon.

I introduced a lot of features such as the category system, caching file data to prevent remapping every time you click on a file, and more.. It would be used as the mapping file instead of modifying the base plugin.

It does a lot more than CodeMap, it is a complete system. It has a dynamic and fixed category system where entries can be added to categories ( and they are output in a certain order so the category header isn't output more than once if you add an entry from one category and then another and back to the first - ie instead of 3 category headers being output and wasting space, it outputs the 2 )..

Category entries can be sorted alphabetically, enabled by default to help find function calls, etc..

Error checking is a part of the system.

Sublime Symbols can also be used to automatically map languages which haven't been mapped by a human.

You can set up custom mappings so projects can use a specific mapper - and there is a rule system coming soon to extend this beyond the current system..

I am working on another system for 1.0.0 which will allow you to define the languages mapped by a mapper so you won't need to remap those - for instance a .php file can have HTML, JavaScript, RegEx ( JS ), PHP, RegEx( PHP ), CSS, SQL languages attached along with a personal mapper in the near future so the base language mapper code never needs to be repeated for a mapper and you can include as many of them as you want per file - which will be bound by file-extension or by the syntax you select. Then depending on your configuration, and the new rules system which is coming soon you could, for instance have vanilla Lua used for all Lua files EXCEPT if inside of a garrysmod/ folder which would then enable GMod_Lua, and if inside of a gamemodes/acecooldev_framework/ path then my AcecoolDev_Framework GMod_Lua mapper could be enabled which adds callbacks for the AccessorFunc Expansion and more..

This system allows you to add callbacks at many different points of the mapping process - for instance, you can add a callback to expand Accessor Functions - ie a piece of code as simple as AccessorFunc.Create( ... ) can create an unlimited number of functions in your code, so they should be displayed in the output panel too..

Other callbacks are some which allow you to alter / deny entries using PreOnAddEntry.. You can change the depth at which the entry is displayed at using CalcDepth, you can change the characters, soon, displayed in front of those entries in case you want to draw lines to make the output look like a tree system ( and 1.0.0 will have a dynamic folding system to fold anything )

1.0.0 will have a new region mapping system, an every character is fair game to be mapped in the output panel instead of having each line exploding on :<line_number> to click and jump to that line in the source-code... in 1.0.0 maps will be hidden so you can create multiple jump-to links for class definitions, nested classes, function definitions, on_hover description boxes shown for each argument, data-types, and more... 

1.0.0 will have a new toolbox feature which really sets this apart from anything else... This feature will allow live-updates to the output panel so say you click search and enter a term, the output panel can filter or mask for that search term and be updated in real-time. There will be useful quick-links just like an IDE tree / mapper so you can click on a class to hide it, or show nothing except entries containing it. 

1.0.0 is very close..

I'm adding this addon now because it is so close... and already, it performs better than CodeMap, doesn't introduce some of the bugs, doesn't use circular imports, allows nesting / inheritance of maps, has callbacks to alter how data is mapped and to detect errors, and more...

Another feature are language maps - links to language wikis and ways to interpret to load the data so when a function is mapped which is in the wiki, all of the important information is added along with descriptions, examples and more.. 

There will be a snippets category, macros category, bookmarks category which can take those entries and assigns those actions on-click by default and more which can be assigned to any of the panels the addon creates that you want so it'll be an efficiency on steroids style of plugin you won't want to do without..

I am working on a solution to the thread locking up ( using the cache system takes 0.1 seconds or less to load a mapped file and on my 10 year old computer mapping a file can take a few seconds - with the threaded system I hope to have all of that in the background and not be noticed along with preprocessing so files can be pre-mapped from the project or open files - whatever you decide in the configuration - by default, open files will be pre-mapped )

Note: Version 1.0.0 will have a lot of new features which will make this look, feel and behave more like an IDE style system and will improve efficiency many times over.

Thank you for taking the time to read this and I hope to see it included in Package Control soon.

Please note: Why isn't it possible for me to define the package folder name as AcecoolCodeMappingSystem while defining the human-readable / searchable term as Acecool's Code Mapping System?


Acecool:
Ah, I see about the tmLanguage fallback - needed for pre 3092 version... I'll work on adding those.... The code already has a backup pre-loading so it looks for tmLanguage or sublime-syntax... So, it wouldn't be a bad idea to have both...

Acecool:
I added tags, so what gives? Should the repo point to src?

Acecool:
Packages added:
  - AcecoolCodeMappingSystem

Processing package "AcecoolCodeMappingSystem"
  - WARNING: '.no-sublime-package' is defined. Please verify that it is *really* necessary


- Ok, this looks good... Yes, the files have to be external - there was one bug I encountered if they weren't either from loading a package... actually, it could've been from the internal plugin reloading system... Or it could've been something else.. I need to go over it again - the exact reason behind needing it extracted is important.

Acecool:
How do I make the public name Acecool's Code Mapping System while retaining AcecoolCodeMappingSystem package folder?

Acecool:
Right now I have the object extract the map for the specific case - so if Python is the language using the User mapper, then it'll extract that default - this poses a problem which is why I had .no-sublime-package as the mappers aren't visible for the user to easily make changes, see live examples, etc...

I'll likely extract all of them, and then move all of the *_User maps to the User/AcecoolCodeMappingSystem/maps/ folder as I originally intended - if a User file doesn't exist, or if it does but it was a previous hash of that file ( unedited ) while leaving a copy of the User file in the base package folder to use as a backup.

As for the issue I was experiencing in the past regarding packed files - The only bugs, and implemented a hack for, or changes to the behavior I seem to have experienced, are:

 - On Single / Double-Click Detection to jump to a line in a source-file after clicking in the output panel - without a small delay in the packed version ( or portable Sublime Text ), regardless of on text command ( async, pre / post modifiers ) used, it doesn't jump to the line you just clicked on, but the line which was previously clicked on. This issue does not occur in the base install of the addon which is fully extracted - this could be related to my settings or my addons - I'll look into why at some point.
 - When a file is mapped, it doesn't take as long on the portable / fresh install of Sublime Text vs the current one, and this is expected because of the number of packages I have installed on my primary editor - once I have threading implemented properly, it won't be noticed.

FichteFoll:
Skimming through this certainly was a ride (and probably why it wasn't reviewed earlier).

> - File: examples/SublimeTextKeyMap.sublime-keymap
> - Exception: No JSON object could be decoded

> These files aren't meant to be processed - they are meant to be used to verify the code mapping system is working... I have removed them for now - I will re-add them when I figure out how to prevent Sublime Text from trying to execute them automatically, and resolve the errors listed... Some of the files are empty which could be why there is this error...

You cannot prevent ST from parsing these files. You either have to rename them to use a different extension or remove them from the project.

> There is no fallback file because of this and because PHP Source is coded this way by default - I will look into a way to dynamically modify the existing files so these files won't need to exist in the main plugin path ( ie create them dynamically with fallbacks too? )

You should just declare that the package is only compatible with ST builds >=3092.

---

https://bitbucket.org/Acecool/acecoolcodemappingsystem/src/8a36c3d3394cd020b43417eb63ea2ebcdd67755a/AcecoolHelper_GetSublimePackageFolder.py?at=master&fileviewer=file-view-default#AcecoolHelper_GetSublimePackageFolder.py-62
This will not work for archived packages since `__file__` will look like `…/sublime-text-3/Installed Packages/AcecoolCodeMappingSystem.sublime-package/AcecoolHelper_GetSublimePackageFolder.py`.

… and that's basically where I gave up. This package is just **unreviewable**. I'm not getting paid enough for this. It has a 13k line file and one with 5k. Despite that, code density is actually pretty small and many trivial algorithms (think of one or two-liners) are factored out into some helper library that I didn't even go into.

So, instead of doing a full review, I'll just ask you a few questions and take that as a good faith declaration of some sort.

1. Do you communicate with a remote server to analyze the code? I presume you don't but I'd like to verify.
2. Did you test your package in an archived state? PackageControl provides a command that packs your package and that you can then deploy into the Installed Packages folder. As I mentioned above, there might be some problems.
3. Can you formulate some actual usage instructions in the readme? I mean, background on why it was created is nice and all, but before that I'd be interested in how to *use* this peace of software. A screenshot would also help
4. Why do you ship syntax definitions for PHP, C++, etc? What is the purpose of those?

PR #7319, opened 2018-10-10
Add "Freeze Ruby Files" package

peagha:
This plugin automatically adds the [magic Ruby comment](https://www.lucascaton.com.br/2016/01/19/what-is-frozen_string_literal-in-ruby/) to Ruby files when saving them:
before save:
```ruby
class MyRubyClass
end
```

after save:
```ruby
# frozen_string_literal: true

class MyRubyClass
end
```
Many linters require this comment and the developer has to remember to add it to every file. With this plugin the developer won't even have to remember ✨

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -1617,6 +1617,17 @@
 			]
 		},
 		{
+			"name": "Freeze Ruby Files",
+			"details": "https://github.com/peagha/freeze_ruby_files",
+			"author": "Philippe Hardardt",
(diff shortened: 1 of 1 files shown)


Thom1729:
You should probably enable/disable based on a setting, not by running a ruby process.

I'd also check first to see if you're in a Ruby file before searching the view.

Thom1729:
@peagha Ping.

PR #7321, opened 2018-10-11
Added BAROC syntax highlighting package

Bushikot:
I didn't found existing package for BAROC syntax highlighting, so here it is.


Bushikot:
Oh, completely forgot about versioning. Fixed.

PR #7322, opened 2018-10-12
Added PipenvPath

dhelonious:

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -1422,6 +1422,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


Thom1729:
I'm not sure what the point is of setting the PATH when a view is activated. Is the same PATH not shared between all views and projects?

Thom1729:
You don't need to check `if int(sublime.version()) < 3000` because the package is only declared for >=3000.

dhelonious:
Thank you for the suggestion. I removed the version check.

PR #7323, opened 2018-10-12
Change EasyClangComplete link

niosus:
This is a small change to a link to EasyClangComplete as I now have set up online documentation with github pages. So now there is no need to point straight to the code.


niosus:
Wait, does it have to be a github link judging from the output from the ci build...

Acecool:
It can be Github, Bitbucket, and probably others.

niosus:
@Thom1729 pushing this PR through a failed check has made my package "missing" on package control. As @Acecool has pointed out it probably is better to keep the pure github link. Can you revert this PR? I was not expecting it to be merged as the check has failed...

FichteFoll:
>Wait, does it have to be a github link judging from the output from the ci build...
>If this is the case, may I ask why? It seems github pages should still be a valid place to point to, right?

As I touched in #7361, the `details` key does much more than set the homepage url. In practice, it is used to:

- set the package name and author, based on the path fragments
- set the readme and issues url
- also set the homepage url, which will be either the same or the one set as the website of your repo description
- set the base url for all release channels, unless overridden

PR #7326, opened 2018-10-13
Remove dired package which no longer exists.

mkleehammer:
The package has not existed for quite some time though the source is available.  I'm freeing up
the project name so someone can take over.

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -782,6 +782,16 @@
 			]
 		},
 		{
+			"name": "dired",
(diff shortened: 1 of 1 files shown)


Thom1729:
If the intent is to replace the existing package with https://github.com/kublaios/dired, would it make more sense to do both with one PR? That would ensure continuity.

Thom1729:
@mkleehammer ping

FichteFoll:
I went ahead and did just that.

PR #7327, opened 2018-10-14
add LinkedData package

blake-regalia:
This package brings syntax highlighting support for the most widely used languages among Linked Data users and developers. https://github.com/blake-regalia/linked-data.syntaxes

Each syntax is generated from a complete implementation of the language's grammar specification, resulting in highly-specific scoping and invalid syntax detection. This is lightyears beyond the basic keyword tokenization provided by the existing package(s) available.

Furthermore, this package embeds the mappings for registered prefix namespaces so that users can easily verify or produce the correct (and intended) prefixes while developing or inspecting the source of an RDF document or query.

Finally, this package will remain actively maintained as new prefixes are registered and new versions of RDF languages are standardized.

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -899,6 +899,17 @@
 			]
 		},
 		{
+			"name": "LinkedData",
+			"details": "https://github.com/blake-regalia/linked-data.syntaxes",
+			"labels": ["semantic web", "rdf", "linked data", "sparql", "turtle", "trig", "n-triples", "n-quads", "shex"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Line Completion",
 			"details": "https://github.com/astropanic/Compline",
 			"releases": [
--- a/repository/l.json
+++ b/repository/l.json
@@ -899,17 +899,6 @@
 			]
 		},
 		{
-			"name": "LinkedData",
-			"details": "https://github.com/blake-regalia/linked-data.syntaxes",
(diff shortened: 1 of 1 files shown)


Thom1729:
I'm not quite sure how this works. The package doesn't appear to contain any `sublime-syntax` files, nor any scripts that would cause Sublime to create them. What am I missing?

blake-regalia:
The assets are built automatically by Travis-CI, see under 'releases': https://github.com/blake-regalia/linked-data.syntaxes/releases

wbond:
@blake-regalia I don't believe this will work with the current GitHub provider and client. Here is what it sees about the repo: https://api.github.com/repos/blake-regalia/linked-data.syntaxes/tags.

You'll probably need to add a step to your CI that manually writes a https://github.com/wbond/package_control/blob/master/example-repository.json with info about each of the packages. If there is just a single package being generated, CI will need to create a zip file that contains all of the necessary files in the proper structure, and then you'll need to write the repository.json and upload it somewhere. Once that is done, that well-known/stable URL can be added to https://github.com/wbond/package_control_channel/blob/master/channel.json#L3.

blake-regalia:
@wbond Thanks for the info! Really helpful, makes sense. I opened a new PR only because it seemed appropriate for this case. However, packagecontrol-bot is rejecting the sublime-package URL because it returns an HTTP 302 redirect -- but this is just GitHub giving the proper HTTP status code to where the built file is hosted on an S3 bucket.  [#7345](https://github.com/wbond/package_control_channel/pull/7345#issuecomment-435567570).

blake-regalia:
Closing in favor of #7345

PR #7332, opened 2018-10-19
Add Terrafmt

p3lim:

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -524,6 +524,17 @@
 			]
 		},
 		{
+			"name": "Terrafmt",
+			"details": "https://github.com/p3lim/sublime-terrafmt",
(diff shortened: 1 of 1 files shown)


Thom1729:
You should check the scope in the `on_pre_save` listener.

p3lim:
@Thom1729 What?

Thom1729:
In `TerrafmtListener`, in the `on_pre_save` method, before running the `terrafmt` command, you should check `view.score_selector(0, 'source.terraform') != 0`.

p3lim:
What would be the benefit of doing that opposed to enabling the command only on the files?

Thom1729:
I guess the `is_enabled` check will do the trick.

PR #7335, opened 2018-10-23
Update t.json, Terminal-command

looking-for-a-job:

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -456,6 +456,17 @@
 			]
 		},
 		{
+			"name": "Terminal-command",
+			"details": "https://github.com/looking-for-a-job/sublime-terminal-command",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": "osx",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


Thom1729:
@looking-for-a-job 	Please address the above error. Once you've added a valid semver tag (I don't think 0.0.0 will work), close and reopen this issue and the bot will recheck it.

Thom1729:
You should add a license.

You should probably also expand the readme at least a bit. It looks like the build system will open Terminal, open a new window, and run the command in that window. A little more explanation would go a long way.

In the Python script:

- Instead of checking `"darwin" in sys.platform.lower()`, you can check `sublime.platform() == 'osx`. Since the package is declared only for the Mac OS, the check probably isn't necessary. If you do keep the check, then when it fails you should display an error instead of silently doing nothing.
- It's not obvious from the readme that the plugin will silently change file permissions. This needs to be documented. Ideally, the user might turn this on and off with a setting.

Thom1729:
@looking-for-a-job 	ping

looking-for-a-job:
Thom1729, done

PR #7336, opened 2018-10-26
add Technion C99

NeatNit:
I made this easy to use build system for my CS intro class, and want to share it with as many people as I can since the alternatives (Code::Blocks, Visual Studio) are a lot of hassle to use the way the class uses them. Adding it to Package Control would be the easiest way to share.

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -360,6 +360,18 @@
 			]
 		},
 		{
+			"name": "Technion C99",
+			"details": "https://github.com/NeatNit/Technion-C99",
+			"labels": ["build system", "technion"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"platforms": "windows",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "TECS",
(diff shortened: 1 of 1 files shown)


NeatNit:
If I add a .bat file, will I need to add `.no-sublime-package` to be able to access it once my package is published on Package Control?

Thom1729:
I can't read the readme. What does this package do? Where does the .bat file fit in?

NeatNit:
The readme has only instructions for installing the package (without Package Control), and instructions for installing MinGW GCC.

This package is just a C build system that compiles a single C source file and runs some tests on it. The .bat file is just the build script that runs gcc to compile and tests the resulting .exe in the requested ways.

The purpose of this package is to be easy to use and fast for students of the intro class, of which I am one.

FichteFoll:
The package won't work as-is because PC installs packages in an archived `.sublime-package` file where you can't access its file contents directly. You'll have to provide a `.no-sublime-package` file to allow for the build system to use the bundled .bat file.

PR #7337, opened 2018-10-26
Added QuickSettings Package

evandrocoan:
https://github.com/evandrocoan/QuickSettings

### DIFF
--- a/repository/q.json
+++ b/repository/q.json
@@ -308,6 +308,18 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


Thom1729:
Why is `.no-sublime-package` necessary?

evandrocoan:
It is not. I removed it.

Thom1729:
You need to release a new tag.

PR #7340, opened 2018-10-28
Add Bats syntax highlight package

sptndc:


Thom1729:
Have you looked at https://github.com/jverdeyen/sublime-bats?

sptndc:
Oh sorry, I think that's only for sublime text 2.

PR #7342, opened 2018-10-30
Add Laravel Latest Migrations

aymericauberger:
This package does the same as the Rails Latest Migration package, but for Laravel apps: it simply opens the last migration file from the migrations directory.

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -286,6 +286,16 @@
 			]
 		},
 		{
+			"name": "Laravel Latest Migration",
+			"details": "https://github.com/aymericauberger/Laravel-Latest-Migration",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Laravel Migrations Snippets",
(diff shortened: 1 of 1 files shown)


Thom1729:
Instead of having a `WindowCommand` and getting `active_view()`, you should have a `TextCommand`.

Have you tested the recursion base case on Windows?

Thom1729:
@aymericauberger ping

aymericauberger:
@Thom1729 Thank you for these corrections!

I'm sorry this is my first Sublime package, I didn't really dig into the documentation but just adapted an existing one. I switched to a `TextCommand`, I hope I did it correctly.

And no I didn't test on Windows, only on OS X, should I mark it OS X only?

I also made another, almost identical package, for opening the latest Laravel log file. Do you think they should be combined into one?

aymericauberger:
I actually found a package that does the same thing for multiple frameworks, it's probably better to keep only this one: https://packagecontrol.io/packages/Latest%20Migration
I submitted a pull request to add Laravel support: https://github.com/zaidakram/latest-migration/pull/1

PR #7343, opened 2018-10-30
update `toml library` to `python-toml`

jgirardet:
according to discussion here : https://github.com/wbond/package_control_channel/pull/7298

### DIFF
--- a/repository/dependencies.json
+++ b/repository/dependencies.json
@@ -750,6 +750,20 @@
 			]
 		},
 		{
+			"name": "python-toml",
+			"load_order": "01",
+			"description": "Python lib for TOML",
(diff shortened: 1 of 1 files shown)


jgirardet:
I can't use it : 
```
Package Control: The dependency 'python-toml' is not currently installed; installing...
Package Control: The dependency 'python-toml' is not available
Package Control: The dependency 'python-toml' could not be installed or updated
error: Package Control
```
I don't know what to change.

FichteFoll:
Can confirm. It's not in https://packagecontrol.io/channel_v3.json either. I can't see why though, so probably something @wbond must investigate on the crawler side.

PR #7344, opened 2018-10-30
added Pest Syntax

dten:

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -834,6 +834,16 @@
 			]
 		},
 		{
+			"name": "Pest Syntax",
+			"details": "https://github.com/dten/pest-sublime",
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"tags": true
+				}
+			]
+		},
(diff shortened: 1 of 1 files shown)


dten:
have since added README

dten:
@wbond I imagine you're busy but how often are these looked at? is there something else that needs asserting?

wbond:
@dten Unfortunately due to my personal life, I have a severe deficit of free time, so I am relatively uninvolved in reviewing.

We have gracious, hardworking volunteers in @FichteFoll and @Thom1729 who do the majority of the work these days. In general requests are done in order of submission. You can get an idea of where you are in line by looking at the open pull requests.

dten:
Thanks for the info Will

Thom1729:
Generally, you would append `.pest` to the scopes in the syntax, e.g. `scope: support.function.override.pest`.

dten:
Yep sure thing

dten:
done

PR #7346, opened 2018-11-05
Add JsModules plugin

luwes:
This change adds my JsModules plugin.
https://github.com/luwes/SublimeJsModules

> JsModules is a Sublime Text Plug-in to convert CJS modules to ES modules.

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -1050,6 +1050,17 @@
 			]
 		},
 		{
+			"name": "JsModules",
+			"details": "https://github.com/luwes/SublimeJsModules",
+			"labels": ["javascript", "convert", "cjs", "es6"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "JSON Key-Value",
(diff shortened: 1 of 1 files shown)


Thom1729:
Why is `.no-sublime-package` required?

luwes:
Because the plugin depends on a NPM library, cjs-to-es6.

https://packagecontrol.io/docs/submitting_a_package#Step_5

> If including executables or shared libraries, add a .no-sublime-package file. Adding this file to the root of your repository will prevent Package Control from keeping your package packed as a .sublime-package file in ST3.

Thom1729:
`.no-sublime-package` is required when one of the files in the package needs to be accessible in the filesystem. For example, you'd need it if you had a `something.js` file in the package and you were invoking `node something.js`. But you don't have any JavaScript files, so I don't see any reason you would need it here.

Other minor issues:

- You don't need an empty `dependencies.json`.
- I don't see why you need the `package.json` or `package-lock.json`.
- Instead of using `WindowCommand` and getting `window.active_view()`, you should use `TextCommand` and `self.view` (and `self.view.window()`). That way, you also don't need to check `if active_view:`.

PR #7347, opened 2018-11-06
add poetry support

jgirardet:
frontend in sublime to work with https://github.com/sdispater/poetry

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -1628,6 +1628,17 @@
 			]
 		},
 		{
+			"name": "poetry",
+			"details": "https://github.com/jgirardet/sublime-poetry",
+			"author": "jgirardet",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


Thom1729:
```python
# Let it like this wainting for toml depedency in package contrl
# https://github.com/wbond/package_control_channel/pull/7298
import sys

sys.path.append(os.path.abspath(os.path.dirname(__file__)))
import toml
```

Can this be removed now that the dependency exists?

jgirardet:
> Can this be removed now that the dependency exists?
there is a problem with the toml dependency. I do not know why. @FichteFoll confirmed it in https://github.com/wbond/package_control_channel/pull/7343

Thom1729:
Is that still not working?

wbond:
It would appear @jgirardet hasn't created any semver tags yet: https://github.com/jgirardet/sublime-toml/tags. Once that is created, Package Control will include it in the channel.

jgirardet:
my bad, tag is added and  I fixed what @Thom1729 requested in poetry

PR #7348, opened 2018-11-07
Add PyenvEnv package

jkr78:
Finds pyenv python version per view/project and allows to set env variable to it.

Difference from other (similar) packages:
- it is pyenv specific
- supports pyenv-virtualenv
- can be used with .python-version file

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -2558,6 +2558,17 @@
 			]
 		},
 		{
+			"name": "PyenvEnv",
+			"details": "https://github.com/jkr78/PyenvEnv",
+			"labels": ["python", "pyenv", "env", "virtualenv"],
+			"releases": [
+				{
+					"sublime_text": ">3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Pygame Completions",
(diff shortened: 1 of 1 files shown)


Thom1729:
The package is for Sublime 3 only, so you don't need Python 2 compatibility. The plugin imports `six`, which is a 2/3 compatibility library that is not built in. There is a Sublime dependency for it, and you could use it, but if you don't need Python 2 compatibility you'd be better off not using `six`.

jkr78:
> The package is for Sublime 3 only, so you don't need Python 2 compatibility. The plugin imports `six`, which is a 2/3 compatibility library that is not built in. There is a Sublime dependency for it, and you could use it, but if you don't need Python 2 compatibility you'd be better off not using `six`.

Updated in tag: 0.0.2

jkr78:
> In `valid_file` and `valid_syntax`, you should pass in the view rather than using `sublime.active_window().active_view()`. In fact, I think that most times you're using `sublime.active_window()`, you can pass in a window or view instead. This avoids possible race conditions.

Updated on tag: 0.0.3

PR #7349, opened 2018-11-09
Add Googler plugin

ashishg-qburst:
A plugin for easy googling, I made this plugin to fix design flaws in similar plugins, and search Google in a more profitable way (targeted at increasing programmer productivity).

Similar Plugins: 

https://github.com/nwjlyons/google-search

* This one allows you to google for any selection or a custom query, but most of the time you want to edit your selection before search, which is not supported. 

This means you will have to first search with the selection, then modify your query in Google and search again. This gets tiresome quickly and you stop using the plugin.

https://github.com/kemayo/sublime-text-2-goto-documentation

* File type based doc searching is available, but yet again queries are not editable.

Both allow prefixing and suffixing but only one prefix or suffix can be configured. A better idea is to allow the user to configure prefixes and suffixes for each file type. This is especially useful to people working on multiple projects across languages.

Features of the new proposed plugin:

* Scope based prefixing and suffixing available
* Queries are editable before getting to Google
* Scope-matched prefixes/suffixes can be turned off entirely
* Triggering the command without a selection gets you straight to the Google homepage
* Better documentation and active maintenance

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -1334,6 +1334,18 @@
 			]
 		},
 		{
+			"name": "Googler",
+			"author": "ashishg-qburst",
+			"details": "https://github.com/ashishg-qburst/googler",
+			"labels": ["google", "google-search"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"details": "https://github.com/SublimeText/GoogleTesting",
 			"releases": [
 				{


Thom1729:
Does this do anything that [Google Search](https://packagecontrol.io/packages/Google%20Search) doesn't?

ashishg-qburst:
> Does this do anything that [Google Search](https://packagecontrol.io/packages/Google%20Search) doesn't?

I have listed the changes and improvements this plugin makes over similar plugins in the original PR comment.

Quoting again about the Google Search plugin: 

> https://github.com/nwjlyons/google-search
>
> * This one allows you to google for any selection or a custom query, but most of the time you want to edit your selection before search, which is not supported.
>
> This means you will have to first search with the selection, then modify your query in Google and search again. This gets tiresome quickly and you stop using the plugin.

The proposed plugin enables you to modify queries in ST before going to Google. 

Also added is scope based prefixing and suffixing of queries. Google Search allows you to have only one prefix or suffix. This is not very helpful in most situations. It makes more sense to have prefixes or suffixes scoped to your file type. Again everything is editable before you actually initiate the search.

bjorn:
What about opening a pull request to the Google Search plugin with your enhancements?

PR #7350, opened 2018-11-09
[YAMLJSON] add plugin to convert yaml to json and back

thomasmeeus:
Hi,

this is a plugin which allows converting json snippets to yaml and back via the Command Palette. I use this functionality a lot when working with Openshift objects.

It's currently just a working draft. Additional features can be added when required.

### DIFF
--- a/repository/y.json
+++ b/repository/y.json
@@ -22,6 +22,17 @@
 			]
 		},
 		{
+			"name": "YAMLJSON",
+			"details": "https://github.com/thomasmeeus/yamljson",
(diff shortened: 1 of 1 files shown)


FichteFoll:
https://packagecontrol.io/packages/SerializedDataConverter does this already (and has plist support). It also has some convenience stuff like loading yaml mappings into ordered dicts and similar.

thomasmeeus:
Damn, I can’t believe I overlooked that. Thanks! Closing this PR as is now is obsolete.

PR #7351, opened 2018-11-11
Add ExecInput package

mheinzler:
This adds the ExecInput package that allows sending input to a process run through the standard build system.

It differs from other packages like [Sublime Input](https://packagecontrol.io/packages/Sublime%20Input) in that it doesn't require any changes to a project's build system or a file's source code.

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -1466,6 +1466,17 @@
 			]
 		},
 		{
+			"name": "ExecInput",
+			"details": "https://github.com/mheinzler/ExecInput",
+			"labels": ["input", "stdin", "build system"],
+			"releases": [
+				{
+					"sublime_text": ">=3176",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "exercism",
 			"previous_names": ["STexercism"],
 			"details": "https://github.com/thosgood/STexercism",


mheinzler:
Sorry for the spamming, that's my first package :smile:.
Should I get rid of this warning by changing the key binding? I already added the context to not overwrite the default key bindings. With that I don't see a problem staying with it.

Thom1729:
Patching the builtin command is probably not the best. Instead, I'd subclass `Default.exec.ExecCommand` and use `EventListener.on_window_command` to intercept calls to the original and redirect them to your subclass.

Also, instead of storing the running command object as a property of `window`, it would be better to store it elsewhere. I suggest a global dict from window IDs to commands. The reference should probably be deleted when the build finishes running.

mheinzler:
Thanks, I didn't know about `on_window_command`. I will prepare a better version in the next few days.

mheinzler:
I tried `on_window_command` but it seems it doesn't help in this case:
1. It doesn't trigger for the `exec` command, but only for the `build` command. I don't see a way to get any kind of access from here.
2. Even if `exec` could be intercepted here, there would still be an issue with other build systems that derive from `ExecCommand` (like [ANSIescape](https://github.com/aziz/SublimeANSI)). They would require special handling which they do not with the current design.

I totally agree with you that patching is a bad idea, but right now I don't think there is better way (without creating another build system or requiring special configuration by the user).

Do you know if there would be a way to integrate such functionality into Sublime Text itself? That would certainly be the best.

FichteFoll:
Your package won't work on ST2 because `Window.active_panel()` is a recent addition. You should clarify that in the `sublime_text` version selector.
(Also `exec` is a keyword in Python 2, so this won't even load due to the import.)

I'm okay with monkeypatching the default ExecCommand because I don't really see another feasible alternative. You'll only run into problems if multiple packages attempt to do so at the same time, in which case the user will have to figure out that having two packages that hack on a single command isn't the best of ideas. It's mentioned in the readme after all.

mheinzler:
> Your package won't work on ST2 because `Window.active_panel()` is a recent addition. You should clarify that in the `sublime_text` version selector.

Right. I pinned the version to 3176 from last year which should be fine.

PR #7352, opened 2018-11-14
Add ShowDefinitionEx

jk4837:
A sublime plugin for visually showing definition, not only symbol location, but also symbol name and parameter.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1514,6 +1514,17 @@
 			]
 		},
 		{
+			"name": "ShowDefinitionEx",
+			"description": "Visually showing definition",
+			"details": "https://github.com/jk4837/ShowDefinitionEx",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"details": "https://github.com/grapegravity/ShowEncoding",
 			"releases": [
 				{


jk4837:
Change binding key.

jk4837:
Change default keymap and add a new tag.

jk4837:
@Thom1729 

> You should add a license.

Added now!

> Is it really necessary to have a dedicated keybind to turn this on and off?

Yes, my plugin only show the definition, when I need the reference,  it will need to turn this off and turn builtin on, or the two popup will affect each other.

>One of these lines of code shouldn't be there. Really, DEBUG should probably be a setting.

Now it's a setting option.

>	global_settings = sublime.active_window().active_view().settings()
This seems backward. Also, why would you want to store in a global variable the settings of the view that was active when the plugin was loaded?

Because using `hide_view.settings().get('syntax', {'extension': hide_view_ex})` can't get correct syntax, and I do not want to get it every time, so just one time when the plugin was loaded.

>Why not use the builtin html module?

Use it now, I am a beginner of Python, don't know there's a builtin html module can use.

>def symplify_path(path):
	s = path.split('/');
	count = 2
	if count*2 < len(s):
		return '/'.join(s[0:count] + ['..'] + s[-count:])
	return path
Will this work correctly on Windows?

Yes, I am working on Windows.

Other bugs are fix too, thanks for the review!

jk4837:
@Thom1729 
Hi, I have been waiting for a month.
Is there anything need to edit?

PR #7363, opened 2018-11-22
Schnapsum

airqz:
It allows you to Insert Schnapsum like Lorem Ipsum in the editor via menu items or keyboard shortcut.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -433,10 +433,11 @@
 		{
 			"name": "Schnapsum",
 			"details": "https://github.com/airqz/schnapsum-sublime-text",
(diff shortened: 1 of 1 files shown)


airqz:
add tag

airqz:
add tag with git

Thom1729:
- You should add a license.
- Is `re` used anywhere?
- Why do you call `self.view.end_edit(edit)`.
- You might declare `words` a  constant outside the command.

airqz:
thank you for your proofreading and review

airqz:
Change invalid key

PR #7365, opened 2018-11-23
Added HOL package

JamesShaker:
HOL (or SublimeHOL on github) provides a development environment for working with the Higher Order Logic Interactive Theorem Proving system (https://hol-theorem-prover.org/). It does this providing a syntax, snippets, as well as an ANSI colourised REPL configured to correctly interact with the HOL system.

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -747,6 +747,16 @@
 			]
 		},
 		{
+			"name": "HOL",
+			"details": "https://github.com/HOL-Theorem-Prover/SublimeHOL",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Hologram Doc Snippet",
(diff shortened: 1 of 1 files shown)


JamesShaker:
Just wondering if there are any issues preventing the acceptance of this pull request?

Thom1729:
Why `.no-sublime-package`?

Is Sublime 2 compatibility important?

Python regexps are already cached, so I don't think there's any benefit to maintaining your own regexp cache.

The `HOL_ANSI_COLORS_CHANGE` listener is going to rebuild files every time the settings change. It might be a good idea to check whether the relevant settings have actually changed. ([sublime_lib.SettingsDict.subscribe](https://sublimetext.github.io/sublime_lib/#sublime_lib.SettingsDict.subscribe) might be handy.)

What is `HolRunExistingWindowCommandCommand` doing?

Thom1729:
@JamesShaker ping.

PR #7366, opened 2018-11-23
add donation links for Sass and LESS

braver:
adds Paypal donation links for Sass and LESS packages.

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -686,7 +686,7 @@
 			"labels": ["language syntax"],
 			"releases": [
 				{
-					"sublime_text": ">3092",
+					"sublime_text": ">=3092",
 					"tags": true
 				}
 			]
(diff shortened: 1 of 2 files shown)


Thom1729:
Did you mean to change the LESS version compatibility?

braver:
@Thom1729 Yes, it should have been done earlier because it changed to the new sublime-syntax format. Sorry, should have added a comment here as well.

FichteFoll:
FYI you still have https://sublime.wbond.net/packages/LESS in the repo's url.

braver:
Yeah, I've got enough access to manage the codebase and releases, but that part of the repo setup is outside my jurisdiction. Also don't care enough to do something about it, it's like a neat little indicator that this package has been around for that long.

PR #7367, opened 2018-11-23
Add gerbus color scheme

gerbus:
A mostly cool sublime color scheme. I looked at other color schemes and I didn't quite find any I wanted. Mine resembles the Pastel On Dark scheme that codeanywhere.com offers.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2518,17 +2518,17 @@
 				}
 			]
 		},
-                {
-                        "name": "Color Scheme - Gerbus",
-                        "details": "https://github.com/gerbus/sublime-color-scheme",
-                        "labels": ["color scheme"],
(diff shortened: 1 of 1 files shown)


Thom1729:
You can ignore that message. The bot needs to be updated to notice `.sublime-color-scheme` files. However, you should declare support for `"sublime_text": ">= 3152",`. You should also add a license.

gerbus:
> You can ignore that message. The bot needs to be updated to notice `.sublime-color-scheme` files. However, you should declare support for `"sublime_text": ">= 3152",`. You should also add a license.

Done.

Thom1729:
Regarding the compatibility note, as of the latest dev build, the core JavaScript syntax does highlight class property arrow functions.

PR #7371, opened 2018-11-26
Add Keras package

MartinThoma:
The Keras package provides snippets for creating MLP and CNN (potentially more in future)

### DIFF
--- a/repository/k.json
+++ b/repository/k.json
@@ -122,6 +122,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


Thom1729:
@MartinThoma ping

MartinThoma:
@Thom1729 I've added a note about the context (Python / Machine learning). I don't think it makes much sense to write more, because it is super well known in this context. If people don't know it, they don't need it.

PR #7372, opened 2018-11-27
Rename sublime-bats package

sptndc:

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -218,6 +218,7 @@
 		},
 		{
 			"name": "Bats (Bash Automated Testing System)",
+			"previous_names": ["Bats"],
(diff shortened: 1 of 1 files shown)


FichteFoll:
There was never a package named "Bats" in the channel.

Did you intend to rename your package *into* "Bats"? If so, the order is reversed.

FichteFoll:
Yeah, you need to swap the two names then. `name` will be the new name while `previous_names` is used to rename a package if that old name is found.

sptndc:
So, I have to change the `name` not the `previous name`?

sptndc:
@FichteFoll Thanks for your review :+1:

PR #7376, opened 2018-11-30
Add OpenSmerge

FloydaGithub:
This is a package open Sublime Merge in Sublime Text.
Matches all git repositories based on the path to the current file and the folders in project.


FichteFoll:
With the latest release of [3181](https://www.sublimetext.com/3dev), I don't think this is needed anymore, is it?

Though, I guess since this operates on the current file path as well it is slightly superior to the current implementation. I like that it also provides a list of matching repos when there are multiple.

As a suggestion, you might want to try resolving the file's path if it is behind a symlink, too.

Thom1729:
@FloydaGithub ping

FichteFoll:
I think this functionality has been sufficiently replaced by core updates. Closing as a timeout.

PR #7377, opened 2018-11-30
Added package Open Sublime Merge

dusk125:
Open Sublime Merge is a simple plugin that opens (or reveals) a Sublime Merge window from inside of Sublime Text 3. This allows for users to quickly pull up the git client with the push of a button.

### DIFF
--- a/repository/o.json
+++ b/repository/o.json
@@ -602,6 +602,18 @@
 			]
 		},
 		{
+			"name": "Open Sublime Merge",
+			"details": "https://github.com/dusk125/sublime-osmerge",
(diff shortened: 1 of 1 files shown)


FichteFoll:
With the latest release of [3181](https://www.sublimetext.com/3dev), I don't think this is needed anymore, is it?

I would also strongly suggest going without a default binding and not submitting your workspace file to source control.
Your Usage section partly describes a different package.

dusk125:
I would agree, I hadn't seen this added yet and the idea just popped into my head. I'm going to close this PR.

Also thank you for the suggestions, I'll keep them in my for my next project.

PR #7380, opened 2018-12-01
Add HISE Script

Levitanus:
is a little project trying to integrate HISE (hise.audio, https://github.com/christophhart/HISE) JS API into the Sublime Text
Features: new JS-based syntax, autogenerated API completions

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -630,6 +630,16 @@
 			]
 		},
 		{
+			"name": "HISE Script",
+			"details": "https://github.com/Levitanus/HISEScript-Sublime",
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "HIVE Log File",
 			"details": "https://github.com/digitalinblue/SublimeHIVELogHighlighter",
 			"releases": [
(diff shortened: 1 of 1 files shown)


Thom1729:
There are several extraneous changes in the channel file.

You're modifying `sys.path` in HISEScript.py. This is strongly discouraged, because `sys.path` is a shared resource. Instead, use relative imports: `import .snippets`.

You don't need to watch for the `supress_completions` setting to change. When the setting changes, the `Settings` object will be updated internally, so the next time you get `settings['supress_completions']` it will return the correct value. If you do need to watch the settings object for some reason, you should make sure that the listener is eventually removed.

Rather than doing path manipulation with strings, you should use `os.path` (or the very handy [`pathlib` dependency](https://github.com/packagecontrol/pathlib).

Levitanus:
Thanks for the review!
updated repository and version tag

Thom1729:
There are extra changes in the channel file. It looks like you may be using some sort of auto-formatter.

Levitanus:
reset to initial branch commit and updated repo one more time

Levitanus:
Can't understand, what is wrong with the tag...

Levitanus:
license added

PR #7381, opened 2018-12-05
Add AnonymousFile plugin

nchauhan890:
Sublime Text 3 plugin to quickly create already-saved files to test code snippets.

*Excerpt from README*:

> **Have you ever thought of a brilliant idea and wanted to test out your code as fast as possible? You'd go through making a new file, writing the code out, saving the file somewhere, running it and then deleting it if the code failed. Now, with Anonymous File, you can quicken this workflow by creating pre-saved files which delete themselves upon closure.**

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -1073,6 +1073,17 @@
 			]
 		},
 		{
+			"name": "AnonymousFile",
+			"details": "https://github.com/nchauhan890/anonymous-file",
+			"labels": ["file creation", "testing"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Ansible",
 			"details": "https://github.com/clifford-github/sublime-ansible",
 			"labels": ["language syntax"],


Thom1729:
What's the reason for `sys.exit(1)`? At best, it won't do anything because of the `raise` before it. At worst, I tried running `sys.exit(1)` in the Sublime console out of morbid curiosity, and it crashed.

You don't need to watch for settings changes just to update variables. When you call `settings.get`, you should always get the latest value.

You might want to leverage the [`tempfile` module](https://docs.python.org/3/library/tempfile.html). This will provide automatic names and so forth.

nchauhan890:
Yes, you're absolutely right about the `sys.exit` call; I've removed it. With the setting changes, the reason I've put it there is to handle 'type checking' and 'normalisation'. The settings values such as `dir` are passed through `os.path.normpath` and any variables like `${packages}` are expanded. If there is a bad input, a message is also displayed for a few seconds in the status bar. By watching settings changes, the messages appear on file save which is beneficial for users' sake. Having said that, the settings are fairly basic and the default probably won't need to be changed 

With regards to `tempfile`, I had a look into it and just couldn't get it to work. I believe what I have made has all the functionality I need like random file names and removing files. The problem I have with `tempfile` is that once the file handle is closed, the file is removed (If I'm not mistaken), which means the plugin would need to have a list of file objects open and isn't really ideal.

Thom1729:
> The problem I have with tempfile is that once the file handle is closed, the file is removed (If I'm not mistaken), which means the plugin would need to have a list of file objects open and isn't really ideal.

You can prevent this by using `NamedTemporaryFile` and passing `delete=False`.

PR #7385, opened 2018-12-07
added intellekt plugin

tushortz:
provides intellisense, code completion and documentation for various languages like java, python, ruby etc,


Thom1729:
Why add to `sys.path`?

tushortz:
> Why add to `sys.path`?

this is to allow external python packages to be used

Thom1729:
`sys.path` is shared among the entire plugin runtime, so adding entries to it is undesirable.

Can you explain how this is being used in the context of the package? There may be alternatives that don't modify shared global state.

tushortz:
> `sys.path` is shared among the entire plugin runtime, so adding entries to it is undesirable.
> 
> Can you explain how this is being used in the context of the package? There may be alternatives that don't modify shared global state.

The current `sys.path` does not provide access to external python modules in the case where there are more than one python versions. Adding the path is optional as there is an `if` statement to check if a path is provided in the `sublime-settings` file.

Adding the extra path allows the package to search for external modules in the specified path

PR #7386, opened 2018-12-08
added PassManager

degami:
added PassManager, a plugin which gets passwords from your passwordstore ( https://www.passwordstore.org )  and inserts the password value into the current view / clipboard


degami:
fixed requested change

degami:
added tag version on repository

PR #7387, opened 2018-12-08
added PassManager

degami:
added PassManager, a plugin which gets passwords from your passwordstore ( https://www.passwordstore.org ) and inserts the password value into the current view / clipboard

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -418,6 +418,17 @@
 			]
 		},
 		{
+			"name": "Pass Manager",
+			"details": "https://github.com/degami/PassManager",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": ["osx", "linux"],
+					"branch": "master"
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


Thom1729:
Rather than using `sublime.active_window()`, it looks like you should be able to use a known window. For example:

```python
class RefreshBrowserViewCommand(sublime_plugin.WindowCommand):
    def run(self):
        window = sublime.active_window()
```

You can use `self.window` instead. This avoids race conditions.

In `FillBrowserViewCommand`, you can use `self.view.window()`.

To get a setting with a default value, instead of `settings.get(key) or default`, you can use `settings.get(key, default)`. This will behave differently if `settings.get(key)` is falsy.

Instead of `WriteViewCommand`, you can use the built-in `view.run_command('append', {'characters': 'whatever'})`.

You could simplify some of your file path logic using the `pathlib` dependency. Totally optional.

degami:
updated with the advices I thought useful

PR #7389, opened 2018-12-08
Fixed my dependencies URLs names not on lowercase convention

evandrocoan:
I bad named my dependencies with PascalCase, this fix it and updates their URLs.


Thom1729:
Does `previous_names` work for dependencies?

hadisfr:
Any news here, after months? :-"

PR #7391, opened 2018-12-09
Added GROQ Syntax Highlighting

alevroub:


Thom1729:
You should add a license.

I'm not sure that `groq.sublime-settings` serves a purpose, because you've already specified the extension in the syntax.

alevroub:
> I'm not sure that `groq.sublime-settings` serves a purpose, because you've already specified the extension in the syntax.

Thanks for clearing that up. A license is now included.

PR #7393, opened 2018-12-10
add cpp-classhelper-plugin

pr0grammr:

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -59,7 +59,10 @@
 		},
 		{
 			"name": "C++ Classhelper",
-			"details": "https://github.com/sawzcode/cppclasshelper-sublime-text-plugin",
+			"details": "https://github.com/pr0grammr/cppclasshelper-sublime-text-plugin",
+			"author": "Fabian Schilf",
+			"labels": ["c++", "automation", "utilities", "file creation"],
+			"description": "Create C++ class with Headerfile",
 			"releases": [
 				{
 					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


Thom1729:
Are there other packages that do similar things?

Thom1729:
It looks like the file path manipulation might be problematic on Windows, or if the package is zipped (as it would be for most users).

Instead of using `open`, you will need to use `sublime.load_resource`, and pass it a resource path (e.g. `Packages/C++ Classhelper/templates/headerfile.template`) instead of a file path. This will correctly load the template whether it is a true file under `sublime.packages_path()` or zipped under `sublime.installed_packages_path()`.

Alternatively (totally optional), the [`sublime_lib` dependency](https://github.com/SublimeText/sublime_lib) provides a [ResourcePath](https://sublimetext.github.io/sublime_lib/#sublime_lib.ResourcePath) class modeled after `pathlib` that may make this easier. Examples:

```python
from sublime_lib import ResourcePath

class CreateCppClassCommand(sublime_plugin.WindowCommand):
    '''
    class for create a C++ class with a sourcefile and headerfile
    '''
    def run(self, **kwargs):
        package_dir = ResourcePath.from_file_path(__file__).parent
        template_dir = package_dir / 'templates'

        ...

class Template:

    def load(self, resource_path):
        self._filename = str(resource_path)
        self._template = resource_path.read_text()
```

Disclosure: I'm the primary author of `sublime_lib`.

Thom1729:
FYI, rather than `self.template_dir = "{}/{}/".format(self.package_dir, self.template_dir_name)` you can `self.package_dir / self.template_dir_name` and get a ResourcePath.

PR #7398, opened 2018-12-11
Register ScalafmtEnhanced, a code formatting plugin for Scala based on scalafmt.

nitrotm:
The goal of this plugin is to bring automatic Scala code formatting closer to what JsPrettier is doing. So this package is a totally new take on formatting Scala code using scalafmt:

- Custom formatting configuration per-project (.scalafmt.conf), with a lookup based on project's directory tree.
- Supports manual formatting of the whole file or one or multiple selected regions.
- Supports automatic 'format on save'.
- Error reporting is included, with status messages and a dedicated html popup showing the eventual errors logs during code reformatting.
- This package manages the background process better (nailgun/scalafmt), especially when the plugin is removed or sublime text is closed, without leaving a hanging process in the system.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -327,17 +327,17 @@
 				}
 			]
 		},
-    {
-      "name": "ScalafmtEnhanced",
-      "details": "https://github.com/nitrotm/st3-scalafmt",
-      "labels": ["scala", "format", "formatting", "formatter", "scalafmt"],
-      "releases": [
-        {
-          "sublime_text": ">=3000",
-          "tags": true
-        }
-      ]
-    },
+		{
+			"name": "ScalafmtEnhanced",
+			"details": "https://github.com/nitrotm/st3-scalafmt",
+			"labels": ["scala", "format", "formatting", "formatter", "scalafmt"],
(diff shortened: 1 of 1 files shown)


Thom1729:
Why is `.no-sublime-package` needed? Can it be avoided?

nitrotm:
The plugin needs to spawn a separated python process (from a script file in the package). Otherwise, I could eventually create a temporary file with the content somewhere (like /tmp in linux). But then the file needs to be removed after sublime shutdown or the plugin is removed (otherwise there will be garbage left on the file-system).

Thom1729:
You can get around this by executing the text of the script:

```python
python_path = sys.executable or 'python3'
code = sublime.load_resource('Packages/ScalafmtEnhanced/scripts/nailgun.py')

subprocess.Popen(
    [python_path, '-c', code],
    ...
)
```

nitrotm:
Ok, good idea. I used a stdin pipe.

Thom1729:
FYI, you shouldn't need the `__future__` imports if you don't need ST2 compatibility.

PR #7401, opened 2018-12-17
Add RubocopDaemon

georf:

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -1871,6 +1871,17 @@
 			]
 		},
 		{
+			"name": "RubocopDaemon",
+			"details": "https://github.com/Lichtbit/sublime_rubocop_daemon",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"platforms": ["osx", "linux"],
+					"tags": "v"
+				}
+			]
+		},
+		{
 			"name": "Ruby Block Converter",
 			"details": "https://github.com/irohiroki/RubyBlockConverter",
 			"labels": ["formatting"],


Thom1729:
Why create a new package? Could your improvements be integrated into the original?

georf:
Sorry, I had a few days off.

The original plugin offers a completely different approach. You can't use both at the same time. Either you use the normal Rubocop process call or you use the daemon call.

Thom1729:
- What is the purpose of `RubocopDaemonEventListener.listener_instance`?
- To test whether a file is a Ruby file, it's probably best to run `view. match_selector(0, 'source.ruby')` instead of checking syntax file names.

Thom1729:
@georf ping.

georf:
Ohh.. I will take a look.

georf:
>     * What is the purpose of `RubocopDaemonEventListener.listener_instance`?

I removed this part of the code.

>     * To test whether a file is a Ruby file, it's probably best to run `view. match_selector(0, 'source.ruby')` instead of checking syntax file names.

Thank you for your hint. Now I use this method to detect the file is a ruby file.

I created a new version `v1.0.1`. Did I do this on the correct way?

Thom1729:
In `daemon_file()`, you never close the file. It would be easiest to use a `with` block here:

```py
try:
    with open(file, "r") as stream:
        return stream.read()
except OSError:
    # whatever
```

georf:
Thanks for your advise. I updated the repo to version 1.0.2

PR #7402, opened 2018-12-17
Added Alabaster Color Scheme

tonsky:
A minimalist color scheme

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -525,6 +525,18 @@
 			]
 		},
 		{
+			"name": "Alabaster Color Scheme",
+			"details": "https://github.com/tonsky/sublime-scheme-alabaster",
+			"donate": "https://www.patreon.com/tonsky",
+			"labels": ["color scheme"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Alan",
 			"details": "https://github.com/M-industries/AlanForSublime",
 			"labels": ["language syntax"],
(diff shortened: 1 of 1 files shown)


Thom1729:
> I guess the bot is running off the old version of st_package_reviewer that didn’t have support for *.sublime-color-scheme yet

Yep.

You can remove `empty.py`.

What is the purpose of the "playground" files?

tonsky:
> You can remove empty.py.

Done.

> What is the purpose of the "playground" files?

Those are the files I test my color scheme on. Any way to exclude them from the package? They are not needed in the installed package.

Levitanus:
> They are not needed in the installed package.

gitignore?

tonsky:
I’d prefer to keep them in the repo as they are part of the development process. Same as pictures that are part of presentation but not needed in the actual package. Maybe something like whitelist.txt or blacklist.txt to explicitly include/exclude files. I could even implement this function if needed, if you point me to the right repo that builds packages

wbond:
@tonsky The area in the code where something like whitelisting or blacklisting would happen would be: https://github.com/wbond/package_control/blob/be02c9b2efa406ed9401f22d38ff477b47f5e211/package_control/package_manager.py#L1293-L1393.

However, unfortunately I'm not sure when the next release will happen as I currently have very little time for open source work, and Package Control releases require a bit of testing since it is a self-updating package, and thus can brick itself.

tonsky:
Ok I removed unnecessary files. Please merge is everything is fine

wbond:
Sorry for the noise here - was just testing the updated st_package_reviewer on packagecontrol.io.

FichteFoll:
FYI, you can disable exporting files in zip-/tarballs using `.gitattributes` since Package Control downloads the zipball of the tag.

tonsky:
thx @FichteFoll I’ll try it out!

PR #7404, opened 2018-12-24
Add IrView package

guilhermeleobas:
A package to generate control flow graphs of LLVM Intermediate Representation directly from Sublime Text.

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -1111,6 +1111,16 @@
 			]
 		},
 		{
+			"name": "IrView",
+			"details": "https://github.com/guilhermeleobas/IrView",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


guilhermeleobas:
Just updated the package

Thom1729:
If a command works on a particular view, it should be a `TextCommand`, not a `WindowCommand`, so that you can use `self.view` instead of `self.window.active_window()`.

guilhermeleobas:
Thanks for the tip. Just updated the plugin.

Thom1729:
The hardcoded `/tmp/` doesn't look cross-platform. Can you use the builtin `tempfile` and `shutil` modules?

guilhermeleobas:
> The hardcoded `/tmp/` doesn't look cross-platform. Can you use the builtin `tempfile` and `shutil` modules?

Thanks for the tip. I have updated the package to be less platform dependent.

PR #7407, opened 2018-12-30
zero-dark new theme

jrvieira:
This colour scheme is now part of a full theme, and the theme part of a family of themes.


wbond:
Yeah I think I reversed the order in the merge

jrvieira:
I created a new pull request in order to solves the issue.

PR #7411, opened 2019-01-01
Add windsor

pranaygp:
[Windsor](https://windsor.io) is a project we're working on to to build a smart, context aware code "sidekick" of sorts in the form of an app that sita next to your text editor and provides a plugin architecture to support deep editor and file system integration, access to authenticated user services, collaboration and more.

You can find out a little bit about us from your [website](https://windsor.io) or feel free to reach out to me at [pranay@windsor.io](mailto:pranay@windsor.io) to get a demo of our product.

### DIFF
--- a/repository/w.json
+++ b/repository/w.json
@@ -529,6 +529,16 @@
 			]
 		},
 		{
+			"name": "Windsor",
+			"details": "https://github.com/windsorio/windsor-sublime",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"details": "https://github.com/SublimeText/WinMerge",
 			"releases": [
 				{


pranaygp:
@wbond wondering how long the process is for a package to get added. We're hoping to let users download our plugin from package control soon :)

Thom1729:
- `import sys` is unused.
- `TextCommand.run` takes args `self`, `edit`, `args`, but `WindsorEditDocumentCommand` seems to ignore the `edit` arg.
- In many places, the plugin is checking for the active view or window when there's already a contextually relevant view or window, or events are ignored when the view that triggered the event is not active when the handler is run. This seems odd to me. If this behavior is deliberate, then so be it.
- If there is an error opening the socket, then the lock will never be released. Instead, you should use the context manager protocol:

```python
with lock:
    if (ws is None):
		websocket.enableTrace(True)
		ws = websocket.WebSocketApp("ws://localhost:61952/",
			on_message = on_message,
			on_error = on_error,
			on_close = on_close)
		ws.on_open = on_open
		wst = threading.Thread(target=ws.run_forever)
		wst.start()
```

pranaygp:
Thanks for feedback @Thom1729. I've [updated the repo](https://github.com/windsorio/windsor-sublime/commit/99d41ea16c086f5cc35c568f9abe6c634ffc09e2)

To address some of your other comments:

> TextCommand.run takes args self, edit, args, but WindsorEditDocumentCommand seems to ignore the edit arg.

The arg is used but with the identifier being "group" instead of "edit":
```python
class WindsorEditDocumentCommand(sublime_plugin.TextCommand):
  def run(self, group, edits=[]):
    for edit in edits:
      region=sublime.Region(edit.get("start"), edit.get("end"))
      self.view.replace(group, region, edit.get("text"))
```
I've tested it to work just fine 👍 

> In many places, the plugin is checking for the active view or window when there's already a contextually relevant view or window, or events are ignored when the view that triggered the event is not active when the handler is run. This seems odd to me. If this behavior is deliberate, then so be it.

Yup, currently the Windsor daemon simply doesn't handle buffers for inactive files so we're dropping the messages from the sender (i.e. this sublime text plugin). Very soon we plan to cache inactive buffers on the Windsor daemon and handle these updates from the sublime text plugin in a way that would allow multiple, slightly different buffers for the same inactive file to all contribute to Windsor's representation. We wrote the event handlers in a way that might facilitate that change more easily soon.

PR #7414, opened 2019-01-02
Added CoverMe package

shauryachats:
CoverMe is a plugin to generate and view code coverage for any testing framework in any programming language.
There are many coverage plugins for Sublime but they are all specific to one programming language or testing framework.

CoverMe has the advantage of ease of extensibility - To view code coverage for a new testing framework, only the commands to be run to generate coverage information, and a parser to interpret that coverage information into a common format which can be interpreted and viewed by CoverMe is required to be implemented.

CoverMe also has the advantage of flexibility - There are some default coverage modes provided for each testing framework for each programming language, which can easily be overriden on a project-to-project basis by overriding the `settings` key in `.sublime-project` files.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3616,6 +3616,16 @@
 			]
 		},
 		{
+			"name": "CoverMe",
+			"details": "https://github.com/shauryachats/CoverMe",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


Thom1729:
You shouldn't modify `sys.path`. The path is shared among all packages. Use relative imports instead.

There are a bunch of stray `print()` calls in `python.py`.

shauryachats:
Incorporated comments. Please review.

PR #7415, opened 2019-01-03
Add NHP Syntax Highlight

hdodov:
Add syntax highlighting for the [NHP templating engine](https://nhpjs.com/) which is going to be officially released soon.


hdodov:
I added a semantic version tag in my repo. How do I trigger the bot again?

PR #7418, opened 2019-01-07
Please add my package into w.json

123581321345589:

### DIFF
--- a/repository/w.json
+++ b/repository/w.json
@@ -850,6 +850,16 @@
 			]
 		},
 		{
+			"name": "WorkTimer",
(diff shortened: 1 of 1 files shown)


Thom1729:
Is there any reason to use a custom `Data.json` file instead of a settings file? The custom implementation is a lot of code, and it looks like it won't work when the package is installed.

Surely it would be easier to use Python's built-in time formatting?

Inside a `WindowCommand`, you can refer to the window using `self.window`. You don't need to (and shouldn't) call `sublime.active_window()`.

Thom1729:
@123581321345589 ping

PR #7419, opened 2019-01-07
Add Formatter plugin

bitst0rm:
Formatter is a Sublime Text 3 plugin to beautify or minify source code.

Features:
  - Work offline


bitst0rm:
Remove key bindings

PR #7422, opened 2019-01-08
Add Formatter plugin

bitst0rm:
Formatter is a Sublime Text 3 plugin to beautify and minify source code.

Features:
  - Work offline

https://github.com/bitst0rm-pub/Formatter

@Thom1729 my previous commit #7421 has been accidentally updates by the pull bot, so it was auto empty. I turned it off now. Could you please review again, Thank you.


Thom1729:
Not sure what happened with the other issue. Doesn't look like I reviewed it.

- You shouldn't modify `sys.path` -- it's shared by all packages. There's a [`pyyaml` dependency](https://github.com/packagecontrol/pyyaml) that should solve your problem.
- Because the package is declared to be compatible only with ST3, it isn't necessary to check the version before running.
- In `RunFormatThread.run`, you should use `with self.command_lock:` instead of manually acquiring and releasing it. This will ensure that it is released if an error is raised.
- Each item in a `Selection` is a `Region`, so in `has_selection` you can just use `sel.empty()`:

```python
    def has_selection(self):
        return any(not sel.empty() for sel in self.view.sel())
```

- Reloading modules is surprisingly subtle. I'd recommend using [AutomaticPackageReloader](https://github.com/randy3k/AutomaticPackageReloader) instead of a hand-rolled solution.
- The logic in `setup_config` looks spot-on.
- I didn't review all of the individual formatters.
- By the time I got here, this PR was already closed. I am confused by what the bot is doing.

bitst0rm:
Thank you for your feedback!
The pull bot i was talking about was this one:
>a GitHub App built with probot that keeps your repository up-to-date with upstream changes via automated pull requests.
https://probot.github.io/apps/pull/
It reverts all of my pull requests with mainstream to keep my package_control_channel repo up to date, so i have to uninstall it. Hope my commit is ok yet.
I will look to improve my plugin in the next release.
Thank you!

PR #7423, opened 2019-01-08
Add Formatter plugin

bitst0rm:
Formatter is a Sublime Text 3 plugin to beautify and minify source code.

Features:
  - Work offline

https://github.com/bitst0rm-pub/Formatter

@Thom1729 my previous commit #7421 has been accidentally updates by the pull bot, so it was auto empty. I uninstalled it now. Could you please review again, Thank you.

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -1496,6 +1496,17 @@
 			]
 		},
 		{
+			"name": "Formatter",
+			"details": "https://github.com/bitst0rm-pub/Formatter",
+			"labels": ["formatting", "beautify", "minify", "css", "scss", "sass", "less", "html", "xml", "javascript", "json", "markdown", "perl", "php", "python", "ruby", "bash", "c", "c++", "c#", "d", "objc", "java"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Fortify Highlighter",
 			"details": "https://github.com/pwntester/FortifyHighlighter",
 			"releases": [


bitst0rm:
Sorry for the repeating the same pull requests, because i got troubles with this one:
https://probot.github.io/apps/pull/
It auto deleted all my commits in my pull requests here :(
Hope its ok yet.

Thom1729:
- You shouldn't modify `sys.path` -- it's shared by all packages. There's a [`pyyaml` dependency](https://github.com/packagecontrol/pyyaml) that should solve your problem.
- Because the package is declared to be compatible only with ST3, it isn't necessary to check the version before running.
- In `RunFormatThread.run`, you should use `with self.command_lock:` instead of manually acquiring and releasing it. This will ensure that it is released if an error is raised.
- Each item in a `Selection` is a `Region`, so in `has_selection` you can just use `sel.empty()`:

```python
    def has_selection(self):
        return any(not sel.empty() for sel in self.view.sel())
```

- Reloading modules is surprisingly subtle. I'd recommend using [AutomaticPackageReloader](https://github.com/randy3k/AutomaticPackageReloader) instead of a hand-rolled solution.
- The logic in `setup_config` looks spot-on.
- I didn't review all of the individual formatters.

bitst0rm:
@Thom1729 Thank you for your review! :)

> * You shouldn't modify `sys.path` -- it's shared by all packages. There's a [`pyyaml` dependency](https://github.com/packagecontrol/pyyaml) that should solve your problem.

The plugin should work offline, therefore i didn't use that dependency.

> * Because the package is declared to be compatible only with ST3, it isn't necessary to check the version before running.

Done. It has been improved.

> * In `RunFormatThread.run`, you should use `with self.command_lock:` instead of manually acquiring and releasing it. This will ensure that it is released if an error is raised.

Done. It has been improved.

> * Each item in a `Selection` is a `Region`, so in `has_selection` you can just use `sel.empty()`:
> 
> ```python
>     def has_selection(self):
>         return any(not sel.empty() for sel in self.view.sel())
> ```

Done. It has been improved.

> * Reloading modules is surprisingly subtle. I'd recommend using [AutomaticPackageReloader](https://github.com/randy3k/AutomaticPackageReloader) instead of a hand-rolled solution.

I did used it, but sometimes i turn it off.

> * The logic in `setup_config` looks spot-on.
> * I didn't review all of the individual formatters.

Thank you :)

Thom1729:
> The plugin should work offline, therefore i didn't use that dependency.

You can vendor the whole thing if you want, just use relative imports instead of modifying `sys.path`. It looks like the library already uses relative imports internally, so you won't have to rewrite them.

bitst0rm:
I use relative imports internally instead of modifying `sys.path` in my last commit.

Thom1729:
A couple of things I missed before:

- Ideally, you should remove the copied stuff from the User directory [when the plugin is uninstalled](https://packagecontrol.io/docs/events).
- I notice that the shared `os.environ` is modified. Can this be avoided?

bitst0rm:
> * Ideally, you should remove the copied stuff from the User directory [when the plugin is uninstalled](https://packagecontrol.io/docs/events).

Users would lost all the config that they work hart to adjust them, therefore i did not build such function in  `plugin_unloaded()`

> * I notice that the shared `os.environ` is modified. Can this be avoided?

This cannot be avoidable, since users could install plugins local outside the default paths. If i recall, SublimeLinter uses the same approach.

Thom1729:
> This cannot be avoidable, since users could install plugins local outside the default paths. If i recall, SublimeLinter uses the same approach.

I presume that you need the environment set up for the `Popen` call in `exec_cmd`. You can pass an `env` argument to `Popen`. This will apply those values to the environment of the child process without changing the environment of the parent.

bitst0rm:
I will look at this tomorrow, its too late now... have thanks & see you again

bitst0rm:
so i changed a few lines to use `Popen env`. The plugin does not touch on `sys.path` and `os.environ` anymore.

PR #7425, opened 2019-01-10
Add FlowTree plugin

maxwelljohn:
I made a package that automatically creates a diagram of your workflow while you're programming.

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -1244,6 +1244,16 @@
 			]
 		},
 		{
+			"name": "FlowTree",
+			"details": "https://github.com/maxwelljohn/FlowTree",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "FlowType",
 			"details": "https://github.com/Pegase745/sublime-flowtype",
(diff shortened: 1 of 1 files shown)


maxwelljohn:
I added a LICENSE file in the FlowTree repository.

Thom1729:
Is ST2 compatibility necessary? I'm concerned about monkeypatching `sublime`.

maxwelljohn:
Thanks for getting to this!  ST2 compatibility isn't necessary.  Let me know if there are further issues. 
 (Or make whatever changes you feel are appropriate, really.)

Thom1729:
Without ST2 compatibility, there's no need to use `Edit.py` (which hacks into Sublime's internals in a worrisome manner).

maxwelljohn:
Oh, I see what you're getting at.  This is my first Sublime plugin and I made use of `Edit.py` because I didn't think there was any other way to do what I was trying to do.  I took another look at the API documentation and I think I figured out the right way to do it.  `Edit.py` is no longer in the repo: https://github.com/maxwelljohn/FlowTree/commit/9878e57c86c0f55331eecd67f1b27f5178941d05

PR #7433, opened 2019-01-15
Add HugoBot plugin

ziishaned:
https://hugobot.io

### DIFF
--- a/channel.json
+++ b/channel.json
@@ -5,6 +5,7 @@
 		"https://bitbucket.org/jjones028/p4sublime/raw/tip/packages.json",
 		"https://bitbucket.org/klorenz/sublime_packages/raw/tip/packages.json",
 		"https://eoa-cdn.s3.amazonaws.com/extensions/packages.json",
+		"https://hugobot.io/artifacts/sublime/packages.json",
 		"https://packagecontrol.io/packages_2.json",
 		"https://packages.monokai.pro/packages.json",
 		"https://raw.githubusercontent.com/20Tauri/DoxyDoxygen/master/DoxyDoxygen.json",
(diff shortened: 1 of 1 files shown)


ziishaned:
Hi @FichteFoll,
Any update on the status of this PR?

Thom1729:
Why add a new channel entry rather than a new package entry?

Also, where is the package repository?

ziishaned:
@Thom1729 i added a new channel entry because this package is a private distribution.

Thom1729:
I'm not sure what you mean by “private distribution”.

The LICENSE file is empty. How is this package licensed?

What is `HugoBot/bin/hugobot-cli`? It's clearly a binary file, and the package entry is not platform-limited. Are you sure that it works on Mac OS, Windows, and Linux?

You're invoking that binary via `os.system()`. You really should use the `subprocess` module instead.

```py
def stop_hugobot_server():
    print("Stop Server")
```

What's the deal with `stop_hugobot_server()`?

The reloading logic appears to be copied verbatim or with only minor modification from GitGutter, but there is no license file for this code.

A line in `annotation.py` seems to be hardcoded with your home directory.

FichteFoll:
>I'm not sure what you mean by “private distribution”.

The package is not hosted on github or bitbucket, so it needs to be maintained in a custom repository file as we don't allow explicit versions and downloads that would have to be updated in the repo on every release.

@ziishaned, I suggest to host your package artifacts at a versioned location because of potential cache or delay issues with people downloading a newer artifact when the version information hasn't been updated by the crawler.

I don't think this *needs* a license for PC to download the file for you. It doesn't redistribute them, although it does create a local copy that any browser would. If there are snippets from GitGuttter involved, its license terms must be adhered to, however.

PR #7438, opened 2019-01-20
Add LuaFormatter

Koihik:
A Lua source code formatter


Thom1729:
It's probably not a good idea to shadow the built-in find command key bindings.

PR #7445, opened 2019-01-25
Add glot plugin

hyrious:
This plugin simply makes use of the api given by https://glot.io,
where users do code running and sharing.
It takes advantages of editing code with powerfull sublime text without openning a browser.

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -876,6 +876,17 @@
 			]
 		},
 		{
+			"name": "Glot",
+			"details": "https://github.com/hyrious/glot",
+			"labels": ["code sharing"],
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


Thom1729:
`from sublime import *` is very dangerous.

Instead of `sublime.active_window()` in a `TextCommand`, use `self.view.window()`.

It looks like the path logic (e.g. `folder = '{}/{}'.format(C.cache_path, id)`) won't work on Windows.

hyrious:
@Thom1729 just pushed changes.
1. expand `*` to `(...)`
2. `active_window()` to `self.view.window()`
3. my PC is Windows. Windows does support `/` instead of `\\` in path. If you've got any error message, tell me please.

PR #7446, opened 2019-01-25
Added SEPath package. 

eudorokhov:
This script allows you to quickly open file paths from the sublime text 3 in the same place or in the file manager.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1094,6 +1094,16 @@
 			]
 		},
 		{
+			"name": "SEPath",
+			"details": "https://github.com/eudorokhov/SEPath",
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


Thom1729:
There is a built-in window command that opens a directory in the OS file browser: `open_dir`, with `dir` and optional `file` arguments. You should probably use this command rather than a custom `open_in_file_manager`. (This also removes the need for platform-specific code.)

eudorokhov:
I used open_dir command. Thanks for help!)

Thom1729:
It looks like you still need to `import os.path`.

eudorokhov:
Of course! Fixed it :)

PR #7450, opened 2019-01-27
Added EasyFileHider plugin

AustinTruex68:
I was not able to find a similar package. It essentially just gives quick access to modifying the file_exclude_patterns node in user settings.

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -127,6 +127,17 @@
 			]
 		},
 		{
+			"name": "EasyFileHider",
+			"details": "https://github.com/AustinTruex68/EasyFileHider",
+			"labels": ["file", "hider", "easy", "clean"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Easygit",
 			"details": "https://github.com/Ivanca/easygit",
 			"releases": [


AustinTruex68:
I added the top level LICENSE file

Thom1729:
What is the purpose of the placeholder file? I'm nearly sure that there must be a simpler way to do it.

AustinTruex68:
@Thom1729 I put it there primarily to separate the settings, as to not overwrite anything already in my settings while I was developing, and liked the idea of having them separated.. It was something I threw together quickly to help out me and a couple co-workers on sublime. I am also not very familiar with the Sublime API and was not sure there was an easy way to fetch a specific setting.

Thom1729:
You can create a package-specific settings file: just name it (e.g.) `EasyFileHider.sublime-settings`. Users can override the defaults with a `EasyFileHider.sublime-settings` file in their `User` package.

You can load the settings via `sublime.load_settings('EasyFileHider.sublime-settings')`. Then, you can persist any changes with `sublime.save_settings('EasyFileHider.sublime-settings')`.

The problem with modifying a file inside the package is that it won't work if the package is installed. It's possible to work around this, but it's much easier to use settings for this sort of thing.

AustinTruex68:
Thanks for the quick response. I was not aware of that (like I said, slapped together a quick plugin for what I was currently doing, and it seemed to work fine on my machine)

I will go ahead and make those changes sometime this week (more likely this weekend) and post a new release to my repo.

Will I need to submit a new pull request, or would a comment on this thread suffice?

Thom1729:
A comment is fine.

AustinTruex68:
@Thom1729 Modified the package to use sublime settings rather than a json placeholder file. Ready for merge to package control.

AustinTruex68:
Just wondering on this @Thom1729

AustinTruex68:
Can anyone approve this... lol

PR #7451, opened 2019-01-29
Added Docs plugin

joaopjt:

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -1143,7 +1143,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"branch": "master"
+					"tags": true
 				}
 			]
 		},
--- a/repository/d.json
+++ b/repository/d.json
@@ -1137,17 +1137,6 @@
 			]
 		},
 		{
-			"name": "Docs",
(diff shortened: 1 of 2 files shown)


Thom1729:
Are there other packages that perform a similar function?

The name “Docs” isn't very descriptive. You should probably use something a little more specific.

Instead of using a custom `.docsconfig` file, why not use project settings?

Building paths with string concatenation is unlikely to work cross-platform.

joaopjt:
I've got changed the plugin name from 'Docs' to 'npmDocs'.

I`m following the exemple of 'waka-time-dashboard'.
Some people just open the folder, instead of creating a project

FichteFoll:
>Building paths with string concatenation is unlikely to work cross-platform.

Fortunately it will in this particular case, as all usages of `/` here are covered by Windows-internal API.

However:

- You should look into using instance variables instead of the `PACKAGES` global.
- You should check for `index == -1` before anything else in `on_done`.
- Remove your `.no-sublime-package` file.

Also, what do you think of "node_modules Readme" as the package name?

Thom1729:
@joaopjt ping

PR #7452, opened 2019-01-31
Adds RSpecImproved

mhenrixon:
I had some trouble with `Better RSpec` and wanted to create my own package for several different reasons. 

1. I want to add more functionality (I have some special requirements in terms of navigating between spec and implementation).
2. I want to learn about sublime plugin development.
3. I need something that can be maintained as I rely on some things to be smooth in my day to day development.
4. Better Rspec has very old configuration files which don't work with sublime text 3.

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -1776,6 +1776,19 @@
 			]
 		},
 		{
+			"name": "RSpecImproved",
+			"details": "https://github.com/mhenrixon/rspec_improved",
+			"labels": ["ruby", "rspec", "file navigation", "snippets", "language syntax"],
+			"readme": "https://raw.githubusercontent.com/mhenrixon/rspec_improved/master/README.md",
+			"donate": "https://paypal.me/mhenrixon",
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


Thom1729:
Is the original RSpec package still being maintained? Would it be possible to improve the existing package rather than adding a new one?

mhenrixon:
@Thom1729 that is a great suggestion, it says they don't maintain it anymore... Not sure if I want to contribute to something that isn't actively used. I'm open for suggestions though

rspec - packages - package control 2019-02-28 20-54-40

Thom1729:
It looks like the author of the RSpec package might be willing for someone else to take it over — it says to file an issue on the tracker. If the author agrees, they could transfer the repository and you could add your improvements to a new version. This would be preferable to having two packages, one of them obsolete.

mhenrixon:
@Thom1729 I completely agree but I am not confident enough with this yet. I'll just attempt to make my package work really well and then I'll see if I can maybe take over using my changes.

PR #7454, opened 2019-02-04
add complete_xml_tag

Neophyliam:
This package can convert non-standard close tag (</>) in an xml file to corresponding standard close tag.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3071,17 +3071,6 @@
 			]
 		},
 		{
-			"name": "complete_xml_tag",
-			"details": "https://github.com/Neophyliam/sublimetext_complete_xml_tag",
-			"labels": ["xml", "formatting"],
-			"releases": [
-				{
-					"sublime_text": ">=3000",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "CompletePHP",
 			"details": "https://github.com/desean1625/CompletePHP",
 			"author": "Desean1625",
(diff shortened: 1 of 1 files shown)


Thom1729:
Instead of matching the file extension, it's probably better to check `view.match_selector(0, 'text.xml')`. You might also make this extensible via settings so that users can use the package with other XML-based syntaxes that might use different scopes. Also, this check should ideally be in an `is_enabled()` method rather than in `run()`.

It's generally better to use standard built-in exceptions than to define your own. In this case, `ValueError` would be appropriate. Also, you shouldn't run `self.view.replace(edit, string_region, new_string)` if `complete_close_tag` failed.

Neophyliam:
@Thom1729 Thanks for your review and advice! I have made some improvements to this package.

* Put the code for checking file type to `is_enabled()` method.
* Check file type using `view.match_selector(point, selector)` method.
* Add a settings file to support other XML-based syntaxes.
* Use `ValueError` instead of my custom exception.
* Handle other types of exception so that `self.view.replace(edit, string_region, new_string)` won't be run if `complete_close_tag` failed.

Thank you :)

Thom1729:
I forgot to mention — package names are generally capitalized unless there's a strong reason not to.

Everything else looks fine.

Neophyliam:
I have made the package name capitalized and updated related files in this package. Hope there’s no other problem this time.

PR #7456, opened 2019-02-06
Add Run Task package

mgutski:
The Run Task package enables Sublime Text users to run custom workspace/project tasks, which can be conveniently configured through JSON file.

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -2140,6 +2140,17 @@
 			]
 		},
 		{
+			"name": "Run Task",
+			"details": "https://github.com/mgutski/sublime-run-task",
+			"labels": ["tasks", "task runner", "shell"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"details": "https://github.com/qfel/RunCommand",
 			"releases": [
 				{


Thom1729:
Why use a custom `tasks.json` file instead of using project files?

For that matter, what does this offer over project-specific build systems?

mgutski:
Hi @Thom1729, perhaps I haven't provided enough detail in the description of my pull request. The project-specific build systems that you mentioned is a great tool for defining custom build system(s) with your own project-specific build process. Unlike tasks, it is not designed for defining custom automated development tasks that you can quickly select with a single key binding whenever needed during the development, which is why I decided to create this plugin. The ability to quickly display a list of available project tasks and selecting the one you see fit is super convenient if you have tasks that are not a part of the build process, and yet they are important part of project development. For example, one may have scripts for generating various project components that are needed during the development way before the build system is started. That said, depending on the project needs, and every project has its own very specific needs, the built-in build systems may not fulfill all requirements. Frequently switching custom build systems through Tools | Build menu is far from optimal, and defining a whole set of custom key bindings for selecting every custom build system doesn't sound convenient too. Defining build system variants is also not an option as custom tasks may not be related to each other.  
The bottom line is that both tools are useful, but they serve a slightly different purpose. The build systems is a very specialized tool for customizing project build process, and Run Task plugin is a more general tool for defining custom tasks that can be freely selected from Sublime quick panel. In the end, I believe that developers should have the ability to choose the tool that is most suitable for their project.

Thom1729:
Makes sense.

Why use `tasks.json` instead of putting that same data in the project settings? It works either way, it just seems like an unnecessary extra moving part.

mgutski:
I assumed that an external plugin should work independently of Sublime-specific files, but if you think that it is fine to do so, then I will be happy to move the tasks data to `.sublime-project` file and simply extend it by adding a fourth top level section called `tasks`.

Does that sound good?

Thom1729:
It's fine to use custom keys in project files. If you do, a common convention is to prefix the key with the name of the package and a dot (e.g. `"Run Tasks.tasks"`).

Thom1729:
@mgutski ping

mgutski:
@Thom1729, I have updated the package's repository. The tasks definitions can now be placed in the `.sublime-project` file as described in the updated README file.

mgutski:
@Thom1729 ping

PR #7457, opened 2019-02-07
Add RustAutoImport

trishume:
My package provides a command to automatically add `use` statement imports in Rust. Basically a single fancy text macro with some smarts.

<https://github.com/trishume/SublimeRustAutoImport>

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -2201,6 +2201,16 @@
 			]
 		},
 		{
+			"name": "RustAutoImport",
+			"details": "https://github.com/trishume/SublimeRustAutoImport",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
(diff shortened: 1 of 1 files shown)


Thom1729:
What's the deal with `Default.history_list`?

trishume:
@Thom1729 I wanted the built in “jump back” (CTRL+minus on Mac) command to work to jump back to where you were before adding the import. The function to add locations to the jump history is only available importing like that.

Thom1729:
Why doesn't `import Default.history_list` work?

trishume:
@Thom1729 it doesn't import it as the path I thought it would. I changed it to use `from Default import history_list` instead and that works. I pushed `1.0.1` with that change.

PR #7458, opened 2019-02-08
Added OpenFilesLister and Fixed Sheller

bantya:
Added [`OpenFilesLister`](https://github.com/bantya/OpenFilesLister) and fixed [`Sheller`](https://github.com/bantya/Sheller)

* Sheller is now fixed to work with Windows only.

### DIFF
--- a/repository/o.json
+++ b/repository/o.json
@@ -631,6 +631,17 @@
 			]
 		},
 		{
+			"name": "OpenFilesLister",
+			"details": "https://github.com/bantya/OpenFilesLister",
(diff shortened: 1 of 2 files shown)


Thom1729:
Sheller seems to be a brand-new package, not an updated existing package. Can you clarify?

In a `WindowCommand`, you should use `self.window` rather than `sublime.active_window()`.

```py
        for sheet in self.window.sheets():
            if sheet and sheet.view():
```

Why not `for view in self.window.views()`?

bantya:
Yes, Sheller is a brand-new package!

PR #7461, opened 2019-02-09
Add SideBarMenuAdvanced Plugin

wisdman:

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1650,6 +1650,18 @@
 			]
 		},
 		{
+			"name": "SideBarMenuAdvanced",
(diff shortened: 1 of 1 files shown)


Thom1729:
Are there other packages that do the same thing?

wisdman:
Yes, there are [SideBarEnhancements](https://packagecontrol.io/packages/SideBarEnhancements) and [Side​Bar​Tools](https://packagecontrol.io/packages/SideBarTools).
**SideBarEnhancements** are too saturated with features, have poor performance and a lot of legacy code.
**SideBarTools** duplicates some existing Sublime Text functionality, which leads to duplication of items in the context menu.
My plugin is a reasonable compromise between existing solutions.

PR #7463, opened 2019-02-13
Updated c.json to include an entry for a new package: Copy, Paste and…

xvrqt:
… Increment

**Similar to:** Duplicate and Increment
**Reason for creation:** That package does not allow for you to input the number of times you want something duplicated resulting in a tedious copy and paste session that is scarcely more efficient than doing the entire process manually. 

This packages provides two functions: one that asks the user to input the number of times to duplicate and increment the selected text and one that duplicates and increments a single time. The previous version also did not have key-bindings.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3448,6 +3448,16 @@
 			]
 		},
 		{
+			"name": "Copy Paste Increment",
+			"details": "https://github.com/xvrqt/copy-paste-increment",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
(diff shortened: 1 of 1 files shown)


Thom1729:
Is it possible to improve the existing package rather than creating a new one?

xvrqt:
I started doing that and ended up changing almost every line so I figured I'd just make a new one.

I'll fork and send a PR request to D&I maintainer, but if he doesn't accept in some time frame, please consider this package instead.

xvrqt:
https://github.com/112KA/duplicate_and_increment/pull/2

PR #7466, opened 2019-02-14
Add LikeGit

poucotm:

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -902,6 +902,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


Thom1729:
Why are there `pyc` files? `LikeGit.py` is doing some pretty intrusive work to avoid loading the normal way.

Thom1729:
@poucotm ping

poucotm:
.pyc files are used to hide some codes like SFTP.

PR #7467, opened 2019-02-14
Add RAMA2200 Assembly

sbhat98:
Adds support for RAMA2200 Assembly syntax highlighting for Georgia Tech's CS 2200 class. This is a different language than LC2200 assembly which is also in use at Georgia Tech.

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -304,6 +304,16 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


Thom1729:
Please add a LICENSE.

Would it make sense to add this to the existing package rather than creating a new one?

Thom1729:
@sbhat98 ping

PR #7468, opened 2019-02-15
Add OpenContextPath

mheinzler:
This package gives the ability to quickly open paths within a text by using the context menu or a keyboard shortcut.

I couldn't find any other plugin which provides the same functionality.

### DIFF
--- a/repository/o.json
+++ b/repository/o.json
@@ -686,6 +686,22 @@
 			]
 		},
 		{
+			"name": "OpenContextPath",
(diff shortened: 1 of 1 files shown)


Thom1729:
What is the purpose of the mousemap? Would it be better to put that as a snippet in the readme?

mheinzler:
It is only supposed to show an example mousemap. I wanted to be able to show this beside the user's mousemap via the main menu "Package Settings", just like with the key bindings or the settings.

If this is a problem, I can remove the file of course and add it to the README.

PR #7473, opened 2019-02-19
Add Registers package

BernardoAbreu:

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -929,6 +929,16 @@
 			]
 		},
 		{
+			"name": "Registers",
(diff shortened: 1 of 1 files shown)


BernardoAbreu:
The keybindings for OSX were changed.

Thom1729:
It looks like this won't work when installed. Instead of storing the “registers” in a JSON file inside the package, you should use the settings API.

Thom1729:
@BernardoAbreu ping

BernardoAbreu:
I updated the package to use the settings API instead.

PR #7483, opened 2019-03-03
Added Jenkinsfile plugin

june07:

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -623,7 +623,7 @@
 		},
 		{
 			"name": "Jenkinsfile",
-			"details": "https://github.com/june07/sublime-Jenkinsfile",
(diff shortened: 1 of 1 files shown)


june07:
Added version tag.  0.1.0

june07:
LICENSE is there?!

june07:
@wbond Do I need to perform some magic here?

Thom1729:
You probably need to release a new tag.

Thom1729:
A few comments:

- Instead of `'win32' in str(sys.platform).lower()`, use `sublime.platform() == 'windows'`.
- In a `TextCommand`, instead of `view = sublime.Window.active_view(sublime.active_window())`, use `view = self.view`.

june07:
@Thom1729 Updated.

PR #7486, opened 2019-03-03
Update d.json

mikasuon:
A Sublime Text 3.X Plugin to handle Microsoft Dynamics 365 Web Resources.

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -34,6 +34,17 @@
 			]
 		},
 		{
+			"name": "D365 Web Resources",
+			"details": "https://github.com/mikasuon/Sublime-Text-D365-Web-Resources",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"platforms": ["windows"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "D3js v4",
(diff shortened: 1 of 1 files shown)


dvhh:
@mikasuon  change your release tag from `v1.0.2` to `1.0.2`

mikasuon:
Hello,

 

Done!

 

/ Mika S

Thom1729:
Is the `adal` module built-in on Windows?

mikasuon:
Hello,

 

No, it’s not.

 

The files in  <https://github.com/AzureAD/azure-activedirectory-library-for-python/tree/dev/adal> https://github.com/AzureAD/azure-activedirectory-library-for-python/tree/dev/adal should be copied to [Sublime Text]\Lib\pythonX\adal\ folder.

 

That’s a little impractical, could be better to include them in the project, but I’m not sure if I’m allowed to include those in my project.

 

Br,

Mika Suonvieri

Thom1729:
It looks like the code is MIT-licensed, which means that you're allowed to redistribute them as long as you include the license. The simplest way to do this would be to have an `adal` directory in the package and use relative imports. You could also create a [dependency](https://packagecontrol.io/docs/dependencies), which would allow other packages to use `adal`.

Thom1729:
@mikasuon ping

PR #7491, opened 2019-03-06
Add Tomato Time

FloydaGithub:
a sublime plugin to use tomato time


Thom1729:
In `CreateTomatoCommand`, you should use `self.view.window()` instead of `sublime.active_window()`.

FloydaGithub:
@Thom1729  Thank you for your advice, I had modify it.

PR #7496, opened 2019-03-09
Added Epitech Header

IAMISSAM:

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -1,3 +1,4 @@
+
 {
 	"schema_version": "3.0.0",
 	"packages": [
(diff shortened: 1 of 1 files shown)


Thom1729:
Rather than using a `WindowCommand` and getting `self.window.active_view()`, why not use a `TextCommand` and get `self.view`?

Sublime keeps comment-related information in `tmPreferences` files. There's an API to get at that information, but it's undocumented. If you do use a hard-coded list, it would probably be best to keep it in settings so that the user can override it.

Thom1729:
@IAMISSAM ping

PR #7497, opened 2019-03-10
added International Keymap plugin

hxss:
[Bitbucket](https://bitbucket.org/hxss/internationalkeymap/src/1.0.0/)
Creates localised keymaps based on sublime keymaps files.

Related discussions:
[Get merged current keymap](https://forum.sublimetext.com/t/get-merged-current-keymap/38767)
[On keymap changed event?](https://forum.sublimetext.com/t/on-keymap-changed-event/42149)

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -838,6 +838,16 @@
 			]
 		},
 		{
+			"name": "International Keymap",
+			"details": "https://bitbucket.org/hxss/internationalkeymap",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Internet Search",
 			"details": "https://github.com/krasun/SublimeInternetSearch",
 			"releases": [


Thom1729:
Will this work when installed/zipped? It looks like it's working with file paths inside the package, which seems like it wouldn't work if the package is zipped.

hxss:
@Thom1729 I was add [new check](https://bitbucket.org/hxss/internationalkeymap/commits/c0eb85ac24e3b6fbb3145ee65024ccb354a0f2f5) for plugin dir

Thom1729:
I don't understand the motivation for the special settings handling that uses the filesystem.

It looks like you're also creating keymap files under `Packages/InternationalKeymap`. Instead, it's better to create them under `Packages/User/InternationalKeymap`. This avoids manually creating new package directories.

hxss:
Do you mean that `Packages/User/InternationalKeymap` is directory or file? Anyway it still will be created manually by my script. So why it should breaks out from the general pattern of storing packages files?

PR #7502, opened 2019-03-12
Add package for Terraform AWS snippets

memoryleak:
Provides Terraform snippets for the AWS provider.

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -602,6 +602,18 @@
 			]
 		},
 		{
+			"name": "TerraformSnippets-AWS",
+			"labels": ["snippets", "terraform"],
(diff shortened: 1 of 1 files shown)


Thom1729:
Sublime will try to load `.py` files at the top level of a package. It looks like `Scraper.py` and `Terraform.py` are not meant to be loaded by Sublime, in which case they should be moved to a subdirectory.

Thom1729:
@memoryleak ping

memoryleak:
> Sublime will try to load `.py` files at the top level of a package. It looks like `Scraper.py` and `Terraform.py` are not meant to be loaded by Sublime, in which case they should be moved to a subdirectory.

The files have been moved to bin/ folder

PR #7505, opened 2019-03-12
Added ukrainian spell check dictonary

keedhost:


Thom1729:
It looks like the package is not alphabetized correctly in the channel JSON.

Thom1729:
@keedhost

PR #7508, opened 2019-03-13
Add Direct Copy plugin

UniFreak:

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -784,6 +784,16 @@
 			]
 		},
 		{
+			"name": "Direct Copy",
+			"details": "https://github.com/UniFreak/SublimeDirectCopy",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Directory Settings",
 			"details": "https://github.com/davepeck/DirectorySettings",
 			"releases": [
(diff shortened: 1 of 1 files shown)


Thom1729:
Please add a license.

UniFreak:
Added MIT license.

Thom1729:
What's the purpose of the odd settings behavior?

UniFreak:
To load the default setting if no user setting is found. 
I don't know whether there is a better way to do this.

Thom1729:
You can put a `DirectCopy.sublime-settings` file in your package with whatever defaults you need. If the user has their own `DirectCopy.sublime-settings` under `User` then it will supersede the defaults key-by-key.

It looks like you should simply rename your `Default.sublime-settings` to `DirectCopy.sublime-settings`. Then, `sublime.load_settings('DirectCopy.sublime-settings').get('entry')` will default to your initial value unless the user's `DirectCopy.sublime-settings` file has an `entry` key. When you save the settings object, it will create/update the user's `DirectCopy.sublime-settings` file, leaving the original intact.

UniFreak:
Thanks for your suggestion. I tried it and it works as expected. Also I need to edit the `Main.sublime-menu` to point the default setting file to `DirectCopy.sublime-settings`.

I've made a new commit.

PR #7511, opened 2019-03-18
Add Brevis CSS Autocomplete

dlzi:

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -1330,6 +1330,17 @@
 			]
 		},
 		{
+			"name": "Brevis Autocomplete",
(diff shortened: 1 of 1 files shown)


Thom1729:
Does `self.data_completions` exist?

dlzi:
It does. Plz check the line 9 of the code.

Thom1729:
That initializes `class_completions`, not `data_completions`.

dlzi:
Yeah.. you are right, sorry. Brevis CSS doesn't use data attributes, so I don't need data attribute completions. I just need classes.

dlzi:
Fixed. Thank you for pointed that out.

PR #7517, opened 2019-03-20
Add Octave Support

apjanke:
This package provides syntax definitions, indent settings, and snippets for GNU Octave.

There's an existing [GNU Octave Completions](https://github.com/tushortz/GNU-Octave-Completions) package, but that only does completions; this package has additional functionality.

### DIFF
--- a/repository/o.json
+++ b/repository/o.json
@@ -141,6 +141,17 @@
 			]
 		},
 		{
+			"name": "OctaveSupport",
+			"details": "https://github.com/apjanke/OctaveSupport",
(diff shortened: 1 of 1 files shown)


Thom1729:
Would it make sense to add new functionality to the existing package rather than creating a new package?

apjanke:
Yes. It's just a question of whether the author of the existing package is willing to accept all these changes. I'll ask them.

apjanke:
Asked the existing package's author: https://github.com/tushortz/GNU-Octave-Completions/issues/3

PR #7526, opened 2019-03-27
Added Nakama Autocomplete package

Superxwolf:
Adds autocomplete and documentation functionality for Nakama server lua functions.

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -12,6 +12,17 @@
 			]
 		},
 		{
+			"name": "Nakama Autocomplete",
+			"details": "https://github.com/Superxwolf/Nakama-Sublime",
+			"labels": ["auto-complete"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


Thom1729:
In `NakamaCompletionEvent`, it's not a good idea to use instance variables to store temporary view-specific state. A single instance will be shared among all views. You might use a `ViewEventListener`, which is specific to each view.

`sublime.load_settings('nakama-doc.sublime-settings')` should always return a `Settings` object (unless the API isn't loaded yet, in which case you'd have bigger problems). So there should be no need to check the return value.

Instead of `return view.settings().get("syntax") == "Packages/Lua/Lua.sublime-syntax"`, you might check the scope: `view.match_selector(0, 'source.lua')`. This will work for users who use third-party Lua syntaxes.

Superxwolf:
I applied the changes you mentioned. As for the syntax check, I check for `'source.lua' in view.scope_name(0)` because some plugins modify the scope, like LuaExtender changes it to `source.luae`. `view.match_selector(0, 'source.lua')` always returned false in that instance.

PR #7530, opened 2019-03-29
Add TSL package

polygoat:
This package adds syntax highlighting and a build system for TSL, a text scraping language for intuitive processing of text files, offering basic data cleanup and normalization tasks to non-programmers.

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -3118,10 +3118,11 @@
 			"name": "TSL Text Scraping Language",
 			"author": "Dan Borufka",
 			"details": "https://github.com/polygoat/TSL",
-			"labels": ["TSL", "text scraping language", "scraping"],
+			"labels": ["TSL", "scraping", "text scraping language"],
 			"releases": [
 				{
 					"sublime_text": ">=3092",
+					"details": "https://github.com/polygoat/TSL/tags",
 					"tags": true
 				}
 			]
--- a/repository/t.json
+++ b/repository/t.json
@@ -3122,7 +3122,6 @@
 			"releases": [
 				{
 					"sublime_text": ">=3092",
-					"details": "https://github.com/polygoat/TSL/tags",
 					"tags": true
 				}
 			]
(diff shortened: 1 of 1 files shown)


polygoat:
The package _does_ contain a LICENSE file though

Thom1729:
It looks like you need to release a new tag.

Thom1729:
`sys.path` is shared between the entire plugin runtime, so you shouldn't modify it. Instead, use relative imports (`from .TSLCore import TaskRunnerCore`).

polygoat:
Thanks, @Thom1729 !

Thom1729:
What's the purpose of this bit at the bottom of TSL.py?

```py
if len(sys.argv) > 1:
	taskFileName = sys.argv[1]

	TSL = TaskRunner(taskFileName)

	TSLPlugins = glob('tsl-plugins/*.py')

	if len(TSLPlugins):
		for TSLPlugin in TSLPlugins:
			exec(open(TSLPlugin, 'r', encoding='utf8').read())
	
	if TSL.task:
		print('Running "%s"...\n' % taskFileName)
		TSL.run()

else:
	print()
```

polygoat:
@Thom1729 I haven't had the chance to document this yet, but this just opens up the ability for auto-discovery & running of TSL plugins in the future

Thom1729:
Sublime will execute that code when the module is loaded, but it doesn't look like it would make sense within the plugin runtime. For instance, `sys.argv` won't generally exist, and `glob` isn't likely to find anything under Sublime's working directory. In addition, `exec`-ing unknown modules in the plugin runtime seems dangerous.

polygoat:
Hi @Thom1729. Please note that so far the script was run through Python externally, so sys.argv would've been populated by Sublime calling the shell command `python TSL.py $filename`. The working directory would be the file from which the build is called. But I have refactored this to be a proper plugin and use Sublime's internal Python engine instead. Please see version `1.2.0`. I have also removed the code that just execs, as it isn't needed at this point and I agree with your concerns.

PR #7533, opened 2019-03-30
add Wesnoth Markup Language

sevu:
Hello. I would like to add a syntax definition to highlighting WML, such as [this one](https://github.com/wesnoth/wesnoth/blob/1.14/data/campaigns/tutorial/scenarios/01_Tutorial_part_1.cfg).


sevu:
@packagecontrol-bot

Thom1729:
This can be merged as-is, but I would suggest changing the name to something a bit more descriptive, like “Wesnoth Markup Language”.

sevu:
Thanks for looking at it. 
I had considered both ways, though I ended up to use the naming scheme of the default packages and to put the details in the description and keywords, with the former one also being displayed in the search results.

PR #7536, opened 2019-04-01
commit n1_corrected

SmileAntoine:


Thom1729:
You need to modify the channel entry: `"sublime_text": ">=3092"`.

Thom1729:
@SmileAntoine ping

PR #7540, opened 2019-04-03
add MultiPaste

cosmozhang1995:


Thom1729:
This won't work when the package is installed. You're modifying files inside the package, but the package will be a zip archive when installed.

Instead, you can simply store the strings using the Settings API.

Thom1729:
@cosmozhang1995 ping

PR #7541, opened 2019-04-04
Add "Show Unicode Name"

ned-martin:
Show the Unicode character name and Unicode code point of the character at the cursor in the status bar in Sublime Text 3


Thom1729:
Please add a license.

ned-martin:
MIT LICENSE added

PR #7542, opened 2019-04-06
Remove chunks add editmode

tonyennis145:
This package is intended to work with the Editmode API (www.editmode.app). It allows users to replace content in their files with helpers which pull the content from an API. The API was previously named Chunks but is now called Editmode, so I have removed the Chunks plugin.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1165,20 +1165,6 @@
 			]
 		},
 		{
-			"name": "Chunks",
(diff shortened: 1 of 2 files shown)


Thom1729:
The new entry should have `"previous_names": ["Chunks"],`.

Thom1729:
@tonyennis145 see the bot output.

PR #7543, opened 2019-04-06
A syntax highligther to work with dart and flutter

LiteKat:

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -185,7 +185,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"tags": true
+					"tags": true 
 				}
 			]
 		},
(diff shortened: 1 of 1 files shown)


Thom1729:
It looks like a Dart package already exists. How does this package differ from the existing one? If this package improves upon that one, then would it make sense to integrate those improvements into the existing package rather than creating a new one?

LiteKat:
Yeah a dart package already exist . I would create a pull request to that package but it has been abandoned and in 4 years the syntax has changed quite a lot.

The main purpose is to bring a updated syntax [it's the same syntax that they use in the vscode dart plugin] . And periodically update it to the latest version now that Dart plugin is quite abandoned

Thom1729:
If this is derived from the VSCode package, then you should at least link it. Is the license also taken from that package?

LiteKat:
> If this is derived from the VSCode package, then you should at least link it. Is the license also taken from that package?

Done, now it is linked in the description and it uses MIT license as the original package

PR #7546, opened 2019-04-08
add braver/ColorHints

braver:
Adds a package that will show inline color boxes. Kinda like ColorHelper but without the gazillion features, kinda like ColorHighlight but better.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2776,6 +2776,15 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


braver:
Sorry, beginner's mistake 🙄 tagged a release now.

Thom1729:
Probably should `from .lib import util` instead of `from ColorHints.lib import util`.

braver:
👌🏻

PR #7548, opened 2019-04-09
add webloc

looking-for-a-job:


Thom1729:
What's the story here? What happened to the old repository?

looking-for-a-job:
Thom1729, probably it was  one of my firsts pr on github, I was tired of waiting, deleted the repository and unsubscribed, and it was merged but not available all this time :)

Thom1729:
Instead of `sys.platform`, you should use `sublime.platform`.

You should also add `"previous_names": "Webloc-open"` to the channel entry.

looking-for-a-job:
fixed

PR #7549, opened 2019-04-10
add JSON Stringify from Nadock/json_stringify

Nadock:
This is a simple tool to convert between JSON and a string serialisation of that JSON data.

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -1103,6 +1103,16 @@
 			]
 		},
 		{
+			"name": "JSON Stringify",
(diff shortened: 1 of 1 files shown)


Thom1729:
Any .py file at the top level of the package will be run when the package loads. It looks like the test file probably shouldn't be.

Nadock:
> Any .py file at the top level of the package will be run when the package loads. It looks like the test file probably shouldn't be.

Yep you're correct, this has been fixed in the latest release

PR #7556, opened 2019-04-19
Add Jasonette Completions

BarryMode:
Adds jasonette completions. I ported [these](https://github.com/felipemullen/jasonette-snippets), which were made for Visual Studio Code.

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -207,6 +207,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


Thom1729:
In addition to a tag, you should add a license and a readme.

BarryMode:
@Thom1729, thanks! I added the tag, license, and readme.

PR #7559, opened 2019-04-25
Add Sublime-API

looking-for-a-job:

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -4062,17 +4062,6 @@
 			]
 		},
 		{
-			"name": "Sublime-API",
-			"details": "https://github.com/looking-for-a-job/sublime-api",
-			"labels": ["build system"],
-			"releases": [
-				{
-					"sublime_text": "*",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "Sublime convert colorcode",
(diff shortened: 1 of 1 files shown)


looking-for-a-job:
curl -X POST https://packagecontrol.io/test_pr/$TRAVIS_PULL_REQUEST.json
curl: (18) transfer closed with 24 bytes remaining to read
The command "curl -X POST https://packagecontrol.io/test_pr/$TRAVIS_PULL_REQUEST.json" exited with 18.

FichteFoll:
I don't quite think I understand the purpose of this package. From the code and the readme, it appears you are providing a build system to execute Python files within ST's Python environment. But what for? When developing a plugin, it's easier to just dump that into the User folder and have ST reload it on modification/save.

There are also quite a few flaws in this, such as the execution being blocking and on the main UI thread and the file handle not being closed, as well as no error handling for unsaved views (where `self.view.file_name() is None`). Not considering that you wrapped the entire thing in a try-catch, which isn't exactly good practice. You might also want to check out `contextlib.redirect_stdout` from the stdlib.

And finally, your package contains the name "Sublime", which is agaisnt our policies. While there is a certain relevance in the name considering we're talking about executing Python code within ST's environment, a name like "Sublime API" doesn't tell you about that at all.

PR #7560, opened 2019-04-28
Add Plugin Geecode

qiuxfeng1985:

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -187,6 +187,16 @@
 			]
 		},
 		{
+			"name": "GeeCode",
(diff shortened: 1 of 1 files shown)


Thom1729:
- What does the plugin do? (I can't read the readme.) Is it designed fora specific programming language?
- The `TipDisplay.py` file contains carriage returns instead of newlines.
- Some of the files are modifying `sys.path`. This is a bad idea, because `sys.path` is shared among the entire plugin runtime. Instead, use relative imports.

PR #7561, opened 2019-04-28
Add "PHP Namespace Monkey"

itsgoingd:
Automatically adds a namespace declaration and optionally a class definition to newly created PHP files in projects using the PSR autoloading standards. Port of my existing Atom package.

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -1141,6 +1141,17 @@
 			]
 		},
 		{
+			"name": "PHP Namespace Monkey",
+			"details": "https://github.com/underground-works/sublime-php-namespace-monkey",
+			"labels": ["php", "utilities"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "PHP Source",
 			"details": "https://github.com/mjkaufer/PHP-Source",
 			"labels": ["php", "utilities", "utils"],


Thom1729:
Maybe I'm missing something, but what causes this to run only when a new PHP file is created?

itsgoingd:
Hey, this runs every time a view is activated, since that is the only event that makes sense afaik.

First thing I do is check if the view is empty and has a php file associated here:
https://github.com/underground-works/sublime-php-namespace-monkey/blob/master/PhpNamespaceMonkey.py#L9

itsgoingd:
Hey, this runs every time a view is activated, since that is the only event that makes sense afaik.

First thing I do is check if the view is empty, so we never modify existing content, and has a php file associated here:

https://github.com/underground-works/sublime-php-namespace-monkey/blob/master/PhpNamespaceMonkey.py#L9

The php file condition itself is a simple file extension check:

https://github.com/underground-works/sublime-php-namespace-monkey/blob/master/PhpNamespaceMonkey.py#L59-L62

Additionally we check if the file ctime is less than a second ago:

https://github.com/underground-works/sublime-php-namespace-monkey/blob/master/PhpNamespaceMonkey.py#L11

We assume that if the ctime is so recent it is a "new" file, this works well for the intended use where you create a new file via right-click option in the sidebar for example, which creates the file and automatically opens the view emitting the activated event (see gif in the original repo https://github.com/underground-works/sublime-php-namespace-monkey).

This is obviously not perfect, but I used the same "workaround" in the original implementation for Atom and it worked very well.

itsgoingd:
@Thom1729 anything more I can do to get this merged?

PR #7563, opened 2019-04-30
Add Zeek Language package

jsiwek:
Provides syntax highlighting for the Zeek scripting language
https://github.com/zeek/zeek
https://zeek.org

### DIFF
--- a/repository/z.json
+++ b/repository/z.json
@@ -44,6 +44,17 @@
 			]
 		},
 		{
+			"name": "Zeek Language",
+			"details": "https://github.com/zeek/zeek-sublime",
+			"labels": ["language syntax"],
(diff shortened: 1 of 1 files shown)


jsiwek:
The license is in a file named COPYING, not LICENSE, but don't mind renaming if that's required.  GitHub itself auto-detects the former as a license file.

Thom1729:
You're using a lookbehind in the `main` context:

```yaml
    - match: (?<=\S)(:)
      scope: punctuation.separator.zeek
```

This will cause the entire context (i.e. the entire syntax, because there is only one context) to use the slower regexp engine. Removing the lookbehind will improve performance.

PR #7566, opened 2019-04-30
Add LSP-vue plugin

predragnikolic:
Vue support for Sublime's LSP plugin.

Currently a user needs to manual configure the vue language server, which is not error prone.
This plugin (language handler) will automate that step.

If you need more info, please say.

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -1598,6 +1598,16 @@
 			]
 		},
 		{
+			"name": "LSP-vue",
+			"details": "https://github.com/sublimelsp/LSP-vue",
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Lua Dev",
 			"details": "https://github.com/rorydriscoll/LuaSublime",
 			"labels": ["lua"],


FichteFoll:
It seems you created a couple similar packages for various syntaxes. What do you think about unifying them into a single package that provides initial configuration for multiple syntaxes?

predragnikolic:
> It seems you created a couple similar packages for various syntaxes.

Yes
> What do you think about unifying them into a single package that provides initial configuration for multiple syntaxes?

I am not exactly sure what you mean by that.

Here is how I understood it.
I create a plugin `LSP setup`,
and in that plugin the user can set up various language handlers?

Is there a different way?
For example, I just discovered that the plugins for SublimeLinter are not located in this repo, 
instead they are located [here](https://github.com/SublimeLinter/package_control_channel/blob/master/org.json).

Can I create a channel with a list of LSP plugins?

FichteFoll:
The idea behind my suggestion was to bundle initial configuration setups for LSP in a single package, since they are really just single Python commands that do some modifications to the LSP settings file and potentially ensure the required binary is installed by running the installation steps. Since the process is very similar and using a central place for these allows for easier additions (and notably does not require submitting and reviewing a new package every time), I think bundling those in a single package would be easier to handle. I don't think providing "LSP packages" in a separate repository.json would be reasonable since for SublimeLinter the authority and review of plugins should fall to the maintainers of SublimeLinter, while I doubt the maintainer of LSP would be interested in maintaining a list of packages that provide initial configurations (and don't interact with LSP in any way afterwards).
Of course, you may still prefer to supply these packages separately. I am merely making a suggestion here.

Currently, the package could contain html, css, vue, json, intelephesense.

predragnikolic:
> a list of packages that provide initial configurations (and don't interact with LSP in any way afterwards) 

Well, some of the LSP-plugins implement custom server functionality, [see](https://github.com/tomv564/LSP/issues/212#issue-275814949), but some of the plugins like the one I submitted (html, css, vue) don't.

> Of course, you may still prefer to supply these packages separately. I am merely making a suggestion here.

Yes, I would still prefer it, but "reviewing a new package every time" makes me think that there should be a better way.

This needs to be discussed with the maintainers of LSP, on how this could be done. 
Should I close all my PR until a solution is found?

FichteFoll:
Let's just invite the maintainers into this PR for their take.

@tomv564 @rwols

predragnikolic:
Hello @FichteFoll,
I got a notification yesterday that LSP-html, LSP-json, LSP-intelephense and LSP-css got merged.

Now that some of the plugins are merged, 
Should we still wait for the response of the maintainers, or?

tomv564:
Sorry it took a while to respond. I feel like publishing a package when all the user needs to do is `npm install -g vls` is a bit unnecessary. 

That said, we haven't looked at helping the user update their server either, which in the VS Code case is done by (bumping) the extension. I'd like to avoid bumping LSP for every single server.

A bundle of other features together with LSP support (like Rust Enhanced) seems ideal to me.

FichteFoll:
I'll leave this to @Thom1729 to decide. I've been too inactive over the past weeks to rule over his decision, although I do still very much favor the "LSP templates in a single package" approach.

Thom1729:
@predragnikolic Could you resolve the merge conflict?

predragnikolic:
@Thom1729 I will tomorrow :)

PR #7567, opened 2019-05-02
Add csv-record-view

mrp130:
Plugin to transform csv row to record view

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -4404,6 +4404,16 @@
 			]
 		},
 		{
+			"name": "CSV Record View",
+			"details": "https://github.com/mrp130/csv-record-view",
+			"releases": [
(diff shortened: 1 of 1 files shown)


Thom1729:
Do any existing packages provide this functionality? (I don't know what “record view” is, so I'm not clear on this.)

mrp130:
@Thom1729 I searched and cannot find existing packages provide this functionality. To make clear, this is example for record view https://github.com/dbeaver/dbeaver/wiki/Data-View-and-Format#table-vs-record-views

Thom1729:
Instead of `sublime.active_window()`, you should use `self.view.window()`.

mrp130:
done @Thom1729

PR #7578, opened 2019-05-08
Add Duplicate File

looking-for-a-job:

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -1602,6 +1602,16 @@
 			]
 		},
 		{
+			"name": "Duplicate File",
+			"details": "https://github.com/looking-for-a-job/sublime-duplicate-file",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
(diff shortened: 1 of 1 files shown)


Thom1729:
- The keybinding conflicts with built-in behavior. Is a keybinding really necessary for this?
- You should add a `.sublime-commands` file.
- Rather than calling `sublime.active_window().active_view()`, make it a `TextCommand` and use `self.view`.
- The `path` property is unused.
- Rather than checking whether the file exists, just try to create it and catch `FileExistsError`. This avoids race conditions.

looking-for-a-job:
done.
- the right question is why the Sublime Text does not support default keyboard shortcut? :) yes, it is really necessary and must be builtin by default :)
- `shutil.copyfile` does not raise `FileExistsError`

Thom1729:
I think that overriding the built-in binding is unfortunately a non-starter. This could be very disruptive for users, and it isn't easy to selectively disable package key bindings.

PR #7579, opened 2019-05-14
Add a MarkdownInsertImage plugin

lizhiyong2000:

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -542,6 +542,7 @@
 			"releases": [
 				{
 					"sublime_text": ">=3000",
+					"platforms": ["osx", "linux"],
 					"tags": true
 				}
 			]
--- a/repository/m.json
+++ b/repository/m.json
@@ -537,6 +537,17 @@
 			]
 		},
 		{
+			"name": "MarkdownInsertImage",
+			"details": "https://github.com/lizhiyong2000/MarkdownInsertImage",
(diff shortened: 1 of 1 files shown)


Thom1729:
It looks like there's a spurious addition in the channel file: `"name": "MarkdownLight",`.

FichteFoll:
1. Your second gif appears to be broken. It doesn't render correctly in FireFox or mpv.
1. The base scope of the markdown syntax is `text.html.markdown`, which your key bindings would never match.
1. You mention that the user needs a Python dependency but you don't mention that they also need to have `python3` in their PATH.
1. You call `get_python_sitepackages()` in `run` but don't capture the return value.
1. You check for jpg files when searching through the clipboard but only use .gif or .png when uploading the file
1. You add the entire system Python's site packages path to ST's python path. That is an absolute **no-go** as it will cause conflicts with potentially different target versions for those libraries and ST's embedded Python.
Instead, you can either call the system python directly to import the necessary modules in a specific script and perform the upload there, or you just directly communicate with qiniu's API endpoints through `urllib` or [`requests`](https://github.com/packagecontrol/requests).
1. There are a couple debug prints throughout the code.
1. Where do you use SideBarEnhancements?

FichteFoll:
@lizhiyong2000 ping

PR #7580, opened 2019-05-14
Add isort package

onanypoint:
Package for the [isort](https://github.com/timothycrosley/isort) command.

Compared to other packages available for this command, it doesn't include the library itself. It uses the user installed on instead (configurable using the settings).

Also, the available packages for this command seems to be currently unmaintained ([thijsdezoete/sublime-text-isort-plugin](https://github.com/thijsdezoete/sublime-text-isort-plugin) for example). By removing the direct inclusion of the isort library, this plugin is less likely to be out of sync with the mainline isort command.

In the pull request #7240  a suggestion was made to "replace the existing isort package instead of adding a clone under a different name". As this is not an updated version of the [thijsdezoete](https://github.com/thijsdezoete/sublime-text-isort-plugin) package, I don't know if this proposition is applicable here.

This is my first package, feedbacks are welcome.

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -1163,6 +1163,17 @@
 			]
 		},
 		{
+			"name": "iSort Python Imports",
+			"details": "https://github.com/onanypoint/iSort-Python-Imports",
+			"labels": ["python", "formatting"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "ITDCHelper",
 			"details": "https://github.com/akalongman/sublimetext-itdchelper",
 			"author": "Avtandil Kikabidze",


onanypoint:
Would it possible to look at this package, please?

Thom1729:
If this package provides substantially the same functionality, then it would make sense to replace the old one.

Also, `get_setting` should really take a view argument rather than getting the current view.

asfaltboy:
Note: a fork switch was also attempted in #7580. As @thijsdezoete does not seem to be responsive, maybe one of us can take over maintaining the forked version?

asfaltboy:
Can we reboot the discussion? @Thom1729 please note that the original author has expressed willingness in switching/transferring the plugin here https://github.com/jcugat/sublime-text-isort-plugin/pull/2#issuecomment-524793411 

I have recently forked off, and added a few features:
https://github.com/asfaltboy/sublime-text-isort-plugin/tree/feat/sort-on-save

@onanypoint would you consider joining our efforts?

I opted for keeping the vendored version here, and am also considering making `isort` a [dependency package][deps] instead of keeping it in the repo. But either way, I think this code is a bit simpler than having to call it from the command line and avoids issues like searching path (or a setting for a custom path specifier).

If you prefer, I can also add your code as a module, to optionally allow users to use a different version (though I don't see many use cases for this with `isort` tbh, as they keep it very simple and most recent changes are configurable).

At this moment, I will submit a PR for the original extension repo, and later we can discuss if we want this as part of a single package (extra features), or as a separate plugin.

[deps]: https://packagecontrol.io/docs/dependencies

FichteFoll:
This definitely looks like a big improvement over the current isort package. You have my recommendation (and permission from the authors) to completely replace it.

Normally, I would suggest mentioning that the user needs to ensure that their system's Python, which they install isort for, needs to be on their PATH, but your error message and the fact that this package is for Python development should make this clear enough.

When using subprocess, you can specify `universal_newlines` and an encoding to be used for decoding the stream instead of doing both manually afterwards. I suggest using `locale.getpreferredencoding()` over using settings because the latter use different identifiers/names than Python. You could use [sublime_lib](https://sublimetext.github.io/sublime_lib/) to convert between them, however.

FichteFoll:
I see this has already been superseded by #7673.

PR #7581, opened 2019-05-17
Adding 'shell format'

marcfrederick:


Thom1729:
It looks like the channel entry is not correctly sorted.

PR #7583, opened 2019-05-19
Add Z80 Assembly (RGBDS)

shinmai:
A Z80 Assembly syntax definitions for the [RGBDS](https://github.com/rednex/rgbds) Game Boy development toolchain.
There is a Z80 syntax definition package available, but it's for the Pasmo cross-assembler, which has a wildly different syntax to rgbasm. Obviously there are dozens of Z80 assemblers, and it probably doesn't make sense to have definitions for each one separately, but I believe RGBDS is popular enough among Game Boy homebrew enthusiasts to warrant this packages existence.

### DIFF
--- a/repository/z.json
+++ b/repository/z.json
@@ -29,7 +29,7 @@
 			"labels": ["language syntax", "z80", "assembly"],
 			"releases": [
 				{
-					"sublime_text": ">=3084",
+					"sublime_text": ">=3092",
 					"tags": true
 				}
 			]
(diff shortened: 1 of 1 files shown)


shinmai:
Mea culpa, I blindly accepted the version quoted [in the unofficial docs](http://docs.sublimetext.info/en/latest/extensibility/syntaxdefs.html) which is 3084. Will amend!

avivace:
@shinmai Hey! Good job on this

I'm the mantainer of the Game Boy development awesome list.

Does your package differ significantly from https://github.com/Bananattack/rgbds_textmate ? If so, can you add your package to the list?

PR #7584, opened 2019-05-21
Add icm-pro language

MarkZaretskiy:
Program icm-pro by MOLSOFT is a really powerful tool for drug-design, bioinformatics etc. It has an own scripting language, but all bioinformaticians have worked with it before  were too lazy to create a syntax file for it. So, I tried.

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -78,6 +78,16 @@
 			]
 		},
 		{
+			"name": "icm-pro",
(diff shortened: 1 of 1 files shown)


Thom1729:
You don't have any tags. You should also add a license.

FichteFoll:
@MarkZaretskiy ping

PR #7593, opened 2019-05-30
Add Earthbound themes

benbusby:
Adds seven color schemes based on the SNES game Earthbound.

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -29,7 +29,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"branch": "master"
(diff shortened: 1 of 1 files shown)


Thom1729:
You should add a license, and also a readme that briefly describes the package. I would recommend also adding screenshots of the themes.

benbusby:
There’s already a readme, license, and screenshots in the root level of the repo, which is where this package is pointed to already. Do they need to be copied to the sublime sub folder in my repo as well?

Thom1729:
My mistake; everything is fine.

PR #7599, opened 2019-06-11
Add 'Elvish Syntax' package

href:
This package adds syntax highlighting for the [Elvish](https://elv.sh) shell.

There is no such package yet for Sublime Text.

Thanks for your consideration! 🤞


href:
Great, thank you 😃

PR #7606, opened 2019-06-20
add CodeStatistics.

absop:
This package is a plugin for SublimeText3 to count the line-number of codes under a sidebar folder.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2222,6 +2222,17 @@
 			]
 		},
 		{
+			"name": "CodeStatistics",
+			"details": "https://github.com/absop/CodeStatistics",
+			"labels": ["analytics", "statistics"],
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "CodeStats",
 			"details": "https://github.com/code-stats/code-stats-sublime",
 			"labels": ["analytics", "statistics"],


Thom1729:
The keybindings are fine because they only apply in the special context you're using,

Generally, instead of checking whether a file exists (or is a file, etc) before performing an operation on the file, it's better to try the operation and catch the error. This avoids race conditions.

Totally optional, but you might find some of [sublime_lib](https://github.com/SublimeText/sublime_lib)'s features useful, particularly [OutputPanel](https://sublimetext.github.io/sublime_lib/#sublime_lib.OutputPanel), [new_view](https://sublimetext.github.io/sublime_lib/#sublime_lib.OutputPanel), and  [ActivityMonitor](https://sublimetext.github.io/sublime_lib/#activity-indicator).

FichteFoll:
It looks like the repo was renamed to (or reuploaded as) https://github.com/absop/CodeCounter. Please update the url.

In that repo, you have a `.no-sublime-package`, but I don't see any reason for there to be one. If you don't have any, please remove it, as it otherwise complicates certain things for users, namely overrides.

Usually, you should also add all commands you expose in a menu to the command palette, so users can access it easily with the keybord only.

Other than that, the package looks good.

absop:
> It looks like the repo was renamed to (or reuploaded as) https://github.com/absop/CodeCounter. Please update the url.
> 
> In that repo, you have a `.no-sublime-package`, but I don't see any reason for there to be one. If you don't have any, please remove it, as it otherwise complicates certain things for users, namely overrides.
> 
> Other than that, the package looks good.

Thanks for your appreciation and advice, I would improve it and reopen a new PR.
I added the file `.no-sublime-package` because there are several share object file compiled by C compilers under the package directory are used. I tried to load them from a packed sublime-package but I failed(I don't have enough knowledge of loading dynamic linked libraries from a sublime-package).

PR #7609, opened 2019-06-24
Add DreamMaker Language Client

SpaceManiac:
This package extends [LSP](https://packagecontrol.io/packages/LSP) with support for the [DreamMaker Language Server](https://github.com/SpaceManiac/SpacemanDMM/tree/master/src/langserver#dreammaker-language-server).

It also provides an integrated browser for the language's reference, a project "object tree" view with the assistance of the language server, and a command to toggle whether a file is included in the project or not.

There are two existing packages which provide syntax highlighting and a build system:
* [BYOND](https://packagecontrol.io/packages/BYOND)
* [Dream Maker](https://packagecontrol.io/packages/Dream%20Maker)

This package is intended to act as a full replacement and thus also provides syntax highlighting, as well as a build system which integrates more tightly with the language server support.

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -1412,6 +1412,17 @@
 			]
 		},
 		{
+			"name": "DreamMaker Language Client",
+			"details": "https://github.com/SpaceManiac/sublime-dm-langclient",
+			"labels": ["language syntax", "linting", "auto-complete", "build system"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Drive Color Scheme",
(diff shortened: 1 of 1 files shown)


SpaceManiac:
The package stores binaries in its folder, and it is therefore my belief that `.no-sublime-package` is warranted.

If there is a different preferred "cache" folder I could use that instead, but I could not find relevant docs.

Thom1729:
In fact, there is a directory for cached package files: `sublime.cache_path()`. You could store downloaded binaries in a `DreamMaker Language Client` directory under there.

SpaceManiac:
Thanks, I'll do it.

PR #7610, opened 2019-06-25
Add YASMJinja2 plugin for YASM syntax highlight

m-messiah:
Plugin for syntax highlight of YASM modified Jinja2 syntax, which uses `<` brackets instead of `{` for readable JSON templates

### DIFF
--- a/repository/y.json
+++ b/repository/y.json
@@ -110,6 +110,17 @@
 			]
 		},
 		{
+			"name": "YASMJinja2",
+			"details": "https://github.com/m-messiah/yasm-jinja2",
+			"labels": ["yasm", "jinja2", "yandex", "formatting", "syntax"],
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


m-messiah:
@packagecontrol-bot LICENSE added

Thom1729:
I see that this is a fork of the existing Jinja2 package. Would it make sense to add the YASM functionality to that package?

m-messiah:
@Thom1729 as far as i know, there is no solution to configure control characters through preferences.
And here we have two different syntaxes, where second (my) may be useless for common Jinja2 users, so it is strange to add this definition to Jinja2 package.

FichteFoll:
Unless you plan to target TextMate, you can remove `tmCommand`, `tmSnippet` and `tmMacro` files (and `info.plist`). Sublime Text doesn't use them.

PR #7613, opened 2019-06-29
Add Notedown package

darryllawson:
Notedown helps you use manage a collection of notes stored as Markdown files. Maybe call it a Personal Wiki? The key feature this adds is **linking between notes**.

Similar packages:

- PlainNotes: https://packagecontrol.io/packages/PlainNotes
  - Different design philosophy.
- Notes: https://packagecontrol.io/packages/Notes
  - Just a syntax highlighting scheme. Doesn't use Markdown.
- Wiki: https://packagecontrol.io/packages/Wiki
  - No longer maintained. Different design philosophy. Does not use Markdown.

PlainNotes is similar but has a different design philosophy; my Notedown package is designed to augment Sublime Text builtin features, not wrap them. E.g. Use the normal Command-P for opening note files, rather provide a different mecahnism.

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -834,6 +834,16 @@
 			]
 		},
 		{
+			"name": "Notedown",
+			"details": "https://github.com/darryllawson/sublime-notedown",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Notepad++ Color Scheme",
 			"details": "https://github.com/evandrocoan/SublimeNotepadPlusPlusTheme",
 			"labels": ["color scheme", "notepad++"],


Thom1729:
Can the tests be moved to a subdirectory?

What is install.py doing?

darryllawson:
"Can the tests be moved to a subdirectory?" I put that file at the top-level because a sub-directory would be redundant, but yes I can, if that is the required convention. Is this because the module pollutes a global namespace?

"What is install.py doing?" It is just a helper script for development use. It comes the package into the Sublime Text Packages directory,

Thom1729:
> I put test_notedown.py at the top-level because a sub-directory would be redundant, but yes I can, if that is the required convention. Is this because the module pollutes a global namespace?

It's not required, and reading through the tests it *probably* shouldn't make any difference — but then, if Sublime doesn't need to load the file, then it may as well be in a subdirectory where Sublime won't try to load it.

darryllawson:
That is a good reason to move that file to a sub-directory (to avoid ST loading it). I'll do that.

Thank you for your time!

PR #7619, opened 2019-07-05
 Add an updated 65c816 assembly syntax

camden7306:
While there already is a 65816 package, this package extends on the syntax, adding highlighting support for hex/numbers, labels, and registers.


Thom1729:
Would it make sense to enhance the existing package rather than adding a new one?

FichteFoll:
Package (and user) doesn't exist anymore.

PR #7620, opened 2019-07-06
Adds ROR Refactor package

zedtux:
This package provides the ability to _refactor_ some code in a Rails app.

After having  selected a part of a line of code, a line of code or multiple lines of code, triggering the `CTRL` + `ALT` + `CMD` + `R` shortcut, a new method definition is created above the current one, with the selected code, and cursors are well placed so when you type you name and call your new method.

Here is an example :

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -1671,6 +1671,17 @@
 			]
 		},
 		{
+			"name": "Ror Refactor",
+			"details": "https://gitlab.com/zedtux/st3-ror-refactor",
+			"labels": ["Ruby on Rails", "Ruby", "Rails", "Refactor", "Refactoring"],
+			"releases": [
+				{
+					"sublime_text": ">3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Ros Snippets",
 			"details": "https://github.com/ckrooss/ros_snippets",
 			"labels": ["snippets", "build system", "language syntax"],


zedtux:
Looks like you're not supporting Gitlab? 🤔

FichteFoll:
That's correct, we don't currently support gitlab. The tracking issue for this is https://github.com/wbond/package_control/issues/719. (Bitbucket is also still broken.)

Instead of providing a single key binding for macOS, you should instead consider adding a command palette entry (and maybe a binding for the other OSes).

FichteFoll:
@zedtux ping

zedtux:
I'll do the update, but that's kind funny you take so long to pay attention to my PR, and now, in 23 days my PR is marked as "stale" ... 😄

zedtux:
BTW why adding a command palette entry is blocking to merge this PR?

That doesn't prevent the package from working, and releasing the new version with that change will not affect this PR AFAIK.

FichteFoll:
> BTW why adding a command palette entry is blocking to merge this PR?

As you mentioned and I have confirmed earlier, we don't support gitlab at this moment. The command palette note was made with the words "should consider", not "must". It would be nice if you considered it, but it would not be a blocker. Regardless, I would have waited until you replied and read my comment before I merged.

> I'll do the update, but that's kind funny you take so long to pay attention to my PR, and now, in 23 days my PR is marked as "stale" ...

I don't think it's far-fetched for users that want their package to be listed to have a greater interest in responding to a single PR comment than I do when I review 80 PRs or more.

zedtux:
> it would not be a blocker

Okay, nice, so if I understood correctly you're not merging this PR because of Gitlab?

> I review 80 PRs and packages or more

Was kind of a joke, I can understand it.

zedtux:
Okay then I don't understand. I'm asking you a question, you're not answering, and and you're closing my PR with a "timeout" label. I could be dumb, but it looks weird to me.

FichteFoll:
My bad, I thought I communicated clearly enough that we don't support Gitlab and you haven't shown an interest in either moving your repo to Github or providing a custom `repository.json` with selfpacked releases. If you are interested in that, then I can of course reopen the issue (or you can submit a new PR), but at the time I was just cleaning up old PRs.

zedtux:
Okay, can you please tell me more about this custom `repository.json`? Do you have a documentation?

FichteFoll:
https://github.com/wbond/package_control/blob/master/example-repository.json

Example package: https://github.com/wbond/package_control_channel/pull/7749

zedtux:
Thank you, I'll have a look.

PR #7622, opened 2019-07-07
Adding AmbientColorScheme

vyatu:
My package changes Sublime's color scheme between light and dark based on Macbook's ambient light sensor readings. I did not find any similar packages and I thought it might quite literally brighten somebody's day!

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -696,6 +696,18 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


vyatu:
.no-sublime-package is defined to make use of a shared library written in C, included in the `lib` subdirectory

PR #7624, opened 2019-07-09
add LikeGit / remove Smart Repeat

poucotm:

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -923,6 +923,17 @@
 			]
 		},
 		{
+			"name": "LikeGit",
+			"details": "https://github.com/poucotm/LikeGit",
+			"labels": ["git", "graph"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "LimitTabs",
 			"details": "https://github.com/max0d41/LimitTabs",
 			"releases": [
--- a/repository/s.json
+++ b/repository/s.json
@@ -2392,17 +2392,6 @@
 			]
 		},
 		{
-			"name": "Smart Repeat",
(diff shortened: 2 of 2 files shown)


Thom1729:
@poucotm You should add a license.

Thom1729:
Why remove Smart Repeat?

Why manually reload modules in LikeGit?

poucotm:
I added the License. 

Why remove Smart Repeat? >> no body uses it. 
Why manually reload modules in LikeGit? >> to ensure updates

Thom1729:
What's the use case for the reloading logic? If it's to make it easier to work on , then [AutomaticPackageReloader](https://github.com/randy3k/AutomaticPackageReloader) would do the trick. Hot-reloading Python is really fiddly and error-prone, so if it's possible to offload the reloading logic to a “standard” implementation, that would be greatly preferable.

FichteFoll:
Unfortunately, because you intend to run `osx/terminal.sh`, it needs to be available as a file and requires the package to be unpacked. Add a `.no-sublime-package` file to tell PC this. Alas, I'm unsure if writing a temporary theme file at `LikeGit/theme/LikeGit.tmTheme` is a good idea, because it will be removed on each update, but it might just work out here. You can try using the cache folder instead (`sublime.cache_path()`).

FichteFoll:
@poucotm ping

poucotm:
I added .no-sublime-package. 
As I know, it cannot apply color_scheme setting for a theme file in cache_path().
I changed to use Packages/User folder instead. Thanks for comment !

FichteFoll:
`os.makedirs` is recursive and with the `exist_ok` parameter you don't need to check whether it exists already.

PR #7626, opened 2019-07-10
Add SPARQL Runner package

hevp:
Add SPARQL Runner Improved package

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -3032,7 +3032,7 @@
 		},
 		{
 			"name": "SPARQL Runner",
-			"details": "https://github.com/cezarsa/sublime-sparql-runner",
+			"details": "https://github.com/hevp/sublime-sparql-runner",
 			"releases": [
 				{
 					"sublime_text": "*",


Thom1729:
Why fork the existing package?

Thom1729:
If the original author isn't interested in maintaining it, then it's possible to transfer the package name to a new repository. Have you reached out to the author, e.g. on the issue tracker?

FichteFoll:
The author appears to be unresponsive (https://github.com/cezarsa/sublime-sparql-runner/issues/6), so we can go forward with transfering package ownership.

Please adjust the internal references to the package's name and do the necessary updates in this PR so we can go ahead with it.

FichteFoll:
@hevp the tags you push are not valid semantic versions. [Semantic versions](https://semver.org) consist of three numeric parts.

hevp:
Indeed, I have updated the tags.

FichteFoll:
Doesn't look like it. https://github.com/hevp/sublime-sparql-runner/tags

hevp:
One more try

PR #7632, opened 2019-07-15
Add "bst syntax" package

hadisfr:
Add "bst syntax" package: sublime syntax definition for LaTeX Bibliography Style (bst) files


hadisfr:
LICENSE has been added to repo.

PR #7634, opened 2019-07-16
Added HttpUnit

gaohuia:
Added SimpleHttpRequester

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -1386,6 +1386,17 @@
 			]
 		},
 		{
+			"name": "HttpUnit",
+			"details": "https://github.com/gaohuia/HttpUnit",
+			"issues" : "https://github.com/gaohuia/HttpUnit/issues",
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Huely Palette Extractor",
 			"details": "https://github.com/Jeloi/huely-sublime-plugin",
 			"releases": [
--- a/repository/s.json
+++ b/repository/s.json
@@ -1937,17 +1937,6 @@
 			]
 		},
 		{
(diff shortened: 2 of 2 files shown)


Thom1729:
Are there other packages that perform a similar task?

gaohuia:
> Are there other packages that perform a similar task?

The answer is YES, but I think SimpleHttpRequester can do better.

https://packagecontrol.io/packages/Http%20Requester

Thom1729:
Would it make sense to improve that package rather than creating a new one?

gaohuia:
> Would it make sense to improve that package rather than creating a new one?

Maybe keep both would be better.  

When people wanna send simple requests, `HttpRequester` would be easy to use, with which even no need to save the request to disk.  If more control is needed or people wanna make something like a document, `SimpleHttpRequester` would be better.

gaohuia:
Hi, just let me know if anything need to add. Looking forward to see it in sublime.

FichteFoll:
>When people wanna send simple requests, HttpRequester would be easy to use, with which even no need to save the request to disk. If more control is needed or people wanna make something like a document, SimpleHttpRequester would be better.

Following this decription, are you sure your package name is appropriate?

Requiring PHP seems like a pretty big downside compared to the other package. Also, because the spawned `php` process needs to access your bundled `.php` file, you must provide a `.no-sublime-package` file to tell PC to unpack your package.

Your `install.py` will get loaded by ST when it most certainly shouldn't. Move it to a subfolder, change its extension, or remove it. (It also uses `sublime` without importing it.)

gaohuia:
Than you, will improve.

gaohuia:
Thank you for your advisement, have done the following changes.

Renamed to HttpUnit. 
Added .no-sublime-package

But I'm not able to get rid of php for now.

PR #7637, opened 2019-07-22
Add MarkdownLink

unlight:
**MarkdownLink** - Convert url to markdown link where description is title from remote url
https://github.com/unlight/sublime-markdown-link

1. Unit exists and tests are passed https://github.com/unlight/sublime-markdown-link/runs/174182925
2. `.no-sublime-package` is necessary because of 3rd party libs, and import does not work in .sublime-package

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -546,6 +546,17 @@
 			]
 		},
 		{
+			"name": "MarkdownLink",
+			"details": "https://github.com/unlight/sublime-markdown-link",
+			"labels": ["markdown", "links", "urls"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "MarkdownLivePreview",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Please consider using [the bs4 dependency](https://packagecontrol.github.io/) instead of vendoring your own version. That way you also don't need to modify `sys.path`, although you revert your changes immediately afterwards to reduce side effects on other plugins.

You don't need `.no-sublime-package`. Please remove it.

FichteFoll:
@unlight ping

unlight:
@FichteFoll

Yes, I'm here. I implemented your suggestions.
But I'm struggling with automated release and unit testing via Github Actions (they changed their syntax and it I could not reproduce it)

I'll ping you when it done. I'll try to solve this by end of the month, if cant. I'll do manual release.

unlight:
@FichteFoll 
Done.
https://github.com/unlight/sublime-markdown-link/blob/master/CHANGELOG.md

PR #7639, opened 2019-07-25
Require version 3.1 for Witness color scheme.

ryanpcmcquen:
This color scheme uses the new JSON format, therefore it is only compatible with `>=3170`.


ryanpcmcquen:
Thank you @FichteFoll!

PR #7640, opened 2019-07-26
Add ViewSetting plugin: https://github.com/fsaad/Sublime-ViewSetting.

fsaad:

### DIFF
--- a/repository/v.json
+++ b/repository/v.json
@@ -327,6 +327,16 @@
 			]
 		},
 		{
+			"name": "ViewSetting",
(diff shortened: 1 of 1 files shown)


FichteFoll:
As a suggestion, you could retrieve all supported settings by reading and parsing the settings file manually via `sublime.decode_value(sublime.load_resource(("Packages/Default/Prefereces.sublime-settings")).keys()` (plus a few settings not in the default file) and then offer them in a selection for a quick panel, for example.

FichteFoll:
You only need to wait for up to 1 hour for the crawler to pick up the new version.

PR #7641, opened 2019-07-27
Add google-java-formatter

ankitshubham97:
Currently, there is no plugin in Sublime Text which provides the functionality of formatting the java code according to Google's google-java-formatter. This formatter is considered standard by many java programmers. This will come in handy for the java programmers who use Sublime Text and are conscious about formatting. Most probably, they would be running this formatter manually as of now to format the code. It would be helpful for such programmers to have a plugin for the same in Sublime Text. I am starting it slow for now, with support for only OSX and Linux(Ubuntu). But I am sure that this project will gain traction and we will make it very resilient soon.

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -1312,18 +1312,6 @@
 			]
 		},
 		{
-			"name": "GoogleJavaFormatter",
-			"details": "https://github.com/ankitshubham97/google-java-format-sublime",
-			"labels": ["google", "java", "formatter"],
-			"releases": [
-				{
-					"sublime_text": "*",
-					"platforms": ["osx", "linux"],
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "Google Reader",
 			"details": "https://github.com/speg/SublimeGReader",
 			"releases": [
@@ -1386,6 +1374,18 @@
 				{
 					"sublime_text": "<3000",
 					"branch": "master"
+				}
+			]
+		},
+		{
+			"name": "GoogleJavaFormatter",
+			"details": "https://github.com/ankitshubham97/google-java-format-sublime",
+			"labels": ["google", "java", "formatter"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": ["osx", "linux"],
+					"tags": true
 				}
 			]
 		},
--- a/repository/g.json
+++ b/repository/g.json
@@ -1312,6 +1312,18 @@
 			]
 		},
 		{
+			"name": "Google Java Formatter",
+			"details": "https://github.com/ankitshubham97/google-java-format-sublime",
+			"labels": ["google", "java", "formatter"],
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


ankitshubham97:
@Thom1729 @wbond I was curious but did I miss something due to which it is not getting into the review process?

FichteFoll:
Your readme claims support for macOS and Linux (Ubuntu), but in this PR you only specified macos. You also say it's for ST2, which has a share of less than 2% of users, afaik. There's nothing in your code that is ST-specific, though, so it's fine.

Do not **ever** use `sys.exit` in a plugin. It will kill the entire plugin runtime.

Since you want to run a separate executable on a file, you need to specify `.no-sublime-package` to tell PC to unpack the package instead of placing it in a `.sublime-package` archive file and you should mention that it requires Java to be installed on your system and in your PATH.

The package will be installed as "Google Java Formatter" via its name field in the package entry, so your hardcoded path of `google-java-format-sublime` will fail.

`self.view.file_name()` already returns a string. You don't need to wrap it in `str`. It may, however, return `None`, which you should check first and then exit early.

FichteFoll:
@ankitshubham97 ping

ankitshubham97:
Recheck

ankitshubham97:
@FichteFoll 
I have tried to address all your concerns.
`Your readme claims support for macOS and Linux (Ubuntu), but in this PR you only specified macos. You also say it's for ST2, which has a share of less than 2% of users, afaik. There's nothing in your code that is ST-specific, though, so it's fine.`
Done

`Do not ever use sys.exit in a plugin. It will kill the entire plugin runtime.`
Done

`Since you want to run a separate executable on a file, you need to specify .no-sublime-package to tell PC to unpack the package instead of placing it in a .sublime-package archive file `
Done

 `you should mention that it requires Java to be installed on your system and in your PATH.`
How do you propose I should do that? Should I check if Java is there in the PATH and if it is not, should I print some error and return?

`The package will be installed as "Google Java Formatter" via its name field in the package entry, so your hardcoded path of google-java-format-sublime will fail.`
Done that but could you please take a look once if I have done it correctly?
 Also, how can I verify if that is done correctly?

`self.view.file_name() already returns a string. You don't need to wrap it in str. It may, however, return None, which you should check first and then exit early.`
Done.

FichteFoll:
> How do you propose I should do that? Should I check if Java is there in the PATH and if it is not, should I print some error and return?

Mentioning it in your readme should be enough. It should be expected from people to read your readme and at least scan for requirements or usage instructions.

> Done that but could you please take a look once if I have done it correctly?

Looks all right.

> Also, how can I verify if that is done correctly?

There is a way to use a local `repository.json` as a package repository source for Package Control, but it's not exactly documented.

FichteFoll:
However, you still need to create a tag with the latest fixes.

FichteFoll:
@ankitshubham97 ping

PR #7642, opened 2019-07-28
Add FlutterEnhancements package

yusdacra:
This package is for basic support for Flutter. It adds hot reload on save, build, run and some other useful commands.


FichteFoll:
You might want to consider using `$XDG_RUNTIME_DIR` instead of the current working directory for your pid file.

PR #7643, opened 2019-07-28
Added TLD Autocomplete plugin to the repo list.

mlew980:
There was no existing plugin that allowed for parsing of .tld (custom java tag library) files and added them to auto completion, including specific attributes they contain. This would improve efficiency when creating .jsp/.jspf files (JSON Servlet pages) and no such plugin exists.

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -2391,6 +2391,17 @@
 			]
 		},
 		{
+			"name": "TLD Autocomplete",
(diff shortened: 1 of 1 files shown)


mlew980:
Added an initial release for the github repo.

mlew980:
Just realized the name contains sublime, will create a new repo and clone the existing one.

FichteFoll:
For the record, we don't care about the repo name, only the package name.

PR #7644, opened 2019-07-28
Added the TLD Autocomplete plugin to the repo list.

mlew980:
The package parses .tld files in a specific folder and uses those for Autocompletion purposes. It also autocompletes the attributes. There is currently not plugins that do this and its' really useful when working with .jsp files when you have 100+ taglibs. I followed the naming conventions and ended up renaming the repo to exclude the word sublime. I created a new repo with initial release.

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -2391,6 +2391,17 @@
 			]
 		},
 		{
+			"name": "TLD Autocomplete",
+			"details": "https://github.com/mlew980/TLD_Autocomplete",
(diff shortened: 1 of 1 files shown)


FichteFoll:
You should aim to make all your menu items available in the Command Palette, so that users can easily and quickly access them with the keyboard only.

You call your `plugin_loaded` file in the main module scope. Don't do that. ST will call this function when its API has loaded.

You should try to use early exits to prevent highly nested control structures, notably in your `on_query_completions`.

FichteFoll:
@mlew980 ping

PR #7645, opened 2019-07-29
base64 decode and encode

dmitrievav:

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -207,6 +207,16 @@
 			]
 		},
 		{
+			"name": "Base64",
+			"details": "https://github.com/dmitrievav/sublime_base64",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Base64 Fold",
 			"details": "https://github.com/DaQuirm/base64-fold",
 			"releases": [


jfcherng:
Isn't it better using https://packagecontrol.io/packages/Codec?
It support far more codes than just base64 and doesn't rely on an `base64` external executable.

dmitrievav:
Hi! 

It is a valid question. Thought, it was not so obvious when I was looking for plugin via package control. Mostly because the name of the plugin is not very relevant to the function I was looking for. So by typing base64 it would be much easy to find this new plugin and improve user experience.

Another argument could be to let people keep their sublime functionality as simple as possible.
So if somebody needs just base64 and nothing else it could be the best solution.

We can see such examples in other text editors, for example: https://marketplace.visualstudio.com/items?itemName=adamhartford.vscode-base64

jfcherng:
Never mind. I am just curious about that `base64` is quite a famous and easy codec algorithm so there should be already a plugin.

---

> So by typing base64 it would be much easy to find this new plugin and improve user experience.

I find `Codec` by seaching `base64` and it's ranked 2 by the way.

image

dmitrievav:
image

jfcherng:
@dmikalova Get your point. It looks like somehow the search results are different :confused:

FichteFoll:
1. You don't need to track whether the readme has been shown when you use Package Control's messages feature, which allows showing a note on install. ([docs](https://packagecontrol.io/docs/messaging))
2. Furthermore, instead of using a custom command to open that file, you could also just use the `open_file` command with `${packages}/Base64/Readme.md`. Same for the changelog.

Other than that, as @jfcherng already mentioned, this functionality is provided by the Codec package already and with much more options to encode/decode as well. I'm not a fan of duplicated behavior, so I'd vote against the addition of this package.

The "Install Package" command from Package Control should primarily be used to actually install packages, not to find them, because the quick panel is limited in what can be searched.

PR #7646, opened 2019-07-29
Add nhb reindent plugin

NickHatBoecker:
The reindentation is not working on MacOS Mojave with several plugins and the bare sublime functionality. So I wrote this plugin.

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -533,6 +533,17 @@
 			]
 		},
 		{
+			"name": "NHB Reindent",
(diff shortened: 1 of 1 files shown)


FichteFoll:
You should load the view's current tab_size setting from `view.settings()`, not from the global preferences file.

You can use `view.replace` instead of `view.erase` and `view.insert`.

Your regex can be simplified to `r"^(\s{1,})"` because it will automatically find the largest match. You should also use a raw string so that escape sequences with backslashes are disabled.

FichteFoll:
@NickHatBoecker ping

PR #7647, opened 2019-08-03
Remove "Yii Framework Docs Search" repository

i-sevostyanov:
This extension does not work anymore.


FichteFoll:
Well, that user has 0 repositories, so ... I guess you could say it doesn't work anymore. Thanks.

PR #7653, opened 2019-08-06
rename "Elm Language Support" to "Elm Syntax Highlighting" / add "Elm Format on Save"

evancz:
## Context

I am the creator of Elm, and I want to have a super reliable Elm plugin that I can point people to from [The Official Guide](https://guide.elm-lang.org/). The initial Sublime Text package for Elm has had many maintainers over the past few years, and as a result, it would benefit from trimming down the features to be more minimal and reliable.

## Design

The design here is to split the most important functionality into two much simpler packages:

- https://github.com/evancz/elm-syntax-highlighting
- https://github.com/evancz/elm-format-on-save

The goal is to end up with (1) an extremely reliable plugin for syntax highlighting that will be easy to install for beginners and (2) have a secondary plugin for people who want a bit more from their workflow.

## Plan

This PR does two things:

1. Rename `Elm Language Support` back to `Elm Syntax Highlighting`
2. Add `Elm Format on Save`

The net effect is that the syntax highlighting part of `Elm Language Support` has been preserved, and the code needed to run `elm-format` on save has been improved quite a bit. All the other parts of the plugin have been trimmed out.

* * *

I have talked this through with @sentience who is the current maintainer of the existing Elm Language Support as listed [here](https://github.com/elm-community/Manifesto/blob/master/maintainers.md), and he has started the process of transferring maintainership to me [here](https://github.com/elm-community/Manifesto/pull/93).

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -603,17 +603,12 @@
 			]
 		},
 		{
-			"name": "Elm Language Support",
-			"details": "https://github.com/elm-community/SublimeElmLanguageSupport",
-			"labels": ["language syntax"],
-			"previous_names": ["Elm Syntax Highlighting"],
+			"name": "Elm Format on Save",
+			"details": "https://github.com/evancz/elm-format-on-save",
+			"labels": ["elm", "elm-format", "formatting"],
 			"releases": [
 				{
-					"sublime_text": "<3092",
-					"tags": "st2-"
-				},
-				{
-					"sublime_text": ">=3092",
+					"sublime_text": "*",
 					"tags": true
 				}
 			]
@@ -622,6 +617,18 @@
 			"name": "Elm Snippets",
 			"details": "https://github.com/rudolfb/elm_snippets",
 			"labels": ["snippets", "elm"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
+			"name": "Elm Syntax Highlighting",
+			"details": "https://github.com/evancz/elm-syntax-highlighting",
+			"labels": ["elm", "language syntax", "highlighting"],
+			"previous_names": ["Elm Language Support"],
 			"releases": [
 				{
 					"sublime_text": "*",
--- a/repository/e.json
+++ b/repository/e.json
@@ -605,7 +605,7 @@
 		{
 			"name": "Elm Format on Save",
 			"details": "https://github.com/evancz/elm-format-on-save",
-			"labels": ["elm", "elm-format", "formatting"],
+			"labels": ["elm", "formatting"],
 			"releases": [
 				{
 					"sublime_text": "*",
@@ -631,7 +631,7 @@
 			"previous_names": ["Elm Language Support"],
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": ">=3092",
 					"tags": true
 				}
 			]


repository/e.json
```diff
@@ -629,6 +624,18 @@
...
 			]
 		},
+		{
+			"name": "Elm Syntax Highlighting",
+			"details": "https://github.com/evancz/elm-syntax-highlighting",
+			"labels": ["elm", "language syntax", "highlighting"],
+			"previous_names": ["Elm Language Support"],
+			"releases": [
+				{
+					"sublime_text": "*",
```
FichteFoll: Since the syntax definition is only for ST3 (and in fact builds >=3092), please adjust this version specifier.

repository/e.json
```diff
@@ -603,17 +603,12 @@
 			]
 		},
 		{
-			"name": "Elm Language Support",
-			"details": "https://github.com/elm-community/SublimeElmLanguageSupport",
-			"labels": ["language syntax"],
-			"previous_names": ["Elm Syntax Highlighting"],
+			"name": "Elm Format on Save",
+			"details": "https://github.com/evancz/elm-format-on-save",
+			"labels": ["elm", "elm-format", "formatting"],
```
FichteFoll: I doubt you'll find any other package that provides the `elm-format` tag, so it might as well be redundant, but won't hurt at least for now.

repository/e.json
```diff
@@ -629,6 +624,18 @@
 				}
 			]
 		},
+		{
+			"name": "Elm Syntax Highlighting",
+			"details": "https://github.com/evancz/elm-syntax-highlighting",
+			"labels": ["elm", "language syntax", "highlighting"],
+			"previous_names": ["Elm Language Support"],
```
FichteFoll: Since you effectively removed functionality from the package, you should add an update message that informs people of the changes and what other package to get. See https://packagecontrol.io/docs/messaging for details.

FichteFoll:
Because the review bot didn't pick up the new packages, here's the review for them when run manually:

---

## Report for elm-format-on-save ######################

### Repository checks ###

No failures

No warnings


### Package checks ###

Reporting 1 failures:
- The package contains 1 Python file(s), but none of them are in the package root and no build system is specified

No warnings


---

## Report for elm-syntax-highlighting #################

### Repository checks ###

No failures

No warnings


### Package checks ###

No failures

Reporting 1 warnings:
- '.sublime-syntax' support has been added in build 3092 and there is no '.tmLanguage' fallback file
    File: src/elm.sublime-syntax


---


TL;DR you need to move plugins to the repository root because they won't get loaded otherwise.

repository/e.json
evancz: That makes sense! I removed it in this commit: https://github.com/wbond/package_control_channel/pull/7653/commits/d57f36b87cf9f5e8a993ff2d0a811e56a3616dc0

repository/e.json
evancz: Good point! I added this documentation in https://github.com/evancz/elm-syntax-highlighting/commit/6486db6b10e0ff45edbcbee530268b1eaf17fd48

repository/e.json
evancz: Adjusted in https://github.com/wbond/package_control_channel/pull/7653/commits/d57f36b87cf9f5e8a993ff2d0a811e56a3616dc0

evancz:
Thank you for running the reviews manually!

I made the changes suggested by you and by the robot.

And thank you for the link about Build Systems! I have been thinking that the final part of this set might be a package called something like `Elm Make Shortcut` that just shows errors in the file you happen to be looking at. I'll need to do some work to ensure that I can get it to a level of quality I'd be comfortable recommending to others though!

P.S. Would I use [`st_package_reviewer`](https://github.com/packagecontrol/st_package_reviewer) to run it myself locally? Happy to do so if it helps!

FichteFoll:
Yes, `st_package_reviewer` is the tool for reviewing packages for ST. You can specify a file path or a repo URL, where in the latter case it also ensures that you have a properly formatted tag and fetches that.

Your changes are looking good, so I'm merging. :rocket:

evancz:
Great, thank you so much!

PR #7655, opened 2019-08-08
added BrailleMusic

martin-mueller:

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -1323,6 +1323,16 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
Seeing how your readme is in English, you should probably adjust the captions of the commands file to be in English as well.

FichteFoll:
@martin-mueller please add a minumum version requirement and remove the ctrl+r binding.

FichteFoll:
@martin-mueller ping

PR #7658, opened 2019-08-14
Add new Xdebug package.

ryanpcmcquen:
This package exists to supersede/replace the unmaintained and broken 'Xdebug Client' (https://github.com/martomo/SublimeTextXdebug) for Sublime. It no longer works with any version of Xdebug released in the last year, and the author seems unreachable, many PRs have sat stagnant against the repo for over a year. I have made attempts to contact the author to no avail. This package merges all standing PRs without conflicts which resolve the showstopper bugs in the current version.

### DIFF
--- a/repository/x.json
+++ b/repository/x.json
@@ -77,6 +77,17 @@
 			]
 		},
 		{
+			"name": "Xdebug Client Plus",
+			"details": "https://github.com/ryanpcmcquen/SublimeTextXdebugPlus",
+			"labels": ["php", "xdebug"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Xkcd",
 			"details": "https://github.com/eivind88/xkcd",
 			"releases": [


ryanpcmcquen:
@packagecontrol-bot, the requested change has been implemented.

ryanpcmcquen:
@thom1729 or @ficthefoll, thoughts?

FichteFoll:
Since the original author/maintainer @martomo seems to be AWOL, I suggest replacing his package instead of adding an alternative. That way current users are converted to the updated and maintained package as well.

@martomo, by our conventions, you have 2 weeks to object to @ryanpcmcquen taking over your slot.

ryanpcmcquen:
@FichteFoll, if @martomo does not respond, do I need to rename my package to match the old name?

FichteFoll:
You would remove your addition and replace the URL of the existing xdebug entry (and add `"tags": true`).

martomo:
@FichteFoll Looking at the proposed changes in @ryanpcmcquen's package it seems that the any older Xdebug installations (prior to 2.7.0) would no longer work with the Xdebug Client package, if it were to be replaced, due to schema in Xdebug version 2.7 being changed from **http** to **https**.

Quickly checking [PECL statistics](https://pecl.php.net/package-stats.php?cid=25%27&pid=214) for Xdebug it shows that version 2.5.5 still has about 25k installs per month, which is just a little less than the total average 30k installs per month for all 2.6 versions.

Perhaps for the time being this package proposed by @ryanpcmcquen can serve as an alternative for newer Xdebug installations (version 2.7.0 or newer).

More appropriate approach would be to implement support for older and newer versions, plan is to take some time for Xdebug Client this weekend and see whether support for both older/newer installations can be added in the existing Xdebug Client package.

FichteFoll:
Thanks for replying.

@ryanpcmcquen could you comment on this new information?

ryanpcmcquen:
@FichteFoll, it would be great if @martomo fixed the package to work with newer versions. I didn't have any troubles with my package and Xdebug 2.6.x, so I'm not sure the protocol change has the effect stated ...

When I view the linked statistics, it looks like each major version has a little over one million installs, which makes me think that people are regularly upgrading the package, but I don't see the monthly download stats mentioned. I'm fine with having both packages or not having this one at all if @martomo finds a good fix.

martomo:
There are 2 major issues with Xdebug Client version 1.1.0 which would not work properly with newer versions of PHP/Xdebug. I've resolved the XML parsing issue already with version 1.1.1, solution for the other issue regarding Xdebug switching from **http** to **https** is clear, just having some problems with integration tests passing on macOS and hope to have that also resolved this week.

ryanpcmcquen:
It appears that @martomo, has resolved all of the outstanding issues as well as made some other excellent optimizations, so I don't see any reason to have my package available anymore, I will probably keep the repo up in case anyone is using it, but @FichteFoll, we can close this for now.

@martomo, thank you for all your work here!

PR #7659, opened 2019-08-15
Add MarkSearch - search through find results while keeping them outlined

robertcollier4:


repository/m.json
```diff
@@ -654,6 +654,17 @@
 				}
 			]
 		},
+		{
+			"name": "MarkSearch",
+			"details": "https://github.com/robertcollier4/MarkSearch",
+			"labels": ["search find mark"],
```
FichteFoll: Labels should be separate list items, not separated by spaces.

PR #7661, opened 2019-08-16
added Urtext

nbeversl:

### DIFF
--- a/repository/u.json
+++ b/repository/u.json
@@ -549,6 +549,17 @@
 			]
 		},
 		{
+			"name": "Urtext",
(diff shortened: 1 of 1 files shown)


FichteFoll:
[Your imports](https://github.com/nbeversl/urtext_sublime/blob/dda1f3f0aa8deabfecc884c98e8ef8f923eaf14d/sublime_urtext.py#L28-L32) should be relative because the package will be installed as "Urtext" with an upper-case U.

The link to documentation in your readme doesn't work, and an example file or screenshot would surely be useful for advertisement.

FichteFoll:
@nbeversl ping

PR #7662, opened 2019-08-19
Update details url of Alive Snippets package

alivevietnam:
Alive Snippets:
- Change details URL
- Change Sublime Text required version
- Change platforms

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -613,14 +613,14 @@
 		},
 		{
 			"name": "Alive Snippets",
-			"details": "https://github.com/alivevietnam/alive-sublime-snippets",
-			"author": "Alive Vietnam ",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Why remove linux?

You should remove the .pyc file because it is only compatible with a specific python version, only used by ST2 and generated on the fly as necessary anyway.

You can use `global settings` instead of `globals()['settings']`. Actually, you don't even need the global because you never use it outside of `init`.

What does that code even do? It seems you use it to enable or disable snippets, but the procedure is quite poor because it
1. Doesn't work on ST3 due to the package being archived there.
2. Is error prone to user config
3. Does a two-step rename instead of only renaming a snippet file when it has been renamed.


I'll merge anyway, since the package is already on the list, but note that as it is it won't work on ST3. You could add `.no-sublime-package` to make it work, however.

PR #7663, opened 2019-08-19
add golang package

yields:

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -1201,6 +1201,19 @@
 			]
 		},
 		{
+			"name": "Golang",
+			"description": "An opinionated plugin for Go",
+			"details": "https://github.com/yields/sublime-go",
+			"author": ["yields"],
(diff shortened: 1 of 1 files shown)


FichteFoll:
You don't need to periodically force-push to the PR to keep it updated. We go through every PR when reviewing anyway.

yields:
> You don't need to periodically force-push to the PR to keep it updated. We go through every PR when reviewing anyway.
> 
> (And I go by submission date.)

Apologies, I remember I read that somewhere I can close and re-open / push, IIRC I've read that here: https://packagecontrol.io/docs/submitting_a_package

Seems like there's no mention of closing/re-opening a PR, so I must've read that somewhere else.

Thanks for merging.

FichteFoll:
Reopening is mentioned in the PR template on this repo and only useful for triggering a review from @packagecontrol-bot.

yields:
ah gotcha, my bad, sorry about that!

PR #7664, opened 2019-08-19
Add MakeTargets

dusk125:
MakeTargets automatically generates a sublime-build build system based on a Makefile in your project. This build system will have variants created for each of the Make targets that are found in your Makefile. Useful for those who wish to build more than Sublime's default 'Make' and 'Make Clean' targets.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -264,6 +264,18 @@
 			]
 		},
 		{
+			"name": "MakeTargets",
+			"details": "https://github.com/dusk125/sublime-maketargets",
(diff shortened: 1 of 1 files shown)


FichteFoll:
When overriding a command's `__init__` it is safer to catch and pass `*args, **kwargs` than specify the actual arguments, because those aren't documented and may change at any point in the future.

If you just requested most of the settings you use when you need them, you wouldn't need the many settings change listeners and could just make decisions based on their value at the time.

Also, have you considered adding the build systems to the current project instead using `window.set_project_data()`?

PR #7665, opened 2019-08-20
create package gototag

dclnet:
submit a package: gototag (quickly jump to any Tag location)

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -1766,7 +1766,7 @@
 			"details": "https://github.com/dclnet/gototag",
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": "*", 
 					"tags": true
 				}
 			]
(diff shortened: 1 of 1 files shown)


dclnet:
fix the error

FichteFoll:
Your package will be installed as "GotoTag", so [this hard-coded path](https://github.com/dclnet/gotoTag/blob/master/GotoTag.py#L8) will not work on case-sensitive filesystems. You may use the module-global `__package__` instead.

Instead of `currentLine.contains(regions[0]) or regions[0].contains(currentLine)` you can use `regions[0].intersects(currentLine)`.

Going to tag id 0 on `ctrl+alt+1` seems unintuitive.

FichteFoll:
@dclnet ping

dclnet:
Sorry, I was so busy that I forgot to reply to you.
I'll fix it later and submit it

PR #7666, opened 2019-08-20
Adds ruby-idiomatic-snippets channel.

mark-d-holmberg:
This package overrides the default ruby snippet for "begin .. rescue .. end" to not rescue Exception but rather StandardError, as rescuing from "Exception" is frowned upon and not the idiomatic way.

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -2011,6 +2011,16 @@
 			]
 		},
 		{
+			"name": "Ruby Idiomatic Snippets",
+			"details": "https://github.com/mark-d-holmberg/ruby-idiomatic-snippets",
+			"releases": [
+				{
+					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


FichteFoll:
If it's a general improvement to the default snippets and the current version strictly inferior, I'm sure a PR at https://github.com/sublimehq/Packages would be accepted. Other than that, it appears Sublime Hq is generally looking to remove snippets from the default packages in the future, so new snippets wouldn't be accepted. However, fixing or changing existing ones is fine.

mark-d-holmberg:
I will definitely do that.

mark-d-holmberg:
https://github.com/sublimehq/Packages/pull/2183

FichteFoll:
Thanks. I'm closing, then.

PR #7668, opened 2019-08-21
Added the Scopus Syntax Highlighter package

Syncrossus:
Scopus is a database of scientific articles that provides an advanced search function which uses a particular syntax. This package defines highlighting for that syntax.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -534,16 +534,6 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
You should refine your scope names a bit and add matches for punctuation characters. Refer to https://www.sublimetext.com/docs/3/scope_naming.html, if you haven't already.

PR #7670, opened 2019-08-23
Add Kirikae

dvcrn:
Plugin to quickly jump between open tabs. Used to treat views as buffers similar to how vim / emacs handles them

### DIFF
--- a/repository/k.json
+++ b/repository/k.json
@@ -255,6 +255,17 @@
 			]
 		},
 		{
+			"name": "Kirikae",
+			"details": "https://github.com/dvcrn/sublime-kirikae",
+			"labels": ["tab"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Kivy Language",
 			"details": "https://github.com/ivlevdenis/kivylng",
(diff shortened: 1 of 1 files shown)


dvcrn:
Fixed the keybindings

dvcrn:
Anything I can do to get this merged?

FichteFoll:
How does it differ from the [fork origin](https://github.com/rajeshvaya/Sublime-Extended-Tab-Switcher)? Would it make sense to replace the old package with this one if it is actively maintained?

dvcrn:
The main idea behind this fork is to correctly implement moving and manipulating of tabs independently of which group they are (if set) similar to how emacs/vim are handling things. For example, when the tab bar is hidden, all tabs can be treated as buffers (which is what I do). If I'm in group 2 but the tab I want to switch to is in group 1, it will get moved transparently to the current group and give focus. 

The fork also fixes the default keybindings and correctly implements previews of tabs on select in different groups (including the part/line that was in focus previously)

Since it's a fork, the core functionality is the same so I'd be ok with replacing the original but the vision I have is a bit more than just a tab switcher

FichteFoll:
>I have is a bit more than just a tab switcher

With this and quite a bit of changed functionality, I guess it makes more sense to add it as a separate package instead of replacing the current one.

PR #7672, opened 2019-08-25
Remove packages of ihodev

deathaxe:
This commit removes ihodev's obsolete/missing packages from packagecontrol.io as I was asked to do so by him.

### DIFF
--- a/channel.json
+++ b/channel.json
@@ -49,7 +49,6 @@
 		"https://raw.githubusercontent.com/Hexenon/FoxCode/master/packages.json",
 		"https://raw.githubusercontent.com/hgraca/sublime-text-2-php-refactor/master/packages.json",
 		"https://raw.githubusercontent.com/himynameisjonas/sublime-ia-writer-opener/master/packages.json",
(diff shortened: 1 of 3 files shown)


jfcherng:
Sad :cry:

deathaxe:
All original local and upstream sources of Boxy and DA UI got deleted. Even the original author doesn't have them anymore. He sent me his a local repo of DA UI, but it only contains LFS links to the no longer existing github repo.

The conclusion therefore is:

1. We could restore the plain Boxy theme from a fork as is, but I don't feel able to maintain it.
2. We are allowed to publish the DA UI, but maintaining is even more complicated due to the closed source structure. Decompiling would require a lot of polishing afterwards. This is why I created _Theme - DAneo_ which clones the look and all the animations of DA UI. It is more or less a copy of the most basic variants of DA UI, with some smaller fixes.

None of the themes suppurt the latest features of ST (git integration, variable support, ...) and won't ever see them.

I'll continue with maintaining _A File Icon_ as it provides a couple of workarounds for some of ST's issues related with `file_type_...` preferences causing missing icons with some themes, but as there is _Theme - DAneo_ as successor of Boxy/DA UI, ... .

PR #7675, opened 2019-08-29
Add Multicommand package

dvcrn:
Plugin for executing multiple commands per keybinding

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -2505,6 +2505,17 @@
 			]
 		},
 		{
+			"name": "Multicommand",
+			"details": "https://github.com/dvcrn/sublime-multicommand",
+			"labels": ["keybinding"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "MultiEditUtils",
 			"details": "https://github.com/philippotto/Sublime-MultiEditUtils",
 			"labels": ["text manipulation", "code navigation", "text selection"],


deathaxe:
How does this package differ from the existing [Chain Of Command](https://packagecontrol.io/packages/Chain%20of%20Command)?

dvcrn:
Looks similar, yeah but Chain Of Command is also abandoned (couple open issues, PR and no license). I don't think another option hurts

dvcrn:
Anything I can do to get this merged?

FichteFoll:
Since you removed the `context` functionality, as ST automatically resolves it for your when you use `Window.run_command`, you should also remove mentions of it in the readme and comments (and mentions of evernote, which is entirely unrelated now).

dvcrn:
Thanks will do!

FichteFoll:
Regarding Chain of Command, the only issue worth resolving seems to be https://github.com/jisaacks/ChainOfCommand/issues/10 (and probably https://github.com/jisaacks/ChainOfCommand/pull/7), which your version *does* address. It not being maintained isn't really an issue since the functionality fundamentally won't change in the foreseeable future.

Because the existing package a) is really basic, b) has probably been referenced a lot in various places over time as a solution for a common problem, and c) both "chain of command" and "multi command" are names you can find when searching this problem, I am okay with having another package fulfill the same job. I could also see replacing Chain of Command, as long as it remains backwards-comaptibl.

dvcrn:
Cleaned up the README and examples to remove context and not needed references

PR #7679, opened 2019-09-01
Added the FixedToggleFindPanel package

evandrocoan:


FichteFoll:
Ctrl+q is a default binding to close the application. Please use something else for your key chords.

evandrocoan:
That keybind is pure evil at at the same time, a very handy/easy to access keybind to make combos. I prefer close it instead of changing it.

FichteFoll:
> That keybind is pure evil at at the same time

I use it very frequently to maintain my ST session on restarts that consists of at least 2 windows. Regardless, it is not in the scope of packages to decide which key bindings a user needs or not.

evandrocoan:
I kept hitting it by accident, closing my Sublime Text and losing my undo history of things, so evil for my tricky fingers.

To close Sublime Text (without losing my windows), I always hit `Alt + F -> Up -> Enter`:
bug2

There is not way I hit/type that by accident like `Ctrl+Q`, which is quite close to `Ctrl+W`.

PR #7681, opened 2019-09-01
Added the ToggleSettingsInBatch package

evandrocoan:
Does everything this package https://github.com/lowerworld/SublimeToggleSettings does, and much more. It is complete rewrite of that old package.

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -2644,6 +2644,16 @@
 			]
 		},
 		{
+			"name": "ToggleSettingsInBatch",
+			"details": "https://github.com/evandrocoan/ToggleSettingsInBatch",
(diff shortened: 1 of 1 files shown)


evandrocoan:
I replaced the default keybindings for zoom with this package own keybindings for zoom.

The difference, is depending on the combination, you will change the zoom only for the current view, window or global. Does not make sense to have several keybindings doing exactly the same thing, i.e., always changing the zoom globally.

FichteFoll:
You should mention that the package adds and replaces key bindings for changing font size.

evandrocoan:
Some other time I will do it. Closing it for now.

PR #7682, opened 2019-09-01
Added the ClearCursorsCarets package

evandrocoan:


evandrocoan:
It is required to mask contents, because that is what this package does. For example, the default behavior is to close any open panel before running `single_selection`. This packages makes it backwards. First it run `single_selection`, only then it closes the open panel.

FichteFoll:
This is the package's primary purpose, so it's okay to override key bindings when it is properly communitcated.

PR #7684, opened 2019-09-02
Added the OverrideEditSettingsDefaultContents package

evandrocoan:


FichteFoll:
I don't exactly foresee a lot of users for this package, but I guess installing a package is still simpler than putting a single plugin in your User folder. :man_shrugging:

PR #7687, opened 2019-09-03
add new KSQL syntax highlighter package

blueedgenick:


blueedgenick:
semver tag added to source repo

PR #7689, opened 2019-09-04
add new KSQL syntax highlighter

blueedgenick:


FichteFoll:
`captures: {0: scope.name}` is equivalent to `scope: scope.name`.

PR #7692, opened 2019-09-13
Update: PresideCMS package readme and version dependency

pixl8-brayden:
Update PresideCMS package readme link and supported version.


FichteFoll:
You bundle a .pyc file, but ST2 and ST3 use different Python versions, which means this is likely not going to work for ST2. I suggest abandoning ST2 support anyway since it has a user base of less than 1% at this point, probably.

pixl8-brayden:
I see. Thanks for the advice. I had updated so it only support for ST3.

PR #7694, opened 2019-09-16
Add package: Lit-Element Syntax Highlighting

JeremyBernier:
This package provides syntax highlighting for Google's `lit-element` and `lit-html` libraries. `lit-html` enables HTML templating within tagged template literal strings, but without syntax highlighting all the developer will see is a string. I did not find any libraries providing syntax highlighting for `lit-html`, so I created my own.

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -1149,6 +1149,18 @@
 			]
 		},
 		{
+			"name": "LitElement Syntax Highlighting",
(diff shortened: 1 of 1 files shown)


JeremyBernier:
Added a tag, hopefully everything is good now

JeremyBernier:
Is everything good now? I've added a version tag already (https://github.com/JeremyBernier/LitElement-Syntax-Highlighting/releases/tag/v1.0.0)

PR #7695, opened 2019-09-21
Add package: Lit-Element Syntax Highlighting

JeremyBernier:
This package provides syntax highlighting for Google's lit-element and lit-html libraries. lit-html enables HTML templating within tagged template literal strings, but without syntax highlighting all the developer will see is a string. I did not find any libraries providing syntax highlighting for lit-html, so I created my own.

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -1149,6 +1149,18 @@
 			]
 		},
 		{
+			"name": "LitElement Syntax Highlighting",
+			"details": "https://github.com/JeremyBernier/LitElement-Syntax-Highlighting",
+			"labels": ["language syntax", "javascript", "lit-element", "lit-html"],
+			"author": "Jeremy Bernier",
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"tags": true
+				}
+			]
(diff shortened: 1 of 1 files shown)


JeremyBernier:
It's been nearly 2 months, is there any intention of reviewing this PR?

FichteFoll:
Yes, sorry for the wait. I've been really busy.

> Note: Only the lit-element.sublime-syntax file is valid. The tmLanguage file will not work (only included it because I had to to get the PR approved by the robot). Unless you're using an older version of Sublime Text, this should not matter.

If it's not valid anyway, just change the minimum ST version requirement to 3092. See also [the releant section in the 'Results help' page](https://github.com/packagecontrol/st_package_reviewer/wiki/Package-checks#sublime-syntax-support-has-been-added-in-build-3092-).

Also, the commit that added a license is not included in any tag yet.

JeremyBernier:
No worries, thanks for getting back to me.

Just updated the minimum ST version, removed the tmLanguage file, and added the license to the tag. Let me know if I missed anything.

PR #7699, opened 2019-09-27
Add "Fluent" package

SirNickolas:
This package provides syntax highlighting and completions for the Fluent file format (`.ftl`). [Fluent](https://projectfluent.org) is an asymmetric localization system designed at Mozilla.

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -1283,6 +1283,17 @@
 			]
 		},
 		{
+			"name": "Fluent",
+			"details": "https://github.com/SirNickolas/SublimeFluent",
+			"labels": ["language syntax", "auto-complete"],
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"tags": true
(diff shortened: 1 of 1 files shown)


SirNickolas:
`Default` bindings invoke `Default/Add Line in Braces.sublime-macro` on `["enter"]` and `["shift+enter"]` when the cursor is placed between the braces: `{}`. This package does [the same][keymap] when the cursor is placed between an arrow and a brace: `->}` — and only in `meta.interpolation.fluent` scope. So this neither conflicts with the default bindings nor produces unwanted effects in other syntaxes.

[keymap]: https://github.com/SirNickolas/SublimeFluent/blob/master/Src/Default.sublime-keymap#L46

FichteFoll:
Thanks for the clarifications. Not sure what the point of adding a snippet binding to `\u` and `\U` is, though.

PR #7701, opened 2019-10-03
Add MAC-1 Syntax

davidholton:
Added syntax support for the MAC-1 assembly language

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -34,6 +34,19 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


davidholton:
@packagecontrol-bot

Forgot to create a release

FichteFoll:
You should change the pop match for the comment to `\n` so it also captures the newline and prevents the auto complete popup from opening at the end of it.

PR #7703, opened 2019-10-03
Added the Scopus Query Compiler package

Syncrossus:
Scopus is a database of scientific articles that provides an advanced search function with its own syntax. The online tool is mediocre for making elaborate queries. This package allows users to write a Scopus query as a combination of many small, manageable queries, and automatically compiles them into a single, longer query.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -534,16 +534,6 @@
 			]
 		},
 		{
-			"name": "Scopus Query Compiler",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Please resolve the merge conflict.

Syncrossus:
Thank you for bringing the conflict to my attention, it has been resolved.

PR #7705, opened 2019-10-05
Add Singularity Definition File Syntax

MorganRodgers:
Adds a new syntax for Singularity definition files. Singularity is a containerization format popular in HPC / academic settings.

https://sylabs.io/


FichteFoll:
You should add `\b` anchors to your keyword matches to ensure you don't partially match a word.

Have you tested whether your `file_extensions` works? Because afaik that does not support regular expressions.

MorganRodgers:
Thanks for the merge and the feedback @FichteFoll. I'll make some updates when I get a chance.

PR #7708, opened 2019-10-14
NetYCE sublime syntax highlighting

bartdorlandt:
This is a syntax highlighter package for NetYCE templates and scenarios.


bartdorlandt:
Apologies. I messed up while trying to fix things.
Added a new PR #7753 instead.

PR #7710, opened 2019-10-15
Added "python-docx" dependency

writeml:
Added an entry for "python-docx" to "dependencies.json"


FichteFoll:
You specify `>=3000` in the range, but also store the dependency insite an `all` folder. Sice ST2 is dead anyway, I'll merge this, but you *may* want to adjust your folder structure.

PR #7714, opened 2019-10-19
Removing clarkewd's Crontab package

nayyarv:
Hi, 

I'd like to deprecate @clarkewd's 'Crontab Syntax' package. It simply doesn't work as advertised anymore. It has been 5 years since it's last update, while the other Crontab package just had a major update last week and is significantly more feature complete.

This is the image on the project
Claim

vs what the old package looks like when installed on a vanilla Sublime install
Old Crontab Package

When searching for Crontab packages for Sublime, you have 2 options presented to you, both at v1.0.0 and seemingly equal, but the difference is stark.

CrontabHighlightSample

I think we want everyone using the old package to move onto the newer one. I still see a handful of installs for this package on https://packagecontrol.io/packages/Crontab%20Syntax and I can't imagine anyone installing this package is happy with it. 

I don't know what the protocol is to deprecate/remove a package, but I'd like to kick off the conversation

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3996,17 +3996,6 @@
 			]
 		},
 		{
-			"name": "Crontab Syntax",
-			"details": "https://github.com/clarkewd/SublimeCrontab",
-			"labels": ["language syntax"],
-			"releases": [
-				{
-					"sublime_text": "*",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "Crypto",
 			"details": "https://github.com/mediaupstream/SublimeText-Crypto",
 			"releases": [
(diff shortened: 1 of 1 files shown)


FichteFoll:
Yes, this seems like a good idea to me. I suggest adding the removed package's name to the superior Crontab package so that users of the other package are migrated as well. Maybe with an update message from @michaelblyons.

By our guidelines, I would not accept a package like this when a definitely superior option already exists. However, since both packages are already on the list, you have 2 weeks to object to this change, @clarkewd.

nayyarv:
Umm, I added the old package name under the new one, but it's now saying both have been removed. Have I done something wrong?

FichteFoll:
Your changes are correct. The review script probably just didn't detect this correctly.

nayyarv:
I think it's been enough time for a merge

PR #7716, opened 2019-10-19
update package "Bootstrap 4 Snippets" with updated, not deprecated repository

alexanderankin:
The url at https://github.com/mdegoo/sublime-bootstrap4 redirects now to https://github.com/degouville/sublime-bootstrap4 which is clearly marked "DEPRECATED | Bootstrap 4 - Sublime Plugin", so i am proposing that this be replaced with my fork - https://github.com/alexanderankin/sublime-bootstrap4.

This snippet workflow is part of my brain like an inextricable cancer. please i am suffering.

after i am opening this package, i am incorporating all of my local changes, and adding release tags to be version matching with bootstrap releases.

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -1080,7 +1080,7 @@
 		},
 		{
 			"name": "Bootstrap 4 Snippets",
-			"details": "https://github.com/mdegoo/sublime-bootstrap4",
+			"details": "https://github.com/alexanderankin/sublime-bootstrap4",
 			"labels": ["snippets","Bootstrap","Bootstrap4"],
 			"releases": [
 				{


alexanderankin:
thanks in advance for taking the time for this! I am not sure how to notify the package management system that there is a new release in my repository but i have added tags to my repository and added messages under the messages folder.

FichteFoll:
I see you have submitted a PR before and it was accepted, but no new tag has been pushed.

Techically, we allow all current owners of a package a 2 week grace period to object to their package being replaced, but the "deprecated" note indicates to me that we probably don't need to do that here. I'm ccing @degouville anyway as an FYI.

I'll note, however, that your latest tag doesn't point to your latest commit either. https://github.com/alexanderankin/sublime-bootstrap4/compare/v4.3.1...master

PR #7717, opened 2019-10-19
Add package: build2

lyrahgames:


FichteFoll:
I strongly advice against using single quotes in YAML strings because of all the escape sequences you need to regular expressions. Either don't use quotation marks at all or double quotes.

PR #7720, opened 2019-10-20
Submitting my package to package control

renidantass:

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -3462,6 +3462,17 @@
 			]
 		},
 		{
+			"name": "TypeScript Transpiler",
+			"details": "https://github.com/renidantass/sublime-typescript-transpiler",
+			"labels": ["typescript", "transpiler"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "TypescriptCompletion",
 			"details": "https://github.com/RonanDrouglazet/TypescriptCompletion",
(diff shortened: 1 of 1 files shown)


FichteFoll:
This looks a lot like a build system in the form of a plugin. Have you checked out build systems before? Aren't there any other packages doing this already?

FichteFoll:
@renidantass ping

renidantass:
Yeah, this is almost a build system... (a error checker on output to show on status bar is a difference)

FichteFoll:
So why not use an actual build system? Or is it just the status bar text you want sepcifically here?

renidantass:
That way is more easy on daily use. As build system, i would have 2 build system (one for embedded tvs and another to typescript).

FichteFoll:
Note that you can add key bindings for any build system (variant) using the `build` command with the `build_system` and `variant` args, if quick switching between the two is an issue.

FichteFoll:
I'm still not convinced this plugin has any reason for existance in its current form. Everything you do in the plugin can be done in a build system, except for printing to the status bar (which isn't even cleared, so it might as well use an output panel for the user to actually dismiss it) and running on save instead of on a separate key (e.g. ctrl+b) on-demand (which we also have https://packagecontrol.io/packages/SublimeOnSaveBuild for).

Besides, there is still the bug with paths containing spaces.

renidantass:
Interesting! I'll fix the bug with spaces anyway, I understand the mistake in project. Thanks for advice!

PR #7725, opened 2019-10-23
Specman e-language syntax for ST3

blutsvente:
Specman e-language syntax for ST3

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -3060,7 +3060,7 @@
 				}
 			]
 		},
-   	{
+		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
You should consider adding punctuation scopes. See https://www.sublimetext.com/docs/3/syntax.html.

blutsvente:
Thanks for merging! 
> You should consider adding punctuation scopes
Any feedback welcome! I believe I used punctuation scopes all over the place. But surely improvements can be done.

PR #7732, opened 2019-10-26
Add Render Tree package

cpeccei:
Adds Sublime Text 3 plugin to render tree structures as text. Uses the anytree python library for rendering.


FichteFoll:
I suggest fetching the tab size from `view.settings().get("tab_size")` instead of hardcoding it to 4 and also handling tab characters.

PR #7734, opened 2019-10-28
Add "RGBDS" package

ISSOtm:
This package provides a syntax definition, snippets/completions and a few more, all for [RGBDS](https://github.com/rednex/rgbds), currently the most popular Game Boy assembly toolchain.

There is already [another RGBDS package](https://github.com/shinmai/sublime_z80_rgbds/) (#7583), but it's very barebones, and mistakenly states RGBDS assembles z80 assembly (it doesn't, the Game Boy's CPU is [neither a 8080](https://youtu.be/HyzD8pNlpwI?t=779), [neither a z80](https://youtu.be/HyzD8pNlpwI?t=830)). Also, this package is much more complete.

I have made some choices about this package I'm not sure about, so feedback would be really appreciated, and I will do my best to correct any mistakes.

I also intend to provide snippets for common, well, snippets of code; I have asked around whether they should be included in this package, and the agreed answer was that they shouldn't unless they're relevant to the syntax itself. For this reason, I will submit another package containing those if this one gets accepted.

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -1464,6 +1464,7 @@
 			"name": "RGBDS",
 			"details": "https://github.com/ISSOtm/sublime-RGBDS",
 			"labels": ["language syntax", "game boy", "rgbds"],
+			"previous_names": ["Z80 Assembly (RGBDS)"],
 			"author": "Eldred Habert",
 			"releases": [
 				{
--- a/repository/r.json
+++ b/repository/r.json
@@ -1461,6 +1461,18 @@
 			]
 		},
 		{
+			"name": "RGBDS",
+			"details": "https://github.com/ISSOtm/sublime-RGBDS",
+			"labels": ["language syntax", "game boy", "rgbds"],
+			"author": "Eldred Habert",
+			"releases": [
+				{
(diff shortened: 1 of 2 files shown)


FichteFoll:
Since the other package's readme states that the included syntax definition only provides a bare minimum and yours definitely looks more completed, I'm inclined to suggest replacing the other package. Unless you say that the Z80 syntax highlighted by that package is significantly different from the one you are targeting, or @shinmai objects within 2 weeks.

ISSOtm:
Looking at the syntax definition, it appears to be targeting the exact same syntax. I will wait two weeks, then!

ISSOtm:
I have noticed a bug with my syntax definition (nested `REPT` blocks end all at the first `ENDR`); since no objection has been brought up, would it be possible to merge this once this bug gets fixed?

FichteFoll:
Yes, you may go ahead with preparing the merge. To do that, either remove your addition and replace the other package's details url, or remove the other package and add its name to your `previous_names` array.

ISSOtm:
Ready to merge! :tada:

PR #7738, opened 2019-11-03
Add Folders Switch

kitzberger:
This plugin enhances Sublime Text 3 to switch between different sets of folders for a given project file. Please see README for details.


kitzberger:
What's the next step here?

PR #7740, opened 2019-11-05
Add CodeCounter.

absop:
## Functions:
- Side Bar Code Counting.
- Side Bar File Size showing.
- Counting directory or file using command "CodeCounter: count dir".
- <kbd>o</kbd> or <kbd>Enter</kbd> open files.
- <kbd>d</kbd> to show a detail counting information for a language.
- mouse click showing.

## More information please read the README file.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1944,6 +1944,18 @@
 			]
 		},
 		{
+			"name": "CodeCounter",
+			"details": "https://github.com/absop/CodeCounter",
+			"labels": ["analytics", "statistics", "status bar", "utilities"],
+			"author": "absop",
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Code Foo",
 			"details": "https://github.com/markandey/codefoo",
 			"releases": [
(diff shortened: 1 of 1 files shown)


FichteFoll:
> I added the file .no-sublime-package because there are several share object file compiled by C compilers under the package directory are used.

Ah, that's a fair use case. I just didn't go that deep, I think.

However, you don't seem to provide a binary for macOS and only for one architecture (most likely only x86_64), so you should adjust the platform restrictions for your package.

absop:
> > I added the file .no-sublime-package because there are several share object file compiled by C compilers under the package directory are used.
> 
> Ah, that's a fair use case. I just didn't go that deep, I think.
> 
> However, you don't seem to provide a binary for macOS and only for one architecture (most likely only x86_64), so you should adjust the platform restrictions for your package.

OK, I have provisionally adjusted the platform restrictions for this package.

absop:
> > I added the file .no-sublime-package because there are several share object file compiled by C compilers under the package directory are used.
> 
> Ah, that's a fair use case. I just didn't go that deep, I think.
> 
> However, you don't seem to provide a binary for macOS and only for one architecture (most likely only x86_64), so you should adjust the platform restrictions for your package.

OK, I have provisionally adjusted the platform restrictions for this package.

PR #7747, opened 2019-11-07
fd-rspec-snippets

aravindaakash:
1. Create a template for a `spec` file used in Rspec
2. Convert the selected lines to Rspec `it` statement

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -374,6 +374,17 @@
 			]
 		},
 		{
+			"name": "Fd rspec snippets",
+			"details": "https://github.com/aravindaakash/convert-to-it",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Unfortunately, you preview gifs are very small and don't link to higher resolution versions either.

Your `convert_to_it` command can be substituted with a snippet that uses the `$SELECTION` variable for the selected text respectively. You can bind that to a key directly using the `insert_snippet` command. Refer to the default key bindings for examples.

I also generally suggest using a separate repo for ST since there is seemingly zero overlap between the two package resources.

FichteFoll:
@aravindaakash ping

PR #7749, opened 2019-11-08
Add python_live_coding package

donkirkby:
Live Coding in Python lets you run your Python code as you type it. For example, this code prints a greeting to my friend, Alice.

    name = 'Alice'                         |name = 'Alice'
    print('Hello, ' + name + '!')          |print('Hello, Alice!')

The plugin adds a second window on the right hand side that shows what happens when you run your Python code. If you change the name in your code, the display on the right updates as you type.

    name = 'Bob'                           |name = 'Bob'
    print('Hello, ' + name + '!')          |print('Hello, Bob!')

The plugin also includes preview modes for matplotlib graphs and Pyglet user interfaces.

I have run the channel repository tools tests, and they pass. I chose manual hosting, because the project contains other plugins as well as the Sublime Text plugin.

For a more detailed introduction to the plugin, see the [getting started] page. Thanks for reviewing my plugin.

### Items from Review Comments
* [x] Convert settings dialog from `open_file` to `edit_settings`.
* [x] Add `python_live_coding` prefix to generic command names.
* [x] Add `Default.sublime-commands` file.
* [x] Don't use `sys.executable`.
* [x] Add `.no-sublime-package` file.
* [x] Refactor `BaseWindowCommand` class.

[getting started]: https://donkirkby.github.io/live-py-plugin/starting_sublime_text

### DIFF
--- a/channel.json
+++ b/channel.json
@@ -4,6 +4,7 @@
 		"./repository.json",
 		"https://bitbucket.org/jjones028/p4sublime/raw/tip/packages.json",
 		"https://bitbucket.org/klorenz/sublime_packages/raw/tip/packages.json",
+		"https://donkirkby.github.io/live-py-plugin/sublime-package/package.json",
 		"https://eoa-cdn.s3.amazonaws.com/extensions/packages.json",
 		"https://packagecontrol.io/packages_2.json",
 		"https://packages.monokai.pro/packages.json",


donkirkby:
Reopening, now that I have added a LICENSE file to the package zip file.

FichteFoll:
I suggest using the `edit_settings` command over two `open_file` commands in your menu file. It is only available in ST3, however, but you are already using a ViewEventListener, which isn't available to ST2. You should also consider adding these command entries to a `.sublime-commands` file for easy keyboard access.

`sys.executable` will *not* yield a path to an exectuable Python interpreter within ST's environment.

Because you are executing a file you bundle in your package with an external interpreter (Python), you must specify a `.no-sublime-package` file to tell PC to unarchive your package. Otherwise it will not be extracted into actual files on the user's disk.

Also note that your `BaseWindowCommand` class *will* be detected by ST's plugin loading algorithm and added as a callable command despite not doing anything. I suggest defining that class as a mixin and using multiple inheritance for the subclasses instead (alternatively define the utility methods there as functions receiving a window object).

A command named `StartCommand` will be registered as `start` in the global command namespace. Please prepend a namespace for generic names like this for your plugin to prevent conflicts.

donkirkby:
Thanks for all the helpful advice, I think I have taken care of all your suggestions.

I have a couple of follow-up questions, though, if you have time to give me some more advice.

1. Where is the `edit_settings` command documented? I found a [backport] of it that gave me enough information to use it, but is it actually documented somewhere?
2. Is it possible to get the configuration of the currently selected build system? Since that's how the user will likely be running their Python script, I'd like to modify that build script to run my live coding command. If there's a helper that builds the command line from the selected build system, that would be even easier for me. For now, I'm just using "python" as the default interpreter, and the user can override that in the settings file.

Thanks again.

[backport]: https://gist.github.com/math2001/6cd5cbb9d2741654c2e994d33c395729

FichteFoll:
> Where is the edit_settings command documented? 

Documentation on commands is sparse at best at the moment. However, since this command is not built-in, you can view its implementation in `Default/settings.py`. Most of the build-in commands can be found in https://github.com/SublimeText/PackageDev/blob/master/plugins_/command_completions/builtin_commands_meta_data.yaml and may or may not have a comment regarding their functionality. Some are also documented in more detailed in the [unofficial docs](https://sublime-text-unofficial-documentation.readthedocs.io/en/latest/reference/commands.html), but that hasn't been updated in a while and is also severely incomplete.

FichteFoll:
> Is it possible to get the configuration of the currently selected build system? Since that's how the user will likely be running their Python script, I'd like to modify that build script to run my live coding command. 

You just provide a build system with your package.
In the usual workflow the user then selects your build system using ctrl+shift+b, after which ctrl+b will continue to use that build system until it is no longer applicable (e.g. when the current syntax isn't Python).

FichteFoll:
Also, I almost forgot, but you'll need to push a new tag with your latest changes to make PC pick them up.

donkirkby:
Thanks again, @FichteFoll, I've pushed a new tag.

The unofficial docs link was helpful. I've seen `Default/settings.py` mentioned a few times, but I can't find it. Where is that?

As for the build system, I assume that many Python developers will have already set up a build system to run the Python script that they're editing. I'd like my plugin to read that build system to see where the Python interpreter is and any extra command-line arguments that they're sending to the script. I don't think I need to include my own build system with my plug in.

FichteFoll:
> Thanks again, @FichteFoll, I've pushed a new tag.

Actually, I totally forgot you were using a custom repository.json when I wrote this, so disregard the tag thing.

> I've seen Default/settings.py mentioned a few times, but I can't find it. Where is that?

It's in the shipped `Default` package. Shipped packages are located in ST's installation folder and you can view these files with the "View Package File" command.

> As for the build system, I assume that many Python developers will have already set up a build system to run the Python script that they're editing.

That is a reasonable assumption. However, I'm not aware of a way to retrieve the currently selected build system. You can compile a list of available build systems by doing whatever ST does manually, but you wouldn't know which one has been selected by the user.

PR #7752, opened 2019-11-10
Add Semantic Higlighter

kapitanluffy:
Plugin for semantic highlighting to variables

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -990,6 +990,18 @@
 			]
 		},
 		{
+			"name": "Semantic Highlighter",
+			"details": "https://github.com/kapitanluffy/sublime-semantic-highlighter",
+			"donate": "https://www.patreon.com/kapitanluffy",
+			"labels": ["text navigation"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Semantic UI",
 			"details": "https://github.com/idleberg/Semantic-UI-Sublime-Text",
 			"labels": ["snippets", "html"],


FichteFoll:
Why do you only define key bindings for Windows?

Setting change event handlers are registered for the entire settings objected and keyed by a string id for later removal, not for some setting name you provide.

As such, you must also unregister setting change handlers in a `plugin_unloaded` function as they will otherwise persist even when the package is unloaded.

What is the point of [this context handler](https://github.com/kapitanluffy/sublime-semantic-highlighter/blob/6566ec20839bbebb4a7cbfd774b643be9f3f4831/src/listeners/view_event.py#L11)?

FichteFoll:
@kapitanluffy ping

kapitanluffy:
> Setting change event handlers are registered for the entire settings objected and keyed by a string id for later removal, not for some setting name you provide.

Hey sorry I got swamped with work. Can you elaborate on this one?

FichteFoll:
It's like adding a key to a dictionary of callbacks that you can remove later by specifying the same key. Adding the same callback twice with different keys will result in the callback being called twice for every change.

kapitanluffy:
Hi @FichteFoll I updated the package. Can we reopen this PR?

FichteFoll:
> Setting change event handlers are registered for the entire settings objected and keyed by a string id for later removal, not for some setting name you provide.

This is still handled incorrectly, although it doesn't break with how it is currently. I suggest to just read the corresponding settings from the settings object directly when you want to use them instead of using a settings listener.

> What is the point of [this context handler](https://github.com/kapitanluffy/sublime-semantic-highlighter/blob/6566ec20839bbebb4a7cbfd774b643be9f3f4831/src/listeners/view_event.py#L11)?

kapitanluffy:
> > Setting change event handlers are registered for the entire settings objected and keyed by a string id for later removal, not for some setting name you provide.
> 
> This is still handled incorrectly, although it doesn't break with how it is currently. I suggest to just read the corresponding settings from the settings object directly when you want to use them instead of using a settings listener.
> 
> > What is the point of [this context handler](https://github.com/kapitanluffy/sublime-semantic-highlighter/blob/6566ec20839bbebb4a7cbfd774b643be9f3f4831/src/listeners/view_event.py#L11)?

I rewrote the whole thing. The context handler is for the keybindings. Let me know if there's a better way of doing that :)

kapitanluffy:
Tried reading the settings directly instead of registering a callback but I realized that it would read the settings file with every cursor move (which I think is very inefficient). 

I think I understood what you are saying now regarding the listeners though.

FichteFoll:
> Also, for the settings listener issue, do you mean that on `add_on_change` it fires the callback everytime _any_ setting is changed on that Setting object?

Yes.



> I rewrote the whole thing. The context handler is for the keybindings. Let me know if there's a better way of doing that :)

The [current implementation](https://github.com/kapitanluffy/sublime-semantic-highlighter/blob/7a19cd995f9b8e7569b92e281afca8ebc6d06e3c/src/listeners/view_event.py#L12) is still the same and a no-op. It will always match when the corresponing "key" is provided, in which case you might as well omit it entirely.

kapitanluffy:
> > Also, for the settings listener issue, do you mean that on `add_on_change` it fires the callback everytime _any_ setting is changed on that Setting object?
> 
> Yes.
> 
> > I rewrote the whole thing. The context handler is for the keybindings. Let me know if there's a better way of doing that :)
> 
> The [current implementation](https://github.com/kapitanluffy/sublime-semantic-highlighter/blob/7a19cd995f9b8e7569b92e281afca8ebc6d06e3c/src/listeners/view_event.py#L12) is still the same and a no-op. It will always match when the corresponing "key" is provided, in which case you might as well omit it entirely.
> 
> > Tried reading the settings directly instead of registering a callback but I realized that it would read the settings file with every cursor move (which I think is very inefficient).
> 
> Hm, that may indeed be a valid reason to have a cache in form of a native Python dict. You could time/benchmark it to be sure, but I understand the reasoning.

I removed the keybinding contexts for now. I fully understand what contexts are for now after reading about it on the docs. 

I also removed a mousemap file that I forgot included in the repo.

PR #7754, opened 2019-11-12
Add "Bibtex Bibliography Sort" package

wolfposd:
Plugin to sort bibliography entries in Bibtex.bib files by author, title, ref-type, ref-name


FichteFoll:
Please resolve the conflict. I can't from Github's web resolver because it doesn't offer me to do so.

PR #7757, opened 2019-11-15
Add JaiTools

RobinWragg:

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -82,7 +82,7 @@
 			"details": "https://github.com/RobinWragg/JaiTools",
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": ">=3092",
 					"tags": true
 				}
 			]
--- a/repository/j.json
+++ b/repository/j.json
@@ -68,17 +68,8 @@
 			]
 		},
 		{
-			"name": "Jai",
-			"details": "https://github.com/GreenLightning/sublime-jai",
-			"releases": [
-				{
-					"sublime_text": "*",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "JaiTools",
+			"previous_names": ["Jai"],
 			"details": "https://github.com/RobinWragg/JaiTools",
 			"releases": [
 				{
--- a/repository/j.json
+++ b/repository/j.json
@@ -78,6 +78,16 @@
 			]
 		},
 		{
+			"name": "JaiTools",
+			"details": "https://github.com/RobinWragg/JaiTools",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Jam Language",
 			"details": "https://github.com/BenjaminSchaaf/jam-sublime-syntax",
 			"labels": ["jam", "language syntax", "build"],


RobinWragg:
Warnings from the bot will be rectified.

RobinWragg:
Two warnings have been rectified. The remaining warning, referring to .no-sublime-package, is necessary for the build system.

RobinWragg:
The build system was removed until it is mature, along with the .no-sublime-package file.

FichteFoll:
How does this package compare to the existing [Jai package](https://packagecontrol.io/packages/Jai)?

RobinWragg:
The existing Jai package only has limited syntax support (e.g. no function tags, as I recall). It has no other features. This new package has much more comprehensive syntax support, Goto Symbol that shows nested functions with indents, and project-wide completions for variables, functions, and parameters.

FichteFoll:
So you're saying it is an elaborate enhancement? @GreenLightning would you agree with replacing your package with this?

GreenLightning:
@RobinWragg Your syntax highlighting seems to be missing some features: here-strings, hexadecimal integer and floating point literals, enums.

Also, certain functions are not recognized (admittedly, it is very complicated, which is the main reason why I haven't done it).

JaiTools

image

Jai

image

Here is the code in text form:

```
NORMAL_STRING :: "Foobar";

// Here-string not highlighted.
HERE_STRING :: #string DONE
This is a very long string that
can span multiple lines and might
contain keywords and stuff like
struct, 12345, #ignored
DONE

Numbers1 :: 123456;
Numbers2 :: 0xf000000000000000; // Not highlighted.
FLOAT32_MIN :: 0h00800000; // Not highlighted.

ExampleStruct :: struct {
	Foo: u8;
}

// Enum not highlighted.
FormatIntBase :: enum {
	DECIMAL     : 0,
	HEXADECIMAL : 1,
	BINARY      : 2
}

GetStdHandle :: (handle: s32) -> s32 #foreign;
WriteFile    :: (handle: s32, buffer: *void, buffer_length: s32, written_result: *s32, overlapped: *void) -> s32 #foreign; // Not highlighted.
array_add    :: (array: *[..] $T, item: T) -> void // Not highlighted.
```

It would be nice if you could add these features. For here-strings see [here](https://github.com/GreenLightning/sublime-jai/blob/868f1b8f2afc8b3c0c002959a50d1ef2fcb0c68b/Jai.sublime-syntax#L53). I also have a large body of real Jai code, which was made available with some of the demos (`demo_renamers_and_static_if` and `demo_type_parameters`, but the latter one contains a bunch of older demos as well), but I don't know if it is still available online. If you don't already have it, I can send it to you via email.

RobinWragg:
Thanks @GreenLightning, that's very helpful. I will add the features you mentioned to my package. Currently, the regex for functions expects an opening brace, as I haven't found confirmation that function declarations can exist without their definitions like they do in C/C++. I will revisit its implementation. Of course, we will have to wait for an official Jai release to ensure the syntax definition is solid, but I am dedicated to making Sublime Text a great Jai editor.

RobinWragg:
Hi @FichteFoll, with @GreenLightning 's blessing I do consider this an elaborate enhancement and I will incorporate @GreenLightning 's suggestions into this package. Many thanks to both of you.

FichteFoll:
> with @GreenLightning 's blessing I do consider this an elaborate enhancement

Unfortunately, I haven't seen an explicit blessing of the sort that would encourage me to go ahead with replacing the current package. Alternatively, you two could cooperate on a single repo (likely @GreenLightning's) and we can just close this PR.

RobinWragg:
What do you think, @GreenLightning ? Would you like me to take responsibility for Sublime's Jai support by replacing the package (and perhaps merging parts of your syntax file), or would you prefer us to collaborate on your package? For what its worth, I plan to add additional features such as build-system integration for in-line errors and warnings.

GreenLightning:
@FichteFoll I think it would be appropriate to replace the package, so that @RobinWragg will be the principal maintainer. I have sent a pull request for here-strings and the other missing features have already been integrated, so everything is looking good.

RobinWragg:
Thanks to both of you for your work and advice, @GreenLightning 's PR has been merged. A new version tag of my package has been created and it is ready to take the place of @GreenLightning 's package. @FichteFoll , should I remove @GreenLightning 's package as part of this PR?

FichteFoll:
Yes, please do that. You can either take over the old name or add it to your new package's `previous_names` array so PC will know to rename existing installations.

RobinWragg:
@FichteFoll The old Jai package has now been removed as part of this PR, and `"Jai"` has been added to the `previous_names` array of JaiTools.

PR #7759, opened 2019-11-17
Added Crackshell package

codecat:
This is a package specifically suited for Crackshell's game engine. It includes multiple syntaxes, including an AngelScript syntax built from the ground up.

I also maintain the original [AngelScript syntax](https://github.com/wronex/sublime-angelscript) by wronex, which I decided not to completely rewrite for 2 reasons:

1. GitHub is using wronex's repository in Linguist, which [does not support `.sublime-syntax`](https://github.com/github/linguist/pull/3214), so I can't update it to this new syntax unfortunately.
2. I added a couple of engine-specific keyword highlights, which wouldn't really fit in a generic highlighter like wronex'.

Plus, this contains a couple other syntaxes too, which provide special highlighting of engine features.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3871,6 +3871,16 @@
 			]
 		},
 		{
+			"name": "Crackshell",
+			"details": "https://github.com/Crackshell/sublime-text",
+			"releases": [
+				{
+					"sublime_text": ">=3103",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Craft Development Snippets",
 			"details": "https://github.com/alecritson/Craft-Sublime-Snippets",
 			"releases": [


codecat:
> does not contain a top-level LICENSE file.

Huh.. [looks fine to me](https://github.com/Crackshell/sublime-text/blob/master/LICENSE)?

codecat:
Sidenote; when installing via "repository URL", this happens due to the name of the Github repository:

image

When someone installs it via Package Control's channel, will it also be that name, or will it be the name of the package itself? If it will be `sublime-text`, should I change anything in the channel definition in order to make that work better?

FichteFoll:
>When someone installs it via Package Control's channel, will it also be that name, or will it be the name of the package itself?

It will use the name you provided to the `name` key via this PR, which is also the same as what @packagecontrol-bot reports.

PR #7760, opened 2019-11-19
Pretty YAML files with node

georf:
The python dumper for yaml files can not modify some styles of yaml
files. The node module js-yaml is better to configure. This plugin is a
little bit slower than the original "Pretty YAML", but you have more
options for styling your code.

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -2202,6 +2202,16 @@
 			]
 		},
 		{
+			"name": "Pretty YAML Node",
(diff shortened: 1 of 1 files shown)


FichteFoll:
I'm not a huge fan of bundling entire `node_modules` wormholes in package distributions compared to having the user install them on their system in parallel, but it seems quite manageable here.

Why don't you use stdin and -out streams for communication with your node script and pass the configuration via command line? That would get rid of the temporary directory.

PR #7761, opened 2019-11-19
Add Tinkerwell plugin

mpociot:
The product can be found at https://tinkerwell.app

This is how the plugin looks like:

sublime

Thank you :)

### DIFF
--- a/channel.json
+++ b/channel.json
@@ -128,6 +128,7 @@
 		"https://raw.githubusercontent.com/zfkun/sublime-kissy-snippets/master/packages.json",
 		"https://s3.amazonaws.com/wallaby-downloads/sublime/packages.json",
 		"https://sequoiastudios.io/sublime_packages/packages.json",
-		"https://sublimerge.com/packages.json"
+		"https://sublimerge.com/packages.json",
+		"https://tinkerwell.app/plugins/sublime/packages.json"
 	]
 }


niladam:
For the ones that are looking to get this sorted faster, in Package Control you can add a repository.

Screenshot 2019-11-19 17 32 46

You are then able to install using your regular install procedure :)

niladam:
hey @mpociot maybe take a look at this ?:) - Manually installing this package doesn't work. And as @wbond said on Twitter, this could be related to the obfuscation of the code..

mpociot:
I updated the package, so it should work now

wbond:
Yesterday when replying on twitter, I tried installing the package. You should not need a .no-sublime-package file from what I could tell. That is only used when you have a shared library or executable you need unpacked. Any `.sublime-*` files work fine in a `.sublime-package` and so do `.py` files.

mpociot:
Yeah I also got rid of that file :)
Thank you!

niladam:
@mpociot Manually installing it using the procedure above, i'm getting this:

Screenshot 2019-11-20 22 13 14

So, this appears to still not be working ?:(

niladam:
@wbond, @mpociot  - this works now!:)

erikverbeek:
Any update if/when we can get this added to Package Control?

FichteFoll:
You should indeed specify a `readme` key for your package with a url to a raw markdown (or other markup? not sure which we support) file that will be rendered on the package's packagecontrol.io entry.

Other than this, I don't see a reason for the package not to work when installed in its current state.

PR #7763, opened 2019-11-19
Add Mql5 Language Pack

tombout:
The Mql5 Language Pack includes auto completion, syntax highlighting and a command to compile Mql5 code. The plugin is similar to the MQL4 Language Pack but for MQL5 which uses other commands and executables to run on.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -2408,6 +2408,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


tombout:
I will do a new pull request due to the warning about missing tmLanguage file.

PR #7764, opened 2019-11-19
Mql5 Language Pack

tombout:
The Mql5 Language Pack includes auto completion, syntax highlighting and a command to compile Mql5 code. The plugin is similar to the MQL4 Language Pack but for MQL5 which uses other commands and executables to run on.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -2408,6 +2408,17 @@
 			]
 		},
 		{
+			"name": "Mql5LanguagePack",
+			"details": "https://github.com/tombout/SublimeMql5LanguagePack",
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
Is there any particular reason for using a plugin to compile the file over specifying a build system? If what you need to do doesn't work from within the standard `exec` command, you should instead consider using your custom command for the `target` key in a build system. See https://www.sublimetext.com/docs/3/build_systems.html for docs. (That way you can also get rid of the key binding.)

FichteFoll:
@tombout ping

tombout:
Sorry for my late response. I will take a look at the link. For now I will close the merge request.

PR #7774, opened 2019-12-02
Add MagentoWorkflow package

vovayatsyuk:
The module automates some stuff that every Magento2 developer should do to get maximum joy during development:

 - Cleans appropriate cache entries when source code is changed
 - Cleans appropriate generated code when needed
 - Sync file changes into docker when source code is hosted outside of the container

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -122,6 +122,17 @@
 			]
 		},
 		{
+			"name": "MagentoWorkflow",
+			"details": "https://github.com/vovayatsyuk/MagentoWorkflow",
+			"labels": ["magento", "magento automation", "docker magento", "magento cache clean"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "MagerunCommander",
 			"details": "https://github.com/nhduy1985/sublime_mageruncommander",
 			"author": "Dustin Tran",


vovayatsyuk:
@packagecontrol-bot

FichteFoll:
Because the command namespace is shared, please prefix your quit generic command names of `clear_cache` and `flush_cache` to prevent collisions.

Use `self.view.window()` to fetch the current window in a TextCommand instead of `sublime.active_window()`.

You might also be interested in using command input handlers instead of input panels for a more modern UX. See `Default/arithmetic.py` for an example.

vovayatsyuk:
Thanks! Done. InputHandler is a much nicer option!

The only thing I didn't fully understand is about `clear_cache and `flush_cache` commands. Can you clarify, please?

FichteFoll:
I meant to suggest renaming your commands to `magento_clear_cache` and similar because the command namespace is global and another package could define a `clear_cache` command as well, resulting in one of the commands being shadowed.

vovayatsyuk:
Don't know how the sublime plugins/python works :(. I thought all the files have their own namespace, that's why we must import them.

These methods are [inside App class](https://github.com/vovayatsyuk/MagentoWorkflow/blob/master/app/app.py#L33), would that help?

FichteFoll:
Command names are detemined from the class's name, e.g. [`FlushCacheCommand`](https://github.com/vovayatsyuk/MagentoWorkflow/blob/master/MagentoWorkflow.py#L62) is named `flush_cache` for command running purposes. Command names are global because they can be called from everywhere.

vovayatsyuk:
Thanks! I've added `MagentoWorkflow` prefix to all commands.

PR #7775, opened 2019-12-06
Added HighlightNumber plugin. It uses the popup feature of ST3…

kalund:


kalund:
Added a version tag with 1.0.0.

FichteFoll:
Your code paths are pretty obscure, but I'll just take your word for that it does what it says.

PR #7777, opened 2019-12-07
Added ProfileManager plugin

adrian-tut:
The Profile Manager plugin allows users to store all configuration files on their google drive account.

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -2104,6 +2104,18 @@
 			]
 		},
 		{
+			"name": "ProfileManager",
+			"details": "https://github.com/adrian-tut/SublimeProfiles",
+			"author": "Adrian Tut",
+			"labels": ["package", "cloud", "sync", "utilities", "synchronization", "profiles", "profile", "manager"],
+			"releases": [
+				{
+					"sublime_text": ">=3092",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Because you vendor a lot of dependencies and modify `sys.path`, you're likely to cause problems with other plugins that use those but either vendor them selves as well or use them from PC dependencies. Ideally you'd use relative imports, but this doesn't seem to be an option here.

I advise you to at least use the dependencies available from PC to reduce the chance of conflicts (since every plugin is limited to the same and predictable version of it). See packagecontrol.github.io/ for a list of the currently available ones. Specifically, I wonder why you need `easy_install` and `pkg_resources`, which are usually part of the installation/distribution toolchain like setuptools and rarely if ever see usage from within vendored packages.

Furthermore, beause PC packages are by default distributed in archived `.sublime-package` files, you should definitely test whether your method of appending to `sys.path` works in that state. If it does not, you need to provide a `.no-sublime-package` file that tells PC to unextract it.
Either way, you definitely **should not** add your own package's path to `sys.path`.

FichteFoll:
@adrian-tut ping

PR #7779, opened 2019-12-08
A proof-validity checker that functions as an IDE for .ax files with Sublime Text.

legalian:
A proof-validity checker that functions as an IDE for .ax files with Sublime Text.
Xaxiom is a new language and currently no addons exist for editing it.

### DIFF
--- a/repository/x.json
+++ b/repository/x.json
@@ -45,7 +45,7 @@
 			]
 		},
 		{
-			"name": "Xaxiom",
+			"name": "Xaxiom", 
 			"details": "https://github.com/legalian/xaxiom",
 			"releases": [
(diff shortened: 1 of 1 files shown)


FichteFoll:
Please add a minimum ST version requirement of 3092.

Scope names should be suffixed by your syntax's base scope name to uniquely identify them (i.e. `keyword.control.ax` instead of `keyword.control`).

Using look-behinds is discouraged because ST's fast sregex engine doesn't implement those. Push into a new context instead where possible (technically always possible).

Your `comments` context uses the `example-c` suffix.

FichteFoll:
@legalian ping

PR #7780, opened 2019-12-09
Add package: B Syntax Highlighting

baleyko:
This package provide basic support of the B programming language syntax (https://en.wikipedia.org/wiki/B_(programming_language))
It contains configuration to highlight the 'b' files, also it contains some snippets.

There is no any packages for support this language in Package Control public repositories.

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -2,6 +2,17 @@
 	"schema_version": "3.0.0",
 	"packages": [
 		{
+			"name": "B Syntax Highlighting",
+			"details": "https://github.com/baleyko/sublime-b-syntax-highlighting",
+			"labels": ["language syntax", "snippets"],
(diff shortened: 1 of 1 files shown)


baleyko:
@wbond please clarify the state of my pr. thank you.

FichteFoll:
The state of your PR was "waiting for review".

I recommend not using the `name` field for syntax definitions and instead writing out the name in the file name because that will be less confusing in the long run. No idea why you specifically lower-cased it.

Your syntax definition uses a lot of look-behind patterns, which are slow to execute. Try pushing into contexts instead or structure it in a way that ending tokens are always matched or don't matter.

baleyko:
Clear snd Thanks for suggestions. I will make some fixes a little later.

PR #7783, opened 2019-12-16
Add package: Syntax Highlighting Scopes Showroom

baleyko:
Syntax Highlighting Scopes Showroom - it's a package those could highlight syntax highlighting scopes, it helps build interop(Sublime, VSCode, Atom) syntax highlighting package

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -5722,6 +5722,16 @@
 			]
 		},
 		{
+			"name": "Syntax Highlighting Scopes Showroom",
+			"details": "https://github.com/baleyko/syntax-highlighting-scopes-showroom",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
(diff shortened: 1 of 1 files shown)


baleyko:
@wbond hey, could you clarify the state of my pr? thanks.

FichteFoll:
In your syntax patterns you use a literal `.` when you need to escape it because you're inside of a regular expression.

You might also want to add a command to the command palette to open the bundled example file, because I doubt there is any other file you would reasonably want to use this syntax for aside from a stripped down version of that.
Use something akin to `{ "command": "open_file", "args": {"file": "${packages}/Syntax Highlighting Scopes Showroom/example.shss"} }`.

baleyko:
@FabioFuschi-SNK good idea, I will add something like that a little later

PR #7784, opened 2019-12-24
adding Microsoft Power Query syntax from https://powerqueryforaccountants.com

dannysummerlin:
Microsoft Power Query syntax from https://powerqueryforaccountants.com

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -1478,11 +1478,11 @@
 		{
 			"name": "Microsoft Power Query Syntax",
 			"details": "https://github.com/dannysummerlin/PowerQuerySublimeSyntax",
-			"labels": ["language syntax"],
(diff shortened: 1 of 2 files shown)


dannysummerlin:
Not sure about that last warning, I added a LICENSE file, so I don't know why it isn't seeing it.

FichteFoll:
> Not sure about that last warning, I added a LICENSE file, so I don't know why it isn't seeing it.

Because you haven't pushed a new tag with the change. Checks are always run against the latest release, not your master branch. However, for the LICENSE it's not too important to be included in the distrubuted archive. (Techically it is for the user, but that's irrelevant.)

If you only have a single capture for the 0 group, `scope` is equivalent.

Try not to use look-behinds because they are slower in ST.

PR #7788, opened 2019-12-26
Add the new Crates package

Kerollmops:
I ran the _ChannelRepositoryTools: Test Default Channel_ test command and the 9152 tests passed.

This package helps Rustaceans retrieve the Rust crates versions. The user can either use the command palette to select a crate version to use in a the versions list, or select the names of multiple crates that will be appended with the last crate version available.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3912,6 +3912,17 @@
 			]
 		},
 		{
+			"name": "Crates",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Good example for using input handlers.

PR #7794, opened 2019-12-29
New package to display numbers in pop-up

nia40m:
My plugin convert and displays a highlighted number in different numeral systems in a pop-up window, what can be helpful to hardware programmers. I didn't found any similar package with a similar functionality.
Plugin can also convert it with a hot-keys, and there is a similar plugin in package list (https://packagecontrol.io/packages/C%20Base%20Converter), but in my plugin is not a main feature, just a bonus. And my plugin can work (display and convert) with only one highlighted number at once.

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -905,6 +905,16 @@
 			]
 		},
 		{
+			"name": "Display numbers",
+			"details": "https://github.com/nia40m/sublime-display-nums",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "DistractionFreeWindow",
 			"details": "https://github.com/aziz/DistractionFreeWindow",
 			"labels": ["fullscreen", "distraction-free"],


FichteFoll:
Only two days ago, I merged a pretty similar package to this: https://packagecontrol.io/packages/Highlight%20Number

However, I believe this to be much superior, because it has better code quality, is more explicit about what it does, allows for more number formats and has a prettier popup (and the name is more decriptive).

@kalund would you be okay if we removed your package in favor of this one?

---

Regarding the package itself, settings change hooks don't operate on setting names. Your handler will be called whenever any setting in the wrapped object changes and the specified `id` is used to unregister the handlers. You also don't unregister the handlers in `plugin_unloaded`. The better approach is to just read the settings as you need them, which also gives more control over which settings object you're looking at exactly (instead of `sublime.active_window().active_view().settings()`).

Many languages support `_` spacing within number literals. You should probably add support for those and strip them before parsing.

You should provide a `.sublime-commands` file with an entry for your command.

Why are the binary numbers underlined?

kalund:
@FichteFoll I agree that this one is similar, and superior in functionality. But my plugin still has some advantages.
1. It shows the highlighted number converted from different bases simultaneously, for example 10 will be shown as converted from hex 10, dec 10 and bin 10.
2. It does not care about the prefix of the highlighted number for conversion. As I often only need to convert a slice of a number this is really helpfull.
3. My popup stays open even if the mouse is moved outside the popup,

I'm sure all 3 points could be incorporated into this plugin, but untill then I would like to keep mine.

nia40m:
@FichteFoll thanks for your help improving my plugin.

I changed all the work with settings. Now I get them right when I need without any hooks and global variables.

I added a support for `_` in numbers.

Also added a `.sublime-commands` file with basic commands.

> Why are the binary numbers underlined?

To display one of the features: _"Pressing any digit of binary number makes it opposite of it's current value"_.

I have a question about folder name: when someone downloads any plugin with **Package Control** what folder name will the plugin have in the folder `Packages`? I have in my `Main.sublime-menu` file hard-coded a folder name `sublime-display-nums`. Should I replace it and to what?

---

@kalund about your tasks:

1. As you could see, my plugin have two modes (`basic` and `extended`). I can add a third one that will look and work like your plugin. User will be able to select which one is to use.

1. I thought that it's better to hide popup window when the mouse is outside of it. I see three possible solutions for that:

    - make it visible when mouse is moved outside of the popup (like you did);
    - make it editable in the plugin settings file (hide or not);
    - leave as it is for my two modes and make it for your third mode as in your plugin.

**What do you both think about that?**

FichteFoll:
>when someone downloads any plugin with Package Control what folder name will the plugin have in the folder Packages?

Good of you to mention that since I forgot to check. Your package is installed as "Display numbers", via the `name` key in this PR.

kalund:
@nia40m It sounds right, I would just make it all configurable from the settings. Does this also include converting from eg. hex without a prefix?
Anyway, when you have a new version ready, let me know and I'll gladly test it.

nia40m:
@kalund yes, it will convert even without prefix.

You can now download the latest version from master branch (https://github.com/nia40m/sublime-display-nums).
You should set a setting `plugin_mode` to **"tabled"** and `hide_on_mouse_move_away` to **false**.

I did it almost like your plugin was.

*And note that the plugin folder name should now be "Display numbers"*.

If you will find bugs or have any suggestions, please let me know.

kalund:
@nia40m Thanks, it looks and works as mine did, I'll use yours from now on. Thanks for the effort.

@FichteFoll You can delete my plugin when this one is merged.

nia40m:
@kalund I'm glad to hear that!

Then I'll tag this version and reopen pull request to launch package control bot to test it.

FichteFoll:
84d626a6d8f690a3b4f333ac69a5586b28925b2e

nia40m:
@FichteFoll Thank you!

PR #7796, opened 2019-12-31
Adding dotsyntax.

seanmorris:
This package allows you to specify syntax highlighting for **custom** dotfiles by creating a `.syntax` file in the project directory. I've seen packages that highlight existing dotfile standards, but not ones that allow configuration for custom dotfiles on a per-project, interpreter-agnostic, pattern-based fashion.

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -1371,7 +1371,6 @@
 			"labels": ["syntax", "hightlight", "config", "per-project", "dotfile"],
 			"releases": [
 				{
-					"branch": "master",
 					"sublime_text": ">=3000",
 					"tags": true
 				}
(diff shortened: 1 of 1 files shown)


FichteFoll:
This falls under a similar problem domain as [ApplySyntax](https://packagecontrol.io/packages/ApplySyntax), but uses a different method to solve it, notably providing folder/project-specific rules on the file system that is more general because it maps extensions.

Your [exit condition in `find_dotsyntax_files`](https://github.com/seanmorris/dotsyntax-sublime/blob/0c2863f1757eb7fc9a3d74ae6513af1a25e34f65/dotsyntax.py#L97) doesn't look like it could handle Windows paths, however.

Using `View.assign_syntax` is preferred over setting the setting directly.

seanmorris:
@FichteFoll I'll look into all that and get back to you!

FichteFoll:
@seanmorris ping

PR #7797, opened 2020-01-01
Add Yosys package

whitequark:
This package provides syntax highlighting for Yosys IR (RTLIL) and scripts.

### DIFF
--- a/repository/y.json
+++ b/repository/y.json
@@ -230,6 +230,16 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
You should include `\n` in your comment pop pattern match to prevent auto completion from showing up while typing at the end of the line.

whitequark:
@FichteFoll Done, thanks for the tip!

PR #7798, opened 2020-01-02
Add Devastate Mode-Z Theme/Colour scheme

CocaineJohnsson:
This package contains a Colour Scheme + minimalist theme combo designed to minimize distractions and improve code-readability at a glance by using carefully selected colours that provide distinctive contrast but that are also easy on the eyes for extended viewing. It's also reasonably colourblind friendly, thought that's more of a happy accident than some grand design.

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -575,9 +575,11 @@
 		{
 			"name": "Devastate Mode-Z",
 			"details": "https://github.com/CocaineJohnsson/sublime-devastate-mode-z",
+			"labels": ["theme", "color scheme"],
+			"author": "Cocaine Johnsson",
 			"releases": [
 				{
-						"sublime_text": "*",
+						"sublime_text": ">=3149",
 						"tags": true
 				}
 			]
--- a/repository/d.json
+++ b/repository/d.json
@@ -573,6 +573,16 @@
 			]
 		},
 		{
+			"name": "Devastate Mode-Z",
+			"details": "https://github.com/CocaineJohnsson/sublime-devastate-mode-z",
+			"releases": [
+				{
+						"sublime_text": "*",
+						"tags": true
+				}
+			]
+		},
+		{
 			"name": "DevastateMini",
 			"details": "https://github.com/shagabutdinov/sublime-devastate-mini",
 			"donate": "https://github.com/shagabutdinov/sublime-enhanced/blob/master/readme-donations.md",


FichteFoll:
Your theme references resource files with the package name "DevastateMode-Z" but via your PR it will be installed as "Devastate Mode-Z" and not match. Actually, the entire prefix you use is `DevastateMode-Z/DevastateMode-Z/` while it should be  `Devastate Mode-Z/Devastate`.

Since this is a fork, I have to wonder if you are complying to the fork origin's license terms.

CocaineJohnsson:
> Since this is a fork, I have to wonder if you are complying to the fork origin's license terms.

It's a fork of[ DevastateMini ](https://packagecontrol.io/packages/DevastateMini)(already in repos) which is MIT licensed, to the best of my knowledge it should be compliant with the license terms, though I'm not a lawyer so maybe not?

The only thing I can think of is that I might want to add a thirdparty folder listing licenses of things I've used, erring on the side of caution I opted to do so. Relevant license files can now be found in thirdparty-license/DevastateMini-license.md 

> Your theme references resource files with the package name "DevastateMode-Z" but via your PR it will be installed as "Devastate Mode-Z" and not match. 

My bad, I don't fully understand the package control system yet. 

Looks like an easy fix on my end though -- should be fixed, I'm not sure if I'm supposed to escape spaces (e.g "Devastate\ Mode-Z" or not, I have erred on the side of not doing so.

> while it should be Devastate Mode-Z/Devastate.

Another easy fix, I didn't correctly merge all changes from testing to github, that's my bad. Sorry.


Sorry for wasting your time, but I believe the issues have been fixed now. Is there anything else I need to do, or do I just await re-review (since this hasn't been closed I'm assuming I should assume it awaits fixes?). A new release (1.1) has been pushed via github containing the fixes to the issues raised, if anything else is needed on my end I will do my best to comply.

FichteFoll:
>It's a fork of DevastateMini (already in repos) which is MIT licensed, to the best of my knowledge it should be compliant with the license terms, though I'm not a lawyer so maybe not?

I didn't check how much the two differ, but your repo did not contain the original copyright notice inside `license.md`, which, according to the terms, needs to be preserved for derivatives. You may add yourself to the note, of course, but the original can only be removed if nothing is left in the repo from the copyright holder to be removed.
I believe adding the original license as you did in a subfolder *could* work, but the method I proposed is more clear.

FichteFoll:
> I'm not sure if I'm supposed to escape spaces

You don't.

> Is there anything else I need to do, or do I just await re-review 

Correct. You also pushed a new tag already, so I can merge this now.

PR #7799, opened 2020-01-02
Rename Hover Image Preview to Image Preview

alvesjtiago:
This PR renames the Hover Image Preview package to Image Preview in order to make it easier to discover.

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -955,16 +955,6 @@
 			]
 		},
 		{
-			"name": "Hover Image Preview",
-			"details": "https://github.com/alvesjtiago/hover-preview",
-			"releases": [
-				{
-					"sublime_text": ">=3000",
-					"tags": true
-				}
-			]
-		},
-		{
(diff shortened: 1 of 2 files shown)


FichteFoll:
I personally find the old name much more descriptive of what it does and having more search names will hardly make it less discoverable when there isn't any competition. Imagine searching for "image hover", which [currently works](https://packagecontrol.io/search/image%20hover).

Either way, tests are failing.

alvesjtiago:
@FichteFoll the reasoning behind it is that the plugin now works without necessarily hovering.  I've just fixed the ordering issue and added a label called hover. Will this provide the same output as https://packagecontrol.io/search/image%20hover? Thank you

FichteFoll:
No, labels are not covered in the search terms, but the description is, so you might want to add some terms there. If you didn't know, the description is sourced from your repo description (but you can also supply it directly here).

PR #7802, opened 2020-01-12
add package: StickWithMarkdownSnippets

UniFreak:

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -3801,6 +3801,16 @@
 			]
 		},
 		{
+			"name": "StickWithMarkdownSnippets",
+			"details": "https://github.com/UniFreak/SublimeMdSnippets",
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
You don't need `.no-sublime-package`. Please remove it.

You can't access `sublime`'s API at module load time ([with exceptions](https://www.sublimetext.com/docs/3/api_reference.html#plugin_lifecycle))

Instead of `sublime.load_settings('Preferences.sublime-settings')`, use `view.settings()`, which ST will resolve automatically and accounts for view-specific settings.

UniFreak:
@FichteFoll Thanks for the tip! I have changed my plugin according to your suggestions, and made the following edits:
- removed `.no-sublime-package` file
- load settings in `plugin_loaded()` function
- replace `sublime.loading_settings()` call with `view.settings()`

I have made a new commit, please review it :)

PR #7805, opened 2020-01-15
add cidr_convert plugin

shenanigans123:
Added a plugin:

Converts IPv4 addresses to/from cidr notation. Detects and converts addresses between the formats:
    192.168.1.0 255.255.255.0
    192.168.1.0/24

Acts on entire document if no text is selected, otherwise only acts within selected area.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1203,6 +1203,16 @@
 			]
 		},
 		{
+			"name": "CIDR Convert",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Your code is pretty inefficient because you do a search for each possible mask instead of finding all candidates and then seeing if you have a mask that you can replace.

shenanigans123:
Thanks i'll look into re-working it.

PR #7812, opened 2020-01-29
Add Markdown Images

xsleonard:
This package renders images inline in markdown files. These images are not popups and do not require mouse interaction to display. It can display the images by default when opening, and refresh them upon saving, or the user can choose to show or hide them from the command palette.  It works for both local and remote files of the PNG, JPG and GIF types.

The intent is to enable automatic image rendering in notes files, similar to other note-taking application that allow image embeds.

The closest plugin I could find is https://packagecontrol.io/packages/ImagePreview.
ImagePreview uses a popup hover to display the image, rather than inserting it into the document's view, and also requires mouse interaction. Since it uses a popup, ImagePreview makes sense to use in any type of file, and has a different usage intent than Markdown Images.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -468,6 +468,16 @@
 			]
 		},
 		{
+			"name": "Markdown Images",
+			"details": "https://github.com/xsleonard/sublime-MarkdownImages",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Markdown Numbered Headers",
 			"details": "https://github.com/weituotian/md_numbered_headers",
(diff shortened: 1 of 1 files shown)


FichteFoll:
I initially meant to suggest adding this to ImagePreview, but I believe the solutions to be distinct enough that one would be likely to want either but not both, so I think it's fine to merge as is.

Pinging @alvesjtiago regardless, in case he would be interested to add this as an optional feature.

alvesjtiago:
This is very interesting, thanks for the ping @FichteFoll. I agree that the feature set is different and it makes sense as a separate plugin but if you'd ever like to discuss a possible integration @xsleonard it would be great. Thank you both.

PR #7813, opened 2020-01-29
Add Fold ABL

RudnikovPaul:

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -1402,6 +1402,17 @@
 			]
 		},
 		{
+			"name": "Fold ABL",
+			"details": "https://github.com/RudnikovPaul/FoldABL",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": ["windows"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Fold Comments",
 			"details": "https://github.com/oskarols/foldcomments",
 			"labels": ["formatting"],


RudnikovPaul:
added version tag

FichteFoll:
You only have key bindings for Windows, but I don't see a particular reason for that.

You might also want to restrict your key bindings to only be active for ABL files using a context specifier.

RudnikovPaul:
> You only have key bindings for Windows, but I don't see a particular reason for that.

Yes, the key bindings for shure don't have to be restricted only for Windows. Fixed that.

> You might also want to restrict your key bindings to only be active for ABL files using a context specifier.

Restricting key bindigs according to file extentions is a bad idea with abl, because, on practice, abl code can have any extension type, it doesn't matter. It would be better to restrict them to current syntax highlighting, but I've not figured out how to do that.

FichteFoll:
> It would be better to restrict them to current syntax highlighting, but I've not figured out how to do that.

That's what I meant. Doing it based on the extension would be much more work in the first place.
[Here's an example for syntax-specific bindings in the JavaScript package.](https://github.com/sublimehq/Packages/blob/759d6eed9b4beed87e602a23303a121c3a6c2fb3/JavaScript/Default.sublime-keymap#L5)

FichteFoll:
@RudnikovPaul ping

RudnikovPaul:
The key bindings are now restricted to abl syntax.

FichteFoll:
Please push a new tag with the change so I can merge.

RudnikovPaul:
Pushed 1.0.1

PR #7815, opened 2020-02-01
Done, a simple todo list package

snehlsen:
Done - a simple Todo list.
Works on all *.todo files. 

Supported Commands:
- Done: Add new todo
- Done: Mark as done
- Done: Add new tag
- Done: Add existing tag
- Done: Set due date
- Done: Show due today

You need a **separator** between done and open tasks, default is **===**

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -1310,6 +1310,16 @@
 			]
 		},
 		{
+			"name": "Done",
+			"details": "https://github.com/snehlsen/sublime-done",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
There is a similarly named package "ToDone", but I suppose this is different enough.

Please add a namespace to your commands in the form `DoneTagCommand` or similar, because command names are global and shared among all plugins.

You may also be interested in using a multi-line string for your HTML template.

snehlsen:
I prefixed all commands with Done. 
Also using a multi-line string now 😉

FichteFoll:
Well, the convention would be different in that you'd TitleCase the command classes, resulting in snake_case command names, but this works.

PR #7817, opened 2020-02-05
add StataLinux

acarril:
Sublime Text 3 plugin that adds support for Stata 13-16 in Linux.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -3674,6 +3674,18 @@
 			]
 		},
 		{
+			"name": "StataLinux",
+			"details": "https://github.com/acarril/StataLinux",
+			"labels": ["stata"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"platforms": ["linux"],
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
As it is, you need to add a `.no-sublime-package` file because you are bundling and executing an executable in your package.
However, instead you could just perform the things your .sh file does inside the Python plugin runtime, where you also have access to `sublime.set_clipboard`. This will make the plugin more self-contained and not require this file, which limits user configuration.

acarril:
> As it is, you need to add a `.no-sublime-package` file because you are bundling and executing an executable in your package.
> However, instead you could just perform the things your .sh file does inside the Python plugin runtime, where you also have access to `sublime.set_clipboard`. This will make the plugin more self-contained and not require this file, which limits user configuration.

I see. I will look into implementing this purely in Python, which definitely would make it more elegant. In the meantime, I have added a `.no-sublime-package` file to publish it as it is.

FichteFoll:
You'll also need to push a new tag with that commit.

acarril:
OK, done.

PR #7818, opened 2020-02-06
Pretty Shell: more appropriate labels

aerobounce:
### Pretty Shell
Modified `labels` to more appropriate terms and order.


aerobounce:
Thanks!

PR #7820, opened 2020-02-09
Added Advanced PLSQL

TheEggi:
There exists a similar package, which is very inactive (last commit in 2016). The main reason why this package was now created is that there exists quite a few ideas (better build system, included SQL tools, better highlights, ...) on how to improve / get this package at least to the same level as the VS Code package. In order to achieve this the possibility of regular updates is mandatory and therefore a new package has been created.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -399,7 +399,7 @@
 		{
 			"name": "Advanced PLSQL",
 			"details": "https://github.com/TheEggi/Advanced-PLSQL",
-			"labels": ["snippets", "plsql", "oracle", "build"],
+			"labels": ["snippets", "sql", "plsql", "oracle", "build"],
 			"releases": [
 				{
 					"sublime_text": "*",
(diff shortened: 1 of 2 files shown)


FichteFoll:
I assume you're talking about https://packagecontrol.io/packages/Oracle%20PL%20SQL? That really only has a syntax definition and if you include that to at least the same feature level, since yours looks to be based on it, I would condone replacing it entirely to give users less choice and only present them the better version. What do you say?

TheEggi:
That sounds perfect to me. Should I just edit this PR in a way that it replaces the original plugin or how should we continue here?

Regarding the syntax: The syntax is based on the existing plugin with some improvements (and it still needs a few more minor fixes for it to be perfect).

FichteFoll:
Yes, just replace the package URL and optionally rename it (by supplying `previous_names`).

Per convention, we grant maintainers a 2 week period to object to their package being replaced. @mulander, if you would like to object or condone this change, please comment.

mulander:
No objections - go ahead. I stopped using Sublime after moving to OpenBSD (no builds for the platform :().

PR #7821, opened 2020-02-09
Add Tailwindcss 1.2.0 autocomplete

MiloudiMohamed:
TailwindCSS v1.2.0 Autocomplete plugin.

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -254,6 +254,18 @@
 			]
 		},
 		{
+			"name": "TailwindCSS Autocomplete",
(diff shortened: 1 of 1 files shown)


FichteFoll:
We can't have three packages whose names only differ in spaces. Honestly, https://packagecontrol.io/packages/Tailwind%20CSS%20Autocomplete already should not have been accepted as-is. Anyway, have you looked at that? Its last release was only months ago.

I could certainly see replacing the old unmaintained package with that package.

MiloudiMohamed:
I agree. though I didn't see this particular one, looks good so I'm closing this PR for now as it's irrelevant.

PR #7825, opened 2020-02-15
Add BazelSyntax

niosus:
This package provides a syntax for `BUILD` and `WORKSPACE` files for Bazel build system. To the degree of my knowledge, there is no package that does it so far. The initial version mostly follows `Python` syntax as skylark (language of Bazel) is a subset of Python, however, it adds an additional check for missing commas in lists and adds some initial handy snippets. More to follow in the next versions.

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -250,6 +250,25 @@
 			]
 		},
 		{
+			"name": "BazelSyntax",
+			"details": "https://github.com/niosus/BazelSyntax",
+			"readme": "https://raw.githubusercontent.com/niosus/BazelSyntax/master/README.md",
+			"donate": "https://www.paypal.com/cgi-bin/webscr?cmd=_s-xclick&hosted_button_id=2QLY7J4Q944HS",
+			"labels": [
+				"bazel",
+				"syntax",
+				"snippets",
+				"language syntax"
+			],
+			"releases": [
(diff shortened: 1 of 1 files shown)


niosus:
LICENSE file is added, but not in the release, hope that is ok. The other two warnings stem from the bindings that are copied over from the Python syntax, hope they are ok too.

FichteFoll:
I couldn't check which actual changes you made to the Bazel syntax definition, because this basically seems to be a copy of the Python definition with all occurances of `python` replaced with `bazel`, even in the URLS (which you might want to revert, as they now point to nothing). Did you change anything other than missing comma detection? Also, since that seems like a useful change to Python as well, do you mind upstreaming that change?

Everything else looks fine.

niosus:
Thanks for suggestions. Sure, I will update the links, I did that as a quick version, so some things might be off. Decided to start somewhere. 

At this point nothing apart from the coma is handled, but there are more things to come at a later point in time. Still mostly snippets I guess. Python syntax is a good enough approximation for a  start. The only reason to change the syntax really is that when using python-specific packages they begin to trigger for bazel files and that was annoying.

I don't think I can upstream the change, however, as in python implicit string concatenation is absolutely legal, while it is not legal for bazel.

PR #7826, opened 2020-02-16
Update Hybris Impex plugin

vladkanash:
Impex language support.
Existing plugin repository was removed a long time ago.
I have contacted the author of the plugin and he said he is no longer going to maintain it

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -1451,12 +1451,12 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
Normally I'd suggest adding a `previous_names` array to indicate a rename, but the old package hasn't been installable for 4 years and only achieved a few hundred installs anyway, so starting anew makes sense to me.

PR #7827, opened 2020-02-17
Add Perl IDE

BenBanerjeeRichards:
Package provides intelligent autocomplete, refactoring, find usages and goto declarations for Perl 5.

The plugin makes use of a separate perl analysis server which can be found at [https://github.com/BenBanerjeeRichards/perl-ide-server](https://github.com/BenBanerjeeRichards/perl-ide-server)

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -804,6 +804,18 @@
 			]
 		},
 		{
+			"name": "Perl IDE",
+			"details": "https://github.com/BenBanerjeeRichards/perl-ide",
+			"labels": ["perl", "completions", "auto-complete"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true,
+					"platforms": ["osx", "linux"]
+				}
+			]
+		},
+		{
 			"name": "perl-Test-Class",
 			"details": "https://github.com/jonasbn/SublimeText-Perl-Test-Class",
 			"labels": ["auto-complete", "formatting", "language syntax", "snippets"],


BenBanerjeeRichards:
I've added a license
.no-sublime-package needed as I have executables in lib/

FichteFoll:
Your plugin unconditionally modifies every view's `word_separators` setting. I cannot accept this. Either provide it as a syntax-specific settings file with the `Perl.sublime-settings` file name, or only set it for Perl files. Besides, you should use `view.match_selector(locations[0], "source.perl")`.

Your command names are quite generic, e.g. `rename_symbol`. Because command names from all plugins share the same namespace, please prefix them to prevent collisions. 

Furthermore, you can define `is_visible` methods on commands that are called when determining which menu items or command palette entries to show when they are requested instead of manually managing a `.sublime-menu` file.

You access ST's API on module load time through the `configure_settings` call in `main.py`, which may not be available on startup. See [the docs](https://www.sublimetext.com/docs/3/api_reference.html#plugin_lifecycle) for details.

FichteFoll:
@BenBanerjeeRichards ping

BenBanerjeeRichards:
Sorry for the delay, that feedback is very useful thank you. I'm putting this project on hold for now as I don't need it on package control for the time being (it was part of my Uni dissertation but due to UK strikes and then the virus it is no longer needed). I'll come back to it in couple months when I'm not so busy and bored in quarantine.

PR #7829, opened 2020-02-18
Renamed Package

daniele-niero:
I just renamed "Project Environment" to "ProjectEnvironment".
It wasn't clever to use spaces in the package's name.
I think remove the space it's better for anyone using the package

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -1032,17 +1032,6 @@
 			]
 		},
 		{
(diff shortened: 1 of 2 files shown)


FichteFoll:
You must reorder the package for its new name. See the travis results for where the package is expected to be.

FichteFoll:
@daniele-niero ping

PR #7830, opened 2020-02-18
Adding Colored Comments

TheSecEng:
Colored Comments allows users to define comments types and a color syntax for them to be highlighted as. Inspired by VS Codes Better Comments.

Intended Use:

image

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2817,6 +2817,16 @@
 			]
 		},
 		{
+			"name": "Colored Comments",
+			"details": "https://github.com/TheSecEng/ColoredComments",
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
You might want to link to a list of provided color-ish scopes. The only official reference I found for them, however, was [in the minihtml docs](https://www.sublimetext.com/docs/3/minihtml.html#predefined_variables).

FichteFoll:
You might want to link to a list of provided color-ish scopes. The only official reference I found for them, however, was [in the minihtml docs](https://www.sublimetext.com/docs/3/minihtml.html#predefined_variables).

PR #7832, opened 2020-02-19
Include plugin to open visual studio code from sublime

stoivo:

### DIFF
--- a/repository/v.json
+++ b/repository/v.json
@@ -559,6 +559,19 @@
 			]
 		},
 		{
+			"name": "Visual studio code",
+			"description": "Open visual studio code from sublime, light integration",
+			"details": "https://github.com/stoivo/sublime-open-in-vscode",
+			"issues": "https://github.com/stoivo/sublime-open-in-vscode/issues",
+			"labels": ["vscode", "visual studio code"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Visual Studio Dark",
 			"details": "https://github.com/nikeee/visual-studio-dark",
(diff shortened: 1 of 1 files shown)


FichteFoll:
The name "Visual studio code VScode" is kinda … redundant, wouldn't you agree?  "Open in VSCode" or "Open in Visual Studio Code" seem much more appropriate.

I'm not sure if providing multiple `--add` arguments in the same argument actually works, but maybe `os.system` can handle it. It is usually recommended to use `subprocess` instead.

stoivo:
I can agree to t the name. For the package I want it to show up either if you type vscode or visual studio code. I did "VScode" first but then a I typed "Visual ..." and added it too. I will update the name to "Open in Visual Studio Code"

Give  me a week and I can move it to subprocess. Thanks for the review

FichteFoll:
Thanks to fuzzy matching, `vscode` will also match `visual studio code`, so that should be covered. For the packagecontrol.io site, you should know that the description (sourced from your repository description) is searched as well, so you can use that for search term optimization.

stoivo:
Turn out os.system work better in this case.

FichteFoll:
I generally don't think so, especially considering the docs even encourage usage of `subprocess` instead of `os.system`, but I'm not gonna block on that.

You do, however, have to change the package name.

stoivo:
Thanks for taking the time to support the Sublime community.

I did rename the package but forgot to push the commit. If you say it's better to use `subprocess` it probably is, so I spent some time to figure it out. I'm not a python programmer so. Thanks again.

FichteFoll:
Good color scheme choice, by the way. :+1:

PR #7835, opened 2020-02-22
Adding aw-watcher-sublime

kostasdizas:
aw-watcher-sublime reports the files and projects you're working on to the open source time tracker [ActivityWatch](https://activitywatch.net)

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -236,6 +236,16 @@
 			]
 		},
 		{
+			"name": "ActivityWatch Watcher",
+			"details": "https://github.com/kostasdizas/aw-watcher-sublime",
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
The "sublime" part in "aw-watcher-sublime" is redundant for a package index for Sublime Text (and ST only). Please remove it.

kostasdizas:
Fair enough, updated.

FichteFoll:
You need to fix the package order now.

Like @packagecontrol-bot mentions, you have a `.no-sublime-package` file but I don't see any need for it as you don't bundle an executable or do other stuff that would require it. Please remove it.

kostasdizas:
I think I misunderstood the purpose of that file. Removed.

PR #7836, opened 2020-02-24
Adds the Better Settings dependency

KuttKatrea:
This module allows a plugin to easily implement OS, Host and
OS/Host-level settings aside the normal user-level settings.

This allows to keep the settings file in the User package, so they can
be synced between machines using different OSes or with specific
configuration for each host without interfering between them.

### DIFF
--- a/repository/dependencies.json
+++ b/repository/dependencies.json
@@ -46,6 +46,20 @@
 			]
 		},
 		{
+			"name": "better_settings",
+			"load_order": "01",
+			"description": "Easily implement OS, Host and OS/Host-level settings aside the normal user-level settings for any plugin.",
+			"author": "KuttKatrea",
+			"issues": "https://github.com/KuttKatrea/sublime-better-settings/issues",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"base": "https://github.com/KuttKatrea/sublime-better-settings",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "boto3",
 			"load_order": "02",
 			"description": "Python boto3 module for Amazon Web Services",


FichteFoll:
I would suggest using `collections.ChainMap` instead of manually chaining .get calls, but I'm not sure if that works for classes that don't implement all the Mapping methods. [`sublime_lib`](https://sublimetext.github.io/sublime_lib/) provides implementations for most, in case this is required.

I also find it weird to instantiate the settings object but then having to `load` them later anyway. Since `sublime.load_settings` also immediately returns a result, I would suggest loading the files in `__init__` and documenting that users shouldn't create instances at the module level during module load time and instead wait for `plugin_loaded`. You also only check your settings attributes for None in `clear_on_change` for some reason.

Your `get_settings_file_path` function ends up producing a path with two consecutive slashes for the User files.

What is motivation for the `basepackage` parameter, which you use `__package__` for in your readme? It's not documented, but its only use seems to be for `open_settings` and assumes for the settings file to be stored in the root. This is not a requirement, so you should instead allow the user to provide an actual path or use `sublime.find_resources` and just a file name.

KuttKatrea:
> I would suggest using `collections.ChainMap` instead of manually chaining .get calls, but I'm not sure if that works for classes that don't implement all the Mapping methods. [`sublime_lib`](https://sublimetext.github.io/sublime_lib/) provides implementations for most, in case this is required.

~~~
File "...\all\better_settings.py", line 58, in get
    return self.chained_settings.get(setting_name, default)
  File "./python3.3/collections/__init__.py", line 795, in get
  File "./python3.3/collections/__init__.py", line 804, in __contains__
  File "./python3.3/collections/__init__.py", line 804, in <genexpr>
TypeError: argument of type 'Settings' is not iterable
~~~

It doesn't seem to work, so I'll have to stick with the current implementation for now, to not to add a full other dependency for only this.

> I also find it weird to instantiate the settings object but then having to `load` them later anyway. Since `sublime.load_settings` also immediately returns a result, I would suggest loading the files in `__init__` and documenting that users shouldn't create instances at the module level during module load time and instead wait for `plugin_loaded`. You also only check your settings attributes for None in `clear_on_change` for some reason.

I extracted this code from one of my plugins so it still leaks out the way it was implemented back there. I'm changing the implementation to use a factory function.

> Your `get_settings_file_path` function ends up producing a path with two consecutive slashes for the User files.

I'm checking it.

> What is motivation for the `basepackage` parameter, which you use `__package__` for in your readme? It's not documented, but its only use seems to be for `open_settings` and assumes for the settings file to be stored in the root. This is not a requirement, so you should instead allow the user to provide an actual path or use `sublime.find_resources` and just a file name.

It's for locating the embedded default settings file for the package, for opening it. 
I think it should always live in the root of the package. 

You can look at https://github.com/KuttKatrea/sublime-toolrunner/tree/feature/better-settings-dependency for an example implementation.

FichteFoll:
> I'm changing the implementation to use a factory function.

It might just be me, but something like

```py
def fun(*args):
    return Class(*args)
```

, where the arguments are exactly identical, definitely looks like a code smell.

> I think it should always live in the root of the package.

For larger packages, such as my [PackageDev](https://github.com/SublimeText/PackageDev/), it makes some sense to store them at a different location. I'm just providing feedback here, however, having reviewed and developed many packages myself.
I mostly question the usefulness for the parameter, since I would almost always prefer using the built-in `edit_settings` command referenced in a `.sublime-menu` or `.sublime-commands` file over making my own plugin to open the default configuration file.

KuttKatrea:
> It might just be me, but something like
> 
> ```python
> def fun(*args):
>     return Class(*args)
> ```
> 
> , where the arguments are exactly identical, definitely looks like a code smell.

Agree.

> > I think it should always live in the root of the package.
> 
> For larger packages, such as my [PackageDev](https://github.com/SublimeText/PackageDev/), it makes some sense to store them at a different location. I'm just providing feedback here, however, having reviewed and developed many packages myself.
> I mostly question the usefulness for the parameter, since I would almost always prefer using the built-in `edit_settings` command referenced in a `.sublime-menu` or `.sublime-commands` file over making my own plugin to open the default configuration file.

I renamed the parameter as its usage is exactly what you describe for, to specify the path where the default settings file is.
Also, I replaced the open_file command with edit_settings. You still need to create your own command because it launches a prompt to ask which scope do you want to edit the settings for.

Also, did a complete overhaul as the original code was based in what I had implemented in another package, so I cleaned it.

FichteFoll:
Definitely looking much better now, although technically your `package_name` argument is just a `settings_name`.

Now you just need to push a valid semver tag (with three numeric parts) and I can merge. :)

Btw, you can use set literals instead of `set([])`.

KuttKatrea:
Done and done ;)

PR #7837, opened 2020-02-24
Fix project maker

pykong:
`STProjectMaker` got renamed to `ProjectMaker`.
This **PR** updates both name and URL to the package.
It also switches to tag-based release.

This is a successor to the previous **PR**: #7014

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -2453,7 +2453,7 @@
 		{
 			"name": "ProjectMaker",
 			"details": "https://github.com/bit101/ProjectMaker",
-			"labels": ["project", "template", "project-templates", "jinja2"],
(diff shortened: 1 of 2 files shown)


pykong:
```md
  - ERROR: The binding ['super+shift+p'] unconditionally overrides a default binding
    - File: Default (OSX).sublime-keyma
```

Should be fixed with `v0.1.2` of `ProjectMaker`.

FichteFoll:
Note that it is okay (in my book) to override key bindings if the package essentially replaces the original behavior (such as keeping a tab history and reopening the last closed tag using that modified behavior, or different paste semantics).

PR #7839, opened 2020-02-26
Add ST3 syntax package for Mandoc

ISSOtm:
This package provides a syntax definition and a couple commands for the [mandoc](https://mandoc.bsd.lv) markup language. Part of the aim was to be at least on par with [Man Page Support](https://packagecontrol.io/packages/Man%20Page%20Support).

This package does not duplicate nor supersede the one linked above: Mandoc is a very different language, that tries to eschew raw [gt]roff commands in favor of more semantic macros. And while the `mandoc` utility is able to process the man language (and I believe groff has some mandoc compatibility?), both languages are completely different.
As explained in a comment at the beginning of the syntax file, since direct roff usage is strongly discouraged in Mandoc documents, and the syntax is large enough as-is, that part is not supported (and will probably be considered "invalid" by the catch-all, which isn't entirely wrong).

All support I could extract from semantics should be provided, with the exception of parsing inline macros, because that's very annoying to do, and not *that* common afaik. There's a TODO comment for it if someone wants to contribute.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -343,6 +343,17 @@
 			]
 		},
 		{
+			"name": "Mandoc",
+			"details": "https://github.com/ISSOtm/sublime-mandoc",
+			"labels": ["language syntax", "mandoc", "man", "man page"],
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "ManiaScript",
 			"details": "https://github.com/Anthodev/Sublime-ManiaScript",
 			"labels": ["language syntax"],


FichteFoll:
You should try to avoid using look-behinds as that causes the slower Oniguruma regex engine to be used instead of ST's own sregex.

Almost 2 years ago, I worked on revising the existing Man Page Support syntaxes myself, but I never truly finished them and thus didn't release. If you don't mind, you could compare your syntax(es) with mine and reuse some of the parts you consider better in my version. I don't have the time to maintain such a package, but it would be sad to see the work go to waste.

[ManPage_FichteFoll.tar.gz](https://github.com/wbond/package_control_channel/files/4256772/ManPage_FichteFoll.tar.gz)

I split them into mdoc and man, because those are what the groff modules are called.

ISSOtm:
> You should try to avoid using look-behinds as that causes the slower Oniguruma regex engine to be used instead of ST's own sregex.

I have considered using `\b` instead, but that hits for example on `.Xy` after the dot. I don't think there's a way to change what `\b` means, but then I don't know how to do besides a look-behind.

I am going to take a look at your code, thanks!

FichteFoll:
When `\b` does not suffice, the next best solution is usually managing contexts in a way the patterns simply can't match, e.g. because they are not active or a different match takes priority.

ISSOtm:
I see the pattern you've used in your syntax; I think mine could use a similar pattern, but it'll take a lot of reworking.

FichteFoll:
Since I haven't taken a closer look, it might also be easier to take my version as a base and add your changes on top. In any case, you can consider my stuff to be MIT-licensed, `(c) 2018 FichteFoll <fichtefoll2@googlemail.com`.

ISSOtm:
That sounds reasonable. I will look into it later this week.

FichteFoll:
@ISSOtm ping

ISSOtm:
This is semi-alive; I've looked into it, but am currently busy with other things. I haven't dropped the idea, but may not have time before ~2 weeks.

FichteFoll:
I'll merge this for now, as it seems to be a more or less reasonable alternative to the currently available synta definition. I don't personally have the time to compile my experiments into a proper package, but I may do so and replace the two currently available packages if I consider it to be a complete replacement of both, unless my work has already been integrated into one of them. That would of course still have a obection period for the two authors/maintainers.

ISSOtm:
I still have your and my syntax definitions on my PC and open on my TODO list, but I am sadly currently busy on important projects. I still plan to merge our efforts later (as in, a couple months), but if you start before then, definitely feel free to ping me (via commit message mention, or anything really), I would definitely have some time for cooperating.

ISSOtm:
[TODO list item closed](https://github.com/ISSOtm/sublime-mandoc/commit/85f19506e3b6cc2ac1d3bb2a3b352522bf63697a)! :)

PR #7849, opened 2020-03-06
Added Solisp plugin

stuin:
Language support for Solisp


stuin:
Readme has been added

FichteFoll:
Note that ST2 can't handle trailing commas in JSON resource files.

PR #7853, opened 2020-03-16
Implemented some features

Nikita-Panyuhin:
Implemented some features from #4310:
- _Remove redundant keys that specify the general default value, notably platforms (also "platforms": ["windows", "linux", "osx"])_ | Done for platforms
- _Translate labels to lower case_ | Currently commented, because it only shows an error, but does not change the value (maybe this is okay?)

Also, changed redundant keys:
- "osx-x32" and "osx-x64" -> "osx"
- "linux-x32" and "linux-x64" -> "linux"
- "windows-x32" and "windows-x64" -> "windows"

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -2518,7 +2518,7 @@
 			"labels": ["language syntax"],
 			"releases": [
 				{
-					"platforms": ["osx", "linux", "windows"],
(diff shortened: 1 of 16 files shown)


FichteFoll:
Thanks!

PR #7854, opened 2020-03-17
added Close All Untouched Files

frame:

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1559,6 +1559,16 @@
 			]
 		},
 		{
+			"name": "Close All Untouched Files",
+			"details": "https://github.com/frame/CloseAllUntouchedFiles-sublime",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Close Oldest File",
 			"details": "https://github.com/jturcotte/SublimeCloseOldestFile",
 			"releases": [


FichteFoll:
That's an interesting use case that I don't think anyone came up with it yet.

Since you run it on all windows, you can make it an ApplicationCommand. You should also check what happens with scratch views.

frame:
Hi @FichteFoll.

Thanks for your feedback! I longed for this close-all-except-unsaved feature for a long time because I'm often previewing a massive number of individual files and while keeping a few unsaved views open where I take notes. Closing tabs individually was quite a chore because the order wasn't quite obvious to me. I originally wrote this plugin two months ago and used it daily without issues so far. Today I finally had the time to read up on how to publish a plugin and how to properly name methods, menu entries and files by looking at many, many other plugins as reference.

As suggested I converted the command type from WindowCommand to ApplicationCommand in Release 1.0.2. I wasn't aware of the difference yet and I never use multiple windows, but it makes sense. I tested it with multiple windows and it works fine. If there's multiple windows the plugin will not close the windows itself, if there's no more tabs left. Which is ok I guess: I don't want to close an opened project-folder or make the user loose access to Ctrl-Shift-T in order to undo the vanishing tabs.

As for scratch views I'm not entirely sure what you mean. Do you mean views like search results on "opened files" which open in their own tab? This very case works fine and they will get closed, however this result is actually what I personally prefer. Are you recommending to skip those views instead? Maybe as an option? Could you give me more examples of scratch views?

Thanks!

FichteFoll:
> I wasn't aware of the difference yet because I never use multiple windows, but it makes sense of course. I tested it with multiple windows and it works fine.

There isn't much of a difference between window and application commands in usage, other than their intended scope. Window commands have the window they were called for as an instance attribute, while application commands do not. You can call any command from a window because of STs internal fallback mechanisms, which looks for text as well as application commands. Text commands will be run on the currently active view, but I'm not sure if this includes widgets or behaves like `window.active_view()`.

> Do you mean views like search results on "opened files" which open in their own tab? This very case works fine and they will get closed, however this result is actually what I personally prefer. Are you recommending to skip those views instead? Maybe as an option? Could you give me more examples of scratch views?

Search results are the prime example of a scratch view. Another one would be update messages of packages provided Package Control or the OverrideAudit pane. I was just making sure you considered that case. I believe the behavior you describe is desirable, because scratch views by definition shouldn't hold valuable data.



If you haven't heard of it, I suggest installing the PackageDev package, which provides some templates for ST resource files, such as commands or menu files, and special highlighting and completions. (Disclaimer: I'm the developer.)

PR #7856, opened 2020-03-18
Add Kite Python Autocomplete package

its-dhung:

### DIFF
--- a/repository/k.json
+++ b/repository/k.json
@@ -276,6 +276,17 @@
 			]
 		},
 		{
+			"name": "Kite Python Autocomplete",
+			"details": "https://github.com/kiteco/KiteSublime",
+			"labels": ["auto-complete", "snippets", "python", "documentation"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Kivy Language",
 			"details": "https://github.com/ivlevdenis/kivylng",
 			"labels": ["auto-complete","build system","language syntax","snippets"],


FichteFoll:
> Modify the path such that our version of jinja2 gets precedence

Why?

You generally vendor a lot of packages, including some already available as [dependencies](https://packagecontrol.github.io/). Please always prefer using those over vendoring as all plugins share the same environment, including the plugin namespace. Besides, any plugin that loaded before your plugin and imported these dependencies will cause your vendored ones to be ignored since they have already been added to `sys.modules`.

---

`package.json` should be generated by PC on installation, but I see that you use it for diagnostics with the version key only, so that's probably fine.

---

I see that you add a global exception handler and usees rollbar to collect them. This must be opt-in, as per our conventions from [the last incident](https://forum.sublimetext.com/t/rfc-default-package-control-channel-and-package-telemetry/30157).

---

Have you verified that your calculation of the package's root path works for a `.sublime-package` archive? Also, do your vendored packages work in that case? There used to be some issues but I don't remember which and how to solve them.

---

I didn't review most of the rest because of the sheer amount of it, but the above issues must be addressed before I can merge.

FichteFoll:
@its-dhung ping

its-dhung:
Thanks for the comment @FichteFoll. Made some changes according to your suggestions:

* Removed the path modification for `jinja2`.
* Made error reporting opt-in, even for existing users.

Further, I had some questions:

> Have you verified that your calculation of the package's root path works for a .sublime-package archive? Also, do your vendored packages work in that case? There used to be some issues but I don't remember which and how to solve them.

I don't plan on distributing this as a `.sublime-package` archive. As I understand it, this isn't required to be listed on Package Control?

> You generally vendor a lot of packages, including some already available as dependencies. Please always prefer using those over vendoring as all plugins share the same environment, including the plugin namespace.

What happens if I remove `jinja2` from the `vendor` directory and a user without Package Control installs the Kite plugin? This is possible if a user installs Kite's Sublime plugin via Kite's builtin plugin manager. If this would cause the local install to fail, should I then include `jinja2` in both `vendor` and also in `dependency-metadata.json`?

FichteFoll:
> I don't plan on distributing this as a `.sublime-package` archive. As I understand it, this isn't required to be listed on Package Control?

By default, PC fetches a zipball of the tag with the highest precedence, checks its content for some metadata such as `dependencies.json` and `messages.json`, writes some metadata to `package-metadata.json` (which it maintaines by itself) and then repacks the package to live under a `.sublime-package` file in the Installed Packages folder. This is the preferred procedure, as it allows users to override certain files of a package, should they so desire, and is described [here](https://docs.sublimetext.io/guide/extensibility/packages.html#removing-packages). If your package cannot do that, for example because it ships executables that other processes need to read or execute, you may include a `.no-sublime-package` in your package's root folder, which tells PC to skip that last step of repacking and makes it place all files in the normal Packages folder. On every package update, that folder is then removed and replaced with the new version.



> What happens if I remove `jinja2` from the `vendor` directory and a user without Package Control installs the Kite plugin? This is possible if a user installs Kite's Sublime plugin via Kite's builtin plugin manager. If this would cause the local install to fail, should I then include `jinja2` in both `vendor` and also in `dependency-metadata.json`?

Just to be clear, it should be `dependencies.json`. And yes, should a user not be using PC and you install the package for them, they will not have `jinja2` available. You could either install jinja2 for them as well in that case or include it in your package and then only import the vendored version if a normal `import jinja2` fails. The same would be true for all your other dependencies.

I didn't know this was a use case you were thinking of. If that is the case, then vendoring is fine in my book, but you should still try to only use them if necessary.

its-dhung:
@FichteFoll thanks for the helpful information. I added a `.no-sublime-package` file and also a `dependencies.json` file. Let me know how everything else looks.

FichteFoll:
The general direction is fine, although you could be using [even more dependencies](https://packagecontrol.github.io/), such as chardet, six, and markupsafe. I can merge once you pushed a new tag with your changes.

its-dhung:
`chardet`, `six` and `markupsafe` are included in the `vendor` directory because they are dependencies of `requests` and `jinja2`. I assumed I didn't need to explicitly list them in `dependencies.json` because the Sublime dependencies already vendors the requirements?

Also I've pushed a new tag as requested!

its-dhung:
That makes sense. It looks like the `requests` dependency already vendors in `chardet`, so I added the dependencies for `markupsafe` and `python-six`. Let me know if there are more changes I need to make.

its-dhung:
@FichteFoll I assume the error downloading from GitHub is unrelated to the actual package itself?

FichteFoll:
That is indeed unrelated and a known problem, but neither myself nor Will has had the time to investigate it in more detail.

PR #7858, opened 2020-03-18
Unifying labels

Nikita-Panyuhin:
## Added the system, that can be helpful for unifying labels:

labels.json contains all labels, that are allowed.

#### How it works:
If you think, that label `C++` is redundant (because `c++` already exists) - remove `C++` from labels.json and run tests - from now on it will check each label and throw an error if the label was not found in labels.json.

However, I have already deleted some keys (and changed existing values). I'm actually not sure about `JavaScript`, but I also removed it because `javascript` was used in most packages.

Therefore, if @FichteFoll or someone else wants to choose between several instances of the same label, here is a list of those that differ only in case (lowercase / uppercase letters):

lowercase|№1|№2|№3
-|-|-|-
ascii|ASCII (2)|ascii (2)|
apex|Apex (1)|apex (5)|
bem|BEM (1)|bem (1)|
beautifier|Beautifier (2)|beautifier (1)|
blog|Blog (1)|blog (1)|
bootstrap|Bootstrap (2)|bootstrap (1)|
c|C (10)|c (10)|
c#|C# (10)|c# (10)|
c++|C++ (10)|c++ (18)|
cjsx|CJSX (1)|cjsx (1)|
cmake|CMake (1)|cmake (2)|
clang|Clang (3)|clang (5)|
clojure|Clojure (1)|clojure (2)|
coldfusion|ColdFusion (1)|coldfusion (2)|
compiler|Compiler (1)|compiler (1)|
debugger|Debugger (2)|debugger (4)|
django|Django (1)|django (2)|
erb|ERB (1)|erb (2)|
es2015|ES2015 (1)|es2015 (2)|
english|English (1)|english (1)|
expand|Expand (1)|expand (1)|
gitlab|Gitlab (1)|gitlab (1)|
ide|IDE (1)|ide (8)|
jsx|JSX (2)|jsx (8)|
legacy|Legacy (1)|legacy (2)|
lorem|Lorem (1)|lorem (2)|
macro|Macro (1)|macro (1)|
magento|Magento (2)|magento (3)|
project management|Project management (1)|project management (3)|
quick|Quick (1)|quick (1)|
r|R (3)|r (2)|
rails|Rails (1)|rails (2)|
ruby on rails|Ruby on Rails (1)|ruby on rails (2)|
svn|SVN (2)|svn (8)|
salesforce|Salesforce (2)|salesforce (5)|
scheme|Scheme (2)|scheme (2)|
terraform|Terraform (1)|terraform (2)|
time|Time (1)|time (1)|
tortoisegit|TortoiseGIT (1)|TortoiseGit (1)|
vhdl|VHDL (1)|vhdl (5)|
virtualenv|Virtualenv (1)|virtualenv (1)|

**P.S.** If approved, I will continue to work on unifying labels. For example, "-" means " " (or " " means "-").


tests/test.py
```diff
@@ -307,6 +307,9 @@ def _test_indentation(self, filename, contents):
     def _test_package(self, include, data):
         name = get_package_name(data)
 
+        with _open("repository/labels.json") as labels_file:
+            labels_list = json.loads(labels_file.read().decode("utf-8", "strict"), encoding="utf-8")
```
FichteFoll: Doing IO for every package test is not a good idea. If we are indeed going forward with this, please cache the contents of the labels file somewhere.
Also, I think a simple textfile with the labels on their own line is simpler and works just as well.

tests/test.py
```diff
@@ -439,11 +439,11 @@ def _test_release(self, package_name, data, dependency, main_repo=True):
...
-                    'base' in data
-                    and ('tags' in data or 'branch' in data)
-                    or ('sha256' in data
-                        and ('url' not in data
-                             or data['url'].startswith('http://')))
+                    'base' in data and
+                    ('tags' in data or 'branch' in data) or
+                    ('sha256' in data and
+                        ('url' not in data or
+                            data['url'].startswith('http://')))
```
FichteFoll: Using line breaks before binary operators is preferred. Please make new changes with the suggested style and don't change existing code.

FichteFoll:
I generally think having a list of allowed labels is a good idea for the current state, because I doubt they are going to change much (or are even used for search, because I have never done so myself).

However, there are still a couple duplicates in your changed list and while I know that labels are case-sensitive, we don't currently have a preference for which case to use. I believe it makes sense to use names in their original form, so programming languages are mostly in title case or allcaps for abbreviations, for example, while other labels describing functionality should be in lower case.

Ideally, I would like to hear feedback from @wbond as well before moving forward with this, but I'm positive for the direction.

Nikita-Panyuhin:
Ok, I understand.

braver:
Just so we paint a realistic picture here:

- This is pretty massive change that's a bit hard to manage. It already conflicts, and will potentially mess up existing PR's.
- The maintainers for package control are busy people, so we're currently focussed on keeping package changes and additions moving forward. 
- I agree it's a good direction, but I also think case should be correct for languages (ie. JavaScript), and that would require further changes to this PR.

So I'm not saying we shouldn't do this thing, but more that "now isn't a good time". I'll keep the PR open for now, but we need to really revisit this. Who knows what the future holds here, but don't hold your breath the entire time 😉

braver:
This idea is already partly covered by https://github.com/wbond/package_control_channel/issues/4310, let’s track it there. GitHub will remember this PR for us, we don’t need to keep it open.

PR #7859, opened 2020-03-18
Add SimpleAdd for computing sum of numeric terms.

fsaad:


FichteFoll:
I'd expect the name to contain "sum" moreso than "add", but this is still okay in my book.

PR #7860, opened 2020-03-19
Added EasyCP

RohitMazumder:
Easy CP package provides a faster way to test our code for competitive programming. It allows us to skip the tedious process of copying each sample testcase from the online judge webpage, then running our code on them, and manually checking if it matches the sample output testcase. It will be possible to do so with just a few clicks using the Easy CP package!

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -36,7 +36,7 @@
 		{
 			"name": "Easy CP",
 			"details": "https://github.com/RohitMazumder/EasyCP",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Technically, .sublime-syntax support has only been added in build 3092, but the syntax isn't essential to your package's functionality, so that's fine. I wonder why the automated tests didn't pick that up, though

PR #7864, opened 2020-03-22
Add ReactCreateComponent

acsands13:
ReactCreateComponent adds a context menu to Sublime's side bar that makes it simple
to create a React Component in a parent folder.

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -524,6 +524,17 @@
 			]
 		},
 		{
+			"name": "React Create Component",
+			"details": "https://github.com/acsands13/sublime-react-create-component",
+			"labels": ["react", "es6", "sidebar"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "React Development Snippets",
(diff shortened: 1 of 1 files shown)


acsands13:
Sorry for closing and reopening this - I wasn't sure how to re-trigger @packagecontrol-bot since it was in "Awaiting requested review" for a bit... I'm guessing it just takes more time.

FichteFoll:
Python dictionaries are not guaranteed to be ordered (until 3.7), so you usually shouldn't rely on a key being at the same index as previously regarding your quick panel selections. While the order is unlikely to change when the dict is not modified, you should either use a list or an ordereddict instead.

acsands13:
@FichteFoll thanks for merging! Do you have a specific line # you were talking about with the dictionary? I think I use OrderedDict throughout already, but maybe I missed one.

FichteFoll:
My bad, I was missing that you already created the dictionaries with an OrderedDict. Since you use the collections version, which is Python 3 only, your package won't work on ST2

acsands13:
Oh gotcha - will update and bump the version this weekend. Thanks again!

PR #7865, opened 2020-03-22
Switch to tag-based release for TrailingSpaces

rchl:
Switch from `branch` to `tags`-based release.

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -2964,7 +2964,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"branch": "master"
+					"tags": true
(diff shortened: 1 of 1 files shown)


rchl:
Current (branch-based) release version is `2020.02.05.17.36.05`.
Tag-based release will have version `v1.1.0`.
I hope ST will consider tag based release the newer one.

FichteFoll:
> I hope ST will consider tag based release the newer one.

Yes, any versioned tag will compare higher than a date-based one, maybe excluding `0.0.0` because I'm not sure about the exact implementation currently.

PR #7867, opened 2020-03-22
Add the Create WebP Image package

TomasBarry:
The `Create WebP Image` package is an interface to the [`CWebP`](https://developers.google.com/speed/webp/docs/using) tool. It can be used on `PNG`, `JPG`, `BMP`, and `GIF` files and it can optionally be opened in the browser to ensure that the conversion has created a good WebP image.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3830,6 +3830,19 @@
 			]
 		},
 		{
+			"author": ["Tomas Barry"],
+			"name": "Create WebP Image",
+			"description": "Convert images to WebP from Sublime Text",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Why don't you use the browser that `webbrowser.get` provides you with when you don't specify one when the user didn't set any?

FichteFoll:
Also, you read the window's variables in `__init__`, but I wouldn't expect new window command instances to be created per active view since you use view-specific variables. Have you verified that this works correctly?

FichteFoll:
@TomasBarry poing

PR #7868, opened 2020-03-23
Add Ancient Pond color scheme

somelun:

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -940,7 +940,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"branch": "master"
+					"tags": true
 				}
 			]
 		},
--- a/repository/a.json
+++ b/repository/a.json
@@ -934,6 +934,17 @@
 			]
 		},
 		{
+			"name": "Ancient Pond",
+			"details": "https://github.com/somelun/ancient-pond",
+			"labels": ["color scheme"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "master"
+				}
+			]
+		},
+		{
 			"details": "https://github.com/Korcholis/Andrew",
 			"releases": [
 				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
.sublime-color-scheme format has been added in the 3100s build range. Please adjust the version requirement.

Even if this color schemes is mostly for you personally, I suggest adding at least the minimal scope coverage [as documented](https://www.sublimetext.com/docs/3/scope_naming.html#color_schemes).

somelun:
Hi, thanks for the feedback, my bad, didn't fix tags in advance.

Also, that is the first time when I communicate with open source that way, so I see, that my last commits were automatically attached to this pull request, but there was a merge commit. I don't know how to deal with this, sorry, but I understand, that I need to get rid of it.

somelun:
> .sublime-color-scheme format has been added in some build in the 3100s build range. Please adjust the version requirement.
> 
> Even if this color schemes is mostly for you personally, I suggest adding at least the minimal scope coverage [as documented](https://www.sublimetext.com/docs/3/scope_naming.html#color_schemes).

Hi, thanks for the feedback, my bad, didn't fix tags in advance, and now have added minimal coverage.

Also, that is the first time when I communicate with open source that way, so I see, that my last commits were automatically attached to this pull request, but there was a merge commit. I don't know how to deal with this, sorry, but I understand, that I need to get rid of it.

FichteFoll:
> Also, that is the first time when I communicate with open source that way, so I see, that my last commits were automatically attached to this pull request, but there was a merge commit. I don't know how to deal with this, sorry, but I understand, that I need to get rid of it.

No, it's okay, since I squash all pull requests anyway (or at least package changes). For future reference, you will likely want to do a combination of rebasing and squashing in the future. There should be dozens of references in case you run into it again.

PR #7871, opened 2020-03-26
Add Tutkain

eerohele:

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -3298,6 +3298,18 @@
 			]
 		},
 		{
+			"name": "Tutkain",
+			"details": "https://github.com/eerohele/tutkain",
+			"labels": ["clojure"],
+			"description": "Clojure socket REPL client",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Twee",
 			"details": "https://github.com/monospaced/sublime-twee",
 			"releases": [


FichteFoll:
Please prefer using already [provided dependencies](https://packagecontrol.github.io/) over vendoring your own versions (pytz). For other dependencies, use relative imports if possible to minimize the effect on the global python environment. (Unfortunately I already see that `pyrfc3339` uses absolute imports internally. :unamused: )

eerohele:
Gotcha: https://github.com/eerohele/tutkain/commit/1921617039b2ce18e1bbf19994e18fa483616f4b

eerohele:
Actually, my Python-fu fails me. With edn_format, I can do:

```python
from .dependencies import edn_format
```

But `edn_lex.py` has:

```python
import ply.lex
```

Do you know if there's a way for me to ensure `edn_lex.py` finds `ply.lex` without me adding `ply` into `sys.path`?

FichteFoll:
Your only viable option would be rewriting the single import of `ply.lex` to a relative one. Since it's a one-line change, that seems pretty doable.

Other than that, you'd have to modify sys.path or sys.modules (better do the former, though).

eerohele:
There's actually quite many places in edn_format where I'd have to make the same kind of change.

I guess I could author Sublime Text dependencies for each of edn_format's dependencies, but I'm not sure I want to take on a maintenance burden like that.

Do you think it would be OK to go with the `sys.path` approach? Or should I take some other route?

Thanks for your help!

FichteFoll:
Yes, I believe in this case it is okay to go with `sys.path` modifications. If it's a rarely used library, it is unlikely to result in conflicts. When it does, however, it is hard to debug because it's not exactly a transparent issue, which is why we aim to reduce as many `sys.path` modifications as possible.

Could you verify that your package works from a `.sublime-package` archive with the `sys.path` modification as well?

eerohele:
Sorry for the delay — work has been pretty hectic lately.

Anyway, would it be OK to keep this on hold for a while longer? I've been thinking about an alternative approach that would reduce the number of dependencies here to just one. I could author it as a provided dependency to eliminate the risk of conflicts.

eerohele:
All right, Tutkain [now has zero dependencies](https://github.com/eerohele/tutkain/commit/95b43b88da26a8f88535c38fa799722092b0809d). Do you think it's still necessary to test it with `.sublime-package`, or was that only because of the `sys.path` thing?

FichteFoll:
Yes, this package shouldn't have issues with loading from `.sublime-package` anymore.

PR #7872, opened 2020-03-27
Add Expand Selection to Line Upwards

aafulei:
Add a new plugin for Sublime Text 3.

Like the built-in `Expand Selection to Line` command. But will expand upwards (instead of downwards).


FichteFoll:
You can expand a multi-line region to entire lines using `View.line`.

PR #7878, opened 2020-03-29
Add TabToSpaceConverter

maliayas:
README provides all the information.

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -167,6 +167,17 @@
 			]
 		},
 		{
+			"name": "TabToSpaceConverter",
+			"details": "https://github.com/maliayas/SublimeText_TabToSpaceConverter",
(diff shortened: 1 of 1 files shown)


maliayas:

FYI. License is mentioned in the README.

FichteFoll:
FYI, you can't just say you use the MIT license because you (or rather others wanting to redistribute your code) need a copyright notice to add to their redistribution to comply with the license's terms. I strongly suggest adding a `LICENSE` file to your repo.

This also frequently happens with code using the GPL and then just copying the verbatim license text without adding/modifying the copyright notice.

maliayas:
Thanks for the notice, FichteFoll. You're right, it's added now. Developers sometimes may not care about licensing their work but this stuff is indeed important.

Cheers!

PR #7886, opened 2020-04-02
add Silence Theme

iuliantatarascu:
light and dark color schemes.


FichteFoll:
Please only use the name "Theme" for ST themes. These are color schemes.

PR #7888, opened 2020-04-03
Add CopyGithubLink package.

ggreer:
This package takes your current open file and cursor position, and if they're in git, builds a Github URL and puts it in the clipboard.

For example: If you're editing line 42 of `github_link.py` and you run "Copy Github Link" from the command palette or the context menu, then `https://github.com/ggreer/sublime-copy-github-link/blob/master/github_link.py#L42` will be in your clipboard. This makes it easy to send a link of the code you're working on.

Example usage:

Run the copy command, then paste in slack/IRC/discord/whatever so others can view the same code.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3707,6 +3707,21 @@
 			]
 		},
 		{
+				"name": "CopyGithubLink",
+				"description": "Easily share Github links to the code you are editing. Includes your cursor's line number in the generated link.",
+				"details": "https://github.com/ggreer/sublime-copy-github-link",
+				"labels": [
+						"code sharing",
+						"github"
+				],
+				"releases": [
+						{
+								"sublime_text": "*",
+								"tags": true
+						}
+				]
+		},
+		{
 			"name": "CopyToSharePointRoot",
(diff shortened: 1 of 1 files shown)


FichteFoll:
This functionality is already provided by [Github Tools](https://packagecontrol.io/packages/Github%20Tools) (and [GitHubinator](https://packagecontrol.io/packages/GitHubinator) to a degree, since that directly opens the browser). How does your package compare to those?

ggreer:
GitHub tools doesn't have the line number in the link that it creates. GitHubinator does get the line number though. I didn't know about that package.

I guess the only difference is that my plugin puts the link in the clipboard instead of opening the browser. This makes it easier to share the link since I don't have to switch to my browser, copy the link, close the tab, then paste the link in chat or an email. I just switch to my chat client and paste.

rchl:
> GitHub tools doesn't have the line number in the link that it creates.

Are you sure? https://github.com/temochka/sublime-text-2-github-tools/blob/master/github_copy_link.py#L20

FichteFoll:
@ggreer ping

PR #7893, opened 2020-04-06
Add GitUrl

kurtmoser:
Open git repos in browser directly from Sublime Text.

Supports GitHub, Bitbucket and GitLab repos out of the box as well as option to configure your own self-hosted git services. Files tracked by git will have context menu option to open url by using either branch or commit as a reference point. Latter is helpful as even if the file changes in the future, the link still keeps pointing to the originally intended part in the code.

---

Some random repos plugin can be tested with:
GitHub: https://github.com/SublimeText/TrailingSpaces
Bitbucket: https://bitbucket.org/atlassian/amps
GitLab: https://gitlab.com/gitlab-org/gitlab
Self-hosted: https://ec.europa.eu/cefdigital/code/projects/esig/repos/dss

Self-hosted git service above requires plugin to be configured (Preferences -> Package Settings -> GitUrl -> Settings) as follows:
```
{
    "domains": {
        "ec.europa.eu": {
            "url": "http://{domain}/cefdigital/code/projects/{user}/repos/{repo}/browse/{path}",
            "url_commit": "http://{domain}/cefdigital/code/projects/{user}/repos/{repo}/browse/{path}?at={revision}",
            "url_branch": "http://{domain}/cefdigital/code/projects/{user}/repos/{repo}/browse/{path}?at=refs/heads/{revision}",
            "line": "#{line}",
            "line_range": "#{line}-{line_end}"
        } 
    }
}
```

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -851,6 +851,17 @@
 			]
 		},
 		{
+			"name": "GitUrl",
+			"details": "https://github.com/kurtmoser/sublime-giturl",
+			"labels": ["git"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "GL Shader Validator",
 			"details": "https://github.com/WebGLTools/GL-Shader-Validator",
 			"releases": [


FichteFoll:
What are the differences to https://github.com/wbond/package_control_channel/pull/7888#issuecomment-612433572 ?

The package you want to create the menu file in won't exist in the install location, because the package is installed as a `.sublime-package` archive in the Installed Packages foler. You have to ensure that it exists before writing to it. `os.path.dirname(__file__)` also won't work.

kurtmoser:
Issue with runtime context menu creation and os.path.dirname(__file__) is fixed at https://github.com/kurtmoser/sublime-giturl/commit/3fb54be3f5fa7b0035be6ae72d05136bae8f6af0

I wasn't aware of Github Tools and GitHubinator. From trying out first one is very GitHub specific and does not support other services (Bitbucket, GitLab, etc). Latter offers indeed similar functionality to my suggested plugin. Major shortcoming I see for GitHubinator is lacking support for self-hosted git services - something I am using daily at my work (self-hosted Bitbucket server in our own intranet, thus GitHubinator would be useless for my use case in its current state).
But perhaps it makes more sense to attempt to add self-hosted git services support to GitHubinator plugin itself then via another pull request.

All in all thanks for mentioning the problem with runtime menu file generation and os.path.dirname(__file__) in .sublime-package archive and feel free to reject this pull request.

FichteFoll:
> But perhaps it makes more sense to attempt to add self-hosted git services support to GitHubinator plugin itself then via another pull request.

That would be lovely. I'm sure at least some of the existing users of that package would appreciate an update and new users wouldn't have to compare two almost identical packages in order to find the one that suits their needs.

PR #7894, opened 2020-04-08
Add Python Boilerplates

bryceandpeas:
Adds a right-click context menu allowing you to quickly insert code for various python libraries and built-ins.

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -2934,6 +2934,16 @@
 			]
 		},
 		{
+			"name": "Python Boilerplates",
+			"details": "https://github.com/bryceandpeas/SublimeText-PythonBoilerplates",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


rchl:
Instead of using `.no-sublime-package` you can use `sublime.load_resource()` API to read arbitrary files from your package.

FichteFoll:
To make your package work inside a `.sublime-package` file, instead of reading from `os.path.dirname(os.path.abspath(__file__)) + '/boilerplates/'`, use `sublime.load_resource`. You could also use [sublime_lib](https://sublimetext.github.io/sublime_lib/#sublime_lib.ResourcePath).

Your command names are faily generic and prone to naming conflicts. Please prefix them with your package name or some recognizable identifier. Actually, instead of creating a dozen commands that only differ in the template to be used, you should only need one and add an argumnent that selects the template to be created from a dict of presets.

Also, I personally don't like context menus for this task and would rather use the command palette, but YMMV.

FichteFoll:
@bryceandpeas ping

PR #7896, opened 2020-04-09
Change host of MCore syntax highlighter

johnwikman:
- Changed the host of the MCore syntax highlighter to use the repository of the official group.
 - Renamed the package in preparation of additional syntax highlighters of other miking languages.
 - Added more descriptive labels to the package.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -1522,7 +1522,7 @@
 			"name": "Miking Syntax Highlighting",
 			"previous_names": ["MCore Syntax Highlighting"],
 			"details": "https://github.com/miking-lang/miking-sublime-text",
-			"labels": ["language syntax"],
+			"labels": ["mcore", "miking", "language", "syntax", "highlighting"],
 			"releases": [
 				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
Are you aware of the ability to transfer repositories? https://help.github.com/en/github/administering-a-repository/transferring-a-repository

I suggest you do do this and comment here again when you have done so.

johnwikman:
The previous repository is just a mirror to the actual [GitLab repository](https://gitlab.com/wikman/mcore-sublime-syntax). I would prefer to not transfer the mirror.

Does package control rely on that the repository history remains unchanged? In that case I could just push the original repository to the new upstream.

FichteFoll:
The advantage of transfering is that Github automatically adds redirects and moves over all extra data from the repo, such as issues or the wiki. Since you don't appear to be using either of them, as it's just a gitlab mirror, it probably doesn't matter, though. PC certainly doesn't care.

I'll merge this and in case you decide you want to transfer, you can do so onto the same URL later.

PR #7900, opened 2020-04-14
A simple plugin that interfaces with dpaste.org pastebin

adder46:

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -1464,6 +1464,16 @@
 			]
 		},
 		{
+			"name": "dpaster-sublime",
+			"details": "https://github.com/adder46/dpaster-sublime",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "master"
+				}
+			]
+		},
+		{
 			"name": "Dracula Color Scheme",
 			"details": "https://github.com/dracula/sublime",
 			"labels": ["color scheme"],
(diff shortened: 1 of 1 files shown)


FichteFoll:
You must use tags and `"tags": true` for releases instead of your master branch.

What about these other packages? https://packagecontrol.io/search/dpaste

In your dependencies file you specify requests for versions 3119 and upwards, but here you also allow installation on ST2.

Users don't need to manually satisfy dependencies when installing via PC.

adder46:
This package uses dpaste.org, some of other packages don't work at all (dpaste.de is shut down) or use dpaste.com.

TheSecEng:
> This package uses dpaste.org, some of other packages don't work at all (dpaste.de is shut down) or use dpaste.com.

In addition to the question, when using `"tags": true` you'll need to generate a Release. That is what PC looks for when using `"tags": true`

adder46:
I generated the release. What are the next steps?

TheSecEng:
Thanks, now you just sit back and wait for it’s approval. Usually pretty quick.

FichteFoll:
`0.1` is not a valid semantic version.

You need to remove the "sublime" part from your package name according to our guidelines, because it is obvious that the package is for sublime text on a sublime text package listing. The other packages I referenced predate this guideline.

FichteFoll:
Since the dpaste API seems to pretty much be the same for all of these sites, you might also want to make the domain configurable. And the expiration time. And add more syntax mappings from the other packages.

Anyway, I will merge the package for now. These are just ideas for improvement.

PR #7901, opened 2020-04-15
Add SUCC

pipe01:
Adds syntax highlighting for the [SUCC](https://github.com/JimmyCushnie/SUCC) language.


pipe01:
Well it does so..

PR #7907, opened 2020-04-25
Add Better PHPunit

dinhquochan:
Better PHPUnit support run PHPUnit on macOS.
Support terminal: `Terminal.app`, `iTerm`, `Alacritty` and `Hyper`.

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -594,6 +594,17 @@
 			]
 		},
 		{
+			"name": "Better PHPUnit",
+			"details": "https://github.com/dinhquochan/better-phpunit",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"platforms": ["osx"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Better RSpec",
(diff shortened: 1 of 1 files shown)


FichteFoll:
```

        file_name = file_name.replace(' ', '\ ')
        phpunit_config_path = phpunit_config_path.replace(' ', '\ ')
```

This code is suspicious to me. You shouldn't need to do this kind of special handling for spaces (and not have to consider other special characters at the same time).

I suggest using `subprocess.Popen` and its possibility to specify paramters as a list over `os.system` and manual fiddling with string parameters. Similarly, instead of `cd`ing into a folder to run  a command, use the `working_dir` argument.

Finally, you must provide a `.no-sublime-package` file, because otherwise your package will be installed as a `.sublime-package` zip archive and your path operations based on `__file__` will point at non-existant files that osascript won't find.

FichteFoll:
@dinhquochan ping

dinhquochan:
Dear,

This is my folk from a popular package in PHP community. So, this is why the package for years is not available in main repository. I don’t know how to edit because i’m not good at python.

Thanks.

PR #7911, opened 2020-04-27
add mariSublime

cg-cnu:
Provides integration with the 3D animation software [Mari](http://foundry.com/products/mari) for sending code using command port.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -427,6 +427,7 @@
 			]
 		},
 		{
+			"name": "Mari",
 			"details": "https://github.com/cg-cnu/mariSublime",
 			"releases": [
 				{
--- a/repository/m.json
+++ b/repository/m.json
@@ -417,22 +417,22 @@
 			]
 		},
 		{
-			"details": "https://github.com/MakiseKurisu/MarieAssembly",
-			"labels": ["language syntax"],
-			"releases": [
-				{
-					"sublime_text": "*",
-					"branch": "master"
-				}
-			]
-		},
-		{
 			"name": "Mari",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Please specify a name override that does not include "Sublime" as per our naming guidelines. "Mari" is still free. Following is a manual run of your reviewer, since the endpoint fails for some reason:


## Report for mariSublime #############################

### Repository checks ###

No failures

No warnings


### Package checks ###

Reporting 3 failures:
- The binding ['ctrl+m'] unconditionally overrides a default binding
    File: Default (Windows).sublime-keymap
- The binding ['ctrl+m'] unconditionally overrides a default binding
    File: Default (OSX).sublime-keymap
- The binding ['ctrl+m'] unconditionally overrides a default binding
    File: Default (Linux).sublime-keymap

Reporting 2 warnings:
- The package does not contain a top-level LICENSE file. A license helps users to contribute to the package.
- The command 'send_to_mariCommand' is not PascalCase
    File: mari_sublime.py
    Line: 53, Column: 1


cg-cnu:
Hi @FichteFoll Thanks for looking into this. I fixed the Pascal case and the name override. The key bindings are still messing with the defaults. Will that be alright or need to be fixed ? If nothing works out I might just remove the default bindings and let the user set them up.

FichteFoll:
Overriding default bindings in general is not acceptable, unless your package specifically replaces the previous behavior for that binding. Since your package does not do that, I suggest not providing a binding and instead explaining what command the user can bind by themselves. You already have a context menu and command palette entry, so maybe that will be enough for most users.

FichteFoll:
@cg-cnu pin

PR #7912, opened 2020-04-28
Add BuildX

cj1128:
BuildX shows sublime build output side by side. It's more convenient and comfortable than the default bottom output panel.

Currently, there is only one package [sublime-text-2-buildview](https://packagecontrol.io/packages/sublime-text-2-buildview) that provides similar functionality but it's unmaintained now and it lacks some features.

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -1611,6 +1611,16 @@
 			]
 		},
 		{
+			"name": "BuildX",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Note that your `content_replace` and `content_clear` commands are not namespaced and might be conflicting with another package providing the same commands. Ideally you'd want to use a common prefix, e.g. `buildx_`.

PR #7913, opened 2020-04-28
[SystemVerilog] Tag for ST3/ST4

TheClams:
Update of my plugin to support ST3 and ST4


FichteFoll:
You don't have a tag matching the `st4-` prefix yet and PC requires at least one release per release channel.

TheClams:
Did not know this requirement: this is now fixed.

PR #7917, opened 2020-05-01
New sublime plugin "Journal" added.

aschulz01:
Journal is a simple Sublime Text plugin that makes it easy to create Journal files for logs, notes, diary, daybooks.... . These files are saved in single location for easy searching and access later.
The files are structured hierachical by date (year\month\day\yyyymmdd.*extension* ). Due to the file naming convention (yyyymmdd) you can - for example - sort the file names by asc and generate a date-ordered summary file.

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -759,6 +759,17 @@
 			]
 		},
 		{
+			"name": "Journal",
+			"details": "https://github.com/aschulz01/Journal",
+			"labels": ["journal", "diary", "day book", "hierarchical file journal"],
+			"releases": [
+				{
+					"sublime_text": "*",
+            				"tags": true
+				}
+			]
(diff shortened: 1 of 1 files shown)


FichteFoll:
Instead of using `/` in some and `\` in other places when joining paths, use `os.path.join` .

Your command to insert a timestamp only inserts at `self.view.sel()[0].begin()` and doesn't handle multiple selections or replacing the current selection for that matter. If you used `view.run_command("insert", {"characters": …})` you would get that for free.

You may want to use the `edit_settings` command for opening settings instead of `open_file`. It's not available for older ST builds (especially ST2), but nobody should still be using a version this old.

FichteFoll:
@aschulz01 ping

PR #7918, opened 2020-05-03
Add ForceMoveToGroup Plugin

tkotosz:
Adds a simple command to allow moving tabs between groups a bit more easily than with the build in "move_to_group"


tkotosz:
@packagecontrol-bot try again

PR #7922, opened 2020-05-05
Add package details for Imaginarium syntax

ethanrobison:
This is a custom syntax highlighting package for users of [Ian Horswill's Imaginarium](https://github.com/ianhorswill/Imaginarium). This package is built for the benefit of his research lab and the students in his class on procedural content generation. That said, Imaginarium is a tool for everyone, and so this syntax highlighting benefits anyone who might care to do casual, constraint-based procedural content generation.

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -312,7 +312,7 @@
 			"description": "Syntax highlighting for Imaginarium",
 			"releases": [
 				{
-					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


FichteFoll:
I don't know this syntax, but usages of variables/identifiers should be named `variable`, not `entity.name`. `entity.name` is for definitions of such.

PR #7923, opened 2020-05-07
adding Penman

writeml:
Hello World!

Penman allows users to write well-formatted prose in plain-text using a lightweight markup language called WriteML. It's ideal for authoring fiction of any length, from short stories to novels.

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -735,6 +735,16 @@
 			]
 		},
 		{
+			"name": "Penman",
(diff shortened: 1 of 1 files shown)


writeml:
I'm closing this PR -- I'm going to refactor Penman for ST4 -- Now that I've had the chance to read through the new docs for it, I realize that it'll be better in the long-run to just start targeting ST4 now than to only half-support ST3 later...

PR #7927, opened 2020-05-11
new package: DeepCodeAI

Arvi3d:
The Sublime package provided by [DeepCode.ai](https://www.deepcode.ai) finds bugs and critical vulnerabilities in your code. We support Java, Python, JavaScript, TypeScript and C/C++.

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -398,8 +398,8 @@
 			"labels": ["code analysis", "security"],
 			"releases": [
 				{
-					"sublime_text": "*",
-					"branch": "master"
+					"sublime_text": ">=3000",
+					"tags": true
 				}
 			]
 		},
--- a/repository/d.json
+++ b/repository/d.json
@@ -395,7 +395,9 @@
 		{
 			"name": "DeepCodeAI",
 			"details": "https://github.com/DeepCodeAI/sublime-plugin",
-			"labels": ["code analysis", "security"],
+			"labels": ["code analysis", "security", "typescript", "javascript", "java", "python", "c/cpp"],
+			"homepage": "https://www.deepcode.ai",
+			"donate": null,
 			"releases": [
 				{
 					"sublime_text": ">=3000",
--- a/repository/d.json
+++ b/repository/d.json
@@ -393,6 +393,17 @@
 			]
 		},
 		{
+			"name": "DeepCodeAI",
+			"details": "https://github.com/DeepCodeAI/sublime-plugin",
+			"labels": ["code analysis", "security"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "master"
+				}
+			]
+		},
+		{
 			"name": "Default File Type",
 			"details": "https://github.com/spadgos/sublime-DefaultFileType",
 			"releases": [


Arvi3d:
@packagecontrol-bot I'm sure we need this file .no-sublime-package

FichteFoll:
It would seem that you need `.no-sublime-package` for [this function](https://github.com/DeepCodeAI/sublime-plugin/blob/34461d960b00dbf07e856ff17a269335d2323bd6/analyze.py#L68-L78), but the referenced `MODULE_DIR` does not seem to be included in your package distribution and I don't see it being downloaded either.

Arvi3d:
this dir is created at run-time. We store python dependencies there.
Maybe I misunderstood something, but i thought .no-sublime-package is required in this case

FichteFoll:
Yes and no. Yes, you probably would need such a file to access files/modules there, but you could and should just store them at a different location, e.g. `os.path.join(sublime.cache_path(), "DeepCode")`, so you don't need `.no-sublime-package`.

Besides, does the user need something else than [this pypi module](https://pypi.org/project/deepcode/)? If so, asking the user to install that manually through `pip install --user deepcode` sounds like a better option to me, which is also what SublimeLinter asks users to do.

Arvi3d:
thank you for valuable comment.
We will change the path as you described.

As for asking users to use pip. I'm not a big fan of this solution:
- This is an additional effort for users
- It's hard to manage versioning and dependency between our plugin and that module. 

I'll write another comment here once that path change is done.
Thank you

FichteFoll:
Will the dependencies you install be separate from the system's Python installation? I see that you use the `-t` option for pip but I want to be extra sure that the user wouldn't have to do perform a manual cleanup after this package was installed.

I also just noticed that you [modify the plugin host's environment](https://github.com/DeepCodeAI/sublime-plugin/blob/93cda9d9fcef022ec44635e40dfd4dbabf802835/deepcode_sublime.py#L18). Please don't do that and instead adjust those for the subprocess calls via its `env` argument only. You also shouldn't need to patch the system's `PATH` variable, especially not unconditionally.

Arvi3d:
Yes, we install dependencies like virtualenv does, in a separate folder only and don’t interfere with system python.

I’ll double check this.

> On 27 May 2020, at 17:58, FichteFoll <notifications@github.com> wrote:
> 
> 
> Will the dependencies you install be separate from the system's Python installation? I see that you use the -t option for pip but I want to be extra sure that the user wouldn't have to do perform a manual cleanup after this package was installed.
> 
> I also just noticed that you modify the plugin host's environment <https://github.com/DeepCodeAI/sublime-plugin/blob/93cda9d9fcef022ec44635e40dfd4dbabf802835/deepcode_sublime.py#L18>. Please don't do that and instead adjust those for the subprocess calls via its env argument only. You also shouldn't need to patch the system's PATH variable, especially not unconditionally.
> 
Thanks. Will fix it.

Arvi3d:
@FichteFoll All requested changes are done(tag v0.0.4).
Could you please check it?

FichteFoll:
Just as a heads-up, you can use `os.path.join` to build OS-agnostic paths, i.e.
```py
# instead of:
MODULE_DIR = "{}{}deepcode_lib".format(sublime.cache_path(), os.path.sep)
# use:
MODULE_DIR = os.path.join(sublime.cache_path(), "deepcode_lib")
```

Arvi3d:
Thank you @FichteFoll 
I updated our code with that fix, added a new version tag. How long does it take usually to be loaded to Package Control?

FichteFoll:
It may take up to 1 hour until the new tag is found.

PR #7929, opened 2020-05-14
Remove 1Self; Plugin no longer active; Domain no longer active

TheSecEng:
**What**
Remove 1Self, sadly, even though in the Top 100, it is not maintained. The domain it calls out too http://app.1self[.]co is no longer active.

### DIFF
--- a/repository/0-9.json
+++ b/repository/0-9.json
@@ -24,18 +24,6 @@
 			]
 		},
 		{
-			"name": "1Self",
-			"details": "https://github.com/1self/sublime-text-plugin",
-			"labels": ["time tracking", "activity"],
-			"previous_names": ["QuantifiedDev"],
-			"releases": [
-				{
-					"sublime_text": "*",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "2pdf",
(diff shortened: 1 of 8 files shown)


FichteFoll:
I'm a bit undecided about removing technically available packages. The package hasn't seen any attention in 5 years, which is fine, but its upstream server isn't available available anymore and it can thus be considered unusable. We don't have any policies for this situation in placed, as I don't think we have ever removed such a package before.

Through https://whois.domaintools.com/1self.co, I found the following information:

```
Registrar Status 	pendingDelete, redemptionPeriod
Dates 	2,259 days old
Created on 2014-03-13
Expires on 2021-03-12
Updated on 2020-05-07
```

Let's give @douglas-vaz some time to sort this out before we jump to action.

TheSecEng:
Yes, I will ping them. I closed this as I’ll re-lone. / submit upon updates / no contact with them.

douglas-vaz:
Apologies for just seeing this. 1self has long expired and is no longer maintained. Please feel free to remove it. Thanks, @TheSecEng @FichteFoll

PR #7939, opened 2020-05-26
add unofficial Deno package

nickypy:
This package provides basic Deno support and Typescript syntax highlighting.

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -512,6 +512,16 @@
 			]
 		},
 		{
+			"name": "Deno",
+			"details": "https://github.com/nickypy/deno-sublime",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Dependents",
 			"details": "https://github.com/dependents/Dependents",
 			"releases": [


rchl:
There is already a package with typescript syntax? How is it different? Since deno has not created its own language, it just wraps JS engine and typescript into one package, it doesn't really make sense to create separate syntax for it. It's just typescript after all.

nickypy:
It currently supports `deno fmt` in the editor via Command Palette. There’s work to integrate other cli commands.

rchl:
It's fine to add a package that is adding those features. I'm talking about the syntax only.

nickypy:
Is the issue that the package includes syntax files? I included those to prevent the need to install another TS package, though I’d be glad to remove them to avoid a possible conflict.

rchl:
One reason it's not a good idea is the fragmentation. It creates confusion for users. It requires them to pick one. If they have both installed, it creates conflicts as it's not very well defined which one is used then.

Also can you guarantee that your syntax will be kept up-to-date or it will eventually stale and become one of those unmaintained ones (it's not easy to keep syntax up-to-date with latest developments of the language).

Also some packages like LSP depend on the syntax path (at least in ST3) and they are configured for existing syntaxes only. If you are adding a new one, those need to updated to take the new syntax into account.

nickypy:
Sounds good. I’ll make those changes

nickypy:
Hey @rchl, repo's updated and made a new release with the changes we discussed earlier.

rchl:
Here goes my personal opinion about this package in its current state: I wouldn't want to use the formatting command manually like that. Normally formatting is handled through formatting plugins that run on save. In the current form this package would be kinda useless to me. But feel free to disagree and continue with it. Maybe it will become something great in the future. :)

FichteFoll:
Personally, I'm fine with running formatting commands on demand and not automatically on every safe operation. However, by the nature of how this works you should only allow running it on saved files when there are no unsaved changes in the current view, i.e. `view.is_dirty() == False`, since the external tool operates on a file level and you don't communicate with it through IO streams.

Thanks to @rchl for the comments on including a copy of the syntax definitions, however.

thgh:
Note that Deno adds some things on top of Typescript making it not 100% the same, for example: imports must include file extensions which is not allowed in TS. Also the Deno global namespace and URL imports.

PR #7942, opened 2020-05-26
Added Lucene Query Formatter

nishant-ns19:
Created and added lucene query formatter plugin for better readability of query object strings.


nishant-ns19:
Greetings, 
How much time will it take to reflect changes in Sublime Text ?

PR #7944, opened 2020-05-28
Add new package "JapanWordStop"

woodmix:
This is a plugin to make word jumps such as ctrl+right appropriate for Japanese strings.

There are plugins for the same purpose, "JapaneseWordSeparator" and "Japanese Word Jump", but the logic for determining the jump stop position is completely different.
They use a library called "tinysegmenter", but "JapanWordStop" simply uses the character type to determine the stop position.

I'd like to offer this plugin as an expansion of choices, not a question of superiority or inferiority.

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -121,6 +121,17 @@
 			]
 		},
 		{
+			"name": "JapanWordStop",
+			"details": "https://github.com/woodmix/JapanWordStop",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"platforms": ["windows", "linux"],
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
Thank you for including information on the differences between the two existing packages as that means I don't have to ask about it.

Please use a more unique identifier in your `add_on_change` call. You also must remove the registered callback inside a `plugin_unloaded` to ensure your references are cleaned up when the plugin is unloaded.

LGTM otherwise.

You might want to at least add an English sentence to the readme explaining what the pacakge does, however.

woodmix:
Thank you for your advice!
I have fixed an issue related to add_on_change.
The English version of the README has been added (mostly by machine translation).

PR #7945, opened 2020-05-29
Add Github Theme

mauroreisvieira:
# GitHub Sublime Text theme

GitHub Sublime Text theme

## Settings

```json
{
    "show_sidebar_disclosure": "false"
}
```


FichteFoll:
From the screenshot, I presume that the theme is adaptive. You may want to mention that in the readme somewhere.

PR #7946, opened 2020-05-31
Add Notepad++ Tab Context plugin

josephmorrill:
Adds some useful tab context menu commands from Notepad++:

- Close Tabs to the Left
- Save
- Save As...
- Rename
- Delete
- Open containing folder in file explorer
- Open containing folder in terminal
- Open in default viewer
- Copy file path
- Copy filename
- Copy directory path

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -920,7 +920,7 @@
 			"labels": ["tab context", "notepad++"],
 			"releases": [
 				{
-					"sublime_text": ">3211",
+					"sublime_text": ">=3000",
 					"tags": true
 				}
 			]
--- a/repository/n.json
+++ b/repository/n.json
@@ -915,6 +915,17 @@
 			]
 		},
 		{
+			"name": "Notepad++ Tab Context",
+			"details": "https://github.com/josephmorrill/SublimeNppTabContextPlugin",
+			"labels": ["tab context", "notepad++"],
+			"releases": [
+				{
+					"sublime_text": ">3211",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Notes",
 			"details": "https://github.com/tbh1/sublime-notes",
 			"labels": ["language syntax", "documentation", "notes"],


repository/n.json
```diff
@@ -914,6 +914,17 @@
 				}
 			]
 		},
+		{
+			"name": "Notepad++ Tab Context",
+			"details": "https://github.com/josephmorrill/SublimeNppTabContextPlugin",
+			"labels": ["tab context", "notepad++"],
+			"releases": [
+				{
+					"sublime_text": ">3211",
```
FichteFoll, line 923: This prevents installation on 3211, the latest available build. Is this what you want?

The last ST3 dev build is 3210 and would be exempted by this selector as well, despite being almost equivalent.

FichteFoll:
For your replacements, you may want to look into `sublime.expand_variables` and `Window.extract_variables` respectively. This will reuse ST's internal expansion algorithm and provide the same variables as available for build systems.

For the settings menu, I suggest using the `edit_settings` command instead of two separate calls to `open_file`. See the default menu file for reference.
Furthermore, you should add a `.sublime-commands` file with that for easier keyboard access.

ST passes `group` and `index` arguments for commands invoked through the tab context menu if `{ "group": -1, "index": -1 }` are provided in the menu file. You don't seem to be using these and instead always operate on the currently active view. Can you check and confirm whether the context menu of an inactive tab does what you'd expect?

repository/n.json
josephmorrill: Thanks! Updated the filter to >=3000

josephmorrill:
@FichteFoll Thanks for the feedback. I made the changes you suggested (with the exception of using Window.extract_variables, which seems to only target the active tab) and tagged version 1.1.0

FichteFoll:
Good point regarding the active view requirement. While the missing variables can easily be filled/replaced with manual code, I guess that could be turned into a feature request for the API.

PR #7949, opened 2020-06-02
Add DlangAuto{FIleName,Import,ModuleName}

vitalfadeev:

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -1124,6 +1124,16 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
I suggest `view.match_selector(0, "source.d")` instead of checking the view's syntax setting for checking whether a command should be available.

vitalfadeev:
view.match_selector(0, "source.d") - ok.
Thanks for suggestion. It helpfull.

PR #7950, opened 2020-06-02
Sublime SVG Preview

chunqiuyiyu:


FichteFoll:
I suggest checking out ImageMagick's `convert` command line tool for an alternative to inkscape when it is not available.

chunqiuyiyu:
@FichteFoll Hi, thanks for your suggestion. Sometimes ImageMagick's `convert` command can not convert SVG to PNG correctly, so I chose Inkscape as background for this plugin.

PR #7951, opened 2020-06-03
Add Nunjucks syntax package

jeremyfourna:
Hello, I forked the package of https://github.com/mogga/sublime-nunjucks that is syntax highlighting for Nunjucks template language. The repository is not maintained anymore.

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -1230,6 +1230,17 @@
 			]
 		},
 		{
+			"name": "Nunjucks Syntax Highlight",
+			"details": "https://github.com/jeremyfourna/sublime-nunjucks-highlight",
+			"labels": ["language syntax", "Nunjucks"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "NuSMV Syntax",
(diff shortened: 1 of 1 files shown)


FichteFoll:
It would have been nice if you made an actual fork and preserved the commit history, so I could have seen which changes you applied instead of having to compare the files manually, which I'm not going to do.

I'll also inform you that this is technically a copyright infringement, because there was no license in the repository you copied from. However, this is just a heads-up, as this seems quite unlikely to involve sensitive property (and PC itself is not violating any terms).

Finally, if you're continuing the existing work, I'd suggest to replace the existing entry for [Nunjucks Syntax](https://packagecontrol.io/packages/Nunjucks%20Syntax) instead of adding a new one. @mogga, you may object to a replacement of your package within the next 14 days.

jeremyfourna:
@FichteFoll I did as you said, I put in the `n.json` the fork of the original package.

The only change on the fork is the support of the `.njk` file name.

FichteFoll:
Note that to make Pacakge Control pick up the changes, you need to push a tag with a higher version. I will merge afterwards.

FichteFoll:
@jeremyfourna ping

jeremyfourna:
@FichteFoll where should I push a tag of higher version?

PR #7954, opened 2020-06-07
Update Tutkain description and labels

eerohele:

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -3316,8 +3316,8 @@
 		{
 			"name": "Tutkain",
 			"details": "https://github.com/eerohele/tutkain",
-			"labels": ["clojure"],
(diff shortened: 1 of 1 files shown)


repository/t.json
```diff
@@ -3317,7 +3317,8 @@
 			"name": "Tutkain",
 			"details": "https://github.com/eerohele/tutkain",
 			"labels": ["clojure"],
-			"description": "Clojure socket REPL client",
+			"description": "An interactive Clojure editing environment for Sublime Text",
+			"labels": ["clojure", "nrepl", "repl"],
```
FichteFoll, line 3321: You now have `labels` twice.

eerohele: That's what you get when you try to make a change like this at midnight… fixed.

PR #7955, opened 2020-06-07
userinterface.js snippets

thoughtsunificator:


FichteFoll:
Well, this is quite a simple one, but I don't think we have a policy regarding some "minimum value" a package needs to add.

The same comment as with your other package applies regarding the lac of any content in the readme.

PR #7959, opened 2020-06-13
Replace Zeal package with a fork

FichteFoll:
Replaces @vaanwd's Zeal package with a fork that I updated.

I submitted a PR 30 days ago at https://github.com/vaanwd/Zeal/pull/30 without a reply, so I'm suggesting to replace the package and offer to maintain it in the future, which is why I adopted it under the @SublimeText org umbrella.

@vaanwd, you may object to this change within the next 14 days.

### DIFF
--- a/repository/z.json
+++ b/repository/z.json
@@ -34,12 +34,16 @@
 		},
 		{
 			"name": "Zeal",
-			"details": "https://github.com/vaanwd/Zeal",
-			"author": "vaanwd",
+			"details": "https://github.com/SublimeText/Zeal",
+			"author": ["vaanwd", "FichteFoll"],
 			"releases": [
 				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
For the record, the referenced PR was merged but the other necessary steps for proper distrubution have not been applied, which means the package isn't in a great position right now. Notably, it simply won't work on ST2 and users were never made aware of their outdated config (if any), because they won't see an update message.

braver:
> the package isn't in a great position right now

Well the fork doesn't worsen the situation. @vaanwd hasn't objected here or really interacted in the original repo. I don't see anything stopping us from merging this PR and thereby completing the takeover. I'll give it another 2 weeks, but will merge unless someone stops me before then.

PR #7961, opened 2020-06-14
Highlime package added

alex-pobeditel-2004:
Good mood plugin which changes current color scheme parameters in real time to make it look *a bit different*


FichteFoll:
0.2 seems pretty aggressive, but that seems intended.

PR #7965, opened 2020-06-16
Make Make Executable work on sub3

HormyAJP:
There's nothing preventing this from working. Just
need to change the repository to support all versions.

N.B. I wrote my own version of this and was going to publish that. However, since this one already exists we might as well just reuse it.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -272,7 +272,7 @@
 			"details": "https://github.com/cockscomb/SublimeMakeExecutable",
 			"releases": [
 				{
-					"sublime_text": "<3000",
+					"sublime_text": "*",
 					"branch": "master"
 				}
 			]


FichteFoll:
The package won't work properly on ST3 because it does not respect lazy loading of the API (and using `Pref = Pref()` is an anti-pattern that really needs to disappear). I suggest you fork the repo, make the required changes and submit a PR to replace the original with your fork. There is [an open issue](https://github.com/cockscomb/SublimeMakeExecutable/issues/4) for that package, but it's been without replies for 3 years.

HormyAJP:
@FichteFoll thanks, understood. I'll do as you suggest. Will close this PR.

Thanks,
Andrew

PR #7968, opened 2020-06-20
Add new package " L042ATeX "

k1242:
The package is a set of plugins, snippets and shotcuts for simplifying using LaTeX for users writing text in Russian. Templates and snippets targeting specially for Russian, so package will be hardly useful for other users. Russian users could begin special environments, commands, etc. using the Russian layout.

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -2,6 +2,18 @@
 	"schema_version": "3.0.0",
 	"packages": [
 		{
+			"name": "L042ATeX",
+			"details": "https://github.com/k1242/L042ATeX",
+			"labels": ["LaTeX", "latex", "russian", "snippets"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": ["windows"],
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


FichteFoll:
I'm sceptical about [this vbs script](https://github.com/k1242/L042ATeX/blob/master/sources/vbs/close_recent_file.vbs). It seems to be used to close a tab for a moved file. Instead, you should use the undocumented `View.retarget` method with the new file name. I don't really understand what you use the change_layout.vbs script for, but if you want to change the layout of an ST window then you can totally do that within ST itself by calling the command that the key is bound to. You can find out via the console by running `sublime.log_commands(True)`.
For the draft article skeleton, you could use [`sublime_lib.ResourcePath.copytree`](https://sublimetext.github.io/sublime_lib/#sublime_lib.ResourcePath.copytree). See the readme referenced at the top for how to use `sublime_lib` in your package.

With that out of the way, it should be possible to target all platforms, in which case you also want to use `os.path.join` to join path segments.

FichteFoll:
@k1242 ping

PR #7969, opened 2020-06-22
Add dprint

dsherret:
This adds a sublime editor extension for [dprint](https://github.com/dprint/dprint), which is a code formatter.

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -1545,6 +1545,16 @@
 			]
 		},
 		{
+			"name": "dprint",
+			"details": "https://github.com/dprint/dprint-sublime",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Dracula Color Scheme",
 			"details": "https://github.com/dracula/sublime",
 			"labels": ["color scheme"],


FichteFoll:
[This code line](https://github.com/dprint/dprint-sublime/blob/5068686f4d3199905c04409f57f4663c691cb0b9/dprint.py#L8) cannot work, as `DprintExec` is undefined at that point.

Instead of just trimming `\r` at the end of the command output, you need to strip it from all line breaks as ST only uses `\n` internally and will render a placeholder for `\r` in your text/code. I suggest using `universal_newlines` in the Popen call, as well as `encoding='utf-8'`, so you don't need to do that manually. 

`view.file_path()` may return `None` for unsaved files.

If you don't have any settings to configure, you shouldn't include a `.sublime-settings` file.

dsherret:
@FichteFoll thanks for the pointers! I will do these fixes soon.

For the purpose of just getting approved would this be able to get merged? I'm asking because it's taken over a month to get to this point and perhaps it might just be better to merge and then I can do the fixes after (as nobody is using this yet).

FichteFoll:
It literally doesn't work in its current state, so no, it cannot be merged. The `\r` issue could also negatively affect usage experience on Windows, although people are of course free to open a ticket at your repo instead.

I apologize for not getting to reviewing this earlier, but in order to maintain a certain level of quality in our package listing, a manual review process is necessary and I am currently the only person to do so. I promise you won't have to wait as long for replies in the future.

FichteFoll:
@dsherret ping

PR #7970, opened 2020-06-24
My New Base64 Encoding Plugin

coekuss:
This plugin encodes and decodes Base64 and provides a function to generate usable Base64 data URLs for embedding within HTML and CSS files. This will save web developers the time of having to type "data:image/png;base64," and using an online Base64 encoder for each time they need to use Base64.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3557,6 +3557,16 @@
 			]
 		},
 		{
+			"name": "CoolBase64",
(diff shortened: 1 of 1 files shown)


FichteFoll:
How does this compare to the existing packages that provide base64 encoding functionalities? https://packagecontrol.io/search/base64

You're not closing file handles properly, which can lead to problems on Windows. Use `open` as a context .

PR #7972, opened 2020-06-29
Thrustmaster's TARGET programming language syntax highlighting and completion

Zander-Labuschagne:

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -299,6 +299,17 @@
 			]
 		},
 		{
+			"name": "Target",
+			"details": "https://github.com/Zander-Labuschagne/target",
+			"labels": ["Thrustmaster", "T.A.R.G.E.T.", "TARGET", "Thrustmaster Advanced pRogramming Graphical EdiTor","HOTAS", "autocompletion", "auto-complete"],
+			"releases": [
+				{
+					"sublime_text": ">=3211",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Task",
 			"details": "https://github.com/samsonw/SublimeTask",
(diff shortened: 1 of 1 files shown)


Zander-Labuschagne:
It looks like I cannot have `/* some text */` in my completions file, is there some other way I can have a commented part in my snippet completion line? I'm sorry if I sound stupid, this is my first editor plugin/addon/package ever and I'm not familiar with YAML or seasoned with JSON.

FichteFoll:
> It looks like I cannot have /* some text */ in my completions file,

You can. [I see you figured out your mistake already](https://github.com/Zander-Labuschagne/TARGET/commit/57ce174c1e0bb27724acd29566e56616910a7f02#diff-a8bcd64292377a5cabe842893d9d936eL251).

Note that by default Package Control does not install pre-relaeses and requires user configuration prior to installing your package. You should push a non-pre-release tag.

Your scoping of punctuation characters is inconsistent, as you don't assign them to the opening match for a string, for example. ([here](https://github.com/Zander-Labuschagne/TARGET/blob/5e0f8e1d7c8202fc3cdba6794ae1ac49be0a4d68/target.sublime-syntax#L111)) You may also want to assign punctuation scopes to grouping characters, such as parens.

You generally only want to scope text tokens, not leading or following whitespace. ([here](https://github.com/Zander-Labuschagne/TARGET/blob/5e0f8e1d7c8202fc3cdba6794ae1ac49be0a4d68/target.sublime-syntax#L135))

Zander-Labuschagne:
Thank you for your input and help. I have made some changes, I hope the changes were in line with what you asked, I am still finding my way around creating syntax definitions.

PR #7973, opened 2020-06-30
Adding Clarity language package

dmitrytorba:
This is a package for supporting the smart contract programming language [Clarity](https://docs.blockstack.org/core/smart/overview.html). 

Repo: https://github.com/dmitrytorba/clarity

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1365,6 +1365,16 @@
 			]
 		},
 		{
+			"name": "Clarity",
+			"details": "https://github.com/dmitrytorba/clarity",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Class Navigator",
 			"details": "https://github.com/malexer/SublimeClassNavigator",
 			"labels": ["code navigation"],


FichteFoll:
Since this seems to be an addon for LSP, I suggest naming it `LSP-Clarity` as per the naming convention. /cc @rwols

rwols:
Yes, it should be named LSP-Clarity because that will make the plugin load at a later point in time than the base LSP plugin. It has no significance at this point but will in the future.

Also consider making a pull request over at https://github.com/sublimehq/packages/ to add `.clar` to the file extension list of Lisp.sublime-syntax.

rwols:
I'm also somewhat confused by the NpmClientHandler inheritance. It seems the language server is written in Rust and has pre-built binaries available on the release page: https://github.com/lgalabru/clarity-lsp/releases/tag/2020-05-27

A similar kind of language server and convenience downloader is LSP-TexLab. You might want to look at how that convenience plugin handles that server: https://github.com/sublimelsp/LSP-TexLab

dmitrytorba:
thanks for the review, I will take a look

rwols:
The amount of LSP-* packages is getting out of hand, so we've made our own package repository. Please submit your pull request there: https://github.com/sublimelsp/repository/

PR #7974, opened 2020-07-03
Add the Abella Package

JimmyZJX:
Support for the Abella theorem prover (https://abella-prover.org/), including syntax highlight and interactions with the proof assistant.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -69,6 +69,17 @@
 			]
 		},
 		{
+			"name": "Abella",
+			"details": "https://github.com/JimmyZJX/SublimeAbella",
+			"labels": ["language syntax", "proof assistant"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "ABNF-sublime-syntax",
 			"details": "https://github.com/ZxxLang/ABNF-sublime-syntax",
 			"labels": ["abnf", "language syntax"],


JimmyZJX:
These key bindings are defaults of another popular GUI prover "CoqIDE", and should be acceptable when using provers. Proof scripts are more focused on the logic (proof states) and not code refactoring or other advanced techniques.

FichteFoll:
You cannot access most of ST's API at plugin load time ([here](https://github.com/JimmyZJX/SublimeAbella/blob/6de233dd8065ae6394bb83be5803a967439313d1/abella.py#L18)). You must defer unti `plugin_loaded` has been called.

`edit_panel` is a pretty generic command name. Since the command namespace is shared, you should prefix that command like your others.

JimmyZJX:
> You cannot access most of ST's API at plugin load time ([here](https://github.com/JimmyZJX/SublimeAbella/blob/6de233dd8065ae6394bb83be5803a967439313d1/abella.py#L18)). You must defer unti `plugin_loaded` has been called.
> 
> `edit_panel` is a pretty generic command name. Since the command namespace is shared, you should prefix that command like your others.

Thanks for the suggestions. I just solved these issues in my master branch. Please kindly review.

JimmyZJX:
> [Your change](https://github.com/JimmyZJX/SublimeAbella/commit/93ba38c32a39d7ed59729b795ea90244ae7bd0ca#diff-a680752278db23e59f830d063aa01112R21) won't work, as `settings` will be treated as a function-local variable and not override the global. You need a `global settings` line before that. You also need to push a new tag with those changes then, so that PC can pick up the new release.

Thanks for pointing out the mistake! I've fixed the problem and published a new release.

PR #7976, opened 2020-07-03
Added EasyWinPath package repo

DeusDeorum1:
I haven\`t found any packages that have a functionality of escaping and unescaping backslahes for win-styled paths, which was a huge pain for me while editing json files. Hence I`ve created my own package that does this. In addition it can open selected path in file explorer or in a default executor (i.e. notepad for .txt)

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -207,6 +207,17 @@
 			]
 		},
 		{
+			"name": "EasyWinPath",
+			"details": "https://github.com/DeusDeorum1/EasyWinPath",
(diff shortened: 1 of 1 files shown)


FichteFoll:
It's unnecessary to print a message that your plugin is loaded because ST itself prints a message for every plugin it loads. (Besides, you plugin is only actually loaded if it can be imported fully, i.e. you'd need the print at the end of the file.)

DeusDeorum1:
@FichteFoll that message was just for debugging which I\`ve just forgotten to delete. Thanks for the advice though, I\`ve learned something:)

PR #7977, opened 2020-07-03
add map preview package

doneill:
This package provides ability to preview spatial map data files in your default web browser on a map.  Current supported map formats are [GeoJSON](https://geojson.org/) & [TopoJSON](https://github.com/topojson/topojson). Open a valid GeoJSON or TopoJSON file and run the Map Preview command from the command pallete, valid files will be parsed and displayed inside users default web browser overlayed a world map.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -408,7 +408,7 @@
 		},
 		{
 			"name": "MapPreview",
-			"details": "https://github.com/doneill/st3-map-preview",
+			"details": "https://github.com/doneill/MapPrevieww",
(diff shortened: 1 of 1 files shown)


doneill:
Added a VERSION file with semver tagged

doneill:
Tagged and created v0.0.5 release

doneill:
Should be good to go now

doneill:
What now?  I have addressed the change requested, do I need to submit another PR?

PR #7980, opened 2020-07-05
CheckiO

oduvan:


FichteFoll:
All plugins in ST share the same namespace, so your `run` and `check` commands are way to generic and prone to conflicts. Prefix them with your package name, for example.

Instead of using `shlex.split` on your string for Popen, why not build a list in the first place?

FichteFoll:
@oduvan ping

PR #7985, opened 2020-07-17
Add Dashboard

learning:
Add Dashboard for Sublime Text 3

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -100,6 +100,17 @@
 			]
 		},
 		{
+			"name": "Dashboard",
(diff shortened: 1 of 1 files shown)


FichteFoll:
The readme is very uninformative. Please include some info on what the package is about, ideally with a screenshot.

When accessing assets from your package via the file system, you must either provide a `.no-sublime-package` file to prevent PC from packing up your package or, ideally, use `sublime.load_resource`.

FichteFoll:
@learning ping

PR #7986, opened 2020-07-19
Add "Swift Format" plugin

aerobounce:
### Added "Swift Format" - Swift file formatter plugin for Sublime Text 3.
- It utilizes [nicklockwood/SwiftFormat](https://github.com/nicklockwood/SwiftFormat).
- Since `swiftformat` is only available on macOS platform, this package is also only available on the platform.
- It is basically a wrapper package with a useful features such as: 
    - Format on Save
    - Showing Error using Phantom API

### Motivation
- As of the time writing this request, there is no formatter plugin dedicated to Swift language.


aerobounce:
Oops, dumb mistake. Fixed. I'll wait for next check time.

FichteFoll:
Instead of having a setting for every CLI option, why not accept a single string or list of additional options to pass to the formatter? That gives the user more flexibility, is forwards-compatible to added options and requires much less code.

aerobounce:
First of all, thanks for reviewing!

You really have a point. Although that was the result of me preferring kind of 'typed' or 'static' way of coding, in this case it was too much.
I've made some changes, could you take a look please?

https://github.com/aerobounce/Sublime-Swift-Format/compare/v1.0.2...v.1.1.0

PR #7989, opened 2020-07-20
CodeFall Color-Scheme Added

EncryptEx:


EncryptEx:
Now I realise that I modified someones code... sorry now I'll change it (  - C Base Converter)

PR #7994, opened 2020-07-24
NoLine (Package Submission)

Weilory:
Sublime Text does not handle white space and blank line generating among particular formats very well. For example, if I convert a txt file to a js file, it would add excess blanks between each line of code. Many stack overflow topics address this issue: https://stackoverflow.com/questions/12008986/sublime-text-2-how-to-delete-blank-empty-lines. 
After install, you can find `NoLine` in Menubar -> Edit. write some dummy codes separated by lines (one or more), then run `NoLine`, it should remove excess blanks meanwhile keep the indentation lines.

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -813,6 +813,17 @@
 			]
 		},
 		{
+			"name": "NoLine", 
+			"details": "https://github.com/Weilory/sublime-noline",
+			"labels": ["blank", "remove blank", "empty line"], 
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		}
+		{
 			"name": "Non Text Files",
 			"details": "https://github.com/bordaigorl/sublime-non-text-files",
 			"labels": ["file navigation", "preview", "file open"],
--- a/repository/n.json
+++ b/repository/n.json
@@ -822,7 +822,7 @@
 					"tags": true
 				}
 			]
-		}
+		},
 		{
 			"name": "Non Text Files",
 			"details": "https://github.com/bordaigorl/sublime-non-text-files",


Weilory:
Sorry I didn't reply in time, busy for last month about examinations. Describe how the issue has happened to me: I have two computers with different OS. The working one is always left at the office, I usually email the files I need to be working on back to the home one. Working one is a HP widows, home one is a Mac. When I went back home, opened mail, download the file to local machine. However, after I open it with sublime, the extra blank lines that were never there before, were added. The pattern is always that add \n after each line, whatever it is a line of code or line break.

Weilory:
@FichteFoll, this issue may has been caused due to email, which the files are encoded before transferring, however not decoded very well when received.

Weilory:
For the method you have suggested: 

```python

        contents = self.view.substr(sublime.Region(0, self.view.size()))
        self.view.erase(edit, sublime.Region(0, self.view.size()))
        self.view.insert(edit, 0, contents.replace("\n\n", "\n"))

```

Sorry, I can't utilise it since the user may have used double line purposely as indentation.

Weilory:
the difference of this plugin to the SO method, is that it <b>distinguishes between double line indentation and additional blank line, and remove only the blank line</b>, without changing any extra of the costumer's original code.

FichteFoll:
> When I went back home, opened mail, download the file to local machine.

That shouldn't happen. I suspect your email program or something else in the process converts `\r` to `\n`, which leaves you with two `\n` for each Windows-style line break. Either way, it does not sound like a ST issue at least. I suggest sending files as attachments, alternatively inside a zip archive, to circumvent this.

> the difference of this plugin to the SO method, is that it distinguishes between double line indentation and additional blank line, and remove only the blank line,

You're right about that, I didn't think of this at the time. However:

> Sorry, I can't utilise it since the user may have used double line purposely as indentation.

I don't see the problem. The code should reproduce your algorithm 1:1 but much more efficiently. Did you try it?
Regardless, my very first suggestion fro performing a simple string replacement inside the editor still applies and I would consider it easier to use than installing this package and more flexible on top of that.

PR #7999, opened 2020-07-28
Add wcmatch dependency

facelessuser:
This adds the dependency for [wcmatch](https://github.com/facelessuser/wcmatch) in Sublime Text.

It provides enhanced file globbing and matching.


FichteFoll:
I generally suggest mirroring the version from upstream, so it can be better compared when and whether it is outdated.

PR #8000, opened 2020-07-29
Add Slyblime package

s-clerc:
This package integrates with the preexisting Slynk server from the Sly IDE to offer a rich Lisp development experience. There is no similar package.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -2378,6 +2378,18 @@
 			]
 		},
 		{
+			"name": "Slyblime",
+			"description": "A rich Lisp development environment by integrating Sly's functionality.",
+			"details": "https://github.com/s-clerc/slyblime",
+			"labels": ["lisp", "auto-complete", "REPL", "references"],
+			"releases": [
+				{
+					"sublime_text": ">=4079",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "SM RAD",
 			"details": "https://github.com/setap/sm-rad",
 			"labels": ["language syntax"],


s-clerc:
@packagecontrol-bot

s-clerc:
@packagecontrol-bot

s-clerc:
* The keybinding things are there for the REPL to work correctly.
* `.no-sublime-package` is strictly necessary.
* There is a `COPYING.md` file.

s-clerc:
> `()` are illegal characters for identifiers, so not sure what you expect [this line](https://github.com/s-clerc/slyblime/blob/668824da66d54e0846a080c33e92fd5b8ef48a04/sly.py#L20) to do.

Unfortunately, that's a mistake. I've committed a fix, and it will appear in the next tagged release.

> You create a loop in [`plugin_loaded`](https://github.com/s-clerc/slyblime/blob/668824da66d54e0846a080c33e92fd5b8ef48a04/sly.py#L213) but then only start it in the [connect command](https://github.com/s-clerc/slyblime/blob/668824da66d54e0846a080c33e92fd5b8ef48a04/sly.py#L109) despite also scheduling tasks [from other commands](https://github.com/s-clerc/slyblime/blob/668824da66d54e0846a080c33e92fd5b8ef48a04/completion.py#L123). This may simply not occur within your plugin lifecycle, but I just want to confirm you thought about this.

Yeah, in the case you cite that `run` function only runs when a completion is generated which only occurs when a connection was established and the loop is running. I do this because nothing can run until we're connected as the async operations all require said connexion. I'm mostly just waiting for 

> Why do you bundle a `Lis`[`p`]`.sublime-syntax`? Is the default one not satisfactory? Why not submit a PR against the default since you're only targetting recent ST builds anyway? 

It isn't and [I did submit a PR](https://github.com/sublimehq/Packages/pull/2312), but to allow new users to start using it as soon as possible I've bundled it. In addition, it's possible that in the future there are more plugin-specific scopes.

> Besides, I strongly suggest to not use the `name` field and instead have ST infer the syntax name from the file name, as that will be generally less confusing, especially wrt the required `.sublime-settings` file name.

Yeah, I can change that, but I was worried that then there are two Lisp entries in the panel which would confuse people.

PR #8003, opened 2020-08-05
Add WebIDL (syntax highlighting)

vinyldarkscratch:
As a part of my work, I work with WebIDL files a lot, so I decided to put together a syntax highlighting file for the language.  This PR introduces this brand new package so that others may use it as well!

### DIFF
--- a/repository/w.json
+++ b/repository/w.json
@@ -262,6 +262,16 @@
 			]
 		},
 		{
+			"name": "WebIDL",
+			"details": "https://github.com/vinyldarkscratch/WebIDL-Sublime",
+			"releases": [
+				{
+					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


vinyldarkscratch:
Odd, I definitely have a readme for my package.  I'd love to get some assistance as to how to resolve the issue, or if I simply need not worry about it!

wbond:
I believe this is hitting a but in the current implementation of the GitHub client, where if no branch info is specified in the URL, it is defaulting to `master`. We'll need to update https://github.com/wbond/package_control/blob/master/package_control/clients/github_client.py#L163-L169 to have it fetch the "primary" branch name from the API rather than assuming it is `master`.

PR #8005, opened 2020-08-06
new package: jumpy

jgburet:
Intent is to navigate easily in your code.

https://github.com/tednaleid/sublime-EasyMotion does not support ST3.
https://github.com/ice9js/ace-jump-sublime is complicate.

This allows to jump across views.
Just like Atom's https://github.com/DavidLGoldberg/jumpy , it is dead simple: just enter what you see in the screen, no need to search with some starting letter or anything.

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -1362,6 +1362,16 @@
 			]
 		},
 		{
+			"name": "Jumpy",
+			"details": "https://github.com/jgburet/jumpy-sublime",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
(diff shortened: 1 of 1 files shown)


wbond:
I'm looking into this now

wbond:
I spent some time working on this yesterday and this morning. Something is funky with the channel server sometimes when downloading things. It worked fine running the development server on my Mac, so I think it is Linux-specific, and likely to do with OpenSSL. Hopefully I'll have it all worked out in the next day or so.

wbond:
Based on the one successful run from the dev server, I believe the automated checks passed.

PR #8006, opened 2020-08-08
Add LSP-svelte

ihodev:


rchl:
Do you envision that the separation for ST3 and ST4 will be needed? I would say that generally, it shouldn't be needed for those packages that use lsp_utils. The lsp_utils itself handles that.

For the languages configuration you can just include configuration for both ST3 and ST4 there and lsp_utils will pick the right one for the currently running version.

rchl:
See https://github.com/sublimelsp/LSP-html/blob/master/LSP-html.sublime-settings for example.

And typically you don't even need that as lsp_utils will automatically convert scopes to document_selector.

ihodev:
> For the languages configuration you can just include configuration for both ST3 and ST4 there and lsp_utils will pick the right one for the currently running version.

Thanks, I didn't know that 🙄 But there is one more reason to use such versioning strategy (actually that's why I decided to give it a try): when you decide to end supporting of st3, no things to do, just do not push new tags for it. Of course there is some overhead, but I do not think that is to big in this case.

ihodev:
But if the reason is to be consistent in terms of versioning `LSP-*` plugins I can revert changes.

rchl:
The separation can be postponed until really needed. No reason to do it yet IMO.

You choose what to do but it's just a hassle IMO to have to maintain and release two versions, especially as they are not any different.

One of the points of lsp_utils was to also handle that case - abstract handling of different ST/LSP versions.

The consistent versioning is not and issue and packages don't sync on that.

ihodev:
I see you points, doing some cleanup, thanks for explanation :)

PR #8010, opened 2020-08-10
Upgrade to Emmet 2

sergeche:
This PR upgrades Emmet package to v2, which now uses different update strategy (via GitHub releases) and lives in new repo.
A follow-up of https://forum.sublimetext.com/t/best-strategy-to-update-emmet-plugin/53015

### DIFF
--- a/channel.json
+++ b/channel.json
@@ -6,6 +6,7 @@
 		"https://bitbucket.org/klorenz/sublime_packages/raw/tip/packages.json",
 		"https://donkirkby.github.io/live-py-plugin/sublime-package/package.json",
 		"https://eoa-cdn.s3.amazonaws.com/extensions/packages.json",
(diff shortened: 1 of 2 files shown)


sergeche:
Any chance to be accepted with this test failing?
Or should I setup proxy server which will return `200` response for this URL?

rchl:
The warning about missing LICENSE seems valid as the artifact referenced from the JSON file (https://github.com/emmetio/sublime-text-plugin/releases/download/v2.0.2/Emmet.zip) doesn't include LICENSE.md.

The warnings about key-bindings are probably not an issue as you are using them in specific context only.
(Well, I know issues where your `tab` handler is conflicting with default "complete" handler but it's sort of expected for those who want to use this package and can be refined later.)

sergeche:
Added LICENSE file to v2.0.7

PR #8014, opened 2020-08-12
Adds the BetterTwig package.

Ultra-Instinct-05:
This PR adds the [BetterTwig](https://github.com/Sublime-Instincts/BetterTwig) Package to Package Control.

This package provides syntax highlighting, snippets, completions & more for PHP twig templates.

There is an existing package already which is [PHP-Twig](https://packagecontrol.io/packages/PHP-Twig), however the package hasn't been updated to keep up with latest syntax improvements via `.sublime-syntax` for ST3. So this package offers,

- An updated & tested syntax file which follows the official scope naming guidelines as closely as possible.
- Snippets for Twig code blocks.
- Completions for built in tests, filters, functions & loop variables.
- Ability to use <kbd>ctrl + /</kbd> for Twig comments.
- Certain key bindings to make life a little bit easier.

I decided to keep `BetterJinja` & this package separate because there are some differences between the templating languages.

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -681,6 +681,17 @@
 			]
 		},
 		{
+			"name": "BetterTwig",
+			"details": "https://github.com/Sublime-Instincts/BetterTwig",
+			"labels": ["syntax", "twig", "highlighting", "php"],
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "BeyondCompare",
 			"details": "https://github.com/npadley/BeyondCompare",
 			"labels": ["diff", "difference", "diff/merge", "file comparison"],
--- a/repository/b.json
+++ b/repository/b.json
@@ -626,6 +626,16 @@
 			]
 		},
 		{
+			"name": "BetterJinja",
+			"details": "https://github.com/Sublime-Instincts/BetterJinja",
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "BetterSave",
 			"details": "https://github.com/adrian-tut/BetterSave",
 			"labels": ["save", "close"],
--- a/repository/b.json
+++ b/repository/b.json
@@ -628,6 +628,7 @@
 		{
 			"name": "BetterJinja",
 			"details": "https://github.com/Sublime-Instincts/BetterJinja",
+			"labels": ["syntax", "jinja2", "highlighting"],
 			"releases": [
 				{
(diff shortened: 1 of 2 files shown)


Ultra-Instinct-05:
Not sure why the commits from the `add_betterjinja` branch is appearing here ? Should the previous PR have been merged first before this ?

rchl:
It's appearing because you've included it in the changes. :)

Ultra-Instinct-05:
Any ideas how I can fix the issue ?

rchl:
Just remove the `BetterJinja` block from this PR.

Ultra-Instinct-05:
I have no idea about this though :-
```
- ERROR: Invalid JSON (with comments)
    - File: resources/settings/BetterTwig.sublime-settings
```
I removed the settings file.

rchl:
It's checking your tagged version, not the master branch.

purplefish32:
I am OK with you replacing my bundle with yours as I no longer use ST (at least as of now) and have lost interest in maintaining the package. If you based your code of of mine (I haven't checked) can you please credit me, that would be cool.

Ultra-Instinct-05:
@purplefish32 Sorry, but your package uses the older `.tmLanguage` format & mine uses the more recent `.sublime-syntax` format. Also this is not converted using ST's built in syntax converter (as that would also preserve all the HTML specific syntax that is scoped in the `.tmLanguage`) and mine uses `with_prototype` to `push` the default HTML syntax and provides scopes only for the Twig specific syntax.

So, this is neither based on your code nor is a port of your work. It is a completely rewritten package.

FichteFoll:
@Ultra-Instinct-05 since we have the OK, please go ahead with the necessary changes.

Ultra-Instinct-05:
@FichteFoll Looks like the warning is for the key binding (which anyway will not trigger in a non twig file). So I think this is good for merging.

FichteFoll:
With "necessary changes" I was actually refering to replacing the existing Twig package with yours (for builds that support the new .sublime-syntax format) and probably adding an update message. See the [INI package](https://github.com/wbond/package_control_channel/pull/8020/files) for a reference.

Ultra-Instinct-05:
(Sorry for the additional questions, since I am doing this for the first time), I have pushed a JSON object with `BetterTwig`, do I have to push a commit that removes that and changes the `details` key of the original `Twig` JSON object to point to the new GitHub URL (while keeping the old Twig package for < 3092 in `releases` key as per the PR you referenced) ?

FichteFoll:
That is correct.

Ultra-Instinct-05:
Ok. All the updates have been done. If there are any other issues, let me know.

PR #8016, opened 2020-08-14
Add Razor C# Syntax package.

LoneBoco:
Adds Razor C# Syntax for Sublime Text 4.  The existing Razor syntax packages are abandoned and/or don't work anymore.

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -460,6 +460,18 @@
 			]
 		},
 		{
+			"name": "Razor C# Syntax",
+			"details": "https://github.com/LoneBoco/razor-sublime",
+			"description": "Razor C# .cshtml language syntax for Sublime Text 4.",
+			"labels": ["language syntax", "razor", "c#", "cshtml"],
+			"releases": [
+				{
+					"sublime_text": ">=4075",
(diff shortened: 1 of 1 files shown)


LoneBoco:
https://github.com/LoneBoco/razor-sublime/commit/6d71fff83539809789e87c6004533b991eb37963

I set the base scope to `embedding.razor.cs` and I updated the C# scopes to `source.cs.embedded.razor`.

rwols:
Does this collide with [this package](https://github.com/michaelblyons/SublimeSyntax-HTML-CSharp)?

/cc @michaelblyons

michaelblyons:
@rwols Kind of, but I'm not going to quibble if @LoneBoco wants his own package. The Razor available in my package is almost entirely the work of @keith-hall PRs. I'm happy to look at more PRs, though, if OP is interested in that.

No pressure from me in either direction.

keith-hall:
A separate package called Razor (as per this PR) would be more discoverable, I think, which is a good thing :tm: But consolidation is useful too, so no pressure from me either way either, in case it matters :)

PR #8020, opened 2020-08-18
Add INI Syntax package

jwortmann:
This package provides syntax highlighting for INI and Windows Registry (.reg) files. There already exist two other packages for this (https://packagecontrol.io/packages/INI and https://packagecontrol.io/packages/REG), but they both use the old .tmLanguage format, provide only very minimalistic highlighting, and don't really follow the scope naming guidelines.

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -679,6 +679,17 @@
 			]
 		},
 		{
+			"name": "INI Syntax",
+			"details": "https://github.com/jwortmann/ini-syntax",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Initializr",
 			"details": "https://github.com/drinks/sublime-initializr",
 			"labels": ["automation", "utils"],
--- a/repository/i.json
+++ b/repository/i.json
@@ -670,19 +670,14 @@
 		},
 		{
 			"name": "INI",
-			"details": "https://github.com/clintberry/sublime-text-2-ini",
-			"releases": [
-				{
-					"sublime_text": "*",
-					"branch": "master"
-				}
-			]
-		},
-		{
-			"name": "INI Syntax",
 			"details": "https://github.com/jwortmann/ini-syntax",
 			"labels": ["language syntax"],
 			"releases": [
+				{
+					"base": "https://github.com/clintberry/sublime-text-2-ini",
+					"sublime_text": "<3092",
+					"branch": "master"
+				},
 				{
 					"sublime_text": ">=3092",
 					"tags": true


jwortmann:
> You could push a context and set a meta_scope on the first matched character that disables the completion.

That would probably work, but it seems a bit awkward in my opinion to add such a scope to the whole file just for this completion. I think I'd like to stay with the few lines of plugin code for this, for simplicity.

> you can of course add an ST version restriction so the package will only be updated if their ST supports the new format.

Oh of course, I actually have this already due to using the `.sublime-syntax` format. :smile:
Do you have experience with replacing packages and know whether Package Control will pick up the correct new version then for former users of the INI and REG packages, because their current version numbers are `2016.07.09.20.30.02` and `2.0.0`, while mine is `1.4.0` at the moment?

Regarding `meta.mapping.key`/`meta.mapping.value` in the default schemes, I opened https://github.com/sublimehq/sublime_text/issues/3590, let's see whether there will be some reactions about it.

I also added the missing file extensions and added a separate syntax and completions for `.editorconfig` files now.

FichteFoll:
> because their current version numbers are `2016.07.09.20.30.02` and `2.0.0`, while mine is `1.4.0` at the moment?

Branch(Commit)-based versions are smaller than `0.0.1` in comparisons, but you'll have to push something higher than `2.0.0` to trigger updates for REG users.

deathaxe:
If the package name INI is to be used for the new package, the content of the current repo should propably be added as `st2` branch to the new repo in order to be able to further provide it for ST2 users via tags.

The package control entry would then look as:

```
		{
			"name": "INI",
			"details": "https://github.com/jwortmann/ini-syntax",
			"previous_names": ["REG"],
			"releases": [
				{
					"sublime_text": "<3092",
					"tags": "st2-"
				},
				{
					"sublime_text": ">=3092",
					"tags": true
				}
			]
		},
```

... with REG being removed.

FichteFoll:
Alternatively, you could also do the following:
```
        {
            "name": "INI",
            "details": "https://github.com/jwortmann/ini-syntax",
            "releases": [
                {
                    "base": "https://github.com/clintberry/sublime-text-2-ini",
                    "sublime_text": "<3092",
                    "branch": "master"
                },
                {
                    "sublime_text": ">=3092",
                    "tags": true
                }
            ]
        },
```
which would be the minimal required effort to keep the old version still installable for the 100 ST2 users on this planet.

jwortmann:
I've updated the PR to replace the INI package. I'm still unsure if it would be beneficial to replace the REG package as well and hence I didn't remove it, because there are a few additional features or possible user configurations which don't exist in my package.

jwortmann:
Maybe a bit off-topic here, but I had another idea of a possible workaround for the `meta.mapping.key string` issue in the default color schemes. The INI package could ship with a supplementary file for each of the default color schemes (e.g. `Mariana.sublime-color-scheme`) with a color adjustment like this:
```json
{
    "rules": [
        {
            "scope": "meta.mapping.key.ini string",
            "foreground": "var(pink)"
        }
    ]
}
```
The `ini` suffix ensures that it only applies to files which are targeted by this syntax, and users could still create an override with a different color in their User package if they want.

A drawback would be, if a user already has a more general override with `meta.mapping.key string` for the default color schemes, that it wouldn't apply to INI files then. Also it would feel a bit odd if a package ships with such override files for the default schemes, and I wonder if such a practice would be acceptable...

FichteFoll:
> The INI package could ship with a supplementary file for each of the default color schemes (e.g. `Mariana.sublime-color-scheme`) with a color adjustment like this:

That would be a possibility, although one that requires careful consideration regardings its potential effects. Regardless, I'm gonna merge the PR as is now. Took long enough already. 🙂

jwortmann:
Right, I guess I will only consider this if there will be an essential amount of requests for it.

PR #8029, opened 2020-09-01
add MapleMBSE MSE repository

blake-regalia:
This package provides syntax highlighting and code completions for MapleMBSE MSE Configuration files. https://github.com/Open-MBEE/MSE/tree/master/sublime-syntax

The syntax is generated from a complete implementation of the language's grammar specification, resulting in highly-specific scoping and invalid syntax detection. At the time of submission, this is the only package that provides support for the MSE Configuration language.

The .sublime-package file for this package 'MapleMBSE MSE Configuration', is built via CI and then posted to the repository's GitHub releases. This repository reflects the correct packages.json for its latest releases and references the URL to the generated sublime-package file.

Repository JSON file included in this PR: https://raw.githubusercontent.com/Open-MBEE/MSE/master/sublime-syntax/channels/sublime/package-control.json

### DIFF
--- a/channel.json
+++ b/channel.json
@@ -80,6 +80,7 @@
 		"https://raw.githubusercontent.com/NicholasBuse/sublime_packages/master/packages.json",
 		"https://raw.githubusercontent.com/nLight/Phing/master/packages.json",
 		"https://raw.githubusercontent.com/noraesae/ClassHierarchy/master/packages.json",
+		"https://raw.githubusercontent.com/Open-MBEE/MSE/master/sublime-syntax/channels/sublime/package-control.json",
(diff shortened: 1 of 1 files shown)


FichteFoll:
There is potential of a race condition between the PC crawler updating the reported version of the channel and the download contents the URL points to being updated, so I suggest to include the version in the download path. Since you seem to be using CI to build the artifacts anyway, it should be possible to include the release version in the path as well. Not a hard requirement, but a suggestion from my side.

blake-regalia:
Yeah okay cool. I went ahead and made asset commits tagged w/ semver so that they get a stable URL, e.g., see [package-control.json#L18](https://github.com/Open-MBEE/MSE/blob/master/sublime-syntax/channels/sublime/package-control.json#L18).

PR #8030, opened 2020-09-02
Add PHP CBF Package

andremacola:
This is a lightweight ST Package to fix your php files with [PHP Codebeautifier](https://github.com/squizlabs/PHP_CodeSniffer/wiki/Fixing-Errors-Automatically) aka `phpcbf`. It is based on the original [PHP_CodeSniffer](https://github.com/andremacola/sublime-PHP_CodeSniffer) package without the `phpcs`.

The advantage of this plugin is that it acts directly in the Sublime Text buffer, avoiding file reloading thus being faster.

https://github.com/andremacola/sublime-PHP_CBF

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -987,6 +987,18 @@
 			]
 		},
 		{
+			"name": "PHP CBF",
+			"details": "https://github.com/andremacola/sublime-PHP_CBF",
+			"issues": "https://github.com/andremacola/sublime-PHP_CBF/issues",
+			"labels": ["php", "phpcs", "formatting"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "PHP Codebeautifier",
 			"details": "https://github.com/Ennosuke/PHP-Codebeautifier",
 			"releases": [


andremacola:
Added license

andremacola (commented):
added license to repo

FichteFoll:
The convention is to name [these command entries](https://github.com/squizlabs/sublime-PHP_CodeSniffer/compare/master...andremacola:master#diff-db82941e6d15b7a54e5c0eb5ef1f756bR7) `Preferences: PHP CBF Settings`.

Since you starte operations from a fork, you aren't allowed to simply attach a license to that and call it your own. We don't want to deal with copyright infringments from our side, especially because they can be very finnicky, but you should be aware of this. Ideally you'd get a license from the original author (by asking for it in an issue or email) and then add your name to it.

andremacola:
> The convention is to name [these command entries](https://github.com/squizlabs/sublime-PHP_CodeSniffer/compare/master...andremacola:master#diff-db82941e6d15b7a54e5c0eb5ef1f756bR7) `Preferences: PHP CBF Settings`.
> 
> Since you starte operations from a fork, you aren't allowed to simply attach a license to that and call it your own. We don't want to deal with copyright infringments from our side, especially because they can be very finnicky, but you should be aware of this. Ideally you'd get a license from the original author (by asking for it in an issue or email) and then add your name to it.

This package is only based on, still.... please see https://github.com/squizlabs/sublime-PHP_CodeSniffer/issues/1 I have the authorization.

FichteFoll:
> This package is only based on

Sure, but that is how copyright functions in most legislations. It was mostly meant as a heads-up anyway in an attempt to spread awareness about license-less repositories, although you are not at fault. The written authorization in the referenced issue is definitely enough for me personally, though, again, I'm not going to precautiously intervene in any such issues until copyright claims are made.

PR #8034, opened 2020-09-07
Added 2 repositories: GIMP Auto-Complete Scheme ... and ... Gimp Auto-Complete Python

civAnimal:


civAnimal:
I have added a 3rd repository : A Sublime Text Auto-Complete plugin for TinyScheme (a dialect of Scheme programming language).

FichteFoll:
Please open a separate pull request for each package in the future.

You probably want to exempt `comment` scopes from your completions, too.

While your completions are created using Python plugins, there isn't any intelligence to them, which is what we commonly refer to as "normal completions" and not "auto-completions". Please adjust the package names to reflect that. Your TinyScheme completions even have a resemblence of snippets (considering they include entire function signatures and some are multi-line).

civAnimal:
Thanks for your observations. I'm going to try and answer them one by one in the following paragraphs.

**Separate PR for each Package**
I'm deeply sorry for the mix-up. I'm new to Github and trying to figure things out without causing too much headache to others. Thanks for being so accommodating.

**Exempt `comment` Scope**
The plugins don't entertain triggers within `comment` scope. They are already exempted. Am I a missing something here, or are you suggesting me to enable them within `comment` scope? Please, do clarify. I'll be happy to make any necessary changes in this regard.

**Adjust Package Names**
I am surprised at your objection at the use of the word "Auto-Complete" in my plugin names. I think they are accurately and descriptively named. "They automatically complete the functions/constants/constructs as the user types a few letters of the relevant item." This facility/feature is widely known and acknowledged by the term "Auto-Complete".

In support of my claim, I am submitting a few definitions of the term "Auto-Complete" by well-reputed linguistic sources as under:

**Oxford**
_"A software function that completes words or strings without the user needing to type them in full."_

**Cambridge**
_"A computer program that automatically finishes a word that someone has started to type."_

**Merriam-Webster**
_"A feature found in many computer programs ... that attempts to predict and automatically complete the current word or phrase as it is being entered by the program's user."_

I firmly stand by the names as they are. They are not misleading, deceptive, or inaccurate.

FichteFoll:
> Exempt `comment` Scope

I was referring to [this line (and similar)](https://github.com/civAnimal/gimp_autocomplete_python/blob/5aae5c4752cce61df81fa535992a15910748aace/gimp_autocomplete_python.py#L33). A user can still manually request compeltions through ctrl+space (or a different OS-dependant binding).

> Adjust Package Names

In an attempt to separate static completions that are unaware of the context they are provided in, we started to differentiate these two concepts both by name and by label. Unfortnately, this only happened a couple years back and was not applired retrospecticely to already existing packages, which is why even the labels sometimes not accurate. 

Regardless, this convention is probably unique to our package listing and most certianly cannot be found in a dictionary, though I wouldn't be too surprised to see it used elsewhere. 

If you insist on your naming, then it would raise the question whether we can even manage this separation, considering it's not applied consistently. It's a decision I don't want to make on my own, that will require some discussion with the community, and that takes effort.

civAnimal:
**Exempt `comment` Scope**
As advised, this change has been applied to all 3 plugins.

**Adjust Package Names**
I appreciate the fact that you/(the team of devs) are trying hard to make a more clear distinction between different categories of plugins. It's good. Everybody should support it. There's nothing wrong with your objective.

"Auto-Complete" is a generic term. Just like "car" is a generic term. Before electricity-driven cars arrived, we used to call them all "cars". Now we call the new ones Electric-Cars. These cars had a distinct feature, so we added that feature-tag to the generic name and started calling them Electric-Cars. The others are still called "cars". Have we imposed a ban on companies, asking them to stop calling their products "cars"? No. That's not how things work. When significant improvements are made to a class of products, the new ones get a new name/category. The old ones are not declassed. They continue to serve their intended purpose. Similarly, you should consider allocating a new and distinct category to those sophisticated variants of Auto-Complete plugins. Perhaps, Advanced-Auto-Complete. You know, something along these lines. I'm sure you'll think of something interesting.

FichteFoll:
I don't have the time or energy to tackle this *very* broad issue, so as I said earlier I'll just give up. It will never see adoption outside of our ST packages bubble anyway.

PR #8039, opened 2020-09-12
Add MKB Package

KeeMeng:
New package for minecraft macro/keybind mod, adds many functions (Syntax Highlighter, Auto Complete, Popup Hints, Indenter, Linter) to help people code with this custom language.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -1817,7 +1817,7 @@
 			"releases": [
 				{
 					"sublime_text": ">=3092",
-					"tags": true
+					"tags": true,
 				}
 			]
 		},
--- a/repository/m.json
+++ b/repository/m.json
@@ -1812,6 +1812,16 @@
 			]
 		},
 		{
+			"name": "MKB",
+			"details": "https://github.com/KeeMeng/MKB-Syntax-Highlighting",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "MLFi",
 			"details": "https://github.com/li-vu/st-mlfi",
 			"labels": ["language syntax", "mlfi", "csml"],
--- a/repository/m.json
+++ b/repository/m.json
@@ -1817,7 +1817,7 @@
 			"releases": [
 				{
 					"sublime_text": ">=3092",
-					"tags": true,
+					"tags": true
 				}
 			]
 		},
--- a/repository/m.json
+++ b/repository/m.json
@@ -1816,7 +1816,7 @@
 			"details": "https://github.com/KeeMeng/MKB-Syntax-Highlighting",
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": ">=3092",
 					"tags": true
 				}
 			]


FichteFoll:
Your scope names are very much non-standard. I strongly suggest to read and apply the [scope naming docs](https://www.sublimetext.com/docs/scope_naming.html). You may also want to add some `\b` word boundary anchors to your matches to not highlight a part of `falseish`.

As it is, you are adding a new top-level menu entry to the main menu. That's quite unconventional and usually those menus are put under a submenu in the Edit or Preferences menus.

Don't the completions in your `.sublime-completions` file and your plugin overlap?

KeeMeng:
I'll change the scope names to the more conventional ones, and double check the matches.
Regarding the auto complete, the `on_query_completions` in the plugin is different from the `sublime-completions` file, if you look closely the auto complete in the plugin is a capitalized version of the completions file. 
Also, the people who are using the plugin asked for the menu to be easier to access, but I'll ask if putting it under Edit is ok. 
Thanks for your comments!

braver:
> Also, the people who are using the plugin asked for the menu to be easier to access,

Sure, everyone wants their favourite commands in the easiest to reach place. That's why everyone can modify pretty much everything in Sublime Text to their liking. You should try to add your commands in places where people can expect to find them, e.g. in Edit or Tools.

Please report back when you've also checked the scope names. The way you've set it up right now, hardly any color scheme will be able to make sense of your syntax and you're missing out on other features as well.

KeeMeng:
Ok I'll change the menu commands. 

But for the scope names, I don't think I can change them. 
Currently the people who use this package copy the colour scheme into the USER colour scheme. I did this so the other languages would look normal, and my MKB language would look like the original language's editor.
If this doesn't make sense, or you have feedback, please comment!

braver:
You're focussing too much on the colors. The syntax definition of a languages, the scope names, are used for many underlying systems in Sublime Text. Features like goto definition, the symbol indexes, autocompletions, and more use the understanding of what's what that a syntax provides. 

And since you're using a fully custom color scheme, of course you can also change the scope names. You have full control there. Just make sure you're specific enough, but what `source.mkb keyword.operator.mkb` looks like is entirely in your hands. There is no need to resort to scopes like `source.mkb.operators` which is essentially gibberish to Sublime Text and all other color schemes.
And that would open it up for users to not be stuck with whatever the "original editor" looked like and use any color scheme that they like.

In Sublime Text it's actually possible to change the color scheme only for a particular language, via syntax specific settings. You could also use that mechanism to set your custom color scheme for this particular syntax. There is probably a way to programmatically write to a language specific settings file and set the "color_scheme" setting automatically for your users (e.g. via a command that you can document in the README).
Extending color schemes is also something that you can automate, but you should have a good reason to do so. A better reason than that _you_ think it should look a particular way.

KeeMeng:
Thank you for the feedback. I have standardized the scope names!

braver:
Thanks! We're ready to go now. One thing you might want to look at is a [.gitattributes file](https://git-scm.com/docs/gitattributes). This allows you to exclude things like PNG's and other stuff from the package. Such files are useful for a readme, but have no function in the package itself. Excluding them via `.gitattributes` makes the download faster and takes up less space on disk.

PR #8042, opened 2020-09-14
Resolve Javascript Modules

thomaswelter:
This plugin for Sublime Text 3 adds completions for es modules, import file paths and browser api's.

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -1332,7 +1332,7 @@
 				{
 					"sublime_text": ">=3200",
 					"platforms": ["*"],
-					"branch": "master"
+					"tags": true
 				}
 			]
 		},
--- a/repository/r.json
+++ b/repository/r.json
@@ -1324,10 +1324,10 @@
 			]
 		},
 		{
-			"name": "resolve_js_modules",
+			"name": "Resolve Javascript Modules",
 			"author": "Thomas Welter",
-			"details": "https://github.com/thomaswelter/resolve_js_modules",
-			"labels": ["auto-complete", "javascript", "browser", "js", "completions"],
+			"details": "https://github.com/thomaswelter/Resolve-Javascript-Modules",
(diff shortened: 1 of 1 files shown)


rchl:
Your package won't work when packaged as you are using file APIs to access files within your package. The package will be a zip file so those won't work. You need to use `sublime. load_resource()` API to load the contents of those files.

(And don't try to write to your package as you do from your `log` function)

thomaswelter:
Thanks for your comments @rchl I fixed the issues you mentioned. I don't think sublime API's are used before plugin_loaded, I assume `on_query_completions` is only called when the sublime API's are loaded.

A new release 1.0.1 is pushed to the repo with the changes

rchl:
You have this code:
```py
sublime.load_resource("Packages/resolve_js_modules/browser_completions.json")
```
but your package will be called `Resolve Javascript Modules` when published so it needs to be:
```py
sublime.load_resource("Packages/Resolve Javascript Modules/browser_completions.json")
```
(you can rename your repo to match)

thomaswelter:
Ah, I was under the impression that the name field was more of a display name and the name restrictions that the packagecontrol docs are talking about applied to the repo name.

Importing names with spaces seems difficult in python so I updated the name of the package to match the source code.

rchl:
> Importing names with spaces seems difficult in python so I updated the name of the package to match the source code.

The name is what is shown on the Package Control page so having a name like `resolve_js_modules` is rather weird.
What's wrong with using spaces? Why would you need to import your own package anywhere?

thomaswelter:
I agree that the snake case style name is weird. Python is not my main language so I had problems getting the esprima import right. I now learned that python also has relative imports and I updated the code to use that instead. The package seems to work when zipped and renamed to `.sublime-package`

This PR now reflects the new name with spaces.

PR #8043, opened 2020-09-14
Add LSP-robloxlua package

alihsaas:


repository/l.json
```diff
@@ -1717,6 +1717,19 @@
...
 		},
+		{
+			"name": "LSP-luau",
+			"details": "https://github.com/alihsaas/LSP-luau",
+			"labels": [
+				"lsp", "roblox", "luau"
+			],
+			"releases": [
+				{
+					"sublime_text": ">=3092",
```
rchl, line 1728: Your package is only compatible with ST4 since you are using LSP API that is only available there.

rchl:
This is based on https://github.com/sublimelsp/LSP-lua right? How do those two compare?

repository/l.json
alihsaas: This uses the server provided in https://github.com/NightrainsRbx/RobloxLsp which adds support for Roblox's Luau

alihsaas:
> 
> 
> This is based on https://github.com/sublimelsp/LSP-lua right? How do those two compare?

This uses the server provided in https://github.com/NightrainsRbx/RobloxLsp which adds support for Roblox's Luau

rchl:
> > This is based on [sublimelsp/LSP-lua](https://github.com/sublimelsp/LSP-lua) right? How do those two compare?
> 
> This uses the server provided in [NightrainsRbx/RobloxLsp](https://github.com/NightrainsRbx/RobloxLsp) which adds support for Roblox's Luau

I wonder if we should consolidate those somehow because the user will likely not understand the difference between two almost alike-named servers.

alihsaas:
> 
> I wonder if we should consolidate those somehow because the user will likely not understand the difference between two almost alike-named servers.
Would LSP-RobloxLuau work?

rwols:
I think it's okay to have two different lua language servers. I don't program in lua myself, but from what I understand searching on the internet there seem to be many dialects of this language:

- The base "lua" language
- A dialect from "Roblox"
- GLua from Garry's Mod
- EmmyLua for type annotations (?) https://github.com/EmmyLua/EmmyLua-LanguageServer

There's a bit of a problem with respect to the syntaxes/languageIDs: if a user has both the base lua language server as well as the roblox lua language server, then they'll both be attached to the same file. How do people handle this in vscode land?

There's opportunity to improve your repository, @alihsaas. See my patch notes: https://github.com/sublimelsp/LSP-lua/blob/master/patch-notes

alihsaas:
You can read more on Roblox's Luau here https://roblox.github.io/luau/

> 
> There's a bit of a problem with respect to the syntaxes/languageIDs: if a user has both the base lua language server as well as the roblox lua language server, then they'll both be attached to the same file. How do people handle this in vscode land?

I don't know, but I have added a syntax definition file, doesn't include type syntax if that helps.

> 
> There's opportunity to improve your repository, @alihsaas. See my patch notes: https://github.com/sublimelsp/LSP-lua/blob/master/patch-notes

Applied the patch. 👍

rwols:
The amount of LSP-* packages is getting out of hand, so we've made our own package repository. Please submit your pull request there: https://github.com/sublimelsp/repository/

PR #8047, opened 2020-09-17
Bemol Language Syntax

fabricioh:
This package adds support to the Bemol programming language (you can check out the language here: https://github.com/fabricioh/bemol-lang).
There are no other packages that do so.


FichteFoll:
Both repositories as well as the PR repo seem to be gone.

PR #8053, opened 2020-09-24
Add plugin "Special Chars"

YanDatsiuk:
Added plugin that escapes special characters.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -3058,6 +3058,16 @@
 			]
 		},
 		{
+			"name": "Special Chars",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Your definition of "special characters" is very domain-specific and only applies to XML and SGML languages like HTML.

What your package does is html-escape problematic characters (with syntactic meaning) with so-called HTML entities, which [we have a couple packages for already](https://packagecontrol.io/search/entity).

YanDatsiuk:
Thanks for the review. I will close the pull request then. If I enhance a functionality or come up with another idea I will do a new one. Thanks again and have a nice day!

PR #8055, opened 2020-09-27
Update Tutkain ST version requirement

eerohele:

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -3227,7 +3227,7 @@
 			"description": "An interactive Clojure editing environment for Sublime Text",
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": ">=4000",
 					"tags": true
 				}
 			]
--- a/repository/t.json
+++ b/repository/t.json
@@ -3222,9 +3222,9 @@
 		},
 		{
 			"name": "Tutkain",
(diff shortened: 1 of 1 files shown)


FichteFoll:
What about your existing and potentially future ST3 users? Does 0.5.3 still support ST3? Because if it does not, then people will already have updated to it.

You can provide an older version of your package for users on older ST builds. There are plenty of references in the default channel for that, but also in the [eample repository file](https://github.com/wbond/package_control/blob/master/example-repository.json).

eerohele:
>What about your existing and potentially future ST3 users? Does 0.5.3 still support ST3? Because if it does not, then people will already have updated to it.

Package Control shows 5 installs in total for the package, and I'm pretty sure they're all me, so there really are no existing users. :P

0.5.3 does not support ST3. I also have no intention of supporting ST3 going forward.

>You can provide an older version of your package for users on older ST builds.

That's good to know, but I don't think it's worth it in this case.

FichteFoll:
> Package Control shows 5 installs in total for the package, and I'm pretty sure they're all me, so there really are no existing users. :P

Well, I don't check package numbers, but 4 of those 5 installs were indeed in the last 3 days, so that's fine with me.

PR #8060, opened 2020-10-02
Added earthly earthfile syntax

vishnugt:
This is a syntax definition for Earthly - https://github.com/earthly/earthly

Earthly is a build automation tool that allows us to execute all our builds in containers. Earthly can be used to create Docker images and artifacts (eg binaries, packages, arbitrary files).

There are no other existing packages for providing syntax highlighting to Earthly

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -34,6 +34,18 @@
 			]
 		},
 		{
+			"name": "Earthly Earthfile syntax",
+			"details": "https://github.com/earthly/sublimetext-earthly-syntax",
+			"readme": "https://raw.githubusercontent.com/earthly/sublimetext-earthly-syntax/main/README.md",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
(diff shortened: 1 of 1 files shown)


braver:
- Since ST2 is essentially dead and there are only like 5 people still out there using it, you may want to provide your syntax as a [.sublime-syntax](https://www.sublimetext.com/docs/3/syntax.html#overview) file. It's much easier to maintain and read, it has more features, and it has enhanced performance. There is a converter built into ST3 (IIRC via Tools > Developer and/or in the command palette).
- You don't need to specify the README location if it's simply in the root of the main branch of your repository and has a "normal" filename.

braver:
@vishnugt please remove the README key. 

Additionally, please consider removing "syntax" from the name. Usually packages that supply syntax and related features simply have the name of the language they support. That it's (also) the syntax is implied.

vishnugt:
Removed readme key and removed "syntax" from the name @braver

braver:
🎉

PR #8063, opened 2020-10-04
Added Visual Studio Code Plus Scheme

vbasky:
# Visual Studio Code Plus Scheme 1.0.0

- First version of colour scheme
- Supports PHP, HTML/Blade, CSS/SASS/LESS, JavaScript (including Vue components), JSON, YML, Markdown and Python

### DIFF
--- a/repository/v.json
+++ b/repository/v.json
@@ -564,6 +564,18 @@
 			]
 		},
 		{
+			"name": "Visual Studio Code Plus Scheme",
+			"details": "https://github.com/vbasky/sublime-vscode-plus",
+			"issues": "https://github.com/vbasky/sublime-vscode-plus/issues",
+			"labels": ["color scheme"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Visual Studio Dark",
 			"details": "https://github.com/nikeee/visual-studio-dark",
 			"labels": ["color scheme"],


rwols:
I noticed you wrote these two color schemes in a .tmTheme format. Why not use the newer and much easier to read .sublime-color-scheme format?

vbasky:
I just ported the theme for the VSCode's native format which is json based to .tmTheme. Have to explore the Sublime Text color scheme not proficient with .sublime-color-scheme format

rwols:
The .sublime-color-scheme format is also json, so that should be a piece of cake.

vbasky:
Since there is no easy way to convert can I change to sublime Color Scheme in subsequent dot release ?

vbasky:
Done modified the .tmTheme to .sublime-color-scheme and tested it

vbasky:
Is there anything else that I need to do on this PR?

rwols:
> Is there anything else that I need to do on this PR?

In terms of this being mergeable, I don't think so. Your PR will be accepted. Just have patience :) this is all done on a voluntary basis.

If you allow me to comment more on your package: there are a lot of scopes targeted that don't apply to common ST syntaxes. For instance, scopes ending in `.cpp` do not target the default C++ syntax, because those end in `.c++`. There's also no `keyword.operator.new` scope in the default syntaxes that I know of. I suppose this is another VSCode specific scope. I'm also missing a general `punctuation.section` and `punctuation.definition` scope.

We're still in the vscode scope space while there should be a transformation to the sublime scope space. http://www.sublimetext.com/docs/3/scope_naming.html

jwortmann:
> There's also no keyword.operator.new scope in the default syntaxes that I know of.

This scope is used in the default syntaxes by C# and there are various 3rd-party syntaxes where it occurs, e.g. the TypeScript, "Babel" (JavaScript) and "Better CoffeScript" packages.

However, besides the `.cpp` you mentioned there are indeed a few other scopes like `header` or `token.error-token` which I haven't seen before and which probably don't apply to Sublime Text. I wonder where they are used by VS Code or by which extensions.

vbasky:
Can the moderator please merge this in?

braver:
Sorry, we're kinda short on "staff". The "issues" property is superfluous here, pointing to the GitHub repo for "details" is sufficient. Please remove it and I'll hit merge.

vbasky:
Done

vbasky:
Please merge the PR if you deem it fit

braver:
Sorry for the delayed response here, but I’ve been busy _not_ being behind a laptop during the holidays. 

So, essentially this looks good. I’d like to consider the name though (naming things 😩). There is already a “Visual studio code” package that does something else. The “plus” here kinda seems to say that it does that _plus_ something else, but it doesn’t: it’s just a color scheme. Is “Visual Studio Code Plus” the best name for this package, or could it maybe be “VS Code Plus Scheme” or something?

vbasky:
Modified the package name

PR #8066, opened 2020-10-05
Add Oceanic Primal Color Scheme

barlog-m:
Oceanic Primal Color Scheme.
A minimalistic color scheme based on [Oceanic Primal palette](https://github.com/oceanic-primal/palette), inspired by [Oceanic Next Color Scheme](https://github.com/voronianski/oceanic-next-color-scheme).

### DIFF
--- a/repository/o.json
+++ b/repository/o.json
@@ -120,6 +120,17 @@
 			]
 		},
 		{
+			"name": "Oceanic Primal Color Scheme",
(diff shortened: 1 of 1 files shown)


braver:
Thanks! You sure you don't want to convert that tmTheme to a sublime-color-scheme? That file format, JSON, is so much easier to work with (e.g. it support variables). Doesn't matter for your submission here, but something to consider for future releases.

PR #8067, opened 2020-10-05
Add Incremental Diff plugin

AmjadHD:
A plugin to quickly diff two views through Sublime text's built-in Incremental Diff.


braver:
Thanks for submitting!

PR #8073, opened 2020-10-15
Added Package MenuPreferences

Penguin98kStudio:
I created this package for making quick panel menus for few features,
1: Temporary Preference Toggling from a sorted (through usage frequency), modifiable list for the current view. I found no such things in the packages lists.
2: I made Quick Panels for File Commands, and other features because as a An Active User you would prefer to find these commands quickly in a single list and sorted through frequency usage.

Please see the readme.md file or contact me for any details. This might be my first published plugin but I have made a lot of them for self use.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -1257,6 +1257,18 @@
 			]
 		},
 		{
+			"name": "MenuPreferences (Menu By Penguin)",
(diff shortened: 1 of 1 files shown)


braver:
Thanks for contributing a package. It's marked as WIP, do you want us to still review it?

PR #8078, opened 2020-10-20
Add GLanguage

JoseCarlosSkar:

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -922,6 +922,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
+					"tags": false,
 					"branch": "master"
 				}
 			]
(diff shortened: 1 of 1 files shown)


rchl:
New packages are required to use tag-based releases (`"tags": true`)

rchl:
And there is a bug with handling repos with `main` branch (rather than `master`). But you shouldn't use branch-based releases anyway.

repository/g.json
```diff
@@ -922,7 +922,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"tags": false,
+					"tags": true,
 					"branch": "master"
```
rchl, line 926: Remove `branch` key also.

PR #8081, opened 2020-10-22
Add screeninfo

ggets:
Add sublime package dependency: add the [https://github.com/rr-/screeninfo](https://github.com/rr-/screeninfo) package as [https://github.com/ggets/sublime-screeninfo](https://github.com/ggets/sublime-screeninfo)

### DIFF
--- a/repository/dependencies.json
+++ b/repository/dependencies.json
@@ -1032,6 +1032,21 @@
 			]
 		},
 		{
+			"name": "screeninfo",
+			"load_order": "99",
+			"description": "Fetch location and size of physical screens - https://github.com/rr-/screeninfo",
+			"author": "ggets",
+			"issues": "https://github.com/ggets/sublime-screeninfo/issues",
+			"releases": [
+				{
+					"base": "https://github.com/ggets/sublime-screeninfo",
+					"tags": true,
+					"sublime_text": "*",
+					"platforms": ["*"]
+				}
+			]
+		},
+		{
 			"name": "select-windows",
 			"load_order": "02",
 			"description": "Python select module for Sublime Text 2 on Windows",


rchl:
I see the code does `import typing` so that would only work with python 3.5+. The dependency runs in python 3.3 environment.

ggets:
> 
> I see the code does `import typing` so that would only work with python 3.5+. The dependency runs in python 3.3 environment.

You're right. I didn't notice that. I'll have try to find a workaround. I see there is a backport. I will try to make use of that.

ggets:
It looks like it's already been thought of [typing module for Package Control](https://github.com/packagecontrol/typing)

rchl:
Oh, didn't know about it.

So I guess it might be OK if you state in your dependency that it has to be required together with the `typing` dependency.

But you'll have to wait for @FichteFoll to confirm that anyway.

ggets:
Well, I tried it and it loaded successfully. Now trying to solve the **dataclass** library problem.

ggets:
Finally got a working version for review.

rchl:
It's not possible for dependencies to depend on other dependencies. The only thing you can do is to state in the readme which dependencies are required when someone wants to use your dependency.

ggets:
> 
> 
> It's not possible for dependencies to depend on other dependencies. The only thing you can do is to state in the readme which dependencies are required when someone wants to use your dependency.

Sorry about that. Fixed.

braver:
Alternatively, if the features you need are supported in later python versions, you could consider releasing for Sublime Text 4. This is currently in beta, so you might not reach a lot of people for a while. But it _will_ allow you to run on [python 3.8](https://github.com/braver/SideBarTools/blob/master/.python-version).

ggets:
> 
> 
> Alternatively, if the features you need are supported in later python versions, you could consider releasing for Sublime Text 4. This is currently in beta, so you might not reach a lot of people for a while. But it _will_ allow you to run on [python 3.8](https://github.com/braver/SideBarTools/blob/master/.python-version).

I have already fixed this by replacing the missing features.

braver:
Let's handle this as part of #8110 and prevent conflicts etc.

PR #8082, opened 2020-10-24
add SynthWave 84 color scheme

lucasvscn:
I like this VS Code color scheme and so I did a Sublime Text version of it.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -5794,7 +5794,7 @@
 			]
 		},
 		{
-			"name": "SyntWave 84 - Colorscheme",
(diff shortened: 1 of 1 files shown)


braver:
You may want to consider converting the color scheme to sublime-color-scheme format. It's JSON, supports variables, and is so much easier to work with. 

Additionally, you need to tag a release with a proper [semver](https://semver.org) version number.

lucasvscn:
As you've requested, I'd added the tags to the repository. I'll consider using sublime-color-scheme on next release. Thank you!

PR #8083, opened 2020-10-24
Add Fmt package

mitranim:
Link to repo: https://github.com/mitranim/sublime-fmt.

Motive (citing from the readme):

Why this exists?

Package Control has special-case formatter plugins for different languages, and the monstrous Formatter with too many batteries included. This makes it hard to add formatters: someone has to make and publish a new plugin every time, or fork a repo and make a PR, etc.

Many formatters just call a subprocess and use stdio. One plugin can handle them all, while letting the _user_ specify any new formatter for any new syntax! This works for `gofmt`, `rustfmt`, `clang-format`, and endless others.

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -1283,6 +1283,16 @@
 			]
 		},
 		{
+			"name": "Fmt",
+			"details": "https://github.com/mitranim/sublime-fmt",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Fn",
 			"details": "https://github.com/jonnyarnold/fn-syntax",
 			"labels": ["fn", "language syntax"],


mitranim:
The bot complains about a missing "license" file. The repo has an "unlicense" file, which is recognized by GitHub and some other services. I could rename it, or we could teach it to the bot. Suggestions?

braver:
Sounds like a really great tool. 

> To avoid potential conflicts, this plugin does not come with hotkeys.

❤️ ❤️ ❤️ ❤️ 

> The bot complains about a missing "license" file.

I think the quickest thing to do here is just to comply with the bot and rename the file to LICENSE.

braver:
Oi, it's picky isn't it, it wants to see an all-caps LICENSE. Well, you can take care of that as you see fit, it doesn't really matter for this PR.

PR #8085, opened 2020-10-25
Fix main branch rename for relikd repos

relikd:
Updated main branch, this PR uses tags instead


braver:
@relikd we require ["semver" semantic versions](http://semver.org) which consist of exactly three numeric parts. Please create new tags (e.g. "v1.0.0").

relikd:
done.

braver:
🎉

PR #8087, opened 2020-10-30
OrgExtended - OrgMode for Sublime Text

ihdavids:
OrgExtended is a much more extensive orgmode than the original orgmode package. It is not finished, but there is a significant body of work here and I think it is fairly useful in its current state.

It adds:
- Full orgmode syntax highlighting.
- Inline image visualization
- Table support through a dependency on Table Edit - I may consume this at some future data, I really want expression support
   in tables and TableEdit while an awesome package is deprecated by the author and thus may be a challenge to extend.
- html and pandoc export functionality
- reveal js export functionality
- Integration with NeoVintageous for those of us that use it.
- TODO cycling
- automatic numbered lists and checkbox management
- property drawers and log drawers
- clocking
- customizable dynamic blocks
- tags and tag filtering
- custom ids
- links both internal and external
- A dynamic agenda with somewhat flexible customizable views
- Extensive navigation and editing commands
- Calendar / scheduling selection widget
- OS / System Notification thread and popups.
- Babel style code execution of code blocks (although this is highly limited and parameter / chaining support is not there yet) 
- Capture templates using snippets and dynamic targetting.
- Refiling support - for GTD style list maintenance.
- Arching support for maintaining task list cleanliness.
- light and dark color schemes

For a really really incomplete and bad set of docs with some illustrations see:
https://github.com/ihdavids/orgextended/blob/master/docs/start.org

### DIFF
--- a/repository/o.json
+++ b/repository/o.json
@@ -946,6 +946,19 @@
 			]
 		},
 		{
+			"name": "OrgExtended",
+			"details": "https://github.com/ihdavids/orgextended",
+			"author": "Ian Davids",
+			"readme": "https://github.com/ihdavids/orgextended/blob/master/docs/start.org",
+			"labels": ["org mode", "notes", "note taking", "authoring", "todo", "tasks"],
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"tags": true
+				}
+			]
+		},
+		{
 			"details": "https://github.com/SublimeText/Origami",
 			"releases": [
 				{


ihdavids:
This has been open for a long time!
How do I get someone to review and commit this for me?

ihdavids:
NOTE the warnings are expected. I redefine tab to "fall through" and do special things depending on context in org files. This is intentional since that is how org traditionally handles folding.

I also use a mousemap to allow following links in an org file with a click. These things seem to trigger warnings but are desired in this format.

PLEASE, can I get a review?

braver:
So, first of all thanks for submitting this package! Looks like a lot of time and effort went into it. Sorry, we've been behind on reviews. Because this is a bit of a large package to review I put it somewhat lower on my todo list. If we're going to do this I want to do it properly. So, take all of this as us taking the opportunity to help the package be the best it can be before we really send it out into the world. 

I might not get through all of this in one go, but here is the first pass. A mix of remarks, and tips, but also a few things that should probably change. I'll number them so you can go through it one by one.

1. The keybinding warnings aren't a problem to me. It's expected for a package like this and you seem to be aware of the challenges here and I trust it's handled well. So all good there.
2. Have you considered/discussed presenting this as an upgrade to the existing orgmode package?
3. You might want to use the .gitattributes to exclude some non-functional files from the package to keep download and storage size down.
4. You don't need to have a packages.json file in your package, or in your repo probably.
5. I'm not sure your usage of git-lfs will work properly, and usually you don't want any large files in a package like this.
6. Make sure your debug logger is off by default.
7. The captions of your commands are CamelCased. It's much more readable if you us spaces etc. Eg. "OrgInsertHeadingSibling" could be "Org Mode: Insert heading sibling". 
8. Similarly "orgextended", "org agenda" etc. names for the syntaxes could be more readable.
9. All your commands seem to be active in all views, but it seems to me that they only have a purpose inside views with the orgextended syntax. 
10. The way you set up the syntaxes and color schemes are a problem. Because all of your scopes are non-standard, you need to use the color schemes you provide. People will expect the org-mode views to be consistent with other views in the editor, and at the very least to have a choice in this matter. You could have a look at packages like GitSavvy that introduce "custom" views with highlighting that plays nice with user's color schemes. 
11. Scopes in your syntaxes, where they are mostly just for highlighting the .org file, should end with something like `.orgmode`. This makes it easier for color schemes to target specific situations in org mode files.

ihdavids:
Wow thank you for reviewing this.
Very exciting! I will respond inline here and then go work to implement these.

I had lost hope that this would ever have the opportunity to become a package.
This thing is definitely an ongoing project that I am actively extending.

1. I am glad to hear that the keybindings warnings will not be an issue. I was worried about that. Tab is the standard for orgmode and it would be annoying to have to dump that. 
2. No, I suppose I could approach the old author and see about replacing his package. OrgExtended is a massive diversion from his package. It uses some stuff inspired from his checkbox support, but it throws out a lot of what he wrote for somewhat more orgmode like items. At this point it would be effectively nuking his repo and replacing what he did with my work. I will however reach out to him and see what he says.
3. For the .gitattributes stuff. Sure, when I wrote the package I didn't think to much about the fact that most sublime packages are zipped. I actually have a link in the package settings that will open up the org docs in sublime as a self documenting bit. I guess that wont work with a zipped package.
4. Okay, I will nuke packages.json.
5. Okay, what would you suggest I do here?
6. Oh sure, yeah, I have gotten used to that being on all the time for my purposes. I will turn it off. Thanks for noticing that.
7. Fair enough, I will swap those for spaces.
8. I am not sure what you mean here? Names for the syntaxs?
9. Fair enough I will filter the ones out that don't make sense outside of org. That is just me being lazy. Some of them do have purpose outside of an org file. Such as capture etc. But you are right most of them are just me being lazy here. I honestly didn't even think to filter them out.
10.  Yes, I am aware my syntaxes and themes are very custom. I will take a look at GitSaavy but I do play some games here where I am depending on specific behaviour (such as foreground color and background color being the same color or certain magical sequences being seen as colored blocks) I might be able to make the orgextended syntax behave more like a standard syntax (although I would like to request that don't do that for version 1 as it represents quite a bit of work) but I expect that views like the agenda view will probably always use a custom color scheme.
11. I am not sure I understand this one. All of the custom scopes start with orgmode.XXXX. Can you help me understand?

braver:
> 1. I am glad to hear that the keybindings warnings will not be an issue.

We advise against keybindings usually, but some packages just need them. As long as they are scoped to a very specific context it's fine.
I do worry about the mousemap though. If they cannot be set to a specific context, I would advise to remove it and instead explain in the README how to add it. Besides, how important can mouse interaction be in something that emulates an Emacs feature? 😉 

> 2. ... I will however reach out to him and see what he says.

I can imagine how this could get really complicated, because then you'd also have to consider compatibility for existing users. So, ideally we have one really good actively maintained org mode package. But it's just as valid to have two different packages, each with maybe a different approach. But I think a merge of the two should at least be considered.

> 5. Okay, what would you suggest I do here?

Actually I don't know. Maybe we'll just have to see. But I can't imagine you really have so many image files in your repo that it demands using git-lfs? 

> 8. I am not sure what you mean here? Names for the syntaxs?

Yes, these:
Screenshot 2021-01-13 at 10 07 31

The all-lower-case names seem like they could use a little attention, e.g. "Org Agenda". 

> 10. Yes, I am aware my syntaxes and themes are very custom.

In the end it's your call, but our experience is that color schemes are a very personal choice. In taste, but also usability (dark vs. light) and accessibility (contrast etc).  You will have trouble finding a large audience for your package if it demands a very specific color scheme. It will probably be the primary thing people complain about and for many it will be the primary blocker against using the package: regardless of the functionality, color schemes are very important.

Some mores thoughts:

You ship several specific color schemes for each particular syntax. It would be easier for you to maintain the color schemes if there was only one. I don't see why that would not be possible technically since you control all the scopes anyway. 
If you ship a single color scheme, users would more easily be able to [override/extend that color scheme](https://www.sublimetext.com/docs/3/color_schemes.html#customization). 

Your color scheme is very different from all of the color schemes that ship with Sublime Text. You may want to consider at least basing the colors off a default and widely used color scheme like Mariana. All the default color schemes come with a set of variables for their colors, which can be easily adapted to your use cases.

You may want to consider generating a color scheme based on the user's preferred color scheme. This could be a color scheme that extends the user's preferred color scheme, but adds scopes of your own. I can see how that becomes tricky though. 

At the very least you should make it easy for someone to create a color scheme or org mode that matches what they want. E.g. maybe you could make "Monokai OrgExtended", "Mariana OrgExtended" and a handful others, and allow users to pick one via your package settings. Or maybe "Monokai Extended" should be able to cover your scopes to make it work.

> 11.  All of the custom scopes start with orgmode.XXXX.

If the syntax is never intended to be targeted by anything else, then that's fine. But if you would make a more standard syntax, `.orgmode` should go at the end of the scope.


---

Other things.

12. Your package settings are still split into two entries, the old "Default" and "User". You can now specify a single menu entry that opens both sides of the settings in a split view. Take a look at [SublimeLinter for an example](https://github.com/SublimeLinter/SublimeLinter/blob/master/Main.sublime-menu#L19).
13. What are all the html and css files for?
14. Have you tested if everything still works when your package is delivered in an archived/zipped format? You can't read and write to your package as you would if it's simply sitting in the Packages directory. You can test this behaviour by [creating the package archive yourself](https://packagecontrol.io/docs/creating_package_files).

ihdavids:
For color schemes. I will work on that but I would prefer not to do that in version 1.0 as that is a fairly large change. If you don't mind I would like to do that in a subsequent version.

I have been talking to the original orgmode author and he seems amenable to merging our plugins. That should work well.
That said, your feedback is valuable to me and I am working (as I can make time) to incorporate all of it. (even if I am going to just merge with his plugin and bypass the need for this review)

Hmmm, yes I should test the zip! I did a while back and things worked, but I have added a ton since then so I really need to run a test. Thanks for reminding me of that.

The html and css files are for exporting. Org mode is a ton of things. One of the things it is capable of is exporting to html, exporting to a reveal.js enabled presentation etc ad nauseum. I have my own html exporter that I wrote for my own purposes since I was struggling getting pandoc to play nice for what I wanted to do. You can make your own folder in the User package and add your own templates / style to export using your own style. My embedded styles may not work in the zip... So I really need to test that.

Oh, I like the split settings stuff. I really need to incorporate that, it's much nicer than my approach.

I have moved the docs out to their own repo (they are... org files of course). That also moves the animated gif's in those docs, so I don't need the lfs support anymore.

> You might want to use the .gitattributes to exclude some non-functional files from the package to keep download and storage size down.

You can exclude the files from the package but not from your repo? I very much want to do that with some test files and my todo lists. How does that work?

Thank you so much for your feedback so far. I am and will be taking everything you say seriously.

braver:
> I might do what you suggest and take the users chosen style and augment it somehow.

I can see how what you're doing it a bit more tricky than usual. There are actually quite a number of examples out there where the selected color scheme is augmented. For foreground colors you can use scope names like `region.yellowish`: [see the 3.1 release notes](https://www.sublimetext.com/3) and you can of course do region highlighting in several ways.

> I have moved the docs out to their own repo

👍🏻 That works :)  Here's an example of how [.gitattributes work](https://github.com/braver/FileIcons/blob/master/.gitattributes).

Thanks for getting back here and explaining stuff. I'm definitely not against merging now and letting you get on with improving the package later. In any case you're definitely taking this seriously :) 
I'll review some things again later this weekend. For now, I'd like to point you to the [Discord server](https://discord.gg/D43Pecu). Lot's of really experienced Sublime Text package developers hang out there and are better at explaining the nitty gritty for lot's of stuff than I am.

braver:
Your schedule / calendar implementation seems to need a few guards :) 
```
  File "/Users/koenlageveen/Library/Application Support/Sublime Text/Packages/OrgExtended/orgdatepicker.py", line 230, in on_changed
    self.dateView.cdate = OrgDateFreeFloating.from_str(text)
  File "/Users/koenlageveen/Library/Application Support/Sublime Text/Packages/OrgExtended/orgparse/date.py", line 567, in from_str
    rv = cls(start, end, active=cls._active_default)
  File "/Users/koenlageveen/Library/Application Support/Sublime Text/Packages/OrgExtended/orgparse/date.py", line 192, in __init__
    self._start = self._to_date(start)
  File "/Users/koenlageveen/Library/Application Support/Sublime Text/Packages/OrgExtended/orgparse/date.py", line 242, in _to_date
    return datetime.datetime(*date)
ValueError: day is out of range for month
  File "/Users/koenlageveen/Library/Application Support/Sublime Text/Packages/OrgExtended/orgdatepicker.py", line 233, in on_changed
    self.dateView.ReShow()
  File "/Users/koenlageveen/Library/Application Support/Sublime Text/Packages/OrgExtended/orgdatepicker.py", line 140, in ReShow
    mid = (now.month - self.midmonth + 1)
AttributeError: 'NoneType' object has no attribute 'month'
```

One final thing that I would change is the name of the (user) settings file. Right now it's lowercase "orgextended", whereas for the end user your package is always named camelcase "OrgExtended". Small stuff, but it's the finishing touches, and something that's harder to change after shipping. 

I guess the best thing to do now is to expand your user base and get some feedback. This PR here looks good, so let's ship it. Just to be sure, let me know if you're really ready for me to hit that "merge" button 👍🏻

ihdavids:
- I think I have a guard in for that exception now. 
- I have renamed the settings file case.
- The mousemap is gone but I have a button to regenerate it.
- I have done a bunch of testing with a zipfile and things all work again. (I did find a few issues)
- I have posted a new release tag of 1.0.2 with all these changes.
- The user folder settings file now needs to be renamed to OrgExtended.sublime-settings.
- NOTE: I kept the case for my syntaxes for orgagenda and the orgdatepicker to be lowercase since those are syntaxes that you shouldn't use yourself, they are used in special windows. The convention makes sense to me and I plan to document it in my docs. I plan to add more "apps" like column select mode etc. All emacs orgmode style features that are really handy. I think having a convention like this will enhance clarity.
- I have added messages for the releases I have done so far.
- I have added spaces to commands et al.
- The logging is off..
- I cleaned up a bunch of prints.


- Sooooo... I think we are good to hit the merge button if you are okay with me doing the syntax stuff you mentioned in a version 1.1.0 at a later date.

ihdavids:
Thanks for pointing me at the Discord Server.
 
I am on there. I am currently using ST4 as I love the relative line number mode with neovintageous and things are generally stable enough with the quick reloading of ST4. I have asked questions up there in the past. (I am the annoying invader zim icon)

braver:
Cool, good job! Let’s ship it! Have fun maintaining and improving your package, it’s really ambitious and it looks like it could make a lot of people happy 👍🏻

PR #8090, opened 2020-11-01
Add TLA+ package

Calvin-L:
This package provides syntax highlighting for the formal modeling language [TLA+](https://lamport.azurewebsites.net/tla/tla.html). I have named the package "TLAPlus" to avoid using a "+" character in a path name.

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -2372,7 +2372,7 @@
 			"labels": ["language syntax"],
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": ">=3092",
 					"tags": true
 				}
 			]
--- a/repository/t.json
+++ b/repository/t.json
@@ -2367,6 +2367,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


braver:
Thank you for submitting a package! Please add a `.gitattributes` file to exclude the test file from the package.

Calvin-L:
Sure thing!

Although... [`.gitattributes`](https://www.git-scm.com/docs/gitattributes) has a lot of capabilities, none of which seem directly related to packaging. Is there documentation about which attributes Package Control looks at? I couldn't find any mention of `.gitattributes` on the Package Control website.

rwols:
You would exclude your syntax test file so that people who install your package through Package Control do not have any extra space wasted. [Example](https://github.com/sublimelsp/LSP/blob/948989cd7020b2ce265478586881233c8104b0a6/.gitattributes#L7)

You can test your changes by running [git archive](https://git-scm.com/docs/git-archive).

PR #8091, opened 2020-11-02
Add BTstack HCI Packet Log Converter

mringwal:
Hi. 

The the package allows to quickly convert a Bluetooth packet log from from an embedded device into a binary representation that can be analyzed with a tool like Wirehshark. 
There's no similar tool in the repo as far as I know as it's a rather niche application. It's helpful though :)

Thanks for consideration,
 Matthias Ringwald, BlueKitchen GmbH

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -1496,6 +1496,17 @@
 			]
 		},
 		{
+			"name": "BTstack HCI Packet Log Converter",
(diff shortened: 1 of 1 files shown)


braver:
Nice, sounds useful indeed! 🎉

mringwal:
Thanks for adding and especially for providing package control all these years.

PR #8092, opened 2020-11-06
Add "Run Behat in Sublime Text"

bfolliot:

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -34,6 +34,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


braver:
- You have a failing check. Please trigger the checks again.
- Please remove "Sublime Text" from the package name, all the packages here are for Sublime Text ;) "Run Behat" is just as descriptive (even more so because it's shorter).

braver:
@bfolliot Please remove "Sublime Text" from the package name. A push will also kick off the review again.

PR #8093, opened 2020-11-07
Add terraform core plugin

marcincuber:
A useful Terraform plugin which supports latest syntax and provides up-to-date snippets

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -637,6 +637,18 @@
 			]
 		},
 		{
+			"name": "Terraform Core",
+			"labels": ["snippets", "terraform", "formatting"],
+			"details": "https://github.com/marcincuber/terraform-core",
+			"author": "Marcin Cuber",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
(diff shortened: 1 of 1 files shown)


rchl:
This looks like a close copy of https://packagecontrol.io/packages/Terraform. The syntax file is exactly the same (but you are missing the fallback for old ST versions).

What made you create a new package rather than improving the existing one? Is the author unresponsive?

It's not a good idea to duplicate packages as it just creates confusion for users and makes it hard to choose the right package.

rchl:
> A useful Terraform plugin which supports latest syntax

BTW. That statement is not true. The syntax you have here (and which is the exact copy of the one from the other package) is not complete and breaks on various constructs.

marcincuber:
Syntax has been reused and will be further expanded to build automated completions for various parts of Terraform. 

In such case, closing PR for now.

rchl:
Consider contributing to an existing package itself. It seems like it's even backed up but Hashicorp (saw commits from core member there). So it should be fairly well supported and active.

PR #8095, opened 2020-11-10
Add enaml syntax highlighter

MatthieuDartiailh:
This adds syntax highlighting for the enaml DSL (https://github.com/nucleic/enaml). Syntax highlighting has been distributed with enaml for several years but a recent request for inclusion in Package control motivated this PR (https://github.com/nucleic/enaml/issues/426).

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -922,6 +922,16 @@
 			]
 		},
 		{
+			"name": "Enaml Syntax Highlighter",
+			"details": "https://github.com/nucleic/sublimetext-enaml-syntax",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Energy Color Scheme",
 			"details": "https://github.com/sandralundgren/Energy-Sublime-Text-theme",
 			"labels": ["theme", "color scheme"],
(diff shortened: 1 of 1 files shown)


braver:
Thanks for submitting a package. You may want to consider converting the tmLanguage to sublime-syntax. It's a yaml based format that is so much easier to work with, more powerful, and even provides a performance benefit. A converter is built into Sublime Text. Something consider for future releases :)

braver:
Oh, please add the "language syntax" label to make it easier to find.

MatthieuDartiailh:
I added the requested label. Regarding the conversion I tried to use the built-in tool but it does not produce any output nor any error message. If you can point me to resources regarding how to proceed I will give it another try.

braver:
You should have a look at the console. The conversion fails because your syntax has some includes that it can't find. E.g. [entity_name_class](https://github.com/nucleic/sublime-enaml-syntax/blob/main/Enaml.tmLanguage#L72).

braver:
Oh, if you do the conversion we need to update the version requirement to ">=3092". A small technicality because there isn't anyone out there using a version older than that, but for correctness sake. Let me know if you want to do that first, or we can simply merge this PR now and then get back to that later.

MatthieuDartiailh:
Since I did not write that parser and I am not familiar with tmLanguage I suggest we merge as is and I will update the version requirement if I switch to the new format.

braver:
Excellent 👌🏻

MatthieuDartiailh:
Thanks !

PR #8097, opened 2020-11-11
Add Guna Widget

poucotm:

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -2243,6 +2243,17 @@
 			]
 		},
 		{
+			"name": "Guna Widget",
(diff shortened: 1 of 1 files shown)


braver:
Because you're shipping closed source, I can't really review the contents of your package. However, on testing it, it does a number of things in "user space" that are a bit unexpected and not documented, like writing to the Preferences.sublime-settings. It's usually a good idea to explain what's going on if your package goes a bit outside the usual boundaries.

Also, how is this different from the existing [Guna](https://packagecontrol.io/packages/Guna) package? Is it just the widget, without the theme?

PR #8098, opened 2020-11-11
Add SublimeLinter-gofumports linter

pztrn:


braver:
SublimeLinter has its own package channel repository, please open a PR there instead:
https://github.com/SublimeLinter/package_control_channel

FYI, you need to make the addition there to the contrib.json file, and add "-contrib-" in the name of your package.

PR #8103, opened 2020-11-15
 Add AutoPEP8 package

thep0y:

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -2311,11 +2311,10 @@
 			"details": "https://github.com/thep0y/SublimeAutoPEP8",
 			"labels": ["linting", "formatting"],
 			"readme": "https://raw.githubusercontent.com/thep0y/SublimeAutoPEP8/master/Readme.md",
-			"issues": "https://github.com/thep0y/SublimeAutoPEP8/issues",
 			"releases": [
 				{
 					"sublime_text": ">=3000",
-					"tags": true
+					"branch": "master"
 				}
 			]
 		},
--- a/repository/a.json
+++ b/repository/a.json
@@ -2307,6 +2307,23 @@
 			]
 		},
 		{
+			"name": "AutoPEP8_2020",
+			"details": "https://github.com/thep0y/SublimeAutoPEP8",
+			"labels": ["linting", "formatting"],
+			"readme": "https://raw.githubusercontent.com/thep0y/SublimeAutoPEP8/master/Readme.md",
+			"issues": "https://github.com/thep0y/SublimeAutoPEP8/issues",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				},
+				{
+					"sublime_text": "<3000",
+					"tags": "st2-"
+				}
+			]
+		},
+		{
 			"name": "AutoPHPDollar",
 			"details": "https://github.com/graarh/sublime-AutoPHPDollar",
 			"releases": [
--- a/repository/a.json
+++ b/repository/a.json
@@ -2307,7 +2307,7 @@
 			]
 		},
 		{
-			"name": "AutoPEP8_2020",
+			"name": "AutoPEP8 2020",
 			"details": "https://github.com/thep0y/SublimeAutoPEP8",
 			"labels": ["linting", "formatting"],
 			"releases": [
--- a/repository/a.json
+++ b/repository/a.json
@@ -2310,7 +2310,6 @@
 			"name": "AutoPEP8_2020",
 			"details": "https://github.com/thep0y/SublimeAutoPEP8",
 			"labels": ["linting", "formatting"],
-			"readme": "https://raw.githubusercontent.com/thep0y/SublimeAutoPEP8/master/Readme.md",
 			"releases": [
 				{
 					"sublime_text": ">=3000",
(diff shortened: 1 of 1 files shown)


repository/a.json
```diff
@@ -2306,6 +2306,18 @@
 				}
 			]
 		},
+		{
+			"name": "AutoPEP8_2020",
```
braver, line 2310: What is the 2020 for here? Also spaces are allowed in names. So why not "AutoPEP8"?

repository/a.json
```diff
@@ -2306,6 +2306,18 @@
...
+			"details": "https://github.com/thep0y/SublimeAutoPEP8",
+			"labels": ["linting", "formatting"],
+			"readme": "https://raw.githubusercontent.com/thep0y/SublimeAutoPEP8/master/Readme.md",
```
braver, line 2313: Since your "details" point to a GitHub repo, and the readme has a normal name and is in the root of the repo, the "readme" key not need.

repository/a.json
```diff
@@ -2306,6 +2306,18 @@
...
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
```
braver, line 2317: I noticed by the way that your code has a constant for the version number, but it is behind vs. the latest tag.

braver (changes requested):

braver:
I notice that you check against the current syntax, probably so you know you're looking at python code. The recommended way to check what language you're in, is to check the [scope](https://www.sublimetext.com/docs/3/scope_naming.html), python being `source.python`, using a selector check. This reliably tells you what kind of code you're looking at, whereas retrieving the "syntax" only tells you the filename of the syntax package. For instance, if someone is using an alternative syntax package, the filename need not be "Python.tmLanguage", but the root scope will still be `source.python`.

braver:
Please respond to the feedback in this PR if you want to still add your package to package control.

repository/a.json
thep0y: Because there is already a plug-in called `AutoPEP8` in the repository, and this `AutoPEP8` has stopped updating, I just updated it.

repository/a.json
thep0y: I will delete the "readme" key.

repository/a.json
thep0y: I will fix it.

braver:
> Because there is already a plug-in called AutoPEP8 in the repository, and this AutoPEP8 has stopped updating, I just updated it.

Oh, I totally overlooked the existing package. In that case we have 2 options.

1. you make a PR against that package with your update. 
2. if that repo is really dead we can look at replacing it with yours.

thep0y:
> > Because there is already a plug-in called AutoPEP8 in the repository, and this AutoPEP8 has stopped updating, I just updated it.
> 
> Oh, I totally overlooked the existing package. In that case we have 2 options.
> 
> 1. you make a PR against that package with your update.
> 2. if that repo is really dead we can look at replacing it with yours.

I update this package just to provide newer formatting functions for python users who use sublime text, so I agree whether to update the old package or replace the old one.

braver:
@wistful we have someone, @thep0y , who wants to contribute to your AutoPep8 package for Sublime Text. There is a pull request in your repo for this.

Could you please do one of the following:
- Participate in merging that PR and release an update
- Invite @thep0y to help maintain the package in your repo
- Or let us know if it’s ok if we replace your package with the new and updated fork

wistful:
Hi Koen,

Thank you for taking an interest in the AutoPep8 package.
The package is a wrapper around `autopep8.py` and mostly requires only upgrading autopep8.py and its dependencies.
The new version 2.3.0 was just released. I'm planning to upgrade dependencies more frequently.
Also, I'm open to PRs with new features and bug fixes.
Contributions and help are very welcome.
I don't see a good reason to transfer the ownership to another person.

Thank you,
Mikhail

wistful:
It looks like [AutoPep8](Auto​PEP8) is the only active package dedicated to python formatting only:

- [Python PEP8 Autoformat](https://packagecontrol.io/packages/Python%20PEP8%20Autoformat) is missing
- [CodeFormatter](https://packagecontrol.io/packages/CodeFormatter) supports python only in ST2.

Of course there is [Anaconda](https://packagecontrol.io/packages/Anaconda), but it would be good to have another lightweight package, so users could choose.

thep0y:
> It looks like [AutoPep8](Auto%E2%80%8BPEP8) is the only active package dedicated to python formatting only:
> 
>     * [Python PEP8 Autoformat](https://packagecontrol.io/packages/Python%20PEP8%20Autoformat) is missing
> 
>     * [CodeFormatter](https://packagecontrol.io/packages/CodeFormatter) supports python only in ST2.
> 
> 
> Of course there is [Anaconda](https://packagecontrol.io/packages/Anaconda), but it would be good to have another lightweight package, so users could choose.

Since you want to update AutoPEP8, it is naturally the best choice. Python formatted packages also have PyYapf, and I am now using yapf to format python code.

braver:
Ok, looks like the way to go now is for you two to get together and figure out the next version of AutoPEP8 👍🏻

PR #8104, opened 2020-11-16
added Latin spelling dictionary

fawzisal:

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -34,8 +34,14 @@
 		},
 		{
 			"name": "Language - Latin",
-			"description": "Latin spelling dictionary for Sublime Text",
-			"labels": ["spell check", "dictionary", "latin"]
+			"details": "https://github.com/fawzisal/SublimeTextLanguageLatin",
+			"labels": ["spell check", "dictionary", "latin"],
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


repository/l.json
```diff
@@ -32,6 +32,19 @@
 				}
 			]
 		},
+		{
+			"name": "Language - Latin",
+			"description": "Latin spelling dictionary for Sublime Text",
+			"author": ["fawzisal"],
```
braver, line 38: You don't need to specify the "author" since "details" already points to your GitHub repo. Also the "description" key is better left out, we also take that from the repo.

braver (changes requested):
It looks like you're replicating some existing spelling packages, and you're including a packages.json. There is no documentation that really covers this, but that file is not necessary. There are also similar packages that don't include one. In any case yours is also incorrect (there is no "details" key inside a "releases" object. Anyway, just remove it.

Additionally, we require [semver](https://semver.org) version numbers, so 1.0.0 instead of 1.0. Please tag a new release.

braver:
@fawzisal please respond to the feedback in this PR, and resolve the conflicts, if you still want to add your package.

braver:
🎉

PR #8107, opened 2020-11-23
TabNine rename to Tabnine, description + labels fields added for better discoverability 

dimacodota:


braver:
If you link to the GitHub repo in "details", you don't need to specify "readme" and "issues". Also, you can add the "description" and "website" info to the GitHub repo:

Screenshot 2020-12-12 at 18 05 24

This way you can manage that information without having to update the package control entry.

rwols:
Isn't the amount of labels a bit excessive? Should we also add all possible language labels to [SublimeLinter](https://github.com/SublimeLinter/package_control_channel/blob/06b141c3eae1cb71af7c37b6f020778b9ce328b7/org.json#L9) and/or [LSP](https://github.com/sublimelsp/repository/blob/c575f4582ccea1535079b746751cb300fe6550a8/repository.json#L21-L27)?

braver:
It _is_ excessive. The package is not specific to a subset of languages, so tagging all languages of the world here just reduces the value of the other labels. It might look like a smart SEO trick but it just makes it less clear what this package is about.

dimacodota:
@braver it is actually all the languages Tabnine supports and was trained on

braver:
Well still, the long list of languages obscures the labels that _are_ useful. In the end though, there are no rules here, so 🤷🏻 .

Are you sure you want to change the author here? It's supposed to point to the author of the package, not necessarily the author of the underlying technology.

In any case please remove the "homepage", and maybe also the "description", keys. Because the "description" already points to the GitHub repo, we will take that information from there and this just duplicates that.

Please consider that we try not to have to update the package control channel for tiny changes to packages. So if the list of languages is very volatile (although there is hardly any major language not already covered), that's not something you want to have in the channel file.

dimacodota:
hey @braver, thanks for the clarification. regarding the author, the answer is yes, we want "Tabnine" to appear as the author of the plugin.

PR #8110, opened 2020-11-25
Add Chromeless Package

ggets:

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1045,6 +1045,16 @@
 			]
 		},
 		{
+			"name": "Chromeless",
+			"details": "https://github.com/ggets/Sublime_Chromeless",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "ChromeREPL",
 			"details": "https://github.com/acarabott/ChromeREPL",
 			"releases": [
--- a/repository/c.json
+++ b/repository/c.json
@@ -1047,9 +1047,13 @@
 		{
 			"name": "Chromeless",
 			"details": "https://github.com/ggets/Sublime_Chromeless",
+			"readme": "https://raw.githubusercontent.com/ggets/Sublime_Chromeless/main/README.md",
+			"labels": ["window manipulation", "fullscreen", "layout", "titlebar", "title bar", "chrome"],
+			"author": "GGets",
 			"releases": [
 				{
 					"sublime_text": ">=3000",
+					"platforms": ["windows", "linux"],
 					"tags": true
 				}
 			]
--- a/repository/g.json
+++ b/repository/g.json
@@ -414,6 +414,26 @@
 			]
 		},
 		{
+			"name": "Gimp Auto-Complete Python",
+			"details": "https://github.com/civAnimal/gimp_autocomplete_python",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
+			"name": "Gimp Auto-Complete Scheme",
+			"details": "https://github.com/civAnimal/gimp_autocomplete_scheme",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Ginkgo Completions",
 			"details": "https://github.com/onsi/ginkgo-sublime-completions",
 			"releases": [
--- a/repository/t.json
+++ b/repository/t.json
@@ -2282,6 +2282,16 @@
 			]
 		},
 		{
+			"name": "TinyScheme Auto-Complete",
+			"details": "https://github.com/civAnimal/tinyscheme_autocomplete",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "tipJS Snippets",
 			"details": "https://github.com/rkJun/sublime-tipJS-snippets",
 			"labels": ["snippets"],


braver:
You're shipping with keybindings (and only for Windows which is a bit odd). They conflict with existing bindings, and it's very common for keybindings to conflict with ones in other packages. We tend to recommend _not_ shipping keybindings, instead explaining how users can add what they want via the README.

Is this PR in sync with the one that just adds [screeninfo](https://github.com/wbond/package_control_channel/pull/8081)?

ggets:
> 
> 
> You're shipping with keybindings (and only for Windows which is a bit odd). They conflict with existing bindings, and it's very common for keybindings to conflict with ones in other packages. We tend to recommend _not_ shipping keybindings, instead explaining how users can add what they want via the README.
> 
> Is this PR in sync with the one that just adds [screeninfo](https://github.com/wbond/package_control_channel/pull/8081)?

I read about keymaps being considered bad practice. So I decided to approach this with a custom setting, to enable them. Right now, by default, it's being set to enabled, though. Maybe I should change it to disabled, then?
The screeninfo PR is a functional version and is required for Chromeless to work. They are both in stable, working stage right now. Is this what you mean in sync?

braver:
> So I decided to approach this with a custom setting, to enable them. Right now, by default, it's being set to enabled, though. Maybe I should change it to disabled, then?

A setting is a good idea, and you can go either way with the default. F11 in particular is already taken by default, so make sure that doesn't conflict or maybe pick another key. In general though, I would prefer to teach people how to make their own keybindings, so they can pick any key that makes sense for them. Editing a keybinding file isn't really any more difficult than editing a settings file to enable a preset keybinding. In the end though, we're not religious about it, just some things to think about. 
And if your package also supports MacOS and Linux, you should provide keybindings for those as well, so that your package behaves consistently.

> They are both in stable, working stage right now. Is this what you mean by 'in sync'?

Well, what I meant was that this PR also adds screeninfo. So we can ignore https://github.com/wbond/package_control_channel/pull/8081 and just focus on this one here.

ggets:
> A setting is a good idea, and you can go either way with the default. F11 in particular is already taken by default, so make sure that doesn't conflict or maybe pick another key. In general though, I would prefer to teach people how to make their own keybindings, so they can pick any key that makes sense for them. Editing a keybinding file isn't really any more difficult than editing a settings file to enable a preset keybinding. In the end though, we're not religious about it, just some things to think about.

I replaced F11, because it provides similar behavior and is more likely to be preferred to fullscreen by one who would install such a package. The setting actually remaps the original F11 shortcut (for the fullscreen toggle function) to CTRL+F11 as well, so that this functionality is not lost.

> And if your package also supports MacOS and Linux, you should provide keybindings for those as well, so that your package behaves consistently.
I haven't really tested it, although I included the code for it, but I am not sure about what the typical fullscreen keyboard shortcut on MacOS and Linux is.


> Well, what I meant was that this PR also adds screeninfo. So we can ignore #8081 and just focus on this one here.
It wasn't intentional, but it's better to handle it here, you're right.

braver:
> I haven't really tested it, although I included the code for it, but I am not sure about what the typical fullscreen keyboard shortcut on MacOS and Linux is.

Well, that's another reason not to ship keybindings 😉 Linux IIRC tends to follow Windows. In MacOS "full screen" means something else. You should not "steal" the default, because that controls the full-screen state of the application window. The mode that looks most similar to full screen on Windows doesn't have a keybinding. In fact, the window layout on MacOS is so different that what your package does might not even works as you'd expect. Maybe later I can get around to test this on MacOS and make a suggestion.

ggets:
> 
> 
> > I haven't really tested it, although I included the code for it, but I am not sure about what the typical fullscreen keyboard shortcut on MacOS and Linux is.
> 
> Well, that's another reason not to ship keybindings 😉

Hence the reason I only included the Windows ones.

> Linux IIRC tends to follow Windows.

Now has included same defaults for Linux. [( Include Linux kb shortcuts )](https://github.com/ggets/Sublime_Chromeless/commit/b2e38c2420a4cdf902aaf70ec8825de1d839b604)

> In MacOS "full screen" means something else. You should not "steal" the default, because that controls the full-screen state of the application window.
> In fact, the window layout on MacOS is so different that what your package does might not even works as you'd expect. Maybe later I can get around to test this on MacOS and make a suggestion.

I am open to suggestions (especially for MacOS as I don't have a test environment).

> The mode that looks most similar to full screen on Windows doesn't have a keybinding.

It doesn't have one, because it's implemented by the Chromeless plugin itself. At least that's the mode I meant by "provides similar behavior". I can't think of another similar mode you might be implying, unless you mean Distraction-free, which IMHO is an entirely different thing.

braver:
> It doesn't have one, because it's implemented by the Chromeless plugin itself.

Sorry, what I meant was that in MacOS the most similar feature ("Zoom"), doesn't have a default keybinding. So, in the OS, not Sublime Text. 

> I am open to suggestions

It's tricky to find a good binding, because most modifiers + F do something with search ("find"). And CMD+CTRL+F already does full-screen and you don't want to hi-jack that. The F* function keys are not something most Mac users use (because the Touch Bar hides them for instance) but you could perhaps use a modifier (e.g. CMD) + F11. 

> Changed default behavior to leave F11 unchanged unless the replace_default_fs_shortcut

Excellent 👍🏻

ggets:
Dropped MacOS from supported OSes list.

braver:
That’s a smart move. What you’re doing here maybe doesn’t even apply, or not in the same way, on MacOS. Ok, let’s ship this thing!

PR #8111, opened 2020-11-27
Add RISC-V Assembly Syntax

fabianschuiki:
Add syntax highlighting for the RISC-V assembly language.

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -1571,7 +1571,18 @@
 					"tags": true
 				}
 			]
-			},
+		},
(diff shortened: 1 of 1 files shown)


braver:
Thanks for submitting a package. Some pointers:
- The preferred label to use here is "language syntax"
- Your Sublime version should be ">=3092", not "*"

Please update your request with these changes!

fabianschuiki:
Done! Thanks for the feedback :tada:

PR #8113, opened 2020-11-30
Live Server added

molnarmark:
This package integrates the Live Server Node package, giving the ability to launch a local development server to serve content directly from Sublime Text.

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -1226,6 +1226,16 @@
 			]
 		},
 		{
+			"name": "Live Server",
+			"details": "https://github.com/molnarmark/sublime-live-server",
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "LiveStyle",
 			"details": "https://github.com/livestyle/sublime-text",
 			"releases": [
--- a/repository/l.json
+++ b/repository/l.json
@@ -1216,22 +1216,22 @@
 			]
 		},
 		{
-			"name": "Live Server",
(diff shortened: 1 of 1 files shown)


braver:
Thanks for submitting, looks useful 👍🏻 As far as a review goes, I bet you tested this and considered differences across operating systems. I wonder how well the emoji will work in the status bar. It's usually also recommended to stay a bit more low-key for persistent status info. Any thoughts there?

molnarmark:
Thanks for the feedback. 

Yes, I did test it, and used it quite a few times myself, but as it always goes, issues, if any will only get discovered if people start using it.

I was thinking about adding a setting to disable the emojis, maybe I should do that whenever I have some free time. I decided to have a persistent status message when the server is running for clarity for the user. If it's there, if they take a long break without shutting it down, they see its there and running, and won't try to start it again wondering if its still running or not.

It's just for the visuals, having another command in the palette to show the status would be a bit overkill in my opinion. Let me know the changes you'd suggest.

rwols:
Just a note: it looks like the server sticky status is not cleared when the plugin unloads.

braver:
> It's just for the visuals, having another command in the palette to show the status would be a bit overkill in my opinion

Sure 👌🏻 

> I was thinking about adding a setting to disable the emojis

It should be pretty easy to move the message strings to a settings file, so people can just do whatever.

> it looks like the server sticky status is not cleared when the plugin unloads

Indeed, there is a `plugin_unloaded()`, but it doesn't clear the the status.

molnarmark:
Thanks! Will definitely fix that. (in a bit though, a bit busy on the weekend)

PR #8114, opened 2020-12-01
Add ZipContents package

mawelborn:
ZipContents is a package for browsing and viewing the contents of zip archives opened in Sublime Text. The only existing package on Package Control that does this, [Zip Browser](https://packagecontrol.io/packages/Zip%20Browser), is no longer available.

### DIFF
--- a/repository/z.json
+++ b/repository/z.json
@@ -166,6 +166,16 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


braver:
Thanks! The old package isn't even in the channel anymore, so we're all green to add this one.

PR #8116, opened 2020-12-05
Add TabNav package

mitchvm:
TabNav is a plugin for keyboard navigation of tabular text data. Quickly move and select "cells" of text in the following formats, without taking your hands off the keyboard:

* Markdown pipe tables
* Org Mode tables
* Textile tables
* CSV files

TabNav also provides the ability to copy only the contents of the table, excluding markup, in a format that can be readily-pasted into other programs, such as Excel.

Demo

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -63,7 +63,7 @@
 			"name": "TabNav",
 			"details": "https://github.com/mitchvm/tabnav",
 			"author": "Mitch Valdmanis-Miller",
-			"labels": ["text navigation", "text selection", "text", "markdown"],
+			"labels": ["text navigation", "text selection", "text", "markdown", "org mode", "textile"],
 			"releases": [
 				{
 					"sublime_text": ">=3000",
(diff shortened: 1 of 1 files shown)


mitchvm:
Regarding all of those key binding warnings, while all of them are enabled by - the nature of the package requires many key bindings - they're only active in very particular contexts, and TabNav offers user settings to disable each key binding individually, should users prefer different key bindings.

mitchvm:
With commit 7ca635a I:

* Rebased off of the latest changes in this repo
* Added the `textile` label to reflect newly added functionality
* Removed one of the key bindings that was getting warnings (<kbd>Alt</kbd>+<kbd>W</kbd> in Windows and Linux)

I also added a [section to the README](https://github.com/mitchvm/tabnav#key-bindings) specifically about the masked key bindings.

mitchvm:
I've removed all of the default key bindings. I'll close this PR and open a new one simply to reduce the noise.

PR #8117, opened 2020-12-06
Add EraBasic syntax package

Fegelein21:
EraBasic syntax package for Sublime Text 3.


braver:
Thanks! Please add the the "language syntax" label to make the package easier to find.

braver:
🎉

PR #8119, opened 2020-12-09
Add Unreal Shader Package

hellokenlee:
Add Unreal Shader Package for UE4 Shader editing and syntax hightlighting and more.

### DIFF
--- a/repository/u.json
+++ b/repository/u.json
@@ -449,7 +449,7 @@
 			"labels": ["Shader","language syntax", "goto file", "Unreal", "UE4"],
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": ">=3092",
 					"tags": true
 				}
 			]
--- a/repository/u.json
+++ b/repository/u.json
@@ -444,6 +444,17 @@
 			]
 		},
 		{
+			"name": "Unreal Shader",
+			"details": "https://github.com/hellokenlee/SublimeUnrealShader",
+			"labels": ["Shader","language syntax", "goto file", "Unreal", "UE4"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Unreal Snippets",
 			"details": "https://github.com/awforsythe/sublime-unreal-snippets",
 			"labels": ["snippets", "Unreal", "UE4"],


braver:
Thank you for adding this package. Some points:

- Your context menu entry is in Chinese, whereas all documentation in the package is English. Unless your audience is restricted to the Chinese language, it usually makes the most sense to primarily use English.
- You provide a keybinding for Alt+G, which very easily clashes with keybindings people may already use. I suggest not providing a keybinding, instead adding an entry to the command palette, and in the README telling people how to configure a keybinding themselves.

hellokenlee:
> Thank you for adding this package. Some points:
> 
> * Your context menu entry is in Chinese, whereas all documentation in the package is English. Unless your audience is restricted to the Chinese language, it usually makes the most sense to primarily use English.
> * You provide a keybinding for Alt+G, which very easily clashes with keybindings people may already use. I suggest not providing a keybinding, instead adding an entry to the command palette, and in the README telling people how to configure a keybinding themselves.

Thank you for providing brilliant suggestions!

- Sure. Context menu entry is now in English.
- I consider this package should work out of the box... Instead of not providing a keybinding, I wonder if is there any way to let user change or call off this keybinding in settings. However, I cannot find any document to do so correctly.

braver:
Well, it's mostly about potential clashes with other packages. Users can always assign any keybinding to do something else. But package developers should consider that there are too few keys on the keyboard to support all packages out of the box. I don't consider it blocking for this PR though.

You could consider adding a "context" key to the keybinding. Since it only works in conjunction with your syntax, you could add

```json
"context":
    [{
        "key": "selector",
        "operator": "equal",
        "operand": "source.us",
        "match_all": true
    }]
```

hellokenlee:
> Well, it's mostly about potential clashes with other packages. Users can always assign any keybinding to do something else. But package developers should consider that there are too few keys on the keyboard to support all packages out of the box. I don't consider it blocking for this PR though.
> 
> You could consider adding a "context" key to the keybinding. Since it only works in conjunction with your syntax, you could add
> 
> ```json
> "context":
>     [{
>         "key": "selector",
>         "operator": "equal",
>         "operand": "source.us",
>         "match_all": true
>     }]
> ```

The "context" key seems like what I need. Than you! Updated in v1.0.5.

PR #8121, opened 2020-12-12
add config for no-response-robot

braver:
This sets up a bot to close "stale" issues and PR's after two weeks automatically. It's a pretty simple and limited bot, but it's a convenient way to set a two week time bomb under issues where the author has gone AWOL. I use it everywhere in SublimeLinter repo's to keep things clean, it's a little time saver. It cannot apply the "timeout" label though, so it does kinda break our protocol in that sense, but overall I'd say it's a win and the "timeout" labels can always be manually applied afterwards.

### DIFF
--- a/.github/no-response.yml
+++ b/.github/no-response.yml
@@ -6,8 +6,5 @@ daysUntilClose: 14
 responseRequiredLabel: "stale"
 # Comment to post when closing an Issue for lack of response. Set to `false` to disable
 closeComment: >
-  This issue has been automatically closed because there has been no response
-  to our request for more information from the original author. With only the
-  information that is currently in the issue, we don't have enough information
-  to take action. Please reach out if you have or find the answers we need so
-  that we can investigate further.
+  This pull request has been automatically closed because there has been no
+  response to our request for information or changes from the original author.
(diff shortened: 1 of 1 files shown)


FichteFoll:
LGTM, although we should update the automatic comment text to suit our needs, i.e. mention "pull request" and use more concise wording.

FYI, I cannot configure apps for this repository.

braver:
> FYI, I cannot configure apps for this repository.

Me neither, so that's probably just @wbond then?

.github/no-response.yml
```diff
@@ -6,8 +6,5 @@ daysUntilClose: 14
 responseRequiredLabel: "stale"
 # Comment to post when closing an Issue for lack of response. Set to `false` to disable
 closeComment: >
-  This issue has been automatically closed because there has been no response
-  to our request for more information from the original author. With only the
-  information that is currently in the issue, we don't have enough information
-  to take action. Please reach out if you have or find the answers we need so
-  that we can investigate further.
+  This pull request has been automatically closed because there has been no
+  response to our request for information or changes from the original author.
```
FichteFoll, line 10: Maybe add something like "you can comment to re-open" (if that's true) or "you may of course open a new pull request" to not discourage people that missed the deadline?

jrappen:
@braver Maybe use <https://github.com/actions/stale> instead? For further reactions <https://github.com/actions/github-script> might help as well.

braver:
🤷🏼‍♂️

PR #8124, opened 2020-12-15
new package: ruler proximity indicator

monsieurluge:
Changes: adding the new "Ruler Proximity Indicator" package data to the existing repositories.

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -2254,6 +2254,16 @@
 			]
 		},
 		{
+			"name": "RulerProximityIndicator",
+			"details": "https://github.com/monsieurluge/ruler_proximity_indicator",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Run",
 			"details": "https://github.com/glorieux/Run",
 			"releases": [


rwols:
Consider adding a `.gitattributes` file in your repository and ignoring the .png pictures. This will prevent the images from ending up in your .sublime-package. [Example](https://github.com/jisaacks/GitGutter/blob/master/.gitattributes)

You'd write something like

```
*.png export-ignore
```

monsieurluge:
> Consider adding a `.gitattributes` file in your repository and ignoring the .png pictures. This will prevent the images from ending up in your .sublime-package. [Example](https://github.com/jisaacks/GitGutter/blob/master/.gitattributes)
> 
> You'd write something like
> 
> ```
> *.png export-ignore
> ```

Done !

braver:
Pretty neat concept! Some things to consider:

You run on `on_selection_modified`. Have you considered using `on_selection_modified_async` instead? It's prevents your plugin from blocking the main thread. I have had instances where that ends up creating additional challenges though. Alternatively, because not every move of the cursor ends up really moving the indicator, you might want to consider bailing out of hitting set_status if there is no visible change. 

You check for `not settings.get('rulers') is None`, but instead you can pass a [default](https://www.sublimetext.com/docs/3/api_reference.html#sublime.Settings) for `settings.get()`. 

You have the 80 chars as a baked in default, so probably your indicator always shows the proximity to 80 even if you don't have any visible rulers. Which I think is actually quite handy. But it's also seemingly undocumented and maybe that would leave some users confused. Maybe put a note about that in the readme?

Have you considered what should happen with multiple cursors? And selections spanning multiple lines? Maybe you can find a way to make it clear what point in the selection is covered by your indicator. Alternatively, because of your intended use case, perhaps it's better to hide the indicator in these instances?

On my system the indicator is really hard to read. I guess you're stuck with the fonts on the system. Maybe there could be a setting for an increased readability using different characters?

Screenshot 2021-01-06 at 19 51 06
Screenshot 2021-01-06 at 19 50 58

monsieurluge:
Thanks @braver for all your hints, issues and remarks !
I'll try to take all of them into account for a future update.

1) I didn't know that so I'm ok with your proposal
2) thanks for the advice
3) this indicator is designed to help when the ruler is not visible so, yes, I'll try to explain the purpose in a better way in the readme file
4) you're right, I'll hide the indicator (or maybe change the appearance) when the selection is not unique
5) I thought about a custom character set using the package settings file, the one currently displayed being the default one

braver:
Okido, I'll let you get to that in your own time. Let's ship this thing!

PR #8125, opened 2020-12-15
Add Tailwind reorder

54nd10:

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -257,7 +257,6 @@
 		{
 			"name": "Tailwind Reorder",
 			"details": "https://github.com/54nd10/Tailwind-reorder",
-			"author": "54nd10",
 			"releases": [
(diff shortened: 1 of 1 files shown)


braver:
Thanks for submitting!

You don't have to specify the "author" key. It's taken from the GitHub repo. So, better to remove it.

You ship with keybindings. Because they tend to conflict with keybindings in other packages, it's not recommended. Instead, we recommend explaining in the README how to set up a keybinding. But, that's not a requirement for this PR per se. However, your commands should also be available via the command palette via a `Default.sublime-commands` file. The command palette is the primary place for users to discover available commands, and in many cases being available there is as good as having a keybinding. And if users have already assigned your keybinding to something else, the command is still available there.

braver:
🎉

PR #8129, opened 2020-12-16
added ContextLangSyntax package

Magorx:
ContextLangSyntax provides some basic syntax highlight for a programming language named 'Context', which is being developed by me as an extension of university homework. It comes alongside a custom color scheme picked up specifically for this language.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3347,6 +3347,16 @@
 			]
 		},
 		{
+			"name": "ContextLang",
+			"details": "https://github.com/Magorx/ContextLang_SublimePackage",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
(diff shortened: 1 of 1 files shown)


braver:
If you're developing the language as a homework assignment, is it relevant for anyone else? You don't need to distribute a package via package control, there are other ways to do so if the audience is limited.

A language that is accompanied by color scheme is usually a red flag. If the syntax is structured correctly this is not necessary. Please read the [Scope Naming guidelines](https://www.sublimetext.com/docs/3/scope_naming.html). If you follow that guide, your syntax will make more sense to Sublime Text, will probably already have good colors, and correct scope names also unlock various features like auto complete and goto jumps.

Magorx:
Sounds fair, thanks

PR #8130, opened 2020-12-16
KuaiJi

ChengAnXu2014:

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -1108,6 +1108,17 @@
 			]
 		},
 		{
+			"name": "EraBasic",
+			"details": "https://github.com/Fegelein21/sublime-erabasic",
+			"labels": ["language syntax"],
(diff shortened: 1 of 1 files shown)


ChengAnXu2014:
Hope this time I can install my own plugin with Package Control.
It will be very cool!

braver:
Thanks!

You ship with a keybinding. This is usually not recommended, as they tend to clash with existing bindings or bindings in other packages: there are not enough keys on the keyboard for every package! So, instead, I would recommend creating a command palette entry ([example](https://github.com/braver/SublimeGitUp/blob/master/Default.sublime-commands)) instead. Then in the README you can explain how users can set up a keybinding for your package if they wish.

braver:
Well, you can look at the keybindings any time you like. Let's just add this package.

PR #8131, opened 2020-12-17
Add ride package with LF syntax highlighting.

sgabl:
Sublime Text 3 package for basic Lingua Franka (LF) syntax highlighting, snippets and auto-completions. Lingua Franka is created by [Franka Emika](https://www.franka.de).

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -1529,6 +1529,17 @@
 			]
 		},
 		{
+			"name": "ride",
+			"details": "https://github.com/frankaemika/ride-sublime",
+			"labels": ["lf", "Lingua Franka", "language syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Riemann Completions",
 			"details": "https://github.com/jarretth/sublime-riemann-autocomplete",
 			"labels": ["completion"],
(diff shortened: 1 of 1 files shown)


repository/r.json
```diff
@@ -1528,6 +1528,17 @@
 				}
 			]
 		},
+		{
+			"name": "ride",
```
braver, line 1532: Why is it called "ride"? Is that a name that would make sense to the target audience? To me, if it is the syntax for a language called "Lingua Franka", why not call the package "Lingua Franka Syntax"?

braver (changes requested):

repository/r.json
sgabl: Hi, ride is the name of our App-Development SDK so the target audience will probably search for that. Syntax highlighting is only the first step of what the extension could provide.

I'm fine if the plugin is for now named to 'Lingua Franka Syntax' and once more features have been added renamed to 'ride'.

rwols: But not that there's also an R-IDE package: https://packagecontrol.io/packages/R-IDE

sgabl: Could also name it 'Franka Ride', would that be ok?

braver: > But not that there's also an R-IDE package

I don't think that's a problem though.

> Could also name it 'Franka Ride', would that be ok?

If "Ride" makes any sense for your context that's all fine. It doesn't really matter what it's called, but for an outsider it just seemed a bit random. So, "ride" is ok 🤷🏻

PR #8132, opened 2020-12-17
Add package: AutoSplit

egnartsms:

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -2390,7 +2390,7 @@
 		{
 			"name": "AutoSplit",
 			"details": "https://github.com/egnartsms/sublime-autosplit",
-			"author": "Serhii Mozghovyi",
+			"author":  "Serhii Mozghovyi",
 			"releases": [
 				{
 					"sublime_text": "*",
--- a/repository/a.json
+++ b/repository/a.json
@@ -2393,7 +2393,7 @@
 			"author": "Serhii Mozghovyi",
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": ">=3000",
 					"tags": true
 				}
 			]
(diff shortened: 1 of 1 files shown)


egnartsms:
[AutoSplit](https://github.com/egnartsms/sublime-autosplit) is a Sublime Text 3 plugin for automatic splitting argument lists across multiple lines and joining them back up. It supports automatic splitting when exceeding the ruler, and arrow indicators suggesting that it's possible to join the arglist up.

repository/a.json
```diff
@@ -2387,6 +2387,17 @@
 				}
 			]
 		},
+		{
+			"name": "AutoSplit",
+			"details": "https://github.com/egnartsms/sublime-autosplit",
+			"author": "Serhii Mozghovyi",
+			"releases": [
+				{
+					"sublime_text": "*",
```
braver, line 2396: If it's a sublime text 3 package, please indicate a required sublime_text version number >3000.

braver (changes requested):
Looks like a really useful package! Thanks for submitting. 

You're shipping with keybindings, which is always a bit tricky. There are just not enough keys on the keyboard to give every command for every package their own binding. The ones you're using already have a function (at least on MacOS), so you would be repurposing keyboard shortcuts people might rely on already. 
I tend to advocate shipping with commands in the command palette (via Default.sublime-commands), and in menu's where applicable. And then via the README, with examples, teach people how to set up keybindings themselves.

repository/a.json
egnartsms: Thank you for valuable suggestions! Done.

egnartsms:
@braver am i not holding things up? It's the first time i submit a PR to a project with this type of build process on GitHub, so want to make sure i've done all i had to this far.

braver:
> am i not holding things up

Nah, I'm just trying not to be behind a laptop the entire time around the holidays.

I notice you renamed the keymap file. I'm not 100% sure that works, I believe ST will just load all the keymap files it finds regardless of their filename. So if you want to have the example the best place to keep that is in the README.

Anyway, that's not blocking here, so let's release this thing!

PR #8134, opened 2020-12-19
Renamed 'DarkDonut Forest Coders' -> 'DarkDonut Color Schemes'

dakshamdev:
We are about to add more colors to it.
So, it must be renamed.
Thanks for Approving 🙂

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -149,7 +149,6 @@
 			"name": "DarkDonut Color Schemes",
 			"previous_names": ["DarkDonut Forest Coders"],
 			"details": "https://github.com/DarkDonut-theme/DarDonut_sublime",
-			"description": "A set of Color Scheme for Sublime Text 3 that fits your flavour! 🎃",
 			"releases": [
 				{
 					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


repository/d.json
```diff
@@ -146,9 +146,10 @@
 			]
 		},
 		{
-			"name": "DarkDonut Forest Coders",
-			"details": "https://github.com/DarkDonut-theme/forest_coders_subime",
-			"description": "Forest Color Scheme for Green lovers 🎃 ",
+			"name": "DarkDonut Color Schemes",
+			"previous_names": ["DarkDonut Forest Coders"],
+			"details": "https://github.com/DarkDonut-theme/DarDonut_sublime",
+			"description": "A set of Color Scheme for Sublime Text 3 that fits your flavour! 🎃",
```
braver, line 152: I would prefer if you updated the description of the GitHub repository, and then remove the "description" key here. It is not required, we will take the GitHub description instead. And that will allow you to change it at will without having to change the package control entry.

braver (changes requested):

repository/d.json
dakshamdev: Done and Thanks 🙂

PR #8135, opened 2020-12-24
Add Nunjucks package

alsolovyev:
Nunjucks syntax highlighting and snippets for Sublime Text 3

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -1183,10 +1183,7 @@
 		{
 			"name": "Nunjucks",
 			"details": "https://github.com/alsolovyev/Nunjucks",
-			"readme": "https://raw.githubusercontent.com/alsolovyev/Nunjucks/master/README.md",
-			"issues": "https://github.com/alsolovyev/Nunjucks/issues",
 			"labels": ["nunjucks", "language syntax", "snippets"],
-			"author": "janeRivas",
 			"releases": [
 				{
 					"sublime_text": ">=3092",
--- a/repository/n.json
+++ b/repository/n.json
@@ -1189,7 +1189,7 @@
 			"author": "janeRivas",
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": ">=3092",
 					"tags": true
 				}
 			]
--- a/repository/n.json
+++ b/repository/n.json
@@ -1181,6 +1181,20 @@
 			]
 		},
 		{
+			"name": "Nunjucks",
+			"details": "https://github.com/alsolovyev/Nunjucks",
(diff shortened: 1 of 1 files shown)


braver:
We already have [a package for that](https://packagecontrol.io/packages/Nunjucks%20Syntax). Yours looks a bit more complete though 😄 If it's ok with @jeremyfourna I think it makes the most sense to replace the existing package with yours. His is a fork of something even older and that makes it harder to do things like issues and PR's so a clean slate looks like a good idea. 

We need to give the existing package author 2 weeks to reply. In the mean time, let's look at the PR itself:
- Because "details" points to a GitHub repo, you don't have to add the "issues" and "readme" keys (unless they're in a non-standard place, which they're not). Please remove them.
- You add the "author" key for on "janeRivas". Since you're seemingly the author of this particular package, the author should point to you (ie. not the language author or something). Again because you add a GitHub repo you don't need to specify the author.

jeremyfourna:
You can replace it no problem for me

alsolovyev:
@jeremyfourna, thank you for your quick reply!
@braver, thank you for your attention! Do I need to modify or add anything else?

braver:
So, what you need to do is remove your new addition, and instead change the details URL of the existing "Nunjucks Syntax" package. Double-check the other info like version requirement. You _can_ rename it if you want to, but it's not necessary. 

You might also want to add a message so that users that are upgraded to the new version also know what has happened and what they can expect from it.

alsolovyev:
After doing some research, I found that packages with similar functionality are named after the name of the language engine they support (e.g., [PUG](https://packagecontrol.io/packages/Pug), [Handlebars](https://packagecontrol.io/packages/Handlebars), [Haml](https://packagecontrol.io/packages/Haml), [Slime](https://packagecontrol.io/packages/Slime)). So, should we follow this naming convention and leave the **"Nunjucks"** name?
Besides, my package contains snippets and preferences (and maybe some additional features in the future), which can confuse some users if the name will be **"Nunjucks Snippets"**.

braver:
Yes, definitely. This is the main support package for that language. Even just the syntax unlocks so much more than just highlighting. So, the rename to just "Nunjucks" is a good idea.

repository/n.json
```diff
@@ -1181,12 +1181,12 @@
 			]
 		},
 		{
-			"name": "Nunjucks Syntax",
-			"details": "https://github.com/jeremyfourna/sublime-nunjucks/",
-			"labels": ["language syntax", "Nunjucks"],
+			"name": "Nunjucks",
```
braver, line 1184: For the rename to work, we need to add `"previous_names": ["Nunjucks Syntax"]`

braver:
Alright, let's roll! 🛥️

PR #8139, opened 2020-12-26
Add Color Scheme 'Limegrass'

saleh49:

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -950,6 +950,17 @@
 			]
 		},
 		{
+			"name": "Limegrass Color Scheme",
+			"details": "https://github.com/saleh49/Limegrass",
+			"labels": ["color scheme"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "LimitTabs",
 			"details": "https://github.com/maximilize/LimitTabs",
 			"releases": [


Ultra-Instinct-05:
If your color scheme only targets ST, then it's probably better to go for a `.sublime-color-scheme` format since that receives updates and new features & the `.tmTheme` format is only supported for legacy reasons.

As to the image screenshots, there is actually no need to ship those to the user so probably could use a `.gitattributes` file to ignore them when PC downloads the package artifacts to the user machine.

I haven't gone through the color scheme, but if you weren't aware, there is a [minimal scope coverage guide](https://www.sublimetext.com/docs/scope_naming.html#color_schemes) which you can follow so that your color scheme looks good for most source files.

saleh49:
> .sublime-color-scheme

Ok, now I have made the changes for the pics which need to ignore.

braver:
If you want to maintain and tweak this theme later, you definitely want to switch to using the `.sublime-color-scheme` format (IIRC Sublime Text even has a built-in converter for it). For one thing it supports variables, which makes it so much easier to keep all the colors straight.

Anyway, it doesn't really matter for the PR although officially you would need to bump the version requirements. So, let me know if you intend to make the switch, otherwise I'll just merge this.

saleh49:
> If you want to maintain and tweak this theme later, you definitely want to switch to using the `.sublime-color-scheme` format (IIRC Sublime Text even has a built-in converter for it). For one thing it supports variables, which makes it so much easier to keep all the colors straight.
> 
> Anyway, it doesn't really matter for the PR although officially you would need to bump the version requirements. So, let me know if you intend to make the switch, otherwise I'll just merge this.

Yes Merge this.

rwols:
Hi @saleh49, I don't think your .gitattributes file works: https://github.com/saleh49/Limegrass/blob/main/.gitattributes

It's missing the `export-ignore` keyword. You can test your work by running

```
git archive --format=zip -o 'Limegrass Color Scheme.sublime-package'
```

PR #8140, opened 2020-12-27
Add AutoMatlab package

asparc:
Short description:
Autocompletion, documentation generation and a command panel for Matlab.

Comparison to related packages:
The two other Matlab-related packages are relatively limited. One is a collection of snippets, and the other scans for function names in a project.
AutoMatlab does the same, and a lot more. 
- It generates autocompletion information from the user's Matlab installation, the project, the current folder(s) and/or the current file. On top of that, it includes elaborate function documentation with the autocompletion, presented in a popup. 
- It provides flexible and context-aware function documentation generation through customizable template snippets.
- It allows to run Matlab commands from within Sublime Text, through integration with AutoHotkey. This is for instance useful to run the current file in Matlab, or to repeat commands from the recent Matlab history through a custom command panel.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -2290,6 +2290,18 @@
 			]
 		},
 		{
+			"name": "AutoMatlab",
+			"details": "https://github.com/asparc/AutoMatlab",
+			"labels": ["auto-complete", "snippets", "documentation", "build system", "matlab"],
+			"releases": [
+				{
+					"platforms": ["windows"],
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "AutoPEP8",
 			"details": "https://github.com/wistful/SublimeAutoPEP8",
 			"labels": ["linting", "formatting"],


asparc:
None of these keybindings interfere with the default keybindings, because 
- the overridden default keybindings have no sensible meaning in the matlab scope,
OR
- the provided context is so specific that the redefined keybinding provides more natural behaviour than the default one

braver:
Alright, I can agree with that I guess. In general shipping keybindings is not recommended because they tend to conflict, if not with the defaults, then with other packages. It's usually better to let users set up their own, and instead focus on providing command via the command palette.

That said, some other minor points. I'm not able to test your package, but just making sure that you've considered these things:

- There's a [print statement](https://github.com/asparc/AutoMatlab/blob/main/auto_matlab_commands.py#L342) that seems superfluous. Might be useful as a logging mechanism, but usually I would recommend not to log anything unless in a debug mode.
- Packages are installed in an archived format. This means you can't read write to the package directory via the file system, and your package is also not in Packages directory like it would if you're developing the package. You can read from the package via `load_resource`. I'm not 100% clear on what you're exactly [writing](https://github.com/asparc/AutoMatlab/blob/main/auto_matlab_commands.py#L308), or where you're writing, but that's something to be careful with.

asparc:
You looked into it in more detail than I expected! Thanks for doing that and thanks for your comments! :-D

About the keybindings:
I tried to keep the keybindings to a minimum (in my personal version, I have quite a few more than the default ones), but I think that the ones I introduced by default are essential in the matlab scope. In the README there are also instructions on the syntax for custom commands (to be used through the command palette or keybindings).

About your remarks:
- Alright, I commented out all print statements that I used for logging. These messages will only briefly appear in the status bar now.
- Yes, I'm aware of the archive format. I think I handled it well, but I'll have to test it once AutoMatlab is packaged as a sublime-package. Two elements of AutoMatlab need to be stored outside a sublime-package archive, though. As the instructions adviced against using `.no-sublime-package`, I did the following instead:
    1. The default Matlab autocompletions (about 3000 of them) are archived in the sublime-package. If the user finds these insufficient, he/she can use AutoMatlab to generate new keybindings based the Matlab installation on his/her system. Those will be stored in `Packages/AutoMatlab/data/matlab_completions` (so outside of the sublime-package archive). Using Sublime's resource mechanism, I then select which `matlab_completions` file I want to use.
    2. To run commands from Sublime in Matlab (i.e. send commands to matlab.exe), I used a tool called AutoHotkey (which is a dependency the user needs to install, as per the README). AutoHotkey uses `.ahk`-scripts, which do not work from an archived state. Therefore, upon the first run, I copy the `.ahk`-script from the sublime-package archive to `Packages/AutoMatlab/ahk/run_in_matlab.ahk`. Incidentally, I think it's actually an advantage for that script to be available outside of the archive, as I can imagine that some users might want to tweak it. 

  As soon as I can test AutoMatlab with the package control, I also intend to add some clean up instructions that remove the `AutoMatlab` folder from the `Packages` folder upon uninstallation, using the Events API `events.remove("Package Name")`.

Hope this clarifies your considerations :-)

braver:
Thanks, excellent. We do try to take the opportunity to ensure a level of quality, and of course to prevent issues stemming from the way packages are delivered.

If you write a file to the same place in Packages/<package name> as the file exists in your package, Sublime will automatically use that to override the file in the package. So I would expect that mechanism to work for your matlab_completions, without you needing any logic to select which file to use. But as you say that’s all a bit easier to figure out once your package is actually shipped.
You can maybe test this locally, by creating the archive yourself:
https://packagecontrol.io/docs/creating_package_files

asparc:
Well, the 'logic' that I was referring to for selecting the right matlab_completions, was nothing more than selecting the first (or last, I forgot) entry of the matlab_completions resources list that Sublime found for me. So indeed, nothing advanced necessary there, Sublime handles that very well on its own :-D

Thanks for the tip! I tried creating a sublime-package and it works like a charm :-) 
A small question in this regard. I'm saving user-generated (system-specific) autocompletion data (as well as the unarchived autohotkey script) in the custom created folder `Packages/Automatlab/...`. Would you advice against this and advocate to use the sublime Cache folder instead? If so, is this cache ever cleaned up (by sublime or windows)? Because that would yield the frustrating scenario where the user would unexpectedly need to regenerate his/her autocompletion data.

braver:
> selecting the first (or last, I forgot) entry of the matlab_completions resources list that Sublime found for me

Sounds good. What I was thinking was that you could also let Sublime Text's package override system handle it. I.e. your defaults would be in `Installed Packages/AutoMatlab/some_resource.file` and you write the customised version to `Packages/AutoMatlab/some_resource.file` and that would would automatically take precedence over the original/default file. Not 100% sure that would apply here though, and your approach is good either way.

I wouldn't advise a package to ever write to the cache directory. I don't believe it's considered writable for packages directly. And in any case, users would assume they can wipe the cache without losing anything other than maybe some state or history (ie. anything that's not volatile anyway). 

In line with my first point, anything in `Packages/AutoMatlab` will effectively replace the same resource from your package. That's what those folders in `Packages` are for: to either add new packages, or override installed packages. If you're storing stuff for the user, ie. their preferences, setting and customisations, that should go into `Packages/User` (and then maybe `Packages/User/AutoMatlab` if it's more than one file).

asparc:
Thanks for that clarification! From your pointers, I understand that I'm indeed overriding some resources in `Installed Packages/AutoMatlab` in `Packages/AutoMatlab`. That's exactly what I intended to be doing :-)

Is any extra action/validation needed from my side before AutoMatlab can be merged into the package_control_channel?

braver:
Nope, I think that concludes our review 😄 Let's ship! 🚤

PR #8142, opened 2020-12-29
Add Debian Packaging Syntax

barnumbirr:
Debian packaging file syntax highlighting.

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -347,6 +347,20 @@
 			]
 		},
 		{
+			"name": "Debian",
+			"description": "Debian packaging file syntax highlighting for Sublime Text.",
+			"author": "barnumbirr",
+			"details": "https://github.com/barnumbirr/sublime-debian",
+			"issues": "https://github.com/barnumbirr/sublime-debian/issues",
+			"labels": ["language syntax", "Debian"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Debug Killer",
 			"details": "https://github.com/adampresley/sublime-debugkiller",
 			"releases": [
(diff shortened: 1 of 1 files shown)


repository/d.json
```diff
@@ -346,6 +346,20 @@
 				}
 			]
 		},
+		{
+			"name": "Debian",
+			"description": "Debian packaging file syntax highlighting for Sublime Text.",
+			"author": "barnumbirr",
+			"details": "https://github.com/barnumbirr/sublime-debian",
+			"issues": "https://github.com/barnumbirr/sublime-debian/issues",
```
braver, line 354: Since you provide a "details" key that points to a GitHub repo, you don't need to specify "issues", "author" or "description": this information is just taken from the repo, which makes it much easier to update later etc. So, please remove those keys.

repository/d.json
```diff
@@ -346,6 +346,20 @@
 				}
 			]
 		},
+		{
+			"name": "Debian",
```
braver, line 350: The name is fine basically, at least it's not taken yet. But since "Debian" is a Linux distro, and not a specific language, just the word "Debian" doesn't say what the package does. Maybe "Debian package syntax" or "Debian Syntax"?

braver (changes requested):

repository/d.json
barnumbirr: I didn't know that was pulled automatically, thanks. Done.

repository/d.json
barnumbirr: Updated to `Debian Syntax`.

braver:
Thanks, that's mostly all set for the channel addition. 

As for your syntaxes, the convention is to have the language at the end of each scope. So in this case probably, `keyword.control.debian` instead of simply `keyword.control`. This allows a color scheme to target a language specific variation easily, while still also matching for the generic `keyword.control`.

barnumbirr:
This has now been fixed upstream, thanks for the help 🙏 .

PR #8143, opened 2020-12-29
Added Laravel Lang package

mariovalney:

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -308,6 +308,17 @@
 			]
 		},
 		{
+			"name": "Laravel Lang",
+			"details": "https://github.com/mariovalney/laravel-lang-for-sublime",
+			"labels": ["laravel", "helper", "snippets", "lang", "l10n", "formatting"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Laravel Migrations Snippets",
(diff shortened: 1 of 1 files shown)


braver:
I'm not 100% what the utility is here, but I don't know Laravel very well.

In any case, usually you _don't_ want to ship keybindings. They tend to conflict with keybindings in other packages, or ones users configured for themselves. If you provide commands, they should show up in the command palette via a `.sublime-commands` file. That makes them discoverable, and in most cases is just as easy to use as a keybinding. You can then explain in the readme how users can set up keybindings if they want them.

Oh, and please add a license file.

braver:
@mariovalney please respond to the feedback here if you still want to add your package to package control.

mariovalney:
> I'm not 100% what the utility is here, but I don't know Laravel very well.

Hi. It's a simple package to help add translation functions during code writting.

> In any case, usually you _don't_ want to ship keybindings. They tend to conflict with keybindings in other packages, or ones users configured for themselves. 

Hi. Removed the default keybindings,

> Oh, and please add a license file.

Added.

--

Thanks for your comments and revision.

braver:
Thanks for responding! So, for users to be able to find your commands, please add a Default.sublime-commands file (have a look here [for an example](https://github.com/SublimeLinter/SublimeLinter/blob/master/Default.sublime-commands)).

mariovalney:
Done, sir!

braver:
🍻

PR #8144, opened 2020-12-31
Update MKB Package Information

KeeMeng:
Sorry for new pull request :/
Hopefully this is ok for merging

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -1818,7 +1818,6 @@
 		{
 			"name": "MKB",
 			"details": "https://github.com/KeeMeng/MKB-Syntax-Highlighting",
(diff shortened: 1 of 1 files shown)


braver:
Why do you require 2 different (but mostly the same) README's?

KeeMeng:
The readme for the github page includes other information which package control does not need, such as manual installation, and also the table feature of markdown doesn't seem to work on package control. 
Hope this is understandable 😀

braver:
Oh, usually it's understood that the manual installation part doesn't apply to package control. It's still good information if someone finds the package via package control. It doesn't hurt to have it and it simplifies a lot to have just one file with that info.

Simple tables tend to work just fine. For example: https://github.com/SublimeLinter/SublimeLinter#key-bindings vs. https://packagecontrol.io/packages/SublimeLinter

braver:
⛵

PR #8145, opened 2021-01-01
Git Extensions Plugin 

LinuxDevon:
Adding in "Git Extensions" repository that I forked. It wasn't being maintained and was broke so I decided to fix up the bugs. I would like this to be available on Package Control instead of using Github. I am hopeful this resolves the issue with not import default settings of the repository as well. Still not sure why it won't import these.

This plugin allows the user to open up the application `Git Extenstions` from sublime. It is useful when wanting to write commits, browse the repo, or any other git actions using the application.

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -557,9 +557,9 @@
 			]
 		},
 		{
-			"name": "Git Extenstions",
+			"name": "Git Extensions",
 			"details": "https://github.com/LinuxDevon/sublime-gitextensions",
-			"labels": ["gitextenstions", "git"],
+			"labels": ["gitextensions", "git"],
 			"releases": [
 				{
 					"sublime_text": "*",
--- a/repository/g.json
+++ b/repository/g.json
@@ -557,6 +557,17 @@
 			]
 		},
 		{
+			"name": "Git Extenstions",
+			"details": "https://github.com/LinuxDevon/sublime-gitextensions",
+			"labels": ["gitextenstions", "git"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


repository/g.json
```diff
@@ -556,6 +556,17 @@
 				}
 			]
 		},
+		{
+			"name": "Git Extenstions",
```
rwols, line 560: Typo

LinuxDevon: Thanks for the catch!

Ultra-Instinct-05:
In your `Main.sublime-menu` https://github.com/LinuxDevon/sublime-gitextensions/blob/master/Main.sublime-menu#L93-L110, instead of using the `open_file` command to open the settings file, I would recommend using the `edit_settings` command since that's recommended and you can open both the default & user settings in a 2 col new window layout at the same time rather than opening them one at a time which can be annoying to a user, who has to switch back & forth between tabs. So something like

```json
{
    "caption": "GitExtensions - Settings",
    "command": "edit_settings",
    "args": {
         "base_file": "${packages}/GitExtensions/GitExtensions.sublime-settings",
         "user_file": "${packages}/User/GitExtensions.sublime-settings",
         "default": "{\n\t$0\n}\n",
    }
},
```

Ultra-Instinct-05:
Also please add a license file.

braver:
> So something like

In fact, you don't need to supply the `"user_file"`, just the base_file and the default.

LinuxDevon:
Just added the license file. Also added the suggestion to `Main.sublime-menu` to use the `open_file` command. 
I left off the `"user_file"` as suggested by braver since I don't want to supply them.

Thank you all for the help on this.

braver:
🚢 🚀

PR #8146, opened 2021-01-02
Add TabNav package

mitchvm:
TabNav is a plugin for keyboard navigation of tabular text data. Quickly move and select "cells" of text in the following formats, without taking your hands off the keyboard:

* Markdown pipe tables
* Org Mode tables
* Textile tables
* CSV files

TabNav also provides the ability to copy only the contents of the table, excluding markup, in a format that can be readily-pasted into other programs, such as Excel.

Demo

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -60,6 +60,18 @@
 			]
 		},
 		{
+			"name": "TabNav",
+			"details": "https://github.com/mitchvm/tabnav",
+			"author": "Mitch Valdmanis-Miller",
+			"labels": ["text navigation", "text selection", "text", "markdown", "org mode", "textile", "csv"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Tabnine",
 			"previous_names": ["TabNine"],
 			"details": "https://github.com/codota/tabnine-sublime",


braver:
Looks really useful thanks! We already have a [Table Editor](https://packagecontrol.io/packages/Table%20Editor) package that does a couple of extra tricks. That one's dead though. Would've been nice to replace it, but your feature set looks pretty solid too and not overcomplicated. 

The only thing I have to add here is about the name. "Tab" for me signals either tab-indentation or editor-tabs, and my first thought was that it would have something to do with switching between different files in the editor. So, personally I would've put "Table" in the name so that it's immediately clear what the package is for. Your call though, let me know what you think.

mitchvm:
> We already have a [Table Editor](https://packagecontrol.io/packages/Table%20Editor) package that does a couple of extra tricks.

Yes, I'm aware of that package, though I don't use it. I believe the feature sets are complimentary, rather than overlapping. My package provides no commands to edit tables; it's really all about selecting and navigating table contents, which is something that package doesn't really provide, AFAIK, or at least in a much more limited capacity.

> The only thing I have to add here is about the name. "Tab" for me signals either tab-indentation or editor-tabs, and my first thought was that it would have something to do with switching between different files in the editor. So, personally I would've put "Table" in the name so that it's immediately clear what the package is for. Your call though, let me know what you think.

I appreciate the feedback. Initially I was going to use the Tab key for lateral navigation, so thought TabNav would be a bit of a play on that as well as "table", but then I moved away from using Tab (even in the recommended key bindings).

I have changed the name to "Table Navigator".

braver:
Excellent, thanks!

The rename probably also affects your package. I see some minor stuff like comments over settings/keybinding files, and maybe the readme. But please also check if you're reading any resources or settings files in a way that is affected by the package name.

braver:
@mitchvm please let me know when you've had a chance to test how your package responds to the change in package name. We need to be sure it doesn't break how you're accessing resources and settings.

mitchvm:
Sorry, I've been extremely busy the past few weeks, and for at least another week or so going forward. I thought that changing the name would only affect how the package would be displayed in Package Control, not the actual function of the package. I'll update the package to use the new name in due course.

braver:
Yes, the name matters for how you access certain resources. No worries, I know you're taking this seriously, so just let me know any time you're ready 👌🏻

mitchvm:
@braver I changed my mind. I'm going to stick with the name "TabNav". I do appreciate your feedback, though.

braver:
Alright! You can always come back and do it later if you still want to then.

PR #8153, opened 2021-01-07
Update Earthbound Themes package path

benbusby:
The previous Earthbound Themes repo has been split up into a theme generation repo, and individual editor submodules. This updates the package path to point to the Sublime-specific submodule.

I also updated the name of the package to have a space between "Earthbound" and "Themes" -- not sure if this is a breaking change for existing users of the theme set or not. If so, I'm happy to change it back.

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -21,9 +21,9 @@
 			]
 		},
 		{
-			"name": "EarthboundThemes",
-			"details": "https://github.com/benbusby/earthbound-themes",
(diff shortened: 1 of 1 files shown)


braver:
To handle the rename transparently for everyone, you can add a “previous_names” key with the old name (you can find examples of this throughout the channel).

benbusby:
Added -- thanks for the tip.

PR #8154, opened 2021-01-08
Mongolian Dictionary package

bataak:

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -33,6 +33,18 @@
 			]
 		},
 		{
+			"name": "Language - Mongolian",
+			"description": "Mongolian spelling dictionary",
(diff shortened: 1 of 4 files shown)


repository/l.json
```diff
@@ -32,6 +32,18 @@
 				}
 			]
 		},
+		{
+			"name": "Language - Mongolian",
+			"description": "Mongolian spelling dictionary",
+			"details": "https://github.com/bataak/sublime-dict-mn",
```
braver, line 38: Thanks, looks good basically. I would prefer it if the description key here were removed. Instead you can enter a description on the github repo, and that will be used by package control. That way you can change the description at any time without having to open a PR here.

braver (changes requested):

PR #8156, opened 2021-01-12
Add ModernFortran

eirik-kjonstad:
Adds a Fortran language syntax (called Modern-Fortran) that supports features introduced in the more recent Fortran language standards - the 2003, 2008, and 2018 standards.

The existing alternative package (Fortran) offers relatively good support for pre-2003 standards. But it does not fully support more recent functionality, such as the extensive object orientation additions of the 2003 and 2008 standards. The Modern-Fortran syntax should be useful for developers who make use of such features.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -1933,6 +1933,17 @@
 			]
 		},
 		{
+			"name": "ModernFortran",
+			"details": "https://github.com/eirik-kjonstad/modern-fortran-syntax",
+			"labels": ["language syntax", "fortran"],
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "ModernPerl",
 			"details": "https://github.com/Blaizer/ModernPerl-sublime",
 			"labels": ["language syntax", "todo"],


braver:
It seems like the [existing package](https://packagecontrol.io/packages/Fortran) is both over-scoped (usually you don't want a linter in the same codebase as a syntax package) and un-maintained. That's an unfortunate combination. It would be a good idea to position your package as a replacement. I don't know much about Fortran... I assume your package also highlights older versions?

Anyway, this looks good. Better go and tell people it's here ;) Thanks!

eirik-kjonstad:
> It seems like the [existing package](https://packagecontrol.io/packages/Fortran) is both over-scoped (usually you don't want a linter in the same codebase as a syntax package) and un-maintained. That's an unfortunate combination. It would be a good idea to position your package as a replacement. I don't know much about Fortran... I assume your package also highlights older versions?
> 
> Anyway, this looks good. Better go and tell people it's here ;) Thanks!

Thanks! Currently, the syntax is intended to highlight Fortran code starting with Fortran 90 ("modern Fortran") because this was where we saw the most need for an improved syntax. It highlights some parts of Fortran 77, but not all. Maybe we will add support for older versions at some point. Could be useful if the existing package is no longer being maintained.

PR #8158, opened 2021-01-13
added new repository for pomodoro timer

debottamKundu:
This package implements a Pomodoro Timer for Sublime Text and enables the user to set up to work in cycles without an intrusive interface. I was looking for a similar package on Sublime repositories and had found this one [package](https://packagecontrol.io/packages/Naetech%20-%20Pomodoro%20Timer%20Plug-in) but the links are all broken.

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -1820,6 +1820,17 @@
 			]
 		},
 		{
+			"name":"PomodoroTimer",
+			"details":"https://github.com/debottamKundu/pomodoro_timer",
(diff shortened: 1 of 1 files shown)


braver:
How about [this one then](https://packagecontrol.io/packages/Pomodoro)?

debottamKundu:
I somehow missed that. thanks!

braver:
Eh, yeah, so I think we won't merge this one then.

PR #8160, opened 2021-01-14
Add Forth

mitranim:
Should be self-explanatory. There doesn't seem to be a Forth syntax on Package Control, there will be now. 🙂


braver:
Again with the "unlicense" ;) Well, whatever, the file is there, let's not let the bot stop us. 

You do probably want to add a "language syntax" label here though.

PR #8162, opened 2021-01-14
New package: FilePrompt

loquens:
Package allows opening and saving files without dialogs using Linux-shell-styled prompt with autocompletion and suggestions

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -692,7 +692,6 @@
 			"name": "FilePrompt",
 			"details": "https://github.com/loquens/subl-open-file-prompt",
 			"labels": ["file navigation", "file manipulation", "utilities", "open files", "file utils", "file suggestions"],
-			"readme": "https://raw.githubusercontent.com/loquens/subl-open-file-prompt/main/README.md",
 			"releases": [
 				{
 					"sublime_text": "*",
--- a/repository/f.json
+++ b/repository/f.json
@@ -692,6 +692,7 @@
 			"name": "FilePrompt",
 			"details": "https://github.com/loquens/subl-open-file-prompt",
 			"labels": ["file navigation", "file manipulation", "utilities"],
+			"readme": "https://raw.githubusercontent.com/loquens/subl-open-file-prompt/main/README.md",
 			"releases": [
 				{
 					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


braver:
Thanks for submitting your package. 

It looks like you're providing commands. It's recommended to always add them to the command palette via a Default.sublime-commands file ([example](https://github.com/SublimeLinter/SublimeLinter/blob/master/Default.sublime-commands)). You can also make it easier for users to add the suggested keybindings, by creating a [Main Menu entry](https://github.com/SublimeLinter/SublimeLinter/blob/master/Main.sublime-menu#L24).

How does this compare to existing packages like [FileDialog](https://packagecontrol.io/packages/FileDialog)?

loquens:
> Thanks for submitting your package.
> 
> It looks like you're providing commands. It's recommended to always add them to the command palette via a Default.sublime-commands file ([example](https://github.com/SublimeLinter/SublimeLinter/blob/master/Default.sublime-commands)). 
Fixed.

> You can also make it easier for users to add the suggested keybindings, by creating a [Main Menu entry](https://github.com/SublimeLinter/SublimeLinter/blob/master/Main.sublime-menu#L24).
Fixed.
Thank you for reviewing.

> How does this compare to existing packages like [FileDialog](https://packagecontrol.io/packages/FileDialog)?
FilePrompt is much simpler, provides only one thing (but customizable), has no dependencies and just works. FileDialog is unsupported now and goes in much bigger package [sublime-enchanced](https://github.com/shagabutdinov/sublime-enhanced). And it is not working (at least on my system) after installation from Package Control.
I've tried to find other packages providing similar functions, but found nothing.

repository/f.json
```diff
@@ -688,6 +688,18 @@
 				}
 			]
 		},
+		{
+			"name": "FilePrompt",
+			"details": "https://github.com/loquens/subl-open-file-prompt",
+			"labels": ["file navigation", "file manipulation", "utilities", "open files", "file utils", "file suggestions"],
+			"readme": "https://raw.githubusercontent.com/loquens/subl-open-file-prompt/main/README.md",
```
braver, line 695: You don't need to provide this readme link. Everything is already found via the repo link in "details".

braver (changes requested):

repository/f.json
loquens: Fixed. Seen it in other packages, so decided to add it.

braver:
Alrighty 🙆🏻‍♂️

PR #8163, opened 2021-01-15
Add "Paste Plus" package

ggets:

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -535,6 +535,20 @@
 			]
 		},
 		{
+			"name": "Paste Plus",
+			"details": "https://github.com/ggets/Sublime_PastePlus",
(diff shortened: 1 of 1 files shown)


braver:
Thanks for submitting. Could you perhaps elaborate what this package adds and how that is something that's not already covered by existing packages? 
It looks like it does one thing (which is good!) but I didn't get _what_ that is from the name of the package (it sounds like a multi-featured thing around pasting stuff). Or is your intention to add more?

ggets:
I plan to add more functionality. Right now it's sole purpose is to provide paste-able URIs from copied files in Windows Explorer.

ggets:
The keyboard shortcuts are not replaced unless this behavior is allowed by configuration option.

braver:
Cool, thanks!

PR #8165, opened 2021-01-16
Add "SVG Viewer" package

YariKartoshe4ka:
[Similar package](https://github.com/chunqiuyiyu/sublime-svg-preview)
### Differences
- Viewing in new tab instead of previewing in popup
- Easy extensibility (easy adding new converter, without coding)
- Possibility of opening SVG as picture by default

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -5300,6 +5300,17 @@
 			]
 		},
 		{
+			"name": "SVG Viewer",
+			"details": "https://github.com/YariKartoshe4ka/sublime-svg-viewer",
+			"labels": ["svg", "view", "image"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "SVG-Snippets",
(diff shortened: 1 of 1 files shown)


braver:
Would it be possible to add these features as alternative modes or features to the existing package? It's a pretty new package and you can perhaps contribute to it. It's not a problem to have competing packages with the same utility if the approach is really different, but I'd like to at least try and consolidate these ideas.

YariKartoshe4ka:
I think it is [impossible](https://github.com/chunqiuyiyu/sublime-svg-preview/issues/4). But in the future (when sublime changes the dependency policy and any PyPI package can be easily added) I'm going to change the approach

wbond:
> when sublime changes the dependency policy and any PyPI package can be easily added

This is unlikely to happen since no one wants to take on the massive responsibility of implementing such a feature of figuring out how to get N different set of dependencies to work together.

braver:
> I think it is impossible

Well, that works for me. At least we've considered it.

PR #8166, opened 2021-01-19
Add Theme One

AmjadHD:


braver:
It might be good to clarify in the README that this package is both a color scheme and a theme. Presumably the user needs to select both for the "full experience". I guess we have some [similar packages](https://packagecontrol.io/packages/Theme%20-%20TwoDark) but none that so elegantly provide both light and dark modes.  So yeah, good job, let's ⛵

PR #8167, opened 2021-01-19
Add "Run As You Type" package

ggets:

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -2306,6 +2306,20 @@
 			]
 		},
 		{
+			"name": "Run As You Type",
+			"details": "https://github.com/ggets/Sublime_RunAsYouType",
+			"readme": "https://raw.githubusercontent.com/ggets/Sublime_RunAsYouType/master/README.md",
+			"labels": ["exec", "interactive", "execution", "command", "command execution", "cmd"],
+			"author": "GGets",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"platforms": ["windows"],
+					"tags": true
+				}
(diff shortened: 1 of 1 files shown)


braver:
You been busy huh? :D 

This sounds a bit familiar. We seem to have some packages that do something similar, like [ToolRunner](https://packagecontrol.io/packages/ToolRunner). Some maybe a bit old like [Code Runner](https://packagecontrol.io/packages/Code%20Runner). Maybe [SublimeREPL](https://packagecontrol.io/packages/SublimeREPL) is in the same vein as well. How does your package relate to these?

ggets:
This is actually an old plugin of mine that I never published. I wrote it, because I couldn't find an existing one with the same functionality.
I will look into each of these plugins to see how similar they actually are.
Basically the way it works is it creates a named output panel ( like the console ) and it evaluates there. You can also pass arguments to the executing program on the fly by the means of an input panel. Do these work like that?

braver:
> Do these work like that?

Seriously, I don't have a clue. It just looks very similar. What I'm looking for is when someone submits a new package, that in the submission the similar packages are at least considered, so that we don't duplicate where a contribution/PR to the other package would also have worked. I don't mind duplication per se, it also facilitates replacement via natural selection: if yours is better, or better maintained, it can overtake the older package that way. So in a way it's on the submitter to "defend" the addition in face of what's already out there. In this case there are probably a couple valid approaches to the same idea, and the existing packages are pretty old, so good chance yours can find its way among them. Let me know what you find when you have a look at the competitors :)

PR #8169, opened 2021-01-21
Theme - One Separate releases for ST3 and ST4

AmjadHD:


AmjadHD:
@braver, if you can merge this now, I would appreciate it. The theme is kind of broken. and I would like it being fixed sooner rather than later.

PR #8172, opened 2021-01-22
add plugin

evercyan:

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -4895,17 +4895,6 @@
 			]
 		},
 		{
-			"name": "subryan",
-			"details": "https://github.com/evercyan/subryan",
-			"releases": [
-				{
-					"sublime_text": "*",
-					"platforms": ["osx"],
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "SubTexting",
 			"details": "https://github.com/willbrazil/SubTexting",
 			"releases": [
--- a/repository/t.json
+++ b/repository/t.json
@@ -2795,6 +2795,17 @@
 			]
 		},
 		{
+			"name": "Tony Water",
+			"details": "https://github.com/evercyan/tony-water",
+			"labels": ["json", "utils"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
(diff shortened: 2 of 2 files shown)


braver:
Hi!

Can you perhaps explain what your package does, and why Tony Water is a good name for it? Not just for me, but also to help others find your package. 

It looks like a collection of string and json manipulation utilities. Perhaps you can elaborate on what your package adds to what is already out there in package control? You already have 2 abandoned packages in package control, maybe it's a good idea to replace one of them with this new one. Also, it would be nice if this one doesn't go missing after a while again...

As for the package itself. You create a context menu. We also recommend adding all commands to the command palette, via a Default.sublime-commands file. This helps users look for and discover commands, and it also enables triggering the commands without reaching for your mouse/trackpad.

evercyan:
> Hi!
> 
> Can you perhaps explain what your package does, and why Tony Water is a good name for it? Not just for me, but also to help others find your package.
> 
> It looks like a collection of string and json manipulation utilities. Perhaps you can elaborate on what your package adds to what is already out there in package control? You already have 2 abandoned packages in package control, maybe it's a good idea to replace one of them with this new one. Also, it would be nice if this one doesn't go missing after a while again...
> 
> As for the package itself. You create a context menu. We also recommend adding all commands to the command palette, via a Default.sublime-commands file. This helps users look for and discover commands, and it also enables triggering the commands without reaching for your mouse/trackpad.

Thanks for ur reply. 
"Tony Water" means "拖泥带水" which is a chinese idiom transliteration.
It's a good name for me because this plugin is a collection of my small tools, but not for a single use sence.
About my 2 abandoned packages, I have no idea how to remove them.

braver:
Thanks! It looks like it makes sense to replace both your existing packages with this new one via a "previous_names" key. That would reconcile the existing packages and move any existing users over to the new one.

evercyan:
> Thanks! It looks like it makes sense to replace both your existing packages with this new one via a "previous_names" key. That would reconcile the existing packages and move any existing users over to the new one.

Thanks~ It's done.

PR #8173, opened 2021-01-23
Update EditorConfig to use tagged releases

sindresorhus:


braver:
Branch based releases are deprecated. Now would be a good time to switch to tagged releases.

sindresorhus:
I prefer to keep using branch releases. I don't foresee many updates for any of these.

braver:
Well deprecated here means you will not be able to do branch based releases at some point in the future. We’re not going to accept changes to different branches. If you don’t expect many updates, certainly creating a single tag is also not that big of a problem.

sindresorhus:
Done

braver:
🙏🏻

PR #8174, opened 2021-01-23
Add syntax mode to colour the CMHG used for building modules in RISC OS.

gerph:
RISC OS CMHG files are descriptions of the header that will be created
for modules written in C. They are a core part of the development
environment used by RISC OS.

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -1584,6 +1584,17 @@
 			]
 		},
 		{
+			"name": "RISC OS CMHG Syntax",
+			"details": "https://github.com/gerph/sublimetext-riscoscmhg-syntax",
+			"labels": ["syntax"],
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "RISC OS Command Syntax",
 			"details": "https://github.com/gerph/sublimetext-riscoscommand-syntax",
 			"labels": ["syntax"],


braver:
Wouldn't it be nicer to have a single RISC OS syntax package that has all of these syntaxes? Now it feels like a RISC OS users would need to hunt down several packages to get all the highlighting.

gerph:
That's a pretty reasonable question and the reason that I didn't do it that way was that I'd already created the Commands syntax, and at that time didn't want to make it a generic 'RISC OS' syntax because it wasn't, and when I created this one I didn't want to go around renaming the old package or creating a new one that was a superset. There's also the fact that I'm (possibly) not the only person that would create syntax modes for RISC OS, so creating a single package that had the name (say) 'RISC OS Syntaxes' would be presumptuous - and untrue, because it would only be a small sampling of them that I felt were useful. Additionally, having choice where users can decide which syntaxes they wish to install is better - if someone doesn't want/need a given syntax, they don't.

So that's kinda my reason why I've done it this way.

The RISC OS Commands module is akin to a bash or DOS shell syntax mode, whereas the CMHG mode doesn't really have an equivalent; it's a description of how you build a given module and it's metadata. I guess you might, if you were generous, describe it as similar to the package metadata for debian packages. It's probably closer to the definition of DLL in Windows, which I vaguely remember.

Would you want to put those metadata syntaxes in the same package as bash or DOS? They're quite different purposes, and aimed at different types of users.

I can see the argument for putting them into the same package, more so when I get around to submitting the BBC BASIC syntax mode later (except of course BBC BASIC isn't specific to RISC OS).

As for 'hunting them down', you'd type in 'RISC OS' in package control's installation or search boxes, and it'd find them so that's less of a concern I think. The search tools somewhat make that issue unimportant :-)

So on balance, I think it's better to have them separate and allow the users the option. For the sake of (at the present time) 2 packages for a system, when there are 10s of packages for Python (to take an example of a language that has many packages for related operations), I don't think there's really that much benefit in merging them.

braver:
Well thank you for really considering my question and answering it so completely. I guess my question is partly fueled my unfamiliarity and now I can see how you would want to keep them separate. So, let's ship this thing!

gerph:
Thank you for understanding.

It's nice to be able to explain and justify a reasoning, even to myself. And especially useful to have the assumptions questioned by someone outside my experience who has the bigger picture of ST in mind. Thank you :-)

braver:
That’s what we’re here for ;) 😄

PR #8175, opened 2021-01-23
Add "devnotes" package

devon-badman:
The core of this package is that it allows you to link between different notes files. These links can be in comments in your code or in other notes files.

My plan is to expand functionality to include syntax and autocompletion, but it currently works as is with existing implementations of these such as [Notes](https://packagecontrol.io/packages/Notes)

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -703,6 +703,17 @@
 			]
 		},
 		{
+			"name": "devnotes",
+			"details": "https://github.com/devon-badman/devnotes",
+			"labels": ["notes", "links"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


braver:
Sorry for not getting around to reviewing this earlier. I guess I don't really understand the utility here, why and or how one would use it. The README doesn't clarify that for me. In any case, I have some questions.
Can you explain why this should exist as something outside of those other packages that already provide rich note taking features? It looks like you might want to propose this as a new feature in those packages.
What does a link to a note look like, especially in non-note files? Since the notes are linked to each other (?) and to code, wouldn't it make sense to always store them relative to a project root and not a hard absolute path? 
You now demand an extension, while sometimes notes are stored without extension. Also, why not simply work with file paths that you don't need to add extensions to?

braver:
Please respond to the questions to continue this review.

PR #8176, opened 2021-01-23
Update Autoprefixer to use tagged releases

sindresorhus:


braver:
Please see my feedback here: https://github.com/wbond/package_control_channel/pull/8173

braver:
Thanks!

PR #8179, opened 2021-01-25
Add TinyCodeCounter

gkpotter:
Add TinyCode Counter

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -2340,6 +2340,18 @@
 			]
 		},
 		{
+			"name": "TinyCodeCounter",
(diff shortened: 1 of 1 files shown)


braver:
To make it easier for users to find the settings for your package, please make a [main menu entry](https://github.com/SublimeLinter/SublimeLinter/blob/master/Main.sublime-menu#L16) and perhaps also an addition to the [command palette](https://github.com/SublimeLinter/SublimeLinter/blob/master/Default.sublime-commands#L38).

gkpotter:
Just added both! I really appreciated the links.

braver:
Excellent!

PR #8180, opened 2021-01-26
Added Frequency Package

krishraghuram:
This package adds a command that allows user to count frequency of each line in the current document.

For example, running the command on,

```
foo
foo
foo
bar
bar
```

will open a new scratch file with the content,

```
{
    "bar": 2, 
    "foo": 3
}
```
I'm also planning to add more features in future (sorted output, word frequency, arbitrary separators(for csv/tsv), output format choice between csv/json etc.)

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -1692,6 +1692,16 @@
 			]
 		},
 		{
+			"name": "Frequency",
+			"details": "https://github.com/krishraghuram/SublimeFrequency",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Freya",
 			"details": "https://github.com/mistydemeo/sublime_freya",
 			"releases": [


Ultra-Instinct-05:
Just a heads up. This
https://github.com/krishraghuram/SublimeFrequency/blob/master/SublimeFrequency.py#L47
will work in ST3, but from ST4 onwards, the `JSON.sublime-syntax` is moved to it's own folder in the `Default` repo. It will be `JSON/JSON.sublime-syntax`.

Ultra-Instinct-05:
> Now I haven't tested my extension for ST2. Also, I just recently got to know about the existence of ST4, and haven't tested on that either. I'm thinking of adding support for ST4 later...

Yeah, I meant you don't have to worry now, but it's something to keep in mind.

> insert does not work with newlines...

I see, now. I am not aware how to ignore to achieve that with `insert` so I guess a dummy `insert` command was needed.

krishraghuram:
Hi, do I have to do anything else for this to be approved/merged :)

krishraghuram:
Kind request for someone to take a look at this PR :)

krishraghuram:
@braver would you be able to merge this PR, or let me know what needs change?

braver:
> How can I specify that this is currently only for ST3 (and not ST2 or ST4)

You can simply ignore ST2, nobody uses that and most packages have dropped support for it without updating their package control entry. So just say < 4000 if you don't want to support ST4. I'm pretty sure there is a syntax to say "between 3000 and 400" but can't find an example right now. 

But you're overthinking it. ST4 is almost completely backwards compatible, and since it's not officially out yet you don't really need to think about your support for it. If for some reason your package doesn't work in ST4 you can always do something about it later. So for now just tag a 1.0.0 and set your entry here to "tags: true". A simple package like yours shouldn't have to worry about all this.

I'm not sure why you're including a packages.json file in your repo? It's not needed for local development, and definitely when distributing through package control.

krishraghuram:
> So for now just tag a 1.0.0 and set your entry here to "tags: true". A simple package like yours shouldn't have to worry about all this.

Done :)

> I'm not sure why you're including a packages.json file in your repo? It's not needed for local development, and definitely when distributing through package control.

I made a mistake while reading https://packagecontrol.io/docs/submitting_a_package and added this file. Removed it now.

braver:
Cool, let's ship this thing! ⛵

PR #8182, opened 2021-01-29
add AMPscript Syntax Highlighter

wvpv:
I'd like to submit my own AMPscript syntax highlighter mainly because the others in the repository have not been updated in years and are incomplete.  My version is completely new, written based on my work co-authoring [The AMPscript Guide](https://ampscript.guide).  Thanks for your consideration!

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -802,18 +802,8 @@
 			]
 		},
 		{
-			"name": "AmpScript Highlighter",
-			"details": "https://github.com/seanja/ampscript-st2",
-			"labels": ["language syntax"],
-			"releases": [
-				{
-					"sublime_text": "*",
-					"branch": "master"
-				}
-			]
-		},
-		{
 			"name": "AMPscript Syntax Highlighter",
+			"previous_names": ["AmpScript Highlighter"],
 			"details": "https://github.com/wvpv/ampscript-syntax-highlighter",
 			"labels": ["language syntax"],
 			"releases": [
@@ -822,7 +812,8 @@
 					"tags": true
 				}
 			]
-		},		{
+		},
+		{
 			"name": "AmxxEditor",
 			"details": "https://github.com/evandrocoan/AmxxEditor",
 			"labels": ["autocompletion", "auto-complete", "amxmodx", "amx mod x", "amxx", "pawn", "build system"],
--- a/repository/a.json
+++ b/repository/a.json
@@ -819,7 +819,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"branch": "master"
+					"tags": true
 				}
 			]
 		},		{
--- a/repository/a.json
+++ b/repository/a.json
@@ -792,18 +792,6 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


braver:
You're right, the existing packages seem outdated. Instead of adding a third, I would prefer replacing one of them. Either is fine I guess, maybe this is the best candidate: https://packagecontrol.io/packages/AmpScript%20Highlighter

Could you try opening a PR on either of the existing repositories? Best case scenario would be for your contribution to be merged. Maybe you could become the new maintainer of the existing repo. If all of that fails (let's give the attempt a week or two) we can replace an entry with your repo. That would move all existing users over to the new an better package, and keeps the ecosystem clean.

wvpv:
I can do that.  I've reached out to SeanJA with a pull request.  I'll report back if he responds.  Thank you!

wvpv:
It's been two weeks and I have received no response from SeanJA regarding my pull request.

braver:
@SeanJA we would like to update your ampscript syntax highlighting package for Sublime Text. Or failing that we intend to replace your package with something new. Please respond to this PR to let us know how you would prefer we handle this, or whether you object to having your package replaced.

braver:
@wvpv if we don’t get a response here either, I think the best thing to do is to replace the existing package with yours. We need to give the original author a week or two again though.

wvpv:
@braver It's been a couple of weeks.  What's next?

braver:
The most simple approach is to simply [replace the details](https://github.com/wbond/package_control_channel/pull/8194/files) of the existing entry to point at your repo. Make sure the latest release in your repo is newer than the old one. That will ensure package control will pick up on the change and propagate it to all users.

wvpv:
Everything's up-to-date in my repo: https://github.com/wvpv/ampscript-syntax-highlighter

wvpv:
So I need to remove my entry in a.json and update the git URL on "AmpScript Highlighter"?

braver:
Just to make sure everything works correctly, you might want to tag a higher version than the [1.0.3](https://github.com/SeanJA/ampscript-st2/releases/tag/v1.0.3) that the original repo is at.

> So I need to remove my entry in a.json and update the git URL on "AmpScript Highlighter"?

Exactly. And also change it from branch based releases to tags.

wvpv:
Can I adjust the title as part of my updates or does it need to stay "Ampscript Highlighter"?

braver:
You can, but the name acts as a key or identifier of the package. So to handle the rename properly you need to provide the old name in via ["previous_names"](https://packagecontrol.io/docs/renaming_a_package#Step_3). This will ensure all existing users are transparently moved over to the new package with the new name.

wvpv:
Alright, I think I did the right thing.

braver:
Jep, looks good 👍🏻

SeanJA:
Whoops... not abandoned, just not really using it anymore so I wasn't paying attention

SeanJA:
Should probably also consider kicking the other one out of the package thing though, it was in worse shape than mine

PR #8183, opened 2021-01-30
Add BBC BASIC syntax mode.

gerph:
BBC BASIC syntax mode is intended for use with the variants of BBC BASIC,
although it is specific to the BASIC V/VI variants.

It should work with...

  * BASIC versions up to 4 for the Electron, BBC Micro and BBC Master
    series (a subset of the keywords available in the syntax),
    usable with https://bbcmic.ro/
  * BBC BASIC for RISC OS.
  * Brandy BASIC (and variants such as Matrix Brandy Basic)
  * BBC BASIC for Windows (with some differences as its syntax has
    been extended).

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -268,6 +268,17 @@
 			]
 		},
 		{
+			"name": "BBC BASIC Syntax",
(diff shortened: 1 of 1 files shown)


braver:
Looks good, thanks!

PR #8184, opened 2021-01-31
add Custom Highlighter

johanrosenson:
Simple package to provide custom configurable color highlighting of any keywords you prefer, I use this plugin to add color highlighning to different css frameworks (in tailwind, example: class red-500, blue-300 and so on) without needing a plugin specifically for that framework.

This is not a replacement for the existing plugin Color Highlight, but more of a companion, i use my plugin and Color Highlight together with great experience.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -4690,6 +4690,17 @@
 			]
 		},
 		{
+			"name": "Custom Highlighter",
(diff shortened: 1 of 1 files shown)


johanrosenson:
Closhing this, I will re-submit soon from different repo

PR #8185, opened 2021-01-31
Add Custom Highlighter

johanrosenson:
This plugin will let the user configure keywords that should be highlighted by a configured color.
It has similarities to plugins such as Color Highlight (which this is a fork of), but is much more limited in scope, on purpose.

I have made sure that this plugin are not overlapping in functionality with other similar plugins, but instead can complement each other.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -4690,6 +4690,17 @@
 			]
 		},
 		{
+			"name": "Custom Highlighter",
+			"labels": ["color", "highlight", "highlighter"],
+			"details": "https://github.com/devlop-ab/custom-highlighter",
+			"releases": [
+				{
+					"sublime_text": ">=3116",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Custom Insert",
 			"details": "https://github.com/yanni4night/sublime-custominsert",
 			"releases": [
--- a/repository/c.json
+++ b/repository/c.json
@@ -4692,7 +4692,7 @@
 		{
 			"name": "Custom Highlighter",
 			"labels": ["color", "highlight", "highlighter"],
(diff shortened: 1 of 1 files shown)


braver:
> Color Highlight (which this is a fork of)

Ha, which borrows a lot from SublimeLinter 2, hence all the references to linters everywhere in the code while this has nothing to do with linting. Anyway, it's a practical approach. Does it still create a copy of a user's color scheme? If so, it it compatible with other packages that do the same (e.g. Color Highlight)?

johanrosenson:
@braver 
> Does it still create a copy of a user's color scheme? 
Yes

> If so, it it compatible with other packages that do the same (e.g. Color Highlight)?
Based on my initial testing i would say it is compatible, but after a longer period of testing i am not 100% sure, to be honest it has been working pretty good, but i have experienced some minor glitches (that i have not had time to look into yet), and they might be because of compatibility issues, or they are due to some other completely unrelated issue, hard to say at this moment.

Anyway, i will close this PR and put this whole plugin on hold until ST4 is released, i don't have the proper time to fine tune it at the moment and i would love to clean up the forked code even more.

braver:
No problem, you can always come back again.

What you want do achieve is hard to do without patching color schemes, but automating that is hard. Especially if you want to consider other packages that take the same approach. I’m not even sure it can be done reliably, although nowadays color schemes can be extended in a way that hasn’t been considered by many of those older packages. Alternatively, you might want to look at phantoms to place visual, colored, markers in code in a way that doesn’t rely on the syntax highlighting mechanism. Color Helper and Color Hints are modern packages you might be able to borrow some ideas from.

PR #8187, opened 2021-02-07
Add SuperSettings

TobyGiacometti:
SuperSettings enables per-directory configuration of Sublime Text. A similar package called [Directory Settings](https://packagecontrol.io/packages/Directory%20Settings) already exists. However, it hasn't been updated in 7 years. An [issue](https://github.com/willbeaufoy/DirectorySettings/issues/4) I consider a deal breaker has been open for 5 years. In addition, SuperSettings has an important feature missing in Directory Settings: per-directory configuration by syntax.

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -875,16 +875,6 @@
 			]
 		},
 		{
-			"name": "Directory Settings",
-			"details": "https://github.com/willbeaufoy/DirectorySettings",
-			"releases": [
-				{
-					"sublime_text": "*",
-					"branch": "master"
-				}
-			]
-		},
-		{
 			"name": "dired",
 			"details": "https://github.com/kublaios/dired",
 			"releases": [
--- a/repository/s.json
+++ b/repository/s.json
@@ -5169,6 +5169,16 @@
 			]
 		},
 		{
+			"name": "SuperSettings",
+			"details": "https://github.com/TobyGiacometti/SublimeSuperSettings",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Suricate",
 			"details": "https://github.com/nsubiron/SublimeSuricate",
 			"labels": [
--- a/repository/s.json
+++ b/repository/s.json
@@ -5171,6 +5171,7 @@
 		{
 			"name": "SuperSettings",
 			"details": "https://github.com/TobyGiacometti/SublimeSuperSettings",
+			"previous_names": ["Directory Settings"],
 			"releases": [
 				{
 					"sublime_text": ">=3000",


braver:
Hm, that repo seems pretty much dead, although @davepeck is still active on GitHub. If "SuperSettings" is a drop-in replacement for "Directory Settings" (sure sounds like it), we have 3 options:

1. You make a PR on the existing package and that get's merged and released
2. The original author might even invite you to maintain that repo
3. Failing all that we replace the existing package with yours

The first one is always a good place to start. If that fails to work out in a week or 2, we'll give the original authors and maintainers a notice that we intend to move to option 3 here and again wait 2 weeks.

TobyGiacometti:
Thanks for your feedback @braver! SuperSettings is close to being a drop-in replacement, but as it stands, it is not (due to some cosmetic differences). To be honest, I do not want to invest time into porting changes to Directory Settings, especially if they won't get merged (@willbeaufoy hasn't actually worked on Directory Settings after taking over maintenance 5 years ago). I would be happy to contact @willbeaufoy to inquire whether he would mind if Directory Settings is replaced. How does that sound to you?

braver:
We try not to have multiple packages that target the same use case unless the approach is significantly different. After several years it’s ok to break settings or make cosmetic changes, especially if it comes with fixes and maintenance in the (near) future. So a replacement sounds like a good idea here. Let’s wait for a week or two to see if the current maintainers object.

TobyGiacometti:
> We try not to have multiple packages that target the same use case unless the approach is significantly different.

Makes a lot of sense.

> After several years it’s ok to break settings or make cosmetic changes, especially if it comes with fixes and maintenance in the (near) future.

Agreed.

> So a replacement sounds like a good idea here. Let’s wait for a week or two to see if the current maintainers object.

Just to clarify: Should I contact the maintainers or has that already been done?

braver:
The mentions in this thread should be enough of a trigger. But it wouldn't hurt if you could to open an issue in the repo.

willbeaufoy:
Hi, yes I don't have any plans to work on Directory Settings any more, so sounds like a good idea to replace it.

TobyGiacometti:
Thanks for your feedback @willbeaufoy!

braver:
@TobyGiacometti easiest would perhaps be to let the existing entry point to your repo, and switch to tag based releases. Let me know if you need any help.

TobyGiacometti:
Sure, I can do that. What exactly will happen with the existing entry once done? And what happens to the users that have Directory Settings installed?

braver:
Next time package control makes its pass over all package repos it will find, in your repo, a version that’s newer than it already had for “Directory Settings”. It’ll update the indexes etc. and all users will receive the new version next time they request updates or restart ST. If there are significant changes you can tell these users about that via the message system.

TobyGiacometti:
Thank you for the helpful explanation, good to know! 2 questions remain:

- Will Directory Settings still exist as a separate entry in the package listing inside Sublime Text or on packagecontrol.io?
- How does the comparison for versions inside `message.json` work. SuperSettings is currently at version 1.0.0, Directory Settings at 2.0.0. If I create a message for version 1.0.0, will that show when users upgrade from Directory Settings to SuperSettings? Or do I have to tag SuperSettings with a version higher than 2.0.0?

braver:
Ah, sorry, I wasn't totally clear. 

There are two ways to do this:

1. Revert the changes proposed in this PR. Instead change the existing Directory Settings entry so that "details" point to your repo, and instead of `branch: master`, you set it to `tags: true`. That way it will see any tag in your repo as newer than any pre-existing version of Directory Settings. Because Directory Settings is on branch based releases whatever they do with tags or releases is invisible to package control. If you switch it to tags, and create a 1.0.0 tag, that will automatically count as a newer version and it will trigger the message. 
2. Remove the existing entry, and in your new entry for SuperSettings add a key: `previous_names: ["Directory Settings"]`. This tells package control that SuperSettings is the same thing as Directory Settings, under a new name. It works the same here with regards to tags/versions.

There is no tangible reason to pick one approach over the other, your pick.

TobyGiacometti:
Thank you very much for your help! I have made all the necessary adjustments...

braver:
Excellent!

TobyGiacometti:
Thanks for merging!

PR #8191, opened 2021-02-11
Added entries for Maya and 3ds Max Debug Adapters

Frenchman98:
These two packages plug into the [Debugger plugin](https://github.com/daveleroy/sublime_debugger/releases), extending its functionality based on the users' needs.

- The [3ds Max plugin](https://github.com/blurstudio/sublime_debugger-3dsMax) allows users to debug Python 2 code running in Autodesk 3ds Max
- The [Maya plugin](https://github.com/blurstudio/sublime_debugger-maya) allows users to debug Python 2 code running in Autodesk Maya

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -380,6 +380,27 @@
 					"tags": true
 				}
 			]
+		},{
+			"name": "Debugger - 3ds Max",
+			"details": "https://github.com/blurstudio/sublime_debugger-3dsMax",
+			"releases": [
+				{
+					"sublime_text": "<4000",
+					"platforms": ["windows"],
+					"tags": "st3-"
+				}
+			]
+		},
+		{
+			"name": "Debugger - Maya",
+			"details": "https://github.com/blurstudio/sublime_debugger-maya",
+			"releases": [
+				{
+					"sublime_text": "<4000",
+					"platforms": ["windows"],
+					"tags": "st3-"
+				}
+			]
 		},
 		{
 			"name": "Decent Color Scheme",
--- a/repository/d.json
+++ b/repository/d.json
@@ -389,6 +389,11 @@
 					"sublime_text": "<4000",
 					"platforms": ["windows"],
 					"tags": "st3-"
+				},
+				{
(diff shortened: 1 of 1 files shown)


braver:
Seems like these are the first plugins for the debugger package in package control. Seems like a good idea to involve @daveleroy here for review. If we can expect many more plugins it might be beneficial to set up a separate channel. Such a channel can be maintained independent of the main channel. And plugins can actually be reviewed (and tested) by someone with knowledge of the main package. Ie. like what has been done for SublimeLinter and LSP.

Frenchman98:
That's a good point. We currently have 2 more adapters being developed, and possibly more in the future, along with a [template](https://github.com/blurstudio/sublime_debugger-plugin) if others want to make their own adapters too, so a separate channel like that might be best.

Frenchman98:
Hey @braver, what should we do if daveleroy doesn't respond? They seem to be actively developing for st4 but not so much for st3 anymore

braver:
As everyone should, ST4 is pretty awesome.

Anyway, we should just publish the plugins here, nothing we can’t change our minds about later. I’ll have a final look tomorrow.

Frenchman98:
It does seem pretty cool! I unfortunately don't have access but I'm looking forward to it releasing. Sounds good though, I'll finalize some changes and tag them today

braver:
Ok, that looks good. I would recommend adding a releases entry for ST4 (e.g. simply tags: true), so you don't need to open a PR here again when you start supporting ST4 in the future.

Frenchman98:
Oh good point, I'll do that right now. I'll add a separate entry under "releases" like they did for the debugger because I believe they changed the debugger's architecture in the ST4 version, which might cause incompatibilities

Frenchman98:
Oh dang, I forgot I need releases for the tests to pass. I'll leave this PR open then and start working on making releases compatible with the ST4 version of the debugger

braver:
Oh, right. I forgot about that. We can merge just the ST3 releases now if you want?

PR #8194, opened 2021-02-16
Takes over ElixirFormatter with new repo

jmnsf:
A fork of [ElixirFormatter](https://github.com/karolsluszniak/elixir-formatter-sublime). That package has some issues, is unmaintained, and is [not accepting fixes](https://github.com/karolsluszniak/elixir-formatter-sublime/pull/5). So I pulled that fixed code, added some new features, and packaged it up again.

It's simply an auto-formatter for elixir code, using the language's official formatter.

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -593,7 +593,8 @@
 		},
 		{
 			"name": "ElixirFormatter",
-			"details": "https://github.com/karolsluszniak/elixir-formatter-sublime",
+			"details": "https://github.com/myskoach/sublime-elixir-formatter",
+			"labels": ["elixir", "formatting", "mix"],
 			"releases": [
 				{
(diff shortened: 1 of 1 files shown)


braver:
The original package is indeed well and truly dead. Our policy in such cases is, if a new package with the same goal and overall approach is submitted, to replace the old package entry with the new one. That will move existing users over to the new package, and ensures the repository stays fresh. 

You can simply change the repo url in the details, and add the new labels. Then, also make sure the newest tag in your repo is a higher version number that the old package. If you need to notify users about maybe changes to settings etc. you can use the messages feature. 

Let me know if you need any help or clarification.

jmnsf:
@braver thanks for the feedback, sounds reasonable. All done.

braver:
Excellent!

PR #8195, opened 2021-02-20
feat: add TailwindCSS Highlighter repository

mauroreisvieira:


deathaxe:
You've been faster. 

I've created a repo at https://github.com/SublimeText/TailwindCSS with some syntax tests added.

mauroreisvieira:
@deathaxe I didn't know I was going to do that, in that case, I will delete my repo. It makes more sense to stay yours since all the code was produced by you.

PR #8196, opened 2021-02-21
Add Tailwind CSS syntax definition

deathaxe:
This package adds Tailwind CSS syntax, which extends ST's default CSS.

It therefore requires ST 4 with CSS sublime-syntax version 2.

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -278,6 +278,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


braver:
Alright 👍🏻 

One minor thing: you might want to "export-ignore" the assets directory as well.

PR #8197, opened 2021-02-21
Add reStructuredText Improved

adamchainz:
This package has gone missing since the Bitbucket removal of Mercurial repositories. It was in the main repository until moving to the author's custom repository in cc98328db6614adb5d96bb668841203ca9a3e83a, but that custom repository has also gone down. I've revived it from my installed copy into [a GitHub repository](https://github.com/adamchainz/sublime-rst-improved).

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -1444,6 +1444,17 @@
 			]
 		},
 		{
+			"name": "RestructuredText Improved",
+			"details": "https://github.com/adamchainz/sublime-rst-improved",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Retro Computers Color Schemes",
 			"details": "https://github.com/idleberg/RetroComputers.tmTheme",
 			"labels": ["color scheme"],


tolomea:
+1

repository/r.json
```diff
@@ -1443,6 +1443,17 @@
 				}
 			]
 		},
+		{
+			"name": "Restructured Text Improved",
```
FichteFoll, line 1447: The original was named "RestructuredText Improved", without a space here. (I know because I have it installed still.) To allow older users to receive updates, I suggest re-using the old name here.

adamchainz: Indeed - well spotted. Fixed.

FichteFoll:
As a final note, you can use this opportunity to switch to a semver-based tagging schema, since date-based version numbers have lower precedence for PC. However, you cannot go back to a major 1 after using `2021.2.21`. Also, versions MUST NOT have leading zeroes. (That's a breaking change between semver 2.0.0-rc.2 and 2.0.0 iirc and our semver library still implements the rc.2 version and I'd rather not have people rely on that.)

adamchainz:
What version would you recommend then? 2.0.0?

FichteFoll:
I'd choose 1.0.0, since you didn't change anything from the base.

adamchainz:
Ah right, I just didn't understand from “However, you cannot go back to a major 1 after using 2021.2.21.”

I've pushed a 1.0.0 tag and deleted the 2021.2.21 tag.

adamchainz:
This will install on top of users who already have the old 2014.9.13 version, right?

FichteFoll:
Correct.

adamchainz:
Thanks for merging.

The listing page on Package Control now has the "Needs Review" label: https://packagecontrol.io/packages/RestructuredText%20Improved

Who can do that review?

FichteFoll:
As of now, only @wbond can do this.

PR #8198, opened 2021-02-21
Add Writer Color Scheme

tonsky:

### DIFF
--- a/repository/w.json
+++ b/repository/w.json
@@ -1043,6 +1043,18 @@
 			]
 		},
 		{
+			"name": "Writer Color Scheme",
+			"details": "https://github.com/tonsky/sublime-scheme-writer",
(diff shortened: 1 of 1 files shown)


tonsky:
BTW is there a way to exclude screenshots/specimens/tests from the Github repo when package is packaged? Something like `.gitignore`, but for sublime packages hosted on Github?

rwols:
Yes. Use a `.gitattributes` to define what you want to ignore while running `git archive` (which is what Package Control runs). Example: https://github.com/jisaacks/GitGutter/blob/master/.gitattributes

braver:
Pretty cool! FYI, you can also change color scheme in syntax specific settings, e.g. if you want to use this color scheme just for Markdown.

tonsky:
Thanks @braver! I realised it a little too late, but yes, it works too

PR #8199, opened 2021-02-22
Add Profile Switcher

tonsky:

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -2332,6 +2332,7 @@
 			"details": "https://github.com/tonsky/sublime-profiles",
 			"donate": "https://www.patreon.com/tonsky",
 			"labels": ["preferences"],
+			"previous_names": ["Switch Settings"],
 			"releases": [
 				{
 					"sublime_text": "*",
--- a/repository/p.json
+++ b/repository/p.json
@@ -2328,6 +2328,18 @@
 			]
 		},
 		{
+			"name": "Profile Switcher",
+			"details": "https://github.com/tonsky/sublime-profiles",
+			"donate": "https://www.patreon.com/tonsky",
+			"labels": ["preferences"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Project PHP ClassBrowser",
 			"details": "https://github.com/degami/ProjectPHPClassBrowser",
 			"releases": [
--- a/repository/s.json
+++ b/repository/s.json
@@ -5475,16 +5475,6 @@
 			]
 		},
 		{
-			"name": "Switch Settings",
-			"details": "https://github.com/t8m8/SwitchSettings",
-			"releases": [
-				{
-					"sublime_text": ">=3000",
-					"tags": true
-				}
-			]
-		},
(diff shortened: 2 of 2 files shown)


braver:
I kinda wonder how this compare with [Switch Settings](https://packagecontrol.io/packages/Switch%20Settings) (besides being a lot newer).

Besides that an older similar package already exists and might be replaced, I don't object merging this per se. However...

I also kinda wonder how much of this wouldn't be better handled using syntax specific setting. Assuming of course that most other modes are like "writing" and that writing mostly being in one syntax. In any case, personally, I'd be looking for an automatic switch like "use this profile for these languages" instead of manually switching modes every time. Which is essentially what syntax specific settings already does, except that it doesn't bundle settings for multiple profiles.

tonsky:
@braver You are right, my case could be handled by syntax-specific settings. I realized that after writing the whole thing, though :) So the question is: could this be potentially useful to someone besides that? I think yes, but can’t think of an use-case.

Re: Switch Settings, I just tried it. It’s exactly the same idea, yes. But I’d say mine is much smoother. E.g. Switch Settings has modal alerts and restarts ST every time you change profile. Mine is silent and invisible.

braver:
Well let’s see if we can update the existing package. Not a whole lot of users, but it would be nice to have a smooth new implementation of the same idea.

@t8m8 would you mind if we replace your Switch Settings package with this new package from @tonsky ?

tonsky:
One of the side effects of using syntax-specific settings for Markdown: in SM markdown diffs are in different font now. Not sure good or bar, certainly not what I’d expected

Screenshot 2021-03-02 at 01 53 27

braver:
Oh, that’s interesting. SM seems to take quite a lot of settings from ST, I’m always a bit surprised by how deep that seems to go.

tonsky:
> @t8m8 would you mind if we replace your Switch Settings package with this new package from @tonsky ?

It’s been a week since you reached out to @t8m8. Is it possible that we publish my package now and wait for the reply (if it ever comes) in the background?

braver:
Not really. The policy requires 2 weeks and its a whole lot of hassle to correct afterwards.

tonsky:
Ok, it’s now been two weeks. For the record, I don’t propose to remove @t8m8 package, I propose to add mine and keep his/hers too for now. And open a separate issue in which you can wait for the reply to remove it later

braver:
Nah, I’d rather just straight out replace the thing. We want to prevent users from having to weed through and perhaps try out different packages that do the same thing. @t8m8 has been active on GitHub these past few weeks, and not responded here. So there really is no reason to keep that package around.

t8m8:
Sorry for the late reply. Switch Settings I developed in the past is now completely out of development, so I am in favor of replacing it with @tonsky package. Thanks for contacting me, I appreciate it.

braver:
@tonsky can you update this PR so that it replaces the existing package with yours?

tonsky:
@braver done

rwols:
braver meant using the `previous_names` key.

tonsky:
@rwols done

tonsky:
@braver done

PR #8200, opened 2021-02-22
Add UCFHighlighter

LDprg:
UCFHighlighter is s syntax highlighter for xilinx ISE Ucf Files, which is used for the pinning of fpgas. Is very usefull for VHDL programmers, which are using Xilinx ISE.


braver:
Usually, the scopes end with the base scope name, e.g. `ucf` in this case. Why did you choose `z80` here?

You probably also want to add the "language syntax" label to make your package easier to find.

PR #8201, opened 2021-02-25
added metro colour scheme package

iJohnPaul:

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -1451,6 +1451,16 @@
 			]
 		},
 		{
+			"name": "Metro",
+			"details": "https://github.com/iJohnPaul/Metro",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


braver:
You probably want to add the "color scheme" labels to make your package easier to find.

You're also still using a tmTheme file. If you're only targeting Sublime Text, you'll want to convert that to a sublime-color-scheme file. That format supports variables and is a 1000x easier to maintain. You can do a conversion within Sublime Text via the Tools>Developer menu.

iJohnPaul:
Yeah, that makes sense. I have converted it now.

jrappen:
You probably want to complete the list of globals. The current docs give you a list:

<https://www.sublimetext.com/docs/color_schemes.html#global_settings:ver-dev>

braver:
Also, please add the "color scheme" labels to make your package easier to find.

braver:
Let me know what you think of that last bit of feedback and we'll re-open.

PR #8203, opened 2021-02-28
Update edge extension source repo

thetutlage:
The edge template engine and all its related repos have been moved to its own organisation http://github.com/edge-js and hence the URL update is required.

Also, I have re-written the extension from scratch to improve the syntax highlighting, add completions and add support for symbols.

- [x] Followed https://packagecontrol.io/docs/submitting_a_package#Step_7
- [x] Repo has a README file
- [x] Releases are tagged

There are chances, that I may have missed something unknowingly. So please provide the feedback and I will be more than happy to work on it 🙂

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -320,7 +320,8 @@
 					"sublime_text": ">=3092",
 					"tags": true
 				}
-			]
+			],
+			"previous_names": ["Edge Syntax Highliter"]
 		},
 		{
 			"name": "Edge Theme",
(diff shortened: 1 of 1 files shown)


braver:
Thanks! One thing: you need to provide a "previous_names" property for the rename to be handled correctly.

thetutlage:
@braver Done. Thanks for the feedback

braver:
🎉 all set!

PR #8204, opened 2021-03-01
QColor - Beta 1.0.0

raphaelquintao:
Awesome Sublime Color Highlighter, Converter and Picker


raphaelquintao:
This is my first pull request here, what i'm doing wrong? 
If someone can point it i appreciate.

braver:
Hi, thanks for submitting your package. We already have a bunch of color tools in the repository. One of which is [ColorHelper](https://packagecontrol.io/packages/ColorHelper). How does yours compare and what does it add to what's already available?

You can remove the "description" from your entry here, we'll just take it from the GitHub repo that the "details" point to.

raphaelquintao:
Mine is a lot simpler than that.  This one has a lot of configurations. 
Mine just works...
I thought any package would be welcome here...

rchl:
Not gonna 

All that trickery [with direct file path checking](https://github.com/raphaelquintao/QColor/blob/d5cbeac1f1e2394ac42f1731eeaeeb3ceb23d7b6/QColor.py#L49-L90) and `chmod`ing some files won't work with the released package as the released package is compressed and there is no direct access to your files from the filesystem (there are ways around that but it's not ideal). Also, the part where you reload stuff in `sys.modules` looks a bit scary.

> Mine is a lot simpler than that.

Is it though if it requires that the user installs some pip dependency in an external (to ST) environment?
And then potentially deal with some path issues and ST not finding that dependency?
I guess you've achieved some level of simplicity in your code but delegating functionality to an external dependency but that dependency can be problematic and it's quite heavy since it's basically a whole web view.
And in your approach, you are also starting an external subprocess to pick a color. And do it in a blocking manner. So not ideal.

> This one has a lot of configurations.

Does a lot of configuration make a package worse? As long as it works out of the box without having to change any setting, I'd say that it doesn't make it more difficult to use.

> I thought any package would be welcome here...

See https://packagecontrol.io/docs/submitting_a_package#Step_1

raphaelquintao:
> All that trickery [with direct file path checking](https://github.com/raphaelquintao/QColor/blob/d5cbeac1f1e2394ac42f1731eeaeeb3ceb23d7b6/QColor.py#L49-L90) and `chmod`ing some files won't work with the released package as the released package is compressed and there is no direct access to your files from the filesystem (there are ways around that but it's not ideal). Also, the part where you reload stuff in `sys.modules` looks a bit scary.

I saw some packages doing that before... 

I guess i  can put the `.no-sublime-package` file to use external app. Right?

> Is it though if it requires that the user installs some pip dependency in an external (to ST) environment?
> And then potentially deal with some path issues and ST not finding that dependency?

That will not happens, just the external app will need the pywebview.

> I guess you've achieved some level of simplicity in your code but delegating functionality to an external dependency but that dependency can be problematic and it's quite heavy since it's basically a whole web view.

Unfortunately to make a descend picker it is necessary to do in an external process, due the sublime limitations. 
I have a plan to make a native picker for every SO, but this is a lot work, so i use webview for now.

> And in your approach, you are also starting an external subprocess to pick a color. And do it in a blocking manner. So not ideal.

I can work on that, but i saw some packages doing it that way.


> Does a lot of configuration make a package worse? As long as it works out of the box without having to change any setting, I'd say that it doesn't make it more difficult to use.

Not worse, but different...

> See https://packagecontrol.io/docs/submitting_a_package#Step_1

braver:
> i saw some packages doing it that way.

Ha, that's not the best defense of course, I've seen packages do all sorts of wild things 😉 

> > And then potentially deal with some path issues and ST not finding that dependency?

> That will not happens, just the external app will need the pywebview.

Let me tell you, path issues _will_ happen. And that's assuming your user has a compatible python/pip setup and env. It would be a good idea to to provide some more context in your readme on what dependency you're actually using and why, and where a user could go to figure things out if they don't get it to work.

Some other points:
- Please avoid using chmod, instead maybe let `python` run those files?
- Please don't ship key bindings. They tend to conflict with those of other packages (there are simply not that many keys on the keyboard) and/or not match conventions for the user's OS. Instead, provide information in the README about the available commands and how to set them up. 
- Make sure all commands are available via the command palette. For many users this is as good as having key bindings, and is more discoverable and intuitive that trying to remember all key bindings for all packages. 
- Please don't create top-level main menu entries. If all packages did that we'd quickly run out of space. Your package settings should go in Preferences > Package Settings. Commands could be placed under Tools, but that's generally not recommended: users go to the Command Palette to find commands. In fact, even your package settings should really be available through the palette as well. 
- The same generally goes for the context menu. There is really limited space here, so if you really need to be in that menu, please take up at most 1 position and group your commands in a submenu if you need more. 
- Users will generally don't care about your package version. Updates generally happen in the background, and it's not something they control either: package control ensures everyone is always on the latest release.

raphaelquintao:
Thanks for the feed back.

> Let me tell you, path issues _will_ happen. And that's assuming your user has a compatible python/pip setup and env. It would be a good idea to to provide some more context in your readme on what dependency you're actually using and why, and where a user could go to figure things out if they don't get it to work.

I improved the README.
 
>     * Please avoid using chmod, instead maybe let `python` run those files?

It's better. I did that.
 
>     * Please don't ship key bindings. They tend to conflict with those of other packages (there are simply not that many keys on the keyboard) and/or not match conventions for the user's OS. Instead, provide information in the README about the available commands and how to set them up.

For that kind of plugin keybinding is really helpful, is just 2 keybindings, and people can easily see that on menu. 
But if you want i can remove it...

>     * Make sure all commands are available via the command palette. For many users this is as good as having key bindings, and is more discoverable and intuitive that trying to remember all key bindings for all packages.

Good one. Did that.

>     * Please don't create top-level main menu entries. If all packages did that we'd quickly run out of space. Your package settings should go in Preferences > Package Settings. Commands could be placed under Tools, but that's generally not recommended: users go to the Command Palette to find commands. In fact, even your package settings should really be available through the palette as well.

Fixed that. I put the main menu on 'Tools' and the Settings under 'Preference > Packages Settings'.

>     * The same generally goes for the context menu. There is really limited space here, so if you really need to be in that menu, please take up at most 1 position and group your commands in a submenu if you need more.

Fixed that.

>     * Users will generally don't care about your package version. Updates generally happen in the background, and it's not something they control either: package control ensures everyone is always on the latest release.

I put that, just for the menu to be more cool. 😀 
I can remove if you think it's disturbs.

braver:
> For that kind of plugin keybinding is really helpful, is just 2 keybindings, and people can easily see that on menu.

Well, every package developer thinks people need keybindings for _their_ package ;) There is no hard rule here, just a very strong recommendation to not ship keybindings and instead teach users how to configure their own.

Thanks for responding to the feedback and demonstrating you take this seriously 👍🏻 One final thing: the "description" key in your entry is superfluous, package control simply takes the description from your GitHub repo. Removing the description from your entry will allow you to update it without having to go through a PR here.

raphaelquintao:
> Thanks for responding to the feedback and demonstrating you take this seriously 👍🏻 One final thing: the "description" key in your entry is superfluous, package control simply takes the description from your GitHub repo. Removing the description from your entry will allow you to update it without having to go through a PR here.

I removed the description.

braver:
🍻

PR #8207, opened 2021-03-05
Added RepeatHint

absop:
This is a plug-in to `show, mark, select, remove` repeat lines in buffer. More information can be found in the `README.md` file under the repository.

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -1274,6 +1274,17 @@
 			]
 		},
 		{
+			"name": "RepeatHint",
+			"details": "https://github.com/absop/SublimeRepeatHint",
+			"labels": ["mark", "show", "remove", "repeat"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


braver:
Thank you for submitting! Some feedback:
- "markbook" isn't an icon in default themes, "bookmark" is. There is also the "circle" icon, that's slightly larger than "dot".
- What is the utility of show_repeat_lines? The popup isn't something you can do anything with. Usually you want such output to appear in the output panel so that it can be copied, and it allows reading and scrolling the code and refer to that output at the same time?
- It's not clear to me what a "repeat line" is. Is it a line that is repeated immediately, or is it a line that appears multiple times in the file?
- You ship with keybindings. Usually on macOS you would use CMD instead of CTRL. However, we also recommend against shipping with keybindings. There aren't that many keys on the keyboard. Also, nobody can remember all the keybindings, especially for utilities you don't use all the time. For most users the command palette is the first place to look, and for things like this that's probably all you need. So, please remove the keybindings and in stead document how to create them in the README.

PR #8210, opened 2021-03-06
add Finer Package Settings entry

celbig:
This package allow for a per project and/or per syntax configuration for any other package. I did not found any other package with this functionnality.

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -930,6 +930,17 @@
 			]
 		},
 		{
+			"name": "Finer Packages Settings",
+			"details": "https://github.com/celbig/finer_package_settings",
+			"labels": ["settings"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Firebase Snippets",
 			"details": "https://github.com/dhcar/firebaseSnippets",
 			"labels": ["snippets"],


celbig:
...
> ```
> Processing package "Finer Packages Settings"
> ```

The licence information is included in the README file

braver:
Per syntax or project settings can already support settings per package. Here's a good example: http://www.sublimelinter.com/en/stable/settings.html#project

This isn't well documented, but ST already merges these settings in a way so that packages can handle them seamlessly. An extra package could be convenient, depending on its implementation, but it certainly isn't necessary.

celbig:
From what I understand of your example this works only for specific packages that implement it. The point of my package is to enable this kind of configuration possibility for all installed packages without modifying the current source code of said packages.

If this is already possible from ST3 I am interested (and will of course close this pull request).

braver:
Hm, you are correct. Most packages will only read and monitor their own settings file and not consider the state of settings per view. 

Alright, digging a bit further... From reading the README I don't have the foggiest on how I should actually use your package 😅  Where do I put the packagesSettings object?
 Also, you require a settings_file reference, but why would I need to strip the extension? Wouldn't it be clearer if it actually references the complete file name (letting your package deal with whatever post-processing you need)? 
 As I understand it your package writes to a particular file in the User space (with a lot of locking/unlocking going on?). That sounds like something worth documenting.

braver:
Please respond to the questions to continue the review.

PR #8211, opened 2021-03-08
Add Open Policy Agent Rego syntax highlighter

hatxor:


braver:
In Sublime Text syntax packages do a lot more that just highlighting. Usually packages are simply named after the language they support, so "OPA Rego" in this case. You would also want to add the label "language syntax" to your entry here.

braver:

PR #8216, opened 2021-03-12
Add: Autofill Arrays

civAnimal:
Fill arrays and other collections quickly and easily.


braver:
Please resolve the conflict. Looks good otherwise.

braver:
Pro-tip: don't re-use your fork of package_control_channel, you're now trying to re-add everything you previously added as well.

civAnimal:
> Pro-tip: don't re-use your fork of package_control_channel, you're now trying to re-add everything you previously added as well.

Could you please describe (or share a link) how to submit a package? What steps need to be taken? I am only familiar with the PR way. I've not been able get any help from Google.

braver:
The PR way is the only way. But you need to keep your fork in sync with the upstream repo. So once your PR is accepted, it’s easier to throw away your fork and then if you want to create a new PR first create a new fork. If you don’t create a clean slate like that before adding something new, you tend to end up with a messy PR like this one.
You can probably also rebase your fork and work it out that way, but that tends to be harder than simply starting over from scratch.

civAnimal:
Thanks.

I guess I should close this one and start over with a new fork.

PR #8217, opened 2021-03-15
Add GNOME Devhelp package

amezin:
[Devhelp][] integration for Sublime Text. Initially based on [Zeal plugin][].

[Devhelp][] is a GNOME app for browsing and searching API documentation.

[Devhelp]: https://wiki.gnome.org/Apps/Devhelp
[Zeal plugin]: https://github.com/SublimeText/Zeal

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -700,6 +700,16 @@
 			]
 		},
 		{
+			"name": "Devhelp",
+			"details": "https://github.com/amezin/sublime-devhelp",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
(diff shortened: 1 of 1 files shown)


braver:
The problem with the keybindings here is that they'll of course clash with those of the Zeal package. For new packages we tend to plea to remove the keybindings and in stead write something in the README about how to users can create them for themselves. I guess otherwise, being a pretty straight port from Zeal, you can't do a lot wrong here.

braver:

amezin:
Disabled key bindings by default in 1.0.1

PR #8218, opened 2021-03-17
Update s.json

ggstec:

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1652,6 +1652,18 @@
 			]
 		},
 		{
+			"name": "SideBarAutoGroup",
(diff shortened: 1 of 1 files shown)


ggstec (commented):
Created LICENSE

rchl:
- You can re-request review from the bot yourself
- The readme says "A web browser page reloading plugin for the Sublime Text 3 editor." which is not really the case.

ggstec:
Modified Readme.md

braver:
Can you explain what your package does and how it provides something that doesn't already exist by default or in other packages? I really don't get from the readme what it does (maybe it needs a gif...) and that might also mean that it makes it hard for others to find your package and see how it solves a problem they might be having.

braver:

PR #8220, opened 2021-03-17
add open_doi package

crazytexer:
This package provides a new command in the context menu. It opens the selected digital object identifiers (DOIs), that are formatted without `https://doi.org/` and look like `10.1103/PhysRevLett.116.061102`, as URLs in your browser.

### DIFF
--- a/repository/o.json
+++ b/repository/o.json
@@ -422,6 +422,16 @@
 			]
 		},
 		{
+			"name": "Open DOI",
+			"details": "https://github.com/crazytexer/open_doi",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Open File in Current Folder",
 			"details": "https://github.com/SublimeText/OpenFileInCurrentFolder",
 			"labels": ["file navigation", "current folder"],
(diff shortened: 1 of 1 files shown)


braver:
Much credits for supporting multiple cursors, people tend to forget about that. 👍🏻  

In your run command you do a check to see if the selection looks like a DOI. You should do this check in an [is_visible](https://www.sublimetext.com/docs/3/api_reference.html#sublime_plugin.TextCommand) check, so that the command doesn't display (and take up space) in the context menu. Also, it's good practice to always include all commands in the command palette, as it's accessible without using the mouse.

Finally, I'd recommend renaming the package to "Open DOI".

braver:
Don’t forget to tag a new release with your latest changes!

crazytexer:
Thank you very much for my pull request processing!

> Don’t forget to tag a new release with your latest changes!

Sorry but do you mean the current release (because it already has some changes as compared to the 3-days-ago version) or any future releases?

braver:
In your repo I can only see the 1.0.0 from a few days ago.

crazytexer:
Yes, exactly! What I am asking is whether I should add a new tag right now (because ~3 hours ago I introduced some code changes according to your review, and the 3-days-ago tag 1.0.0 is technically older than that code update) or not. This is probably a very amateurish, if not to say silly, question, and I am sorry about it.

braver:
Package control will only pick up the releases from your GitHub repo. So anything newer than your latest release will not reach any users. So, anytime you do something new in your package you should make a new release. Then, usually an hour or two later, package control will see that release and distribute it to all users of your package.

PR #8221, opened 2021-03-18
Add ST3 CloseGroupAndAppend plugin

davidhcefx:
## Description

A ST3 plugin to close current group (pane) and append all tabs to the **BACK** of the previous group. (The default only prepend to the **FRONT**)

demo-animation

---

**Usage**: It can be bound to a key-binding or called from the Command Pallete.
**Repo**: https://github.com/davidhcefx/CloseGroupAndAppend

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1536,6 +1536,16 @@
 			]
 		},
 		{
+			"name": "CloseGroupAndAppend",
+			"details": "https://github.com/davidhcefx/CloseGroupAndAppend",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "CloseMinimapOnMultiView",
 			"details": "https://github.com/alkuzad/CloseMinimapOnMultiView",
 			"labels": ["code navigation"],


braver:
Hm, this sounds a bit niche, but 🤷🏻 You introduce one command but the Default.sublime-commands has about 4. What's going on there?

davidhcefx:
@braver The other three are the default commands in `View -> Groups`. I added them because I think it was strange to provide a command for closing groups without providing other commands for group creation. But you're right, this has the risk of conflicting with other plugins, so I think maybe I'll comment those out.

braver:
Oh, you're right. In that case I actually agree that it would make sense to have those commands in there, in the sense that you're enhancing the grouping feature via commands in the palette. There is a tiny chance of conflicts, but it should probably be fine. You can go either way with that 🤷🏻 
Another tiny detail is that you might want to remove certain files (gifs in this case) from the package via `.gitattributes` ([example](https://github.com/sublimelsp/LSP/blob/main/.gitattributes)). This way they won't be in the package as it ships to users, saving on precious storage, bandwidth and CO2.

davidhcefx:
Okay, thanks for your suggestion!

PR #8222, opened 2021-03-19
add liquidsoap package

saranrapjs:
This package adds support for [liquidsoap](http://liquidsoap.info) file syntax.


braver:
Please add a "language syntax" label, that will make it easier for users to find your package and understand what it's for.

saranrapjs:
@braver done!

PR #8223, opened 2021-03-19
Add POVRay syntax hilight

leoheck:
This PR adds Pov-Ray syntax highlight.

This is a well-known project http://www.povray.org/ that does not have any indication of a syntax highlight or any other feature here for Sublime-Text.

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -1948,6 +1948,17 @@
 			]
 		},
 		{
+			"name": "PovRay",
+			"details": "https://github.com/leoheck/sublime-povray",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Power System Tools Syntax",
 			"details":"https://github.com/dparrini/sublime_powersystems",
 			"labels": ["language syntax", "auto-complete"],


leoheck:
It was using tag `x.y` instead of `x.y.z`. Now it is ok. How do I trigger this CI again?

braver:
Thanks for submitting!

Hit that button:
Screenshot 2021-03-20 at 11 19 34

That's the first time I see the comments specified in a sublime-syntax file instead of tmPreferences, I didn't know that was a thing?

Your `entity.name.function` is missing `.povray` at the end.

leoheck:
Thanks for the feedback @braver and for teaching me how to restart the CI.
The missing scope was added too.

leoheck:
Any recommendation here? What should I do?
I tried to add these files `*-YAML.PLIST-tmLanguage` but it didn't help here.

```
  - WARNING: '.sublime-syntax' support has been added in build 3092 and there is no '.tmLanguage' fallback file
    - File: POVRay-Syntax.sublime-syntax
  - WARNING: '.sublime-syntax' support has been added in build 3092 and there is no '.tmLanguage' fallback file
    - File: POVRay-Comments.sublime-syntax
   ```

**Comment by braver** [comment]
I think maybe we should disable those warnings. Nobody uses a version older than 3092. You can ignore them for now.

Ready to merge?

**Comment by rwols** [comment]
Are you sure those comments work?

**Comment by braver** [comment]
Oh, damn, forgot about those, got tied up in the 3092 warning.

They’re not in any documentation I know about, but neither are the old TM comments.

**Comment by braver** [comment]
By the way, there is no need to prefix all files with "POVRay-".

**Comment by braver** [comment]
As for the comments, you want to have a file called Comments.tmPreferences, with these contents:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<plist version="1.0">
<dict>
	<key>name</key>
	<string>Comments</string>
	<key>scope</key>
	<string>source.povray</string>
	<key>settings</key>
	<dict>
		<key>shellVariables</key>
		<array>
			<dict>
				<key>name</key>
				<string>TM_COMMENT_START</string>
				<key>value</key>
				<string>// </string>
			</dict>
		</array>
	</dict>
</dict>
</plist>
```

leoheck:
Comments were working for me. But I will use your suggestions. Let me fix these things.

leoheck:
Remove unnecessary files. Files were renamed too. Comments file is using tmLanguage now.

rwols:
Were those comments still working when you removed the .sublime-syntax file? I’m not convinced it has any effect. In which case, you must already have some other tmPreferences file somewhere that already handles comments.

leoheck:
Now I have 2 instances of the POVRay, I renamed one to identify these things and I am using it now. How can I check and remove the second one so I can guarantee it works fine? I've tested during the last commit and the comments were working (as yesterday)

rwols:
You can run View Package File from the Command Palette to browse the “catalog” of ST.

leoheck:
Oh, I was able to find some replicated instances of my repo inside the config folder.
Now I just have one and its cache. Comments are still working. I can see only one PovRay on Sublime.
image

leoheck:
Cool, thanks for this trick. Yeah, files are right, since I have renamed them.
image

PR #8225, opened 2021-03-20
Add Koto package

irh:
This PR adds the Koto package, providing syntax highlighting support for the [Koto language](https://github.com/koto-lang/koto).

### DIFF
--- a/repository/k.json
+++ b/repository/k.json
@@ -445,6 +445,17 @@
 			]
 		},
 		{
+			"name": "Koto",
+			"details": "https://github.com/koto-lang/koto-sublime",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "KRL (Kinetic Rule Language) Syntax",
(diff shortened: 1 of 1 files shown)


jrappen:
Current docs:

<https://www.sublimetext.com/docs/syntax.html#ver-dev>

Also you probably want to add tests and use GitHub actions as is done for the default packages shipping with SublimeHQ products:

- <https://www.sublimetext.com/docs/syntax.html#testing:ver-dev>
- <https://github.com/sublimehq/Packages>
- <https://github.com/sublimehq/Packages/tree/master/.github/workflows>
    - target URL will change, but you knew that

irh:
Hi @jrappen, thanks for taking a look but it's not clear to me if there's a problem? You gave me a docs link without describing an issue, then started the next point with 'Also', which makes me think a sentence is missing from your reply.

jrappen:
* Docs link was meant as a replacement for the link you provide in the syntax file at the top.
* No sentence missing, add optional tests if you like. Whatever works for you.

irh:
OK thanks, I've updated the docs link, and I'll consider adding tests next time I need to update the package.

PR #8227, opened 2021-03-20
Added "blif_sis" package

mario33881:
This package adds support for BLIF, SIS' Berkeley Logic Interchange Format.
> SIS is a tool that optimizes and simulates logic gate and finite state machine networks

Specifically, it's features are:
* Syntax highlighting
    
* Keywords autocompletion
    
* Template snippets

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -965,6 +965,17 @@
 			]
 		},
 		{
+			"name": "blif_sis",
+			"details": "https://github.com/mario33881/sublime-blif_sis",
+			"labels": ["auto-complete", "language syntax", "snippets"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "BlinkScript",
 			"details": "https://github.com/diegoinacio/blinkScript-sublime",
 			"releases": [


mario33881:
Yesterday I saw [this pull request](https://github.com/wbond/package_control_channel/pull/8223):

one suggestion was to use a ```Comments.tmPreferences```
file for comments instead of using the ```.sublime-syntax``` file...

I tried removing the comments section
on my ```.sublime-syntax``` and ```.tmLanguage```
files and then add the ```.tmPreferences``` file:

everything worked the same except for the comments:
there's no syntax highlight...

This is the file content:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<plist version="1.0">
<dict>
  <key>name</key>
  <string>Comments</string>
  <key>scope</key>
  <string>source.blif</string>
  <key>settings</key>
  <dict>
    <key>shellVariables</key>
    <array>
      <dict>
        <key>name</key>
        <string>TM_COMMENT_START</string>
        <key>value</key>
        <string># </string>
      </dict>
    </array>
  </dict>
</dict>
</plist>
```

> ```View Package File``` shows the ```Comments.tmPreferences``` file:
>
>

mario33881:
> requested a review from packagecontrol-bot because I realized I forgot to add the file format to some "scope" props.

braver:
> everything worked the same except for the comments:
> there's no syntax highlight...

Not 100% sure, is there a question in there? 😄 You need tmPreferences file for the comments.

mario33881:
Basically what I meant to ask for is if I actually need to add that file
because it seems to do nothing...

The comments rule inside the ```.sublime-syntax``` 
and/or ```.tmLanguage``` are syntax hightlighting comments
but the ```.tmPreferences``` alone doesn't highlight comments.


I've added it anyway and tagged a new version.

braver:
Oh, no, the tmPreferences doesn't do anything for highlighting. You need to have that in your sublime-syntax. That the tmPreferences file does is basically just define what the commands "comment this line" or "comment this selection" should do, ie. what the keybinding CTRL+/ or CMD+/ does.

mario33881:
Oh, ok. I didn't understand that...

Everything should be fine now: syntax highlighting is working and the file was added

PR #8228, opened 2021-03-21
Added SmartBASIC language syntax file for SmartBASIC language.

eriklins:

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -4736,6 +4736,16 @@
 			]
 		},
 		{
+			"name": "SublimeSmartBASIC",
+			"details": "https://github.com/eriklins/sublimetext-smartbasic-syntax",
+			"releases": [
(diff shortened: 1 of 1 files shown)


braver:
You might want to add the "language syntax" label. Also, we require a [semver](https://semver.org) format tag, e.g. `1.0.0`.  Package looks good otherwise 👍🏻

eriklins:
I added the label "language syntax" as requested and changed the tag 
from 1.0 to 1.0.0 to (hopefully) comply with the valid semver tag rules. 
It still gives an error during check. No idea what's still missing.


Am 30.03.2021 um 20:19 schrieb Koen Lageveen:

braver:
The review process seems to crash on something, but your package and the tag looks fine. 🤷🏻

PR #8229, opened 2021-03-21
Add: AutoFill Arrays

civAnimal:
Fill arrays and other collections, like lists, tuples, sets, etc, quickly and easily. Compatible with all mainstream languages.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -2083,18 +2083,6 @@
 			]
 		},
 		{
-			"name": "AutoFill Arrays",
-			"details": "https://github.com/civAnimal/autofill-arrays",
-			"author": "civAnimal",
-			"labels": ["snippets", "auto-complete", "text manipulation"],
-			"releases": [
-				{
-					"sublime_text": "*",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "Auto Hide Sidebar",
 			"details": "https://github.com/xunker/AutoHideSidebar",
 			"releases": [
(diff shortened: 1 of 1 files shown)


braver:
Thanks for contributing. There are some issues with this, in general it's just a bit too simple to go into package control.
Your suggestions are loaded unconditionally, which means they crowd the suggestions everywhere. There is also no way for users to control where your suggestions show up.
In stead of FruitsA, FruitsB, FruitsC, etc. it would be nicer if it would just be "Fruits" randomised. Maybe you'd also want to provide a way to configure the length of the array? And maybe customise the entries?
The strings are double-quoted, you probably need to introduce a setting there for people who need/want single quotes.
I also wonder if the auto-complete suggestions are the right place for this. Suggestions tend to be language-specific and smart, this is neither. Maybe it would be better if this functionality was provided via the command palette.
Finally, you might also want to consider proposing this as a new feature of the existing [Random Everything](https://packagecontrol.io/packages/Random%20Everything) package.

civAnimal:
After reading your observations, it seems safe to conclude that this plugin won't be granted entry. Under the circumstances, I see no point in keeping this PR open any more.

General Public Notice:
Until forced-evicted, this plugin shall be maintained and improved (available as a free download) at: https://github.com/civAnimal/autofill-arrays

braver:
Alright! The package has potential, but to be included in package control that raises the bar a bit. You're always free to share your ideas on the forums etc. though.

PR #8230, opened 2021-03-22
Add Italian Dictionary

nixxo:
Add Italian Dictionary to the spell checker

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -33,6 +33,18 @@
 			]
 		},
 		{
+			"name": "Language - Italian - Italiano",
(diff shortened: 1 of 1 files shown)


braver:
2016.02.10 isn't a 100% valid [semver] version:
https://regex101.com/r/Ly7O1x/3/

nixxo:
updated my release tag from 2016.02.10 to 16.2.10 to maintain the same version as the original dictionary but in a format that is compatible with the regex. Hope is ok.

braver:
Excellent 👌🏻

PR #8231, opened 2021-03-22
Add QuickView package

jwortmann:
This is a port of the Quick View feature which can be found in the Adobe Brackets editor to display preview popups on mouse hover for images and CSS colors.

There is already https://packagecontrol.io/packages/ImagePreview for image previews, but my package for example doesn't require disk write access for images from the internet, and the popup style and behavior are tweaked to be similar to Brackets.
And there are of course several packages for the color previews, but those which I tried all use either regions or phantoms, and not hover popups.
Also this is afaik the only package which is able to display the color boxes with a transparent checkboard pattern background for colors with alpha value ;-)

### DIFF
--- a/repository/q.json
+++ b/repository/q.json
@@ -368,6 +368,17 @@
 			]
 		},
 		{
+			"name": "QuickView",
+			"details": "https://github.com/jwortmann/quick-view",
+			"labels": ["color", "image", "preview"],
+			"releases": [
+				{
+					"sublime_text": ">=3170",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "QuickXDev",
 			"details": "https://github.com/leitwolf/QuickXDev",
 			"releases": [


braver:
Thanks, that looks really useful! Some feedback:
- It wasn't immediately clear to me what would trigger a preview. Apparently hovering over something does it, but is there a way to trigger that from the keyboard? I would love to see a command in the command palette, also because it would explicitly trigger the preview immediately.
- Which brings me to the image previews: in many projects paths to an image don't follow a relative path from the current file to the image file. At least, I believe that's how the preview resolves the image. When that fails, and you don't get a preview, there is no feedback. Because the preview isn't instantaneous you're left wondering if something is wrong with the package, the image path, or generally why you're not getting a preview. Maybe you should flash a message in the status bar if the image cannot be resolved?
- For some image formats you require a third party utility. I would suggest moving information about those from the settings file to the readme. That would reduce the repetition there and allows you to link to information on how to install those utilities on the various platforms. 
- You may need to do a little work on your popups, I'm seeing a border around the outside:
Screenshot 2021-04-08 at 16 08 09

jwortmann:
Thanks for the feedback and the screenshot!

> It wasn't immediately clear to me what would trigger a preview. Apparently hovering over something does it, but is there a way to trigger that from the keyboard? I would love to see a command in the command palette, also because it would explicitly trigger the preview immediately.

I implemented the preview popups to only trigger on mouse hover over a supported CSS color or image URI, because that's the way it works in the Brackets editor. But as suggested, I will add a command for it to the command palette when applicable, so that users can add a keybinding if they want.

The color previews operate on scopes from the CSS syntax and should work for files that use/embed these scopes (CSS, HTML, PHP) as well as a few third-party syntaxes which comply with the CSS scopes (Sass, Less, Stylus). It should work for all colors that Sublime (minihtml) understands natively, as well as for all other valid CSS colors thanks to facelessuser's coloraide dependency.

> Which brings me to the image previews: in many projects paths to an image don't follow a relative path from the current file to the image file. At least, I believe that's how the preview resolves the image. When that fails, and you don't get a preview, there is no feedback. Because the preview isn't instantaneous you're left wondering if something is wrong with the package, the image path, or generally why you're not getting a preview. Maybe you should flash a message in the status bar if the image cannot be resolved?

Image previews should work not only for relative paths, but also for absolute URLs and data URIs. The previews are triggered only for applicable regions in a document, i.e. for HTML tag-attribute strings, the `url` function in CSS and for Markdown images, but not for random links in plain text (again, this is how it works in Brackets).

There is also a setting if a user wants to extend the applicable image link scopes and I will probably adjust the logic if the preview gets triggered from the command palette when I add a command for it. But I think I would add a status message only if it fails for the manual trigger, otherwise it might be too distractive when the user coincidentally hovers over incomplete or invalid links.

Some aspects of the package are still in development and support for the SVG, WebP and AVIF image formats was added just recently. I also just improved the parsing for data URIs and hopefully the previews should work for all or almost all valid link formats now. I will push a new tag soon or at latest when this will be accepted to Package Control.

> For some image formats you require a third party utility. I would suggest moving information about those from the settings file to the readme. That would reduce the repetition there and allows you to link to information on how to install those utilities on the various platforms.

Yes, I initially wanted to avoid requiring third party utilities, and I found a pure Python SVG rasterizer that I tried to embed into the plugin. But unfortunately it required some backporting to Python 3.3 and its only dependency Numpy isn't compatible with the totally outdated Numpy version on Package Control... And after adding some workarounds for the unsupported Numpy functions in that ancient version I couldn't get it to work satisfyingly, so I discarded that idea and rely on installed third-party programs now. I will tweak the readme a bit as soon as I know that the settings and supported utilities & formats are finalized.

> You may need to do a little work on your popups, I'm seeing a border around the outside:

If the popups have a border, then this border is forced by the theme (the Default and Adaptive themes don't do that), and there is nothing I can do about that in the plugin. But I realized that Sublime seems to add a shadow for popups on Linux and MacOS, and in case you're not using a custom them then I assume this "border" might be the shadow you are seeing in the screenshot.
Again, there is unfortunately nothing I can do about it via the API or minihtml content. The only way to prevent this would be to maybe restrict the package availability to Windows, but that wouldn't really be a satisfying solution either.

I'd love to further tweak the popup style, but some thing are just API limitations:
* A transparent background for popups isn't supported by ST, why I paint the background with the color scheme's background as a workaround instead, which might slightly hide some text in the line below (not that big of an issue, but still not optimal).
* If the popup would extend over the bottom of the screen, ST will move it on top of the specified position. I would like to move the little triangle pointer to the bottom of the popup when that happens, but ST doesn't notify about it and I think there isn't any other way to find out as well.

braver:
Thanks for the thoughtful reply. 

I guess that’s why most pop-ups don’t have the pointer nubbin. Indeed I’m using one of the default themes on MacOS. I don’t think the pointer adds much, if the pop-up is triggered by your cursor hovering over something it’s pretty clear what that pop-up is about. In fact, similar to exposing all commands via the command palette: that’s an area where you need to be mindful about how ST and other packages in the ecosystem work. Hardly anybody will recognize this as something that Brackets does similarly, and it’s not important: you’re building for ST now and people expect your package to be similar to things they already use here. For instance, have a look at the pop-ups in SublimeLinter, or LSP. Maybe it’s a better idea to align with that, even if that removes a particular design detail you wished worked the way you wanted it to.

jwortmann:
Yeah, styling popups in a satisfying way that works on all platforms and versions without any issues seems really difficult.
I've already struggled with minihtml popups earlier because unfortunately the implementation - especially for user-overrides - [is](https://github.com/sublimehq/sublime_text/issues/2935) [full](https://github.com/sublimehq/sublime_text/issues/3651) [of bugs](https://github.com/sublimehq/sublime_text/issues/3650) (I'm sorry for the many GitHub references), mdpopups [doesn't want to follow](https://github.com/facelessuser/sublime-markdown-popups/issues/70) the best practices for minihtml as described in the ST docs (and even if, things wouldn't work due to a ST bug :sweat_smile:), and when working with unknown content e.g. from language servers you can never be sure that things work as expected.

I wouldn't really mention the LSP popups as a good example though, because sometimes their behavior seems just to be random too and I don't know what causes that.

lsp

The native popups in Brackets just look much more beatiful than popups from Sublime plugins imo, and therefore I tried to replicate that style here. Perhaps I'll add a setting to make the pointer nubbin optional, but I think I'd prefer to keep in enabled by default then.

Maybe you can wait another week or two before merging this PR, so I have some more time to tweak some aspects of the package :)

braver:
Yeah, I bet it’s a lot easier to create beautiful UI in something based on a browser... 🤷🏼‍♂️

Anyway, sure, let’s put this on hold. Just give me a nudge if you’re ready for a new round of review.

wbond:
@jwortmann Is that GIF from build 4101? I thought we had tackled most of the popup layout issues, but I definitely want to know if we hadn't.

jwortmann:
@wbond Sorry, the gif is still from ST 3. I assumed the suboptimal popup dimensions only in some cases are either caused somehow by LSP or mdpopups itself, but I appreciate if it's fixed on ST 4.

jwortmann:
I have added a command and an entry in the command palette to trigger the preview popups for the current selection or cursor position. There is no key binding set by default, but if a user adds one, it can be used to toggle the previews. When triggered from the command palette or a key binding, it should also display a message in the status bar in case the preview fails for some reason.

I also added another setting to configure the popup style, so that it can hopefully look well on all platforms now.

And there were two bugs which are fixed now and prevented the popups in some special occasions previously. I hope that everything will work as expected now.

***

PS. I did not intend to sound rude in my recent comment; I was just frustrated at that time for wasting two hours unsuccessfully trying to get the SVG conversion to work in pure Python. Even though there might be an occasional bug here and there for power users in ST or a missing feature compared to editors based on web technologies, I can say that Sublime is still my favorite editor by far :)

braver:
> PS. I did not intend to sound rude in my recent comment;

It’s all good, I can definitely relate with the frustration. The world of development is full of little things that you think _should_ work a certain way but don’t.

Anyway, I’ll give it another look-see 👀 Sounds like some good improvements 👍🏻

braver:
Oh, boy, sorry, but I _will_ get around to it one of these days!

braver:
Alright, looks really good now. Maybe the settings are worth a little note in the README, because the popups don't look good by default on macOS (but excellent if you tweak it a little). Solid enough to publish now in any case, I'm sure you'll tweak it further later. Thanks for such a great addition!

--- 

By the way, as a daily user of LSP and some other package that use popups extensively, on the latest 4xxx releases, I can ensure you the popups are really quite solid ;)

jwortmann:
Thanks! I've just pushed another tag so that installations from Package Control will get the latest updates.

Maybe I should adjust the settings to use the most simple style for popups by default. Feedback from other users is certainly helpful for that. I have some more ideas for possible features to add, like previews for CSS gradients and to resolve color variables in Sublime themes (for color schemes it should work fine already), but that will take a bit of work.

And if I remember correctly, I've read somewhere that popups on ST4 can also have a transparent background, so there is probably more room for improvements. I'll have to experiment with that later.

PR #8232, opened 2021-03-23
Add "SQL Prettify" package

stubedston:

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -3362,6 +3362,17 @@
 			]
 		},
 		{
+			"name": "SQL Prettify",
+			"details": "https://github.com/stubedston/sublime-sql-prettify",
+			"labels": ["sql", "formatting", "formatter", "beautify"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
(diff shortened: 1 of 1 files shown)


braver:
It would probably be better to introduce python-sqlparse as a dependency, instead of shipping it inside your package.
Please don't ship key bindings. They tend to conflict with other packages, it's usually better to have the commands in the palette and then tell people how to set up their own bindings. 
Have you considered adding this functionality to the existing [Formatter](https://packagecontrol.io/packages/Formatter) package?

stubedston:
> 
> 
> It would probably be better to introduce python-sqlparse as a dependency, instead of shipping it inside your package.
> Please don't ship key bindings. They tend to conflict with other packages, it's usually better to have the commands in the palette and then tell people how to set up their own bindings.
> Have you considered adding this functionality to the existing [Formatter](https://packagecontrol.io/packages/Formatter) package?

Thanks for this. For some reason I didnt see Formatter when I initially looked. Agree it would be best to add functionality there.

braver:
Ok, let’s try that first. You can always come back here and reopen if that route doesn’t work out.

PR #8233, opened 2021-03-25
Add File Tab Rename

LDprg:
File Tab Rename is a small plugin for simple File renaming via Tab Context.

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -720,7 +720,7 @@
 			]
 		},
 		{
-			"name": "File Tab Rename",
+			"name": "FileTabRename",
 			"details": "https://github.com/LDprg/FileTabRename",
 			"labels": ["rename", "file rename"],
 			"releases": [
--- a/repository/f.json
+++ b/repository/f.json
@@ -720,6 +720,17 @@
 			]
 		},
 		{
+			"name": "File Tab Rename",
+			"details": "https://github.com/LDprg/FileTabRename",
+			"labels": ["rename", "file rename"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


braver:
We should probably replace this entry instead of adding something new (nobody uses ST2 anymore): https://packagecontrol.io/packages/RenameTab

LDprg:
https://packagecontrol.io/packages/RenameTab does only rename the Tab, but not the file name. It is quite simple package and works with ST3 too. 
My plugin renames the File (and the Tab to match File name). 

Nevertheless have no idea what I need to do in oder to replace this package, because I haven't done this yet. And I am not sure about replacing the package is the best solution, maybe renaming my package is a better way.

braver:
Oh, right, that only renames nameless tabs.

Alright, let's just take it as is, some feedback:

- The context menu entry should be "Rename…" (<- with an ellipsis, not 3 periods), to signal that it starts a process that needs more input. And to be inline with the sidebar context menu.
- I'm thinking this might be a nice addition to [SideBarTools](https://github.com/braver/SideBarTools), not strictly because it has anything to do with the sidebar, but because it also fills in some of the context menu file management gaps in ST's default setup. What do you think?

LDprg:
I have changed the context menu entry to "Rename File…".

I think makeing this an addition to SideBarTools is great. Do I have only to make a PR to SideBarTools?

braver:
> Do I only have to make a PR to SideBarTools?

I'd love to have the "Copy filename", "Move", "Duplicate" etc. commands also work on the tabs if you're able to do so with your solution in this package. That would really add value, and make a great addition to that package. So if you want you could make a PR there.

That said, I believe in small packages that just solve a problem without trying to do everything. So, I'm going to also merge this one here.

PR #8240, opened 2021-04-05
Add Mint Language

Orgmir:
This package adds syntax highlight support and a few snippets to the [Mint language](https://www.mint-lang.com).

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -1680,6 +1680,17 @@
 			]
 		},
 		{
+			"name": "Mint Lang",
+			"details": "https://github.com/Orgmir/mint-sublime-text",
+			"labels": ["language syntax", "snippets"],
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Minteresting",
 			"details": "https://github.com/daytonn/Minteresting",
 			"labels": ["theme", "color scheme"],


braver:
- The label you're looking for is "language syntax". 
- Any particular reason you push specifically to the default HTML package, instead of embedding the `text.html.basic` scope?

Orgmir:
Thanks for the feedback @braver and thanks for checking my code! I have updated the label to "language syntax".

Regarding the embedding, do you have more info on where can I find the available scopes?

I was following the documentation "Including Other Files" in https://www.sublimetext.com/docs/3/syntax.html I didn't know there where scopes available by default.

I checked the docs for the embed keyword, but replacing the current implementation with the embed keyword and `text.html.basic`  doesn't work the same way.

```yaml
html:
    - match: '(?i)(?=:(\s*)Html(\s*){)'
      embed: text.html.basic
      escape: '(?<=>})|^\s*}$'
```


| BEFORE |  AFTER |
|:------:|:------:|
|  Screen Shot 2021-04-09 at 8 18 38 am  | Screen Shot 2021-04-09 at 8 18 03 am |

braver:
Markdown and PHP in the default packages have some examples: https://github.com/sublimehq/Packages

Orgmir:
Thanks for the examples. I have force pushed a new commit with the correct labels, and changed the sublime syntax file to use `text.html.basic`.

It was missing a `scope:` prefix! Now it highlights the html correctly.

```yaml
html:
    - match: '(?i)(?=:(\s*)Html(\s*){)'
      push: scope:text.html.basic
      with_prototype:
        - match: '(?<=>})|^\s*}$'
          pop: true
```

braver:
Excellent! Don’t forget to tag a new release 🚀

PR #8241, opened 2021-04-08
Add "Hammock Color Scheme"

AlanLynn:

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -145,6 +145,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


braver:
If you're only targeting Sublime Text, you probably want to convert that to a sublime-color-scheme format (ST has a convertor built in in the Tools menu). The new format supports variables and a bunch of other quality of life improvements that make making and maintaining a color scheme much more fun.

PR #8243, opened 2021-04-09
Add PxToRem package

xiyuesaves:

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -2716,6 +2716,16 @@
 			]
 		},
 		{
+			"name": "PxToRem",
+			"details": "https://github.com/xiyuesaves/px-to-rem",
+			"releases": [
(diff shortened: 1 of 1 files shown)


braver:
We already seem to have [some](https://packagecontrol.io/packages/rem-unit) [packages](https://packagecontrol.io/packages/REM%20PX) that provide this functionality. What does your package add to what's already out there?

xiyuesaves:
Fixed the bug that `rem-unit` could not convert negative numbers and added keywords to determine whether the plugin is enabled

braver:
Well that sounds like you should open a PR on the existing package so it can be improved for all users of that package. Please try that first. And if that doesn’t work out for some reason, come back here and we can discuss some alternatives.

PR #8244, opened 2021-04-09
Add Rhai syntax highlighting.

schungx:
Adds syntax highlighting for the [Rhai](https://github.com/rhaiscript) scripting language.

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -1529,6 +1529,16 @@
 			]
 		},
 		{
+			"name": "Rhai Syntax Highlighting",
+			"details": "https://github.com/rhaiscript/sublime-rhai",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "RhetosDSL",
 			"details": "https://github.com/Hugibeer/RhetosDSLSyntax",
 			"labels": ["language syntax"],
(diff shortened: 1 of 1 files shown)


schungx:
I'm wondering if there is anything we need to do to merge this?

rwols:
I see that the vscode counterpart has language server integration. Consider looking into the LSP package and experiment if you can make that langserver work out of the box as well.

schungx:
Thanks.Will make the recommended changes. 

Not very sure how to generate a sublime-syntax file from tmLanguage... Can you recommend a converter?

The LSP is far away and so far it only has syntax highlighting... Should I wait until there is also an LSP before I submit again?

schungx:
> * Please add a "language syntax" label to make your package easier to find.

Done.

> * You can remove the "syntax highlighting" part from your package name. A syntax in ST does a lot more that simply highlighting. Also, you might want to add snippets and completions later. Most language packages simply have the name of the language they support, i.e. simply "Rhai".

Done with the renaming. 

> * If you're only targeting Sublime Text you probably want to convert this to a sublime-syntax file. It's a yaml format that is much easier to maintain, supports a number of situations and constructions that tmLanguage doesn't, has built-in tests that are easy to work with, and it's also more performant. ST has a built-in convertor in the Tools menu.

Done.  It is now switched to a `sublime-syntax` file.

braver:
Excellent!

PR #8245, opened 2021-04-09
New gitlab package: Jq wrapper

Jiehong:
The test default channel does not pass, because it does not accept gitlab as a url. However, I saw a PR long ago saying gitlab is now supported.

https://github.com/wbond/package_control/pull/1481/files

So, from August 2020. Perhaps the test is not up to date?

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -760,6 +760,17 @@
 			]
 		},
 		{
+			"name": "Jq",
+			"details": "https://gitlab.com/jiehong/sublime_jq",
+			"labels": ["interactive", "jq"],
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "JQ Syntax",
 			"details": "https://github.com/zogwarg/SublimeJQ",
 			"labels": ["language syntax", "jq", "stedolan"],


repository/j.json
```diff
@@ -759,6 +759,17 @@
...
 			]
 		},
+		{
+			"name": "Jq",
+			"details": "https://gitlab.com/jiehong/sublime_jq",
+			"labels": ["interactive", "jq"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "master"
```
rwols, line 769: branch-based releases are deprecated.

FichteFoll:
Yeah, the tests and the reviewer haven't been updated for gitlab yet. I (or someone else) will have to at least satisfy the local tests before merging this PR. We can ignore the pacakge reviewer for now, but the local tests will make all subsequent CI builds fail otherwise.

---

1. Are you sure you want to concatenate jq's stdout and stderr? stderr is likely to contain error messages and only have content if there were errors to begin with (and then stdout will likely be empty)
2. Instead of decoding and replacing newlines manually, use the `encoding` and `universal_newlines` arguments of `Popen`.
3. You may need to run `jq.exe` on Windows since the automatic extension detection is applied by Windows's shell, cmd.
4. The JSON syntax has been moved to a separate package for ST4. Also, `set_syntax_file` is deprecated. Instead, you may also use `apply_syntax("scope:source.json")`, but only on ST4.
5. You can feed jq text through stdin without having to create a temporary file.
6. Modifying the view in-place on every input of the input panel is likely to cause a lot of changes that the user then needs to undo manually (unless ST automatically groups them), so that experience may be suboptimal. You are resetting the contents if the user cancels, but their undo stack will still be messy.
7. In your apply command, you unnecessarily iterate over multiple regions even though it's always going to be the whole file. Instead, what if you took the regions from `view.sel()`? Note that regions from the selection only update itself (from modifications) until they are retrieved through `__getitem__` or iterating over them, so be aware of that when making changes to multiple regions.

Jiehong:
@FichteFoll thanks for reviewing the plugin itself as well!

1. I think I can check the return code to decide to print stdout or stderr, but that works as well. I like to see the output in case of errors, to know how to fix it.
2. ah, I didn't know that! I'll look into it
3. I tried the plugin on both windows and linux, and it works in both cases without .exe so far.
4. can I put apply_syntax in a try-except to make it work in both ST3 and 4?
5. I didn't manage to make it work with stdin, but I'll look into it more, as I agree having a temporary file is not great
6. I've clearly noticed the shit being added to the cancel stack for that file. I couldn't make it both interactive and not mess up the cancel stack so far. I think I might have to duplicate the view first instead, but I also don't like much plugins that duplicate things like that. Is there a way to remove a set of modifications in the cancel history to avoid this mess? Or to force grouping them?
7. I think I started iterating, mainly because of a copy paste, but also I thought it would be nice to handle this on a user selection, but it's quite difficult, as the size of the text changes. So I'll indeed remote the iteration!

I wished the API reference for Sublime had a few examples for each (that page: https://www.sublimetext.com/docs/3/api_reference.html#sublime.View), also with common patterns....

I don't mind waiting for the repository to accept gitlab in the tests before this is merged.

FichteFoll:
2) The parameters are old, so 3.3 is fine. Since 3.8, there is a new `text=True` parameter that combines the two.
3) All right then.
4) apply_syntax is a very old API, so yes. You'll need an ST version check for the parameter, however. I suggest using `"scope:json.source"` on ST4 and what you have currently for ST3.
5) Make sure to pass `stdin=subprocess.PIPE` and then just call `process.communicate(text)`.
6) It's a complicated situation. For InsertNums, I tried to unwind my previous changes for previewing by running the `undo` command from plugin code, but you still need to be careful you're only undoing *your* changes and you need to beware of the user running your command twice. This is just a heads-up, though, and not a requirement to pass review.


The API docs mention at the top that the default plugins can serve as examples, observable with the "View Package File" command. 

You may also be interested in the updated docs that have version-specific information https://www.sublimetext.com/docs/api_reference.html.

FichteFoll:
I also just noticed that your tag is not a valid semantic version. It needs 3 numeric parts.

FichteFoll:
Support for gitlab in checks has been added in acd6ba73115c5e834db370bbd3747fb38908f64e.

Jiehong:
@FichteFoll I've fixed the tags, rebased, but the bot seems to say my "details", aka my repository url, is invalid. 

I've also fixed most of the issues in the code thanks to you review, so thank you for that.

FichteFoll:
Yeah, we'll have to ignore the reviewer bot for now, as updating that is more involved and I didn't find the time for that today (and would like to refrain from giving an ETA on it as well). Either way, it onl triggers for packages that were *added* in a PR and not for all CI runs like the checks that failed before, so we can merge this after the manual review I just did.

Jiehong:
Thanks again @FichteFoll

rchl:
I wonder if it would make sense to rename the package.

Searching for `jq` returns a lot of `jquery` packages.
Also, I would suggest adding instructions in the readme on how to install this package. More specifically, what is the package name one should look for?

Jiehong:
@rchl : it really is just called "jq" only. But it doesn't seem to appear on Package Control so far, but idk why.

However, I can update its Readme to add manual installation steps for now.

rchl:
> @rchl : it really is just called "jq" only. But it doesn't seem to appear on Package Control so far, but idk why.

Yes. I didn't mean to not remove `jq`, just add some extra keywords to make it easier to filter. Not sure what would those keywords be. Maybe `jq toolkit`?

It's fine that it doesn't appear on PC. It can take an hour or two to appear.

> However, I can update its Readme to add manual installation steps for now.

I meant that you could add instructions on how to install it from Package Control. It's just the name of the released name of the package is not obvious for someone looking at the readme on gitlab.

Jiehong:
@FichteFoll Is there anything else to do on my side to have it appear on Package Control? I can see [Mint Lang](https://github.com/wbond/package_control_channel/commit/d4fb3c66d0eff6c1f372466bc6d592f4db1d55ed) appearing, which was merged at around the same time.

Perhaps another part needs to be changed for gitlab packages to be seen?

If I can do anything to help, just let me know.

FichteFoll:
Hm, I'm not sure. Everything looks correct from my side. However, it appars that your package is the first one to be hosted on Gitlab and it may be that the entire pipeline isn't updated yet? I did check the [packagecontrol.io code](https://github.com/wbond/packagecontrol.io/search?q=gitlab), however, and that seems to be up to date.

PR #8247, opened 2021-04-11
Added LuaVEXT

MajorVictory:
Provides Lua 5.3 syntax highlighting, snippets, and autocompletion for Battlefield 3: Venice Unleashed's custom Lua engine with additional support for EBX dump text files.

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -1783,6 +1783,19 @@
 			]
 		},
 		{
+			"name": "LuaVEXT",
+			"details": "https://github.com/MajorVictory/LuaVEXT",
+			"labels": [
+				"auto-complete", "language syntax", "snippets", "lua", "ebx"
+			],
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


braver:
Thanks for contributing. Some feedback:
- Your snippets are registered for the "source.luae" scope, but no such scope is defined in your package (probably just copied that from LuaExtended?).
- Your main syntax presents the "source.lua" scope, but it seems to be a more specific derivative from "normal" Lua. To present the syntax as basically the same as Lua, but different, you might want to change that scope to "source.lua.vext". That means most other packages will recognise it as Lua (source.lua still matches your scope), but it can also be more specifically targeted for cases where it's different.
- Your EBXVEXT syntax registers itself for .txt files. That seems a bit too broad. I would suggest leaving it to just "ebx". Any user that regularly want to open txt files in this syntax can always configure ST to see all .txt as EBXVEXT (via "open all with current extension as \<syntax\>").

MajorVictory:
I've implemented your suggestions. Should I resubmit through another pull request?

braver:
No need for a new PR, your entry here doesn’t change.
Looks good, let’s ship this thing.

PR #8248, opened 2021-04-12
Add ST4 tags for JS Custom.

Thom1729:


braver:
Don't forget to actually tag something with that prefix!

PR #8251, opened 2021-04-15
Added plugin "Rename Open File"

kadadov:

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -1222,6 +1222,16 @@
 			]
 		},
 		{
+			"name": "Rename Open File",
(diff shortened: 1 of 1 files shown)


kadadov:
Added a tag. What else do I need to do?

rchl:
ST3 already has a "Rename file" functionality. You can invoke it by right-click a file in the sidebar or from the command palette.

And  ST4 has it too, slightly improved as it selects the file name by default.

braver:
Indeed, what does this package add that is not already available?

PR #8252, opened 2021-04-18
Add ST4 version of babel-sublime.

Thom1729:


repository/b.json
```diff
@@ -32,6 +32,10 @@
 				{
 					"sublime_text": ">=3000",
 					"tags": true
+				},
+				{
+					"sublime_text": ">=4000",
+					"tags": "st4-"
```
rchl, line 38: I haven't seen that convention used before, where both ranges match the ST4 version. Are you sure that Package Control handles it correctly? Typically the ranges would be exclusive (<4000 & >=4000)

rchl, line 38: Also, would it make sense to make the ST3 use the tag prefix and make the ST4 use just plain version tags since ST4 release is supposed to be just around the corner?

Thom1729: There are already a large number of non-prefixed tags in the repository, and I think it makes the most sense to use a new prefix going forward rather than renaming all of the existing tags.

Thom1729: I actually haven't tested it. It doesn't seem correct to mark the existing versions as incompatible with ST4.

rchl: I don't see why the old ones would need to be renamed.

If the master branch is currently compatible with ST3 then fork off it to create an ST3 branch and then use the master to develop the ST4-compatible version.

Then create an ST3-prefixed tag for ST3 and make a switch here to use it.

rchl: But it's not necessarily being defined here which versions (tags) are compatible but rather which versions you want to serve to specific ST versions.

Thom1729: I've changed it.

Thom1729: I've changed it.

repository/b.json
```diff
@@ -30,7 +30,11 @@
 			"labels": ["language syntax", "javascript"],
 			"releases": [
 				{
-					"sublime_text": ">=3000",
+					"sublime_text": ">4000",
```
rchl, line 33: should be "lower than".

Thom1729: Oops! Fixed.

Thom1729:
@braver Sorry to bug you, but it looks like I misnamed the tag (st3-v10.x.x instead of st3-10.x.x) and now the package is listed as missing. (As it happens, I did the same thing for JS Custom.) I've created new tags, but I think that someone has to manually remove the missing condition?

braver:
Oh, I hope that would solve itself actually. There is some other manual review that needs to be done that only @wbond seems to be able to do. But I guess in a few hours time a new run will discover the tags?

Thom1729:
It does look like Babel has moved to “Needs Review”. I fixed its tags first, so hopefully JS Custom won't be far behind.

braver:
Oh, that's good. Progress. The review thing I can't do, I think that's Will's job. Or hobby, not job 🤔

Thom1729:
What is a hobby, if not a job that doesn't pay you?

braver:
If you want to track progress on the manual review, you might want to open an issue here: https://github.com/wbond/packagecontrol.io/issues. I assume Will will go through all of the outstanding reviews in due time.

PR #8253, opened 2021-04-19
Adding Data Wizard package

nickklaskala:

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -274,6 +274,16 @@
 			]
 		},
 		{
+			"name": "Data Wizard",
+			"details": "https://github.com/nickklaskala/DataWizard",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Date",
 			"details": "https://github.com/paulollivier/sublimetext-date",
 			"releases": [
--- a/repository/d.json
+++ b/repository/d.json
@@ -230,6 +230,16 @@
 			]
 		},
 		{
+			"name": "Data Wizard",
+			"details": "https://github.com/nickklaskala/DataWizard",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},		
+		{
 			"name": "Data Wrangler",
 			"details": "https://github.com/AlexLamson/DataWrangler",
 			"releases": [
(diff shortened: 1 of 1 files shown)


nickklaskala:
Add a tag. let me know if there is anything else I need to change.

braver:
You can use [.gitattributes](https://github.com/SublimeLinter/SublimeLinter/blob/master/.gitattributes) to remove your gif files from the package, saving on storage space and traffic.

You might also want to document the available commands in a way that makes it easy to add to keybindings (I don't think most users will be able to figure it out from your readme as it stands).

How does it compare to existing packages like [Table Editor](https://packagecontrol.io/packages/Table%20Editor)?

nickklaskala:
I have added the gifs to the gitattributes. I have also updated the readme for recommended keybindings.

I think that table editor is a slightly different use case.  This package is to for anyone who is constantly opening and inspecting delimited files(typically ETL tasks).  Being able to justify , mask data, and collapse columns and save a file is a huge time save over importing it into some other program(god forbid excel).  Also being able to pivot data just makes sense when trying to look at an entire record when there are 150 columns.  Further when combining tools like pivot, pop, justify and sample with existing ST3 tools and packages like split-into-line, sort-lines, highlight-duplicates and tabnav you can get into some really wizard like text editing.

braver:
I'm a bit concerned about your baked in lists of cities, addresses, first names, etc. Ie. your random data seems quite random, and the lists are so big they kinda add weight that seems out of place. Wouldn't you rather move those out to a datafile that you can read on-demand? Or let some other random data [generator package](https://packagecontrol.io/packages/Random%20Everything) handle that?

nickklaskala:
thats a valid concern. the Random Everything package solves the same usecase.  i'll trim it down to just the scramble tools.

braver:
Alright, let me know when you're ready to continue the review and get this thing published!

nickklaskala:
It is ready.

braver:
- If you have any changes you want to publish, please tag a new release so Package Control will pick it up.
- Your package description and README say "st3", but since st4 is out now and your package probably also supports that, maybe just say "Sublime Text"?
- Maybe prefix your commands in DataWizard.sublime-commands with "DataWizard" to help people find them and distinguish them from other commands.

braver:
You missed an ST3 reference:

Screenshot 2021-06-09 at 14 34 46

PR #8255, opened 2021-04-20
Add 'LÖVELY2D'

kevinfiol:
[Link to repo](https://github.com/kevinfiol/LOVELY2D)

A LÖVE2D plugin for Sublime Text 4 that provides autocomplete and API information. Based on [LÖVE-API](https://github.com/love2d-community/love-api).

Similar to [Lua Love](https://packagecontrol.io/packages/Lua%20Love), but smaller in scope and with autocomplete that behaves significantly different to that plugin. Also intended to only with with Sublime 4.

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -1615,6 +1615,16 @@
 			]
 		},
 		{
+			"name": "LOVELY2D",
+			"details": "https://github.com/kevinfiol/LOVELY2D",
+			"releases": [
+				{
+					"sublime_text": ">=4102",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "LPC",
 			"details": "https://github.com/abathur/SublimeLPC",
 			"labels": ["language syntax"],
--- a/repository/l.json
+++ b/repository/l.json
@@ -1617,6 +1617,7 @@
 		{
 			"name": "LOVELY2D",
 			"details": "https://github.com/kevinfiol/LOVELY2D",
+			"labels": ["language syntax", "lua"],
 			"releases": [
 				{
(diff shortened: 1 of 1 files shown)


kevinfiol:
Added LICENSE file.

braver:
We also have the pretty ancient and seemingly dead package [SublimeLove](https://packagecontrol.io/packages/SublimeLove). How does that compare? I like removing dead packages from the registry if something new looks like a good replacement for existing users.

kevinfiol:
> 
> 
> We also have the pretty ancient and seemingly dead package [SublimeLove](https://packagecontrol.io/packages/SublimeLove). How does that compare? I like removing dead packages from the registry if something new looks like a good replacement for existing users.

@braver Thanks for the response. In my experience, SublimeLove autocomplete never worked for me in ST3 or ST4. Admittedly, my plugin still has some bugs, but the basic autocomplete functionality is there (my first ST plugin 😄). Additionally, SublimeLove (like Lua Love) also has build system support, which my extension, LOVELY2D, does not attempt to implement.

Also, please let me know if it's an issue that LOVELY2D only supports ST4 and not ST3. I only use ST4 these days, so I wanted to use the new ST4 features.

braver:
Ok, let's just leave the existing packages as they are. 

There are no requirements to support any particular version, supporting only the latest v4 builds makes a lot of sense if that's what you're using. That's where the future is anyway. 👍🏻

braver:
Oh, before I forget, please add the "language syntax" label to your entry, perhaps also the "lua" label. That makes it that much easier to find, and to understand what it's for.

PR #8257, opened 2021-04-22
Nice Sort

ChengAnXu2014:

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -1108,6 +1108,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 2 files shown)


braver:
Your PR seems to (accidentally?) add two packages. 

I also have a hard time seeing the general purpose of the NiceSort package. It seems highly specific to number files in that way. Why would other users of ST and package control need this functionality?

braver:
Please respond to the feedback!

PR #8259, opened 2021-04-27
Add "Trusted-Coding Static Analysis" package

innovation-madnext:

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -3249,11 +3249,11 @@
 		{
 			"name": "Trusted Coding",
 			"details": "https://bitbucket.org/madnext/trustedcodingsublime",
+			"labels": ["security", "static", "analysis", "code"],
 			"releases": [
 				{
 					"sublime_text": "*",
-					"tags": true,
-					"platforms": "*"
+					"tags": true
 				}
 			]
 		},
--- a/repository/t.json
+++ b/repository/t.json
@@ -3247,6 +3247,17 @@
 			]
 		},
 		{
+			"name": "Trusted Coding",
+			"details": "https://bitbucket.org/madnext/trustedcodingsublime",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true,
+					"platforms": "*"
+				}
+			]
+		},
+		{
 			"name": "TSL Text Scraping Language",
 			"author": "Dan Borufka",
 			"details": "https://github.com/polygoat/TSL",


braver:
Please, elaborate: what kind of package and why should it be added?

innovation-madnext:
The core of the package is that it can analyze statically PHP and Python to find security flaws. When scan has finished, it show all the flaws linked to OWASP and/or CWE weakness, so, the user can see a brief description and related information to those weaknesses.

repository/t.json
```diff
@@ -3246,6 +3246,17 @@
...
 			]
 		},
+		{
+			"name": "Trusted Coding",
+			"details": "https://bitbucket.org/madnext/trustedcodingsublime",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true,
+					"platforms": "*"
```
braver, line 3256: If platforms is "*" please leave it out.

repository/t.json
```diff
@@ -3246,6 +3246,17 @@
 				}
 			]
 		},
+		{
+			"name": "Trusted Coding",
+			"details": "https://bitbucket.org/madnext/trustedcodingsublime",
```
braver, line 3251: You might want to add some labels for categorization of your package.

braver (changes requested):
Your package seems to have neither a description nor a README. Without those people will not be able to discover your package and know what it's for, let alone know how to use it.

You ship with keybindings. This is not recommended because you will inevitably conflict with other packages. Also, it does not sound like this is something people need to have at their finger tips at all times. Besides, keybindings are not discoverable. Which is why we have a command palette. Please at the very least add your commands there.
Something similar goes for the context menu. Unless you really need to be in there, you should try to stay out of it: it's a very limited space and everyone wants their package in that menu. You already have entries in the Tools menu, which is great, and if you're in the command palette, usually that's more than enough.

It feels like you're either providing settings via that menu, or actually providing a wizard. If you're exposing setting that way, please consider making a settings file, or supporting project settings if you expect people to want different settings for different cases.

rchl:
- Your package is configured to connect to a local API server - https://bitbucket.org/madnext/trustedcodingsublime/src/d516341c2f93f09c427c2d8b2c06dc448391ad62/TrustedCoding.py#lines-204:205
 - If your package is sending the whole checked code into some remote API server you should make it very clear in the readme. And also have a license that is being specific on how you handle the data (I'm assuming it's your server)

repository/t.json
innovation-madnext: I just added some labels. Thanks! :)

repository/t.json
innovation-madnext: Ok! Removed in the new commit

innovation-madnext:
> 
> 
> Your package seems to have neither a description nor a README. Without those people will not be able to discover your package and know what it's for, let alone know how to use it.
> 
> You ship with keybindings. This is not recommended because you will inevitably conflict with other packages. Also, it does not sound like this is something people need to have at their finger tips at all times. Besides, keybindings are not discoverable. Which is why we have a command palette. Please at the very least add your commands there.
> Something similar goes for the context menu. Unless you really need to be in there, you should try to stay out of it: it's a very limited space and everyone wants their package in that menu. You already have entries in the Tools menu, which is great, and if you're in the command palette, usually that's more than enough.
> 
> It feels like you're either providing settings via that menu, or actually providing a wizard. If you're exposing setting that way, please consider making a settings file, or supporting project settings if you expect people to want different settings for different cases.

Removed context menu and keybindings. There is only two ways to execute the package options now, command palette and main menu.
Also added a README in the package repository with description, usage,... (PENDING LEGAL)

innovation-madnext:
> 
> 
>     * Your package is configured to connect to a local API server - https://bitbucket.org/madnext/trustedcodingsublime/src/d516341c2f93f09c427c2d8b2c06dc448391ad62/TrustedCoding.py#lines-204:205
> 
>     * If your package is sending the whole checked code into some remote API server you should make it very clear in the readme. And also have a license that is being specific on how you handle the data (I'm assuming it's your server)

Fixed: Endpoint removed, now pointing to prod.
Pending: I'm waiting for the legal department. At the moment, it has only been indicated in the README that the code will be sent to our backend, where it will be analyzed and processed. So, if You want, this PR can be hold on until the legal part is ready...

braver:
Looks good to me. I don't have a project on hand to test it with, but it looks sane enough and has nothing that should harm performance or anything. The information in the README is clear enough, thanks for that. A license isn't a requirement, but a recommendation. I also don't necessarily need to wait for your legal department to clear it. So, as far as I can see it's good to go. Let me know if you're ready to merge this.

braver:
>  Let me know if you're ready to merge this.

So, yeah, give me a ping when you're ready...

innovation-madnext:
Sorry to keep you waiting! Legal department delayed the response a lot... The "Terms and Conditions" are ready in the repo's README :) I would really appreciate it if you accept the PR :) Thanks!

PR #8260, opened 2021-04-27
Add Unicode Character Input plugin

sylbru:
This package allows for inserting any Unicode character via its hexadecimal code point. The command opens an input handler that includes a preview of the character before inserting it. It also works with multiple cursors/selections.

image

Under Linux, there is a shortcut to do that, that works in most apps, except in Sublime Text. This plugin by default binds the command to that same shortcut (ctrl+shift+u).

Another package ([Unicode Character Insert](https://packagecontrol.io/packages/Unicode%20Character%20Insert)) sounds similar judging by its name, but it’s just displaying a table of ~65 hardcoded characters that you can then insert, you cannot chose any other character.

### DIFF
--- a/repository/u.json
+++ b/repository/u.json
@@ -210,6 +210,16 @@
 			]
 		},
 		{
+			"name": "Unicode Character Input",
+			"details": "https://github.com/NiavlysB/sublime-unicode-character-input",
+			"releases": [
+				{
+					"sublime_text": ">=3154",
+					"branch": "master"
+				}
+			]
+		},
+		{
 			"name": "Unicode Character Insert",
 			"details": "https://github.com/Sorbing/sublime-unicode-character-insert",
 			"releases": [
--- a/repository/u.json
+++ b/repository/u.json
@@ -215,7 +215,7 @@
 			"releases": [
 				{
 					"sublime_text": ">=3154",
-					"branch": "master"
+					"tags": true
 				}
 			]
 		},


TheSecEng:
A few things

1. Did you consider creating a PR on the other plugin to add this functionality?
2. Please remove the print statement found on Line 22. If this is for debugging, consider implementing a debug setting with debug logging
3. For the command palette entry, consider renaming it to give indication to what plugin it is associated with `UCI: Enter character` perhaps

braver:
In addition to the above, keybindings are not recommended. Ideally you leave them out of your package and instead put a note in your README about how user can set them up themselves if they want. That way yours won't clash with other packages someone might have installed.

NiavlysB:
@TheSecEng Thanks for the feedback!

1. I figured a PR didn’t make much sense because the other plugin provides a functionality that’s quite different and still serves the same purpose, so I wasn’t sure how they should cohabit. Not sure how to deal with that, I admit. Also it seems unmaintained so I decided to make that plugin for myself, then to share it via Package Control.
2. Oops, I’m removing that!
3. What about simply `Unicode Character Input`? The name of the plugin is the name of its only command, in the end.

NiavlysB:
@braver I’ll do that, thanks for the advice!

TheSecEng:
That sounds good to me. I think it just helps organize the command palette and easier to know where it’s coming from.

braver:
> the other plugin provides a functionality that’s quite different and still serves the same purpose

That package seems to have a similar or overlapping goal, but the approach is completely different. However maybe they could be two modes of the same package? Maybe, since it does seem unmaintained, you can become a maintainer on that package. Or we could replace it with yours if that doesn’t work out. I think we should at least try to reconcile the two.

NiavlysB:
Right, it would be great to merge both packages. I’ll suggest that to them.

The only thing is that in my opinion that other package’s functionality should be reimplemented in a better way: currently the list of insertable characters is hardcoded in [a HTML file](https://github.com/Sorbing/sublime-unicode-character-insert/blob/master/unicode-characters.html), where it should be configurable. I might try to do that.

braver:
Sounds like a plan 👍🏻

braver:
Alright, let me know how that works out. We can always re-open this one.

sylbru:
Since a recent version of ST, it seems like something has changed in the text input handling, and the native Ctrl-Shift-U from Linux (GTK? IBus?) is now working perfectly! Therefore my plugin isn’t needed anymore. 

image

I would be curious to hear more about this though, I couldn’t find any info in the changelog on what seems like a pretty big improvement to me!

PR #8262, opened 2021-05-02
Add GidOpen package

jongiddy:
Context menu entry to open or go to a file that appears in existing text in a Sublime Text window. For example, on Linux, right-click on the text `/etc/timezone` in an existing file will open a new tab to show (and edit) the file `/etc/timezone`.

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -414,6 +414,17 @@
 			]
 		},
 		{
+			"name": "GidOpen",
+			"details": "https://github.com/jongiddy/sublime-gidopen",
+			"labels": ["file navigation"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Gimp Auto-Complete Python",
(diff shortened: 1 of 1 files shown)


braver:
Have you seen the [HyperClick](https://packagecontrol.io/packages/HyperClick) package? How does yours compare?

jongiddy:
The HyperClick package seems to be specific to supported languages. Performing the simple steps described in the description above does not appear to trigger HyperClick to do anything.

GidOpen works in any window, including text files, log files, terminal output. 

https://user-images.githubusercontent.com/3735327/117058278-5876d280-ad16-11eb-9257-e3231f485928.mp4

braver:
Oh, that does look to be a lot more powerful. Some feedback:

- The `gidopen_in_view` command would translate in python to `GidopenInViewCommand` in camelcase.
- Your command is only available via the context menu. I would recommend also adding the command to the command palette so that it is accessible via keyboard as well.
- Your command does a lot of work that delays the context menu. You'd want to redesign that... maybe the user doesn't need to know exactly what path you've found?

braver:
Please respond to the feedback to continue the PR.

PR #8264, opened 2021-05-04
Add 'Show Minimap Setting' package.

ryanpcmcquen:
This package allows the Minimap to be toggled with a setting (rather than just the Command Palette toggle).

For example:
```
    "show_minimap": false
```

@packagecontrol-bot please review!

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -1057,6 +1057,16 @@
 			]
 		},
 		{
+			"name": "Remember Minimap Setting",
+			"details": "https://github.com/ryanpcmcquen/sublime_text_remember_minimap_setting",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Remify",
 			"details": "https://github.com/brousalis/remify",
 			"releases": [
--- a/repository/r.json
+++ b/repository/r.json
@@ -1047,21 +1047,21 @@
 			]
 		},
 		{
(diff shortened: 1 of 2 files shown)


braver:
Eh, so this means if you toggle the minimap off, it doesn't stay off for you?

By the way, ST3 runs Python 3, so UTF8 is default. Of course you don't specify ST >2, but nobody uses both ST2 and package control.

ryanpcmcquen:
@braver, if you wanted the toggle to work as it does now, you would just disable this plugin (or remove it). Main use case here is people who don't really want or use the Minimap but don't want to have to toggle it all the time.

ryanpcmcquen:
Another use case would be setting `show_minimap` to `false` for certain languages, like Markdown for instace.

braver:
Yeah, sure, but I seemingly only have to toggle it off once, not all the time. (that said I haven't tested over a long period of time)

> Another use case would be setting show_minimap to false for certain languages, like Markdown for instance.

Now _that's_ an interesting use case.

braver:
Might I suggest a different name? "Show" is also an active verb, and it sounds like it will show me something. Maybe "Remember Minimap (Toggle|Mode|Visibility)"?

ryanpcmcquen:
Sure! I am open to any name. "Remember Minimap Setting" seems descriptive enough, I will change it and update this PR.

ryanpcmcquen:
@braver, it has been updated!

braver:
🎉

PR #8266, opened 2021-05-06
nix-shell package

NorfairKing:
This package contains a `nix-shell` environment switcher like in https://marketplace.visualstudio.com/items?itemName=arrterian.nix-env-selector

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -712,6 +712,17 @@
 			]
 		},
 		{
+			"name": "Nix Shell",
+			"details": "https://github.com/niteoweb/sublime-nix",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": ["osx", "linux"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Nocturnal Color Scheme",
 			"details": "https://github.com/noct/nocturnal-color-scheme",
 			"labels": ["color scheme"],


NorfairKing:
How do I get the bot to recheck?

NorfairKing:
@packagecontrol-bot recheck please? :D

braver:
It's not clear to me from the README what problem this package solves. Can you provide some details?

There is plenty documentation about installing packages, so that you can have proper instructions in your own README (mostly you just clone a repo into the Packages dir, symlinks are not recommended for various reasons).

NorfairKing:
> It's not clear to me from the README what problem this package solves. Can you provide some details?

Absolutely.
This package allows you to enter a nix-shell environment from inside sublime.
This is handy for devs who use nix-shell to set up their dev dependencies.
(Why you might want to use nix is a very deep rabit hole, but the jist is: reproducibility.)


I'm pretty sure I will have done some things wrong, because I don't use sublime myself.
I made this for @niteoweb's @zupo. 
 
> There is plenty documentation about installing packages, so that you can have proper instructions in your own README (mostly you just clone a repo into the Packages dir, symlinks are not recommended for various reasons).

Could you point me to that so I can set it up correctly?
Does it still work if the source is in a subdir?

braver:
> I'm pretty sure I will have done some things wrong, because I don't use sublime myself.

Eh, so did you test this and do you intend to maintain this and support it going forward? Package control is for public packages for all users of Sublime Text. If you have a very narrow audience, there is no need to publish your package here. And one thing we do "demand" before adding a package is that developers exhibit a willingness to perfect and support their package.

zupo:
For what it's worth, I'm the one paying @NorfairKing for this, and I intend to keep using and paying for it. Been on Sublime for over a decade, and it's still my favorite. Recently switched to nix-shell, and I can see myself using nix-shell for the foreseeable future. This package helps a ton.

NorfairKing:
> Eh, so did you test this and do you intend to maintain this and support it going forward?

I didn't test it but I certainly tried it out and it seems to work for our purposes so far..

As for maintenance, that will be @zupo's @niteoweb's responsibility.

zupo:
Also, to be clear, I'm using it daily, it definitely works.

braver:
Well that's good to hear 😄 I have to say though, this is a bit out of the ordinary for package control. You don't need a package to be on package control, you can get distribute a package by hosting your own channel or you can very simply subscribe to updates for a single GitHub repo. When a package _is_ proposed for addition here we do an extensive review of the functionality so that the package plays nice in the eco system and users of package control can be sure stuff in there is more than just a personal project of the creator. We definitely go quite a bit beyond "yep, seems to work" here. In fact, the most importart thing we look for is that the author is able to work with us during that review to implement the changes. This also ensures that (at least for the immediate future) bugs can get fixed etc.

Right now, we have an author here who knows nothing about Sublime Text, so he can't cooperate in a review. And the proposed maintainer might not be able to do so either because they didn't build it. And is the package even useful for anyone outside the current user base? So, for now I'm going to reject this package and close the pull request. I'm totally willing to reconsider that move, but to do so someone who has full access to the package repo needs to step forward in a way that we can really do the review and get this package fully up to snuff.

NorfairKing:
@braver that's fine by me, but could you still point me to the docs about how to document how to install a package by `git clone`-ing like you said/

braver:
https://www.sublimetext.com/docs/packages.html

In Sublime Text via the main menu go to Preferences > Browse Packages. This opens the packages directory, and you can simply clone a package repo into this directory.

PR #8268, opened 2021-05-06
Switch Test Double package to use tags

rosston:
I'd like to switch this package to using tags. The "ChannelRepositoryTools: Test Default Channel" command in Sublime Text passes.

Note that [master](https://github.com/testdouble/sublime-test-double/tree/master) is even with [0.0.1](https://github.com/testdouble/sublime-test-double/tree/0.0.1) (the [only tag](https://github.com/testdouble/sublime-test-double/tags)), so there will actually be no changes.

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -731,7 +731,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"branch": "master"
+					"tags": true
 				}
 			]
 		},


braver:
Sure, good idea. Are you active in that project, or is someone in the testdouble team who can confirm the change?

rosston:
@braver Good question! I just realized that my membership in the testdouble org wasn't publicly visible. I now show up on https://github.com/orgs/testdouble/people.

I'll also go ahead and ping @searls to confirm this change, as he's [the only one who has committed that repo](https://github.com/testdouble/sublime-test-double/commits/main) and he's in the testdouble org as well.

searls:
Can confirm, @rosston is authorized to do this!

braver:
Excellent!

PR #8269, opened 2021-05-09
added linkedterminal

tobiasrenkin:
LinkedTerminal is a plugin for Sublime Text on Linux that connects an external terminal to Sublime Text. The plugin includes a build target that will execute build commands in the linked terminal and a "send" command that sends snippets of code to the terminal. In contrast to some existing solutions that rely on automated copy and paste functionality, LinkedTerminal uses pipes for communication between Sublime Text and the terminal. It does not require any external packages for its core functionality.

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -1071,6 +1071,24 @@
 			]
 		},
 		{
+			"name": "Linked Terminal",
+			"details": "https://github.com/tobiasrenkin/sublime-LinkedTerminal",
+			"labels": [
+				"terminal",
+				"shell",
+				"bash",
+				"console",
+				"zsh"
+			],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": ["linux"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "LinkerScript",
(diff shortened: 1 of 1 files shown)


braver:
@tobiasrenkin thank you for submitting your package. So, as I understand it, it can do two things. On the one hand you can run build systems in that "linked terminal" instead of ST sending it to a background process. And the other thing is you can send lines of code to that terminal for it to execute. A bit (a lot?) like the existing [External REPL](https://packagecontrol.io/packages/External%20REPL) package, but without the tmux requirement?

Some other points:
- Packages are delivered in an archived format. Have you considered if your package still works in that way, or if you require a `.no-sublime-package` file to prevent it being delivered that way?
- Your commands are now only available through keybindings. We recommend against shipping keybindings: they tend to conflict with those of other packages and they are hard to discover and remember. In stead, make sure all your commands are available via the command palette. Then, in your README or elsewhere, explain how the user can set up their own key bindings.
- It's common practice to present a main menu entry, and in most cases even a command palette entry, for editing settings files.

braver:

PR #8270, opened 2021-05-09
adding moon-light theme

developedby-sam:

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -137,7 +137,12 @@
 		{
 			"name": "MagentoWorkflow",
 			"details": "https://github.com/vovayatsyuk/MagentoWorkflow",
-			"labels": ["magento", "magento automation", "docker magento", "magento cache clean"],
+			"labels": [
+				"magento",
+				"magento automation",
+				"docker magento",
+				"magento cache clean"
+			],
 			"releases": [
 				{
 					"sublime_text": ">=3000",
@@ -578,7 +583,7 @@
 		{
 			"name": "MarkdownAssistant",
 			"details": "https://github.com/code-reaper08/MarkdownAssistant",
-			"labels": ["markdown","snippets","snippet"],
+			"labels": ["markdown", "snippets", "snippet"],
 			"releases": [
 				{
 					"sublime_text": ">=3103",
@@ -619,7 +624,13 @@
 		{
 			"name": "MarkdownEditing",
 			"details": "https://github.com/SublimeText-Markdown/MarkdownEditing",
-			"labels": ["markdown", "github markdown", "gfm", "multimarkdown", "color scheme"],
+			"labels": [
+				"markdown",
+				"github markdown",
+				"gfm",
+				"multimarkdown",
+				"color scheme"
+			],
 			"releases": [
 				{
 					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


developedby-sam:
This is a sublime theme and comes with its own color scheme, which is dark yet with vibrant colors. The colors are a lot easy on eyes than the themes already out there..!

braver:
Please don't change any entries other than the one you want to add!

developedby-sam:
> Please don't change any entries other than the one you want to add!

yeah, I guess it got auto formated..
should i do another commit or start a new pull request.?

braver:
Another commit is fine. In the end we’ll just squash everything anyway.

developedby-sam:
I guess, it is fixed now.

braver:
Yes, all that remains is to tag your first release!

jwortmann:
A few hints:
* The resource path of the package will be determined by the "name" property in this PR, i.e. it will be `Packages/Mooon Light Theme/`. There are a few invalid paths in your package, for example in "Widget - mooon-light.sublime-settings": `"color_scheme": "Packages/mooon-light/widgets/Widget - mooon-light.stTheme",`.
* The textures in the theme refer to `ayu/assets/`, which means that the theme only works if "ayu" is installed as well. I guess you would either want to describe that in the Readme, or use only assets from inside your package.
* A `dependencies.json` file without any dependencies in it has no real use
* The names of the theme and color scheme in the command palette are taken from the file names. You might want to adjust the filenames unless you want them to be lowercase and with the dash
* I would say the "popup_css" in the color scheme can be a bit of a "danger zone", because it can have unintended side effects if used. Here is how the phantoms for build errors from Default/exec.py on ST3 currently look with your color scheme:
screenshot

developedby-sam:
> Yes, all that remains is to tag your first release!

yeah! Done.

developedby-sam:
> A few hints:
> 
>     * The resource path of the package will be determined by the "name" property in this PR, i.e. it will be `Packages/Mooon Light Theme/`. There are a few invalid paths in your package, for example in "Widget - mooon-light.sublime-settings": `"color_scheme": "Packages/mooon-light/widgets/Widget - mooon-light.stTheme",`.
> 
>     * The textures in the theme refer to `ayu/assets/`, which means that the theme only works if "ayu" is installed as well. I guess you would either want to describe that in the Readme, or use only assets from inside your package.
> 
>     * A `dependencies.json` file without any dependencies in it has no real use
> 
>     * The names of the theme and color scheme in the command palette are taken from the file names. You might want to adjust the filenames unless you want them to be lowercase and with the dash
> 
>     * I would say the "popup_css" in the color scheme can be a bit of a "danger zone", because it can have unintended side effects if used. Here is how the phantoms for build errors from Default/exec.py on ST3 currently look with your color scheme:
>       screenshot
> 
>     * The files in the "images" folders are several MB large. You should use a .gitattributes file with `export-ignore` to not distribute them when the package is installed from Package Control

Thanks for the hints 😃, fixed most of them
Please review once again.

jwortmann:
You should not "export-ignore" the LICENSE file, it needs to be distributed with the package.

Besides that, the package looks good to me. The final approval is required from braver, of course.

developedby-sam:
yup!! thanks!! 😄

braver:
I’m about to hit “merge” just to get it out of the way, but be sure to actually those final changes. Package control will only pick up on your tags, so those file path fixes won’t reach your users unless you tag a release.

developedby-sam:
> I’m about to hit “merge” just to get it out of the way, but be sure to actually those final changes. Package control will only pick up on your tags, so those file path fixes won’t reach your users unless you tag a release.

I created new tag with all those changes, when those file path fixes will be able to reach to the users?

PR #8271, opened 2021-05-09
add aaLogPanel

gwenzek:
Add LogPanel plugin.
This allows to configure logging of ST pacakges.
Logging configuration is using the standard Python configuration.

This plugins also provide log handlers to write the logs to a ST panel.
Takes care of the two plugins hosts.

https://github.com/gwenzek/LogPanel/

The pacakge is named aaLogPanel to be loaded first.
If you have other ideas on how to achieve this I'm interested.
(I also considered making this package a dependency, but I'm not sure it's simpler)

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -15,6 +15,18 @@
 			]
 		},
 		{
+			"name": "aaLogPanel",
+			"details": "https://github.com/gwenzek/LogPanel",
+			"author": ["gwenzek"],
+			"labels": ["logging"],
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "ABAP",
 			"details": "https://github.com/flaiker/abap-st3",
 			"labels": ["language syntax"],


braver:
If you're proposing a general package that other packages can use, a dependency is the way to go. This would allow other package developers to declare they need this logger. I'm sure what you did here also "works", but it's a bit of a hack. 
Also, what kind of packages will be using this? Usually packages have their own logging requirements, and it's pretty easy (as far as I know) to implement some logging.

rwols:
I’m not a lawyer, but your chosen license, “AGPL”, states that any package using this must use AGPL as well. That’s quite restricting. Why not use MIT or Boost?

gwenzek:
> If you're proposing a general package that other packages can use, a dependency is the way to go. This would allow other package developers to declare they need this logger. I'm sure what you did here also "works", but it's a bit of a hack.
> Also, what kind of packages will be using this? Usually packages have their own logging requirements, and it's pretty easy (as far as I know) to implement some logging.

Setting up logging is easy if you're fine with a static config.
Generally I don't think this plugin should be required by any package, but if you want to do so please let me know and maybe switching to a dependency would be a good idea.
I was thinking of this plugin as a tool for Package author to see the log of the plugins they are using / developing.

gwenzek:
> I’m not a lawyer, but your chosen license, “AGPL”, states that any package using this must use AGPL as well. 

Hmm, that wasn't my intention. To clarify this I switched to LGPL which explictly allow importing this package.

braver:
> I was thinking of this plugin as a tool for Package author to see the log of the plugins they are using / developing.

In that case, the intention could be more clearly stated in the README. And I also think it might be a bit out of scope of package control. In any case I'm scared of the idea that we'll end up having more and more packages starting with "a" "aa" "aaa" "aaaa" in attempts to load early enough for other packages to use. If a package developer needs enhanced logging, maybe they would want to import your utility in a different way?

braver:

PR #8272, opened 2021-05-13
Add Kubeseal plugin

p3lim:
This plugin allows the user to run the kubeseal command from Bitnami Sealed Secrets directly on a Secret manifest in Sublime Text.

### DIFF
--- a/repository/k.json
+++ b/repository/k.json
@@ -517,6 +517,16 @@
 			]
 		},
 		{
+			"name": "Kubeseal",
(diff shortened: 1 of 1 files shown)


braver:
Thanks for contributing!
A tip: reading from settings is actually really fast, if you need a specific value that's usually returned from memory immediately. And you're not doing anything "on keyup" or anything. So load the settings whenever you need to access a property, because that ensures you get the current value. Loading the settings only when the package loads basically means people would have to restart ST whenever they change settings.

braver:

PR #8273, opened 2021-05-16
Added Estomato package

rakibulhaq:
This package generates an aggregated summary of the total effort required from a selected list having man hour or man days mentioned in specified format adding the result at the bottom of the selection text.

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -1316,6 +1316,16 @@
 			]
 		},
 		{
+			"name": "Estomato",
+			"details": "https://github.com/rakibulhaq/Estomato",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Esuna Framework Snippets",
 			"details": "https://github.com/Idered/esuna-snippets",
 			"labels": ["snippets"],
--- a/repository/e.json
+++ b/repository/e.json
@@ -1318,6 +1318,7 @@
 		{
 			"name": "Estomato",
 			"details": "https://github.com/rakibulhaq/Estomato",
+			"labels": ["text manipulation", "text selection", "aggregation"],
 			"releases": [
 				{
 					"sublime_text": "*",


braver:
Thanks for submitting. Looks pretty specific to me, do you have an idea for how many people will look for and find this in package control? Also, how they would know to look for your package specifically?

As for the package itself, I'm seeing main menu, context menu and keymap entries. 
- What's missing is probably the most important one, the command palette (usually in Default.sublime-commands). 
- The main menu entry is essentially ok, but "Selection" is mostly for manipulating selections. What you're adding here is something that belongs in Edit (because it inserts text), or Tools.
- The context menu I would advise against. That menu gets crowded pretty quickly, and I don't see any way for your command to remove itself from the menu if it's not applicable. Do you really _need_ to be in the context menu?
- Similar with the keybindings. There are too few keys on the keyboard to accommodate all packages. It doesn't look like something people need at their fingertips at all times. In fact, most people can't remember keybindings for stuff they don't use a lot, and the command palette is almost as quick as a key binding anyway.

rakibulhaq:
Hi, Thank you for detailed feedback. This package may be for fewer audience right now as it only would help developers who uses sublime text for doing feature estimation. But, i will be adding more aggregation operation on selected text, i.e getting sum of monetary values inserted, total sizes (megabytes / gigabytes) etc.

For people finding my package, I think different types of numbers (days,hours, later dollars, sizes as mentioned) aggregation from text can be a point. Please let me know if you have any suggestion for me as I am new to this :D .

- Added command pallet entry and removed all others as suggested. Only from command pallet is enough. Thanks for the tip!

Please let me know if any other improvement is required.

braver:
The description of the package is clear enough I guess, but it might help to add some [labels](https://packagecontrol.io/browse/labels) to your entry. 

Although I like the Estomato name, it sounds like it's not fully future proof for your plans. You're free to stick with it if you want to though. So, yeah, look at some label and let me know if you're ready to merge.

rakibulhaq:
Hi, thanks! I have now added some labels to identify my package better.

Also I would like to stick with the name as it is, hope it is okay. Thanks for your help.

PR #8275, opened 2021-05-21
Updated SortBy

Doi9t:
Updated SortBy (converted to tags instead of a branch)

Thanks!

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -2836,10 +2836,20 @@
 		{
 			"name": "SortBy",
 			"details": "https://github.com/Doi9t/SortBy",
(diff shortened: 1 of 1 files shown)


repository/s.json
```diff
@@ -2836,10 +2836,20 @@
 		{
 			"name": "SortBy",
 			"details": "https://github.com/Doi9t/SortBy",
+			"description": "A Sublime Text plugin that allows you to sort lines with methods that are not present by default",
```
rchl, line 2839: Description is automatically fetched from the repo so I think it's preferred to not add it here.

Doi9t: I removed the description; thanks!

PR #8276, opened 2021-05-21
Add WebAgent Syntax

seanfagan:
Syntax highlighting for the WebAgent 2 scripting language.

### DIFF
--- a/repository/w.json
+++ b/repository/w.json
@@ -110,7 +110,7 @@
 			]
 		},
 		{
-			"name": "WebAgent Syntax",
+			"name": "WebAgent",
 			"details": "https://github.com/seanfagan/webagent_syntax",
 			"labels": ["webagent", "language syntax"],
 			"releases": [
--- a/repository/w.json
+++ b/repository/w.json
@@ -110,17 +110,6 @@
 			]
 		},
 		{
-			"name": "WebAgent",
-			"details": "https://github.com/seanfagan/webagent_syntax",
-			"labels": ["webagent", "language syntax"],
-			"releases": [
-				{
-					"sublime_text": "*",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "Web Audio API Autocomplete",
 			"details": "https://github.com/hoch/waa-autocomplete",
 			"labels": ["autocomplete", "snippets"],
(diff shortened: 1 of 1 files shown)


Ultra-Instinct-05:
3 points about the package
1. You can use a `.gitattributes` file to prevent shipping files like screenshots (since they are technically not needed and just used to visually demonstrate package features) https://git-scm.com/docs/gitattributes
2. It looks like your package doesn't follow [semantic versioning](https://semver.org/). Every tagged release needs to have a semver https://packagecontrol.io/docs/submitting_a_package#Step_4
3. It looks like your syntax makes extensive use of negative lookbehinds. Note that using negative lookbehinds causes the syntax engine to fall back to the slower Oniguruma engine from the faster custom one (sregex). You can probably use lookaheads (though I haven't delved too much into the syntax itself)

repository/w.json
```diff
@@ -109,6 +109,17 @@
 				}
 			]
 		},
+		{
+			"name": "WebAgent Syntax",
```
braver, line 113: You can just call it "WebAgent": languages that implement support for a specific language are simply called after that language.

braver (changes requested):

seanfagan:
Thank you both.
- I've fixed the versioning. 
- I've added a `.gitignore`, which always helps, but I don't think it prevents my image from shipping like @Ultra-Instinct-05 suggested. Would I want something [like this](https://github.com/danro/LESS-sublime/blob/master/.gitattributes), with a line `*.png     export-ignore`?
- Thanks for the heads-up about lookbehinds. I use them everywhere right now as my custom way of detecting word boundaries, because this language considers "-" a word character, but "\b" does not. I'm sure it's possible to replace the lookbehinds but it's beyond me right now.
- I've fixed the name in `w.json`

repository/w.json
seanfagan: fixed

braver:
> Would I want something like this, with a line *.png export-ignore?

Yes, that's the one you want.

braver:
You may want to learn about [testing syntaxes](https://www.sublimetext.com/docs/syntax.html#testing). Also, on [Discord](https://discord.gg/HcmwdVK) there are a bunch of people who like helping out with syntaxes etc.

braver:
Ah,  the test fails, you need to place your entry right before `WebAssembly Text Syntax`.

seanfagan:
@braver Gotcha, I think I've fixed the ordering issue (and added a .gitattributes file).

braver:
Excellent. Let's ship, and you can figure out the lookbehinds in your own time. It's worthwhile though, well setup syntaxes for the new custom engine are blazing fast.

braver:
Of course it's immediately after merging that I think that maybe it should have been lower case `w` in webAgent?

PR #8277, opened 2021-05-22
Update Tutkain labels and description

eerohele:


braver:
While you're at it, you can also remove the description here. Package control will just take it from the repo directly (so you can make edits more easily later).

eerohele:
Thanks for the tip! Didn't know that. Force-pushed a commit that removes the description.

PR #8278, opened 2021-05-22
Add "Phoenix LiveView Snippets" package

sardaukar:
Adds snippets to make it easier to work with [Phoenix LiveView](https://github.com/phoenixframework/phoenix_live_view) code.

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -981,6 +981,16 @@
 			]
 		},
 		{
+			"name": "Phoenix LiveView Snippets",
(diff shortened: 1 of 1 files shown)


braver:
About your README: users can manually install such a package by putting the directory in Packages, no need to zip it up really.
You might want to add some "labels" to clarify that this works in the elixir syntax etc.

sardaukar:
Thanks for the feedback, will fix these soon

sardaukar:
README on repo has been updated, added some labels here too.

PR #8279, opened 2021-05-23
Add Move Half Pages.

Finnerale:
This package adds a `move_half_pages` command, which behaves like `move` with `"by": "pages"` and an additional `percentage` parameter (0.5 by default).

I find the normal 100% page scroll quite distracting and this just feels a lot nicer to me. I was quite surprised that there was no such package for Sublime yet, so I thought I'd publish mine.

The package can be found here: https://github.com/Finnerale/Sublime-MoveHalfPages

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -2490,6 +2490,16 @@
 			]
 		},
 		{
+			"name": "Move Half Pages",
+			"details": "https://github.com/Finnerale/Sublime-MoveHalfPages",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Move Tabs",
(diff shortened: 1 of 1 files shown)


braver:
I was initially wondering "what does moving a page mean? and what's a half page", but I understand now it moves the cursor by half a page 🙄 😄 

We require [semver](http://semver.org/) tags (i.e. 1.0.0, not 1.0).

Finnerale:
> I was initially wondering "what does moving a page mean? and what's a half page", but I understand now it moves the cursor by half a page roll_eyes smile

Yes, the name is a bit stupid but at least it is descriptive :sweat_smile: 

> We require [semver](http://semver.org/) tags (i.e. 1.0.0, not 1.0).

Oh right, I've updated it

braver:
Excellent 👍🏻

PR #8280, opened 2021-05-24
add LuauSyntaxHighlighting package

zainthemaynnn:
LuauSyntaxHighlighting package

this package intends to add syntax highlighting for the [luau subset of lua](https://roblox.github.io/luau/) which includes a bunch of extra globals and some keywords. although I'm aware that [this package already provides a similar purpose](https://packagecontrol.io/packages/ROBLOXLua) it's no longer maintained and the owner is inactive; this one stays up to date with network requests. while I thought about making a fork of it I had no interest in continuing support for snippets or anything like that. although I will be glad to do that if you'd prefer.

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -1817,6 +1817,17 @@
 			]
 		},
 		{
+			"name": "LuauSyntaxHighlighting",
+			"details": "https://github.com/zainthemaynnn/luau-syntax",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "LuaVEXT",
 			"details": "https://github.com/MajorVictory/LuaVEXT",
 			"labels": [


braver:
Have you tried opening a PR or otherwise coordinating with the author of that package? If that doesn’t work, we can look into replacing that package with yours, assuming you intend to maintain it and make sure the snippets continue to work (if they’re useful enough).

zainthemaynnn:
sure, I’ll see if I can try and reach out. if not I suppose maintaining whatever is already there shouldn’t be too bad, I’ll just replace the syntax file.

rwols:
Have you also considered adding this "Roblox" environment to lua-language-server and using LSP-lua instead?

rwols:
See also https://github.com/sublimelsp/repository/pull/2 where we almost added a special LSP-roblox language client

zainthemaynnn:
thanks, that actually sounds like the best idea because I made it with that language server in mind. although, will LSP-lua be the right place to put it? it could go there for the time being but I think that the unreleased client will be the perfect place to bundle this if it can get up and running.

rwols:
I am not too familiar with Lua and how it should adapt to different environment.

But I can make a comparison to Python. With Python you write an interface file ending in .pyi and that file describes the high-level functions and classes available for a module with the same name. I imagine it works similar for Lua. The maintainer of lua-language-server is trying to make his server adapt to different environments: sumneko/lua-language-server#409

From what I understand from the linked discussion, you set the setting `Lua.workspace.library` to some Lua file (?) and the server will then understand the functions/classes available in that file.

zainthemaynnn:
oh I see. if that works the plugin could pretty easily extend to support multiple frameworks, I will look into that.

braver:
Let me know what you want to do here. If you still want to add this syntax, let's try to replace the existing one.

rwols:
I was wondering that as well a few days ago @zainthemaynnn. Did things work out by using `Lua.workspace.library`? If so, maybe make a pull request to explain in the README of LSP-Lua how that would work to set up with WoW, Roblox, GMod, love2d, etc

zainthemaynnn:
> I was wondering that as well a few days ago @zainthemaynnn. Did things work out by using `Lua.workspace.library`? If so, maybe make a pull request to explain in the README of LSP-Lua how that would work to set up with WoW, Roblox, GMod, love2d, etc

sorry about this, forgot it was still open. yeah, actually I got one of those files working a few days ago. I can use the same thing I did in this package but parse one of those workspace files instead to achieve the same syntax highlighting. I'll submit a pull request to that package soon. in the meantime, I don't think we would really need this package once that's finished so I'll close this.

PR #8281, opened 2021-05-25
Add Acme color scheme & theme

multiversecoder:
Add Acme, a new color scheme and theme inspired by Plan9 and Acme for Sublime Text.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -180,7 +180,7 @@
 		{
 			"name": "Acme",
 			"details": "https://github.com/multiversecoder/Acme",
-			"labels": ["color scheme", "themes"],
+			"labels": ["color scheme", "theme", "acme"],
 			"releases": [
 				{
 					"sublime_text": "*",
--- a/repository/a.json
+++ b/repository/a.json
@@ -178,6 +178,17 @@
 			]
 		},
 		{
+			"name": "Acme",
+			"details": "https://github.com/multiversecoder/Acme",
+			"labels": ["color scheme", "themes"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Acme Color Scheme",
 			"details": "https://github.com/zaynali53/acme-color-scheme",
 			"labels": ["color scheme", "javascript"],


braver:
We already have an Acme color scheme. Is this the same thing? Or are you adding a theme?

multiversecoder:
@braver, I know there's already an acme-inspired color scheme, but I've never been satisfied with it. What I'm trying to do is make the experience consistent adapting theme and color scheme to Sublime Text.

braver:
Have you tried opening a PR at the existing Acme repo and incorporating your improvements there?

multiversecoder:
@braver, No, because I developed everything from scratch returning to Sublime Text after years of IDEs, mg, vim and acme.

I anticipate that I also plan to add more features to this package, primarily an advanced mouse interface and also panels to make Sublime more suitable for an Acme-like experience.

I'm reluctant to fragmentation but, I plan to develop a unique project so I think it's ok to take a specific direction.

braver:
Ok, I can see that, this should be a completely different package. I guess the word [Acme](https://en.wikipedia.org/wiki/Acme_Corporation) didn't mean anything specific to me. My main concern is that people looking for Acme won't understand (from the name) that the existing color scheme has nothing to do with the Bell Labs Acme editor, and won't be able to figure out how it relates to you new package. Or the existing [Acme Plumbing](https://packagecontrol.io/packages/Acme%20Plumbing), which it seems like you intend to replace. 

So, I guess we're looking for a different name?

One additional tip: you can exclude images you only use for documentation from the package using `.gitattributes`. Saves on storage etc.

multiversecoder:
@braver. I thank you for the conversation. 

I think "Acme" individually makes it clear to the user that the package refers to "Acme from Bell Labs" and will makes Sublime Acme-like.

If there are problems with the single word "Acme", I need to think for a couple of days about a new name to replace it. 

Regarding the mouse interface, now I have only a prototype of the package using Python3.8, that offers security and concurrency... but it is far from something definitive.

braver:
Maybe it's just me, but I associate "acme" with the fake corp from [Looney Tunes](https://looneytunes.fandom.com/wiki/ACME) 😄 

Some suggestions: Plan9 Acme, Acme Emulator, Acme-like, Neo Acme ... 

"Acme" maybe isn't a real problem, but it would be better to have a name that really signals "this is the package that will make work/look like the Bell Labs Acme editor", since the Acme name is effectively already taken even though it's just a (completely unrelated) color scheme. But think about it, maybe you can come up with something.

multiversecoder:
@braver, The first time I saw the word "Acme" again I immediately thought about Looney Tunes, being a fan.
I won't deny that even now my mind occasionally wanders to the countless times I've seen some scene where there was some weird product.

I'll think about the naming tomorrow.

Thank you.

braver:
Let me know if you intend to change anything here. Otherwise, let's just ship this as it is.

multiversecoder:
Hi @braver, sorry I haven't been in touch but, I have several things that have kept me away from this pr.

For me it would be very fine to ship "Acme" as it is.

braver:
Oh, one final thing: the label for themes is `theme` (singular, not plural), you'll want to change that for "SEO"

multiversecoder:
@braver, I changed the label

PR #8282, opened 2021-05-25
Enable GhostText on ST4

fregante:
I verified that this works correctly in Sublime Text 4.

Closes https://github.com/fregante/GhostText/issues/210

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -408,7 +408,7 @@
 			"previous_names": ["ChromeTextArea"],
 			"releases": [
 				{
-					"sublime_text": ">=3000",
+					"sublime_text": "*",
 					"tags": true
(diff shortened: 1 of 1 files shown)


braver:
Eh, >=3000 should already match ST4 (builds >4000)?

fregante:
Apparently not since it's marked as "ST3". You tell me 😅 

https://packagecontrol.io/packages/GhostText

Screen Shot

And it does not appear in the installable packages dialog in ST4

braver:
Ah, that icon is just an indication of the minimum version it seems. There is no requirement to specify support for ST4, your entry in package control is correct. I'll mention with maintainers of the site that the icons are confusing.

PR #8283, opened 2021-05-27
Add Taskfile plugin

biozz:
This plugin parses `Taskfile.yml` and allows running tasks from that file. More info on Taskfile - https://taskfile.dev/.

The version requirement is `>4000` because I am using Python version 3.8.

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -387,7 +387,7 @@
 				{
 					"sublime_text": ">4000",
 					"tags": true,
-					"platforms": ["osx"]
+					"platforms": ["linux", "osx", "windows"]
 				}
 			]
 		},
--- a/repository/t.json
+++ b/repository/t.json
@@ -397,7 +397,7 @@
 				{
 					"sublime_text": ">4000",
 					"tags": true,
-					"platforms": ["linux", "osx", "windows"]
+					"platforms": "*"
 				}
 			]
 		},
--- a/repository/t.json
+++ b/repository/t.json
@@ -381,6 +381,17 @@
 			]
 		},
 		{
+			"name": "Taskfile",
+			"details": "https://github.com/biozz/sublime-taskfile",
+			"releases": [
+				{
+					"sublime_text": ">4000",
+					"tags": true,
+					"platforms": ["osx"]
+				}
+			]
+		},
+		{
 			"name": "Task Timer",
 			"details": "https://github.com/jcowgar/sublime-task-timer",
 			"releases": [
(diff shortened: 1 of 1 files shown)


braver:
Thanks for submitting this package. 

`pyyaml` is available as a [dependency](https://packagecontrol.io/docs/dependencies), maybe your package can use that instead of relying on the user installing it on their system?

biozz:
Sorry, I overlooked this part of the docs.

I added `pyyaml` as a dependency, removed vendored package and prepared the new version.

PS: impressed how much thought went into package control ecosystem.

braver:
> impressed how much thought went into package control ecosystem

Well it has been around for a while :D So much cool stuff you can do here. 

As a point of general advice, something I experienced with my own package: not everyone appreciates the update messages. Nowadays I tend to limit it to things that really "break the API", ie. changes to settings users should know about. 

```
if not folders or len(folders) > 1:
```

I would recommend maybe flashing a message in the status bar, but otherwise simply defaulting to (attempt to) run in the first folder in the project. That way your package still functions mostly even if someone temporarily has another folder added to their project. Alternatively, you might be able to figure out which folder to run in from the `active_view`, by comparing its path to the folders in the project.

Finally, you seem to have some files in your repo (e.g. the gif) that are not actually used in your package. You can remove them from the package as it's delivered to users via `export-ignore` in a [.gitattributes file](https://github.com/SublimeLinter/SublimeLinter/blob/master/.gitattributes).

braver:
Oh, and the package seems to be out of alphabetical order.

biozz:
> not everyone appreciates the update messages

Good to know. I didn't add a message for this set of minor fixes :D 

> I would recommend maybe flashing a message in the status bar

I actually do just that, there is a `status_message` call right after the condition.

> otherwise simply defaulting to (attempt to) run in the first folder in the project
> which folder to run in from the active_view, by comparing its path to the folders in the project

I think I can add this as a feature in the following releases. So I would like to focus on adding the package to the package control and improve it later. Here is a filed issue https://github.com/biozz/sublime-taskfile/issues/1#issue-932640584

Would that be ok for you?

> you seem to have some files in your repo (e.g. the gif) that are not actually used in your package

This is a nice feature, didn't know about it. I added `.gitattributes` and tested locally with `git archive HEAD`.

> the package seems to be out of alphabetical order.

Moved it after `Task Timer` and before `Taskwarrior`.

braver:
> Would that be ok for you

sure!

Please have a look at the failed test. The platforms should be replaced with “*”. You can probably request a re-run of the test yourself to make sure everything is in order.

PR #8285, opened 2021-05-28
Renamed package and upgrade ST version

kinoute:
Now that ST 4 is available, I'm changing my Color Scheme to be a whole Theme (with also the previous Color Scheme), hence the change of the package's name.

I also upgraded the requirements since this theme will be only for ST4.


braver:
Ideally, the way to approach this is to tag a version for st3, with an `st3-` prefix for example, and have normal tags continue support for st4+. So you'd have 2 entries in "releases". There are a bunch of examples out there already with this approach.

kinoute:
I was able to make the theme and color scheme work on both ST3 and ST4, therefore, only the rename of the package is needed now. I updated the PR.

braver:
Excellent

PR #8286, opened 2021-05-29
ChoiceScript Syntax for Sublime 3/4

fawkesy:

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -960,6 +960,16 @@
 			]
 		},
 		{
+			"name": "ChoiceScript Syntax",
+			"details": "https://github.com/fawkesy/choicescript-syntax",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "choo snippets",
 			"details": "https://github.com/kareniel/sublime-choo-snippets",
 			"releases": [
--- a/repository/c.json
+++ b/repository/c.json
@@ -952,7 +952,7 @@
 		{
 			"name": "ChoiceScript",
 			"details": "https://github.com/fawkesy/choicescript-syntax",
-			"previous_names": "Choice Script",
+			"previous_names": ["Choice Script"],
 			"releases": [
 				{
 					"sublime_text": ">=3092",
--- a/repository/c.json
+++ b/repository/c.json
@@ -952,11 +952,11 @@
 		{
 			"name": "ChoiceScript",
 			"details": "https://github.com/fawkesy/choicescript-syntax",
+			"previous_names": "Choice Script",
 			"releases": [
 				{
 					"sublime_text": ">=3092",
-					"tags": true,
-					"previous_names": "Choice Script"
+					"tags": true
 				}
 			]
 		},
--- a/repository/c.json
+++ b/repository/c.json
@@ -950,22 +950,13 @@
 			]
 		},
 		{
-			"name": "Choice Script",
-			"details": "https://github.com/mrsonord/choicescript",
-			"releases": [
-				{
-					"sublime_text": "*",
-					"branch": "master"
-				}
-			]
-		},
-		{
-			"name": "ChoiceScript Syntax",
+			"name": "ChoiceScript",
 			"details": "https://github.com/fawkesy/choicescript-syntax",
 			"releases": [
 				{
-					"sublime_text": ">=3000",
-					"tags": true
+					"sublime_text": ">=3092",
+					"tags": true,
+					"previous_names": "Choice Script"
 				}
 			]
 		},


braver:
We already have a syntax for that (although a rather old one): https://packagecontrol.io/packages/Choice%20Script. Have you considered contributing to the existing package?

fawkesy:
> We already have a syntax for that (although a rather old one): https://packagecontrol.io/packages/Choice%20Script. Have you considered contributing to the existing package?

The original creator of that package hasn't been around since 2014, and my understanding was that I'd have to work with them if I wanted to tweak/modify their package? So unfortunately that hasn't been an option for me, and beyond that, I feel that there are quite a few differences between theirs and mine.

Theirs only highlights some ChoiceScript syntax, while mine highlights all of it, and mine also includes a syntax-specific color scheme and settings and language-specific autocompletes and auto-indents. It's a considerably more up-to-date and fully developed package for writing ChoiceScript in Sublime 3/4.

Also, I am new at this so I'm unsure of how much of a difference this makes, but their filetypes for the contents of their package are `.tmLanguage` and `.JSON-tmLanguage`, whereas I used the more recent `.sublime-syntax` and other Sublime-specific filetypes. There is remarkably little overlap between the actual content of our packages; in my opinion anyway, so I understand that must be taken with a grain of salt.

Additionally -- and this is a minor thing, I know -- the name of theirs is technically misspelled, as the language is called "ChoiceScript," no space, haha.

Given the large gap in time between these two packages and the difference in contents and function, I do feel that mine should still at least be considered to be added to Package Control.

That said, I am willing to do what is necessary to get this accepted! I can't contact @mrsonord about contributing to their package -- and the contents of mine would essentially overwrite theirs in the process of "expanding" upon it -- but if there are other steps I can take to reconcile our packages, please let me know! I really don't mean to be difficult!

braver:
Not difficult at all! If the existing package for the same functionality is defunct, and transferring the existing repo isn’t an option or even practical, we can replace the existing entry. That would also move all users over to the new package, everyone wins.

So, we already pinged @mrsonord here. The process now is that we give him 2 weeks to respond and maybe propose a different approach. After that time we should update this PR so that the existing package is removed, and your new package takes it over by using the “previous_names” key. 

In the mean time we can review your package. I’ll get back to that when I’m on an actual computer 😉

fawkesy:
Thank you so much! That sounds perfect to me, and I have no problem whatsoever waiting!

If we were to replace their package with mine, would it be possible to change the name? Just to remove the space in "Choice Script" so it's "ChoiceScript," as that's the proper name, haha.

braver:
Sure! Have a look at this example: https://github.com/wbond/package_control_channel/commit/95cfa498e13846be5e431065ce5d2bc45fb5bd0d

fawkesy:
@braver -- It's been two weeks, and @mrsonord hasn't objected. Where do we go from here?

braver:
You can update your PR here as follows:

- remove the existing entry for Choice Script
- add a previous_names key to your entry with the value `['Choice Script']`
- rename your entry to ChoiceScript (language support packages are simply named for the language, no "syntax" or other suffixes)

And your repo as follows:

- There is no need to ship a `.sublime-package` file. Please remove that, and move everything from the ChoiceScript folder to the root of the repo.
- For manual installation, people can then simply clone your repo into their Packages directory. Users should not manually move things to the `Installed Packages` directory. In fact you'd want to stay away from that dir entirely and leave that for packages installed via package control.
- You're shipping a color scheme along with the syntax. That's a bit of a red flag, you should be able cover all those cases in the syntax in a way that users can use any color scheme they like. In fact, I'm inclined to not accept a syntax package that ships with a color scheme. It pays off to put in some (usually not a lot) of effort to make sure that your syntax works with any color scheme. If your package finds any number of users, it will be the first thing they'll ask for because there are so many reasons people use one color scheme over the other in addition to simple preference (light/dark, contrast, the way their eyes work, what looks good on the monitor they use, etc. etc.). So you might as well do that before being accepted into package control.
- You might want to consider shipping the color scheme separately.

fawkesy:
@braver 
> There is no need to ship a `.sublime-package` file. Please remove that, and move everything from the ChoiceScript folder to the root of the repo.

Done!

> For manual installation, people can then simply clone your repo into their Packages directory. Users should not manually move things to the `Installed Packages` directory. In fact you'd want to stay away from that dir entirely and leave that for packages installed via package control.

I'll change the `readme.md` to reflect this, thank you!

> You're shipping a color scheme along with the syntax. That's a bit of a red flag, you should be able cover all those cases in the syntax in a way that users can use any color scheme they like.

The syntax works perfectly with other color schemes. The color scheme I'm shipping with it simply adds additional --- and, perhaps to some, extraneous --- highlighting, that does still work with other languages, though it was designed specifically for _this_ language. As it is right now, users have to manually choose to use my included color scheme, as it is not on by default. It's just an optional feature that's there if they want it, without needing to install something else.

The majority of people who develop with this language are generally not tech savvy --- for many, this is their first language after some very basic HTML --- so the fewer things they need to install, the better.

If you still believe that the color scheme should be separate, I can do that, but including it with everything else was a deliberate decision on my part.

---

Before I update the PR, I want to make sure that everything works for the packagecontrol-bot. I've already fixed the `.sublime-package` issue, but I'm having trouble with the following:
```
  - ERROR: Invalid JSON (with comments)
    - File: ChoiceScript/ChoiceScript.sublime-completions
    - Exception: Invalid control character at: line 8 column 26 (char 126)
```
This is, admittedly, my first time working with JSON in any capacity, so I'm unfortunately not very familiar with it. The character it is referring to is an `*` in reference to the language's own syntax. It's in a `.sublime-completions` file, and it is used as part of a "contents" string, where I include a snippet of the language's syntax.

It works perfectly fine on my end, and I've played with it quite a bit, so I'm unsure of what needs to be done to resolve this. I'm sorry for being such a bother with all this, but can you possibly point me in the direction of any documentation or anything that will help me figure this out?

Also:
```
  - WARNING: '.sublime-syntax' support has been added in build 3092 and there is no '.tmLanguage' fallback file
    - File: ChoiceScript/ChoiceScript.sublime-syntax
```
Will this be fixed when I update the PR to reflect that this package only works from that build of Sublime on?

Lastly, is it _required_ that I include a LICENSE.md, or simply strongly encouraged?

Thank you again so, so much for all your help and patience throughout this! I really do appreciate it!

braver:
> Done!

Excellent 👍🏻 

> It's just an optional feature that's there if they want it, without needing to install something else.

Alright, thanks for the explanation. My reasoning here is mostly this: if your color scheme is unique and good, you'd reach a lot more users if shipped separately. There is no requirement here, other than that a syntax should work well enough in any color scheme, which you've already covered. 

> Will this be fixed when I update the PR to reflect that this package only works from that build of Sublime on?

You can ignore that particular message, there is no need to support builds 3092 and older. We have a PR open to remove that feedback, but people have been busy with other things.

> Lastly, is it required that I include a LICENSE.md, or simply strongly encouraged?

Encouraged, not required. If your work is essentially open source, it's easy enough to put an MIT license on it. 

> Exception: Invalid control character at: line 8 column 26 (char 126)

That's odd. Your completions file isn't valid JSON because you have a `,` after the final property in each object (e.g. on line 10 after `"details": "Create a new *choice",`) but that's not the problem it seems to be complaining about (and ST has a custom JSON parser that can deal with this just fine). 

There is some good documentation now for these files: https://www.sublimetext.com/docs/completions.html#completion_files
But it doesn't cover this case, and it isn't mentioned in the _escaping_ section. If it works for you, and you fix it so it passes a general JSON validation (like [this one](https://jsonlint.com)), let's just ship it and figure that one out later if it does cause issues.

fawkesy:
I've fixed the `.sublime-completions` file! At least, it passes that JSON validation just fine! I'm pretty sure the issue is completely resolved (I hope so, anyway), because it seemed to have actually been entirely my fault. I'm glad I figured it out! Lesson learned.

I'll update the readme, add a license, then edit the PR some time today or tomorrow. Will I need to @ you for that once it's done?

braver:
I’ll get notified when you push anything to this PR, so do what you need to do and I’ll pick it ip from there 👌🏻

braver:
Take a look here for how to use previous_names: https://github.com/wbond/package_control_channel/blob/4e5928320dad7d0613a09b82987fafc30f458d43/repository/a.json#L6

braver:
Actually, it should be an array ;)

fawkesy:
Your patience with me and my many mistakes is so greatly appreciate!!

braver:
Alright, now that that’s all set, there are some things in the syntax we should look at. Sorry I didn’t get around to that before.

1. Are you sure you want to register for the txt extension?
2. Your scopes look fine (without really digging into the language), but should have a suffix specific to your syntax. That allows color schemes to provide tweaks for your syntax only. So “keyword.control.choisescript” for instance (IIRC there are more languages that would match “.cs”), instead of just “keyword.control”.

fawkesy:
> Are you sure you want to register for the txt extension?

I am.

> Your scopes look fine (without really digging into the language), but should have a suffix specific to your syntax. That allows color schemes to provide tweaks for your syntax only. So “keyword.control.choisescript” for instance (IIRC there are more languages that would match “.cs”), instead of just “keyword.control”.

Should I just add `.choicescript` to the end of all my scopes then?

braver:
> Should I just add .choicescript to the end of all my scopes then?

sounds like a plan 👍🏻

fawkesy:
All right, that's all done!!

Thank you again so, so much for all your help!! I am so grateful! And I have really learned a lot throughout this entire process.

braver:
No problem!

Final thing: you may want to update the selectors in your preferences to also include that new suffix. Something you can look at later, this is ready to go here 👌🏻

PR #8288, opened 2021-06-02
add python-black

thep0y:
A `black`-based package with only one formatting function may be more elegant than other formatting packages.
There is an additional function: use the command to generate a black configuration file.

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -3127,6 +3127,19 @@
 			]
 		},
 		{
+			"name": "python-black",
+			"details": "https://github.com/thep0y/python-black",
+			"labels": [
+				"formatter"
+			],
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Python2UnicodeFixer",
 			"details": "https://github.com/allieus/Python2UnicodeFixer",
 			"releases": [


braver:
Ah, that's better :)

If this is just a black formatter, that we're talking about, it would seem the same functionality is already provided by the existing [Formatter](https://packagecontrol.io/packages/Formatter). Although I can appreciate the elegance of a single purpose package, we do need to consider the options we present someone looking for a formatter, and how he/she is going to select one over the other. And also, instead of multiple maintainers putting effort into supporting the same (or greatly overlapping) functionality, they might be better off collaborating on the same package. 

As for your particular package, some first thoughts before going any deeper:
- You ship with keybindings. This is generally discouraged because they tend to conflict with other packages. In stead, you're usually better off supplying the command via the palette (all done here, 👍🏻) and in your README or elsewhere explaining how someone might add the keybindings themselves.
- You have a setting for max-line-length, and I wondered if it wouldn't be better to let that be handled by one of the config files most python tools already respect (a setup.cfg file for instance). 
- Indeed, those settings look like they map directly to command line arguments for black. Wouldn't it be easier if you had a single setting for such command line arguments, so that you don't have specifically map everything that black already understands?

thep0y:
I agree with you on "multiple maintainers putting effort into supporting the same (or greatly overlapping) functionality, Collaborative on the same package."

So if the package can't be uploaded to packagecontrol. io, I'll continue to use it. The configuration priority in sublime-settings is lower than that in pyproject.toml, and if pyproject.toml is present in the project, this configuration file will be used instead of the configuration in sublime-setting.

The original design was to read the file code, and then use the formatted code to overwrite the original code. Later, I found that I could directly call the command line to achieve this. For formatting a file, the command line is very simple. As mentioned in the previous paragraph, the configuration in sublime-text is not used when executing commands in a project with the pyproject.toml configuration file that black uses by default.

The configuration in sublime-setting is only used if there are no configuration files associated with black.

thep0y:
I agree with you on "multiple maintainers putting effort into supporting the same (or greatly overlapping) functionality, Collaborative on the same package."

So if the package can't be uploaded to packagecontrol. io, I'll continue to use it. The configuration priority in sublime-settings is lower than that in pyproject.toml, and if pyproject.toml is present in the project, this configuration file will be used instead of the configuration in sublime-setting.

The original design was to read the file code, and then use the formatted code to overwrite the original code. Later, I found that I could directly call the command line to achieve this. For formatting a file, the command line is very simple. As mentioned in the previous paragraph, the configuration in sublime-text is not used when executing commands in a project with the pyproject.toml configuration file that black uses by default.

The configuration in sublime-setting is only used if there are no configuration files associated with black.

braver:
It’s not that it can’t be added per sé, again I think a single use package is elegant. But if we do add it, it does need to pass a certain level. So that’s what this discussion is about, to find out what’s optimal here.

It does kinda sound like in essence most of what you do here can be accomplished by a build system. Have you considered formatting a buffer instead of a file on disk, ie to allow formatting without necessarily saving first?

thep0y:
> It’s not that it can’t be added per sé, again I think a single use package is elegant. But if we do add it, it does need to pass a certain level. So that’s what this discussion is about, to find out what’s optimal here.
> 
> It does kinda sound like in essence most of what you do here can be accomplished by a build system. Have you considered formatting a buffer instead of a file on disk, ie to allow formatting without necessarily saving first?

I am thinking of formatting the file buffer, it may be updated in a later version.

thep0y:
In version 0.0.3, I deleted the default binding key, deleted the duplicate configuration, and used the buffer to complete the formatting. But there is still only one formatting command.

thep0y:
> It’s not that it can’t be added per sé, again I think a single use package is elegant. But if we do add it, it does need to pass a certain level. So that’s what this discussion is about, to find out what’s optimal here.
> 
> It does kinda sound like in essence most of what you do here can be accomplished by a build system. Have you considered formatting a buffer instead of a file on disk, ie to allow formatting without necessarily saving first?

 thep0y commented 8 hours ago
In version 0.0.3, I deleted the default binding key, deleted the duplicate configuration, and used the buffer to complete the formatting. But there is still only one formatting command.

braver:
One approach to the keybindings problem is to retain your file and main menu entry, but to comment out your proposed binding. That way you give users something to discover within ST, and makes it easier to set up for them. Just a thought.

OK, looking good 👍🏻

jfcherng:
I would like to mention https://packagecontrol.io/packages/sublack since it seemed not to be mentioned here. I read these all conversations but I still think this is just a less-featured `black` plugin.

braver:
sublack has been discussed in a previous PR (this is the second attempt at adding this package). The approach here is different and much more simple (no daemons etc), and therefore we decided to add it to the channel.

PR #8289, opened 2021-06-03
Chromeless package: dropping linux support

ggets:


ggets:
Wonder if there is a way to include another release for linux in the array of releases, bound to an older release tag?

braver:
Platforms being inside the releases array it should be possible to put all Linux users on a prefixed tag. You should search a bit through the repo for an example, I’m sure it has been done before.

While you’re at it, you might also want to remove the readme key, it’s entirely superfluous since the README is in a normal location.

If you want to rename your package you need to provide “previous_names”.

braver:
That should do it, let's see how that goes down... I'm slightly worried your 2.x release would have already rolled out to Linux users. But you can always create a new tag. Your entry here is a-okay 👌🏻

PR #8292, opened 2021-06-07
Fix pkg name webAgent

seanfagan:
The casing on my package name was wrong (my bad). "webAgent" is correct.

### DIFF
--- a/repository/w.json
+++ b/repository/w.json
@@ -131,7 +131,7 @@
 			]
 		},
 		{
-			"name": "WebAgent",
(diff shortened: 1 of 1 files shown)


seanfagan:
See PR #8276 for more context.

braver:
We'll need to add a "previous_names" array with the previous name for this to work well.

seanfagan:
@braver Gotcha, I've added the previous name array. Thanks for your continued help

PR #8295, opened 2021-06-08
Package for Bunch.app on macOS

ttscoff:

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -1661,6 +1661,17 @@
 			]
 		},
 		{
+			"name": "Bunch",
+			"details": "https://github.com/ttscoff/Sublime-Bunch",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": ["osx"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Bunting Language",
 			"details": "https://github.com/xhacker/Bunting.tmLanguage",
 			"labels": ["language syntax"],


braver:
Thanks for submitting your package. We'll need some more context here though. What is the Bunch app? Why is it both a syntax and a color scheme, and a theme, and why does it also ship with keybindings and a bunch of commands (and what to the commands do?)? Why would someone navigating the package control site decide to install this? So many questions! 😄 You'll probably want to add some more information to the description of your repo, and some more information (and screenshots?) in your README probably also wouldn't hurt.

ttscoff:
I'll add more detail to the README, but here's the pitch:

[Bunch](https://bunchapp.co) is a Mac app that has its own "language" 
for automation on a Mac using plain text files. Its syntax is unique, so 
it needed a syntax definition, and the scopes I defined are unique (and 
admittedly poorly named), so it needed at least one theme that 
incorporated them to be useful.

The commands provide some autocompletion based on scope (including 
completion of application names available on the user's machine), 
navigation of Bunch documents, and snippets to complete some common 
Bunch commands.

Only people already using Bunch will have any use for this package, but 
for those who are, it will be a huge convenience.

Will update the README today.

ttscoff:
Updated README and repo description. 
https://github.com/ttscoff/Sublime-Bunch/

ttscoff:
Ok, I've pushed tag 1.0.5. I rebuilt the syntax scopes to work with more 
standard naming conventions, and it works with any theme now. I left the 
two (updated) custom themes in the package and included instructions for 
using them in the README, but the settings file no longer forces one. 
There are a couple of specific sub-scopes that are nice to style, and 
the included themes take advantage of that, but make no assumptions 
about the user's preferences.

I removed both of the keybindings that were set up and moved them into 
the README for the user to opt in to. Both of the commands just forced 
autocomplete to trigger with limited completions when entering specific 
scopes. Just convenience methods, user can add them if desired.

-Brett

braver:
Alright, that sounds a lot better already 👍🏻 

Are you sure you don't want to ship the theme separately?

ttscoff:
This package has a limited scope of appeal, and the themes are useless 
without the syntax definition, so it seems like asking users to install 
two packages would be less useful than just providing those who do 
install it with the option to use an included theme. Is there any real 
concern/drawback to including them?

rchl:
At least in ST4 the theme can be easily customized using `UI: Customize Theme` from the command palette. Providing instructions for how to add customizations specific to this package would be much better as then user could use his own theme with the benefits that you want to provide.

braver:
> the themes are useless without the syntax definition

that I don’t get… because the colors wouldn’t match otherwise?

ttscoff:
> At least in ST4 the theme can be easily customized using `UI: Customize Theme` from the command palette. Providing instructions for how to add customizations specific to this package would be much better as then the user could use his own theme with the benefits that you want to provide.

As in just list the specific scopes that could be added? I've actually never played around with `UI: Customize Theme`, but will take a look at it.

That said, I really don't see any issue with _also_ including an "opinionated" theme that the user is free to ignore.

ttscoff:
FYI, I revamped the included themes to be `.sublime-color-theme` JSON, which should make it easier to copy/paste bits into `Customize Theme`. Left the PLIST versions there as I'm not sure how backward-compatible the JSON format is.

braver:
> No, because they reference specific scopes like `punctuation.definition.embedded_snippet` to provide styling that `punctuation.definition` doesn't.

Oh, right... you have a _color scheme_ here, not a _theme_. A theme in Sublime Text terminology is the styling of the UI, the sidebar, search panels, etc. A color scheme is for the syntax highlighting. In that case, ignore my previous question, it totally makes sense to ship with a specific color scheme here. You'd want to get that corrected in your documentation though.

ttscoff:
Got it, my bad.

-Brett
________________________________

braver:
No problem, you're hardly the first to get confused by that 😄 Let me know when you've got that sorted and then we can ship this thing!

ttscoff:
Pushed an updated README.

braver:
don't forget your repo description (the "about" thing)

ttscoff:
> don't forget your repo description (the "about" thing)

Good catch, fixed.

braver:
🎉

rwols:
Wouldn't it be slightly better if https://github.com/ttscoff/Sublime-Bunch/blob/10998f277597798e0ed94b571e0fed4e5a0092c0/Bunch.sublime-syntax#L68 were `support.function` ?

rwols:
Forgot `.bunch` suffix here https://github.com/ttscoff/Sublime-Bunch/blob/10998f277597798e0ed94b571e0fed4e5a0092c0/Bunch.sublime-syntax#L291 (and below as well). I think that's all from me :)

braver:
You say you converted the tmThemes, but maybe you forgot to push them?

rwols:
> I guess I was a little confused by the description of the .support scope. Is it the preferred scope for builtins?

Yes. For built-in functions/callables/commands.

ttscoff:
Ok, took care of @rwols suggestions. 

I apparently had a global .gitignore for `*.sublime-*` that I forgot about. I assume it was supposed to prevent adding `-workspace` and `-project` files, but it ended up ignoring a couple of snippets and the `.sublime-color-scheme` files. Fixed now, everything should be there.

ttscoff:
Side note: @braver and @rwols I appreciate the attention to detail on this stuff. I added a package years ago (MarkdownEditing) but it must have been a lot more "wild west" back then as a ton of this stuff went unnoticed and had to get sorted out _after_ users had already started finding and installing it. I haven't created a package since then, so I'm no pro at this, and you guys have really helped.

braver:
Excellent. You can safely remove the tmTheme files, there probably aren't any installs of Sublime Text versions out there that will pick that over the newer sublime-color-schemes.

braver:
No problem. The amount of time we can put into a review varies from time to time, I guess you're lucky it's Friday and I have nothing better to do 😆 

🛥️

ttscoff:
> Excellent. You can safely remove the tmTheme files, there probably aren't any installs of Sublime Text versions out there that will pick that over the newer sublime-color-schemes.

Done.

PR #8296, opened 2021-06-16
delete "Sync Settings" support on st2

mfuentesg:


braver:
It's not really necessary, ST2 is considered deprecated implicitly. But thanks for the update!

PR #8298, opened 2021-06-16
Add 'Programatic Key Binding' plugin

VonHeikemen:
This plugin allows the user the ability to declare their keybindings using Python

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -2360,7 +2360,7 @@
 		},
 		{
 			"name": "Programatic Key Bindings",
-			"details": "https://github.com/VonHeikemen/SublimeProKeyBindings",
+			"details": "https://github.com/VonHeikemen/sublime-pro-key-bindings",
 			"donate": "https://www.buymeacoffee.com/vonheikemen",
 			"labels": ["keymap"],
 			"releases": [
--- a/repository/p.json
+++ b/repository/p.json
@@ -2359,6 +2359,18 @@
 			]
 		},
 		{
+			"name": "Programatic Key Bindings",
+			"details": "https://github.com/VonHeikemen/SublimeProKeyBindings",
+			"donate": "https://www.buymeacoffee.com/vonheikemen",
+			"labels": ["keymap"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Project PHP ClassBrowser",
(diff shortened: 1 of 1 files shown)


braver:
Hi, thanks for your submission. Sorry if I'm slow, but just so I understand what you're providing here, why would someone use your package to edit keybindings instead of editing them manually? Ie. your README does a pretty good job of explaining how everything works, but maybe an example use case would help "visualise" it :)

VonHeikemen:
For me the main benefit comes from the ability to use variables. I have a lengthy configuration which became a bit of a pain to manage in a json format. I wanted a way to store "context objects" to do something like this:

```python
def keybinding(bind):
  # Navigate the sidebar vim-style 
  bind(["h"], "move", sidebar_focused, by="characters", forward=False)
  bind(["l"], "move", sidebar_focused, by="characters", forward=True)
  bind(["k"], "move", sidebar_focused, by="lines", forward=False)
  bind(["j"], "move", sidebar_focused, by="lines", forward=True)

# Context objects
sidebar_focused = {
  "key": "control",
  "operand": "sidebar_tree"
}

#  more code ...
```

In this short example it may not look like a big deal but if you have 15 context objects (and many bindings), having them in variables with meaningful names it's really useful.

A nice side effect of this approach is that we can use functions, for loops, conditionals... so if you wanted to take it one step further, you could do it.

```python
def keybinding(bind):
  # use `ctrl+k` and `ctrl+j` to move up and down everywhere
  move(bind, up=["ctrl+k"], down=["ctrl+j"])

  # more bindings ....

def move(bind, up=[], down=[]):
  contexts = [
    sidebar_focused,
    overlay_focus,
    auto_complete
  ]

  for ctx in contexts:
    bind(up, "move", ctx, by="lines", forward=False)
    bind(down, "move", ctx, by="lines", forward=True)


# Context objects
sidebar_focused = {
  # ...
}

overlay_focus = {
  # ...
}

auto_complete = {
  # ...
}
```

braver:
Nice, sounds useful 👍🏻

braver:
Shouldn’t it be “programmatic”, double m?

VonHeikemen:
you're right, it should be double m. I'll fix that.

braver:
🤔 final consideration, because you can’t easily change it later: maybe the storage dir for the keybindings should be named after the package name (instead of the repo name)?

VonHeikemen:
ok. It has been updated.

PR #8299, opened 2021-06-16
Added package Vintage Relative Lines.

bicarlsen:

### DIFF
--- a/repository/v.json
+++ b/repository/v.json
@@ -394,6 +394,18 @@
 			]
 		},
 		{
+			"name": "Vintage Relative Lines",
+			"details": "https://github.com/bicarlsen/vintage_relnums",
+			"issues": "https://github.com/bicarlsen/vintage_relnums/issues",
+			"labels": [ "vintage", "line numbers" ],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Vintage Solid Caret",
 			"details": "https://github.com/muffinmad/VintageSolidCaret",
 			"author": "muffinmad",
--- a/repository/v.json
+++ b/repository/v.json
@@ -396,11 +396,10 @@
 		{
 			"name": "Vintage Relative Lines",
 			"details": "https://github.com/bicarlsen/vintage_relnums",
-			"issues": "https://github.com/bicarlsen/vintage_relnums/issues",
 			"labels": [ "vintage", "line numbers" ],
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": ">=3000",
 					"tags": true
 				}
 			]


Ultra-Instinct-05:
Are you aware that ST4 has a new setting `relative_line_numbers` to have relative line numbers ?

bicarlsen:
I had heard talk of it, but wasn't sure if it was implemented. Perhaps this is still a useful setting for those using ST3, though. Shall I set the build number to be `>3000` and `<4000`?

bicarlsen:
You can view the discussion about why I submitted to Package Control [here](https://github.com/bicarlsen/vintage_relnums/issues/1)

Ultra-Instinct-05:
> Perhaps this is still a useful setting for those using ST3, though

ST3 indeed doesn't have that ability so the package is useful for those wanting relative line numbers and want to use ST3. But with the public release of ST4, I don't see a point to use ST3 anymore.

> Perhaps this is also useful if people only want relative line numbers in vintage mode.

If there is some way to detect that Vintage mode is turned on, I am guessing you could enable `relative_line_numbers` and then disable it when a user comes out of Vintage mode (Just a thought).

But it looks like your package adds more features to it. I don't use vim or relative line numbers feature for daily work, so that's just my 2 cents however 🙂.

bicarlsen:
As @Ultra-Instinct-05 mentioned before, the original version of this plugin did not offer much functionality above the new ST4 option `relative_line_numbers`. However after making some significant improvements, I now believe it does offer enough additional functionality to be worth having a plugin for.
**Functionality**
 + Only active in Command mode, where switching between line numbering modes is more important.
 + Easy toggling between phantom and native line numbers.
 + Easy toggling between `hybrid`, `relative`, and `absolute` number modes using key bindings. Toggling is smart, detecting if phantoms or native lines are being used.
 + When using phantoms, custom coloring and markings are used for the current lines, lines above, and lines below. Other settings for numbering span, font face, and padding give more styling options.

repository/v.json
```diff
@@ -393,6 +393,18 @@
 				}
 			]
 		},
+		{
+			"name": "Vintage Relative Lines",
+			"details": "https://github.com/bicarlsen/vintage_relnums",
+			"issues": "https://github.com/bicarlsen/vintage_relnums/issues",
```
braver, line 399: Because your issues are in the default location, please don't link to them here (ie. remove this line)

repository/v.json
```diff
@@ -393,6 +393,18 @@
...
+			"labels": [ "vintage", "line numbers" ],
+			"releases": [
+				{
+					"sublime_text": "*",
```
braver, line 403: `"> 3000"` would probably more correct here?

repository/v.json
bicarlsen: Thank you for the input. I'll make the change.

repository/v.json
bicarlsen: Yes, thank you. I hadn't read through the ST2 Docs yet, but it appears it isn't compatible with it. I'll make the change.

braver:
Alright. Some final details to consider:

- We usually advise against shipping with keybindings because they tend to conflict with other packages. You can opt to not ship with any and instead refer to the README so that users can learn how to set them up themselves. It doesn't seem essential to me to have the available commands under a keybinding, but I may be wrong. 
- You have some PNG's in the docs. You can exclude such non-functional files from the package (saving on storage, bandwidth and CO2) using a `.gitattributes` file.

bicarlsen:
> Alright. Some final details to consider:
> 
> * We usually advise against shipping with keybindings because they tend to conflict with other packages. You can opt to not ship with any and instead refer to the README so that users can learn how to set them up themselves. It doesn't seem essential to me to have the available commands under a keybinding, but I may be wrong.
> * You have some PNG's in the docs. You can exclude such non-functional files from the package (saving on storage, bandwidth and CO2) using a `.gitattributes` file.

Both these issues have been addressed in the newest push.

Tha k you for the guidance,

braver:
Indeed. Thanks! One detail: "Default Key Map" would then become "Suggested key map"? You can finish that up later, I'll merge this right now.

bicarlsen:
Just fixed that as well :).

PR #8300, opened 2021-06-17
Add SubGrep plugin

houcheng:
Hi,
I added a plugin for grepping based on current selection text or regular express.
Thank you.


houcheng:
Hi,
Release tag is added, please pull. Thank you.

PR #8301, opened 2021-06-23
Adding ric-script syntax highlighter

Ricardicus:
I have a programming language under development that I am very proud of called [ric-script](https://github.com/Ricardicus/ric-script).

I am also a huge fan of Sublime Text and use it everyday. 

The link that i added to the file leads to this repo: [https://github.com/Ricardicus/ric-script-syntax](https://github.com/Ricardicus/ric-script-syntax)

I have added a .sublime-syntax file at the root of a github repo, it can be viewed at https://github.com/Ricardicus/ric-script-syntax/blob/master/ricscript.sublime-syntax .
It should be used for syntax highlighting only.

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -1584,7 +1584,7 @@
 		},
 		{
 			"name": "ricscript",
-			"details": "https://github.com/Ricardicus/ric-script",
+			"details": "https://github.com/Ricardicus/ric-script-syntax",
 			"labels": ["language syntax"],
 			"releases": [
 				{
--- a/repository/r.json
+++ b/repository/r.json
@@ -1583,7 +1583,7 @@
 			]
 		},
 		{
-			"name": "ric-script",
+			"name": "ricscript",
 			"details": "https://github.com/Ricardicus/ric-script",
 			"labels": ["language syntax"],
 			"releases": [
--- a/repository/r.json
+++ b/repository/r.json
@@ -1583,6 +1583,17 @@
 			]
 		},
 		{
+			"name": "ric-script",
+			"details": "https://github.com/Ricardicus/ric-script",
+			"labels": ["language syntax"],
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


Ricardicus:
Just added a semver tag based on feedback from the packagecontrol-bot

braver:
Does your language have users external to your team such that it adds practical value to have a package control entry?

Ricardicus:
Honestly, to the extent of my knowledge there aren't any external users that I don't know personally yet :)
It seems that I am trying my luck here I suppose.

braver:
Well, having your package in the public repository isn't required, there are tons of other ways to distribute packages. It does however come with some extra responsibility etc. Your README and everything is very limited, and all together it doesn't look like this really needs to be in here. 
I mean, it's fine and we can add it, just to be sure that's really what you want. In any case you should remove the `ricscript-syntax.sublime-project`, for instance via `.gitattributes`.

Ricardicus:
I understand, and you are right; it doesn't need to be there. Apart from rocking my ric-script world, instilling me with a sense of accomplishment and motivation, not much would change under the sun if it was added. I might be a little bit greedy at this point but yes, I want it there. I removed the mentioned file and I also added a license, which was missing. As you mentioned; there are other ways to make this work without adding it to the source repo. However, it does require some extra steps it seems like but it will be the way of doing this if it will not pass. I won't be devastated if you don't let this in, don't worry about it, but I might try again at a later point when the project has grown even further.

Ricardicus:
I kind of don't feel like it doesn't belongs now too, haha.
Looking at other projects, they typically seem to have a massive community
behind them. I might have jumped the gun.

braver:
Well we do have some smaller projects here too, and I totally understand the cool factor. No decision here is final either,  you can always come back and we’ll reconsider.

PR #8302, opened 2021-06-25
Add IRC log Syntax package

barnumbirr:
IRC log file syntax highlighting.

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -1117,6 +1117,17 @@
 			]
 		},
 		{
+			"name": "IRC Syntax",
+			"details": "https://github.com/barnumbirr/sublime-irc",
+			"labels": ["language syntax", "IRC"],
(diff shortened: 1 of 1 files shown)


braver:
Are you sure you want to register for the txt extension?

barnumbirr:
It's not ideal but most IRC client export their log files to plain text.

braver:
I don’t know exactly how it behaves, but as long as it doesn’t automatically becomes the default for txt files that’s fine. That would be annoying. And that might be a good reason to not set the extension, or something irc-specific.

barnumbirr:
I've now released [v1.0.1](https://github.com/barnumbirr/sublime-irc/releases/tag/1.0.1) which removes file extension.

braver:
Excellent 👌🏻

PR #8303, opened 2021-06-25
Add "Simple Fuzzy" package

ukyouz:
Fuzzy search line-based jump for the current file and project files.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1840,6 +1840,16 @@
 			]
 		},
 		{
+			"name": "Simple Fuzzy",
+			"details": "https://github.com/ukyouz/SublimeText-SimpleFuzzy",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Simple Ifdef",
 			"details": "https://github.com/rumly111/simpleifdef",
 			"releases": [
--- a/repository/s.json
+++ b/repository/s.json
@@ -1844,7 +1844,7 @@
 			"details": "https://github.com/ukyouz/SublimeText-SimpleFuzzy",
 			"releases": [
 				{
-					"sublime_text": ">=3000",
+					"sublime_text": ">=4107",
 					"tags": true
 				}
 			]


TheSecEng:
Looks like a neat package. I have a few initial reactions which are pretty minor

1. You have a print statement in the code. Generally we prefer package authors to implement a debug logic to allow users to enable or disable console logging for information. Please either remove the print statement or look at debugging logic. https://github.com/ukyouz/SublimeText-SimpleFuzzy/blob/00f34a848a79c5faad6caf4f090d8eda77cb2de5/SimpleFuzzy.py#L93
2. You have several blank lines at the end of your file. Generally it's good practice to have 1 empty line at the end of files. 
3. You have `ripgrep` as required, but within your codebase you have fallbacks for this logic. Perhaps you can update the readme to include something to the effect of `Ripgrep is optional but highly recommended for optimal performance, git is used as a backup function, and lastly pythons os library is used as a last resort`
4. Since you have built in fallbacks for functionality, potentially let the enduser specify the preferred method via a settings file. If they want this plugin to prefer git over os or rg or vice versa

ukyouz:
Thanks for your kindly reply and I have fixed all of them at version `1.1.1`.

1. Add debugging logic by window command `simple_fuzzy_debug_toggle`
2. Remove all the tailing empty lines
3. Update `README.md`
4. Add user setting key `simple_fuzzy_ls_cmd` for customization

Hopefully it is useful for all Sublime Text users :)

TheSecEng:
LGTM

jrappen:
While you are at it, you could make this require Build 4107 or later and use python 3.8. Below a list of changes from `py3.3` to `py3.8`, but looking at your code I assume you knew that.

Up-to-date docs are at: <https://www.sublimetext.com/docs/index.html>

---

#### asyncio

- Futures.
- Coroutines with `async` / `await`.
- `async for` / `async with`.
- Async generators.
- Event loop.

#### Type hints

- `typing` module.
- Variable annotations.
- postponed evaluation: `from __future__ import annotations`.

#### Other

- Formatted string literals (`“f-strings”` vs `f"Hello, {name}!"`).
- Matrix multiplication operator (`a @ b`).
- Unpacking is much more flexible.
- Various internal changes to module loading.
- Dictionaries are now ordered by default. `**kwargs` also preserve order.
- Assignment expressions (“walrus operator” `:=`).
- Positional-only parameters.
- dataclasses.
- enum.
- `pathlib`.

---

In:

https://github.com/ukyouz/SublimeText-SimpleFuzzy/blob/55ace2a56294caa2991d988e58d4e75f4352ca70/SimpleFuzzy.py#L42-L44

you might want to get the current active view just once.

ukyouz:
OK, I noticed that some sublime's API are relatively new (ie. sublime.ListInputItem), so I updated the required version to 4107. And yep, I indeed only need to get current active view once :)

PR #8305, opened 2021-07-01
Update URL for Nextflow package

peterk87:
Updating URL for Nextflow package from https://github.com/peterk87/sublime-nextflow to https://github.com/nf-core/sublime/

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -440,7 +440,7 @@
 		},
 		{
 			"name": "nextflow",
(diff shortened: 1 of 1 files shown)


braver:
Wouldn’t it be cleaner to transfer the repo instead of forking it? Anyway, just double-checking here, let me know if all is in order and I’ll push the button.

peterk87:
Hi @braver everything is in order. Forking the repo to the nf-core organization was okay with the org maintainers. Thanks for checking.

PR #8306, opened 2021-07-03
Update URL for 'RustAutoComplete'

typowritter:
The original repository of 'RustAutoComplete' has been transferred, updating the URL.


braver:
It hasn't been transferred, it's forked. Which is not optimal, but also works. Can you point me to a confirmation that the original author is ok with this?

braver:

PR #8308, opened 2021-07-04
Add StyleLint-Formatter, tailwindcss Class Sorter

LetsZiggy:
I'm sorry for the joint commits. I intended for a separate pull request but absentmindedly pushed a new commit while this pull request is open.

**StyleLint-Formatter**
- Uses stylelint / stylelint_d to format stylesheets
- Forked from and inspired by [ESLint-Formatter](https://github.com/TheSavior/ESLint-Formatter)

**tailwindcss Class Sorter**
- Sorts tailwindcss classes according to selected sorting type (eg: recess)
- Sorts each class variants according to recommended variant order in [tailwindcss doc](https://tailwindcss.com/docs/configuring-variants#ordering-custom-variants)
- Sorts similar functioning classes according breakpoints

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -3990,6 +3990,18 @@
 			]
 		},
 		{
+			"name": "StyleLint-Formatter",
+			"details": "https://github.com/LetsZiggy/StyleLint-Formatter",
+			"issues": "https://github.com/LetsZiggy/StyleLint-Formatter/issues",
+			"labels": ["css", "formatting", "stylelint"],
+			"releases": [
+				{
+					"sublime_text": "*",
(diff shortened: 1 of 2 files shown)


jrappen:
* I assume you should rather split this PR into two. One each per added package.

---

* please add `.gitattributes` files as in <https://github.com/sublimelsp/LSP/blob/main/.gitattributes> to NOT ship all those helper files to end users receiving the package via Package Control

LetsZiggy:
Okay.
I'll add `.gitattributes` and resubmit pull request individually.
Thanks

PR #8310, opened 2021-07-05
Adding Looplex Lawtex Plugin to Package Control Channel

Ivo-Deus-Looplex:
This package is the public release of our ST3 plugin for Lawtex scripts, our proprietary language to write smart documents. We've been developing and utilizing this package internally for quite a while by manual installations and our goal now with the publication of this package is to not only facilitate the installation of this plugin for our team, but to also make it readily available for our clients.


jrappen:
* Please add a license file.
* Recent Sublime Text Build is 4109, you might want to review your code. Compare: <https://www.sublimetext.com/>
* Recent sublime syntax version is 2. Compare: <https://www.sublimetext.com/docs/syntax.html#ver-dev>
* You might want to add a github action workflow. Compare https://github.com/sublimehq/Packages/blob/master/.github/workflows/ci.yml
* You might want to break the regex lines in your syntax file for greater legibility. Compare the default packages at: <https://github.com/sublimehq/Packages>

PR #8312, opened 2021-07-07
Add SubGrep plugin

houcheng:

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -4093,6 +4093,17 @@
 			]
 		},
 		{
+			"name": "SubGrep",
+			"details": "https://github.com/houcheng/SubGrep",
+			"labels": ["grep", "regular expression", "search", "find"],
+			"releases": [
+				{
+					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


jrappen:
You might want to add a `.gitattributes` file to **NOT** ship the included GIF via Package Control.

Random example:

<https://github.com/sublimelsp/LSP/blob/main/.gitattributes>

houcheng:
> You might want to add a `.gitattributes` file to **NOT** ship the included GIF via Package Control.
> 
> Random example:
> 
> https://github.com/sublimelsp/LSP/blob/main/.gitattributes

Hi,
I've added the file. Thank you.

braver:
We don't recommend shipping with keybindings, especially for packages that don't completely need them. They tend to conflict with those of other packages, or even existing defaults. Just make sure your commands are in the command palette, that's enough for most users. Then maybe document how to make the bindings in the README, so users can choose their own bindings.

braver:

PR #8313, opened 2021-07-08
Add Hippie Autocomplete

Suor:
Rebinding `Tab` is the whole point. Some context on the plugins goal is [here](https://forum.sublimetext.com/t/st3-style-autocomplete-in-st4/57774).

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -627,7 +627,7 @@
 			"labels": ["auto-complete"],
 			"releases": [
 				{
-					"sublime_text": ">4000",
+					"sublime_text": ">4050",
 					"tags": true
 				}
 			]
--- a/repository/h.json
+++ b/repository/h.json
@@ -622,6 +622,17 @@
 			]
 		},
 		{
+			"name": "Hippie Autocomplete",
+			"details": "https://github.com/Suor/sublime-hippie-autocomplete",
+			"labels": ["auto-complete"],
+			"releases": [
+				{
+					"sublime_text": ">4000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Hipster",
 			"details": "https://github.com/bjarneo/hipster",
 			"releases": [


kaste:
I think this is too idiosyncratic atm. 
 
- The `<tab>` binding here conflicts for example when the user has other plugins installed which also want to use `<tab>`.  For example, Terminus' autocomplete or GitSavvy.  This is although these other plugins bind with a very specific context because the plugin here binds with a super low specificity.  (Then order of definition matters which *for the end-user* is basically random. Technically it's sorted alphabetically, not sure if the python environment, 3.3 vs 3.8, counts as well.)

- The `tab` should not bind for the Goto panels.  It can bind for input panels but in that case the index for the `view` (which is the input handler view) is almost useless.

- The "indexer" or completions works for some languages but not for example for "CSS".  This is because you use `view.word` but the indexer just splits on `\w`.  (E.g., `\w` does not split on `-` but `word()` does on some syntaxes.)  I don't know the correct solution here; for consistency it would be: when you use `word()` also use `word_separators`.  That should work for any language, think clojure, haskell, languages which look really different, t.i. not like a c-derivative, and allow any character in an identifier.  Maybe something like `missing?` in ruby would be captured as expected.  Maybe a completely different, simpler approach would be to just and only include the `-`, and then not use `word()` but a custom implementation. 

- An evergrowing global index and re-indexing on `deactivate` (t.i. for example when you open a Goto panel) seems very inefficient.

On a side-side-side note:  "hippie" autocomplete is a thing in emacs, but a different thing.  That might be confusing.  This here is just autocomplete word from buffer, both emacs and vim have these built-in;  iirc, called `dabbrev` in emacs.

I just pushed a `pu` branch https://github.com/kaste/sublime-hippie-autocomplete/tree/pu which tries to solve some of these issues.

Suor:
Thanks you for taking a look.

However I don't want it to work only in Python, I use it in lots of file types including plain text ones, and want them to share words. And in Python file I expect it to work in strings and comments too.  That's why `Tab` needs to be bound. We may exclude more contexts, but unfortunately there is no way to set priority. BTW, this is the way autocomplete was working in ST2/3 and the way it was set up.  

I've considered possible performance issues, but never really met any. Reindexing is so fast that could be done on any change, as in the original snippet from Sublime forum thread, reindexing on deactivate is even less of an issue. Collecting fuzzy matches may be made lazy, I never did that because it was never slow. Theses optimizations are open but not necessary. Making global index to not grow will require more complex structures, maybe storing a word set for each view, it's possible but I don't think it's really needed.

The meaning of the word might be changed, I don't do CSS nor Lisp now so didn't bother so far. But respecting word characters should be the easiest and good enough way. However, then inserting words across scopes might become less useful, which we can solve by making plugin smarter, but at some point autocompletions may become less predictable, which goes against the original goal of making it fluent.

repository/h.json
```diff
@@ -621,6 +621,17 @@
 				}
 			]
 		},
+		{
+			"name": "Hippie Autocomplete",
+			"details": "https://github.com/Suor/sublime-hippie-autocomplete",
+			"labels": ["auto-complete"],
+			"releases": [
+				{
+					"sublime_text": ">4000",
```
jrappen, line 630: This should be `>=4050` (first public dev build) or `>=4107` (first public stable build) at least.

Suor: Fixed

braver:
Sorry for letting this sit here for so long, we've been struggling with capacity and also with the question of how to approach a package like this. Seems to this is fine to ship, so let's 🏎️

Suor:
Thanks.

PR #8316, opened 2021-07-11
Added "EXAPUNKS Syntax" package

cafuneandchill:
A package for [Sublime Text 3](https://www.sublimetext.com/) that provides syntax highlighting for the assembly language used in the game [EXAPUNKS](http://www.zachtronics.com/exapunks/).

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -1432,16 +1432,6 @@
 			]
 		},
 		{
-			"name": "EXAPUNKS Syntax",
-			"details": "https://github.com/cafuneandchill/exapunks-syntax",
-			"releases": [
-				{
-					"sublime_text": "*",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "Exalt",
 			"details": "https://github.com/eerohele/exalt",
 			"labels": ["xml", "html", "xslt", "validation", "formatting"],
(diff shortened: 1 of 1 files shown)


braver:
Syntaxes do much more in Sublime Text than just highlighting. And packages for specific languages usually provide more than just the syntax. Therefore, simply name your package after the language it supports and leave out words like “syntax” or “highlighting”.

The current version of ST is 4, and your package probably also supports that. Maybe just leave out the version in your documentation.

For manual installation I would never recommend putting stuff in the User package. That’s for overrides, or personal stuff. Simply instruct people to clone or download your repo into the Packages dir.

cafuneandchill:
Thank you for the recommendations! I never would have thought that naming a package would have such specifics.

> Syntaxes do much more in Sublime Text than just highlighting. And packages for specific languages usually provide more than just the syntax. Therefore, simply name your package after the language it supports and leave out words like “syntax” or “highlighting”.
> 
> The current version of ST is 4, and your package probably also supports that. Maybe just leave out the version in your documentation.
> 
> For manual installation I would never recommend putting stuff in the User package. That’s for overrides, or personal stuff. Simply instruct people to clone or download your repo into the Packages dir.

braver:
Naming things, cache invalidation and off by one errrors 😉

PR #8317, opened 2021-07-12
Changed the position of ST2 & ST4, to fix the logic issue

Doi9t:
Hello,

I just changed the order of ST2 & ST4 to prevent ST2 user of receiving the latest version of my plugin (since `<3000` and `<4000` are conflicting).

I wanted to have 3 tags, should I use only ST2 & ST4 ?

Thanks!


braver:
You’d want to set a range of versions then for st3. You could also consider just dropping st2 since nobody uses that anymore. And for ST4, maybe just set tags to “true”?

Doi9t:
Thanks, I choose to keep ST2, since I have already support for it; Added the range `3000 - 3999` to `st3`

PR #8318, opened 2021-07-15
Add package HSX Syntax Highlighting

jpe90:
This package enables HTML syntax embedded inside [HSX quasiquotes](https://ihp.digitallyinduced.com/Guide/hsx.html) in Haskell source code to be highlighted. All HTML is currently highlighted as a string.

Before:
2021-07-15-112133_1920x1080_scrot
After:
2021-07-15-111738_1920x1080_scrot

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -986,6 +986,17 @@
 			]
 		},
 		{
+			"name": "HSX Syntax Highlighting",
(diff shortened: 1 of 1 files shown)


jpe90:
Identified some issues while using this that need to be cleaned up, sorry for the noise. Closing this PR until I figure out a better way to implement this functionality.

PR #8319, opened 2021-07-18
Add ryanolsonx/sublimetext-zenburn-theme

ryanolsonx:
Ported the zenburn color scheme to sublime text. I have a VS Code version and would love to provide this in package control for sublime text users. This is vastly different than the existing zenburn theme. My theme is a lot closer to the Vim color scheme (that this was ported from).

### DIFF
--- a/repository/z.json
+++ b/repository/z.json
@@ -104,6 +104,17 @@
 			]
 		},
 		{
+			"name": "Zenburn Color Scheme",
+			"details": "https://github.com/ryanolsonx/sublimetext-zenburn-theme",
+			"labels": ["color scheme"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Zenoss",
 			"details": "https://github.com/cluther/SublimeZenoss",
 			"releases": [


braver:
The existing package is actively maintained, why not open a PR there?

ryanolsonx:
Okay - I'll work with that package author to see if I can incorporate my changes into that repo. If not, would we be able to get this in as a separate package, @braver?

braver:
Sure, there is nothing fundamentally against multiple versions of the same idea, but all around it’s better if we can work together and present users with a single really good version.

braver:
I’ll close this for now, just ping me to let me know if you worked it out or if we need to look at adding your version as well.

ryanolsonx:
@braver  So far no response from @colinta after 3 days. I created an issue https://github.com/colinta/zenburn/issues/8 and a PR https://github.com/colinta/zenburn/pull/9. If there's no response by Jul 27th (after 1 week), would you mind re-opening this?

braver:
We usually keep a 2 week timeout period for these things. @colinta seems to regularly check in in GitHub, so let's give him a little time.

rwols:
Three days is very short for people doing this in their spare free time ;)

ryanolsonx:
@rwols I was thinking 1 week would be reasonable -- not 3 days.

I'm happy to wait 2 weeks 😃

ryanolsonx:
@braver Do you mind re-opening this?

braver:
Alright, let's do this. 

- You have quite a few files that aren't functional to the package. You can exclude them from the package using [.gitattributes and export-ingore](https://github.com/SublimeLinter/SublimeLinter/blob/master/.gitattributes), saving some storage bits and CO2 ;) 
- About the name, I think it should signal this is the new and improved Zenburn. "Zenburn New", "Zenburn Improved", something like that?
- And have a look at the feedback from the bot.

ryanolsonx:
@braver  Alright - so I've fixed up a lot of the issues. 

- I removed the "demo" directory that wasn't really needed.
- It has a license file now.

Let's talk about the name. I don't feel like the existing Zenburn color theme is really "Zenburn." Mine is a more accurate port of what Zenburn is all about and really matches up with what was originally created for Vim. With that said, would we be able to keep my proposed name? I feel like it would be misleading to users for them to see Zenburn Improved or something like that when I haven't changed the nature of the Zenburn theme itself (and as it's not a variant of Zenburn).

Thoughts?

braver:
Well you still have a lot of screenshots in the package, I would encourage you to use the `.gitattributes` for that. I also kinda doubt that they're useful to new users.

I agree with your reasoning about the name though, let's keep it as it is.

ryanolsonx:
Awesome.

I'll remove those screenshots. It should be much better now.

ryanolsonx:
Thanks a lot @braver!

PR #8321, opened 2021-07-21
Added Looplex Lawtex Plugin

Ivo-Deus-Looplex:
This package is the public release of our ST3 plugin for Lawtex scripts, our proprietary language to write smart documents. We've been developing and utilizing this package internally for quite a while by manual installations and our goal now with the publication of this package is to not only facilitate the installation of this plugin for our team, but to also make it readily available for our clients.

We had a [previous pull request](https://github.com/wbond/package_control_channel/pull/8310) with some feedback that has been addressed:
- Added a license file in Portuguese (Brazil), which will be translated into English at a later date.
- Code has been reviewed for Sublime Text Build number 4109, with no apparent compatibility issues. We have, then, upgraded our version requirements to Sublime Text 3 version 4.0 since we've opted to develop the plugin further using this version as reference.
- Lawtex syntax file reviewed for version 2, with no apparent compatibility issues. We've made some changes regards the regex expressions, and we will upgrade the file with changes as they become necessary.
- We've opted not to include a GitHub action workflow at the moment, but will include one in the future.
- We've managed to break the regex expressions by using variables, but whenever we've attempted to make multi-line expressions (with the '|-' prefix much alike the many examples provided), ST3 would find errors reading our syntax file. We've left the multi-line expressions commented and if there's any feedback regards this issue, that'd be much appreciated.

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -1588,6 +1588,22 @@
 			]
 		},
 		{
+			"name": "Looplex Lawtex Plugin",
+			"details": "https://github.com/looplex/looplex_lawtex_ST3_plugin",
+			"homepage": "https://www.looplex.com.br/",
+			"author": "Looplex S.A.",
+			"labels": [
+				"looplex",
+				"lawtex"
+			],
+			"releases": [
+				{
+					"sublime_text": ">=4109",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "LoremGibson",
 			"details": "https://github.com/erdostom/LoremGibsonSublime",
 			"releases": [


jrappen:
* In general, I'm not sure copy-pasting the license file from your website and/or other services is a good idea. As it is a waste of time and energy for me personally to read through your license, I assume others won't as well. Custom licenses are a first blocker for potential contributors regardless of whether or not they actually allow contributions. I also assume you're neither interested in contributions nor grant anyone the right to do so in your license anyway, so take this as a general advice and do whatever works for you.
* I won't thoroughly review a plugin shipping a closed source binary by installing it on any of my machines. Nor would I advise anybody else, including your clients. But that's me. Whatever works for you.
* It's neither `Sublime Text 3 Build 4109` nor `Sublime Text 4 Build 4109` but simply `Sublime Text`, as SublimeHQ has switched to a subscription model for licensing.

---

For your YAML indentation problem with the syntax file compare <https://yaml.org/spec/1.2/spec.html> and an example:

image

You might want to install the PackageDev package:

<https://github.com/SublimeText/PackageDev>

---

I'm not sure why your package should be on Package Control channel. Adding instructions for a **manual installation** would've made more sense in my opinion, as your clients can still update the package through Package Control without it actually being made publicly available on the channel.

braver:
Your README doesn't explain what your package does. Does it really need to be public?

Ivo-Deus-Looplex:
Hey, sorry for the delayed response; and yes, after discussing the feedback from these pull requests (and being informed that is it possible to update the package via Package Control even if it is installed manually), we agree that keeping the plugin as a manual installation is the best course of action.

However, this process of updating the package through Package Control was new to us until jrappen kindly mentioned it, so if anyone has some documentation or references we can look into in order to apply it to our plugin, that would be appreciated. Either way, thanks for your feedback and hope you guys have a nice week!

braver:
You might want to have a look at the "Add repository" section of the [Package Control Usage docs](https://packagecontrol.io/docs/usage). You can add GitHub repos directly and users will receive any updates to the repo just like they would for packages managed via the package control channel.
If you have multiple packages, you can even set up your own channel by simply hosting a JSON that lists the packages on a public URL. 

I'll close the PR. If you have any questions, don't be afraid to ask!

PR #8323, opened 2021-07-22
Add Thread Dump Analyser plugin

vasugr:
This is a thread dump analyser plugin, which takes a thread dump text file as an input and produces analytic data. I have made similar plugin for VScode as well , you can find it [here](https://marketplace.visualstudio.com/items?itemName=Vasu.tda)

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -2234,6 +2234,20 @@
 			]
 		},
 		{
+			"name": "Thread Dump Analyser",
+			"author": "Vasu",
+			"details": "https://github.com/vasugr/Thread-Dump-Analyser",
+			"readme": "https://raw.githubusercontent.com/vasugr/Thread-Dump-Analyser/main/README.md",
+			"issues": "https://github.com/vasugr/Thread-Dump-Analyser/issues",
+			"labels": ["text manipulation", "formatting","Thread dump"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "ThreatConnect Exchange Snippets",
 			"details": "https://github.com/fhightower/tcex-snippets",
 			"labels": ["snippets"],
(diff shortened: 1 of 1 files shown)


repository/t.json
```diff
@@ -2233,6 +2233,20 @@
 				}
 			]
 		},
+		{
+			"name": "Thread Dump Analyser",
+			"author": "Vasu",
+			"details": "https://github.com/vasugr/Thread-Dump-Analyser",
+			"readme": "https://raw.githubusercontent.com/vasugr/Thread-Dump-Analyser/main/README.md",
+			"issues": "https://github.com/vasugr/Thread-Dump-Analyser/issues",
```
braver, line 2241: Your README and issues are in the default location, so you don't need to link to them here (ie. please remove these lines)

vasugr: Thanks. I have made the changes.

rchl:
Just out of curiosity, is the thread dump format universal? If not then would it make sense to specify what is supported?

braver:
We don't recommend shipping with keybindings, especially for packages that don't completely need them. They tend to conflict with those of other packages, or even existing defaults. Just make sure your commands are in the command palette, that's enough for most users and usually as good as or even better than keybindings (in any case a lot easier to discover). Then maybe document how to make the bindings in the README, so users can choose their own bindings.

You have some .DS_Store files in your repo, usually you want to have those covered by `.gitignore`.

You have some PNG's in the docs. You can exclude such non-functional files from the package (saving on storage, bandwidth and CO2) using a .gitattributes file.

Python files in subdirectories are not hot-reloaded, making it a bit harder during development. It doesn't look like you really need a "models" directory right now?

braver:

vasugr:
I'll make the aforementioned changes by month end.

braver:
September is gone. Any news?

vasugr:
Thanks for the reminder. I will complete it by 2 weeks. Thank you

braver:
Please let us know when you're ready, we can re-open the PR then.

PR #8325, opened 2021-07-24
Add Cottle Syntax

deathaxe:
This commit adds Cottle a text-to-speach scripting language for Elite:Dangerous.

The syntax was developed as result of a request at

- https://stackoverflow.com/questions/67027410/sublime-text-3-custom-syntax-for-cottle-hard-to-start
- https://forum.sublimetext.com/t/problem-developing-syntax-highlight-for-cottle/57051

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3706,6 +3706,17 @@
 			]
 		},
 		{
+			"name": "Cottle",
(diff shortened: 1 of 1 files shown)


braver:
The common label for a syntax would be "language syntax".

braver:
🎉

PR #8326, opened 2021-07-24
Add SurfaceSyntax

cschmatzler:
Adds basic syntax highlighting for [Surface](https://surface-ui.org/), the component library for Elixir's Phoenix web framework.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -5171,6 +5171,16 @@
 			]
 		},
 		{
+			"name": "SurfaceSyntax",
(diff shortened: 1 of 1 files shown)


braver:
We already have some Elixir packages that cover some of the related syntaxes. Would it make sense to add it to one of those existing projects? If we publish it separately please remove "syntax" from the name and simply name it after the language. You may also want to look into adding some tmPreferences for things like comments. I noticed from the readme you would like to embed the Markdown syntax, do you need some pointers there?

PR #8327, opened 2021-07-25
Update MarkdownEditing

deathaxe:
Appears Package Control has issues with previous version checks.

ST3176 doesn't see any tags but only branches, which makes it impossible for users on it to opt-in to alpha releases.

Hence a `st2-2.2.10` tag was added to the most recent commit which is compatible with ST prior to ST3176.

This commit registers `st2-` prefix for all ST builds from 2000 to 3175.

Hope, ST3176 sees tags then.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -622,8 +622,8 @@
 			"labels": ["markdown", "github markdown", "gfm", "multimarkdown", "color scheme"],
 			"releases": [
 				{
-					"sublime_text": "<3176",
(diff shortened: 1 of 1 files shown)


braver:
not really sure here either, but let's try?

deathaxe:
I guess you're right.

https://github.com/wbond/package_control/issues/1578

braver:
Huh, that’s interesting 🤔

PR #8330, opened 2021-07-28
Add Avrae Utilities package

Croebh:
This adds some utilities for writing code for the Avrae discord bot.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -2474,6 +2474,17 @@
 			]
 		},
 		{
+			"name": "Avrae Utilities",
+			"details": "https://github.com/Croebh/avrae-sublime",
+			"labels": ["avrae", "language syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "AvrAssembler",
 			"details": "https://github.com/vyivanov/AVR-ASM-Sublime",
 			"labels": ["language syntax"],
--- a/repository/a.json
+++ b/repository/a.json
@@ -2481,7 +2481,7 @@
 			"donate": "https://ko-fi.com/croebh",
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": ">=3093",
 					"tags": true
 				}
 			]
--- a/repository/a.json
+++ b/repository/a.json
@@ -2476,7 +2476,9 @@
 		{
 			"name": "Avrae Utilities",
 			"details": "https://github.com/Croebh/avrae-sublime",
-			"labels": ["avrae", "language syntax"],
+			"labels": ["avrae", "language syntax", "utilities"],
+			"author": "Joseph Keen (Croebh)",
+			"donate": "https://ko-fi.com/croebh",
 			"releases": [
 				{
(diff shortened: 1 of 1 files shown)


braver:
Usually we add a language-specific suffix to scopes so that they can be targeted specifically.

Croebh:
I'm not entirely sure what that means, for the language specific thing. Is there a page where I can read up on it?

braver:
For instance, all scopes here end with .css, so that color schemes can make changes specifically for your syntax:

https://github.com/sublimehq/Packages/blob/master/CSS/CSS.sublime-syntax

Croebh:
Beautiful, thank you Braver, I'll set that up tonight

Croebh:
Think I have that sorted now, not entirely sure because I use embeds to use json/python code. When I tried adding it to the embeds I lost their individual schemes so.

braver:
Ah, no you don't want them there. The embeds keep their own scope. Looks good as it is 👍🏻 

Python files in subdirectories are not hot-reloaded, making it a bit harder during development.
I'm also seeing some curious commands to open settings files. That's not really how you interact with settings (and keybindings) in Sublime Text, you may need to check up on [some documentation](https://www.sublimetext.com/docs/api_reference.html#sublime.Settings), or look at [some examples](https://github.com/braver/SideBarTools/blob/master/SideBar.py#L10). Packages contents cannot be read via file paths because they're shipped in an archived format.

However, those commands don't seem to be used. In fact, you're not exposing any way to change settings. Usually you would set up a main menu entry, (something like [this](https://github.com/braver/SideBarTools/blob/master/Main.sublime-menu)), and preferably also a command palette entry (like [this](https://github.com/braver/SideBarTools/blob/master/Commands.sublime-commands#L33)).

Croebh:
Ah, I have it in the command pallete in ``avrae.sublime-commands`` to set the token directly, but using the normal edit_settings option is almost definitely the way to go.

I based my plugin on another ones (not a clue which one now) to learn how to do the settings. I'll play around with this. Thanks again Braver

Croebh:
There! That was really easy, and cleaned up my repo a lot haha.

braver:
Good job, that cleans up a lot 👍🏻 

Some last things (I think 🤞🏻):
- You define a top level menu item for your utils. Please put them in Tools.
- I'm wondering if you maybe should announce in the README that this package makes a remote connection? Maybe it's to be expected in this context, but it's unusual for an editor and people like to know exactly what leaves and enters their system.

Croebh:
@braver Thats very valid! I moved the menu and added a notice to the readme. So used to developing these things for myself that I never even considered it.

braver:
> So used to developing these things for myself 

Yeah, that tends to be the core of the upgrade we do in the review process here.

I think you should remove the `"caption": "-",` separator in the tools menu. Entries added by non-default packages can usually be grouped together just fine.

Croebh:
> I think you should remove the `"caption": "-",` separator in the tools menu. Entries added by non-default packages can usually be grouped together just fine.

Ah gotcha, I just checked another package I had installed that added to the menu and it had it haha.  Its now removed

braver:
🍽

PR #8331, opened 2021-07-29
Add "Toggle Dark Mode"

ElMassimo:
This pull request adds [_Toggle Dark Mode_](https://github.com/ElMassimo/sublime-toggle-dark-mode), a plugin that provides a command to toggle dark mode within Sublime.

It leverages Sublime Text 4's `auto` setting for `theme` and `color_scheme`, as well as the new `ui_info` API to detect OS dark mode preference, making it compatible with all platforms.

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -2624,6 +2624,16 @@
 			]
 		},
 		{
+			"name": "Toggle Dark Mode",
+			"details": "https://github.com/ElMassimo/sublime-toggle-dark-mode",
+			"releases": [
+				{
+					"sublime_text": ">=4096",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Toggle Delphi File",
 			"details": "https://github.com/alefragnani/toggle-delphi-file",
 			"releases": [
--- a/repository/t.json
+++ b/repository/t.json
@@ -2626,6 +2626,9 @@
 		{
 			"name": "Toggle Dark Mode",
 			"details": "https://github.com/ElMassimo/sublime-toggle-dark-mode",
+			"author": "Maximo Mussini (elmassimo)",
+			"donate": "https://paypal.me/maximomussini",
+			"labels": ["toggle", "color scheme", "dark mode", "utilities"],
 			"releases": [
 				{
 					"sublime_text": ">=4096",


braver:
So how does this work (I don't get it from the README)? I set up an auto light/dark theme and color scheme, and this adds a command to switch between those three modes?

Your python will always run in python 3.x (and inside Sublime Text), so this is completely superfluous:

```py
#!/usr/bin/env python
# coding: utf-8
```

I also wonder, especially for such a small feature, why you use a main.py pointing to the actual code in `src`? That breaks the auto-reloading you can get for python in the root directory for ST packages, making it thousand times harder to develop the package. It looks like you inherited the boilerplate from a much larger project. All of that baggage also makes it harder for anyone to contribute though.

We don't recommend shipping with keybindings, especially for packages that don't completely need them. They tend to conflict with those of other packages, or even existing defaults. Just make sure your commands are in the command palette, that's enough for most users. Then maybe document how to make the bindings in the README, so users can choose their own bindings.

ElMassimo:
Thanks for the feedback @braver, I've updated the readme to clarify that what it does is allow you to manually toggle dark mode independently from the current OS preference.

I've also updated the usage instructions, and removed the default key bindings.

As you mentioned, I created the plugin using the structure of an existing one, it's the first time I attempt to publish a Sublime plugin and I wasn't sure how files are loaded.

braver:
Oh, right, that sounds pretty useful. 

Essentially, for Sublime Text a "package" is a directory in the Packages directory. It will load all python files it sees in that directory, and set them up to reload every time they change. You can use python files from other subdirectories, but those don't get hot-reloaded like that and it's generally something to avoid unless your package is so complex it becomes a good trade-off. 

Things like menus and commands generally go into a subdirectory to keep things clean, but a hidden directory (`.sublime` here) doesn't make that much sense. That said, for a simple project you usually just put things in the root ([example](https://github.com/braver/SideBarTools)).

A package_control_channel.json file is also not needed. You can very easily work on packages locally by putting them in the Packages directory, and for distribution outside package control you can just point people to your repo. So there usually is nothing that will actually read that file.

ElMassimo:
Thanks for the explanation! So for a package this simple it would suffice with having everything in the root and sublime will automatically detect the menus and commands, that makes sense!

So the labels for the package and the author should be added here instead?

braver:
> So the labels for the package and the author should be added here instead?

Most stuff will be simply lifted from the GitHub repo (via details), but labels, releases and donate you really want to add in this PR here.

braver:
🎉

ElMassimo:
@braver Thanks for the guidance and the patience!

PR #8332, opened 2021-08-01
Expand Selection to Line Upwards plugin supports all ST versions, including latest ST4

aafulei:
May I request to change the `"sublime_text"` field from `>=3000` to `*` as it is more general, and, as I just tested it, it works for ST4 as well. No change in the code.

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -1574,7 +1574,7 @@
 			"releases": [
 				{
 					"platforms": "*",
-					"sublime_text": ">=3000",
+					"sublime_text": "*",
 					"tags": true
 				}
 			]


rchl:
It's fine as it is. `>=3000` includes ST4 already.

aafulei:
> It's fine as it is. `>=3000` includes ST4 already.

I don't disagree but there is a `ST3` tag to the right of my plugin on the Package Control website, which is a little ambiguous. Therefore, if it is possible I still would like to request the change 🙏, especially considering I have just tested it and confirmed it works on ST4.

rchl:
But using `*` would also make it available to ST2 which is probably not correct. Although at this point probably nobody cares about ST2.

rwols:
I would say this is a UI issue for the packagecontrol.io website then? https://github.com/wbond/packagecontrol.io/issues

aafulei:
I figured out that these are trully misleading labels on Package Control website.

When you hover the mouse over the labels --
`2` means only works for ST2.
`4` means only works for ST4.
However, `3` means works for ST3+.

aafulei:
Anyway, as long as the plugin shows up on `Package Control: Install Package` on ST4, I don't mind leaving it as it is. I am closing the PR. Thank you for the discussion.

PR #8335, opened 2021-08-03
Added C & C++ Advanced package

dannftk:
The package provides syntax highlighting and language hints for C++17 and C11 at now. The rest C/C++ related packages in the Package Control seem to have been abandoned by their maintainers. I am ready to support this package and consider requests from users

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2,6 +2,16 @@
 	"schema_version": "3.0.0",
 	"packages": [
 		{
+			"name": "C & C++ Advanced",
+			"details": "https://github.com/dannftk/cppAdvanced",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "C Base Converter",
 			"details": "https://github.com/flau/c-base-converter",
 			"author": ["Daniel Connolly"],


braver:
You can ignore the version warnings here, but the others are more important.

Any particular existing package you’re looking at? [C Improved](https://packagecontrol.io/packages/C%20Improved)? If you’re a drop in replacement for any dead packages I’d love to replace them instead of adding something new.

rwols:
I'm also wondering what the problem is with the prepackaged syntaxes: https://github.com/sublimehq/Packages/blob/master/C%2B%2B/C%2B%2B.sublime-syntax

dannftk:
Many people might be using old packages including C Improved. Actually there are C++-related packages those are also stale. I would leave them as they are. My package is something that is prone to replace these abondoned packages in the future, but we need to start with something not touching old packages I think

dannftk:
@rwols, I actually missed that C++ prepackaged is being developed. I am closing the PR

rwols:
The prepackaged C++ syntax isn't perfect. Perhaps you can use your talent to improve it: https://github.com/sublimehq/Packages/issues?q=is%3Aissue+is%3Aopen+%5BC%2B%2B%5D

PR #8336, opened 2021-08-05
Add Nelua syntax

edubart:
[Nelua](https://github.com/edubart/nelua-lang) is a minimal, efficient, statically-typed and meta-programmable systems programming language heavily inspired by Lua, which compiles to C and native code.

This package provides syntax highlighting for Nelua source files.

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -215,6 +215,17 @@
 			]
 		},
 		{
+			"name": "Nelua",
+			"details": "https://github.com/edubart/nelua-sublime",
+			"labels": ["nelua", "language syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Nemerle",
 			"details": "https://github.com/xeno-by/sublime-nemerle",
 			"releases": [


braver:
You can ignore the version warning, versions below 3092 aren't really supported in a serious way.
The license file is definitely recommended, but not required.
A question about your syntax: shouldn't the scopes end with `.nelua` instead of `.lua`, so that color schemes etc targeting Lua will not accidentally also target your syntax?

edubart:
> You can ignore the version warning, versions below 3092 aren't really supported in a serious way.

Ok.

> The license file is definitely recommended, but not required.

I've added the LICENSE file to the repository.

> A question about your syntax: shouldn't the scopes end with `.nelua` instead of `.lua`, so that color schemes etc targeting Lua will not accidentally also target your syntax?

Making the scopes end with `.lua` is intentional for two reasons:

1. One reason is exactly that, to reuse color schemes targeting Lua, because the language embeds Lua in its preprocessor while the original Lua syntax will not work because preprocessor can have incomplete statements.
2. The other reason is to make easy for me to maintain the package while tracking updates in the original Sublime's Lua syntax file. As the syntax is a super set of the Lua syntax, the file differences between Lua syntax file are minor if I don't change that, so it's easy to see differences in a patch file, and also to update new changes from Lua syntax. If I kept everything ending `.nelua` would add some extra work to maintain.

braver:
Alright, excellent 👍🏻

rwols:
> the syntax is a super set of the Lua syntax

@edubart, have you thought about using `extends: Packages/Lua/Lua.sublime-syntax`?

edubart:
> @edubart, have you thought about using `extends: Packages/Lua/Lua.sublime-syntax`?

I wasn't aware of that feature, but I think copying `Lua.sublime-syntax` and doing the necessary changes can be more free of errors, like in case a future Lua syntax change breaks my syntax. But now that I know I may revisit this possibility in the future, thanks for letting me know.

PR #8338, opened 2021-08-06
Add ryanolsonx/sublimetext-goodwolf-theme

ryanolsonx:
This is a Sublime Text color scheme that is a port of the dark version of good wolf (https://stevelosh.com/projects/badwolf/). It also includes a light version of good wolf. Good wolf is a minimalist syntax highlighting color scheme that only highlights some things so you can focus on what's important.

I don't think anyone has ported Good Wolf to Sublime Text.

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -1361,6 +1361,17 @@
 			]
 		},
 		{
+			"name": "Good Wolf Color Scheme",
+			"details": "https://github.com/ryanolsonx/sublimetext-goodwolf-theme",
+			"labels": ["color scheme"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Google Apps Scripts",
 			"details": "https://github.com/revolunet/sublimetext-google-apps-scripts",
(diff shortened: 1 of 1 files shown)


braver:
Well, there is [Wolf](https://packagecontrol.io/packages/Wolf%20Theme) (sadly incorrectly labeled as a theme, it's a color scheme). I think it's unrelated though.

It looks like you have some non-functional files in your repo, you can exclude them from the package as it's shipped to users, saving on bandwidth, storage and CO2, using `. gitattributes` and `export-ignore`.

ryanolsonx:
What files should I ignore? I get the obvious ones (like the itermcolors and slack.csv) but would I need to keep README and screenshot in there? Or would you advise that I export-ignore those too?

braver:
Usually the images are a good thing to exclude. You could exclude the readme I guess but I wouldn't go that far. It doesn't matter for the packagecontrol site, that just reads from your repo (not the package). It's not all that important either, but it's good to understand the mechanism especially for packages that tend to contain quite a few more bits in screenshots that in actual functionality.

Anyway, the package looks good so let's 🚢

ryanolsonx:
Sweet. Thanks!

PR #8339, opened 2021-08-07
Add PrintDebugger

futureprogrammer360:
PrintDebugger is a plugin that allows easy insertion of both debugging and empty print statements through keybindings. It is fully customizable, with language syntax able to be modified and new syntax able to be added.

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -2264,10 +2264,10 @@
 		{
 			"name": "PrintDebugger",
 			"details": "https://github.com/futureprogrammer360/PrintDebugger",
-			"labels": ["debug", "print", "output"],
+			"labels": ["debug", "print", "output", "log", "logs", "console"],
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": ">=4107",
 					"tags": true
 				}
 			]
--- a/repository/p.json
+++ b/repository/p.json
@@ -2262,6 +2262,17 @@
 			]
 		},
 		{
+			"name": "PrintDebugger",
+			"details": "https://github.com/futureprogrammer360/PrintDebugger",
+			"labels": ["debug", "print", "output"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Prisma",
 			"details": "https://github.com/prisma/sublimeText3",
 			"releases": [


braver:
Thanks, that looks really useful. Some questions:

- Understanding what your package does would be much easier if you could include a gif in the README to illustrate the interaction.
- You're really focussed on the key bindings. We don't recommend shipping with key bindings because they tend to conflict with other packages. And for most maintainers it's not possible to test and verify correct default bindings on all three platforms (and to decide on good defaults that match with what users expect on each platform because that tends to be quite different). I would suggest removing the key bindings and in stead documenting in the README how users can set up their own.
- Your readme mentions that things can be customised but isn't clear as to _how_ to do so. Your install message actually seems to be clearer in that sense.
- Maybe it's me (probably 😄 ), but I really don't understand this in the settings: "second element is the number of times the left arrow needs to be pressed to move cursor from end of line to where you start typing"?
- You're trying to match either scope or extension. Any specific reason for the extensions? I would expect scopes to pretty much cover all cases.

jrappen:
FYI Build 4103 added the `chain` command (compare: <https://www.sublimetext.com/dev>), which should let you clean up your python file.

repository/p.json
```diff
@@ -2261,6 +2261,17 @@
 				}
 			]
 		},
+		{
+			"name": "PrintDebugger",
+			"details": "https://github.com/futureprogrammer360/PrintDebugger",
+			"labels": ["debug", "print", "output"],
+			"releases": [
+				{
+					"sublime_text": "*",
```
jrappen, line 2270: since you use `py3.8`, this should be at least be `"sublime_text": ">=4050"` (first dev build) or even `"sublime_text": ">=4107"` (first stable build) which also includes the new `chain` command from Build 4103.

futureprogrammer360:
Thanks for the feedback @braver!

> * Understanding what your package does would be much easier if you could include a gif in the README to illustrate the interaction.

I've added 2 GIFs in the README. From other PRs, I see it might be a good idea to add a `.gitattributes` file, is that applicable in this situation? (Sorry if this is a dumb question, I haven't used `.gitattributes`  before 😅)

> * You're really focussed on the key bindings. We don't recommend shipping with key bindings because they tend to conflict with other packages. And for most maintainers it's not possible to test and verify correct default bindings on all three platforms (and to decide on good defaults that match with what users expect on each platform because that tends to be quite different). I would suggest removing the key bindings and in stead documenting in the README how users can set up their own.

I included the default keybindings only because I use them about once a minute 😆. Anyways, I removed them now and documented how to add keybindings in the README.

> * Your readme mentions that things can be customised but isn't clear as to _how_ to do so. Your install message actually seems to be clearer in that sense.

I modified the README's Customization section, please let me know if it's still unclear.

> * Maybe it's me (probably 😄 ), but I really don't understand this in the settings: "second element is the number of times the left arrow needs to be pressed to move cursor from end of line to where you start typing"?

That's my bad. The plugin basically moves the cursor from the end of line back a few times into where the user can start typing after inserting a print statement. I changed the design now so the print statement syntax in settings uses `$0` for where the cursor should end up like in snippets. Maybe it's a bit more intuitive now?

> * You're trying to match either scope or extension. Any specific reason for the extensions? I would expect scopes to pretty much cover all cases.

The extensions are a fallback. Some packages, like [Djaneiro](https://packagecontrol.io/packages/Djaneiro) for instance, change the scope of Python files from `source.python` to `source.python.django`, which would cause the plugin to stop working. The extensions are used in those situations.

Please let me know if there's anything else I should change.

futureprogrammer360:
Thanks for your suggestions @jrappen! I couldn't find any documentation on the `chain` command besides the change log, can you point me in the right direction?

I totally forgot that Python 3.8 support was only added in ST4, thanks for pointing it out! Looking back, the only place where `py3.8` is needed is one f-string which can be easily swapped out for something `py3.3` supports. Would it be a good idea to make that change so all ST versions are supported and more people can use the plugin?

braver:
> I see it might be a good idea to add a .gitattributes file, is that applicable in this situation?

No such thing as a dumb question 😉 It's always a good idea to remove images only used for documentation, or files only used during development and testing, from the final package using gitattributes.

> I removed them now and documented how to add keybindings in the README.

Excellent! It's nothing against the key bindings in principle, just that in practice key bindings bundled in packages are a bit of a hassle once you use more than few packages.

>  modified the README's Customization section, please let me know if it's still unclear.

Works for me like this 👍🏻 

> I changed the design now so the print statement syntax in settings uses $0 for where the cursor should end up like in snippets.

Oh, now I get it! 😄 Yeah, a marker like `$0` makes total sense, much clearer now.

> source.python to source.python.django, which would cause the plugin to stop working

Well actually, it shouldn't. Those scopes are chosen that way to indicate that both are actually python and both will match for `source.python`. So if that breaks your package, maybe your scope matching needs some attention.

> Would it be a good idea to make that change so all ST versions are supported and more people can use the plugin?

I'm always in favour of pushing the envelope. We can't still be using 3.3 ten years from now. If you support ST3 you'll have a bigger audience for now, but you also don't get to use 3.8 (or ST4 specific API's) until you drop ST3 support.

futureprogrammer360:
Thanks for the reply @braver.

> It's always a good idea to remove images only used for documentation

I've added a `.gitattributes` file. Let me know if it needs any change.

> So if that breaks your package, maybe your scope matching needs some attention.

I intentionally made it so they're recognized as different scopes. It's an easy fix to change that, but it'll cause some problems like `source.c` and `source.c++` or `source.js` and `source.json` potentially pointing to the same syntax.

> I'm always in favour of pushing the envelope. We can't still be using 3.3 ten years from now. If you support ST3 you'll have a bigger audience for now, but you also don't get to use 3.8 (or ST4 specific API's) until you drop ST3 support.

Cool, I've corrected the ST version to `">=4107"` as @jrappen suggested.

futureprogrammer360:
I just realized that I forgot to push the `.gitattributes` file to GitHub 😂. Everything should be up to date now, sorry about that!

braver:
> but it'll cause some problems like source.c and source.c++ or source.js and source.json potentially pointing to the same syntax

Ha, if that were true the entire concept of scopes wouldn’t work 😉. If you go through the API (ie. you’re not doing substring matching yourself like you do now), that will never happen. If it helps, think of it as CSS classes and selectors. Have a look at match_selector and score_selector in the API reference.

futureprogrammer360:
> Ha, if that were true the entire concept of scopes wouldn’t work 😉. If you go through the API (ie. you’re not doing substring matching yourself like you do now), that will never happen. If it helps, think of it as CSS classes and selectors. Have a look at match_selector and score_selector in the API reference.

That makes sense, I should spend more time on the API reference. Please give me some time to fix the scope matching. I'll let you know when the issue is fixed. Thanks!

futureprogrammer360:
@braver, I've changed the scope matching to use `match_selector` and it should now work properly. Please let me know if there's anything else that isn't working.

braver:
Excellent, looks good to me. Don't forget to tag a release with all your latest changes.

futureprogrammer360:
Got it, thank you!

PR #8340, opened 2021-08-08
Add Exact Quick Find

aafulei:
I just wrote a Sublime Text plugin `Exact Quick Find` which I hope will benefit more Sublime Text users.

This plugin improves the built-in `Quick Find` command by allowing users to specify `Case Sensitive` and `Whole Word` flags.

These flags are always displayed on the status bar and can be easily toggled with keyboard shortcuts.

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -1438,7 +1438,6 @@
 			"labels": ["search", "text navigation", "text selection"],
 			"releases": [
 				{
-					"platforms": "*",
 					"sublime_text": "*",
 					"tags": true
 				}
--- a/repository/e.json
+++ b/repository/e.json
@@ -1432,6 +1432,19 @@
 			]
 		},
 		{
+			"name": "Exact Quick Find",
+			"details": "https://github.com/aafulei/sublime-exact-quick-find",
+			"author": "Aaron Fu Lei",
+			"labels": ["search", "text navigation", "text selection"],
+			"releases": [
+				{
+					"platforms": "*",
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Exalt",
 			"details": "https://github.com/eerohele/exalt",
 			"labels": ["xml", "html", "xslt", "validation", "formatting"],


repository/e.json
```diff
@@ -1431,6 +1431,19 @@
...
 			]
 		},
+		{
+			"name": "Exact Quick Find",
+			"details": "https://github.com/aafulei/sublime-exact-quick-find",
+			"author": "Aaron Fu Lei",
+			"labels": ["search", "text navigation", "text selection"],
+			"releases": [
+				{
+					"platforms": "*",
```
braver, line 1441: If you support all platforms you can simply not specify it.

braver:
It looks like you have some non-functional files (e.g. images) in your repo, you can exclude them from the package as it's shipped to users, saving on bandwidth, storage and CO2, using `.gitattributes` and `export-ignore`.

Maybe you should consider starting your Main menu entries with a separator (`{ "caption": "-" },`), or maybe even grouping them together into a sub menu. I would expect most users will either interact with this via the palette, and actually I expect most would want to set up keybindings for this. So maybe your main menu entries could be there for discovery, but take up a little less space. Your call though!

Pretty cool and useful package, thanks!

aafulei:
@braver Thanks for the like and the feedback! The comments are very helpful and to the point. Allow me some time to consider how to improve it. Should be within this week. Will let you know once it's ready again.

aafulei:
Hi @braver I spent 3/4 of the past weekend improving this. Have added quite a few more useful features. With the grow of the code size, more tests are needed. The doc needs to be reworked too. Thus I need more time. Possibly another week as I am usually busy over the weekdays. Thank you.

braver:
Cool, no hurry! Just let me know when it’s shippable 👌🏻

jrappen:
Also, do not override existing captions as in:

https://github.com/aafulei/sublime-exact-quick-find/blob/a1806d681ff32f9483284c36b4591d9bf6f9f610/Main.sublime-menu#L2-L4

but simply reference the `"id"`.

jrappen:
Since you explicitly use `py3.8` you can make use of f-strings and other newer features.

aafulei:
@jrappen Thank you for the helpful remarks! Point taken. I am working on the project. For Python versions, indeed I prefer compatibility (especially for ST3). F-string is a great feature. But to stay compatible with Python 3.3 I have to refrain from using it. Maybe there is no need to explicitly have that `.python-version`.

aafulei:
I would like to do a status update here: I think I have done the code development. It's 10 times the features, the code, and the power than the original submission. Allow me some time to prepare a good doc that matches the efforts in the code. Thank you very much!

aafulei:
When everything is ready, I will leave a note here and will re-request reviews.

repository/e.json
aafulei: The line `"platforms": "*",` has been removed. Thanks.

aafulei:
> It looks like you have some non-functional files (e.g. images) in your repo, you can exclude them from the package as it's shipped to users, saving on bandwidth, storage and CO2, using `.gitattributes` and `export-ignore`.

Have tagged all the images with `export-ignore` in `.gitattributes`. Thanks!

> 
> Maybe you should consider starting your Main menu entries with a separator (`{ "caption": "-" },`), or maybe even grouping them together into a sub menu. I would expect most users will either interact with this via the palette, and actually I expect most would want to set up keybindings for this. So maybe your main menu entries could be there for discovery, but take up a little less space. Your call though!
> 
> Pretty cool and useful package, thanks!

Have grouped all the main menu entries together into a sub menu, and have added multiple separators (`{ "caption": "-" },`). Thanks!

aafulei:
Hi @braver the plugin is ready for shipping. Could you please take a look again?

jrappen:
Since you use py3.8 this should be ST4 only.

aafulei:
> Since you use py3.8 this should be ST4 only.

@jrappen I have a different understanding. My understanding is that setting`3.8` in file `.python-version` is only a **suggestion**, which means, if both Python 3.3 and 3.8 are supported, Python 3.8 will be used. So if ST 4050 sees this `.python-version`, it will use Python 3.8. If ST3 sees this `.python-version`, it will continue to use Python 3.3. Won't it?

Source: https://www.sublimetext.com/docs/api_environments.html

The code has been run and tested on both ST3 and ST4.

jrappen:
Yes, then it's fine. I'd assume most users will have migrated to ST4 soon anyway. So you might as well save yourself the hassle of supporting `py3.3` going forward.

Whatever works for you, though.

aafulei:
> Yes, then it's fine.

Thanks for the confirmation. 

Some explanation - Compatibility is a priority for me. So all the syntax and standard library calls are carefully taken to make sure they work with Python 3.3.

Beyond that, I do assume Python 3.8 might bring some performance boost, maybe even for some same library calls. So if there is a choice, I would like to run the code in Python 3.8 environment.

braver:
What you could do with the key bindings file is keep the normal file and comment everything out. And then have a menu and command palette entry for the bindings as if you were shopping them. Users could then easily find your example bindings and copy them over to their personal bindings file to enable them.

aafulei:
Hi @braver thank you for the feedback. I have made modifications according to your suggestions.

In fact previously I did have a `Recommended-sublime-keymap.json` file that will open as the base file if the user clicks `Preferences > Package Settings > Exact Quick Find > Key Bindings`.

Since you suggested, I explicitly changed that `.json` file to 3 identical `.sublime-keymap` files.

Besides, the command names for key bindings and the menu / command palette entry names have been carefully designed that they are strictly identical except for letter case, and the spaces replaced with `_`. Thus, I feel there is no need to list the entry names again alongside the commands, which would be cumbersome. Though, I have added several comments as headers for different sections.

braver:
Excellent 👍🏻

PR #8342, opened 2021-08-10
Add ryanolsonx/sublimetext-syntax-off-theme

ryanolsonx:
I've created a color scheme that turns off syntax highlighting. This is something that I wasn't able to find in packagecontrol already. So I created my own.

I didn't add .gitattributes because I've kept this repository small and only have two small screenshots.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -5773,6 +5773,17 @@
 			]
 		},
 		{
+			"name": "Syntax Off Color Scheme",
+			"details": "https://github.com/ryanolsonx/sublimetext-syntax-off-theme",
+			"labels": ["color scheme"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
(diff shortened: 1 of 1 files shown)


ryanolsonx:
I just added a semver tags. Woops! I forgot to do that.

braver:
We may already have something like this, but I can't find it. It's a nice touch you kept the diff colours in there. Nice!

rwols:
What about block comments though?

braver:
> What about block comments though?

That's a slippery slope man! 😄

ryanolsonx:
> What about blocks comments though?
😱

ryanolsonx:
In all seriousness, no highlighting for block comments. 😉 

I want this to be like running `:syntax off` in vim. I'm planning to make another theme called "Syntax Less" that will feature highlighting for comments and _maybe_ something for strings.

PR #8343, opened 2021-08-10
Add BinViewer package

AetelFinch:
This plugin allows the user to see the binary representation of decimal and hexadecimal numbers

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -813,6 +813,7 @@
 		{
 			"name": "BinViewer",
 			"details": "https://github.com/AetelFinch/BinViewer",
+			"labels": ["binary notation", "bin"],
 			"releases": [
 				{
 					"sublime_text": ">=3000",
--- a/repository/b.json
+++ b/repository/b.json
@@ -811,6 +811,16 @@
 			]
 		},
 		{
+			"name": "BinViewer",
+			"details": "https://github.com/AetelFinch/BinViewer",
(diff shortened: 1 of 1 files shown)


braver:
Neat idea!

I don't think you need to but a `deadbeef` in there.
Your command is only accessible via keybindings. We recommend against keybindings. They tend to conflict with other packages, and for things you don't use every day they're usually impossible to remember. Instead, make sure your command is accessible via the command palette, and maybe also via the main menu. In your README you might the explain how users can set up a keybinding if they want to.
You might also want to implement `is_applicable` for your command, to detect if it would work at all for the current selection. 
Also remember that in ST you can have multiple cursors. You get bonus points if you can handle that elegantly without simply resorting to only reporting on the first selection. 
It would seem useful to be able to copy the binary value to a clipboard. You can find an example of how to do that in the `show_scope_name.py` plugin that's in the default package.

AetelFinch:
Okay, thanks for the review. I'll try to fix it

braver:
Good job on the multiple cursors and the copy button!

Tip: you don't need multiple menu entries for each platform. [example](https://github.com/SublimeLinter/SublimeLinter/blob/master/Main.sublime-menu#L29)

You can fix that in a future version, let's start shipping this thing 👍🏻

AetelFinch:
Thanks for your feedback 😊

PR #8344, opened 2021-08-11
Add "JamfLog" syntax

jorks:
Hello 👋 
JamfLog is a Sublime Text syntax highlighter for Jamf Pro server logs. Simply put, it makes logs readable by humans 🤗.
There were no similar plugins and I have successfully run all the pre-tests. Tested on SB3 & SB4 for Mac/Win/Lin. Feedback welcomed.
Thank you for your time 👍

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -79,6 +79,17 @@
 			]
 		},
 		{
+			"name": "JamfLog",
+			"details": "https://github.com/jorks/sublime-jamflog",
+			"labels": ["jamf","syntax","highlighting"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Jamon",
 			"details": "https://github.com/drhayes/SublimeJamon",
 			"releases": [
--- a/repository/j.json
+++ b/repository/j.json
@@ -81,10 +81,10 @@
 		{
 			"name": "JamfLog",
 			"details": "https://github.com/jorks/sublime-jamflog",
-			"labels": ["jamf","syntax","highlighting"],
+			"labels": ["jamf","log","highlighting"],
 			"releases": [
 				{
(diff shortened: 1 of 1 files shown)


repository/j.json
```diff
@@ -78,6 +78,17 @@
 				}
 			]
 		},
+		{
+			"name": "JamfLog",
+			"details": "https://github.com/jorks/sublime-jamflog",
+			"labels": ["jamf","syntax","highlighting"],
```
braver, line 84: "language syntax" is the canonical label for such a package.

braver:
- Please don't override a user's color scheme for your syntax, or ship any other settings that are not required to work with your syntax. Users can set their own syntax specific settings if they want, but language package shouldn't change things by themselves.
- The fact that you're shipping a color scheme at all is a red flag. Did you follow the scope names guidelines? You should make sure your syntax works with any color scheme. In fact I would very much advise against shipping a color scheme here.
- If your color scheme is any good for other syntaxes, you might want to consider shipping that in its own package.
- Your ascii art is really cool, but does it need to preface all your files?
- The copy right stuff has no place in the syntax file. You already have a license file that should cover all what's needed here.

jorks:
Thank you for the ideas to improve compatibility with a wider range of colour schemes, I am in the process of updating my scopes and will report back shortly.

jorks:
Sorry, it took me a minute to get back to this project. Thanks for keeping the PR open.

The advice has paid off and the syntax file now implements scopes in a much more compatible and standard way. Stacking scopes worked really well to support a wider set of Colour Schemes and Themes. See GIF below.

I have pushed a new tag (v1.0.4) for this update. Let me know if you need anything further. 

Thank you, James

braver:
Looks good, let's ship it. It does look like you need to figure out what to do with the color scheme though. Maybe you could recommend using it in the README?

PR #8345, opened 2021-08-11
Update Maude syntax

ngraymon:
Small updates to the previous Maude syntax package
the primary one being that comments can now be toggled
using the default shortcut

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -1062,7 +1062,7 @@
 		},
 		{
 			"name": "Maude Syntax Highlighting",
-			"details": "https://github.com/jpsikorra/maude_syntax_highlight",
+			"details": "https://github.com/ngraymon/maude_syntax_highlight",
 			"labels": ["maude", "language syntax"],
 			"releases": [
 				{


ngraymon:
My aim is to slowly flesh out the repository, borrowing from this [repo](https://github.com/adamlazz/language-maude) for atom; while following the documentation for [scope naming](https://www.sublimetext.com/docs/scope_naming.html) and [syntax](https://www.sublimetext.com/docs/syntax.html). Let me know if I need to provide any additional information, thanks!

braver:
You seem to be taking over this package. Does the original author agree?

ngraymon:
> 
> 
> You seem to be taking over this package. Does the original author agree?

I do not know. I was not able to find contact information otherwise I would have reached out. I've added an issue to the original repository as an attempt at contacting them.

braver:
Yes, an issue or Pull Request on the original package’s repo would be the first thing to attempt. If that doesn’t work out, and/or @jpsikorra doesn’t want to maintain his package anymore, we can always revisit this PR.

ngraymon:
> 
> 
> Yes, an issue or Pull Request on the original package’s repo would be the first thing to attempt. If that doesn’t work out, and/or @jpsikorra doesn’t want to maintain his package anymore, we can always revisit this PR.

Okay, thanks for the information!

PR #8346, opened 2021-08-12
Add support for CookLang syntax

dubadub:
Syntax highlighting for the [CookLang markup language](https://cooklang.org)

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3402,6 +3402,16 @@
 			]
 		},
 		{
+			"name": "CookLang",
+			"details": "https://github.com/cooklang/CookSublime",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Cool",
 			"details": "https://github.com/princjef/sublime-cool-highlighter",
 			"releases": [


dubadub:
@packagecontrol-bot please, review

Ultra-Instinct-05:
A few points about your syntax (Just from skimming it)

1. https://github.com/cooklang/CookSublime/blob/main/Cook/Cook.sublime-syntax#L5 - It looks like this language is some kind of a markup language & not a programming language. `source.` is usually given for programming languages and `text.` for markup languages e.g. HTML & Markdown. (But I guess it's fine)

2. https://github.com/cooklang/CookSublime/blob/main/Cook/Cook.sublime-syntax#L6 - You don't need to specify a `version` key if you know your syntax is using `version: 1`, it's already understood you are in version 1. Only if you want to opt into `version: 2`, do you specify that key. https://www.sublimetext.com/docs/syntax.html#compatibility. This will also probably not work in ST3 since that key was introduced in ST4 (Not sure though)

3. It looks like there are several places where you pop more than 1 context off the stack. Note that popping arbitrary # of contexts off the stack is an ST4 specific feature. In ST3, one is only allowed `pop: true`. So, you can't use `sublime_text: "*"` in your package settings. Either you need to restrict it to > ST4 or rework your syntax to use only `pop: true` to make it both 3 & 4 compatible.

dubadub:
> A few points about your syntax (Just from skimming it)
> 
> 1. https://github.com/cooklang/CookSublime/blob/main/Cook/Cook.sublime-syntax#L5 - It looks like this language is some kind of a markup language & not a programming language. `source.` is usually given for programming languages and `text.` for markup languages e.g. HTML & Markdown. (But I guess it's fine)
> 2. https://github.com/cooklang/CookSublime/blob/main/Cook/Cook.sublime-syntax#L6 - You don't need to specify a `version` key if you know your syntax is using `version: 1`, it's already understood you are in version 1. Only if you want to opt into `version: 2`, do you specify that key. https://www.sublimetext.com/docs/syntax.html#compatibility. This will also probably not work in ST3 since that key was introduced in ST4 (Not sure though)
> 3. It looks like there are several places where you pop more than 1 context off the stack. Note that popping arbitrary # of contexts off the stack is an ST4 specific feature. In ST3, one is only allowed `pop: true`. So, you can't use `sublime_text: "*"` in your package settings. Either you need to restrict it to > ST4 or rework your syntax to use only `pop: true` to make it both 3 & 4 compatible.

Hi @Ultra-Instinct-05. Thank you for your review, it's really useful. I [made](https://github.com/cooklang/CookSublime/compare/v0.0.1...v0.0.2) all the changes as you suggested.

braver:
Looks good, thanks!

PR #8347, opened 2021-08-12
Add HDL_Automation

ToTamire:
# HDL_Automation
Verilog and SystemVerilog automation with Sublime Text 4

* Auto format file
auto_format_file
* Auto format selections
auto_format_selections

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -335,6 +335,18 @@
 			]
 		},
 		{
+			"name": "HDL_Automation",
+			"details": "https://github.com/ToTamire/HDL_Automation",
+			"labels": ["text manipulation", "formatting"],
+			"releases": [
+				{
+					"platforms": ["windows", "linux"],
+					"sublime_text": ">=4000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "HDLProject",
 			"details": "https://github.com/bootsiaz/sublime-hdlproject",
 			"author": "Andrew Carter",


braver:
Is your package also about other kinds of automation that just formatting? 
Have you looked at [HDLProject](https://packagecontrol.io/packages/HDLProject), and might it be a good idea to maybe integrate with [Formatter](https://packagecontrol.io/packages/Formatter)?

ToTamire:
Code formatting is a first step. This package will do much, much more. 
I check all packages for sublime and I am not satisfied. They are good for hobbyists. I will add three or four packages which together will make full verilog environment.
After few commits my package will do more than code format so I don't think that integration with formatter is good idea.

braver:
Alright 👍

- For context menu items use is_visible to remove unavailable commands instead of is_enabled.
- Why is there a dependencies.json with your own package?
- For your settings menu entry you can have a single entry that will open the settings in a split view (there are plenty examples out there).
- It’s common practice to also expose the settings in the command palette.
- Your settings are all prefixed with the package name. That’s not necessary, they’re already scoped to your package.
- I think you load settings only on initialization and you might miss on changes a user made. In any case you don’t need to worry about performance here, settings are always served from memory. So maybe you can optimize a bit less here and reduce the complexity a bit. In fact you have so much duplication here, almost halve your code is about reading settings. If you document your settings well you can simply let your package error out if someone puts in an incorrect value instead of spending so much effort dealing with it. If your dead set on validating settings, I would suggest printing out an error somewhere so the user can fix their error (instead of simply falling back to a default value).
- A detail, but I find the underscore in the package name a bit inelegant, you can use a space there.
- You seem to match files by their extension. Wouldn’t it be easier to match them by syntax (ie. using scope matching)?

ToTamire:
> * For context menu items use is_visible to remove unavailable commands instead of is_enabled.

Done.

> * Why is there a dependencies.json with your own package?

Idk. Removed.

> * For your settings menu entry you can have a single entry that will open the settings in a split view (there are plenty examples out there).

Indeed. Done.

> * It’s common practice to also expose the settings in the command palette.

Done.

> * Your settings are all prefixed with the package name. That’s not necessary, they’re already scoped to your package.

You are right. Prefixes removed.

> * I think you load settings only on initialization and you might miss on changes a user made. In any case you don’t need to worry about performance here, settings are always served from memory. So maybe you can optimize a bit less here and reduce the complexity a bit. In fact you have so much duplication here, almost halve your code is about reading settings. If you document your settings well you can simply let your package error out if someone puts in an incorrect value instead of spending so much effort dealing with it. If your dead set on validating settings, I would suggest printing out an error somewhere so the user can fix their error (instead of simply falling back to a default value).

Load settings fixed, but I will not touch settings validation. Hardware description language is different from programming language. If you make a mistake in programming language, you can program device again. If you make a mistake in hardware description language, you release from the factory bilion devices with bug on gate level and there is not way to make changes in silicon. I have to validate settings as accurately as possible. 

> * A detail, but I find the underscore in the package name a bit inelegant, you can use a space there.

Unnoticeable in snake case / screaming snake case language.

> * You seem to match files by their extension. Wouldn’t it be easier to match them by syntax (ie. using scope matching)?

Sublime text don't have build in Verilog / SystemVerilog syntax. Users have to download it from web. There are few syntaxes with different scope names. At the moment it is easier to match them by extension.

braver:
> There are few syntaxes with different scope names.

Which are you seeing then? We really try to prevent duplicates in package control and all I can find are [source.verilog](https://packagecontrol.io/packages/Verilog) and [source.systemverilog](https://packagecontrol.io/packages/SystemVerilog), which seems correct. 

> I have to validate settings as accurately as possible.

Ok, I can see where you're coming from with that. It also sounds like it might be kind of important to let the user know that an error was made and that the code will decide to throw away the setting and use the default value. A simple print to the console can save a user a lot of headaches.

Also, this screams for a little helper function to dedupe a bit:

```py
# all this 16x
setting = self.settings.get("case_items_alignment")  # get setting
if type(setting) == str:  # check if setting is a string
    setting = setting.strip()  # remove leading and trailing whitespaces
    setting = setting.lower()  # convert to lowercase
    if setting in ['align', 'flush-left', 'preserve', 'infer']:  # check if setting is correct
        return setting  # return correct setting
return 'infer'  # otherwise return default setting

# could be
def check_setting(key, allowed_values, default):
  setting = self.settings.get(key) 
  if type(setting) == str:
    setting = setting.strip()
    setting = setting.lower()
    if setting in allowed_values:
        return setting
  print('HDL Automation: Incorrect setting for {}, reverting to the default value!'.format(key))
  return default

# and this 16x
check_setting('case_items_alignment', ['align', 'flush-left', 'preserve', 'infer'], 'infer')  
```

braver:

ToTamire:
Reverting to the default value now prints to the console.

braver:
Excellent 👍🏻

PR #8352, opened 2021-09-02
Add Firebase Rules Syntax

jingyuexing:

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -593,6 +593,17 @@
 			]
 		},
 		{
+			"name": "ElixirDocs",
+			"details": "https://github.com/reisub/sublime-elixir-docs",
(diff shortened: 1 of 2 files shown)


jingyuexing:
image

braver:
Looks good in general but you'll want to look at the report from the bot. And maybe consider using `.gitattributes` to "export-ignore" the png from the final package to save on some storage and bandwidth.

jingyuexing:
> Looks good in general but you'll want to look at the report from the bot. And maybe consider using `.gitattributes` to "export-ignore" the png from the final package to save on some storage and bandwidth.

already added `.gitattributes` file

braver:
Having a LICENSE file is a recommendation, not a requirement. You can look into that in your own time. For now, let's 🚢

PR #8355, opened 2021-09-03
Add FullwidthTabSpaces

woodmix:
I would like to add a plugin to solve a minor inconvenience related to full-width characters and the "translate_tabs_to_spaces":true setting.

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -1823,6 +1823,16 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


braver:
Sorry for letting this sit here for so long, we've been struggling with capacity and also with the question of how to approach a package like this. Seems to this is fine to ship, so let's 🏎️

PR #8358, opened 2021-09-05
added openssl configuration file syntax

finitelife:
This package simply highlights the openssl configuration file syntax as it is slightly different from the INI syntax.
This includes default words, proper comment escaping, quoting, variable highlighting, section/name/value pair coloring, multiline values, directives, etc

### DIFF
--- a/repository/o.json
+++ b/repository/o.json
@@ -859,6 +859,23 @@
 			]
 		},
 		{
+			"name": "OpenSSL",
+			"details": "https://github.com/finitelife/openssl-sublimetext",
+			"readme": "https://github.com/finitelife/openssl-sublimetext/blob/master/README.md",
+			"author": "finitelife",
+			"labels": [
+				"language syntax",
+				"color scheme",
+				"formatting"
+			],
+			"releases": [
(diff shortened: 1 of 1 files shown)


finitelife:
"CI / trigger-review (pull_request) Failing after 3s — trigger-review"
Why is this happening? I'm not seeing any specific reason here vs any other accepted/closed pull requests.

repository/o.json
```diff
@@ -858,6 +858,23 @@
 				}
 			]
 		},
+		{
+			"name": "OpenSSL",
+			"details": "https://github.com/finitelife/openssl-sublimetext",
```
braver, line 863: The author and readme are just the defaults so we get those from the repo URL. So you can remove them here.

braver (changes requested):

braver:
- author isn't actually an [allowed header key](https://www.sublimetext.com/docs/syntax.html) (put it in the readme or license)
- why does your package contains a sublime-color-scheme, and a settings file?
- a package-metadata.json has no place in a package

PR #8360, opened 2021-09-06
Added plugin "Inline Images"

haferburg:

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -715,6 +715,16 @@
 			]
 		},
 		{
+			"name": "Inline Images",
+			"details": "https://github.com/haferburg/Sublime-Inline-Images",
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Inline Python",
 			"details": "https://github.com/apiad/Sublime-InlinePython",
 			"releases": [


haferburg:
It's been 10 weeks, and the PR hasn't been merged. Is there a problem with the PR or the package? Did I do something wrong?

Ultra-Instinct-05:
Here are a couple of areas where you can improve.

1. The `test` directory mainly seems to have files & images that are used for testing purposes and not needed by the package in anyway. You can use `.gitattributes`'s `export-ignore` flag to prevent shipping such files to the end users.
2. https://github.com/haferburg/Sublime-Inline-Images/blob/main/Main.sublime-menu#L18-L27 Please don't use the `open_file` command to add 2 different menu items for settings. When it comes to settings, the best experience comes when they are viewed side by side. Consider using the `edit_settings` command instead.

```json
"children": [
    {
        "caption": "Settings",
        "command": "edit_settings",
        "args": { "base_file": "${packages}/Inline Images/Inline Images.sublime-settings" },
    },
]
``` 
3. https://github.com/haferburg/Sublime-Inline-Images/blob/main/messages.json Please don't use the README as an install message. Markdown is not a suitable way of showing package install/update messages. Consider using a simple plain text file instead. References: https://packagecontrol.io/docs/messaging

Haven't gone into great detail about the plugin itself, but otherwise looks okay.

braver:
Your README itself links to similar packages, but you don't explain why we need to add yours. We try to prevent duplication.

PR #8362, opened 2021-09-09
New ST4 Compatible release of vue-syntax-highlighter

skyronic:
The Vue Syntax Highlight had to be updated for ST4 compatibility. This change creates a separate version installable only for users with Sublime Text Version > 4000

### DIFF
--- a/repository/v.json
+++ b/repository/v.json
@@ -766,8 +766,12 @@
 					"tags": "oldsyntax-"
 				},
 				{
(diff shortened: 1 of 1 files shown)


rchl (approved):

rchl:
I've been using the new branch for the last month so got it tested pretty well.

rchl:
Right. I've had `st4` branch checked out manually so didn't notice.

PR #8366, opened 2021-09-17
Add missing readme and issues link for jq gitlab package

Jiehong:
This package correctly appears on packagecontrol.io, but its readme and issues link is not automatically determined (see https://packagecontrol.io/packages/Jq).

This fixes that.


braver:
I guess it doesn't pick those up from GitLab. 🙆🏻‍♂️

PR #8370, opened 2021-09-21
Add GraphQL Playground

kapitanluffy:
Plugin for GraphQL development inside Sublime Text


kapitanluffy:
>  - WARNING: '.no-sublime-package' is defined. Please verify that it is *really* necessary

Since I am planning to use node packages for parsing the graphql schema in the future

> - WARNING: The binding ['ctrl+enter'] is also defined in default bindings but is masked with a 'context'

This is used for running the query under the cursor similar to what Graphiql does

rchl:
> > * WARNING: '.no-sublime-package' is defined. Please verify that it is _really_ necessary
> 
> Since I am planning to use node packages for parsing the graphql schema in the future

That is a bad idea since all the files within the `Packages` are indexed and can affect the general performance of ST. So having hundreds or thousands of files in there is not good.

You could adopt the convention that LSP has started and put files in `Cache/../Package Storage`. See https://github.com/sublimelsp/LSP/blob/cd3a7f348e2a7faaa824d3c2945b0c64ae2e9068/plugin/core/views.py#L128-L137

The files there are not indexed.

kapitanluffy:
Thank you. I'll give it an update.

kapitanluffy:
@rchl I removed the `no-sublime-package` for now. Is there anything else I need to do for the bot to recheck it?

rchl:
Create new release in your repo and request review from the bot on this page.

rchl:
Went ahead and tried it. Some issues (you can decide if those should block the release):
 - The documentation mentions shortcuts using `ctrl` key but on mac it's `command` instead
 - The `.grapqhlconfig` can be in [various formats](https://graphql-config.com/introduction) but the package only supports JSON format.
 - The parsing of lines is too string [here](https://github.com/kapitanluffy/sublime-graphql-playground/blob/cc5d434dea8e76879d65f078e2519502ed2eaaf5/graphql_playground.py#L198). It doesn't account for something like `query($countryCode: String!) {` which is a query that can group multiple queries within it. All my queries looked like that so I haven't stopped testing at this point.

kapitanluffy:
> 
> 
> Went ahead and tried it. Some issues (you can decide if those should block the release):
> 
>     * The documentation mentions shortcuts using `ctrl` key but on mac it's `command` instead
> 
>     * The `.grapqhlconfig` can be in [various formats](https://graphql-config.com/introduction) but the package only supports JSON format.
> 
>     * The parsing of lines is too string [here](https://github.com/kapitanluffy/sublime-graphql-playground/blob/cc5d434dea8e76879d65f078e2519502ed2eaaf5/graphql_playground.py#L198). It doesn't account for something like `query($countryCode: String!) {` which is a query that can group multiple queries within it. All my queries looked like that so I haven't stopped testing at this point.

Thank you @rchl for testing it out. I updated the README for the keybinds and graphql config formats. I also fixed the pattern for parsing query operation names.

I reverted the tags to `1.0.0` and will be releasing the yml/tml support probably this weekend if time permits.

rchl:
> I also fixed the pattern for parsing query operation names.

Note that that won't fix compatibility issue with queries like that:
```
query($countryCode: String!) {
    album(countryCode: $countryCode, limit: 6) {
        items {
          title
        }
    }
}
```

since it tries to parse and run query from the first line which isn't really a query on its own.

kapitanluffy:
It seem to work on my end 
image

rchl:
Getting `"message": "Unknown operation named \"\"."` with my query.

Some possible improvement (maybe only in debug mode?) could be to show the full query that was executed.

Also the variables view is strangely disconnected from the main gql file. That would probably be a big change but being able to enter a layout like the original playground has would probably be the best. That is, enable split view with the variables view included.

kapitanluffy:
Yeah. I should probably add a roadmap thing for this. 
- Debug mode is planned. There was way to inspect the http request
- The variables view detaching is a known bug 😟 ..I was pretty sure I already fixed that before but I'll prioritize it.
- Support other graphql config formats. I already have toml and yaml dependencies installed 👍
- Ability to customize the headers (for auth? I haven't tried auth on gql personally though)

Maybe try to open and close the graphql file? If possible maybe you can provide me on discord that part of the script? If not you can always add a print statement on
the file `graphql_playground@sendRequest`

As for the layout, I initially thought I should use splitting to groups to make it look literally like playground but I find it to be way more complicated than just grouping the tabs.

rchl:
I have been triggering the query while on the second line (`album(...`) and that doesn't work. So the keyboard shortcut allows that but annotation is only shown on the main query block. Should be consistent.

BTW. You can use `primary` key for all platforms since it will evaluate to either ctrl or command automatically.

kapitanluffy:
@rchl can you try it again? I added a `debug` setting to preview the sent payload. I also tried running a similar query in the second line and it works. Maybe because it was sending an empty string instead of `None` before. Thank you.

kapitanluffy:
So, I updated the package with the following..

- Add support for `toml` and `yaml` graphql config files
- Add support for [passing headers](https://graphql-config.com/schema#passing-headers)

braver:
Sorry for letting this sit here so long, but thanks for working through the feedback with @rchl. You can always improve the package some more of course, but for now just let's put it out here 🏎️

PR #8373, opened 2021-09-22
Transfer WolframLanguage package to WolframResearch

bostick:
Hi, I work for Wolfram Research and I am the developer for our official Sublime package.

I would like to transfer the WolframLanguage package from https://github.com/ViktorQvarfordt/Sublime-WolframLanguage to https://github.com/WolframResearch/Sublime-WolframLanguage, so I am opening this PR. 

Here is a message I sent to will@wbond.net and CCed viktor.qvarfordt@gmail.com on August 31, but I have not yet heard a reply.

> Hello,
> 
> 
> I am the developer of the official Wolfram Language package for Sublime Text:
> 
> https://github.com/WolframResearch/Sublime-WolframLanguage
> 
> 
> We (Wolfram Research) are interested in taking over the WolframLanguage package in Package Control:
> 
> https://packagecontrol.io/packages/WolframLanguage
> 
> and I'm not sure what the process is to achieve this. (or indeed if this is even possible!)
> 
> 
> 
> This is a fork of the original package from Viktor Qvarfordt:
> 
> https://github.com/ViktorQvarfordt/Sublime-WolframLanguage
> 
> 
> 
> You can see that there has not been any development for a couple of years.
> 
> Wolfram Research is interested in taking over ownership of this package and providing ongoing development and support.
> 
> We have been in communication with Viktor and he has agreed to transfer ownership to Wolfram Research.
> 
> 
> Has Package Control dealt with kind of thing before?
> 
> What should be the next step?
> 
> 
> Brenton

Maybe transferring can be as easy as merging this PR?

### DIFF
--- a/repository/w.json
+++ b/repository/w.json
@@ -668,7 +668,7 @@
 		},
 		{
 			"name": "WolframLanguage",
-			"details": "https://github.com/ViktorQvarfordt/Sublime-WolframLanguage",
+			"details": "https://github.com/WolframResearch/Sublime-WolframLanguage",
 			"labels": ["language syntax"],
 			"releases": [
 				{


rchl:
Note that since you've included and LSP Plugin in your code, that makes it compatible with ST4 and above so the minimum version needs to be updated.

rchl:
Actually, the code is even importing from the LSP package without catching the import error so it will just crash if the LSP package is not updated.

So if the LSP part is optional then this needs to be handled differently. And if the LSP part is required then that should be stated in the readme (there is no way to make package depend on another one automatically).

bostick:
Thank you for taking a look at this! I do want to keep LSP optional, so I will add error handling for when LSP is not installed.

ViktorQvarfordt:
I agree to this

braver:
Oh, alright! @bostick let me know when you’re ready for me to hit that button.

bostick:
I pushed up a change to handle LSP not being installed, so I think that addresses the only concern.

This is the first package that I have pushed to Package Control, so I'm not sure if I'm missing anything else.

If this looks good to everyone else, then let's hit that button!

rwols:
It should also require a minimum ST build version of 4000, seeing as you’re using constructs from LSP that are only available since that version.

rwols:
On second though, I guess the ImportError catch will take care of that.

rwols:
Note that I have plans to switch to the python 3.8 runtime when package control libraries can run in 3.8: https://github.com/sublimelsp/LSP/issues/1389

when that happens, all helper packages must be upgraded to the py38 runtime as well.

bostick:
Correct, I'm only trying to support the latest version of LSP. If there are any errors at all, then there simply won't be any LSP support.

braver:
Sounds good all around. 
Just as a tiny tip/enhancement, you might want to make use of `.gitattributes` to exclude some files from the package archive to save some storage and data for users (e.g. docs and anything else that isn't part of the package functionality).

If all of this works, tag that latest version and we're good to go.

bostick:
I added a `.gitattributes` file and all of the `export-ignore` entries that were needed. I did not know that Package Control used `git archive` to make releases. That is good to know.

I also moved the tag, so I believe everything is ready.

bostick:
Thanks!

I see it is now pointed to the new repo.

I was not thinking and I added `messages.json export-ignore` to `.gitattributes`. I did not get any messages when I installed the package, and I think that I should have left messages in the archive. I will fix that up in the next release.

PR #8375, opened 2021-09-24
Add clean-css

1000ch:
Add [clean-css](https://github.com/1000ch/Sublime-clean-css) that provide [`cleab-css/clean-css`](https://github.com/clean-css/clean-css) functionality on Sublime Text.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1340,6 +1340,17 @@
 			]
 		},
 		{
+			"name": "clean-css",
+			"details": "https://github.com/1000ch/Sublime-clean-css",
+			"labels": ["css", "formatting"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "CleanCSS",
 			"details": "https://github.com/stolksdorf/CleanCSS",
 			"releases": [


braver:
We'd really prefer if you try not to have to use the `.no-sublime-package` file. Installing node_modules into the client machine, and into the User package directory is really inelegant. Generally if you need some node utils or packages, the approach is to describe the dependencies in the readme and have users install them on their machine in a way that they're available via PATH. That way you're also off the hook in terms of keeping them up to date (vulnerabilities etc) and the node environment as a whole, and can just talk to the cli interface of the utils.

braver:
By the way, sorry for letting this sit here for so long, but we're a bit short on maintainers and were not really sure how to approach some more complicated packages.

1000ch:
@braver Thank you for your feedback.

> Generally if you need some node utils or packages, the approach is to describe the dependencies in the readme and have users install them on their machine in a way that they're available via PATH. That way you're also off the hook in terms of keeping them up to date (vulnerabilities etc) and the node environment as a whole, and can just talk to the cli interface of the utils.

I understand the point. It may cause additional instruction to install the package for users if we mention another point, and users still need to update the package to the latest by their own effort. Also, it would be slightly difficult to keep the package work because the package cannot specify the version of cli installed globally on users' environment (`clean-css` for here).

I personally would like to continue this approach from my perspective as a package developer because of the above reasons. Do we still have to avoid `.no-sublime-package`?

braver:
Ok, just so long as the alternatives are considered there isn't much more to say against this approach. Let's ship it!

PR #8381, opened 2021-09-30
Rename "ayu" to "Theme - Ayu"

dempfi:

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -2566,17 +2566,6 @@
 			]
 		},
 		{
-			"name": "ayu",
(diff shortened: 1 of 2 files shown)


dempfi:
@braver do you happen to know how autoinstallation will work? Will the users need to update the themes in their settings?

deathaxe:
Package is renamed automatically and sublime-theme files are set in settings without path, so transition should be seamless.

braver:
Please respond and/or pass the tests to continue the PR

PR #8382, opened 2021-09-30
Remove all references to my repositories

kallookoo:
I have decided to delete my plugins and their corresponding repositories.

### DIFF
--- a/repository/w.json
+++ b/repository/w.json
@@ -294,16 +294,6 @@
 			]
 		},
 		{
-			"name": "WebSearch",
-			"details": "https://github.com/kallookoo/WebSearch",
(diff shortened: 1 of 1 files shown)


CoodyCZ:
> I have decided to delete my plugins and their corresponding repositories.

Hi, is there any option to get your great wordpress package for sublime?

kallookoo:
@CoodyCZ I just restored them to my profile and there are two repositories. One generates all of the other's files. If you want to take over, I will transfer them to you without problems.

braver:
I can imagine users would still like to have these. Let me know if you want to continue with the removal or something else.

PR #8387, opened 2021-10-07
Add Plugin - Expand and Edit

aafulei:
I would like to submit a new Sublime Text plugin `Expand and Edit`. This plugin offers a set of convenient semantics that combines expand and edit commands. With it, you can expand to word/subword and cut/copy/paste/swap with one command.

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -1438,7 +1438,6 @@
 			"labels": ["search", "text navigation", "text selection"],
 			"releases": [
 				{
-					"platforms": "*",
 					"sublime_text": "*",
 					"tags": true
 				}
--- a/repository/e.json
+++ b/repository/e.json
@@ -1549,6 +1549,18 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


aafulei:
Hi @braver it's me -- author of the [`Exact Quick Find`](https://packagecontrol.io/packages/Exact%20Quick%20Find) plugin. When you have time, would you please take a look at this plugin? There are several machines on which I would like to deploy this `Expand and Edit` plugin -- from package control. Thank you very much!

braver:
Alright, looks pretty handy and doesn't seem to really overlap with anything already in PC. Also you did everything correct in terms of readme, gitattributes, menu entries and great job on the keybindings (ie. not shipping any but plenty of examples) 👍🏻 I'm not going to really dig into your code, it looks clean enough and it seems to work. If you want more feedback there we have Discord channel with very active users. 

So, let's ship it!

PR #8388, opened 2021-10-08
Add QuickFileSwitcher

magnusdahlstrand:
I tried out a few different packages that seemed to do the same thing as this plugin does, but a few I couldn't get to work at all and a couple of others required confusing configurations to work. This plugin lacks configuration and works across all file extensions automatically.

### DIFF
--- a/repository/q.json
+++ b/repository/q.json
@@ -419,6 +419,18 @@
 					"branch": "master"
 				}
 			]
+		},
+		{
+			"name": "QuickFileSwitcher",
+			"details": "https://github.com/magnusdahlstrand/SublimeQuickFileSwitcher",
+			"readme": "https://github.com/magnusdahlstrand/SublimeQuickFileSwitcher/blob/main/README.md",
+			"labels": ["switch", "toggle", "quick", "utility"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
 		}
 	]
 }


Ultra-Instinct-05:
Please fix the CI failing tests. After adding a repository entry, you need to run
`ChannelRepositoryTools: Test Local Repository (Current File)` to make sure everything is in alphabetical order.

A couple other points as well.

1. Why do you have a `.no-sublime-package` file ? From what I gather, it should generally be used only in specific situations. Your plugin seems simple so I don't understand the reason.

2. https://github.com/magnusdahlstrand/SublimeQuickFileSwitcher/blob/main/Default.sublime-keymap#L2 I believe <kbd>alt + 0</kbd> is a default key binding. Please use some other non conflicting binding.

3. https://github.com/magnusdahlstrand/SublimeQuickFileSwitcher/blob/main/Main.sublime-menu#L16-L49 - Please don't use the `open_file` command to add 2 different menu items for settings. When it comes to settings, the best experience comes when they are viewed side by side. Consider using the `edit_settings` command instead.

Your plugin looks really similar to the built in `switch_file` command (Switching files between same named files with different extensions). Any reason your package provides better support than `switch_file` (From a look at the code, it just seems a barebones version of `switch_file`) ?

magnusdahlstrand:
Cheers for having a look at this @Ultra-Instinct-05, and thanks for your input.

1. The `.no-package-file` was from the plugin I based this off of, I guess it's not needed.

2. I had a look through the default keybindings (I'm on Mac) and couldn't find anything bound to `alt+o`.

3. I wasn't aware of said command, where in the documentation should I look for it, I had a look but couldn't find it.

4. You're absolutely right! I had no idea the `switch_file` command with `extensions` args existed. My plugin doesn't need any configuration, but for my use-case the configuration is minimal, so I'll scrap the plugin.

Thanks!

rchl:
> Sorry, I thought that was alt + 0, but even then alt + o is also bound to switch_file on Windows. Looks like Mac doesn't use alt + o

Depending on the keyboard layout, alt+o can produce accented character.

PR #8389, opened 2021-10-08
add Project And Workspace Management

tshrpl:
I would like to submit a Sublime Text plugin `Project Management` for, as the name suggests, managing sublime project and workspace files

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -86,16 +86,6 @@
 			]
 		},
 		{
-			"name": "Project Management",
-			"details": "https://github.com/tshrpl/ProjectManagement",
-			"releases": [
-				{
-					"sublime_text": ">=3000",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "PackageResourceViewer",
 			"details": "https://github.com/skuroda/PackageResourceViewer",
 			"releases": [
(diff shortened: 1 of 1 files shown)


repository/p.json
```diff
@@ -85,6 +85,16 @@
 				}
 			]
 		},
+		{
+			"name": "Project Management",
```
rchl, line 89: The entries are supposed be sorted by "name".

tshrpl: i have fixed that now

tshrpl:
sorry for the late reply, I was busy with school work

I have changed the plugin name from `Project Management` to `Project And Workspace Management`. Hopefully this solves the issue.

rwols:
I’d go further and suggest to make a pull request with your desired changes in the existing Project Management package.

tshrpl:
I found an [issue](https://github.com/randy3k/ProjectManager/issues/32) in the old plugin repository requesting the same features that i implement. It seems it's not possible to add these functionalities because the old plugin saves the project/workspace files in a centralized location, and my plugin saves them inside the project folder itself.

rchl:
Can you explain why storing projects in a centralized location is not compatible with switching workspaces?

@randy3k was open to implementing that idea...

tshrpl:
Well, it is not exactly incompatible. Storing the project/workspace files inside each project gives users easy access to the files to save them in version control or create new build systems (like a visual studio solution file). in fact, someone already 
made a [pull request](https://github.com/randy3k/ProjectManager/pull/92) in randy3k's repository, but the code changes were significant and they were unable to merge pr.

braver:
I can follow that. Some people like having their project files in one place, but it's just as common to store them in the project root dir. I would recommend really clearly communicating that difference in philosophy, because you're starting with ~60k installs less and it's not necessarily immediately clear why one would use your package over the other options. Other than that, your package looks good so let's ship it.

jrappen:
@tshrpl Your package is not compatible as stated here with `"sublime_text": ">=3000"`. Please fix your package or fix this here with a new PR and a limit to ST4.

tshrpl:
Hey thanks for pointing that out, I'm not sure why my package is not compatible with ST3 so for now I'm limiting it to ST4 and opened a pr for the same

PR #8391, opened 2021-10-12
Add Xcode build configuration, strings and project syntax

peterthomashorn:
There currently are no packages available which provide the syntax definition for *.strings, *.xcconfig or *.pbxproject files.

### DIFF
--- a/repository/x.json
+++ b/repository/x.json
@@ -45,12 +45,45 @@
 			]
 		},
 		{
+			"name": "Xcode Build Configuration",
+			"details": "https://github.com/peterthomashorn/xcode-build-configuration-language-for-sublime-text",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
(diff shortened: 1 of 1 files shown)


braver:
Looks good basically, but wouldn't it be nicer to have these syntaxes in one package? I can imagine that might make it easier to users, assuming one would usually want support for all the Xcode files at once.

peterthomashorn:
I see that valid point. Though the definitions are all independent from each other. I have not consciously come across such "bundle packages" yet. I would prefer sticking to the principle of one package for one syntax definition rather than a blurry "bunch of Xcode syntax definitions" package. I think it improves the transparency of changes for the users in the end.

And as I just found out I need to rename the "Xcode Strings" package. [That is not a file format specific to Xcode but the platform in general](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/LoadingResources/Strings/Strings.html). I will do.

braver:
Alright, I don’t disagree 👌🏻

PR #8392, opened 2021-10-14
Add Astro package

Destiner:
The package provides syntax highlighting for [Astro](https://astro.build), which is a frontend framework for static sites.

Syntax grammar is based on the official [syntax file](https://github.com/snowpackjs/astro-language-tools/blob/main/packages/vscode/syntaxes/astro.tmLanguage.json) for VSCode.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -1932,6 +1932,17 @@
 			]
 		},
 		{
+			"name": "Astro",
+			"details": "https://github.com/Destiner/sublime-astro",
+			"labels": ["language syntax", "astro"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Async Mindset Snippets",
 			"details": "https://github.com/punkave/sublime-async-mindset",
 			"labels": ["snippets"],


braver:
Maybe move the syntax into the top level directory?
Wouldn't it be nicer to extend the existing HTML syntax? (it looks like this is basically HTML but with addons?) Ie. similar to template languages like liquid?

Destiner:
Thanks for feedback.

>Maybe move the syntax into the top level directory?

Done.

As for extending HTML syntax, I think you're right. I'm not sure what was the intent of the original author, I merely converted it into sublime-friendly format. I'd keep the current format to ease maintenance of the package.

braver:
Well maybe extending isn’t possible or practical in VS Code, I wouldn’t know. It’s not a problem but probably means your highlighting of HTML (and maybe other embedded languages like JS and CSS) is a bit more limited than what people are used to.

Destiner:
That might be true, although personally when working on .astro files I didn't notice anything off. Maybe if someone will have issues, they could open an issue in the package repo?

braver:
Ah, the convertor fails over `\G`. Well if it works now it will probably stay working for a while. It's a shame to have to add new tmLanguage syntaxes to the repository, but there is no rule against it. 🤷🏻

PR #8393, opened 2021-10-22
Add Nuke-Tools-St

sisoe24:
A Sublime Text package that allows to send python or blinkscript code to be executed inside Nuke.

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -1149,6 +1149,18 @@
 			]
 		},
 		{
+			"name": "NukeToolsST",
+			"details": "https://github.com/sisoe24/Nuke-Tools-ST",
+			"author": "virgilsisoe",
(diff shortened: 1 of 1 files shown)


braver:
Tip: you can use `.gitattributes` to "export-ignore" some files (test, documentation, ideas config) so they don't end up in your package, saving space and bandwidth. 
We recommend against shipping keybindings because they tend to conflict with other packages. Instead document in your readme how users can set up their own bindings. Also, your command probably doesn't need to be in every menu, especially not unconditionally in the context menu. Space is limited!

sisoe24:
Thank you for your time!

I've implemented all of your suggestions. Should I create a new release tag or create back the initial one?

braver:
Just create a new tag 👍🏻

sisoe24:
ok. let me know if there is something else. 

thanks

braver:
Nope, looks good 👍🏻

PR #8394, opened 2021-10-24
Add Clojure Sublimed

tonsky:
### Checklist

- [x] Local tests passed
- [x] Used `tags: true`
- [x] Have a README.md
- [x] Similar packages: see below

### About

Sublime Clojure is nREPL client and syntax definition for Clojure. Latter is required for REPL to operate: detect form boundaries, find namespace symbol. nREPL is the most popular REPL server in Clojure ecosystem.

The main features of this plugin are inline evaluation (via annotations and phantoms) and convenience (parallel evaluation, formatted stacktraces, execution time)

Screenshot 2021-10-25 at 01 37 11

### Similar packages

[Sublime REPL](https://packagecontrol.io/packages/SublimeREPL):
 
- not specific to Clojure, but can run Clojure
- requires local lein project
- manages process for you
- no remote REPL

[Tutkain](https://packagecontrol.io/packages/Tutkain):

- Uses streaming Socket Server REPL
- Requires an extra REPL panel to operate
- Can’t keeps multiple eval results on a screen simultaneously
- Can’t show stack traces inline in editor
- Can’t eval several forms in parallel
- Can’t eval non well-formed forms (e.g. `(+ 1 2`)
- Can’t eval infinite sequences
- Can’t redirect `*out*`/`*err*` to `System.out`/`System.err`

Thanks!

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1480,6 +1480,18 @@
 			]
 		},
 		{
+			"name": "Clojure Sublimed",
+			"details": "https://github.com/tonsky/Clojure-Sublimed",
+			"labels": ["clojure", "repl", "nrepl"],
+			"donate": "https://www.patreon.com/tonsky",
+			"releases": [
+				{
+					"sublime_text": ">=4075",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "ClojureDocSearch",
 			"details": "https://github.com/Foxboron/ClojureDoc-Search",
 			"releases": [


tonsky:
There’s now a blog post that explains details of package implementation and capabilities https://tonsky.me/blog/sublime-clojure/

Is this PR blocked by failed CI? I tried to look into it but don’t understand if it’s caused by my package, my PR or by CI itself.

rwols:
There is also LSP with the clojure-lsp language server.

tonsky:
LSP does not provide Eval, which is the main purpose of REPL and this package

tonsky:
1 month anniversary of this PR 🎂
Please talk to me. Any feedback is appreciated, even “we’ll take a look in the next five years” is better than silence
Thanks 🙏

braver:
> even “we’ll take a look in the next five years” is better than silence

I don't like making promises I can't keep, and silence is better than "thanks, we'll maybe look at it eventually". We don't have the capacity currently to properly review more complicated packages, sorry.

Maybe have another look at the [naming rules](https://packagecontrol.io/docs/submitting_a_package#Step_2)?

The package otherwise looks mostly ok I guess, it's shame the bot doesn't want to test it for some reason.

tonsky:
Thank you for replying!

It’s actually very helpful, even without any promises:

- I know know my package was seen, e.g. I didn’t make any stupid mistake when submitting
- I know I shouldn’t expect it to be merged any time soon, so I can find other means to distribute
- I understand the situation better and can worry less

Thanks!

RE: name, I have seen the rules but still couldn’t come up with a better name. It’s a package for Clojure, but simple “Clojure” would conflict with built-in package. I am happy to change to “Clojure” if it wouldn’t.

RE: using space in the name, documentation says “search indexer will properly split words” but when I run “Install Package” I see lots of names not split: “RustFormat”, “RustAutoComplete” etc. Also lots of packages use space already and don’t seem to have a problems, and I like how it looks with space much more. “SublimeClojure” or “sublime_clojure” both look like package name made for computers, “Sublime Clojure” looks like it was made for humans.

I would change it if it’s a stopper, of course, but if it’s just a recommendation, I would like to keep “Sublime Clojure”.

braver:
The space isn't a problem, as you noticed plenty of packages use a space. "Sublime Clojure" doesn't communicate anything, the default "Clojure" is also "Sublime" 😉. We usually see something like "Better Clojure" or "Clojure Improved".

braver:
> but if it’s just a recommendation ...

It's a very strong recommendation: if we don't uphold that guideline we just get more and more package with the Sublime prefix. In the context of package control, which is what we're talking about here, "Sublime" is truly an empty term and just adds noise. `Clojure Sublimed` is cute though, I like that. So, work with us a bit here please.

tonsky:
Renamed to Clojure Sublimed

tonsky:
Thank you 🙏 🙏 🙏 🎉 🎉 🎉

PR #8396, opened 2021-10-28
Add Django Snippets Pro Package

Ali-Aref:
Hi,
This is **Django Snippets Pro** package for supporting Django snippets in Sublime Text. However there are other packages doing the same for sublime text but, it supports the newest **Django3**, **models**, **forms**, **fileds**, **python and html files**, **imports**. it has **file templates** and ad-don autocompletions and much more.

It's an all in one package where developers don't need any extra package for HTML or for Django and so on. All Django related stuffs are categorised and implemented here.

The package is written on understandable syntax with comments for developers whom want to have contribution and for those want to customise for their extra needs.    

It's fully tested! and I'm always available to helps deves for new features requests and solving any issues if availed.    

I am a 5 year Django developer and I were using my package almost for 2 years and decided to share these snippets with other sublime users. I hope It be approved for Django developers better experience.

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -1057,6 +1057,17 @@
 			]
 		},
 		{
+			"name": "Django Snippets Pro",
+			"details": "https://github.com/Ali-Aref/Django-Snippets-Pro",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": ["*"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Django Starter",
 			"details": "https://github.com/geekpradd/Sublime-Django-Starter",
 			"releases": [
--- a/repository/d.json
+++ b/repository/d.json
@@ -1059,7 +1059,7 @@
 		{
 			"name": "Django Snippets Pro",
 			"details": "https://github.com/Ali-Aref/Django-Snippets-Pro",
-			"labels": ["django", "python", "snippets", ],
+			"labels": ["django", "python", "snippets"],
 			"releases": [
 				{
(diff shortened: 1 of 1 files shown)


braver:
Is it not possible to provide these improvements as an upgrade to one of the existing packages?

Ali-Aref:
Actually It's better to be a stand alone package because
-  The completion and triggers differs form other packages (as I checked other packages, but on some parts it's similar to "djaneiro", however djaniero package is outdated and doesn't supports new features, in addition on other snippets it has it own completion style which differs form any other package. if I want to upgrade an existing package that package will completely loss it's completion style.)
- I'm just 22 and more energetic to update for newer versions of django. or if there were any issues or  feature request I will always be available to maintain this package.

braver:
Well that first argument makes sense. As for the second one I would say you could also direct that energy towards contributing to other packages 😉 . That said, I agree the addition makes sense.

So, one thing to change is that you should set up your preferences files to open in split view. [Example](https://github.com/aziz/SublimeHyperClick/blob/master/Main.sublime-menu#L20).

Ali-Aref:
Menu is updated! now it's opening on split windows.

braver:
Thanks! Sorry for letting this sit here so long but the bot doesn't want to play nice with this one. We're looking at it, so sit tight for a while 🤞🏻

Ali-Aref:
@braver any updates ?

braver:

Ali-Aref:
Ok I hope so..

PR #8397, opened 2021-10-28
Add Comment Marks package

maegul:

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2980,6 +2980,18 @@
 			]
 		},
 		{
+			"name": "Comment Marks",
+			"details": "https://github.com/maegul/comment_marks",
+			"labels": ["comments", "code navigation"],
+			"author": "maegul",
+			"releases": [
+				{
+					"sublime_text": ">=4050",
+					"tags": true
+				}
+			]
+		},
+		{
 			"details": "https://github.com/ajayexpert/comment-snippet",
 			"labels": ["snippets"],
 			"releases": [


maegul:
Not trying to nag here ... but have I done something wrong with this submission/pull-request?

braver:
Please resolve the conflict and maybe have a look at our [new checklist](https://github.com/wbond/package_control_channel/blob/master/.github/PULL_REQUEST_TEMPLATE.md)

braver:

maegul:
New Checklist:

	- Keybindings only available through menu (which is detailed in the README)

repository/c.json
```diff
@@ -3003,6 +3003,18 @@
...
+			"name": "Comment Marks",
+			"details": "https://github.com/maegul/comment_marks",
+			"labels": ["comments", "code navigation"],
+			"author": "maegul",
+			"releases": [
+				{
+					"sublime_text": ">=4050",
+          "tags": true
+        }
+      ]
```
braver, line 3015: indentation got kinda messy here

braver:
It really helps if you tell us a little bit about your package. In any case you should add a little description to your GitHub repo, that's picked up by package control and helps people find your package and understand what it's for. 
It also looks like your key bindings would also get loaded (ST doesn't really care about the filename and will probably load all `*.sublime-keymap` files it can find.

maegul:
> It really helps if you tell us a little bit about your package. In any case you should add a little description to your GitHub repo, that's picked up by package control and helps people find your package and understand what it's for.

* What's wrong with the already existing [readme](https://github.com/maegul/comment_marks/blob/0.1.9/README.md)?

> It also looks like your key bindings would also get loaded (ST doesn't really care about the filename and will probably load all `*.sublime-keymap` files it can find.

* This isn't the case on my machine (FWIW).  
* Also the `Brackethighlighter` package does something similar (see [Menu File here](https://github.com/facelessuser/BracketHighlighter/blob/223ffd4ceafd58686503e3328934c039e959a88c/Main.sublime-menu#L252)), which appears to work as intended (ie, the `Example.sublime-keymap` file is not loaded) on my machine.
* [Unofficial Docs](https://docs.sublimetext.io/reference/key_bindings.html#naming-keymap-files) indicate that keymap files must start with `"Default"`.
* [Official Docs](https://www.sublimetext.com/docs/key_bindings.html#user_bindings) on User defined key binding files, are consistent with this, though not conclusive.

Some clarity around this might be good, especially if it's going to change.  Otherwise, do you have any suggestions on how to do something similar?  What I like about this approach is that it's low maintenance for package authors but still convenient for users given the item in the Preferences Menu.

braver:
> Unofficial Docs indicate that keymap files must start with "Default".

I learned something new 😄 I didn’t know it was specific about the file names for keymaps. I don’t have any other information than you have access to, so I would expect that if it works as intended now it should stay that way (at least for this major version of ST). So all good there 👍🏻

> What's wrong with the already existing readme?

The description is used for these lines:

Braver - Search - Package Control

maegul:
> I learned something new 😄 I didn’t know it was specific about the file names for keymaps. I don’t have any other information than you have access to, so I would expect that if it works as intended now it should stay that way (at least for this major version of ST). So all good there 👍🏻

Cheers.  Though I do feel like I (and any other package author relying on this) ought to be more certain than I am ... might duck into the discord on this at some point.

> The description is used for these lines:

Thanks!  Sorry for my daftness ... I had forgotten about the GitHub descriptions.  Done now ... [repo here](https://github.com/maegul/comment_marks)

jrappen:
Not sure why this PR used mixed indentation.

jrappen:
Also wrong order in this PR was fixed in wbond/package_control_channel@f63fd2b.

PR #8398, opened 2021-10-29
Add package rbx-lsp

Nexure:
This package enables support for Roblox's custom lua implementation in sublime text, which allows for auto-completion and other semantics, and highlighting through LSP. There are no other packages that do this, except for one which is outdated (by 5 years), since then they've improved upon their language, now dubbed "lua-u" and added a custom type system. This package support all the new changes, and adds more features.

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -462,6 +462,17 @@
 			]
 		},
 		{
+			"name": "rbx-lsp",
+			"details": "https://github.com/Nexure/sublime-rbx-lsp",
+			"labels": ["roblox-lsp", "rbx-lsp", "roblox", "lsp"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "RDate Editor",
 			"details": "https://bitbucket.org/SergeyEg/date-editor",
 			"releases": [


jwortmann:
This PR should be moved to https://github.com/sublimelsp/repository, and the naming convention for the LSP helper packages would be to use a "LSP-" prefix. There is already an open PR for [LSP-robloxlua](https://github.com/sublimelsp/repository/pull/2) which you probably didn't know about, but that PR seems to be kept on hold for some reason. I'm not sure what's the current status there and what would be required to get that package or yours merged to be available in Package Control, though.

rwols:
The blocker is that the Lua syntax isn’t actually applicable to “Roblox Lua” because they’re different syntaxes. But maybe I’m too idealistic about these things 😉

Nexure:
> 
> 
> The blocker is that the Lua syntax isn’t actually applicable to “Roblox Lua” because they’re different syntaxes. But maybe I’m too idealistic about these things 😉

The language is still based primarily off lua, with their own custom modifications to it in order to make the language more efficient. They even have announced they are planning to open source their language in the future. Information about Roblox Lua (LuaU) can be found [here](https://luau-lang.org/). However, I believe it'd be good for both the sublime's and Roblox's community to have something on sublime to make their version of the language to be actually supported with intelligent features through LSP. Though, while looking at the PR you linked it seems to be quite dead as of lately, and lacking some features such as auto-updating functionality to keep the API in sync with Roblox's API, etc. As of right now, the only good IDE you can use as an alternative to Roblox Studio, is Visual Studio Code. I think it'd be great to get a PR like this approved in order to allow developers (and i know few that use sublime), to get that sort of functionality you'd get in in VSC also in sublime.

rwols:
Auto-update sounds appealing. Make a PR in the LSP packages repo so we can continue the discussion there.

jwortmann:
According to the given link, Luau is mostly backwards compatible and isn't very different from normal Lua. But if the different syntaxes is a problem, then you could create another syntax for Luau with an adjusted base scope, and use this adjusted scope in the "selector" setting for LSP. This would ensure that the language server only starts for files with the Luau syntax. For example like this:

```yaml
%YAML 1.2
---
name: Luau
scope: source.lua.luau
version: 2

extends: Packages/Lua/Lua.sublime-syntax

file_extensions:
  - luau # which file extension does Luau use?

variables:
  builtin_types: \b(?:any|nil|boolean|number|string|thread)\b

contexts:
  statements:
    - meta_prepend: true
    - match: \bcontinue\b(?!\()
      scope: keyword.control.continue.luau

  infix-operator:
    - meta_prepend: true
    - match: '(?:[-+*/%^]|\.\.)='
      scope: keyword.operator.assignment.luau
      push: expression-begin

  maybe-type-annotation:
    - match: ':'
      scope: punctuation.separator.type-annotation.luau
      set:
        - match: '{{builtin_types}}'
          scope: meta.type-annotation.luau storage.type.builtin.luau
          pop: 1
        - match: \(
          scope: punctuation.section.parens.begin.luau
          set:
            - match: \)
              scope: punctuation.section.parens.end.luau
              pop: 1
            - match: \,
              scope: punctuation.separator.sequence.luau
            - match: '{{builtin_types}}'
              scope: meta.type-annotation.luau storage.type.builtin.luau
        - include: else-pop
    - include: else-pop

  function-parameter-list:
    - match: \(
      scope: punctuation.section.group.begin.lua
      set:
        - meta_scope: meta.group.lua
        - match: \)
          scope: punctuation.section.group.end.lua
          set: maybe-type-annotation
        - match: ','
          scope: punctuation.separator.comma.lua
        - include: reserved-word-pop
        - match: '{{identifier}}'
          scope: variable.parameter.function.lua
          push: maybe-type-annotation
        - match: \.\.\.
          scope: constant.language.lua
    - include: else-pop

  # TODO add support for type annotations after local variables
```

PR #8399, opened 2021-11-01
Add Mindustry Assembly Package

gigamicro:
May not meet standards? should be at least reasonably up to spec

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -1558,6 +1558,16 @@
 			]
 		},
 		{
+			"name": "Mindustry Assembly",
+			"details": "https://github.com/gigamicro/Mindustry4Sublime",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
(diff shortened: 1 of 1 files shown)


gigamicro:
... that looks like an http error? is that something I can fix?

braver:
No, the bot just doesn't want to play for some reason. We're looking at it. Meanwhile, please have a look at our [new checklist](https://github.com/wbond/package_control_channel/blob/master/.github/PULL_REQUEST_TEMPLATE.md) and maybe tag a release.

braver:

rwols:
FYI you can enable bot integration to run for first-time contributors from the repo settings. I think that would save a lot of hassle.

braver:
Oh they finally made a setting for that? I don’t have access to those settings, maybe @wbond can hit that toggle. That would indeed save a lot of time.

PR #8400, opened 2021-11-01
Add 0x0

axsuul:
Thanks for your consideration!

https://github.com/axsuul/sublime-0x0


braver:
You don't really have to consider python2, Sublime Text 2 doesn't really have users anymore and ST3 runs python3.
You might want to consider also outputting the response elsewhere because it's easy to accidentally clear/overwrite the clipboard.
Other than that, looks good.

PR #8403, opened 2021-11-03
Update svelte-sublime releases

corneliusio:
Uses a separate release branch to maintain support for the Svelte package and ST3.


braver:
Branch based releases are not supported, but you can prefix your tags with e.g. st3- to create a separate stream of releases.

braver:
Please switch to tag based releases also for the st3 versions.

PR #8404, opened 2021-11-06
Add 42-norminette Linter

GlaceCoding:
This is a Linter for the `42 norminette` which is [a school linter](https://42.fr/en/homepage/).

https://github.com/baptistejamin/sublime-42-norminette


braver:
SublimeLinter plugins should be submitted here: https://github.com/SublimeLinter/package_control_channel

PR #8405, opened 2021-11-08
Add poor-mans-tsql-formatter package

marioosh-net:
This plugin is using Node.js library to make Microsoft Transact-SQL look better.
It is using fomatting like in http://architectshack.com/PoorMansTSqlFormatter.ashx

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -1856,6 +1856,16 @@
 			]
 		},
 		{
+			"name": "poor-mans-tsql-formatter",
+			"details": "https://github.com/marioosh-net/sublime-poor-mans-tsql-formatter",
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


braver:
Looks like our bot doesn't feel like it today... we're looking into it. In the mean time, why not have a short look at our [new checklist](https://github.com/wbond/package_control_channel/blob/master/.github/PULL_REQUEST_TEMPLATE.md) to see if you can maybe nip and tuck some things about your package.

braver:

braver:
Also, what won't help is that you haven't tagged any releases. Please open a new PR if you want to try again, we have a new guide in there that will help you go through the process a bit better too.

PR #8407, opened 2021-11-10
CodexAI

necarlson97:
Write code alongside cutting edge AI, leveraging [OpenAI's 'Codex'](https://openai.com/blog/openai-codex/).
This package simply uses the OpenAI API (which currently requires an OpenAI login, and beta access) to provide a code writing companion. It can fill out boilerplate, provide algorithmic solutions, help you transition to new programming languages, and guide your coding process. Simply highlight the code you want continued, press the keybinding, and see how the AI might complete your work.

I found no existing sublime package that hooked up to Codex, nor any other of OpenAI's offerings.

Apologies for any misstep - this is my first sublime package. I wrote it initially only for myself, but found such joy in using it, it seemed a shame not to share. I am happy to make any changes needed.

Thank you for your help.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2300,6 +2300,17 @@
 			]
 		},
 		{
+			"name": "CodexAI",
+			"details": "https://github.com/necarlson97/codex-ai-sublime",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": ["linux"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "codic",
 			"details": "https://github.com/airtoxin/codic-sublime",
 			"releases": [


Ultra-Instinct-05:
A few points

1. Why do you have a `.sublime-package` as a part of the package ? When a user downloads the package via PC, it is automatically zipped as `.sublime-package` & put in `Installed Packages` directory.

2. Please use a `.gitattributes` file using the `export-ignore` keyword to avoid exporting files like these https://github.com/necarlson97/codex-ai-sublime/blob/main/sublime-codex.gif since they are for demonstration purposes and serve no other purpose.

3. You don't technically need a `.python-version` file. If it's not present, ST will assume that the package runs under `3.3` host.

4. https://github.com/necarlson97/codex-ai-sublime/blob/main/codex_ai.py From looking at the plugin, seems to be just a request to this CodexAI API and getting back the data. So people are kind of trusting this end point with their code. Since the API might take time to fetch the results, you could run the task in a background thread to avoid UI freeze ups (Just like how PC does when fetching package data).

necarlson97:
Perfect, this is great feedback - can't make these changes tonight, but as soon as I get time, I will look at this in more detail. Thanks!

necarlson97:
Responding to the points:
1. Gotcha, removing
2. Sounds good
3. Understood, removing
4. These are good points - I am adding disclaimer to the readme to explicitly state that the code is being sent unencrypted to an outside organization, and how this is especially dangerous for HIPPA, confidential info, etc.

As for asynchronously fetching from the API - that would be lovely, but the naive attempt of simply calling a thread does not seem allowed, as as the `sublime.Edit` throws a:
`ValueError: Edit objects may not be used after the TextCommand's run method has returned`
when using `self.view.replace(edit, region, stuff)`

I imagine there is a specific way to perform async edits, and I am [searching for it](https://www.sublimetext.com/docs/api_reference.html), but if you know of an example project that does this well, or have any advice, that would be awesome.
`set_timeout_async` seems promising, but even when moving the fetch call to another command, and using soemthing like:
```
def insert_codex_response():
    self.view.run_command('async_codex')
sublime.set_timeout_async(insert_codex_response, 0)
```
it still seems to freeze the UI...

Thanks again!

Ultra-Instinct-05:
As an example of fetching results from an API (& showing a spinner of sorts in the status bar, like what Package Control does)
https://forum.sublimetext.com/t/how-to-access-thread-results/48860

Ultra-Instinct-05:
Again, do note that my points are not hard & fast rule for the package being accepted (I just try to help with reviews, I don't have merge permissions), but addressing them should result in
1. Quality package being delivered to users.
2. Reducing the time of those who have merge permissions to merge the package to PC.

necarlson97:
Perfect - your comments have been very helpful. I'll continue working and update when these improvements have been pushed.
Thanks again!

necarlson97:
Alright, awesome, those improvements have been [pushed to the repo](https://github.com/necarlson97/codex-ai-sublime/tree/v1.1.0). Thanks!

necarlson97:
Hey - just checking in. Been a little while - anything else anyone needs from me on this?

braver:
- Please look for examples on how to open settings in split view.
- Are you sure you need a keybinding? It's not documented so people will not find (or remember) it. 
- Commands should be available via the command palette (and if they're there you hardly need a keybinding anyway).

necarlson97:
Thanks for the feedback!
1. Pardon, I don't think I understand the first point - is there something incorrect with the settings file? When I open the project settings, it opens in a new window with the side-by-side view - that seems to be the same behavior as (some) other packages.
2. I personally use think the experience works nicely with a keybinding, but will definitely add a command palette so users have choice. The keybinding is documented in the "Use" header of the readme
3. 👍 Added the `codex-ai.sublime-commands` to make codex available from command palette
4. Also, added MIT `LICENCE.txt` file

See v1.2.0 for improvements. Thanks again for your help!

braver:
> is there something incorrect with the settings file

Oh, sorry, never mind... I didn't look closely enough, it's fine 👍🏻 

> The keybinding is documented in the "Use" header of the readme

👍🏻

PR #8411, opened 2021-11-15
Add `CommandsBrowser` package.

Ultra-Instinct-05:
This PR adds the `CommandsBrowser` package to Package Control.

`CommandsBrowser` is a Sublime Text package that allows a user to view all the available core/plugin commands for Sublime Text and Sublime Merge, along with their documentation/source.

Useful if you are into plugin/package development.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2980,6 +2980,17 @@
 			]
 		},
 		{
+			"name": "CommandsBrowser",
+			"details": "https://github.com/Sublime-Instincts/CommandsBrowser",
+			"labels": ["commands", "plugin/package commands", "browser"],
+			"releases": [
+				{
+					"sublime_text": ">=4121",
+					"tags": true
+				}
+			]
+		},
+		{
 			"details": "https://github.com/ajayexpert/comment-snippet",
 			"labels": ["snippets"],
 			"releases": [


braver:
I have a bit of a problem with the key binding on principle, but also because bindings on ctrl are so uncommon on a Mac (ie. you'd really need to consider proper keybindings for all platforms). And nobody remembers chords like that anyway unless you use the tool daily. Setting up your own binding is usually the first step in memorizing it.

Other than that I'm going to assume that it works (you seem to know what you're doing on that end better than I can review anyway), and that you thought long and hard about maintaining all that metadata?

Ultra-Instinct-05:
I can remove the key bindings if that's an issue though. I agree that setting up your own key bindings is the best way forward since that's what your most likely to remember.

> Other than that I'm going to assume that it works (you seem to know what you're doing on that end better than I can review anyway), and that you thought long and hard about maintaining all that metadata?

Well, there is no other way of getting the command details without manually maintaining a list. Even `PackageDev` has to rely on a yaml file with the metadata to provide AC for commands. But I'd anyways consider this close to a complete list. The documentation still needs work, but that's being improved slowly.

braver:
> I can remove the key bindings if that's an issue though.

I guess it's fine because it's a bit of an unlikely chord. I would prefer to comment out the binding and leave it as a suggestion for someone to copy-paste to their own keymap (and maybe tweak the combo). You can do that in your own time though, let's ship it.

PR #8412, opened 2021-11-18
Replace broken and unmaintained direnv package with a working one

misuzu:
[This package](https://github.com/misuzu/direnv-subl) provides proper [direnv](https://direnv.net/) integration.
The [existing package](https://github.com/zchee/sublime-direnv) has broken (it's not even using direnv) implementation and is not maintained.

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -915,10 +915,11 @@
 		},
 		{
 			"name": "Direnv",
-			"details": "https://github.com/zchee/sublime-direnv",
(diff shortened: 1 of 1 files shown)


braver:
@zchee do you agree with the change of ownership? Please respond within 14 days, or we'll probably just move forward.

@misuzu if you want to rename it you need to also provide "previous_names".

zchee (approved):
LGTM
yeah, I'm stop maintain. thanks.

zchee:
cc: @braver

PR #8414, opened 2021-11-19
Add a theme plugin package info named "Vine Theme" based on "Boxy Theme".

xufanchn:
Update description: Add the theme plug-in package information named "Vine Theme" based on "Boxy Theme". The original "Boxy Theme" theme was removed from the ST plug-in market because the original author deleted the code repository on github. However, my friends and I liked this theme very much, so I fork the source code and repackaged it.

### DIFF
--- a/repository/v.json
+++ b/repository/v.json
@@ -384,6 +384,17 @@
 			]
 		},
 		{
+			"name": "Vine Theme",
+			"details": "https://github.com/xufanchn/sublime-vine-theme",
+			"author": "xufanchn",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Vintage Escape",
 			"details": "https://github.com/tonymagro/VintageEscape",
 			"releases": [


braver:
If this is a straight repackage, and the original license allows it (I believe it's MIT so it should be good), we can replace the existing Boxy Theme entry with this one. How's that sound to you?

xufanchn:
> If this is a straight repackage, and the original license allows it (I believe it's MIT so it should be good), we can replace the existing Boxy Theme entry with this one. How's that sound to you?

xufanchn:
> > If this is a straight repackage, and the original license allows it (I believe it's MIT so it should be good), we can replace the existing Boxy Theme entry with this one. How's that sound to you?

yep, I can accept it, the original intention is the same, I just want to use this theme more simply.

braver:
Alright, for this to fly, I think the best would be to not introduce a new package, but replace the details link for the old Boxy entry. You would need to go through things like missing images in the readme, and double checking the package name in things like themes and settings. Also make sure build tools and such are excluded from the package using `.gitattributes`. 

In any case, the entry you propose doesn't pass mustard either because all the documents still talk about "Boxy" and not "Vine". So one way or the other some work to do still.

xufanchn:
ok, thanks. I will check it and resubmit.

PR #8013, opened 2020-08-12
Adds the BetterJinja Package.

Ultra-Instinct-05:
This PR adds the [BetterJinja](https://github.com/Sublime-Instincts/BetterJinja) Package to Package Control.

This package provides syntax highlighting, snippets, completions & more for [Jinja2]() templates.

There is an existing package already which is [Jinja2](https://packagecontrol.io/packages/Jinja2), however the package hasn't been updated to keep up with latest syntax improvements via `.sublime-syntax` for ST3. So this package offers,
- An updated & tested syntax file which follows the official scope naming guidelines as closely as possible.
- Snippets for Jinja2 code blocks.
- Completions for built in tests, filters, functions & loop variables.
- Ability to have Jinja2 line & block comments.
- Certain key bindings to make life a little bit easier.

PR #8415, opened 2021-11-19
Add `BetterJinja` (Part 2)

Ultra-Instinct-05:
Sorry that it took so long. Around the time when I first submitted `BetterJinja`, I just got into my job, so there was a bit of flux and challenge with looking into OSS.

The original PR was https://github.com/wbond/package_control_channel/pull/8013
And based of it, I got the permission to collaborate on the repo, but I will continue with the same suggestions that was offered when `BetterTwig` was reviewed.

1. For ST2 and ST3<3092, point the URL to the old `Jinja2` package https://github.com/kudago/jinja2-tmbundle
2. For ST3 > 3092, use `BetterJinja`

I am choosing this path because its the easiest & hassle free IMHO.

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -685,7 +685,7 @@
 				{
 					"base": "https://github.com/kudago/jinja2-tmbundle",
 					"sublime_text": "<4107",
-					"branch": "master"
+					"tags": true
 				},
 				{
 					"sublime_text": ">=4107",
--- a/repository/j.json
+++ b/repository/j.json
@@ -680,10 +680,15 @@
 		},
 		{
 			"name": "Jinja2",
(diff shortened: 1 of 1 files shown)


Ultra-Instinct-05:
On second thoughts, since ST already has a few ST4 stable releases, I don't think I want to go down the road of ST3 maintenance burden. So changed the ST Build numbers to reflect it.

braver:
Indeed it sounds like a whole lot to trouble with very little profit.
The original package also has tags, I don't think you should change it to branch based release here.

Ultra-Instinct-05:
> The original package also has tags, I don't think you should change it to branch based release here.

Addressed

PR #8416, opened 2021-11-20
added openssl configuration file syntax

finitelife:
This package simply highlights the openssl configuration file syntax as it is slightly different from the INI syntax.
This includes default words, proper comment escaping, quoting, variable highlighting, section/name/value pair colouring, multiline values, directives, etc

### DIFF
--- a/repository/o.json
+++ b/repository/o.json
@@ -859,6 +859,21 @@
 			]
 		},
 		{
+			"name": "OpenSSL",
+			"details": "https://github.com/finitelife/openssl-sublimetext",
+			"labels": [
+				"language syntax",
+				"color scheme",
+				"formatting"
+			],
+			"releases": [
+				{
+					"sublime_text": ">=3103",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "OpenType Feature Bundle",
 			"details": "https://github.com/brew/opentype-feature-bundle",
 			"releases": [


finitelife:
I've resubmitted this as my last PR failed and I accidentally deleted my local repository. author and readme keys deleted, and removed any json files in local repo.

finitelife:
added v1.0.0 tag to personal repository

braver:
Thanks. This feedback still stands though:

> why does your package contains a sublime-color-scheme, and a settings file?

finitelife:
> Thanks. This feedback still stands though:
> 
> > why does your package contains a sublime-color-scheme, and a settings file?

it has additional scopes to highlight

braver:
The built-in scopes shouldn't be treated as some minimal set. There are scopes in there for invalid/illegal input, punctuation, strings and everything else you need. Users want to use the color scheme they like, and they want to use dozens of languages. Nobody wants to have to use a specific color scheme to fully use a specific syntax. In effect, we do not usually accept syntaxes that ship with color schemes, there is hardly ever a good justification for it.

braver:
Please let us know if you want to continue this PR, and maybe have a look at the [new checklist[(https://github.com/wbond/package_control_channel/blob/master/.github/PULL_REQUEST_TEMPLATE.md).

PR #8417, opened 2021-11-22
Add Wrap

futureprogrammer360:
Wrap allows users to wrap the text under cursor or selected with brackets (type is customizable). A placeholder will be inserted before the starting bracket. This is useful when passing a variable into a function/method call as argument or as an index into a list/array.

Example:
Original code: `x` (`x` is selected or under cursor)
After `Wrap`: `abc(x)` (`abc` is selected)

### DIFF
--- a/repository/w.json
+++ b/repository/w.json
@@ -971,6 +971,16 @@
 			]
 		},
 		{
+			"name": "Wrap",
+			"details": "https://github.com/futureprogrammer360/Wrap",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Wrap Plus",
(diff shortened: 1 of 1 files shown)


futureprogrammer360:
I might be understanding things wrong, but aren't comments allowed in the configuration files?

The `sublime-keymap` files are there for users to use as reference when adding their own keybindings. I didn't ship the plugin with default keybindings since that was discouraged in previous packages. At the same time, I didn't want to delete the files altogether since that'd cause problems with the `Preferences: Wrap Key Bindings` command.

futureprogrammer360:
Ah that makes sense, thank you @jwortmann!

jwortmann:
You need another tag for packagecontrol-bot to see the changes.

futureprogrammer360:
Right, forgot about that 😅

PR #5346, opened 2016-03-20
take over VintageLines

co-dh:
The tmanderson/VintageLines seemed abandoned.
I have a pull request sit for 40 days.
I've implemented new features and would like to take over.

PR #8418, opened 2021-11-22
Take over Squirrel

Enduriel:
The [original author](https://github.com/iAmRoland/Sublime-Squirrel) of the squirrel plugin is unresponsive (hasn't replied to my email for months and 0 activity on any of their easily viewable social media for years). I've worked on significantly improving the package since I started working with Squirrel and think it should be used as the default for Package Control.

Hope I'm doing this right, used [this](https://github.com/wbond/package_control_channel/pull/5346) as a template.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -3474,7 +3474,12 @@
 			"labels": ["language", "syntax", "snippets", "completions", "squirrel"],
 			"releases": [
 				{
-					"sublime_text": "*",
+					"base": "https://github.com/iAmRoland/Sublime-Squirrel",
+					"sublime_text": "<3092",
+					"branch": "master"
+				},
+				{
+					"sublime_text": ">=3092",
 					"tags": true
 				}
 			]
--- a/repository/s.json
+++ b/repository/s.json
@@ -3469,8 +3469,8 @@
 		},
 		{
 			"name": "Squirrel",
-			"details": "https://github.com/iAmRoland/Sublime-Squirrel",
-			"author": ["micheg", "Roland Piirsoo"],
+			"details": "https://github.com/Enduriel/Sublime-Squirrel",
+			"author": ["Enduriel", "micheg", "Roland Piirsoo"],
 			"labels": ["language", "syntax", "snippets", "completions", "squirrel"],
 			"releases": [
 				{
--- a/repository/s.json
+++ b/repository/s.json
@@ -3475,11 +3475,11 @@
 			"releases": [
 				{
 					"base": "https://github.com/iAmRoland/Sublime-Squirrel",
-					"sublime_text": "<3092",
+					"sublime_text": "<4107",
 					"branch": "master"
 				},
 				{
-					"sublime_text": ">=3092",
+					"sublime_text": ">=4107",
 					"tags": true
 				}
 			]
--- a/repository/s.json
+++ b/repository/s.json
@@ -3474,11 +3474,6 @@
 			"labels": ["language", "syntax", "snippets", "completions", "squirrel"],
 			"releases": [
 				{
-					"base": "https://github.com/iAmRoland/Sublime-Squirrel",
-					"sublime_text": "<4107",
-					"branch": "master"
-				},
-				{
 					"sublime_text": ">=4107",
 					"tags": true
 				}


Ultra-Instinct-05:
@iAmRoland Are you okay with @Enduriel taking over the maintainer ship of the package ? If so, please respond within 14 days.

iAmRoland:
Sorry, was supposed to get back but completely forgot.

Yeah you can go ahead and transfer the maintainer ship. The changes in your branch seem pretty neat, if I ever use Squirrel again I'll try it out 😎

braver:
Thanks. 

Please don't ship color schemes with your syntax. There are hundreds of color schemes out there and people really want to use what they like. And people like to use their editor for loads of different languages. If you stick to the scope naming guidelines there is usually no need to tweak color schemes, and it allows color schemes author to effectively optimize for a wide range of languages.

Enduriel:
@iAmRoland awesome ty! I genuinely thought something had happened to you since you were inactive on every social media you had linked so glad to hear you're okay.

@braver that is fair, in theory that should be the case, for the two specific themes I included it unfortunately was decidedly *not* so I bundled them together for the people I was sharing it with but I'll move it to a different repo.

Enduriel:
Done

Ultra-Instinct-05:
I forgot to mention another thing. Since your package is using the `.sublime-syntax` format, you need to specify `"sublime_text": ">=3092"` since the format was introduced in that build. To continue for any ST2 users (for whatever goddamn reason), you can do something like this.

```json
"releases": [
    {
        "base": "https://github.com/iAmRoland/Sublime-Squirrel",
        "sublime_text": "<3092",
        "branch": "master"
    },
    {
        "sublime_text": ">=3092",
        "tags": true
    }
]
```

jrappen:
Also, please add a license.

Enduriel:
@Ultra-Instinct-05 done and @jrappen done.

Also I would argue if someone is still using ST2 they deserve anything that happens to them :P

repository/s.json
```diff
@@ -3469,12 +3469,17 @@
...
-			"author": ["micheg", "Roland Piirsoo"],
+			"details": "https://github.com/Enduriel/Sublime-Squirrel",
+			"author": ["Enduriel", "micheg", "Roland Piirsoo"],
 			"labels": ["language", "syntax", "snippets", "completions", "squirrel"],
 			"releases": [
 				{
-					"sublime_text": "*",
+					"base": "https://github.com/iAmRoland/Sublime-Squirrel",
+					"sublime_text": "<3092",
+					"branch": "master"
```
braver, line 3479: Actually, ST < 3092 is not supported either, there is no need to take those versions into account here. Especially not via branch based releases.

braver (changes requested):

braver:
@Ultra-Instinct-05 is strictly correct, but we don't enforce support for < 3092. "ST3 support" is assumed to mean versions newer than that.

jrappen:
If you ask me, you're wasting your time with ST3 support. I believe the vast majority has already moved on.

I'd target `>=4107` (first public stable ST4 Build).

braver:
Honestly, after a package is submitted nobody verifies the compatibility anymore. Make life easy for yourself and just support whatever you’re using yourself. Nobody using an ancient unsupported version of Sublime Text is looking for new packages.

Enduriel:
I think from what I understand the thing I'd most prefer to go with would be @jrappen's suggestion, as if I use features that have been added later than that (which I might end up doing at some point but it'd definitely be a conscious decision) I can cap the release for those versions to a specific tag from what I understand?

And like I don't know, someone might have a rando reason to not use the latest latest version of ST4 which in my opinion is fair enough. If they're using ST4 it will be fine (the sublime-syntax file was first done for my own personal stuff before ST4 even released).

Is that alright?

braver:
I would say just drop the release for < 4107 and if someone has some random reason to use an outdated version of ST you can tell in them (in the readme) how to manually install a specific version of your package. You don't need to do everything via package control, just make life easy for yourself.

Enduriel:
@braver I mean it doesn't cost me anything to point older versions to the other repo just in case, not sure why not to do that?

braver:
Sure, whatever you like. I just want to get rid branch thing.

braver:

Enduriel:
Sorry for taking so long, @braver I've removed the section like you asked.

PR #8419, opened 2021-11-23
Adding V (vlang) syntax highlighting package

elliotchance:

### DIFF
--- a/repository/v.json
+++ b/repository/v.json
@@ -2,6 +2,17 @@
 	"schema_version": "3.0.0",
 	"packages": [
 		{
+			"name": "V",
+			"details": "https://github.com/elliotchance/vlang-sublime",
+			"labels": ["language syntax", "v", "vlang"],
+			"releases": [
+				{
+					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


elliotchance:
@wbond it's been 7 days now - is there any way I can get this reviewed faster?

braver:
I'll send you a Paypal link, send me some money and we can maybe work something out 😉 

Not sure a single character is the best way to get found but 🤷🏻 
Have you tried converting your syntax to the new sublime-syntax (yaml) format? That way you can also get rid of the JSON.

elliotchance:
@braver I'm actually pushing a fork so that it doesn't have to manually be installed. If it works, I'm planning (when I get around to it) rewriting the syntax to sublime format, but also adding more feature like building and debugging - if possible.

Do I need to put in a license now to proceed?

I know "V" is short but that that is the name of the language... https://vlang.io/

braver:
> Do I need to put in a license now to proceed?

No, it's recommended but not required. Let's ship it!

PR #8420, opened 2021-11-23
Add the Create WebP Image package

TomasBarry:
A redo of https://github.com/wbond/package_control_channel/pull/7867 since that one went stale.

The necessary updates have been made and I have verified that this package works.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3938,6 +3938,18 @@
 			]
 		},
 		{
+			"author": "Tomas Barry",
+			"name": "Create WebP Image",
+			"labels": ["WebP", "cwebp"],
+			"details": "https://github.com/TomasBarry/Sublime-create-webp-image",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "CreativeCrocodile",
 			"details": "https://github.com/Abban/CreativeCrocodile",
 			"releases": [


Ultra-Instinct-05:
A couple points, but otherwise seems fine.

1. https://github.com/TomasBarry/Sublime-create-webp-image/blob/master/.DS_Store - I suppose this is something MacOS specific and can be gitignored ?

2. https://github.com/TomasBarry/Sublime-create-webp-image/blob/master/CreateWebpImage.sublime-settings#L9 - 99% of the times, settings in Sublime Text Packages are snake_case by default. But whatever suits you.

3. https://github.com/TomasBarry/Sublime-create-webp-image/blob/master/CreateWebpImage.py#L89 - You use the `input` method only when you are implementing an `InputHandler`. Otherwise, it's not required.

Ultra-Instinct-05 (approved):

TomasBarry:
Thanks for the approval @Ultra-Instinct-05. What needs to happen to get this merged in?

repository/c.json
```diff
@@ -3937,6 +3937,19 @@
 				}
 			]
 		},
+		{
+			"author": ["Tomas Barry"],
```
braver, line 3941: I didn't know we supported arrays here, also doesn't seem necessary in this case.

repository/c.json
```diff
@@ -3937,6 +3937,19 @@
...
+			"name": "Create WebP Image",
+			"description": "Convert images to WebP from Sublime Text",
```
braver, line 3943: You can put this in the GitHub repo description instead for ease of maintenance.

braver:
I think you'd want to be more clear about the prerequisite. From experience I guess I need to install that utility in a way that puts cwebp in my $PATH, but for many users that's a few hurdles to many.

repository/c.json
TomasBarry: Addressed.

repository/c.json
TomasBarry: Addressed.

TomasBarry:
@braver, I've addressed your comments and have included the latest changes as part of the release (a lapse on my part). The usage instructions have been updated to more explicitly reference the official documentation so that I always point to the latest instructions instead of potentially having stale instructions in my README.

Is there anything else I need to cover?

braver:
Nope, all good 👍🏻

PR #8421, opened 2021-11-24
Add SmarterLineMoves package

trych:
A Sublime Text package that overrides Sublime's default line swapping to work in a more predictable way and adds additional commands for moving lines of text.

https://github.com/trych/SmarterLineMoves

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -2435,6 +2435,17 @@
 			]
 		},
 		{
+			"name": "SmarterLineMoves",
+			"details": "https://github.com/trych/SmarterLineMoves",
+			"labels": ["text manipulation", "formatting"],
+			"releases": [
+				{
+					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


braver:
Alright, doesn't always work as smoothly as the default but I can see the value proposition here.

trych:
> Alright, doesn't always work as smoothly as the default but I can see the value proposition here.

Do you mean smoothly in terms of performance? Or in terms of usability? Open for improvement suggestions!

Thanks for merging!

braver:
I especially noticed some stuttering when moving a selection up against the top of the viewport. Although of course in the default implementation most of a multi-line selection moves out of view, it does scroll smoothly. No idea if the API provides any way for you to improve on that though.

trych:
Thanks. Hadn't noticed something like that in my tests, but will keep an eye on it.

PR #8422, opened 2021-11-27
Add plugin Paste As Column

TheClams:
This plugin is just a slight change of the default sublime behavior where in case of multiple line paste it will automatically add space to ensure the block of text is kept aligned, and it also does not require having as many line selected as line in the clipboard.

It was an old plugin hosted on bitbucket, removed last year when bitbucket stopped supporting mercurial repo and some users reached out to get it back.

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -449,9 +449,7 @@
 		},
 		{
 			"name": "Paste As Column",
-			"description": "Paste clipboard content in block mode (like emacs rectangle mode)",
 			"details": "https://github.com/TheClams/PasteAsColumn",
-			"issues": "https://github.com/TheClams/PasteAsColumn/issues",
 			"labels": ["text manipulation", "clipboard", "paste"],
 			"releases": [
 				{
--- a/repository/p.json
+++ b/repository/p.json
@@ -448,6 +448,19 @@
 			]
 		},
 		{
+			"name": "Paste As Column",
+			"description": "Paste clipboard content in block mode (like emacs rectangle mode)",
+			"details": "https://github.com/TheClams/PasteAsColumn",
+			"issues": "https://github.com/TheClams/PasteAsColumn/issues",
+			"labels": ["text manipulation", "clipboard", "paste"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Paste as One Line",
 			"details": "https://github.com/joshdavenport/PasteAsOneLine",
 			"releases": [


Ultra-Instinct-05:
1. https://github.com/TheClams/PasteAsColumn/blob/master/messages.json#L2 - Ideally you'd want a `.txt` file to show as a message rather than markdown, because markdown is not a good way of showing end messages to users on Package installs/upgrades. A simple `.txt` file with install messages/upgrade release notes should be good enough IMO.

Ultra-Instinct-05:
Also you don't need to close & reopen the PR to get `packagecontrol-bot` to rerun the CI checks (If you weren't aware of it that is). Simply press

image

repository/p.json
```diff
@@ -447,6 +447,19 @@
 				}
 			]
 		},
+		{
+			"name": "Paste As Column",
+			"description": "Paste clipboard content in block mode (like emacs rectangle mode)",
+			"details": "https://github.com/TheClams/PasteAsColumn",
+			"issues": "https://github.com/TheClams/PasteAsColumn/issues",
```
braver, line 454: This is the default location for issues, do you don't have to provide it here. You can also leave out the description, it'll take that from your GitHub repo.

braver:
You can also consider adding a keymap file but with the binding commented out (and a menu entry to open it of course). That way users looking for the keybinding can more easily find it inside Sublime Text, and can just copy over the binding to their personal file.

braver:

TheClams:
Which modification are needed ? This plugin was approved like 5 years ago, and I litterally just reverted the removal and change the URL to be up-to-date. I can remove the issues and details but my understanding is that it changes nothing.
As for the keymap it is very subjective: it is part of the readme which is open automatically when the user install it, so I do not see any need to do more. And for the remark about markdown vs text I don't understand it: markdown is just a convention to format text file.

braver:
> This plugin was approved like 5 years ago

Just a month and a bit ago, and by a bot. You need approval from humans, which takes time, and you didn't respond to any of the feedback.

> I can remove the issues and details but my understanding is that it changes nothing.

It keeps superfluous information out of the registry, so that definitely changes something. This is not a recommendation, it's a requirement.

> And for the remark about markdown vs text I don't understand it

The markdown isn't rendered in any way, so it's recommended to stick to plain text. The installation message is something users only see once. Rendering the README there isn't recommended, just saying "thank you for installing" together with a link to the readme is a much better experience for the end user.

TheClams:
When I say the plugin was approved 5 years ago, it is actually 10 years ago: https://packagecontrol.io/packages/Paste%20as%20Column

And as for the readme there is no rendering needed, this is the way I write all my text file.

braver:
> it is actually 10 years ago

Oh… well that’s long enough ago to warrant a fresh review 😂

PR #8423, opened 2021-11-30
Sublime PDML

tajmone:
Sublime Text syntax support for Christian Neumanns's PDML 
(Practical Data and Markup Language):

https://github.com/tajmone/Sublime-PDML

In the future, the package will also support upcoming official PDML extensions, through syntax inheritance, which is why this package is marked as requiring ST4, and also the reason why the Basic PDML syntax was kept minimal.

- Basic PDML syntax support:
    + File extension: `.pdml`.
- Mandatory PDML settings:
    + UTF-8 encoded files.
- Keyboard shortcuts:
    + <kbd>Ctrl</kbd><kbd>Alt</kbd><kbd>n</kbd> &rarr; new-node snippet.

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -644,6 +644,17 @@
 			]
 		},
 		{
+			"name": "PDML",
+			"details": "https://github.com/tajmone/Sublime-PDML",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Pebble",
 			"details": "https://github.com/jbrooksuk/Sublime-Pebble",
 			"releases": [


braver:
Please add the "language syntax" label. 

Please remove the keybinding, in stead add a command palette entry. I do wonder why this functionality should be part of a syntax package?

tajmone:
> Please add the "language syntax" label.

You mean to the repository?

> Please remove the keybinding, in stead add a command palette entry. I do wonder why this functionality should be part of a syntax package?

What's wrong with having a key binding to quickly insert a node? Nodes are the only core elements of the PDML Basic. Having it as a command palette entry would defeat the whole purpose of saving keystrokes to enter nodes.

braver:
There are too few keys on the keyboard to support all packages and you risk overriding that of another package. Usually it's better to help users set up keybindings themselves, e.g. by providing commented out examples in a keymap file and a little note in the README. That also helps people memorize the binding and maybe set up something that makes sense for them.

repository/p.json
```diff
@@ -643,6 +643,16 @@
 				}
 			]
 		},
+		{
+			"name": "PDML",
+			"details": "https://github.com/tajmone/Sublime-PDML",
```
braver, line 648: You'd want a label there to indicate the kind of package.

tajmone: Done! (also rebased on latest master head before amending and force pushing)

braver:
> There are too few keys on the keyboard to support all packages and you risk overriding that of another package

That said at least they're scoped. But note that on a Mac it would be more natural to use super/CMD instead of ctrl.

tajmone:
# Key-Bindings

> That said at least they're scoped. But note that on a Mac it would be more natural to use super/CMD instead of ctrl.

PDML is not available for macOS, only Windows on Linux.

> There are too few keys on the keyboard to support all packages and you risk overriding that of another package. Usually it's better to help users set up keybindings themselves, e.g. by providing commented out examples in a keymap file and a little note in the README. 

The question of whether packages should provide ready-to-use kind bindings or whether the burden of setting them up should be up to the end users seems one of the controversial topics about packages, where package guidelines rub against the real everyday users needs. some of the most popular ST packages ship with dozens of key-binding presets ([MarkdownEdit], to mention but one), and if end users had to manually set all those keys it would be a real drawback.

The key-binding I've added is context specific to PDML, so in the worst case scenario it could override a same-bound key of some third package, which would be a problem only if the package was providing some syntax-agnostic universal functionality which would be highly beneficial to PDML files (which I doubt, since PDML isn't your typical syntax).

Having to occasionally deal with conflicting key-bindings is not something out of the ordinary for non US-users — even the default ST key binding need heavy remapping if you don't have a US keyboard (I have an Italian keyboard and had to fix many of the core key bindings, including those for keys that are present on the layout but for some reason don't map correctly).

> That also helps people memorize the binding 

I doubt so. I often have to refer to a package's README (from within the package or online) to lookup for the associated keys which I don't use often. If I had to manually set all those keys I'd also need to manage a document listing all the custom key-binding (which BTW I do, but fortunately only for the native ST bindings which I had to fix because they don't work on the Italian keyboard).

> and maybe set up something that makes sense for them.

That's a fair point, but the argument would apply to any binding (from ST native choices to all packages).

I personally prefer packages that provide all the required key bindings, especially if well documented, and only change them if they clash with my keyboard, some other package or custom binding I've created.

[MarkdownEdit]: https://sublimetext-markdown.github.io/MarkdownEditing/usage/#key-bindings

braver:
> seems one of the controversial topics about packages

It's not really controversial though. Package ecosystems always run into the limitations of they keyboard and conflicting packages. Some let it go, we take the position that there is a better way.

> I personally prefer packages that provide all the required key bindings

That's fine for you, but for a big ecosystem it doesn't scale. Packages will continually conflict over popular key combinations.   Sure, there are already thousands of packages in the repository with key bindings, but we can learn and improve over time to not make the mistakes that were made before even if we can't go back and fix them.

You could say "if it conflicts for a user just let them figure it out", or "everything already sucks for non-US users anyway", but that's not user-friendly. Users don't have to manually do much at all if they can simply copy-paste over your examples and maybe tweak some things. At least they won't lose any key bindings they're used just by installing your package (without any warning or explanation of what took their bindings away). 

So all were asking here is to play nice in the ecosystem.

All of that said, your binding here are acceptable.

tajmone:
> All of that said, your binding here are acceptable.

Indeed, I wouldn't add a global key-binding with the same ease, since I'm aware of the issues you mentioned.

> You could say "if it conflicts for a user just let them figure it out", or "everything already sucks for non-US users anyway", but that's not user-friendly.

That's actually a topic I value very much, i.e. accessibility (for different locales and special needs), but unfortunately the "everything already sucks for non-US users anyway" is what most non-US users have to deal with.

In a PDML related project ([PML](https://www.pml-lang.dev/)) we're actually investigating the issue of which characters are more user-friendly in terms of their availability in different key-board layouts, and you'll be surprised to learn that there hasn't been much research on this topic either.

On the one hand, there's a long tradition of poor software decision making, where developers sit down and arbitrarily decide what "user friendly" amounts to based on their everyday reality, which usually translated to English/US keyboards first. E.g. lightweight markup syntaxes (markdown and friends) heavily rely on backticks and tildes, although these are seldom found in non-US keyboards, which means having to resort to Alt-codes to access these chars.

On the other hand, the hardware manufacturers seem to give precedence to currency symbols over programming/logic symbols — from what I saw, every keyboard has dedicated keys for the `$`, `£`, `€` and `¥`, whereas few have a key for `~` or `` ` ``. It would have probably been more user friendly if markdown adopted the dollar as a code block delimiter, rather than tilde, since currency symbols are omni-present in any keyboard layout. Yet, they are never used in key-bindings, even though we can trust to find them on most keyboards worldwide.

I've worked with Arabic and Hebrew keyboards in the past, and things just get worst. From discussions with other users on this topic, it seems that Chinese users are suffering the most from these "user-friendly" standards; I was told that they simply don't have expectations for their needs to be considered, and they've come up with custom solutions to work around the problems via custom drivers that can switch keyboard behaviour according to need.

So, this is a widespread problem with a long history — surely not specific to ST, nor one that ST could solve with some guidelines. The thing that struck me most about all this is the actual numbers of world speakers of the languages most neglected by "user friendly" standards — Chinese and Arabic are on top of the list of most spoken languages in the world. But then, when it comes to decisions making in the industry, I guess decisional power is all on one side, and reason on the opposite side. 

> Users don't have to manually do much at all if they can simply copy-paste over your examples and maybe tweak some things. At least they won't lose any key bindings they're used just by installing your package (without any warning or explanation of what took their bindings away).

> So all were asking here is to play nice in the ecosystem.

Probably a truly user-friendly solution in this respect might require a different approach to how key-shortcuts are mapped to the actual keyboard — i.e. instead of having to redefine the bindings, provide a settings file to alter how these symbols are mapped to hardware keys, specifically for the ST application. This would allow to keep the same binding definitions and redirect them to specific keys found on the actual keyboard.

Of course, what you said about conflicting settings still holds true. What I'm trying to say is that many users are already tackling with conflicts with the vanilla key-bindings, so sticking to the guidelines in packages means that the same problematic keys are being propagated over and over, forcing custom interventions on a per-package basis, whereas being able to remap the symbols to custom keys would solve the problem at its roots.

PR #8424, opened 2021-11-30
Fix .gitattributes for Windows

tajmone:
Add to `.gitattributes` the `* -crlf` rule to disable CRLF behaviour on Windows machines where Git is set to `core.autocrlf`, otherwise users won't be able to submit commits with LF EOLs, due to current `.gitattributes` and `.editorconfig` settings.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 text eol=lf
+* -crlf


braver:
Thanks, but your PR also adds a package, that seems unintentional.

tajmone:
> PR also adds a package, that seems unintentional.

Opps! Must have forgot to branch from `master` and used my PR branch instead!

FIXED! I've now amended the commit. 

(sorry for that, things that happen when working late at night to cut corners)

PR #8425, opened 2021-12-01
update sublime text version from >=3000 to >=4000 in Project And workspace Management Plugin

tshrpl:


jrappen:
There is no public Build 4000.

tshrpl:
so what is the lowest ST4 version?

jrappen:
* `stable` channel: <https://www.sublimetext.com/download>
* `dev` channel: <https://www.sublimetext.com/dev>

FichteFoll:
There may not have been a build 4000, but this selector is still fine.

PR #8426, opened 2021-12-03
SphinxRefmate package added : Changes Made

phughes3866:
SphinxRefmate is a cross-referencing manager/aggregator for use within Sphinx Documentation projects that use the standard restructuredText in their source files. 

SphinxRefmate can gather, on demand, five types of references from the Sphinx Doc's source and configuration files, and then present these to the user in a Sublime Text quick-panel for selection/insertion. More details in the package's README.md file.

A similarly named package 'Sphinx Ref Helper' is already available on packagecontrol.io, but this is limited to aggregating references of one type from one Sphinx Doc project, through trawling all the source files. SphinxRefmate is operationally different in that it aggregates references from Sphinx Docs config files (conf.py) and build time inventories (objects.inv). SphinxRefmate is much faster and expansible through this approach. SphinxRefmate can gather more types of references than Sphinx Ref Helper, and from multiple Sphinx Doc projects at once (piggy-backing off the intersphinx_mapping variable associated with the standard Sphinx Intersphinx extension).

SphinxRefmate is never going to be needed in the same degree as food, clothes and shelter. It is however, likely to be a popular plugin for Sublime Text editors of Sphinx Doc projects.

Thank you for your considerations.

Paul Hughes

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -3149,6 +3149,17 @@
 			]
 		},
 		{
+			"name": "SphinxRefmate",
+			"details": "https://github.com/phughes3866/SphinxRefmate",
+			"labels": ["sphinx"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "SPICE Circuit Simulator",
 			"details": "https://github.com/leoheck/sublime-spice",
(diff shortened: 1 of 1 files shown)


phughes3866:
Requested changes made and committed to main branch.

PR #8430, opened 2021-12-05
Fix: Rename branch for package `Toggle Words`

gordio:


braver:
Branch based releases are deprecated and as such we also don't accept changing the branch: just tag a version (e.g. 1.0.0) and change the entry to `"tags": true`.

gordio:
@braver can I use branch definition? My "release" branch is `main`? Off course I add "tags: true" and add latest release tag.

braver:
If you have tags the branch pointer becomes useless, so you can’t have both.

PR #8432, opened 2021-12-10
Added Algo2-TADs package

Sponja-:
Adds support for a spanish dialect of [Abstract Data Type](https://en.wikipedia.org/wiki/Abstract_data_type) specification, used by students of the COMP930004 class in the University of Buenos Aires. This includes syntax highlighting and some snippets, as seen in the [GitHub repository](https://github.com/Sponja-/Algo2-Sintaxis-TAD-st3).

I'm not sure if non-english packages can be submitted to the default channel. If this isn't the case, I'd appreciate any recommendations on other distribution methods.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -582,7 +582,7 @@
 			"labels": ["language syntax"],
 			"releases": [
 				{
-					"sublime_text": ">=3000",
+					"sublime_text": ">=3092",
 					"tags": true
 				}
 			]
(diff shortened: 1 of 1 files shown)


braver:
Why do you keep the legacy syntax around? New users won't need it, so maybe remove it. 
Please try to use `.gitattributes` to remove images and test files from the package (ie. everything end users don't need).
Non-English is fine, just severely limits your audience.

Sponja-:
Thanks for the feedback!
I added a `.gitattributes` file and updated the release version dependency.

PR #8436, opened 2021-12-18
Add Dictionary

futureprogrammer360:

My package is [Dictionary][4], a plugin that shows definitions of words in popups.

My package is similar to [MacDictionary][5]. However, it should still be added because [Dictionary][4] is not OS-dependent and is not limited to just macOS.

[2]: https://semver.org
[4]: https://github.com/futureprogrammer360/Dictionary
[5]: https://packagecontrol.io/packages/MacDictionary

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -808,6 +808,17 @@
 			]
 		},
 		{
+			"name": "Dictionary",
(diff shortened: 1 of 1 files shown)


braver:
Thanks!

PR #8438, opened 2021-12-19
Add Linguist Syntax

hronro:

My package is a bunch of syntaxes.

[2]: https://semver.org

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -1083,6 +1083,17 @@
 			]
 		},
 		{
+			"name": "Linguist Syntax",
+			"details": "https://github.com/hronro/sublime-linguist-syntax",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "LinkerScript",
 			"details": "https://github.com/jbw3/SublimeTextLinkerSyntax",
 			"labels": ["language syntax"],


braver:
We don't generally accept bundled packages like this. Usually it's a selection made once, mostly randomly, and then never maintained. Why should we accept your package? Why does your bundle make sense and how are you going to improve all those syntaxes over time?

hronro:
@braver 
I don't maintain those syntaxes, them comes from [github/linguist](https://github.com/github/linguist) and are maintained by GitHub.com (so they likely will be well-maintained).

However those syntax can not be used in Sublime Text directly, so in my repo there's a script to convert them to Sublime compatible syntaxes. I also plan to add a CI, which can automatically update my repo when there's a new release from github/linguist.

hronro:
@braver 
Please let me know if this is not acceptable in package control.

braver:
I guess I don’t understand why someone would use this in Sublime Text? The source syntaxes aren’t even supposed to be used in an editor, and there is no reason to assume they’re compliant with Sublime Text scope guidelines. So it’s a quick way to get loads of syntaxes, but I don’t see the use case.

hronro:
@braver 

The languages bundled in Sublime Text (github.com/sublimehq/Packages), are of very high quality.

But as a result, Sublime Text will only bundle a very limited number of languages.

I strongly agree with this, it avoids Sublime Text bundling low quality languages.

But I also do need more language support. And those language supports don't need to be as high quality as the ones bundled in Sublime Text, I just need their syntax highlighting to work in most of time.

Instead of going to Package Control and searching for and installing every language I encounter that Sublime Text doesn't support, Linguist Syntax provides out-of-the-box syntax highlighting for most languages, and those syntax highlighting should be fairly accurate, since their syntax definitions come from GitHub.com. I haven't tested them all, but I think most of them should work well.

I think Linguist Syntax provides an option for users. If the user wants each of his syntax definitions to be accurate and high quality, he can go to Package Control and search for and install a package for each language he wants to add. (This isn't always accurate, and I've seen several niche languages in package control that are not maintained any more and the syntax definition is out of date.)

braver:
Alright, we can simply add this to package control and if people find it useful that’s all good. However, you do need to pass the tests first.

hronro:
@braver 

Could you please help check if this is an issue of the tests themselves?

```
- ERROR: Invalid Plist
    - File: syntaxes/Julia.tmLanguage
    - Exception: not well-formed (invalid token): line 203, column 297
```

I opened `syntaxes/Julia.tmLanguage`, and  line 203 column 297 is an special character `゛`. (I guess it might because the test doesn't recognize those special characters?)

I load `syntaxes/Julia.tmLanguage` in Sublime Text 4 4121, and it works petty well.

Untitled

And that special charactor is also in the original Julia syntax highlighting:

https://github.com/JuliaEditorSupport/atom-language-julia/blob/8faf348415476d471187df2522cfda86982ce8e3/grammars/julia.json#L164

braver:
I can’t help you there. It might just be a limitation of the plist format. Converting to the newer sublime-syntax (yaml) format might help.

hronro:
@braver If it's a limitation of plist format, I guess the `Julia.tmLanguage` should be invalid for Sublime Text. However, as you can see in the screenshot I post above, `Julia.tmLanguage` actually works pretty well in Sublime Text 4121, which means it is valid.

hronro:
The result I got from searching Internet, it's plist(XML) support use unicode directly as long as they are encoded in UTF-8. So it seems like this is a limitation of the test, not a limitation of plist.

braver:
Alright, 🤷🏻 whatever then. Let's 🚀

PR #8439, opened 2021-12-19
Add Dockerfile syntax highlighting

snxx-lppxx:

Syntax highlighting did not work for me, and I decided to write my own.

[2]: https://semver.org

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -1292,6 +1292,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


braver:
We already have such a package. Please consider contributing improvements to that package, instead of adding a new package. If that doesn't work out for some reason, you can always come back here, and we can discuss options.

PR #8441, opened 2021-12-23
Replace Sublime-Gitignore with updated fork (again)

vilhelmen:

This is a gentle takeover of @kevinxucs's Gitignore package (if they don't mind, hi :wave:), which hasn't been updated in four years. See #1433 for their takeover of the original.

This update drops ST2 support, removes sublime-package support (overrides didn't work anyway), and (should) now automatically compile and push updates from github/gitignore quarterly via Actions. It packages the root gitignores (templates in common use) as well as the global gitignores (templates for various editors, tools, and operating systems).

[2]: https://semver.org

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -767,11 +767,11 @@
 		},
 		{
 			"name": "Gitignore",
-			"details": "https://github.com/kevinxucs/Sublime-Gitignore",
+			"details": "https://github.com/vilhelmen/Sublime-Gitignore",
 			"releases": [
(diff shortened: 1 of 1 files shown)


braver:
@kevinxucs you ok with this?

vilhelmen:
Will we be able to move forward if we don't hear back from the previous author? Or should I change the upload name?

braver:
No, let's do it, it's taken long enough now.

PR #8443, opened 2021-12-31
YamlToJson from ryandao and Power Query syntax from Kenneth Scrubbard

dannysummerlin:
- [x] I'm helping the package's author and/or maintainer.

My packages are two open source packages from other authors who have not updated or published in a long time.


dannysummerlin:
I have added a license file to the YamlToJson repo, but the Power Query repo already had one so I'm not sure what that warning is about

braver:
I'm not against publishing forks of existing work like this, but the lack of licenses explicitly allowing that in the original work complicates it. 

> but the Power Query repo already had one so I'm not sure what that warning is about

I don't see a license in the [original repo](https://github.com/PowerQueryforAccountants/PowerQuerySublimeSyntax)? 

I'm no lawyer, but seems the cleanest way to do this is to publish the repos ad verbatim, ie. without adding a license of your own. We don't require the license, it's just a recommendation to prevent confusing situations like this. So, you don't really need to add a license if it's not your own work.

dannysummerlin:
Ah, thanks for the suggestion! I just swapped in the original repos so that any licensing is clear from there, let me know if anything else needs tweaking!

braver:
Oh, sorry, that’s not quite what I meant. The entries here should point to your forks. That way they can’t suddenly break if they decide to move them or whatever. But you simply fork them without adding license files, and we publish them that way.

dannysummerlin:
Oh haha I'll get this right one day! I dropped the licenses, thanks for walking me through all this!

braver:
No problem! I'm only seeing the YamlToJson addition now, is that correct?

dannysummerlin:
Oh yeah, I guess so - I had missed that the Power Query bit was already in there, remembering things is overrated

braver:
Oh, you’re right. Well good 👍🏻

PR #8444, opened 2022-01-01
Add CompetitiveProgrammingParser Package

codeuniverse101:

My package is [CompetitiveProgrammingParser](https://github.com/codeuniverse101/CompetitiveProgrammingParser), a plugin that parses test-cases, problems and contess from various online judges.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3124,7 +3124,6 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"platforms": ["*"],
 					"tags": true
 				}
 			]
--- a/repository/c.json
+++ b/repository/c.json
@@ -1340,6 +1340,17 @@
 			]
 		},
 		{
+			"name": "clean-css",
+			"details": "https://github.com/1000ch/Sublime-clean-css",
+			"labels": ["css", "formatting"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "CleanCSS",
 			"details": "https://github.com/stolksdorf/CleanCSS",
 			"releases": [
@@ -1476,6 +1487,18 @@
 				{
 					"sublime_text": "*",
 					"branch": "master"
+				}
+			]
+		},
+		{
+			"name": "Clojure Sublimed",
+			"details": "https://github.com/tonsky/Clojure-Sublimed",
+			"labels": ["clojure", "repl", "nrepl"],
+			"donate": "https://www.patreon.com/tonsky",
+			"releases": [
+				{
+					"sublime_text": ">=4075",
+					"tags": true
 				}
 			]
 		},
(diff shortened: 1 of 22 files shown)


braver:
In those cases where you don't comply with the recommendations, can you explain why? E.g. why do you add key bindings and sidebar menu items?

braver:
Please respond to the feedback to continue the PR.

Taara-Sinh-Aatrey:
> * [ ]  My package doesn't add context menu entries. *
> * [ ]  My package doesn't add key bindings. **

**We won't be able to comply with the above two recommendations. In competitive programming, speed is an important factor and that's why it is very important for the user to have context menu entries and key bindings in order to do the task faster.**

----------------------------------------------------------------------------------------------------------------------------

> * [ ]  Any commands are available via the command palette.
> * [ ]  Preferences and keybindings (if any) are listed in the menu and the command palette, and open in split view.

**We are working on these two recommendations and will let you know once we complete it.**

codeuniverse101:
* [x]  I'm the package's author and/or maintainer.
* [x]  I have have read [the docs][1].
* [x]  I have tagged a release with a [semver][2] version number.
* [x]  My package repo has a description and a README describing what it's for and how to use it.

***

* [ ]  My package doesn't add context menu entries. *
* [ ]  My package doesn't add key bindings. **

We won't be able to comply with the above two recommendations. In competitive programming, speed is an important factor and that's why it is very important for the user to have context menu entries and key bindings in order to do the task faster.

***

* [x]  Any commands are available via the command palette.

All the commands except those invoked from the sidebar are now available via the command palette. Please let me know if it is
possible to make the sidebar commands availabe via the command palette too.

* [x]  Preferences and keybindings (if any) are listed in the menu and the command palette, and open in split view.

Settings (.sublime-settings), keybindings (.sublime-keymap) and Command palette (.sublime-commands) files can now be
opened from both menu as well as command palette.

This package is now ready from our side. Please review and merge.

codeuniverse101:
We had run the tests as mentioned [here](https://packagecontrol.io/docs/submitting_a_package#Step_7) after making the last set of changes and the test results were okay but I can see that some checks by the bot were not successful. 

Please let us know how we can deal with these unsuccessful checks.

repository/c.json
```diff
@@ -3118,6 +3118,17 @@
 				}
 			]
 		},
+		{
+			"name": "CompetitiveProgrammingParser",
+			"details": "https://github.com/codeuniverse101/CompetitiveProgrammingParser",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": ["*"],
```
braver, line 3127: If you support all platforms, please just remove the key here.

maybe a push on the PR also kicks the review-bot into action 🤞

braver:
1.0 isn't a valid semver tag, you need 3 digits.

jrappen:
Instead of adding default context menu entries and key bindings for your package and as a result for every end user, you should instead add documentation on how the end user makes use of the commands you have now added via context menus and/or key bindings while maybe adding the current defaults as suggestions to said documentation.

repository/c.json
codeuniverse101: Done

codeuniverse101:
I like your idea of not providing the keybindings and adding a documentation of adding new keybindings while also providing the default as suggestion because otherwise the default keybindings may conflict with the user's custom keybindings.

However, I don't agree with you on not providing the context menu entries. This package is for competitive programmers mainly and in competitive programming, speed is an important factor. The context menu entries is very important for the user in order to do the task faster which is why almost everyone would need them, not providing them will result in end user going through the documentation and adding the same entries themselves.

codeuniverse101:
Yeah, sorry about that. I have now updated the tag to v1.0.0

PR #8451, opened 2022-01-11
Merged "Xcode Build Configuration" and "Xcode Project".

peterthomashorn:
As suggested during initial publishing on Package Control and later on in a dedicated issue.

### DIFF
--- a/repository/x.json
+++ b/repository/x.json
@@ -45,8 +45,9 @@
 			]
 		},
 		{
-			"name": "Xcode Build Configuration",
-			"details": "https://github.com/peterthomashorn/xcode-build-configuration-language-for-sublime-text",
+			"name": "Xcode Build Configuration and Project Syntax Definitions",
+			"previous_names": ["Xcode Build Configuration", "Xcode Project"],
+			"details": "https://github.com/peterthomashorn/xcode-languages-for-sublime-text",
 			"labels": ["language syntax"],
 			"releases": [
 				{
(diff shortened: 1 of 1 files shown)


repository/x.json
```diff
@@ -45,8 +45,9 @@
 			]
 		},
 		{
-			"name": "Xcode Build Configuration",
-			"details": "https://github.com/peterthomashorn/xcode-build-configuration-language-for-sublime-text",
+			"name": "Xcode Build Configuration and Project Syntax Definitions",
```
braver, line 48: Lol, ok, bit long for a name but that works. "Xcode Builds and Projects" maybe?

jrappen:
> as suggested ...

Do you expect people to look up your prior discussions? Would have been easier to just provide the links to those as does everybody else across GitHub.

repository/x.json
peterthomashorn: Yeah, I know, but somehow I would like to keep it clear this is only about syntax definitions, not full support of such or anything else. 😅

peterthomashorn:
> > as suggested ...
> 
> Do you expect people to look up your prior discussions? Would have been easier to just provide the links to those as does everybody else across GitHub.

No. I had the tabs open for copying and pasting the links. I just simply forgot that because I was juggling with four repositories, some issues and pull requests simultaneously.

repository/x.json
peterthomashorn: In case an additional syntax is added I will definitely think about the title again to make it shorter.

PR #8452, opened 2022-01-11
Add SMPL Syntax plugin, syntax definition for the SMPL language

ofekih:

My package is a very simple syntax package for the SMPL programming language. It does not add a new style. I previously created a similar package for the LOLCODE programming language, although that package also includes snippets / code completions.

[2]: https://semver.org

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -2550,9 +2550,9 @@
 			]
 		},
 		{
-			"name": "SMPL Syntax",
+			"name": "SMPL",
 			"details": "https://github.com/ofekih/smpl-sublime",
-			"labels": ["smpl"],
+			"labels": ["smpl", "language syntax"],
 			"releases": [
 				{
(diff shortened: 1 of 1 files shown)


repository/s.json
```diff
@@ -2549,6 +2549,17 @@
 				}
 			]
 		},
+		{
+			"name": "SMPL Syntax",
+			"details": "https://github.com/ofekih/smpl-sublime",
+			"labels": ["smpl"],
```
braver, line 2555: I would recommend adding the "language syntax" label here, and removing "Syntax" from the package name: by convention language support packages are simply named after the language because they usually do more than just the syntax like snippets and even the syntax itself unlocks various other features like highlighting but also autocomplete and goto definition.

ofekih: thanks for the feedback, done :)

PR #8454, opened 2022-01-13
Add Sophia package

dincho:

My package is for [Sophia](https://github.com/aeternity/aesophia) contract programming language for aeternity blockchain.

[2]: https://semver.org

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -2829,7 +2829,7 @@
 			"labels": ["blockchain", "contracts", "aeternity"],
 			"releases": [
 				{
-					"sublime_text": ">=3000",
+					"sublime_text": ">=3092",
 					"tags": true
 				}
 			]
--- a/repository/s.json
+++ b/repository/s.json
@@ -2824,6 +2824,17 @@
 			]
 		},
 		{
+			"name": "Sophia",
+			"details": "https://github.com/aeternity/sublime-sophia",
+			"labels": ["blockchain", "contracts", "aeternity"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Sort JavaScript Imports",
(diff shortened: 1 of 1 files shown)


braver:
You should end all your scopes with something specific to your syntax, e.g. `.Sophia`, so that other packages and syntaxes can make adjustments for your particular syntax.

ElectricRCAircraftGuy:
@dincho, in case you don't know what @braver is talking about, all he means is this, I think:

In this file: https://github.com/aeternity/sublime-sophia/blob/master/Sophia.sublime-syntax

Add `.sophia` to *every* single `scope` entry. Ex:

Change FROM this:

```yaml
  brackets:
    - match: \]
      scope: punctuation.section.brackets.end
      pop: true
    - include: main

  braces:
    - match: \}
      scope: punctuation.section.braces.end
      pop: true
    - include: main
```

TO this:

```yaml
  brackets:
    - match: \]
      scope: punctuation.section.brackets.end.sophia
      pop: true
    - include: main

  braces:
    - match: \}
      scope: punctuation.section.braces.end.sophia
      pop: true
    - include: main
```

This doesn't break anything for various Color Schemes, because the first matching color scheme name will be used, meaning that if a particular color scheme doesn't have an entry for `punctuation.section.brackets.end.sophia`, the formatting one level up at `punctuation.section.brackets.end` will be used instead, so long as it exists. If it doesn't exist in the color scheme, it goes up another level, and so-on, until a matching name is found, and that formatting will be used.

Adding `.sophia` to the end of all `scope`s allows color scheme maintainers and users to add customisations for your particular package by simply adding entries for your `*.sophia` scopes.

ElectricRCAircraftGuy:
Good luck getting your package approved. I'll be submitting my own shortly.

dincho:
The request seem clear and fair, I just forgot adding it in a rush, also I'm delaying a bit to service it as I'm working on other stuff right now.

dincho:
@braver fixed, thank you

PR #8456, opened 2022-01-13
Replace `Scheme` with a new repo

absop:

My package is [Scheme](https://github.com/absop/ST-Scheme)

My package is similar to [Scheme](https://github.com/egrachev/sublime-scheme) However it should still be added because it is better (supports more features) and still being maintained.

[2]: https://semver.org

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -414,12 +414,13 @@
 		},
 		{
 			"name": "Scheme",
-			"details": "https://github.com/egrachev/sublime-scheme",
-			"labels": ["language syntax", "snippets"],
+			"details": "https://github.com/absop/ST-Scheme",
+			"labels": ["language syntax", "snippets", "formatting", "formatter", "code navigation", "build system", "lisp"],
+			"author": ["absop", "jfcherng"],
 			"releases": [
 				{
-					"sublime_text": "*",
-					"branch": "master"
+					"sublime_text": ">=4099",
+					"tags": true
 				}
 			]
 		},


braver:
In cases like these, if the original package isn't maintained (and you cannot be added as a maintainer, or transfer the repo, etc.), we can try to replace it. Would that work here (trying those other options first)?

absop:
> In cases like these, if the original package isn't maintained (and you cannot be added as a maintainer, or transfer the repo, etc.), we can try to replace it. Would that work here (trying those other options first)?

Thanks for reply. I am contacting the author of the original package to replace it.

egrachev:
Hello there! What I should do for you?

braver:
There are many ways to go about this. If you guys agree that Schemer is a good improvement on Scheme, maybe do a PR and consider adding @absop as a maintainer?

absop:
> There are many ways to go about this. If you guys agree that Schemer is a good improvement on Scheme, maybe do a PR and consider adding @absop as a maintainer?

Although they share the same feature of Syntax Highlighting, they are almost two completely different packages. I don't think do a PR is advisable. Maybe we should keep both of them and leave the choice to the users.

braver:
We usually try to avoid that. To the end user they serve an identical purpose. It is not clear why you would pick one over the other. If there is a very good reason to keep an 8 years old package available we can do that, but for most users it will be beneficial to get the newest and most complete support for the language they’re using.

absop:
If @egrachev agree, I would like to replace Scheme with Schemer. But let me give up [Schemer](https://github.com/absop/ST-Scheme) and to manage the [Scheme](https://github.com/egrachev/sublime-scheme), I can't do it.

egrachev:
> If @egrachev agree, I would like to replace Scheme with Schemer. But let me give up [Schemer](https://github.com/absop/ST-Scheme) and to manage the [Scheme](https://github.com/egrachev/sublime-scheme), I can't do it.

I agree

repository/s.json
```diff
@@ -414,12 +414,14 @@
 		},
 		{
 			"name": "Scheme",
-			"details": "https://github.com/egrachev/sublime-scheme",
-			"labels": ["language syntax", "snippets"],
+			"description": "Syntax Highlighting, smart snippets, code formatting, code navigation and more for the Scheme language",
```
braver, line 417: You don’t have to put a description here, package control will take the description of the GitHub repo, allowing to change it at will.

absop: Ok. Thank you for your remind.

braver:
Alright, final thing would be to tag a release with your latest changes.

absop:
> Alright, final thing would be to tag a release with your latest changes.

I've done it.
@braver

egrachev:
Good

PR #8457, opened 2022-01-13
Renamed `CodeCounter` to `CodeLines`

absop:

My package was originally **CodeCounter** and I've renamed it [CodeLines](https://github.com/absop/ST-CodeLines)

[2]: https://semver.org

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2038,13 +2038,13 @@
 			]
 		},
 		{
-			"name": "CodeCounter",
-			"details": "https://github.com/absop/CodeCounter",
+			"name": "CodeLines",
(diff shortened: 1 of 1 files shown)


repository/c.json
```diff
@@ -2204,6 +2191,18 @@
 				}
 			]
 		},
+		{
+			"name": "CodeLines",
+			"details": "https://github.com/absop/ST-CodeLines",
```
braver, line 2196: You need to add a "previous_names" key for CodeCounter here so that existing users are moved over to the new name of the package. Look in the repository for examples.

absop: Thanks for your reminding, I have done that.

PR #8458, opened 2022-01-13
Add Lumi to the repository

smadeja:

My package is a minimalistic, light colour scheme with constant luminance

[2]: https://semver.org

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -1905,6 +1905,16 @@
 			]
 		},
 		{
+			"name": "Lumi",
+			"details": "https://github.com/smadeja/lumi",
+			"releases": [
+				{
+					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


repository/l.json
```diff
@@ -1904,6 +1904,16 @@
 				}
 			]
 		},
+		{
+			"name": "Lumi",
+			"details": "https://github.com/smadeja/lumi",
```
braver, line 1909: Please add a labels key with "color scheme", to make it easier to understand what kind of package this is and make it easier to find.

braver (changes requested):

repository/l.json
smadeja: Absolutely, it's labelled now

PR #8461, opened 2022-01-17
Add a isorted package

rimvaliulin:

My package is [isorted](https://github.com/rimvaliulin/isorted)

My package is similar to [isort](https://github.com/thijsdezoete/sublime-text-isort-plugin). However it should still be added because it's seems like abandoned and use [fixed and old isort library v3.9.6](https://github.com/thijsdezoete/sublime-text-isort-plugin/blob/master/isort/__init__.py) (current isort version 5.10.1), which you can't easily update. My package works much more like [sublack](https://github.com/jgirardet/sublack) and use external [isort](https://github.com/PyCQA/isort) library. 

[2]: https://semver.org

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -1223,6 +1223,16 @@
 			]
 		},
 		{
+			"name": "Isorted",
+			"details": "https://github.com/rimvaliulin/isorted",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "ITDCHelper",
 			"details": "https://github.com/itdc/sublimetext-itdchelper",
 			"author": "Avtandil Kikabidze",
--- a/repository/i.json
+++ b/repository/i.json
@@ -1223,8 +1223,9 @@
 			]
 		},
 		{
-			"name": "Isorted",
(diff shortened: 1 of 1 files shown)


braver:
I can agree on adding this because it's a different (and cleaner) approach. However, make sure your command is listed in the command palette so people can find and use it easily.

rimvaliulin:
I renamed package, and added missing files due global .gitignore.

braver:
Alright, that’s better. I would like to propose removing the keybinding though. It’s a very generic combination that users or other packages might already be using for something else. At the very least it should be made conditional for python. But typically it’s a better idea to remove the binding itself and help users set up a binding if they want to. Your main menu entry should also have an entry for editing the keymap, and you could for instance comment out your binding. That way it’s presented as a suggestion and it makes it really easy for users to add it if they want to, or tweak the combination to their liking.

rimvaliulin:
I maked keybindings optional.

braver:
You need to tag a new release when you make changes. Only tags are picked up and distributed by package control.

rimvaliulin:
I tagged a new release, is there anything wrong?

PR #8462, opened 2022-01-21
Add Competitive Programming Lite

JameelKaisar:
- [ ] My package doesn't add context menu entries. `See Context Menu Below`
- [ ] My package doesn't add key bindings. `See Key Bindings Below`

About:
A simple, easy-to-use Sublime Text plugin to assist in Competitive Programming.

Repository:
https://github.com/JameelKaisar/Competitive-Programming-Lite

Context Menu:
The plugin has no entries in context menu in default mode. However in Competitive Programming mode, the plugin has two context menu entries: CP Test and CP Run. When user exits the Competitive Programming Mode, the menu entries are no longer available.

Key Bindings:
The plugin has four key bindings and these are necessary for creating and navigating through questions (files) efficiently. Competitive Programming usually has tight time constraints and having key bindings for most used functions is a reasonable thing to have. The key bindings are listed in the menu and the command palette and open in split view.

[2]: https://semver.org

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3142,6 +3142,17 @@
 			]
 		},
 		{
+			"name": "Competitive Programming Lite",
+			"details": "https://github.com/JameelKaisar/Competitive-Programming-Lite",
+			"labels": ["competitive programming", "competitive coding"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Compile Jasper Template",
 			"details": "https://github.com/MettleUp/CompileJasperTemplate",
 			"releases": [


JameelKaisar:
## Version 1.1.0 Released

### Changelogs
- Added Support for C
- Added Support for Java
- Minor Fixes

braver:
We already have some [competitive programming packages](https://packagecontrol.io/search/competitive), how does yours compare?

JameelKaisar:
## Existing Plugins

1. [Cpp​Fast​Olympic​Coding](https://packagecontrol.io/packages/CppFastOlympicCoding)
    - Excellent plugin for those who code in C++
    - Languages: C++ only
2. [Competitive Programming](https://packagecontrol.io/packages/Competitive%20Programming)
    - Basic plugin
    - Languages: C++ only
3. [CPP Competitive Programming Snippets](https://packagecontrol.io/packages/CPP%20Competitive%20Programming%20Snippets)
    - Snippets only
    - Languages: C++ only
4. [Get​Code](https://packagecontrol.io/packages/GetCode)
    - Problem parser
    - Platforms: Codechef, Codeforces, SPOJ, Virtual Judge

## Competitive Programming Lite
- Languages: C, C++, Java, Python
- Ability to test and compare program output with expected output with a single command.
- Ability to generate multiple template files with a single command.
- Ability to add /edit/delete custom templates to speed up coding.
- Easy to add support for more languages (see [C](https://github.com/JameelKaisar/Competitive-Programming-Lite/commit/378a6155eef666ef6b22b8b1364f23cc3fde5b6b), [Java](https://github.com/JameelKaisar/Competitive-Programming-Lite/commit/33f34f35100ce1ca8140d6700ec024678a5177cd))

braver:
Thanks. Can you maybe tell us why .no-sublime-package is needed? Just from browsing through the repo it seems everything should be doable without.

JameelKaisar:
There are couple of reasons why .no-sublime-package is needed:
1. [Adding Files] The user can add new templates. The templates are files stored in the plugin folder and are used as boilerplate codes to speed up coding.
2. [Renaming Files] There are two entries in the Context Menu (CP Test and CP Run) when user is in CP Mode. These options are not needed when user is not in CP mode. I couldn't find a way to get the current build system (see [Getting the current build system?](https://forum.sublimetext.com/t/getting-the-current-build-system/5886)) and decide whether Context Menu entries are relevant or not. The plugin renames the "Context.disabled-menu" file to "Context.sublime-menu" to enable the menu entries whenever user enters the CP Mode and vice versa.

braver:
> The user can add new templates. The templates are files stored in the plugin folder 

That’s not a good idea. Updates of the package can then conflict with user files and uninstalling the package also removed the user data. Think of your package as an app: you don’t store user data in the same place as your app code. In ST user data goes in the User package. You can put a directory in there and your package can read files from it too.

> and decide whether Context Menu entries are relevant or not

Do you even really really need the context menu? It’s usually not the best place to put functionality, especially if it isn’t tied to an exact click-location. If your commands are in the command palette most users are able to find it, access it quickly, even without touching the mouse.

Maybe you need to slightly compromise on your features, but like the user templates an un-packaged package is usually a problem waiting to happen.

JameelKaisar:
Both of your points are valid. I will make the necessary changes first then ping you back.

rwols:
> The templates are files stored in the plugin folder and are used as boilerplate codes to speed up coding.

How is this different from clicking on Tools > Developer > New Snippet ?

JameelKaisar:
Templates are different from Snippets because when we create a template, 4 to 5 additional files are created under `.cpl_files` folder. These include standard input/output files, the expexted output file and others files that help in comparing the output, etc. I prefer using snippets for the algorithms which I use the most while doing competitive programming. [CPP Competitive Programming Snippets](https://packagecontrol.io/packages/CPP%20Competitive%20Programming%20Snippets) plugin is based on the very same approach.

jrappen:
Instead of adding default context menu entries and key bindings for your package and as a result for every end user, you should instead add documentation on how the end user makes use of the commands you have now added via context menus and/or key bindings while maybe adding the current defaults as suggestions to said documentation.

jrappen:
Please do not override existing menu entries. Simply targetting by id should suffice.

JameelKaisar:
1. Yes. I will remove Context Menu entries as suggested by @braver.
2. Also I'll add documentation instead of adding default Key Bindings.
3. I'll make changes to target menu entries by id only.
4. No, the videos don't need to be shipped to end user. Videos do not appear in the release files. I have maked videos folder as `export-ignore` in `.gitattributes` file. Would that be enough or should I remove the videos folder from the repo also? @jrappen
5. I will make the necessary changes and then test the compatibilty with previous versions of Sublime Text.

braver:
Please report back after you make those changes and resolve the conflict of this branch with master.

JameelKaisar:
## Version 1.2.0 Released

### Changelogs
- Change from 'No Sublime Package' to 'Sublime Package'
- Remove Context Menu Entries
- Remove Default Key Bindings
- Add Default Template for Key Bindings
- Remove Help Videos from Repository
- Target Menu Entries By ID Only

JameelKaisar:
@braver I made the required changes and also resolved the conflict of this branch with master. You can merge this PR now.

PR #8463, opened 2022-01-21
Add XDCHighlighter

LDprg:

My package is a Highlighter for Vivados XDC files which are used for VHDL pinning in Xilinx FPGAs.

[2]: https://semver.org

### DIFF
--- a/repository/x.json
+++ b/repository/x.json
@@ -68,6 +68,17 @@
 			]
 		},
 		{
+			"name": "XDCHighlighter",
+			"details": "https://github.com/LDprg/XDCHighlighter",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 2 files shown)


jrappen:
Why not target `"sublime_text": ">=4107"`, ST3 is already dead?

braver:
The syntax is trivial, so if it works on ST3 it works 🤷🏼‍♂️ You can always come back and make it st4 only.

repository/x.json
```diff
@@ -67,6 +67,17 @@
 				}
 			]
 		},
+		{
+			"name": "XDCHighlighter",
```
braver, line 71: Syntax packages do more than simply highlighting, they enable various other features as well. Hence, language support packages should simply be named after the language. E.g. “css” and not “css syntax” or “css highlighting “.

braver (changes requested):

repository/x.json
LDprg: Oh. I will rename the package.

PR #8470, opened 2022-01-31
Added SIFTKeypointSyntaxHighlight Package

lchauvin:

My package is a syntax highlight for 3D SIFT-Rank Keypoint file format

[2]: https://semver.org

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1710,6 +1710,18 @@
 			]
 		},
 		{
+			"name": "SIFTKeypointSyntaxHighlight",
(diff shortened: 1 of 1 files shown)


braver:
`example-c` looks like a really odd scope, you sure that's correct? 

You can have spaces in the name of your package for readability. Also, language packages should just be named after the language they support, because the syntax enables many other features besides the syntax and highlighting itself. So maybe "SIFT Keypoint"?

braver:
Closed for inactivity, if you want to try again just reply here.

PR #8473, opened 2022-02-03
add more labels for Ethereum package

davidhq:
Hi ! Does adding these labels 

`"labels": ["language syntax", "solidity", "vyper", "smart contracts", "blockchain"],`

influence search terms?

It would be great if people could find this package through search for "solidity".

I get quite some issues in regard to this and people also install `EthereumSoliditySnippets` package by mistake at first which does not include syntax for Soldity smart contract programming language.

My package is ...

[2]: https://semver.org


braver:
Yes, I believe labels are also considered when searching.

PR #8474, opened 2022-02-07
Add Meson Build package

colinkiama:

My package is Meson Build. It adds Meson build system commands to Sublime Text and adds syntax highlighting to Meson configuration files.

My package is similar to Meson (the package) however it should still be added because the Meson package repo is missing and it was only concerned with the configuration files. The goal of the Meson Build package to both enhance the configuration file experience and allow users to interact with Meson through Sublime Text.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -1350,7 +1350,7 @@
 			]
 		},
 		{
-			"name": "Meson Build",
+			"name": "Meson",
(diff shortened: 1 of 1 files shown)


braver:
Looks like we should replace the existing (but missing) Meson package with yours. Essentially just rename yours to "Meson" and we're good to go.

colinkiama:
Thanks, just made the changes now.

PR #8475, opened 2022-02-08
Add Pretty Lua

aerobounce:

This package is Lua formatter which wraps [StyLua](https://github.com/JohnnyMorganz/StyLua) and with some additional features.
There are some Lua formatter in Package Control though, there is no packages that uses StyLua.

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -2204,6 +2204,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


braver:
Maybe mention "stylua" in your repo's description to help users distinguish your package from the alternatives.

PR #8478, opened 2022-02-14
Add Hacklang syntax

albertxing:

My package is a language syntax for [Hacklang](https://hacklang.org/)

[2]: https://semver.org

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -46,6 +46,17 @@
 			]
 		},
 		{
+			"name": "Hacklang Syntax",
+			"details": "https://github.com/albertxing/sublime-hacklang",
+			"labels": ["language syntax", "language", "hacklang"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Hacklang Typechecker and Autocompletion",
 			"details": "https://github.com/SiebelsTim/hack-sublime",
 			"labels": ["language syntax", "linting", "auto-complete", "language", "hacklang"],


deathaxe:
Looks like a pretty dated TextMate syntax which fails on pretty much every test case from ST's default PHP. My understanding is Hack being an extension to PHP so should highlight with most of its syntax constructs, no? How to ensure this syntax satisfies least highlighting qualites without any syntax tests being present?

albertxing:
Thanks for your feedback. I'm not familiar with syntax tests — what failures are you referring to?
You're correct that Hacklang is for the most part backwards compatible with PHP syntax. I manually compared Hacklang to PHP syntax highlighting for a couple of random open-source PHP files (e.g. https://raw.githubusercontent.com/googleapis/google-api-php-client/main/src/Utils/UriTemplate.php) and the results are comparable.

The syntax is directly derived from the canonical VSCode Hacklang package: https://github.com/slackhq/vscode-hack/blob/master/syntaxes/hack.json

deathaxe:
There are two major concerns.

1. The `*.tmLanguage` format used by VSCode uses some more recent TextMate features, which are not supported by Sublime Text. Most of them may result in compile errors, but that's not for sure in all circumstances. Some of them may cause "runtime" errors.
2. The `*.tmLanguage` format has numerous limitations compared to `*.sublime-syntax` format.

When trying to use Hack.tmLanguage to highlight https://github.com/sublimehq/Packages/blob/master/PHP/tests/syntax_test_php.php many highlighting issues can be observed. Without dedicated syntax tests for Hack it is hard to tell, whether those are expected due to language differences or whether it indicates other kinds of syntax definition related issues.

braver:
I've merged to ship the package. You can continue to do quality improvements and review of the package itself though.

I forgot to check the package name. The convention is to name language support packages without "syntax" or "highlighting" because language packages also enable completions, goto definition, and many more features. Therefor I quickly changed "Hacklang Syntax" to simply "Hacklang" after merging this PR. Hope that's ok ok with you, otherwise let me know here or in a new PR.

PR #8479, opened 2022-02-16
Limit Microsoft Typescript package to ST2 and ST3

deathaxe:
This package is unmaintained and outdated.

A better syntax definition is already shipped with ST4.
Intellisense features are provided by LSP and LSP-typescript.

So lets limit the package's availability to ST2 and ST3.

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -3567,7 +3567,7 @@
 			"labels": ["auto-complete", "code navigation", "formatting", "language syntax", "snippets", "typescript", "javascript"],
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": "<4000",
 					"tags": true
 				}
 			]


rchl:
I agree with that change but I think it might be a good idea to give them a heads up in the repo, stating the reasons also there. I might do that unless someone else volunteers.

rchl (approved):

braver:
I agree but I’d really like to give them some time to respond here before we merge this.

rchl:
At the same time, no response in a long time is pretty much equivalent to admitting that they don't support it anymore.

I would also mention that "orta", who was working on typescript in MS until recently gave it thumbs up. Thumb since he's no longer at MS, we probably can't take it as an official statement. :)

braver:
> At the same time, no response in a long time is pretty much equivalent to admitting that they don't support it anymore.

totally agree

PR #8481, opened 2022-02-18
add NotePad

va9iff:

My package is ...

[2]: https://semver.org

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -896,6 +896,17 @@
 			]
 		},
 		{
+			"name": "NotePad",
+			"details": "https://github.com/va9iff/SublimeNotePad",
+			"labels": ["productivity"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Notepad++ Color Scheme",
 			"details": "https://github.com/evandrocoan/NotepadPlusPlusColorScheme",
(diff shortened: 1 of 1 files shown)


braver:
How does this compare to the built-in "distraction free mode", or these options?
https://packagecontrol.io/packages/DistractionFreeWindow
https://packagecontrol.io/packages/distractionless

va9iff:
You can open a new NotePad view with 1 command and start writing here. It is window specific, so you can have separate post-it and work however you want in your current workspace.

But the biggest pro of NotePad over distractionless is that it doesn't have any listeners. I believe the packages should have as less listeners and background jobs as possible in sublime text. Cuz the speed and ligthweightness is the advantage of st for me.
I even have "copied" some packages to work by only calling 3rd party tools cuz dedicated packages were a bit bloated.

So, that's the biggest hit of NotePad to distractionless. But distraction Free Window looks very similar :) 
But in the gif It didn't toggle the gutter. But NotePad does.
Anyways, I love Sublime Text. Here's my contribution. If isn't useful, no problem :')

But NotePad is logically convenient like post-it of notepad++
I couldn't find a package to turn st to notepad. Probably someone else will too

braver:
Alright, thanks. You skipped some stuff in the checklist like tagging a release. Please double check the requirements for submitting a package.

va9iff:
I've done the tagging a release. If there is something else I'm missing, please let me know.

PR #8482, opened 2022-02-18
added Mystify

M49ICKPIxi3:

My package is ...

[2]: https://semver.org


jrappen:
Have you considered using [`sublime_plugin.ListInputHandler`](https://www.sublimetext.com/docs/api_reference.html#sublime_plugin.ListInputHandler:ver-dev) instead of context menu entries?

PR #8483, opened 2022-02-19
Add FormatX

cj1128:

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -1569,6 +1569,17 @@
 			]
 		},
 		{
+			"name": "FormatX",
+			"details": "https://github.com/cj1128/sublime-formatx",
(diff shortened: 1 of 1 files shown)


cj1128:
Can you give some links? I don't find any formatter like FormatX in package control.

FormatX is formatter agnostic, the main usage is that the user configures the shell command, FormatX runs the command with the file as stdio and uses the stdout to replace the file. It doesn't bundle any formatters.

braver:
That sounds a lot like Fmt:
https://packagecontrol.io/packages/Fmt

jrappen:
Please do not override captions for "Preferences" and "Package Settings" in your menu file.

Simply targetting by id should suffice.

cj1128:
I think `Fmt` solves my problem so there is no need for `FormatX` now.

PR #8484, opened 2022-02-21
adding Exp language support

NemsLucky:

My package is ...

[2]: https://semver.org

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -1760,6 +1760,16 @@
 			]
 		},
 		{
+			"name": "Exp language support",
+			"details": "https://github.com/dev-univers/Exp-language-support",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Extend Selection",
 			"details": "https://github.com/anly2/sublime-extend-selection",
 			"author": "aanchev",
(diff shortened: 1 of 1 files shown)


braver:
Please actually fill in the template and tell us a bit about your package.

jrappen:
Please add git tags to the git repo as required (and configured).

---

The `*.sublime-syntax` files are not supported by the syntax engine of all versions of Sublime Text.

---

Your syntax file is missing some tags:

<https://www.sublimetext.com/docs/syntax.html#ver-dev>

---

You might want to add continuous integration tests. Compare:

* <https://www.sublimetext.com/docs/syntax.html#testing:ver-dev>
* <https://github.com/sublimehq/Packages/blob/master/.github/workflows/ci.yml>
    * be aware Packages are in subdirectories there
    * for root directory testing also compare: <https://github.com/braver/SublimeSass/blob/master/.github/workflows/continuous-integration.yml>

Currently you can test:

* indentation
* symbols
* syntax

The best (short) overview is currently in the still pending pull request to the default Packages for SQL:

<https://github.com/sublimehq/Packages/pull/3046/files>

where you should have a look at the files in `./SQL/tests/`. Click the conversation tab at the top of the page for some explanations within the conversation with regards to testing.

PR #8487, opened 2022-02-26
SSH-Panel Package

Haiquan-27:
add SSH-Panel

My package Package functions:

1. View, Add, Edit files on remote SSH host
2. View remote file detailed properties
3. Browse with HTML panel like sidebar
4. Verify remote host fingerprint
5. Support password and public-private key pair authentication methods

The function is similar to the *RemoteTree*

difference:
* support on the Linux and Windows
* support all available openssh server operating systems
* the details of the server can be displayed

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -3524,6 +3524,21 @@
 			]
 		},
 		{
+			"name": "SSH-Panel",
+			"details": "https://github.com/Haiquan-27/SSH-Panel",
+			"author": "Haiquan",
+			"issues": "https://github.com/Haiquan-27/SSH-Panel/issues",
+			"readme": "https://raw.githubusercontent.com/Haiquan-27/SSH-Panel/main/README.md",
+			"labels": ["file navigation", "file creation", "remote"],
+			"releases": [
+				{
+					"sublime_text": ">=3211",
+					"platforms": ["windows", "linux"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"details": "https://github.com/bmc/ST2SyntaxFromFileName",
 			"releases": [
 				{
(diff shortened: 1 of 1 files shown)


Haiquan-27:
Should I wait for the automatic test?

repository/s.json
```diff
@@ -3523,6 +3523,21 @@
 				}
 			]
 		},
+		{
+			"name": "SSH-Panel",
+			"details": "https://github.com/Haiquan-27/SSH-Panel",
+			"author": "Haiquan",
+			"issues": "https://github.com/Haiquan-27/SSH-Panel/issues",
+			"readme": "https://raw.githubusercontent.com/Haiquan-27/SSH-Panel/main/README.md",
```
braver, line 3531: You don't need "issues" or "readme" here since they're just the defaults for GitHub.

braver:
- Can your dependency be handled through the dependencies feature of package control, or maybe by the user installing it on their machine?
- Would be nice to be able to access the settings though the main menu.

Haiquan-27:
I don't support Mac because I haven't found a way to compile library files on Mac. In fact, as long as I can compile library files, I can support Mac platform. I haven't solved this problem recently, and it may be solved in the next version.

Can I accept my current submission?

braver:
Sure, sounds reasonable. Please have a look at the other feedback too.

repository/s.json
Haiquan-27: i have deleted " issues" and "readme"

braver:
> Can your dependency (on paramiko) be handled through the dependencies feature of package control, or maybe by the user installing it on their machine?

🤔❔

Haiquan-27:
Dependencies can be accessed by users from [here](https://github.com/Haiquan-27/SSH-Panel-doc-annex) Self download, which provides the dependent files required by the platform supported by the package

In addition, because Windows requires additional python DLL file, so dependencies feature of package control is not applicable

PR #8495, opened 2022-03-13
Add AnyTest package

timfjord:

The main idea of this package is to automatically detect a test framework for the given file and run it.
Basically it is a Sublime Text interpretation of the [vim-test](https://github.com/vim-test/vim-test) Vim plugin.
Currently it supports only 7 test frameworks(`espec`, `exunit`, `jest`, `pytest`, `pyunit`, `minitest`, `rspec`)  but more test frameworks are coming very soon.

There might be some packages for individual test frameworks(like [TestRSpec](https://packagecontrol.io/packages/TestRSpec), etc) but the idea of `AnyTest` is to handle as many test framworks as possible in one pacakge.

[2]: https://semver.org

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -1258,6 +1258,19 @@
 			]
 		},
 		{
+			"name": "AnyTest",
+			"details": "https://github.com/timfjord/AnyTest",
+			"labels": [
+				"jest", "rspec", "pytest", "testing", "exunit", "minitest", "espec", "pyunit"
+			],
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


braver:
Looks interesting! Just because I feel like I need to nitpick on something, I'd like to see a setting that disables the context menu entries. Maybe that's just a pet peeve, but that area does get crowded. Or maybe remove the separator?

braver:
Cool 👌🏻

PR #8502, opened 2022-03-25
Add package `klog`

jotaen:

My package provides support for working with klog files in Sublime Text. For context: [klog is an open-source file format for time tracking in plain text files](https://klog.jotaen.net/).

This package is mainly for providing syntax highlighting, but it also comes with a few sensible settings, e.g. regarding indentation.

Since klog is a data format, the package provides specific colour definitions (overrides) for the Mariana and Monokai theme. The syntax highlighting is not arbitrary, but it is supposed to illustrate the meaning of the data: for example, a negative time duration should be red (to indicate that it is subtractive), whereas a positive time duration should be green (to indicate that it is additive).

### DIFF
--- a/repository/k.json
+++ b/repository/k.json
@@ -310,6 +310,16 @@
 			]
 		},
 		{
+			"name": "klog",
+			"details": "https://github.com/jotaen/klog-sublime",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "knitr",
 			"details": "https://github.com/andrewheiss/SublimeKnitr",
 			"labels": ["knitr", "latex", "r", "markdown"],
--- a/repository/k.json
+++ b/repository/k.json
@@ -314,7 +314,7 @@
 			"details": "https://github.com/jotaen/klog-sublime",
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": ">=3092",
 					"tags": true
 				}
 			]


braver:
- your scopes should end in something that specifically identifies your syntax:
  - `constant.numeric` ➡️ `constant.numeric.klog`
  - `meta.klog.duration.negative` ➡️ `meta.duration.negative.klog`
- please double check if you're matching the `https://www.sublimetext.com/docs/scope_naming.html`
- your overrides only work for one color scheme, what about the hundreds of different color schemes people use? it's probably better to investigate scopes that convey the meaning you're looking for, maybe even using the color scopes ... have a look at [this discussion](https://github.com/wbond/package_control_channel/pull/8344#issuecomment-907390201)

braver:
Yeah, ok, colorish scopes are no good. 🤔 

> So I’m actually not sure what the right way forward would be?

It’s worth the effort to try really really hard not to have to ship color schemes with syntaxes. There are so many downsides to it… so let’s try to think this one through.

Have you had a look at scopes used for diffs? 

Or, if you do keep the override, can you use existing variables instead of introducing new colors? That way overrides users made themselves to those variables should also apply to yours (I know of numerous cases where users tweak a color scheme for contrast).

jotaen:
> Have you had a look at scopes used for diffs?

Ah cool, I haven’t thought of those.

I’ve updated the repo with a version *without* included colour overrides. I have tested it with the default colour schemes, and I think it would be viable like that, although not optimal. (But my original approach wasn’t optimal either, so there is that.) In Mariana/Monokai it’s pretty close actually, for Celeste it doesn’t work out so well.

Screenshot 2022-04-01 at 10 00 39

Screenshot 2022-04-01 at 10 00 26

[I’ve added a section to the README](https://github.com/jotaen/klog-sublime#colour-overrides), in which I explain how users can customise specific colour scopes. For that purpose, I assigned additional, dedicated scopes everywhere (`markup.other.____.klog`), to make the customisation process more convenient for users. These scopes are not consumed by the package itself, however.

braver:
Sounds like the right approach. That way color scheme authors and users can make any adjustments they like to improve what your syntax looks like. At least that’s how the system is set up to work. Realistically though, the default color schemes will never be updated, so maybe well defined overrides there (using variables, not hardcoded color) can be acceptable.

jotaen:
Awesome, thanks for helping me work this out. I also think the current solution should work well enough for users.

PR #8503, opened 2022-03-26
New Package : Remove Workspace

SmartManoj:

My package is  Remove Workspace

[2]: https://semver.org

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -1213,6 +1213,16 @@
 			]
 		},
 		{
+			"name": "Remove Workspace",
+			"details": "https://github.com/SmartManoj/RemoveWorkspace",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Rename Buffer",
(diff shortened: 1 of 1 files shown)


SmartManoj:
Solves https://stackoverflow.com/questions/11744365/removing-projects-in-sublime-text-2-and-3

rwols:
I believe this is builtin in ST4

Ultra-Instinct-05:
Yes. For ST4, you can just use `Project -> Open Recent -> Remove Deleted` to get rid of deleted projects or workspaces (FWIW SM also has the same ability to clear dead repositories using (`File -> Open Recent -> Clear Missing`).

For ST3, there was this earlier https://github.com/STealthy-and-haSTy/SublimeScraps/tree/master/session_cleaner. So the plugin could be useful for ST3, but given that a decent number of stable ST4 releases are present, I wouldn't use ST3 and there is no need of this plugin in ST4.

SmartManoj:
Just a note, the difference from that package is it automatically closes the app and the workspaces can be selected.

PR #8504, opened 2022-03-26
Add Sublime REST Client project

yeraydiazdiaz:

My package allows you to define and send HTTP requests directly from Sublime Text 4.

My package is similar to RESTer HTTP Client. However it should still be added because its development seems
to have stopped several years ago and thus does not target Sublime Text 4. Its code also uses the standard library for most of the HTTP request heavy lifting which seemed unnecessary and a potential security issue when `urllib3`
exists.

[2]: https://semver.org

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -1422,6 +1422,16 @@
 			]
 		},
 		{
+			"name": "REST Client",
+			"details": "https://github.com/yeraydiazdiaz/sublime-rest-client",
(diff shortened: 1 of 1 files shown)


yeraydiazdiaz:
I've updated the repo and requested a review from the bot, but maybe it needs a bump?

yeraydiazdiaz:
Thanks! 🎉

PR #8505, opened 2022-03-27
Changing Fennel package branch

gbaptista:

I'm just changing the default branch for the `Fennel` package from `master` to `main`.


gbaptista:
Oh, I use tags, don't need a branch. Nevermind.

PR #8508, opened 2022-03-29
Pull requests for "Pieces for Sublime" plugin, a plugin that saves code snippets and texts from Sublime text editor

moon-bui-pieces:

My package is ...

[2]: https://semver.org

### DIFF
--- a/channel.json
+++ b/channel.json
@@ -2,6 +2,7 @@
 	"schema_version": "3.0.0",
 	"repositories": [
 		"./repository.json",
+		"https://bitbucket.org/pi3c3s/plugin_sublime/src/host_test_branch/packages.json",		
 		"https://donkirkby.github.io/live-py-plugin/sublime-package/package.json",
 		"https://emmetio.github.io/sublime-text-plugin/registry.json",
 		"https://eoa-cdn.s3.amazonaws.com/extensions/packages.json",
--- a/channel.json
+++ b/channel.json
@@ -2,7 +2,6 @@
 	"schema_version": "3.0.0",
 	"repositories": [
 		"./repository.json",
-		"https://bitbucket.org/pi3c3s/plugin_sublime/src/host_test_branch/packages.json",		
 		"https://donkirkby.github.io/live-py-plugin/sublime-package/package.json",
 		"https://emmetio.github.io/sublime-text-plugin/registry.json",
 		"https://eoa-cdn.s3.amazonaws.com/extensions/packages.json",
@@ -125,6 +124,7 @@
 		"https://raw.githubusercontent.com/xgenvn/InputHelper/master/packages.json",
 		"https://raw.githubusercontent.com/zfkun/sublime-kissy-snippets/master/packages.json",
 		"https://s3.amazonaws.com/wallaby-downloads/sublime/packages.json",
+		"https://storage.googleapis.com/app-releases-59612ba/plugin_sublime/release/packages.json",
 		"https://sublimerge.com/packages.json",
 		"https://tinkerwell.app/plugins/sublime/packages.json"
 	]
--- a/repository/p.json
+++ b/repository/p.json
@@ -1518,17 +1518,6 @@
 			]
 		},
 		{
-			"name": "Pieces",
-			"details": "https://bitbucket.org/cristion-pieces/plugin_sublime",
-			"labels": ["pieces", "snippets"],
-			"releases": [
-				{
-					"sublime_text": "*",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "Pine",
 			"details": "https://github.com/gerardroche/sublime-pine",
 			"labels": ["language syntax"],


moon-bui-pieces:
My package is Pieces for Sublime

There are no packages like it in Package Control. It is a plugin that saves the user's code snippets and texts to an application, and the user can access their saved snippets from the context menu in Sublime text to insert code/text and increase their workflow productivity

moon-bui-pieces:
Hi Braver! Thank you for your response to our pull request. I am more than happy to answer your questions and shed some more light into our Sublime plugin specifically, and our whole software product in general.

To answer your first question:
  - We added a channel because we are choosing to manually host the plugin.
  - Following the guides from step 6 in the **Package Control Documentation** for submitting a package, we added the URL to “channel.json”.
  - As there was no mention of it in the documentation, I was not aware that you can add to the repository files for manual hosting, so it would be helpful if you can confirm on that matter!

As for the similarities to **SnippetMaker**, my understanding is that **SnippetMaker** allows you to save snippets via the **Command Palette**. After inputting the name and description for the snippet, it is saved as a file under the “User” folder. 

While **Pieces** also allows developers to save snippets, our product focus is quite different in a few ways: 
  - We want developers to stay in their flow, so our users can save a snippet from Sublime simply by highlighting the code and right-clicking to save a snippet.
  - We do a lot of ML magic on this snippet - snippets are auto-classified by language and auto-tagged with keywords.
  - Users can re-use their snippets by simply right-clicking on their list of snippets and pasting into their file.  Users can also download our desktop app for more functionality, like saving screenshots of code and extracting the code inside via OCR.
  - Our Sublime roadmap is quite ambitious - we want to launch automatic saving of snippets, code auto-complete and more.  All of these features are already live in our VS Code plugin by the way, which is a trending top plugin on their marketplace.
  - **Pieces** is available on IDEs such as **Visual Studio Code**, **IntelliJ** and **UltraEdit**. We also have extensions on **Chrome**, **Safari** and **Edge** for maximum workflow productivity. Just recently, we were featured [in a video with **FreeCodeCamp**](https://www.youtube.com/watch?v=g1vy03ZY5mM&t=1559s&ab_channel=freeCodeCamp.org) if you would like to check us out there!

I hope this was somewhat helpful in answering your questions. If you have any further concerns, please don’t hesitate to reach out to us at _moon@pieces.app_. We welcome any feedback and would love to hear more from you!

braver:
Oh, right. We already have a package for Pieces though: https://packagecontrol.io/packages/Pieces

moon-bui-pieces:
We are aware of that, however that plugin was developed by a previous intern and is made public on Bitbucket, which was not our intention. It does not work as it, and we are not sure how the procedure goes for deleting the plugin for situations like this. It would be great if you can help shed some light on the matter!

braver:
Alright, we can remove or replace the existing package. If we remove the existing entry and you introduce a new one with the same name everyone who has already installed the package will move to the new one seamlessly.

As per our guideline packages should not have "Sublime" or "For Sublime" in their name: this is the package repository for Sublime Text, all packages here are for Sublime. So rename your package to simply Pieces would be the best for everyone.

> I was not aware that you can add to the repository files for manual hosting

Not sure what you mean here? What kind of files?

moon-bui-pieces:
> Alright, we can remove or replace the existing package. If we remove the existing entry and you introduce a new one with the same name everyone who has already installed the package will move to the new one seamlessly.

This is a great solution! Thank you very much!

> * Why do you add a channel and not simply to one of the repository files?

I was referring to this, did I misunderstand your point?

rchl:
Doesn't look like your channel actually provides the package URL so it wouldn't work as is.

FichteFoll:
Your specified `repository.json` (`packages.json`) will not work, because

- `details` only support URLs to github, bitbucket, or gitlab repositories. If you want to manually host the package, you'll need to provide [a different set of keys](https://github.com/wbond/package_control/blob/master/example-repository.json#L340-L369) in the `releases` list. Furthermore, you'd want to at least specify a `homepage` url and optionally one for `issues`.
- The URLs currently specified in `detials` and `readme` don't appear to be working.
- Please adjust the name to "Pieces", as discussed with @braver before. Please also remove the current entry for "Pieces" from p.json in this PR.

moon-bui-pieces:
One more question, I am a little confused when referring to "repository.json". Where should I be adding the new key changes? p.json, repository.json or do I create my own.json?

rchl:
Your `messages.json` file is not a valid JSON file as it has trailing comma. Package Control reads that file using python's JSON module so it will not like that.

rchl:
I don't think you should be shipping `Package Control.sublime-settings` and `Package Control.user-ca-bundle` in your package.

moon-bui-pieces:
While I'm at it with fixing the package for push, I'd like to ask what the content encoding/mime type for the README should be, text/markdown or application/octet stream?

rchl:
> Understood! I have it in there because I wanted the "SubNotify" plugin to be installed with our plugin for notification purposes. Can you suggest a different way of doing this?

Read https://packagecontrol.io/docs/dependencies

moon-bui-pieces:
@braver @rchl @FichteFoll , I've updated the changes accordingly. You should be able to see it reflect from the channel link, please let me know if there are still any issues. Thank you so much for the help!!

rchl:
Can you instead just declare the dependencies inside your package using the `dependencies.json` file?

That's because I feel like you are not doing it correctly right now. I don't think you should be defining the dependency `SubNotify` in your channel file if it's a dependency that already exists and is available in package control. You should only be declaring dependency within your own package entry.

moon-bui-pieces:
> Can you instead just declare the dependencies inside your package using the `dependencies.json` file?

Just made the changes, it should be available now!

moon-bui-pieces:
request review @packagecontrol-bot

moon-bui-pieces:
Would you be able to take a look at the updated plugins fixes now and inform me if it works properly @braver @rchl @FichteFoll ? Thank you very much really appreciate it!!

moon-bui-pieces:
Hey I haven't heard back in a while and I was hoping you guys can give me a status update on the situation with the plugin. Is there anything I can do to help speed up the process or make it simpler somehow @braver  @rchl  @FichteFoll ?

jrappen:
No need to mention everyone again and again.

---

* The coding style across all files is all over the place and inconsistent.
* Why duplicate your channel file into the package you're shipping to end users?
* Why not host the code for your plugin and the channel file on GitHub or Gitlab like everyone else?
* Since you (target ST4+ and) use py3.8 specifically, why not make use of it in your Python code?
* `official_plugin` line 31 should be `r"^\s*$"` not `"^\s*$"`
* Your util method `find_platform` is redundant, use [`sublime.platform()`](https://www.sublimetext.com/docs/api_reference.html#sublime).

FichteFoll:
Sorry for the wait here, I don't frequently review packages anymore and was busy with other things.

Kind of like @rchl indicated but different: The thing you specified as a dependency, "SubNotify", is a Package and not a dependency (or "library", as they will be called in the future to combat this confusion), which will not work. The package will fail to install because a dependency with that name is not available. If you have a dependency on a different package, you should express that dependency in the readme and have safe-guards in code against it missing (and maybe prompt the user about installing it should it be a hard requirement). Additionally, I'd advise to only specify dependencies in either `dependencies.json` *or* `repository.json`, but that doesn't matter here anyway.

The following are optional suggestions for your code:

- When opening a file in write mode (i.e. with `w`), a `+` has no effect. The file is truncated as it is opened. Seeking to the beginning of the file before closing it in general and closing it explicitly when it is opened through a `with` block is also redundant. (util.py:49)
- Building the context menu file path and ensuring its parent folder's existance could be refactored into a reusable function.
- `os.makedirs` has an `exist_ok` parameter that makes it not fail when the directory already exists, making previous checks redundant.
- During deletion/initial load, instead of rewriting the file with a basic structure, then re-reading the file to add the menu items for the snippets, it would be simpler if you included the basic structure in your "write context menu" function
- I suggest employing a simple linter such as `flake8` for your code to clean up unused imports (and ensure a more consistent code style).
- Instead of using `View.erase` and later `View.insert`, you could use `View.replace`. (official_plugin.py:56-65)
- If you already know the contents of a snippet at the time you create the menu entry, you could also use the previously fetched `snapshot` for the text to insert instead of making another HTTP call. If the snippets are supposed to update frequently in the background, you should update the context menu snippet list periodically anyway (e.g. with a `sublime.set_timeout_async` chain).
- As the next step, providing a command palette command with an input handler or opening a quick panel with all snippets for the user to search through seems like a reasonably addition. See [here](https://docs.sublimetext.io/guide/extensibility/plugins/input_handlers.html) for documentation on the former and [here](https://www.sublimetext.com/docs/api_reference.html) (`show_quick_panel`) on the latter.
- When a user has multiple selections and opts to insert text, the usual behavior is to insert the text for all selections and not just the first. (InsertSavedPiecesCommand)

- Managing packages hosted on GitHub with the `"details"` feature is generally much simpler as you do not need to update the archive download and manually edit the repository.json. Instead, you would simply create and push a new tag for a new release. You aren't using any feature exclusive to manually-hosted repository.json files.

The following changes are **required** before we can merge this:

- In util.py:23 you replace the entire file for each for-loop iteration. You should do it only once after the loop has terminated.
- When using `urllib.request.*`, you need to `import urlib.request` as well, otherwise the code in HTTPClient.py working could just be coincidence (e.g. if the full module has been imported by some other plugin before).
- The resource that https://storage.googleapis.com/app-releases-59612ba/plugin_sublime/release/Pieces.sublime-package points to should not be changed because otherwise there could be a race condition between when a version update is registered by Package Control and someone installing the newest version of the archive marked as a previous version.

---

Well, I suppose lenghty reviews like this are why I eventually got burnt out on doing these regularly, but I hope you learned a thing or two about Python and the Sublime Text ecosystem.

moon-bui-pieces:
@FichteFoll thank you so much for the detailed response. This is very helpful. I will review and make the necessary updates accordingly.

braver:
Let us know when those changes are made, we can re-open this thread then, if needed.

PR #8509, opened 2022-03-29
Add dprint

dsherret:

My package is for [dprint](https://dprint.dev)

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -1564,6 +1564,16 @@
 			]
 		},
 		{
+			"name": "dprint",
+			"details": "https://github.com/dprint/dprint-sublime",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Dracula Color Scheme",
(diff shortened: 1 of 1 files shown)


braver:
- Would be nice to access settings via the main menu.
- Oh, on second look the settings are empty?
- Since it is slow-ish, maybe running on-pre-save should be optional?

dsherret:
Thanks. I'm actually not going to bother with this for now. I need to upgrade the interface and it would break the current implementation fo the sublime plugin and I don't have the bandwidth for it.

dsherret:
Actually, I just remembered that this is implemented by formatting via `dprint fmt --stdin` so I'm not breaking this interface.

@braver what settings are required? I don't think there are any.

> Since it is slow-ish, maybe running on-pre-save should be optional?

It's only slowish depending on the plugin. For the main default ones it should be pretty fast.

braver:
If that’s essentially all it does, can the same not be achieved via the Fmt package?

dsherret:
That looks like a better solution and then I won't have to maintain this. Thanks!

PR #8511, opened 2022-03-30
Add package `prismicio-snippets`

lihbr:

My package is a collection of snippets to template and work with [Prismic.io](https://prismic.io)

[2]: https://semver.org

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -2338,7 +2338,7 @@
 			]
 		},
 		{
-			"name": "Prismic.io snippet",
+			"name": "Prismic.io snippets",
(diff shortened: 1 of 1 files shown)


braver:
Would be nice to just ship the Sublime Snippets and not everything else as well. Also, maybe the package name should be "snippets" plural?

lihbr:
> Would be nice to just ship the Sublime Snippets and not everything else as well.
Other IDE might have their requirements regarding what gets shipped but I'll make sure to shave off as much as possible. Do you mean the dotfiles also? Can add them if needed, let me know ☺️

> Also, maybe the package name should be "snippets" plural?
For sure! Our bad on that one.

PR #8513, opened 2022-04-01
Add package `xit!`

jotaen:

My package is for working with [`[x]it!` files, which is a format for todos and check lists](https://xit.jotaen.net/).

It provides:

- Syntax highlighting
- Commands for toggling the status of a todo item (when the cursor is placed in the item scope)
  - The command auto-saves the file by default, but that can be turned off via the `xit_auto_save` setting.
- “Smart” completions for due dates, where typing e.g. `3d` completes to `-> 2022-04-04`, i.e. the date 3 days from now
- Some sensible default settings that are specific for the `.xit` file format

I’d love for the package name to be `[x]it!` rather than `xit!`, but I suppose that’s not possible, is it?

### DIFF
--- a/repository/x.json
+++ b/repository/x.json
@@ -100,6 +100,16 @@
 			]
 		},
 		{
+			"name": "xit!",
+			"details": "https://github.com/jotaen/xit-sublime",
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Xkcd",
 			"details": "https://github.com/EivindArvesen/xkcd",
 			"releases": [


jotaen:
> Repo link: [xit!](https://github.com/jotaen/xit-sublime) [Results help](https://github.com/packagecontrol/st_package_reviewer/wiki/Package-checks)
> 
> ```
> Packages added:
>   - xit!
> 
> Processing package "xit!"
>   - WARNING: The binding ['enter'] is also defined in default bindings but is masked with a 'context'
>     - File: Default.sublime-keymap
>   - WARNING: The binding ['enter'] is also defined in default bindings but is masked with a 'context'
>     - File: Default.sublime-keymap
>   - WARNING: The binding ['shift+enter'] is also defined in default bindings but is masked with a 'context'
>     - File: Default.sublime-keymap
>   - WARNING: The binding ['['] is also defined in default bindings but is masked with a 'context'
>     - File: Default.sublime-keymap
> ```

Regarding this warning, please let me know if I can do something about it.

My intention was that these bindings only have an effect when the file is a `.xit` file, because they only make sense then.

braver:
> Regarding this warning, please let me know if I can do something about it.

No it's good actually, the warnings do mention that your bindings are specific to a context. We are really careful about introducing bindings or overriding existing ones, but specific contexts sometimes need that (like here). If anything I would like to advocate adding bindings that feel more native on MacOS (e.g. use CMD/super instead of CTRL). 

I wouldn't say there's "nothing like it" on package control, we have various orgmode and similar packages. But the world never has enough checklist/todolist options 😉  

You're in luck, this must be the quickest package addition in the history of package control 😆

jotaen:
> You're in luck, this must be the quickest package addition in the history of package control 😆

Awesome, it’s amazing that this went through so quickly! 🎉 (The file format is currently [featured on HackerNews](https://news.ycombinator.com/item?id=30879327), so this is good timing.)

I also eagerly applied all the things that I learned from [my other package submission (klog)](https://github.com/wbond/package_control_channel/pull/8502), so thanks again for taking the time!

braver:
🥂

jotaen:
Sorry for bothering you again… I had made a small mistake: the original package name was just `xit`, and so the internal file paths were all wrong in version `1.0.0`, since the package is now actually called `xit!`.

I’ve corrected that via [version `1.1.0`](https://github.com/jotaen/xit-sublime/releases/tag/1.1.0) ([diff](https://github.com/jotaen/xit-sublime/compare/1.0.0...1.1.0)). Will that update automatically be picked up at some point, or do I have to actively do something?

jotaen:
Nevermind, the update went through now!

braver:
Yes it takes an hour or two maybe, but package control will pick up any new tags and roll them out to users.

PR #8514, opened 2022-04-07
added django-snippets-pro package

Ali-Aref:
This is Django Snippets Pro package for supporting Django snippets in Sublime Text. However there are other packages doing the same for sublime text but, it supports the newest Django3, models, forms, fields, python and HTML files, imports. it has file templates and ad-don auto-completions and much more.

It's an all in one package where developers don't need any extra package for HTML or for Django and so on. All Django related stuffs are categorized and implemented here.

The package is written on understandable syntax with comments for developers whom want to have contribution and for those want to customize for their extra needs.

It's fully tested! and I'm always available to helps deves for new features requests and solving any issues if availed.

I am a 5 year Django developer and I were using my package almost for 2 years and decided to share these snippets with other sublime users. I hope It be approved for Django developers better experience.

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -1080,6 +1080,18 @@
 			]
 		},
 		{
+			"name": "Django Snippets Pro",
+			"details": "https://github.com/Ali-Aref/Django-Snippets-Pro",
+			"labels": ["snippets", "python", "django"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": ["*"],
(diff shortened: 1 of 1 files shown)


braver:
You haven’t tagged a release yet. 

Your package also ships with syntaxes, not just snippets. In the ancient tmLanguage format no less (please have a look at the built in converter for the newer much better format). So maybe this package needs a different name?

PR #8516, opened 2022-04-09
Add Pep package

pedrorgirardi:

Thank you! 🙏 

https://github.com/pedrorgirardi/Pep

[2]: https://semver.org

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -744,6 +744,16 @@
 			]
 		},
 		{
+			"name": "Pep",
+			"details": "https://github.com/pedrorgirardi/Pep",
+			"releases": [
+				{
+					"sublime_text": ">=4094",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "PEPE Assembly",
 			"details": "https://github.com/patriq/Sublime-PEPE-Assembly",
 			"labels": ["language syntax", "snippets"],


braver:
Maybe tell us a little bit about your package and what it adds to the existing Clojure support packages we already have.

Other considerations before we continue with a proper review:
- Package control requires semver version, ie. 3 digits not 2.
- Sublime Text uses Python extensively, and within the Python ecosystem "pep" has a very specific meaning. I would also not have guessed your package has anything to do with Clojure, but that's maybe just me not knowing anything about Clojure. Altogether though, while "Pep" is short and sweet, it might not be a good name for the package. More functionally descriptive names tend to be more successful because they can be easily found and more easily understood.

pedrorgirardi:
Thanks for looking at this @braver ! Much appreciated. 🙏 

> Maybe tell us a little bit about your package and what it adds to the existing Clojure support packages we already have.

 I think **Pep** is a complementary Clojure plugin. Existing plugins e.g. **Tutkain** and **Clojure Sublimed** extend Sublime Text to give an interface to a Clojure REPL, but it doesn't do any static analysis. Pep doesn't do anything with a Clojure runtime but it provides analysis of your code.

 I have Tutkain and Pep in my setup - Tutkain lets me "poke" at the runtime, and Pep helps me check and navigate my code.

> Package control requires semver version, ie. 3 digits not 2.

Right! I can cut a release using `0.6.0`.

> Sublime Text uses Python extensively, and within the Python ecosystem "pep" has a very specific meaning. I would also not have guessed your package has anything to do with Clojure, but that's maybe just me not knowing anything about Clojure. Altogether though, while "Pep" is short and sweet, it might not be a good name for the package. More functionally descriptive names tend to be more successful because they can be easily found and more easily understood.

It's very intentional that the plugin name does not have any "Clojure" in its name. I understand PEP would be a confusing name in the Python ecosystem, but I don't think that's a problem in Clojure - nobody that I've spoken to in the Clojure community about Pep raised this issue. 

There are very few Clojure programmers using Sublime Text, and I think people that would have an interest in the Pep plugin will look specifically for it.

There are some examples of a plugin not having anything related to Clojure in its name e.g.: *Calva* and *Tutkain*.

Can I help with anything else?

braver:
Alright, let's go. Two things:

- please resolve the merge conflict
- remember to tag changes you want to publish, package control only pushes new versions out based on tags

pedrorgirardi:
> Alright, let's go. Two things:
> 
>     * please resolve the merge conflict
> 
>     * remember to tag changes you want to publish, package control only pushes new versions out based on tags

Great! Thanks :)

I didn't see any merge conflict, though.

PR #8520, opened 2022-04-15
adding "elm query" package 1.0.0

gacallea:

My package is similar to to Hoogle Search, however it should still be added because there are no packages specific to Elm Packages. I will soon add functionality for Elm Search too.

@packagecontrol-bot

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -671,6 +671,17 @@
 			]
 		},
 		{
+			"name": "Elm Query",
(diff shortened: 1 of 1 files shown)


gacallea:
ps: on a second thought, while this package was **inspired** by hoogle search, it is actually **unique.** just in case it could be rejected for the wrong reasons... 😄

gacallea:
Fixes gacallea/elm-query#1

PR #8524, opened 2022-04-19
Added simple-black

phil-bell:
- [x]  If my package is a syntax it doesn't also add a color scheme. ***

My package is a plugin for the Python formatting library [Black](https://github.com/psf/black).

My package is similar to `python-black` however, it should still be added because it is a simpler implementation that relies on the user's installation of black rather than its own. The main benifits of this is it will reliably work the same as the black cli tool when it comes to config and users can easily pick which version of black they would like to use.

[2]: https://semver.org

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1812,7 +1812,7 @@
 			]
 		},
 		{
-			"name": "Simple Black",
+			"name": "simple-black",
 			"details": "https://github.com/phil-bell/simple-black",
 			"releases": [
 				{
--- a/repository/s.json
+++ b/repository/s.json
@@ -1816,7 +1816,7 @@
 			"details": "https://github.com/phil-bell/simple-black",
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": ">=4000",
 					"tags": true
 				}
 			]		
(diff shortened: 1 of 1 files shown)


jrappen:
Have you considered sending a pull request to `python-black` with your changes, instead? Asking because this is the third `black` package for Sublime. The first one died already. Not sure whether adding a third is the best way forward.

jrappen:
Your README has a lot of typos. Maybe you could fix those?

phil-bell:
Heya, `python-black` goes about their plugin in such a different way I would expect a PR to move it towards black cli would be too large to be accepted. I also think it would be difficult to make their implementation as consistent and adaptable as this approach. 

Sorry for the typos, I think I've fixed them all now.

deathaxe:
Explicitly running `black` on a file can easily be achieved via build system.

The point of python-black is to provide a package without external dependencies.

braver:
In fact, [Fmt](https://packagecontrol.io/packages/Fmt) probably already supports this. IMO, there is no need for specific formatter packages if they don't offer anything beyond simply running a file through the command line utility. Because Fmt does it better (more flexibility in settings, more continued support) and as @deathaxe says a package is entirely unnecessary anyway because build systems can already do the same.

PR #8525, opened 2022-04-22
Add Image Link Insertion

jtojnar:
- [N/A] I use [.gitattributes][3] to exclude files from the package: images, test files, sublime-project/workspace.

My package allows quick insertion of many images selected using file chooser dialogue into a markdown document.

I was not able to find any packages like it in Package Control.

[2]: https://semver.org

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -237,6 +237,17 @@
 			]
 		},
 		{
+			"name": "Image Link Insertion",
+			"details": "https://github.com/jtojnar/sublime-image-link-insertion",
+			"author": "Jan Tojnar",
+			"releases": [
+				{
+					"sublime_text": ">=4075",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Image2Base64",
 			"details": "https://github.com/tm-minty/sublime-text-2-image2base64",
 			"labels": ["image manipulation", "base64", "image", "converting"],


jtojnar:
It looks like the CI failed for unrelated reason?

braver:
Your repo doesn't have any tags, see checklist item "I have tagged a release with a [semver](https://semver.org/) version number."

repository/i.json
```diff
@@ -236,6 +236,17 @@
 				}
 			]
 		},
+		{
+			"name": "image-link-insertion",
```
braver, line 240: You might want to consider calling it simply "Image Link Insertion"?

braver:
Since your functionality only applies to Markdown, it might make sense to limit the visibility of the commands to that syntax.

jtojnar:
> Your repo doesn't have any tags, see checklist item "I have tagged a release with a [semver](https://semver.org/) version number."

Looks like I forgot to push the tags, fixed.

jtojnar:
> Since your functionality only applies to Markdown, it might make sense to limit the visibility of the commands to that syntax.

The default settings apply to Markdown but the syntax can be customized (I still need to figure out how to best make syntax-specific settings). I have changed the menu item appropriately.

repository/i.json
jtojnar: Done, thanks for the suggestion.

braver: Ah, that looks better. Don't forget that this name will now be the path and identifier for your package, so that affects where Sublime will look for settings files etc.

braver:
> but the syntax can be customized

As in the way it inserts a link can be customized? I can see how your feature would apply elsewhere, eg. in HTML. You might want to have a look at how [HyperClick](https://github.com/aziz/SublimeHyperClick/blob/master/HyperClick.sublime-settings) provides customization for various different languages inside one settings file.

repository/i.json
braver: Oh, and you might want to update your command palette entries as well. Let me know when that's done, and we should be ready to ship.

jtojnar:
> As in the way it inserts a link can be customized? I can see how your feature would apply elsewhere, eg. in HTML. You might want to have a look at how [HyperClick](https://github.com/aziz/SublimeHyperClick/blob/master/HyperClick.sublime-settings?rgh-link-date=2022-07-03T18%3A31%3A55Z) provides customization for various different languages inside one settings file.

Right. When I used Atom, I had LaTeX-syntax-specific config, where I overrode the tags to LaTeX syntax. But as far as I can tell, in Sublime, package settings are completely independent of Sublime settings and cannot be changed in `Preferences` » `Settings – Syntax Specific`.

I would prefer not having the extension config branched like HyperClick like that but I am not sure if that is feasible.

braver:
> package settings are completely independent of Sublime settings and cannot be changed in Preferences » Settings – Syntax Specific.

Actually, it's a bit hidden but it is actually possible. You can put package specific settings in syntax specific and project specific settings just like any other settings (in fact you can put package settings in the main settings file). Look at [SublimeLinter](http://www.sublimelinter.com/en/stable/settings.html#project) for an example. You flatten the path to the setting and prefix it with the package name.

jtojnar:
Thanks, released 0.2.0 that adds support for [syntax-specific configs](https://github.com/jtojnar/sublime-image-link-insertion/commit/4eeb2136e1bf2a63f10f75d2c648a7a50a559979).

PR #8527, opened 2022-04-30
Open URL plugin: move st3 support onto st3 branch, st4 version released with tags

kylebebak:
This package, https://github.com/noahcoad/open-url, already exists. I'm just changing the `releases` dict to move code for the ST3 version of the plugin onto the `st3` branch. The Sublime Text 4 version of the plugin is now the one with `"tags": true`

[2]: https://semver.org

### DIFF
--- a/repository/o.json
+++ b/repository/o.json
@@ -649,12 +649,16 @@
 			"labels": ["open folders", "open files", "shortcut", "power user", "web search", "url"],
 			"releases": [
 				{
-					"sublime_text": ">=3000",
-					"tags": true
-				},
-				{
 					"sublime_text": "<3000",
 					"branch": "st2"
+				},
+				{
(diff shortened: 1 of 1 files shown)


braver:
Please don’t “at” me, there are also other reviewers, all doing this in our spare time without compensation. We’ll get to it when we get to it.

in this case please stop using branch based releases, that’s deprecated (and also not a good way to work). You should switch to prefixed tags.

braver:
Good point about the docs, that could be more explicit. I’ll make a mental not of clarifying that. It’s an implicit side effect of the implementation of semver, but not everyone is familiar with the details of that specification.

kylebebak:
It didn't come off as cranky, just direct, and it pointed me in the right direction. Thank you! 😄

PR #8528, opened 2022-04-30
Add Purescript Format plugin

sepisoad:
this is a small plugin that perform code formatting on purescript codes


braver:
Please have a look at the Fmt package, it can probably already handle the formatted you use.

PR #8530, opened 2022-05-02
Added package PEPEasm

Jonyleo:

My package is syntax highlighting for the PEPE assembly programming language, used in a introduction to computer architecture class in IST at university of Lisbon.

My package is very similar to [PEPE Assembly](https://github.com/patriq/Sublime-PEPE-Assembly), however it should still be added because the original package has fallen behind of the new features of the language.

I've tried getting in contact with the maintainer, via github issues and e-mail, but I have not gotten a response.

I would try to get the original package replaced with mine, however I can't find a way to request that.

For this reason I decided to make my own package that addresses this issue. 

I would also like to add that I work on the official compiler for this language, if that adds any weight to my application!

[2]: https://semver.org

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -755,6 +755,16 @@
 			]
 		},
 		{
+			"name": "PEPEasm",
+			"details": "https://github.com/Jonyleo/SublimePEPEAsm",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Perfect Code",
 			"details": "https://github.com/chaserogers/PerfectCodeColorScheme",
 			"releases": [
--- a/repository/p.json
+++ b/repository/p.json
@@ -745,18 +745,8 @@
 		},
 		{
 			"name": "PEPE Assembly",
-			"details": "https://github.com/patriq/Sublime-PEPE-Assembly",
-			"labels": ["language syntax", "snippets"],
-			"releases": [
-				{
-					"sublime_text": "*",
-					"tags": true
-				}
-			]
-		},
-		{
-			"name": "PEPEasm",
-			"details": "https://github.com/Jonyleo/SublimePEPEAsm",
+			"details": "https://github.com/Jonyleo/PEPEAsm",
+			"labels": ["language syntax"],
 			"releases": [
 				{
 					"sublime_text": "*",


braver:
Thanks for your submission. I have seen your issue in the original repo. In such cases we typically give the original author, @patriq in this case, some time to object (say 2 weeks). Then we can proceed to replace the existing package with the new one.

Two things to do in preparation for the replacement to work:
- Ensure the latest tag in your repo is a higher version than the last one in the old repo.
- Change your submission in this PR to simply replace the repo url of the existing entry to point to your repo.

You can ignore those warnings from the bot. The version compatibility issue is not really relevant anymore and a license is optional (though recommended).

Your syntax should be slightly tweaked though. To enable package authors and color scheme designers to specifically target your syntax, all your scopes should end with something like “.pepeasm”.

Jonyleo:
Thank you for the feedback, I will push the new version shortly. I do have a question, I want to have simple completion for words that already exist in the file, the documentation says that sublime does that as default but it does not seem to happen.

Jonyleo:
I think I've done all you asked correctly, thank for the feedback.

I'm not great with github but I believe my latest tag in the PEPEAsm repo points to the latest commit!

patriq:
Hello! Sorry this issue was somehow filtered by my inbox.  Also sorry @Jonyleo in-case I have ignored any of your communication attempts (I have only now seen the issue created in my repo).

I am not sure about the similarity of our projects. The language I added syntax and snippets for was this very niche assembly for this very niche microprocessor made to teach the [Introduction to Computer Architecture course](https://fenix.tecnico.ulisboa.pt/disciplinas/IAC317/2015-2016/1-semestre) at the Instituto Superior Técnico University in Lisbon.

As stated in the README of my project:
```
This is a syntax definition for assembly that works on a processor called PEPE which was designed by my Introduction to Computer Architecture teacher. Thus you should only use it if you are actually developing for it.
```

So unless you are a student from this specific course I would rather keep ownership and name of the package.

Jonyleo:
Hello @patriq, thanks for the response.
I am indeed a student of that class(took it a few years ago) and I've been working with the teacher to develop new language features for the processor that are missing from your package.

I believe that puts me in a good position to own this package.

I will say however, that when I took the class, your package was a great tool to aid our work. Thank you for your contribution.

I would love to add a note to my repo thanking your previous work, let me know if you would like that!

patriq:
@Jonyleo sounds good!! Then what I would suggest is that I just transfer ownership of the original package to you, keeping the name PEPE Assembly and the original git commit history.

What do you think?

Jonyleo:
I would be overriding all the files. My grammar is not based on yours, it's redone from scratch.
With that being said I'm not 100% sure on the pros of your suggestion

patriq:
So users wouldn't have to download a whole new package, getting the updated version right away.

Also keeping download history is a nice feature.

braver:
Indeed, that would be the best course of action as it keeps everything in one place and all references to this package will continue to point to the latest version.

PR #8531, opened 2022-05-04
Add RegexMatch package

mercurykd:

My package is Testing regular expressions

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -952,6 +952,16 @@
 			]
 		},
 		{
+			"name": "RegexMatch",
+			"details": "https://github.com/mercurykd/RegexMatch",
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "RegexPortfolio",
 			"previous_names": ["Regex Portfolio"],
 			"details": "https://github.com/KaminoU/regex_portfolio",


braver:
Very interesting package idea! Some questions:
 
- Do you really need the keybinding?
- The feedback UI seems a bit limiting for larger regexes with more capture groups. Maybe you could utilize the output panel for more details about the matches?
- Are you fully highlighting the regex, like with escape sequences etc?

mercurykd:
> * Do you really need the keybinding?

no

> * The feedback UI seems a bit limiting for larger regexes with more capture groups. Maybe you could utilize the output panel for more details about the matches?

good idea

> * Are you fully highlighting the regex, like with escape sequences etc?

didn't really understand the question

mercurykd:
add license and delete keybinding

mercurykd:
add:
- explain in panel
- display groups like ~()~

mercurykd:
removed global variables
color scheme patch
redundant events removed

braver:
> > Are you fully highlighting the regex, like with escape sequences etc?

> didn't really understand the question

If you look at a regex in say Python, sequences, certain characters and escapes are highlighted with different colors. In your package it's just a string. Would be nice to have syntax highlighting for the regex. Since you probably only support Python regexes anyway (which is something you might want to clarify in your readme, the underlying regex engine can be quite relevant), you might be able to apply the same syntax in this context?

mercurykd:
> If you look at a regex in say Python, sequences, certain characters and escapes are highlighted with different colors

do you mean highlight the regex pattern itself?
image

braver:
yes!

braver:
Well, consider those improvements in your own time. I think we should just ship it now as is.

PR #8535, opened 2022-05-07
Add backwards compatibility for Atomic Soft Tab Nav package

ruiokada:

This pull request is not to add a new package but to introduce backwards compatibility to an existing package. For the initial release of my package, I only tested on Sublime Text 4 but I have now confirmed that it works on Sublime Text 3 as well.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -2026,7 +2026,7 @@
 			"details": "https://github.com/ruiokada/sublime-text-atomic-soft-tab-nav",
 			"releases": [
 				{
-					"sublime_text": ">=4000",
+					"sublime_text": ">=3000",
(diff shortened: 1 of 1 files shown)


Ultra-Instinct-05:
I'd really recommend not maintaining backwards compatibility with ST3. There have been several stable ST4 releases so far, so there is effectively no need to use ST3 anymore.

ruiokada:
> I'd really recommend not maintaining backwards compatibility with ST3. There have been several stable ST4 releases so far, so there is effectively no need to use ST3 anymore.

Okay, I suppose if a user is inclined to use ST3 then they can just do a manual install. I will close this PR then.

PR #8537, opened 2022-05-11
added a fork of unmaintained package AdvancedNewFile

kendfss:


braver:
If @skuroda is indeed definitely no longer maintaining AdvancedNewFile I would much rather pick one of these solutions:

- you become a contributor/maintainer of the existing repo
- failing that we replace the repository URL of the existing package with your fork

We generally don't duplicate packages. 

(also, but moot since we won't consider adding your package, please don't format the entire file in a PR like this)

rchl:
I would even suggest forking it to https://github.com/SublimeText repo as it's high profile one and it would be easier to ensure that it stays relevant. Not to mention that there is some risk associated with letting "random" (sorry, no offense) people take over very popular packages.

FichteFoll:
@skuroda I have added you to the SublimeText org. Feel free to prepare a PR that updates the PC entry and move your repository there (GitHub sets up redirects but not reliably for the API).

skuroda:
File #8541. Let me know how you want to coordinate the move to minimize disruptions (or if you'd rather me do it first and let you you get to the merge at your leisure).

kendfss:
I'm closing this because I'd assumed @skuroda wasn't frequenting github after checking the issues. 
The feature I chose to implement was simply restoring the ability to create anonymous files, and I was likely not going to address the open tickets.

Regards, and thanks for writing an otherwise stellar utility! I genuinely think it should replace the default New File command in sublime.

skuroda:
@kendfss , I appreciate you trying to enhance the plugin. Please do keep an eye on the other PR so we can get the enhancement in for others to use.

PR #8538, opened 2022-05-12
re-add AlgoLang package now that is has been moved to github

qraynaud:

My package is added to support syntax for an algorithmic language.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -588,6 +588,18 @@
 			]
 		},
 		{
+			"name": "AlgoLang",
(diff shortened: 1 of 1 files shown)


qraynaud:
I just added a LICENSE file!

braver:
Oh, that's odd. Was the original entry removed? Is this just the same package but moved, or is there any change in content we should consider?

braver:
Eh...

> Is this just the same package but moved, or is there any change in content we should consider?

qraynaud:
Sorry, I missed your previous message. It’s the same content but moved.

PR #8539, opened 2022-05-15
Add Hare language highlights

artursartamonovs:

My package is ...

[2]: https://semver.org

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -231,6 +231,17 @@
 			]
 		},
 		{
+			"name": "Hare Highlight",
+			"details": "https://github.com/artursartamonovs/hare-highlight",
+			"labels": ["language syntax"],
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


braver:
Well that's a bit minimal, but it should work 👌🏻

repository/h.json
```diff
@@ -230,6 +230,17 @@
 				}
 			]
 		},
+		{
+			"name": "Hare Highlight",
```
braver, line 234: Language packages are simply named after the language they support. Any other words like "syntax" or "highlight" etc. are superfluous and misleading: a syntax/language package unlocks many more features in Sublime Text that just the highlighting and typically includes things like snippets etc. So you can just name your package "Hare".

artursartamonovs: Updated name to "Hare" in package channel
Added LICENSE file to syntax/language package repository

PR #8540, opened 2022-05-17
Add MarLant

retifrav:
- [x] I'm the package's author and/or maintainer
- [x] I have have read [the docs][1]
- [x] I have tagged a release with a [SemVer][2] version number
- [x] My package repository has a description and a README describing what it's for and how to use it
- [ ] My package doesn't add context menu entries
- [x] My package doesn't add key bindings
- [x] Any commands are available via the command palette
- [x] Preferences and keybindings (if any) are listed in the menu and the command palette, and open in split view
- [x] If my package is a syntax it doesn't also add a color scheme
- [x] I use [.gitattributes][3] to exclude files from the package: images, test files, sublime-project/workspace

My package is a plugin for working with SubRip/SRT subtitles. It provides commands for inserting new titles, splitting and joining existing ones, renumbering titles ordinals, shifting all the titles timings and subtitles validation. It also provides a SubRip/SRT language syntax.

There are no packages like it in Package Control. I found these two: [SRT](https://packagecontrol.io/packages/SRT) and [SubVal](https://packagecontrol.io/packages/SubVal), but they provide a single functionality/feature each.

My package does add context menus:

- in tab context menu: 2 items for file-based operations
- in text area menu: 1 item which is a group for several nested items

If you'd prefer it didn't add context menus, that is not a problem, I can delete them from sources and add instructions for users in README.

I know that `author` property should not be provided for a GitHub-hosted package, but my GitHub username isn't quite what I wanted, and so I've set it.

[2]: https://semver.org

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -800,6 +800,18 @@
 			]
 		},
 		{
+			"name": "MarLant",
+			"details": "https://github.com/retifrav/marlant",
+			"labels": ["subrip", "srt", "subtitles", "language syntax"],
+			"author": "retif",
+			"releases": [
+				{
+					"sublime_text": ">=4099",
+					"tags": "v"
+				}
+			]
+		},
+		{
 			"name": "Marshal Command Code",
 			"previous_names": ["MinecraftCommandCode"],
 			"description": "Syntax highlighting for Minecraft Commands",


braver:
Make sure all your scopes end with a consistent suffix (`.srt` in this case) so your syntax can be targeted by package and color scheme authors.

Context menu entries are acceptable, and these seem to be useful within the context. However, make sure that they are conditional and only show up in files with your syntax.

Same for the Tab Context: if these commands show up every time that could be annoying. Users use Sublime Text for loads of different and unrelated things and no package should take up space in general context menus. Unless I missed something, I would be in favor or removing these menu items.

Make sure all your commands are available in the command palette. Users will look for your features there, or maybe in the tools menu. If you have that covered, context menu items and key bindings will be of secondary importance.

retifrav:
> make sure all your scopes end with a consistent suffix (`.srt` in this case)

If I got that right, this is already done, more details in the next item.

> make sure that they are conditional and only show up in files with your syntax

That's how it is already, all the commands are conditional and are available only for `text.srt` scope. Here's how I did it:

```
def is_enabled(self) -> bool:
        return self.window.active_view().match_selector(0, "text.srt")

def is_visible(self) -> bool:
        return self.window.active_view().match_selector(0, "text.srt")
```

> make sure all your commands are available in the command palette

That's also already done.

PR #8541, opened 2022-05-22
Move AdvancedNewFile pointer to SublimeText organization.

skuroda:
Move the AdvancedNewFile plugin from it's current location (in my [space](https://github.com/skuroda/Sublime-AdvancedNewFile)) to the SublimeText organization.

Note at the time of filing this the project has not been moved to allow for coordination and minimize disruption to end users.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -407,7 +407,7 @@
 		},
 		{
 			"name": "AdvancedNewFile",
-			"details": "https://github.com/SublimeText/Sublime-AdvancedNewFile",
+			"details": "https://github.com/SublimeText/AdvancedNewFile",
 			"releases": [
 				{
 					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


deathaxe:
For this PR to get merged the repository needs to be handed over to SublimeText organisation first.

When using `Transfer ownership` to move the repository all current urls continue to work.

I'd suggest to remove `Sublime-` prefix, once the repo is transfered.

FichteFoll:
The order isn't hugely important for the process, but if the rename happens first we can ensure during the PR merge that everything has been set up correctly and aside from the API, Github does indeed setup redirects which makes verifying it much easier.

Could you please remove the "Sublime-" prefix after transfering, as @deathaxe suggested, and update the PR for the new name? I will also add you as a maintainer of that repo once it has been transferred, so you can continue adding collaborators as you see fit.

skuroda:
Repo has been transferred and PR updated. Will do the rename once I have access, or feel free to change it yourself.

FichteFoll:
All done, thanks for your cooperation. :+1

PR #8544, opened 2022-05-31
Add Sublime Tips

kapitanluffy:

My package is **Tips**

[2]: https://semver.org


kapitanluffy:
@packagecontrol-bot

jrappen:
Please use a sane target for supported Sublime Text Builds, I suggest `>=4107`.

jrappen:
You should also add the git helper files to `.gitattributes` with export-ignore.

jrappen:
Not sure about the value of the tip: `Sublime Text is the second best editor!`

jrappen:
The tips you provide seem to target Windows specifically, yet the plugin's installation isn't limited to Windows only in this PR.

What are your plans in this regard?

kapitanluffy:
@jrappen updated the repo. It now supports osx.

kapitanluffy:
@braver sorry. removed it 😅

braver:
You could simplify a bit further and remove the settings file. Users can simply uninstall or disable the package if they don't want hints. 

I would think that making hints about key bindings would be a bit more involved than simply switching what key the "primary" or "super" key is? For Win/Linux chords are used more often, whereas on MacOS there are many more modifiers available and chords are usually avoided. And typically a Mac keyboard doesn't have Home or End keys (you use CMD + arrow keys instead). I'm sure you double checked your list, but this approach wil really limit what you can cover. And of course as soon as someone starts customizing the hints are no longer correct (presumably at that level you no longer need or want hints).

But it looks like a fun little, and harmless, package, so I wouldn't let its limitations stop us from adding it.

One final consideration: typically the same for such a feature would be "Tips", the dreaded "Show tips at start up" popups etc. 😉 Wouldn't that be a better name? "Hints" in software development are usually more contextual, whereas this is simply random.

PR #8548, opened 2022-06-12
Add PyScript

Ultra-Instinct-05:

My package is [PyScript](https://github.com/Sublime-Instincts/PyScript) which provides syntax highlighting for [PyScript](https://pyscript.net/)

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -3004,6 +3004,17 @@
 			]
 		},
 		{
+			"name": "PyScript",
+			"details": "https://github.com/Sublime-Instincts/PyScript",
+			"labels": ["syntax highlight", "python", "pyscript", "language syntax"],
+			"releases": [
+				{
+					"sublime_text": ">=4126",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "PySide",
 			"details": "https://github.com/DamnWidget/SublimePySide",
 			"releases": [


jrappen:
Was there a reason for shipping an empty gitignore file? Try:

```gitignore
# macOS

.DS_Store
Thumbs.db

# Windows

[Dd]esktop.ini

# Python

*.cache
*.log
*.py[co]

# Sublime Text

*.sublime-workspace
```

Ultra-Instinct-05:
> Was there a reason for shipping an empty .gitignore file? Try: 

None of those files are present in the repository, so I don't really see a point in trying to ignore something not present.

> Are you sure about the base scope of your syntax being the same as the one of the syntax you're extending from?

For this, I am a little unsure. `text.html.python` sounds wrong because HTML also has other embedded syntax like CSS, JS, JSON etc. So I just stick to `text.html.basic` for now.

> Why use .html as an extension for your test file when you set .htmlpy in the syntax? Test works because the first line with the config overrides the wrong extension of the file.

`.htmlpy` was simply set just so I could have some extension to test against. PyScript, doesn't have a unique extension. It's still a regular HTML file.

> Whatever is in ./resources/syntax/ can (and should?) be moved to the root.

Personal preference. All the resource files are present in a `resources` directory segregated by the type of resource file. This just happened to have only one syntax, so there is just a single `syntax` folder. This structure makes sense for me and I just want to stick to it.

PR #8549, opened 2022-06-13
Added Highlight Custom Chars Unicode repo

Hootrix:

My package is similar to "Highlight Dodgy Chars" However it should still be added because 

1. Customize the highlight color
2. Customize the highlight text
3. The highlighted characters are modified to unicode encoding, similar to the effect in vim
4. Easy selection of special characters at the beginning of the line

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -1568,6 +1568,16 @@
 					"tags": true
 				}
 			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


braver:
Please have a look at the output from the CI job.

braver:
👋🏻

PR #8553, opened 2022-06-20
Add StatusBarClock package

lukstep:
- [x] I have read [the docs](https://packagecontrol.io/docs/submitting_a_package).
- [x] My package doesn't add context menu entries. See Context Menu Below.
- [x] My package doesn't add key bindings. See Key Bindings Below.

My package is [StatusBarClock](https://github.com/lukstep/StatusBarClock) 

My package is similar to [Fuzzy Clock](https://github.com/SarthakU/fuzzy-clock-sublime) However it should still be added because it has more options to customize.

[2]: https://semver.org

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -3791,7 +3791,7 @@
 			"labels": ["status bar", "clock"],
 			"releases": [
 				{
-					"sublime_text": ">=3000",
+					"sublime_text": ">=4107",
 					"tags": true
 				}
 			]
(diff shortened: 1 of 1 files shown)


jrappen:
Not sure if you already looked at:

<https://github.com/deathaxe/sublime-clock>

which isn't on PackageControl itself, but similar to this.

---

You could use py3.8 for this for `"sublime_text": ">=4107"`, compare:

<https://www.sublimetext.com/docs/api_environments.html>

(or deathaxe's clock above, which uses py3.8) as Sublime Text 3 is already outdated and the vast majority of users has already switched to newer Builds.

lukstep:
Yes, I saw mentioned package. It's similar but not available on PackageControl.

braver:
Looks good, thanks for your submission 👍🏻

PR #8554, opened 2022-06-20
Add NativeScript Core XML Syntax

CatchABus:

My package is [NativeScript Core XML Syntax](https://github.com/CatchABus/sublime-nativescript-xml-syntax).

This package adds syntax highlighting for [NativeScript Core](https://docs.nativescript.org/tutorial/plain.html#getting-started-with-nativescript) XML files.

Syntax is identical to the default one for XML with the addition of highlighting view model bindings.
The highlight color is the one used for CDATA tags.

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -112,6 +112,16 @@
 			]
 		},
 		{
+			"name": "NativeScript Core XML Syntax",
+			"details": "https://github.com/CatchABus/sublime-nativescript-xml-syntax",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "NativeScript Snippets",
 			"details": "https://github.com/tsvetan-ganev/nativescript-sublime-snippets",
 			"labels": ["auto-complete", "snippets"],
--- a/repository/n.json
+++ b/repository/n.json
@@ -116,7 +116,7 @@
 			"details": "https://github.com/CatchABus/sublime-nativescript-xml-syntax",
 			"releases": [
 				{
-					"sublime_text": "">=3092",
+					"sublime_text": ">=3092",
 					"tags": true
 				}
 			]
--- a/repository/n.json
+++ b/repository/n.json
@@ -116,7 +116,7 @@
 			"details": "https://github.com/CatchABus/sublime-nativescript-xml-syntax",
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": "">=3092",
 					"tags": true
 				}
 			]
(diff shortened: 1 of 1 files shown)


CatchABus:
@packagecontrol-bot

CatchABus:
@packagecontrol-bot

CatchABus:
@packagecontrol-bot

CatchABus:
@packagecontrol-bot

repository/n.json
```diff
@@ -111,6 +111,17 @@
 				}
 			]
 		},
+		{
+			"name": "NativeScript Core XML Syntax",
```
braver, line 115: Language support packages can just be named after the language, without suffixes like "syntax", "language", "highlighting" etc. In this case I would lean towards "NativeScript XML"?

braver:
Have you considered extending or including the default XML syntax instead of duplicating most of it?

Your main scope is `text.xml`, but since you're extending the basic syntax for that, you might want to have a more specific one like `text.xml.nativescript`. Everything that supports  `text.xml` would also support your syntax, but yours can now be targeted specifically as well.

I'm a bit worried about this replacing the default XML syntax for many users, for files that are not necessarily NativeScript. May or may not be a problem, especially if this is truly an extension of the default and maybe only for specific file extensions.

repository/n.json
CatchABus: NativeScript is a mobile framework with multiple flavors (NativeScript Vue, Svelte, etc). NativeScript Core is a flavor itself, hence the `NativeScript Core` reference.

NativeScript Core consists of JS, CSS, and XML files. In XML, there is a custom expression syntax where attribute values are enclosed in double curly braces (e.g. `{{ myVariable }}`), that is why I focus on syntax.

CatchABus:
> Have you considered extending or including the default XML syntax instead of duplicating most of it?
> 
> Your main scope is `text.xml`, but since you're extending the basic syntax for that, you might want to have a more specific one like `text.xml.nativescript`. Everything that supports `text.xml` would also support your syntax, but yours can now be targeted specifically as well.
> 
> I'm a bit worried about this replacing the default XML syntax for many users, for files that are not necessarily NativeScript. May or may not be a problem, especially if this is truly an extension of the default and maybe only for specific file extensions.

NativeScript is an existing framework that makes uses of XML files for UI rendering stuff.
This is currently a really small plugin that adds support for highlighting double-curly brace expressions `{{ }}` so it's a very small addition for xml in common. I can understand if it's a problem.

I could find a solution when plugin contains more features that affect the general use of XML.

repository/n.json
braver: Alright, makes sense then 👍🏻

PR #8555, opened 2022-06-26
(chore) rename Tint => Tint Terminal

joshgoebel:
- [x] My package _adds only bindings for scratch views it creates._
- [x] Preferences and keybindings (if any) are listed in the menu and the command palette, and open in split view. _The upcoming update resolves this._

This is just the renaming (and refreshing of a package already in the library).

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -2378,7 +2378,7 @@
 			"name": "Tint Terminal",
 			"previous_names": ["Tint"],
 			"details": "https://github.com/joshgoebel/sublime-tint-terminal",
-			"author": "joshgoebel",
+			"author": "Josh Goebel",
 			"labels": [
 				"terminal",
 				"shell",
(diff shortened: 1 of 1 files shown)


repository/t.json
```diff
@@ -2375,19 +2375,23 @@
 			]
 		},
 		{
-			"name": "Tint",
-			"details": "https://github.com/yyyc514/Tint",
-			"author": "jgoebel",
+			"name": "Tint Terminal",
+			"previous_names": ["Tint"],
+			"details": "https://github.com/joshgoebel/sublime-tint-terminal",
+			"author": "joshgoebel",
```
braver, line 2381: Since the author here is just your GitHub username, you can leave it out.

joshgoebel: I see a lot of people using their liternal name perhaps I should change it to "Josh Goebel" then?

braver: Yes, you can also do that 👍🏻

joshgoebel: Better?

braver: 👍🏻

PR #8558, opened 2022-06-30
Transformice Lua API syntax package (FormatTransformice)

artssan:

My package is ...

[2]: https://semver.org

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -1573,6 +1573,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


braver:
Please check those open items in the list and tell us why you don't comply. Also, fill in some info on those dots so that we know what your package is about.

braver:
Please respond to the feedback so we can continue the review.

PR #8559, opened 2022-07-01
Added CompareBuff package

nexional:
SublimeText package to compare two buffers using external comparison tool


nexional:
Hi, For "CI / trigger-review" failure could you provide what's the exact error we have & what can be done to fix it?

PR #8562, opened 2022-07-01
Added CompareBuff package

nexional:

My package is CompareBuff - [SublimeText package to compare two buffers using external comparison tool]
Link: https://github.com/nexional/CompareBuff

My package is similar to https://packagecontrol.io/packages/BeyondCompare However it should still be added because -
a. It allows you to chose two files/buffers you want to compare from any window
b. Not limited to Beyond Compare tool. You can configure any external comparison tool in settings
c. Can also work on just the selected area

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3129,6 +3129,20 @@
 				}
 			]
 		},
+        {
+            "name": "CompareBuff",
+            "details": "https://github.com/nexional/CompareBuff",
+			"readme": "https://github.com/nexional/CompareBuff/blob/master/README.md",
+			"issues": "https://github.com/nexional/CompareBuff/issues",
+            "author": "VK",
+            "labels": ["compare", "comparison", "beyond compare", "diff", "buffer"],
+            "releases": [
+                {
+                    "sublime_text": "*",
+                    "tags": true
+                }
+            ]
+        },
 		{
 			"name": "Compare Side-By-Side",
 			"details": "https://bitbucket.org/dougty/sublime-compare-side-by-side",
--- a/repository/c.json
+++ b/repository/c.json
@@ -3142,10 +3142,7 @@
 			]
 		},
 		{
-			"name": "CompareBuff",
 			"details": "https://github.com/nexional/CompareBuff",
-			"readme": "https://github.com/nexional/CompareBuff/blob/master/README.md",
-			"issues": "https://github.com/nexional/CompareBuff/issues",
 			"author": "VK",
 			"labels": ["compare", "comparison", "beyond compare", "diff", "buffer"],
 			"releases": [
--- a/repository/c.json
+++ b/repository/c.json
@@ -3129,25 +3129,25 @@
 				}
 			]
 		},
-        {
-            "name": "CompareBuff",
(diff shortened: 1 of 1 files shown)


repository/c.json
```diff
@@ -3141,6 +3141,20 @@
 				}
 			]
 		},
+		{
+			"name": "CompareBuff",
+			"details": "https://github.com/nexional/CompareBuff",
```
braver, line 3146: Since the info can be retrieved simply from this GitHub repo url, you can remove "name", "readme" and "issues".

braver:
Thanks for your submission.

I'm not convinced the keybindings and context menu entry are required here. We typically prefer to have the keybinding presented as something you can enable (and tweak) if you want to, to prevent overlapping with bindings of other packages. As for the context menu item, I would like to at least see a setting to enable/disable it. Everyone likes to have their package in all the menus, but the context menu should stay small and efficient and users need to be in control over what shows up in there.

nexional:
@braver thanks for your input. Regarding context-menu I don't seem to find a way to dynamically control it. All I see in the documentation is that we can create/remove a file Context.sublime-menu. Is there a way to control it programmatically which can be driven by user setting?

braver:
>  Is there a way to control it programmatically which can be driven by user setting?

Yes, you can add [is_visible](https://www.sublimetext.com/docs/api_reference.html#sublime_plugin.Command.is_visible) to the command and check an setting there.

nexional:
@braver Uploaded the [package](https://github.com/nexional/CompareBuff) with following changes:
- No default key-binding. If user desires it can be defined
- No default context-menu entry. It can now be controlled in settings
- All settings can we changed via Command Palette as well

Please review & let me know your comments.

nexional:
Can anyone review & approve it?

nexional:
> Can anyone review & approve it?

@wbond

braver:
Please don't ping anyone, we'll get to it when we get to it, nobody is getting paid here.

braver:
Packages are delivered in a zipped format and cannot be written to. If you want to enable/disable context menu items, implement `is_visible()` as I suggested earlier.

nexional:
> Packages are delivered in a zipped format and cannot be written to. If you want to enable/disable context menu items, implement `is_visible()` as I suggested earlier.

Thanks for your reply. I checked is_visible(). It's not practical since it hides command-palette entries as well thus giving user no option to run the package-command unless he has a command shortcut is defined which is not provided by default.

braver:
> Thanks for your reply. I checked is_visible(). It's not practical since it hides command-palette entries as well 

Well, only if you use one command. But you can have on command call the other, so you would create a separate command for the context menu and hide that based on the setting. 

What you're doing probably works now that you explain it. It will write an override into the user's packages directory. It's not as elegant though.

nexional:
> Well, only if you use one command. But you can have on command call the other, so you would create a separate command for the context menu and hide that based on the setting.
It doesn't look appealing to me that we put duplicate commands in command-palette (assuming context menu one is enabled)

nexional:
Updated to use is_visible() method now.

PR #8564, opened 2022-07-03
Move Astro package to SublimeText organization

deathaxe:
As Liquid, Astro has received some work to be based on Sublime Text's core syntaxes.

The author has been informed about the work and the idea about moving the repo via https://github.com/Destiner/sublime-astro/issues/5 without response so far.

[Astro](https://github.com/SublimeText/Astro) was therefore moved today to prepare release of latest changes/improvements.

We might still wait a bit, maybe we get some response in near future. Not sure.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -663,7 +663,7 @@
 					"branch": "master"
 				}
 			]
-		},	
+		},
 		{
 			"name": "Alrighty Snippets",
 			"details": "https://github.com/alrighty/alrighty-snippets-sublime",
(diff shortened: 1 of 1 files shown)


braver:
We tend to keep 2 weeks as a fair warning period around here and that has been respected in that issue. Maybe I was a bit quick on the trigger here, but we can always make alternative moves if the original author really objects.

FichteFoll:
@Destiner, as you can read above, we have replaced your package on the default PC channel with a fork to be maintained by the community. Usually, you would have had 2 weeks time to react on the issue, but unfortunately this time the process ended up slightly different.

If you are unsatisfied with this move, feel free to voice your concerns here within the next 2 weeks and we will revert the change or sort things out in a differnt way.

Destiner:
Sounds good 👍

PR #8565, opened 2022-07-04
added ProjectNotes

budRich:

My package is ...

[2]: https://semver.org

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -2596,6 +2596,17 @@
 			]
 		},
 		{
+			"name": "ProjectNotes",
+			"details": "https://github.com/budlime/ProjectNotes",
+			"labels": ["project", "notes", "todo"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "ProjectTreeTemplater",
 			"details": "https://github.com/julianxhokaxhiu/sublime-projecttreetemplater",
 			"releases": [
--- a/repository/p.json
+++ b/repository/p.json
@@ -2601,7 +2601,7 @@
 			"labels": ["project", "notes", "todo"],
 			"releases": [
 				{
-					"sublime_text": ">=3000",
+					"sublime_text": ">=4000",
 					"tags": true
 				}
 			]


braver:
Please tell us a little about your package, and have a look at the bot test results.

budRich:
i think the CI checks went through at first, but after i made the last commit (changed `"sublime_text": ">=3000",` to `"sublime_text": ">=4000",`) i got the red cross.

---

This package provides two commands:  
`project_notes_open` and `project_notes_list`  

---

## default settings

```JSON
{
  "notes_dir_path": "...",
  "note_file_name": "todo",
  "add_note_dir_to_project": true,
  "note_dir_name": "notes"
}
```
---

## project_notes_list

This will display a quickpanel listing the names
of projects which have a note named **$note_file_name**.
And open that file upon selection.

## project_notes_open

When the command is executed it will open the file:  
`$notes_dir_path/$PROJECT_NAME/$note_file_name`  

If there is no project associated with the active
window the command will not do anything.

If **notes_dir_path** is not a path to an existing
directory (example: `~/Documents/notes`), it will
default to: SUBLIME_PACKAGE_PATH/User/notes
(`~/sublime-text/Packages/User/notes`).

If **add_note_dir_to_project** is set to true the directory
containing the note will get added to the current project
with the display name defined for **note_dir_name**.

## Why?

I often find myself wanting to keep notes, lists
of links and other files related to a project, in its
own directory, *outside* the project, so i don't accidentally
commit those files to **git**.

This simple package makes that workflow seamless.

## keybinding

There is no keybindings enabled by default, I personally use
<key>Alt</key>+<key>n</key>:  
```JSON
{ "keys": ["alt+n"], "command": "project_notes_open" }
{ "keys": ["alt+shift+n"], "command": "project_notes_list" }
```

braver:
Thanks for the information, I'll need to get back to that at a later time. For now though, your tags need to adhere to the [semver](https://semver.org) specification (ie. 1.0.0, 1.0.1 etc)

budRich:
ok, i created a new release (no change in source) with a valid semver
https://github.com/budlime/ProjectNotes/releases/tag/2022.07.04

braver:
Lol semver sucks indeed. y.m.d probably passes the regex, but that's not semver. The concept is that you (e.g.) update the first number only for new versions that break compatibility in some way, not just because it's a new year. This isn't a discussion about the merits of semver, users of Sublime Text and Package Control will have expectations about what those version numbers mean. If you don't want to play along, just number them 0.1.0, 0.2.0, 0.3.0 etc or something.

budRich:
mfw releasing the 2022th version and is not believed.
also version 0.3.0 released.

budRich:
ok now there are not tags.....

PR #8566, opened 2022-07-06
Update w.json (-> added "Wikidot Syntax Highlighter")

MrAbadede:

My package is ...
a syntax highlighter for the Wikidot markup code. There are no packages at all for Wikidot.

[2]: https://semver.org

### DIFF
--- a/repository/w.json
+++ b/repository/w.json
@@ -469,6 +469,17 @@
 			]
 		},
 		{
+			"name": "Wikidot Syntax highlighter",
+			"details": "https://github.com/MrAbadede/sublime-wikidot-syntax-highlighting",
+			"labels": ["syntax", "highlight", "wikidot"],
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"branch": "master"
+				}
+			]
+		},
+		{
 			"name": "Wild Cherry",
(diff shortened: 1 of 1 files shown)


braver:
Please have a look at the Scope Naming guides, make sure your scopes are made specific to your syntax, and when you're ready tag a (correct semver numbered) release.

repository/w.json
```diff
@@ -468,6 +468,17 @@
 				}
 			]
 		},
+		{
+			"name": "Wikidot Syntax highlighter",
```
braver, line 472: Language support packages should just be named after their language. They do a lot more than simply add colors. In this case simply "Wikidot".

repository/w.json
```diff
@@ -468,6 +468,17 @@
...
+			"details": "https://github.com/MrAbadede/sublime-wikidot-syntax-highlighting",
+			"labels": ["syntax", "highlight", "wikidot"],
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"branch": "master"
```
braver, line 478: Branch based releases are not supported. Please set this to `"tags": true` and tag a release using proper semver version numbering.

braver:

PR #8572, opened 2022-07-18
Added HoverDocs

gladclef:

My package is ...

[2]: https://semver.org

Bonus points:
* Did a search for "documentation" on PackageControl.io and could not find a comparable package. The closest was OmniDocs.
* I can't put the commands from the .sublime-keybindings in the command pallette, due to the need to have all the information available in the EventListener.

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -1011,6 +1011,16 @@
 			]
 		},
 		{
+			"name": "HoverDocs",
+			"details": "https://github.com/gladclef/HoverDocs",
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "How Do I Code Search",
 			"details": "https://github.com/jlangston/SublimeHowDoI",
 			"releases": [


gladclef:
Repository link:
https://github.com/gladclef/HoverDocs

gladclef:
My package is a way to view symbol definitions (eg function parameters) and docstrings inline.

braver:
Don't we have LSP and such for that these days?

gladclef:
I didn’t know about LSPs! Thank you so much for teaching me something new!

I would still argue that this is useful in cases where LSPs don’t work very well, such as on company controlled computers where you don’t have the ability to install nodejs or the environment setup is too complex for the LSP to understand. If you agree with this sentiment, I can add something to the readme about how this package duplicates some of the functionality of the lsp package, but is useful in certain use cases such as those listed above.

braver:
I'm not against the idea, I certainly know of projects where LSP is overkill, unpractical, or simply doesn't work, where a "dumber" approach actually works better (for the that's a big part of what make Sublime Text awesome: it unlocks many of such features without the overhead of LSP, and you trade accuracy and richness of features). I don't think I've seen a recent attempt at something like this at least. If you look for "Doc" though we have lots of packages that do something with doc blocks. 

So, what kind of languages would your package support? Might be prudent to be up front about any limitations here.

gladclef:
I agree that part of what I like about Sublime is that it can be “dumb” or “smart”, depending on what you need.

It should support any language that has “comment” or “parameters” in it’s syntax scope names, although it’s worth noting that I’ve only tried it with python so far.

gladclef:
I like these suggestions. I’ll implement them and post a new message here once the updates are pushed.

gladclef:
I implemented the following changes:
* remove comment markings for per-line markings (eg "#" for python or "//" for c)
* remove pydocs markings for python (""")
* added link to LSP in readme
* added license
* add limitation that this has only been tested with python in README
* other small fixes

gladclef:
Ah, sorry! I didn’t update the tag! I’ll post an update with the tag tonight.

gladclef:
Alrighty @braver! I think this is ready for re-evaluation, thanks!

braver:
Cool, good job, let's ship it!

gladclef:
I agree that part of what I like about Sublime is that it can be “dumb” or “smart”, depending on what you need.

It should support any language that has “comment” or “parameters” in it’s syntax scope names, although it’s worth noting that I’ve only tried it with python so far.

PR #8574, opened 2022-07-23
Try to fix duplicate "Paste as Column" package entry

deathaxe:
packagecontrol.io lists "Package as Column" and "Package As Column". The former one is dead link to bitbucket. It also causes Package Control to fail installing it as it seems to look for old package.

I guess it's caused by a casing issue.

This commit is an attempt to trigger a cleanup. Not sure whether it works now that two packages exist already though.

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -451,6 +451,7 @@
 			"name": "Paste As Column",
 			"details": "https://github.com/TheClams/PasteAsColumn",
 			"labels": ["text manipulation", "clipboard", "paste"],
+			"previous_names": ["Paste as Column"],
 			"releases": [
 				{
 					"sublime_text": ">=3000",


braver:
Worth a shot 🤷🏻‍♂️

FichteFoll:
Where is the second package specified? If it's inside a third-party `repository.json`, not much we can do, but if it's inside one of our json files we should remove the other entry. Although the tests would have failed if that was the case.

deathaxe:
It appears it just has been renamed from `Paste as Column` to `Paste As Column` at some point. My hunch is some casing issue causing it to be listed as new entry, therefore.

But it may be something else, because packagecontrol.io still lists several packages which have been removed from registry long time ago.

deathaxe:
Duplicate item seems gone, but with all its 40k installs stats

PR #8577, opened 2022-07-24
Update QML repository URL

ratijas:
New syntax, new repo, new maintainer

[2]: https://semver.org

### DIFF
--- a/repository/q.json
+++ b/repository/q.json
@@ -69,11 +69,11 @@
 		},
 		{
 			"name": "QML",
-			"details": "https://github.com/skozlovf/Sublime-QML",
+			"details": "https://github.com/SublimeText/QML",
 			"labels": ["language syntax", "snippets"],
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": ">=4107",
 					"tags": true
 				}
 			]


jfcherng (approved):

FichteFoll:
Don't you want to prepare the entry for the new syntax currently in https://github.com/SublimeText/QML/pull/10 ? I.e. either restrict the supported ST version to ST3 or 4 (dependingg on the usage of branching) or add a legacy channel for the previous version that people on older ST versions can install.

ratijas:
> Don't you want to prepare the entry for the new syntax currently in https://github.com/SublimeText/QML/pull/10 ? […]

I… dunno?  Do I?  I barely know what are you talking about. What is the market share of old ST installations, and what are possible reasons against upgrading?

Not to mention that this plugin's target audience are ST users who build KDE from git master (i.e. me, let's be realistic here xD)

FichteFoll:
There isn't really a reason not to upgrade other than legacy OS support, or laziness, or in general being unable to update software. That said, none of this is a reason for us to break their existing, working installation by updating one of their packages to a version they cannot use anymore. For that reason, PC has a mechanism of declaring which ST builds are supported by a package and only allows installation (or updates) for supported builds.

Thus, you may either provide a legacy version for older ST builds pre-ST4 by adding a [separate release channel](https://github.com/wbond/package_control/blob/cfaaeb57612023e3679ecb7f8cd7ceac9f57990d/example-repository.json#L260-L288) or update the version selector to only allow ST4 versions (which currently says "all builds").

ratijas:
tbh I don't think an old version is worth keeping. It didn't do much expect highlighting JavaScript keywords and operators, and _sometimes_ uppercase identifiers as classes.

However, specifying lower bound is a good idea. So, which predicate that would be? `"sublime_text": ">=4000"` ?

FichteFoll:
I suggest limiting it to the first stable ST4 build or higher
.

ratijas:
> the first stable ST4 build or higher

And that would be [`Build 4107`](https://www.sublimetext.com/download)?

ratijas:
I checked the source code of this repo locally, grepped, and found a lot of plugins using 4107 as their lower bound. Makes sense to me.

PR #8580, opened 2022-07-27
Add LSP-copilot

TheSecEng:
- [ ] ~~If my package is a syntax it doesn't also add a color scheme. ***~~

My package is ... There are no packages like it in Package Control.

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -1733,6 +1733,17 @@
 			]
 		},
 		{
+			"name": "LSP-copilot",
+			"details": "https://github.com/TheSecEng/LSP-copilot",
+			"labels": ["AI", "auto-complete", "github"],
+			"releases": [
+				{
+					"sublime_text": ">4126",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Lua Dev",
 			"details": "https://github.com/rorydriscoll/LuaSublime",
 			"labels": ["lua"],


rwols:
Can you add an entry here instead?

https://github.com/sublimelsp/repository

repository/l.json
```diff
@@ -1732,6 +1732,17 @@
 				}
 			]
 		},
+		{
+			"name": "LSP-copilot",
+			"details": "https://github.com/TheSecEng/LSP-copilot",
+			"labels": ["AI", "auto-complete", "github"],
+			"releases": [
+				{
+					"sublime_text": ">4126",
```
jfcherng, line 1741: I guess `>=` 4126 since it's the latest stable build.

jfcherng: I wonder whether we want to prepend a min version now like https://github.com/sublimelsp/repository/blob/9391a520ca86a2e7ce3599a367689d4cbbf711cb/repository.json#L59-L68 so that if we bump the min version someday in the future, it won't be looks odd.

jfcherng: I wonder whether we want to prepend the min ST version to the tag at this moment like https://github.com/sublimelsp/repository/blob/9391a520ca86a2e7ce3599a367689d4cbbf711cb/repository.json#L59-L68 so that if we bump the min version, it won't looks odd.

jfcherng:
Adding to [sublimelsp/repository](https://github.com/sublimelsp/repository) is a green channel so that it can appear on Package Control hours later.

rwols:
Well, in case we ever switch to py38 it’s also good to have an overview of the existing LSP packages.

PR #8581, opened 2022-07-30
added HoverLine package

nexional:

My package is ...
HoverLine: Sublime Text package to hover over a line to see line in its entirety

[2]: https://semver.org

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -1011,6 +1011,18 @@
 			]
 		},
 		{
+			"name": "HoverLine",
+			"details": "https://github.com/nexional/HoverLine",
+			"author": "VK",
+			"labels": ["hover", "tooltip", "line", "word-wrap", "wrap", "popup"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "How Do I Code Search",
 			"details": "https://github.com/jlangston/SublimeHowDoI",
 			"releases": [


kaste:
(Disclaimer: I'm not an official reviewer here.)

- `word_wrap` can have 3 states: true, false, and "auto".  Typically (by default) it should be "auto" which is "trueish" and therefore your plugin bails out.   I would bail out only for the literal `True` value and otherwise check `view.match_selector(0, "source") which is roughly what "auto" is about and gives practical good results.

E.g.

```
        if not self.is_enabled() or view.settings().get('word_wrap') is True:
            return

        if not view.match_selector(0, "source"):
            return
```

- the plugin is written in a *very* idiosyncratic style. This adds an unnecessary burden to review IMO and you should consider just using an auto-formatter like `black` even though such formatter produce completely boring source code.

- the usage of `locals()` is over the top here, you can just access the variables by their name. E.g. `locals()['toggle_package']`is just `toggle_package`.  E.g. `if True not in locals().values(): return` does nothing here.  

- a settings change observer is unnecessary here because the callback `is_enabled` has no side-effect -- it is just a getter.  

Other than that; I think this plugin makes some sense and it appears useful.

kaste:
Lastly:  I would probably make the core functionality a `TextCommand` (which we can bind to a key or so) so that if we disbale `on_hover` we could still trigger it.  Mainly for keyboard only users.

repository/h.json
```diff
@@ -1020,6 +1020,18 @@
 				}
 			]
 		},
+		{
+			"name": "HoverLine",
+			"details": "https://github.com/nexional/HoverLine",
+			"author": "VK",
```
braver, line 1026: Who is VK?

braver:
Please respond to the feedback so we can continue the review.

repository/h.json
nexional: Will fix it to userid

kaste:
Also remove repository/h.json.orig

PR #8583, opened 2022-08-02
Add AutoF-Strings

vraneb:

My package is AutoF-Strings

[2]: https://semver.org

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -2302,6 +2302,16 @@
 			]
 		},
 		{
+			"name": "AutoF-Strings",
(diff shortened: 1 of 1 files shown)


FichteFoll:
Instead of enabling the functionality per syntax name, enbling it via a scope selector might be more flexible. Additionally, the current regexps you employ cannot differentiate between f-strings and `str.formt` strings. The latter usually have either nothing between the braces or a number, making them easy to distinguish from f-strings, but when the keyword syntax is used, they become indistinguishable.

PR #8590, opened 2022-08-15
Add SnailSVN-Helper

newnight:

My package is SnailSVN helper 

[2]: https://semver.org


braver:
What does your package do exactly? I don't really understand it and the Readme doesn't help. But I don't know what SnailSVN does so maybe it's just me.

Why is there a context menu? That makes more sense for interacting with selections etc., not for entire files. A menu item in the File menu would make more sense. And in any case all commands should be available via the command palette.

repository/s.json
```diff
@@ -2565,6 +2565,19 @@
 				}
 			]
 		},
+		{
+			"name": "SnailSVN_Helper",
+			"description": "This is a SnailSVN Helper for Sublime",
```
braver, line 2570: That much is clear from the name, but what does it do?

braver: Also, you can remove this and instead simply fill in a description on your GitHub repo. That will be used instead.

braver:

newnight:
SnailSVN is a TortoiseSVN-like Apache Subversion (SVN) client for Mac OS X, implemented as a Finder extension. 

official website is: https://langui.net/snailsvn/

this package will allow users to directly perform SVN operations in Sublime

braver:
Alright, thanks. I have made several other points in my earlier feedback, please address those as well.

PR #8596, opened 2022-08-18
Add LaravelPintFixer package repository

OthmanAdlan:

This package is specifically to run Laravel Pint over PHP files to fix their coding style according to configurable preset of standards.

[2]: https://semver.org

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -375,6 +375,17 @@
 			]
 		},
 		{
+			"name": "Laravel Pint Fixer",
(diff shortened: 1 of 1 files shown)


braver:
We have a [similar submission](https://github.com/wbond/package_control_channel/pull/8638) already, and the same feedback also applies:

Thanks for your submission, looks really useful. However, the package [Fmt](https://packagecontrol.io/packages/Fmt) already exists and would probably be able to provide this functionality as well, with more settings and flexibility.

PR #8599, opened 2022-08-19
Tidy up dependencies.json

deathaxe:
Tidy up dependencies.json

This PR introduces a common key sort order for all dependency entries to make transition to scheme 4.0.0 easier.

libraries: name, description, author, issues, load_oarder, releases

releases: base, platforms, sublime_text, tags

It doesn't change content.

### DIFF
--- a/repository/dependencies.json
+++ b/repository/dependencies.json
@@ -4,153 +4,159 @@
 	"dependencies": [
 		{
 			"name": "anytree",
-			"load_order": "01",
 			"description": "Python Anytree module - https://github.com/c0fec0de/anytree",
 			"author": "c0fec0de",
 			"issues": "https://github.com/c0fec0de/anytree/issues",
+			"load_order": "01",
 			"releases": [
 				{
-					"sublime_text": "*",
 					"base": "https://github.com/nbeversl/anytree_sublime",
+					"sublime_text": "*",
 					"tags": true
 				}
 			]
 		},
 		{
 			"name": "arrow",
-			"load_order": "01",
 			"description": "Python arrow module - https://github.com/crsmithdev/arrow",
 			"author": "douglas-vaz",
 			"issues": "https://github.com/douglas-vaz/arrow/issues",
+			"load_order": "01",
 			"releases": [
 				{
-					"sublime_text": "*",
 					"base": "https://github.com/douglas-vaz/arrow",
+					"sublime_text": "*",
 					"tags": true
 				}
 			]
 		},
 		{
 			"name": "backports_lzma",
-			"author": "dvhh",
-			"load_order": "50",
 			"description": "lzma modules for sublime text 2 and 3",
+			"author": "dvhh",
 			"issues": "https://github.com/dvhh/sublime-backports.lzma/issues",
+			"load_order": "50",
 			"releases": [
 				{
 					"base": "https://github.com/dvhh/sublime-backports.lzma",
-					"tags": true,
-					"platforms": ["windows"],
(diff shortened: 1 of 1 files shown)


repository/dependencies.json
```diff
@@ -4,1357 +4,1418 @@
...
-			"load_order": "01",
 			"description": "Python Anytree module - https://github.com/c0fec0de/anytree",
 			"author": "c0fec0de",
 			"issues": "https://github.com/c0fec0de/anytree/issues",
+			"load_order": "01",
 			"releases": [
 				{
-					"sublime_text": "*",
 					"base": "https://github.com/nbeversl/anytree_sublime",
+					"platforms": ["*"],
```
FichteFoll, line 14: Was there a particular reason for adding this to all entries where it was missing? E.g. considerations for ARM architectures and changing the default in 4.0.0?

deathaxe: Just want to have all entries to be of equal structure as it makes adding `python_versions` easier. We could also remove it from all, wich only contain asterisks, but I like explicity in general.

FichteFoll: The issue I have with it is that if we want it to be explicit, we also need to enforce it (in the tests), otherwise the next PR adding a dependency might not use it and we're back to where we started.

That said, I generally prefer reasonable defaults that do not need to be modified and not having to specify a platforms key to mean "install for all" seems like a reasonable default that reduces noise in the actual file.

deathaxe: So `"sublime_text": "*"` would also need to be stripped everywhere, too, as it is also a reasonable default. This PR is an attempt to introduce a common structure (used keys and their order) in at least libraries.json.

Expilicit keys were choosen as it is what will be needed for `"python_versions"` in scheme_version 4.0.0. It must list supported python versions explicitly. So it made sense to me to treat the other parts the same.

see: https://github.com/deathaxe/package_control_channel/blob/four-point-oh/repository/libraries.json

Order and presence of those wildcard keys is a mess across the repo.

FichteFoll: Yes, `sublime_text` is required because we request package authors and dependency authors to make a conscience choices about that, especially after ST3 was introduced with different Python versions and supporting both at the same time was not the norm. The same also holds true for the `python_versions` key, so I'm definitely on board with requiring that.

Now, I presume that `sublime_text` will be automatically softly restricted to something above 3000 due to the PC update not being available for ST2, otherwise `"sublime_text": "*", "python_versions": ["3.3"]` wouldn't make sense.

As for platforms, I still believe the default of allowing for all platforms is sensible, but seeing how basically every other key is required for dependencies I'm not strongly tied up on it.

braver:
@deathaxe and @FichteFoll can you please decide what to do with this PR? I'd hate for it to become yet another "good intentions, but impossible to roll out" PR that just sits here with no action for 5 years, that would make it just a waste of energy.

FichteFoll:
I want(ed) to strip out the additions of redundant `platform` keys and merge it. @deathaxe could you restore your branch?

deathaxe:
This PR was just meant to make transition to 4.0.0 scheme easier. As this will take a while anyway there's no real use for it anymore.

With new python_version string being a required key, I am uncertain whether we should enforce explicity on `platform` as well especially as `sublime_text` is required already, too. Either require everything or nothing, I'd say.

FichteFoll:
Hm true, with `python_version` joining the list of required keys, you've convinced me enough that this will be worthwhile. For now, I manually applied the diff and added a test to require the key in #8676.

PR #8601, opened 2022-08-20
Add Smithy IDL syntax support.

albe-rosado:

My package provides syntax support for [Smithy IDL](https://awslabs.github.io/smithy/2.0/index.html).

[2]: https://semver.org

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -2534,6 +2534,20 @@
 			]
 		},
 		{
+			"name": "Smithy",
+			"description": "Syntax Support for Smithy IDL.",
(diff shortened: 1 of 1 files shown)


albe-rosado:
Hi, I updated my extension repo with a new commit , addressing the warnings, would this workflow be rerun? @braver

braver:
You can ignore those warnings, the legacy syntax file is irrelevant by now. However, running the test again does demonstrate that package control will only see changes if you release a new tag. So let's ship by now, and remember to tag any future changes you make to the package!

PR #8607, opened 2022-08-25
New Ansible syntax for Sublime Text 4

killua99:

My package is an updated Ansible syntax, the other package is abandoned.

[2]: https://semver.org

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -1158,6 +1158,17 @@
 			]
 		},
 		{
+			"name": "Ansible Neo",
+			"details": "https://github.com/killua99/ansible-neo-sublime-syntax",
+			"labels": ["syntax"],
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


killua99:
what else does it need to advance ?

braver:
> what else does it need to advance

Patience padawan 🤗

braver:
@clifford-github your Ansible package for Sublime is really old. If you don't intend to update it, can you maybe

- give @killua99 access to your repo so he can continue your work and create new versions of the package
- or allow us to replace your package with his neo package

If we don't get a reaction within some time (typically 2 weeks, or whenever after that period we have time to revisit this PR) we'll opt for the second option.

clifford-github:
@braver I added @killua99 as contributor to the project.
The second option is OK from my side. Do I have to do something?

braver:
Oh, that's excellent. The ideal situation IMO is if the original work is continued in that a repo. That way everyone knows where to find it. Also, in that case we don't have to do anything here. Thanks all!

PR #8610, opened 2022-08-30
feat: catppuccin colorscheme package

ghostx31:

My package is ... Catppuccin colourscheme. It can be viewed at [catppuccin/sublime-text](https://github.com/catppuccin/sublime-text).

[2]: https://semver.org

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -416,6 +416,16 @@
 			]
 		},
 		{
+			"name": "Catppuccin", 
+			"details": "https://github.com/catppuccin/sublime-text",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "CBP Syntax Highlighter",
 			"details": "https://github.com/destruc7i0n/CBP_Sublime",
 			"labels": ["minecraft", "language syntax"],
--- a/repository/c.json
+++ b/repository/c.json
@@ -416,8 +416,10 @@
 			]
 		},
 		{
-			"name": "Catppuccin", 
+			"name": "Catppuccin color schemes", 
 			"details": "https://github.com/catppuccin/sublime-text",
+			"author": ["Catppuccin"],
+			"labels": ["color scheme"],
 			"releases": [
 				{
 					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


stfnx:
This one is almost two months old now... when will it get merged?

ghostx31:
We are also waiting for it to get merged. Don't know if I should ping the maintainer to have a look at this and get this merged.

repository/c.json
```diff
@@ -415,6 +415,16 @@
 				}
 			]
 		},
+		{
+			"name": "Catppuccin", 
+			"details": "https://github.com/catppuccin/sublime-text",
+			"releases": [
```
braver, line 421: Please add the color-scheme label.

repository/c.json
```diff
@@ -415,6 +415,16 @@
 				}
 			]
 		},
+		{
+			"name": "Catppuccin",
```
braver, line 419: Typically it's a good idea to add "color scheme", or in this case "colors schemes" in the name.

braver:
I think the assets directory and probably also the python file should be excluded from the package using `.gitattributes` file.

> Soothing pastel theme for Sublime Text

In Sublime Text parlance your package isn't a theme, it's a (collection of) color scheme(s). So I would suggest using the correct wording here also.

ghostx31:
I've removed the `assets` folder and the python file so that should be clear. I've also added the correct labels in the `c.json` file as requested. 

> In Sublime Text parlance your package isn't a theme, it's a (collection of) color scheme(s). So I would suggest using the correct wording here also.

About this, the `Soothing pastel theme for <App Name>` is the description convention we follow at Catppuccin so I don't think we need to remove that?

repository/c.json
```diff
@@ -415,6 +415,18 @@
...
 			]
 		},
+		{
+			"name": "Catppuccin color schemes", 
+			"details": "https://github.com/catppuccin/sublime-text",
+			"author": ["Catppuccin"],
+			"labels": ["color scheme"],
+			"releases": [
+				{
+					"sublime_text": "*",
```
txtsd, line 425: This should be `"sublime_text" : ">=3100"` because `.sublime-color-scheme` support was added in ST 3.1

txtsd (approved):

braver:
Nice 👍🏻

PR #8612, opened 2022-09-04
Add ParquetPython

dogversioning:
- [X] Any commands are available via the command palette. (None)
- [X] Preferences and keybindings (if any) are listed in the menu and the command palette, and open in split view. (There are none)
- [X] If my package is a syntax it doesn't also add a color scheme.
- [X] I use .gitattributes to exclude files from the package: images, test files, sublime-project/workspace.

My package is similar to sublime-parquet. However it should still be added because is uses a different underlying mechanism for decoding parquet files which may be useful to some users. Additionally, it is verified as supporting Windows, and the original package has not seen a non-readme update in six years.

[2]: https://semver.org

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -405,6 +405,17 @@
 			]
 		},
 		{
+			"name": "ParquetPython",
+			"details": "https://github.com/dogversioning/sublime-parquet-python",
+			"labels": ["parquet"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Particle Reverse Polish Language",
 			"details": "https://github.com/KubeRoot/Sublime-PRPL",
 			"labels": ["language syntax", "completions"],


braver:
> However it should still be added because is uses a different underlying mechanism for decoding parquet files which may be useful to some users.

The existing package is untouched for 6 years, that's never a good sign. If your package solves the same problem for end users I would still first look into having one good package. Can you maybe try to add your features to the existing package, maybe try to become a contributor to that repo? Also, would your package be a drop-in replacement? We try really hard not to duplicate packages.

braver:
Well it does sound like it's quite a bit different then. But maybe you could make the choice between one and the other a setting in that original package. For end users that's just the most convenient. We try (in contrast with say the marketplace for VS Code) to make it so users don't have to try a dozen different packages, all with the same purpose, to find one that works for them.
If that doesn't work out, we can of course add your package. As long as it's clear why one would use yours over the other.

dogversioning:
Makes sense - I'll try that approach and ping this PR if I don't here anything back after a month or so.

dogversioning:
Ok, this is going into the main package, closing this PR.

braver:
Cool, thanks!

PR #8617, opened 2022-09-27
Adjust SuperSettings readme URL

TobyGiacometti:
I have recently switched to using a `main` branch instead of a `master` branch. Since Package Control takes the readme from the `master` branch by default, I am now explicitly providing the readme URL.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -5280,6 +5280,7 @@
 		{
 			"name": "SuperSettings",
 			"details": "https://github.com/TobyGiacometti/SublimeSuperSettings",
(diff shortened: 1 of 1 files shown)


braver:
> Since Package Control takes the readme from the master branch by default

Does it naively use master or the default branch? I'm not sure but it looks like it's seeing your readme just fine.

TobyGiacometti:
Looks like it took a while to detect the correct branch. It definitely didn't see it for a few days after the adjustment. Not sure when it actually grabbed it since I did not check regularly.

PR #8628, opened 2022-10-22
Add LionDocs Plugin

Jalkhov:

My package is a Plugin to help to the translators community from mdn. Offers tools to make more easily the work of translate documents.

[2]: https://semver.org

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -1153,7 +1153,6 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"platforms": "*",
 					"tags": true
 				}
 			]
--- a/repository/l.json
+++ b/repository/l.json
@@ -1148,6 +1148,17 @@
 			]
 		},
 		{
+			"name": "LionDocs",
+			"details": "https://github.com/liondocs/liondocs-sublime",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "lioshiTheme",
 			"previous_names": ["lioshi Color Scheme"],
 			"details": "https://github.com/lioshi/lioshiScheme",


repository/l.json
```diff
@@ -1147,6 +1147,17 @@
 				}
 			]
 		},
+		{
+			"name": "LionDocs",
+			"details": "https://github.com/liondocs/liondocs-sublime",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": "*",
```
braver, line 1156: `"*"` is assumed, so you don't need to specify "platforms" here.

braver:
> Any commands are available via the command palette

Wouldn't it make sense to make the command available via the palette?

> My package doesn't add context menu entries. *

Eh, why did you check this while your package actually does introduce context menu entries? Also, it would be nice to have a setting to remove them.

> I use [.gitattributes](https://www.git-scm.com/docs/gitattributes#_export_ignore) to exclude files from the package

Your package seems to have several files that could be excluded this way.

Jalkhov:
> Wouldn't it make sense to make the command available via the palette?

For now I would just like to handle it like this, depending on the reception, later on it can be implemented.

> Eh, why did you check this while your package actually does introduce context menu entries? Also, it would be nice to have a setting to remove them.

You are right, my mistake.

> Your package seems to have several files that could be excluded this way.

For example...Please.

repository/l.json
Jalkhov: Removed

braver:
> For example...

`.editorconfig`, `.github/*`, `*.toml`

> For now I would just like to handle it like this

File-level (ie not context specific) command definitely need to be in the command palette though. 

Also, does your package need the context menu items, or would it make sense to be able to turn then off with a setting? Every package wants to be highly visible, but that menu would get really long and annoying if we did nothing about it.

Jalkhov:
> `.editorconfig`, `.github/*`, `*.toml`

Ok, you have reason, thanks.

> File-level (ie not context specific) command definitely need to be in the command palette though.

> Also, does your package need the context menu items, or would it make sense to be able to turn then off with a setting? Every package wants to be highly visible, but that menu would get really long and annoying if we did nothing about it.

Yes...it could be...but, isn't that more the developer's decision? I found it more comfortable to facilitate the Plugin options via Context Menu, I have been testing with my colleagues and none of them said anything about it, now, if it is a kind of standard to create plugins for ST, then the thing changes, but in the documentation of the api I did not see anything about it, unless I have overlooked it.

> or would it make sense to be able to turn then off with a setting?

Is that possible?

braver:
> now, if it is a kind of standard to create plugins for ST, then the thing changes

The documentation from ST doesn't really point you in a specific direction. Their interest is in describing how to do anything basically. Our concern here in package control (which is organizationally completely separate) is to create and maintain an ecosystem of packages. The main reason we review new packages is precisely this, to ensure all packages fit in and play nice in the wider ecosystem. Over time 4 important points we look for have become:

- ensure menu entries are in the proper structure and consistently labelled, using the split-view feature where possible
- ensure the context menu is not overloaded with entries and ensure that users have control over what entries are in there (because that menu stops being useful if it has 20+ things in it)
- ensure packages don't add keybindings but give users control over which bindings do what (because the keyboard doesn't have enough buttons and packages can override the main system as well as each other)
- ensure all commands are accessible via the command palette, if only because it means a users doesn't need to reach for their mouse to do something

> Is that possible?

Certainly. For instance you could add an `is_visible()` method to the command class that checks for the setting. Similarly you can use such a method to hide your command if it's not useful to a given context.

Jalkhov:
@braver Ready, I hope it does not seem that I am a little stubborn hehe I just like to know the why of things, and you have been able to explain it in the best way, based on what you tell me I may take two paths to go in line with the ecosystem:

1. Add the commands to the command palette and hide the context menu when not needed.
2. Add the commands to the command palette and remove the context menu completely.

Thank you very much for the information and support, as soon as I have everything ready I request the revision.

braver:
Excellent. It's no problem at all to explain things, I'm sure none of this is entirely obvious and our maintenance time tends to go towards reviewing submissions more than towards documentation. Let me know when you're ready to ship 👍🏻

Jalkhov:
@braver Everything is ready (on my side), I decided to remove the context menu, I also used .gitattributes, and all the commands are in the command palette.

Issue #6907, opened 2018-01-31
Add my fork of the BufferScroll to package control


evandrocoan:
@titoBouzout I have forked 2 of your packages I daily use and fixed some stuff which was not working. These are the packages:

1. https://github.com/evandrocoan/BufferScroll
1. https://github.com/evandrocoan/WordCount

Currently the `WordCount` package is already on Package Control from the fork of @philipbelesky:
1. https://github.com/philipbelesky/WordCount

But there are no new commits on it. @philipbelesky, can I add my fork in-place of yours?

If you prefer I can add both under a different name like `View Settings Save` and `View Status`.

philipbelesky:
@evandrocoan my fork was just for the purposes of publishing to Package Control so I wasn't planning on doing anything more than mirror @titoBouzout 's repo. I'm a bit apprehensive about swapping out which repo is linked to as `WordCount` on package control; to my mind tito has claim to that name given they initially published those packages. Ideally your PR would be merged but if not I think they should make the call as to which package is listed as `WordCount`

titoBouzout:
relevant for word count https://github.com/titoBouzout/WordCount/pull/56

Maybe you can add these using a different name, as gonna be confusing if you change stuff and there's two packages with the same name behaving differently. gl

evandrocoan:
Thanks for the response. I could then:

1. Keep the name and push a major release for both packages announcing breaking changes.
1. Add them both with new names.

@FichteFoll, which one of these options would accept merging?

FichteFoll:
Thanks for the references, but I do not have the capacity to review them.

If @titoBouzout thinks the changes are fine and can agree to you taking over the old packages, then let's do that. This is largely preferrable because all existing users of the packages will benefit from continued maintenance and fixed bugs. Less used features that were potentially removed could be communicated through update messages.
I suggest to provide a summary of your changs for @titoBouzout to review (you tried in https://github.com/titoBouzout/WordCount/pull/56, but try to be more concise without leaving out things. Then you can go on about the reasons for the changes in a different sections.

If this somehow doesn't work out, option 2 is also fine.

braver:
@evandrocoan Sorry, this has been sitting for a while. By now both original packages are really ancient. 

Can you make major releases for both, add message about anything that might break for existing users, and then create PRs to replace the existing WordCount package? 

[...]

PR #8629, opened 2022-10-25
Added the package WordingStatus

evandrocoan:

My package is a rewrite of WordCount (https://github.com/wbond/package_control_channel/issues/6907).

Fixes https://github.com/evandrocoan/WordingStatus/issues/5

[2]: https://semver.org

### DIFF
--- a/repository/w.json
+++ b/repository/w.json
@@ -750,18 +750,6 @@
 			]
 		},
 		{
-			"name": "WordCount",
-			"details": "https://github.com/philipbelesky/WordCount",
-			"homepage": "https://github.com/titoBouzout/WordCount",
-			"author": "titoBouzout",
-			"releases": [
-				{
-					"sublime_text": ">=3000",
-					"tags": true
-				}
-			]
-		},
-		{
 			"details": "https://github.com/SublimeText/WordHighlight",
 			"releases": [
 				{
--- a/repository/w.json
+++ b/repository/w.json
@@ -761,7 +761,8 @@
 		{
 			"name": "WordingStatus",
 			"details": "https://github.com/evandrocoan/WordingStatus",
-			"author": "evandrocoan",
+			"previous_names": ["WordCount"],
+			"author": ["titoBouzout", "evandrocoan"],
 			"releases": [
 				{
 					"sublime_text": ">=3000",
--- a/repository/w.json
+++ b/repository/w.json
@@ -773,7 +773,6 @@
 		{
 			"name": "WordingStatus",
 			"details": "https://github.com/evandrocoan/WordingStatus",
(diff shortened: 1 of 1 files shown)


repository/w.json
```diff
@@ -759,6 +759,18 @@
 				}
 			]
 		},
+		{
+			"name": "WordingStatus",
+			"details": "https://github.com/evandrocoan/WordingStatus",
+			"homepage": "https://github.com/evandrocoan/WordingStatus",
```
braver, line 765: That "homepage" is not necessary since it's just the GitHub repo url.

braver:
Why should this exist next to the existing WordCount package? Can't your rewrite be contributed back to that repository?

braver:
Oh, that's a "Tito" package, and it's archived... of course you can't contribute back. Shall we just replace the existing package with yours?

evandrocoan:
Some may like the original. My code was a complete rewrite, and the functional requirements should be backward compatible, but there is a non-functional requirement not respected. For big files, my version does not often refresh, while the original version refreshes the exact word counting within the same period.

braver:
Honestly I'd like to phase out dead and archived packages and yours offers the exact same functionality (with minor changes that sounds like improvements to me). What you could do is ship as a major version upgrade, and have an upgrade-message explaining the take-over, and steps users can take to manually install the old package (e.g. check out that repo in the packages dir).

braver:
Maybe one place where your package can be improved is in factoring out the Preferences class. I don't know why it exists in the original package. Sublime's interface to settings serves the values straight from memory, and updates the values immediately when they change: you cannot improve upon the performance of the API. And the interface also intelligently supports defaults. This would significantly reduce your lines of code, reduce potential misdirection, and probably make things like changing settings, view-only settings, and project-level settings work automatically and more reliably than your own implementation here.

evandrocoan:
The CI test failed but it does not look related to may changes: 
image

braver:
Your changes don't look good right now, you're touching more than just your package. 

There are two possible approaches:

1. We keep the old name, but update the entry to point to your repo (and maybe some other adjustments). 
2. We remove the old entry and add yours, effectively changing the name (doesn't seem necessary to me). You would also need to add a "previous_names" in your entry with the name of the old entry in it, to transfer current installations to your package. 

Approach 1 seems simplest to me.

evandrocoan:
As I already renamed the package, I am keeping the new name, and I added the "previous_names" key.

I also rebased with `develop` to try to fix the python error, but now looks like another package is not correctly sorted:			
image

FichteFoll:
Master was fixed, please rebase again.

braver:
Looks good, let's roll 👍🏻

PR #8632, opened 2022-11-03
Adding AskCodi extension data

assistivai:
- [ ] My package doesn't add key bindings.

I don't use git attributes since the project has no media files.

My package is AskCodi extension which
1. Converts Natural language (comments, statements etc.) to convert to code,
2. Generates documentation for the code,
3. has a manual keybinding trigger for code completion (manual autocomplete)
4. Generates unit test for the code,
5. Provides time complexity for the code,
6. Provides an explanation for the code.

Website: https://askcodi.com
Web app: https://app.askcodi.com

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -1874,6 +1874,17 @@
 			]
 		},
 		{
+			"name": "Ask Codi",
+			"details": "https://github.com/AskCodi/Sublime-AskCodi",
+			"author": "AskCodi",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "ASL Syntax Highlight",
 			"details": "https://github.com/Fansaly/ASL-Syntax-Highlight",
 			"author": "Fansaly",
--- a/repository/a.json
+++ b/repository/a.json
@@ -1874,9 +1874,9 @@
 			]
 		},
 		{
-			"name": "Ask Codi",
+			"name": "AskCodi",
 			"details": "https://github.com/AskCodi/Sublime-AskCodi",
-			"author": "AskCodi",
+			"author": "Assistiv AI",
 			"releases": [
 				{
 					"sublime_text": ">=3000",


repository/a.json
```diff
@@ -1873,6 +1873,17 @@
 				}
 			]
 		},
+		{
+			"name": "Ask Codi",
+			"details": "https://github.com/AskCodi/Sublime-AskCodi",
+			"author": "AskCodi",
```
braver, line 1879: If the author is just the GitHub username, you can leave it out here.

braver:
>  I have tagged a release with a [semver][2] version number.

Actually, no, you didn't ;)

braver:
Since your package seems to be sending code and other content to a remote server, potentially (or actually) tracking activities of uses, it would be a good idea to explain in detail explicitly how that works, what users should look out for with sensitive content (ie. when working on closed-source/proprietary code), etc.

assistivai:
Hi, can you please provide details on what is causing the issue for trigger-review?

braver:
This: https://github.com/wbond/package_control_channel/pull/8632#issuecomment-1337123457

assistivai:
I have added the tag as per with guidelines for the semantic version number. Is this enough for the trigger-review check?

assistivai:
I have changed the key bindings that throw the error. It should work without conflict now.

braver:
Please tag a release with you latest changes so the bot can verify.

assistivai:
tagged the release with latest changes.

braver:
Alright, thanks.

- It would be nice to have a setting to turn off the context menu, although I can see how selecting text for use with the AI, and then right-clicking makes sense for the most common workflow.
- It seems problematic to me that the settings have an API key in them already. I would assume users would need to provide their own API keys?
- There is no way for users to edit the settings (via the menu and the command palette)? That seems problematic to me, for the device_id for instance?
- Note that the settings file shipped with your package is not writeable. In fact, your package will be delivered "packed" (zipped really). If you need to write files somewhere (e.g. as a cache), you do so elsewhere. 
- I think it might be a good idea to more clearly and explicitly state that this tool is going to send code to the service. In many work environments it's not acceptable if that happens, for instance on accident. 
- It looks like you need to set up this tool for a certain project. People use ST for all sorts of stuff. You need to imagine someone might work on a project, then maybe write some documents, then maybe review some other unrelated project. Your package seems to authenticate as soon as ST starts. What happens if someone has multiple projects going with AskCodi? Have you considered using ST's project settings? You might want to explore turning your package on and off for different projects using ST's built-in cascade of settings.
- I'm not convinced every message from AskCodi needs an emoji. Imagine what it would look like if every package that someone uses was that colorful, it would get annoying soon.

assistivai:
1. It would be nice to have a setting to turn off the context menu, although I can see how selecting text for use with the AI, and then right-clicking makes sense for the most common workflow.

We have added a config file in the home directory for users to update the settings for context and individual features of AskCodi. The readme contains its details in "How it Works" and "Change configuration"

2. It seems problematic to me that the settings have an API key in them already. I would assume users would need to provide their own API keys?

The settings file is not to be edited by the user, apikey is default for all users and is used in server authentication, not user authentication.

4. There is no way for users to edit the settings (via the menu and the command palette)? That seems problematic to me, for the device_id for instance?

The settings file is not to be edited by the user, device id is unique and if edited by user, there will be authentication problems since 2 users can have same device id in that case.

5. Note that the settings file shipped with your package is not writeable. In fact, your package will be delivered "packed" (zipped really). If you need to write files somewhere (e.g. as a cache), you do so elsewhere.

We only write in settings file the first time authentication using sublime.load_settings(SETTINGS_FILE) and sublime.save_settings(SETTINGS_FILE). Users are not allowed to change these settings.

6. I think it might be a good idea to more clearly and explicitly state that this tool is going to send code to the service. In many work environments it's not acceptable if that happens, for instance on accident.

Mentioned in the readme "How it works"

7. It looks like you need to set up this tool for a certain project. People use ST for all sorts of stuff. You need to imagine someone might work on a project, then maybe write some documents, then maybe review some other unrelated project. Your package seems to authenticate as soon as ST starts. What happens if someone has multiple projects going with AskCodi? Have you considered using ST's project settings? You might want to explore turning your package on and off for different projects using ST's built-in cascade of settings.

Users can turn off individual AskCodi features already. Although AskCodi only works on manual triggering by user. Only authentication is an automated process which is required at the beginning to initialize AskCodi.

8. I'm not convinced every message from AskCodi needs an emoji. Imagine what it would look like if every package that someone uses was that colorful, it would get annoying soon.

Removed the emojis from most of the messages.

braver:
Alright, thanks for the explanation. 

You seem to be using the settings for something they’re not meant to do (cache information managed by the program), and then introduce a new configuration file to handle what the settings file should be doing. I would suggest you dig into some more documentation around these subjects, and maybe explore examples of existing packages. The only reason for sublime-settings files is for users to edit the file to store their preferences. Please reconsider how you’re approaching some aspects of the package with that as a starting point.

assistivai:
To clarify, should there be only one settings file that contains all the plugin information and user information and should be editable by the users and the plugin? Does this editable file needs to be edited via the menu or command palette, or is it okay if it is just editable by the user directly in the file?

braver:
User settings should be stored in a sublime-settings file in the User package directory and editable via menu and command palette. Sublime Text packages do not store anything in the user home directory, but in the Packages directory. You can store additional information in any kind of file you like, just don’t “abuse” the user settings pattern for anything non-editable.

braver:
Here’s an example of a sublime-settings file that stores additional data, overridable for user-specific data, besides the main user settings for the package:
https://github.com/sublimelsp/LSP/blob/main/language-ids.sublime-settings

Which are in a [separate file](https://github.com/sublimelsp/LSP/blob/main/LSP.sublime-settings). Ie. you can have multiple settings files and use the API’s to work with them. Users are always able to see and modify the files of course, they’re on their machine.

assistivai:
We have added the user settings to be editable via menu. Also added instructions in Readme under "Change user settings"
Now there is only one settings file.

braver:
Your menu item for settings still uses separate entries for default and user settings. We expect a single entry these days, that opens in split view. It’s also user friendly to document each setting using comments in the default settings file.

assistivai:
While testing, I only set user settings path. Are you saying that I need to open 2 settings, default and user, where default settings point to "{package directory}/AskCodi/AskCodi.sublime-settings" while user settings point to "{package directory}/User/AskCodi.sublime-settings"? When user clicks on Settings in Menu, both these files in split view?

braver:
Have a look at this for an example:

https://github.com/SublimeLinter/SublimeLinter/blob/a3d57f4a05799e0bbfbe13e0d71f7b3a1aa8229c/Main.sublime-menu#L18

assistivai:
Hi, I have default and user settings in split view for AskCodi. Also commented on the files for better understanding.

PR #8642, opened 2022-11-24
add yaral syntax package

duongbkak55:
- [x]  If my package is a syntax it doesn't also add a color scheme. ***

My package is highlighting for Yara L2 syntax, this was using for google cloud platform for create the rule.
My package is nearly to Yara Rule Syntax, but it should still be added because this is specific version of Yara into Yara L2 (https://cloud.google.com/chronicle/docs/detection/yara-l-2-0-overview), customize for using in google cloud rule.
Also, my package add some customize scope for clearly theme and special language.

[2]: https://semver.org

### DIFF
--- a/repository/y.json
+++ b/repository/y.json
@@ -78,6 +78,17 @@
 			]
 		},
 		{
+			"name": "Yara L2 Syntax",
+			"details": "https://github.com/appliedsciencegroup/sublimetext-package-yaral2-syntax-highlighting",
+			"labels": ["Yara", "Yaral" ,"Yara L2", "syntax", "snippets"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Yara Rule Syntax",
 			"details": "https://github.com/nyx0/YaraSyntax",
 			"labels": ["Yara", "syntax", "snippets"],
--- a/repository/y.json
+++ b/repository/y.json
@@ -80,7 +80,7 @@
 		{
 			"name": "Yara L2 Syntax",
 			"details": "https://github.com/appliedsciencegroup/sublimetext-package-yaral2-syntax-highlighting",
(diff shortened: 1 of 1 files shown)


repository/y.json
```diff
@@ -77,6 +77,17 @@
 				}
 			]
 		},
+		{
+			"name": "Yara L2 Syntax",
+			"details": "https://github.com/appliedsciencegroup/sublimetext-package-yaral2-syntax-highlighting",
+			"labels": ["Yara", "Yaral" ,"Yara L2", "syntax", "snippets"],
```
braver, line 83: `"syntax"` -> `"language syntax"` is the canonical label.

braver:
- Why does your package also include a color scheme? This is not necessary, if you follow the scope naming guide you never need to customize a color scheme to make it work.
- Your syntax alternates between `yara` and `yaral`, can you explain why?

duongbkak55:
> * Why does your package also include a color scheme? This is not necessary, if you follow the scope naming guide you never need to customize a color scheme to make it work.
> * Your syntax alternates between `yara` and `yaral`, can you explain why?

* I want to addition some color for highlighting some special syntax. If this is a requirement condition for package, I can using existed scope in default color schema.
* About my syntax alternates between `yara` and `yaral`, sorry for my mistake. Because YaraL2 is nearly to Yara Rule Syntax, so I using some rule base from that without replacement. I will correct it.

braver:
> If this is a requirement condition for package, I can using existed scope in default color schema.

Yes, it's unexpected for syntax package to require a specific color scheme (there are hundreds out there because people have their own preferences) or to modify any existing color scheme. For the same reason, plus people tend to use dozens of syntaxes.. you can imagine what a mess it would get if every syntax has its own opinion on the highlighting.

braver:
Yes, that looks good, thanks. Please change the “syntax” label to “language syntax”, and we’re good to go.

repository/y.json
duongbkak55: I changed syntax label to language syntax.
Thanks

PR #8643, opened 2022-11-25
Add Advanced Ansible Snippets

memoryleak:
- [ x] I'm the package's author and/or maintainer.
- [ x] I have have read [the docs][1].
- [ x] I have tagged a release with a [semver][2] version number.
- [ x] My package repo has a description and a README describing what it's for and how to use it.
- [ x] My package doesn't add context menu entries. *
- [ x] My package doesn't add key bindings. **

My package is similar to AnsibleSnippets. However it should still be added because it has a more complete list of snippets and is up to date with the new Ansible syntax.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -373,6 +373,17 @@
 			]
 		},
 		{
+			"name": "Advanced Ansible Snippets",
+			"details": "https://github.com/memoryleak/AdvancedAnsibleSnippets",
+			"labels": ["auto-complete", "snippets"],
+			"releases": [
+				{
+					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


braver:
> However it should still be added because it has a more complete list of snippets and is up to date with the new Ansible syntax.

Can't your snippets be added to that existing package?

memoryleak:
The other package follows a completely different syntax style, thus I don't think that it would make sense.

braver:
Alright, it’s not a terrible idea to have sets of snippets in different packages anyway 🤷🏼‍♂️

PR #8654, opened 2022-12-07
try to run ci with latest python

braver:
since 3.6 is missing

reference: https://github.com/actions/setup-python/issues/544

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v1
         with:
-          python-version: 3.6
+          python-version: 3.11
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -11,7 +11,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v1
```
FichteFoll, line 12:
```suggestion
        uses: actions/setup-python@v4
```
should get rid of the deprecation warning and the latest for `actions/checkout` is 3.

FichteFoll (approved):
Can be merged once tests passed :+1:

braver:
Thanks! Bumped those two things and all looks green.

PR #8655, opened 2022-12-08
Add Nim-lang package

vanyle:

My package is similar to Nimlime.
However it should still be added because NimLime is abandoned and does not support Sublime Text 4 (It's targeted towards Sublime Text 2). A lot of features like autocompletion, documentation generation and tooltips are missing or broken. Contributing to NimLime is non-trivial and making a new package is simpler.

[2]: https://semver.org

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -647,17 +647,6 @@
 			]
 		},
 		{
-			"name": "Nim-lang",
-			"details": "https://github.com/vanyle/SublimeNim",
-			"labels": ["nim", "language syntax", "auto-complete", "snippets"],
-			"releases": [
-				{
-					"sublime_text": ">=4000",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "NimLime",
 			"details": "https://github.com/nim-lang/NimLime",
 			"labels": ["nim", "language syntax", "completions"],
@@ -665,6 +654,17 @@
 			"releases": [
 				{
 					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
+			"name": "NimPlus",
+			"details": "https://github.com/vanyle/SublimeNim",
+			"labels": ["nim", "language syntax", "auto-complete", "snippets"],
+			"releases": [
+				{
+					"sublime_text": ">=4000",
 					"tags": true
 				}
 			]
--- a/repository/n.json
+++ b/repository/n.json
@@ -2,6 +2,17 @@
 	"schema_version": "3.0.0",
 	"packages": [
 		{
+			"name": "Nim-lang",
+			"details": "https://github.com/vanyle/SublimeNim",
+			"labels": ["nim", "language syntax", "auto-complete", "snippets"],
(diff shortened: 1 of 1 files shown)


braver:
Is NimLime truly abandoned? There seems to be a pretty big PR going to update it. Have you tried contributing to that effort?

vanyle:
The scope of this package is a bit different from what NimLime does. NimLime aims to provide basic editing support for Nim: syntax highlighting, and integration with the build system. While this plugin also has these features, the aim is a bit different.
The current pull request on NimLime aims to update the syntax to the current new features of Nim without adding features to the plugin.
The Nim compiler provides by default some tools (namely Nimsuggest and the `nim check` command) that allow for static analysis and fancy suggestions without needing an LSP.
Most sublime plugins (I think?) that support a language are designed to be simple and give this basic support as Sublime Text is not an IDE but a text editor.
This plugin gives the user the option to have these "IDE" like features if they need (all the features can be toggled in a config file).
There are some redundancy between this plugin and NimLime because I think it's easier for a user to only install a single package for a language that gives what they need (depending on if they want something closer to an editor or an IDE) then first installing a plugin for basic support and another plugin for "advanced" features.

AmjadHD:
@vanyle, Other than that PR, I was planning to rename and simplify the package. And make it work alongside nimlsp.
So I would prefer if you can help with this.

vanyle:
@AmjadHD I can do both!
I just like to work on my own package so that I can quickly add features and distribute them based on problem I encounter when writing Nim (currently, that probably means figuring out a way to add auto-imports even if it will probably we quite hard!), and without having to worry about handling contribution, reviews, etc...
I have plans to simplify the syntax to use snbf and if I do, I will probably add also submit PR to NimLime. I need to check how far you are on the simplification. I'm not really writing a lot of syntax highlighting as I find automatic documentation/error highlighting/suggestion more exciting but I noticed a few highlighting bugs that will need fixing at some point (even though they are not very bothering)

Also, I will probably change the name to Nimplus to avoid confusion. What name would you change NimLime to ? Just "Nim" ?

AmjadHD:
Just "Nim" since it's the official package, But I didn't discuss this with the owners yet.

braver:
So… management summary, we’re going with two packages, one official one that’s maybe going to be renamed, and another one with more features and higher velocity, and you’re looking to contribute back to the official package whenever that seems like a good idea? Works for me, just let me know when to push “merge”.

vanyle:
Yup, this is fine with me! I'm ready for the merge!

braver:
Alright! Almost there. You may want to update your readme, settings file and maybe some other places that have the name of the package in it: they currently don’t match the NimPlus name and that might confuse users.

vanyle:
Updated !

braver:
> changing the url will become more and more impractical

GitHub is pretty good at redirecting usually, but it's definitely nice to have these things aligned. If the bot is happy too I'll hit merge.

PR #8657, opened 2022-12-11
Add OpenAI package.

yaroslavyaroslav:
Tests are passed.

My package is client for OpenAI code completion(?) API, that is takes selected text within Sublime Text, sends it to the OpenAI server and then pastes the whatever comes from the server.

It could be any of the following:
- Appending of a given bunch of text
- Inserting new text within the selected code replacing the placeholder.
- Editing selected text by the command passed through input panel.

My package is similar [Codex](https://github.com/necarlson97/codex-ai-sublime). However it should still be added because it's providing supports of new API provided by OpenAI, while Codex supporting only the completion one.

[2]: https://semver.org

### DIFF
--- a/repository/o.json
+++ b/repository/o.json
@@ -680,6 +680,17 @@
 			]
 		},
 		{
+			"name": "OpenAI completion",
+			"labels": ["code completion", "openai"],
+			"details": "https://github.com/yaroslavyaroslav/OpenAI-sublime-text",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "OpenBazelBUILDFile",
 			"details": "https://github.com/macrael/OpenBazelBUILDFile",
 			"labels": ["file navigation", "bazel"],


yaroslavyaroslav:
@braver Just wonder whether it is time doesn't come yet, or do you have an issue to that package? I mean I see that there is a merged PRs after this one

braver:
Yeah, sorry, I’m hitting the low hanging fruit first, and saving this for when I actually have the attention span to do a proper review. I’m getting to it though, hang in there!

braver:
Your keymap file is empty, maybe just remove it. You can of course help users setting up bindings by providing an example in the readme.

I’ll quote this from another AI service package review:

> Since your package seems to be sending code and other content to a remote server, potentially (or actually) tracking activities of uses, it would be a good idea to explain in detail explicitly how that works, what users should look out for with sensitive content (ie. when working on closed-source/proprietary code), etc.

Finally, I wonder if the existing CodexAI package couldn’t be improved upon with your contributions. It has seen only 28 installations so far. We try not to have “competing” packages, and push contributors to work together as a community (ie. open source as it’s meant to be, not a dumping ground or a “marketplace”). I’m not against adding this package per se, but as an outsider to me it looks like it would be better to have one really good Codex implementation rather than several somehow overlapping ones.

braver:
> There's no placeholder within selected text, there's has to be exact one.

There's no placeholder within selected text, there's has to be *exactly* one.

yaroslavyaroslav:
> there are some other places where the language could be improved or clarified, I might send you a PR later

I'm struggling to get it, whether it's on me, or should I wait your PR that fix those?

braver:
> or should I wait your PR that fix those

don’t wait for me! 😄

yaroslavyaroslav:
> Your keymap file is empty, maybe just remove it. You can of course help users setting up bindings by providing an example in the readme.

Easy-peasy. I've deleted it just yet, but I'll push and release (by tagging such commit) a bit later, while I'm still in testing of reimplementing my last feature — provide Completion completion (sic) into output panel, rather handling the buffer selection. This one works some kind of `?` help in Jupyter notebook.

> > Since your package seems to be sending code and other content to a remote server, potentially (or actually) tracking activities of uses, it would be a good idea to explain in detail explicitly how that works, what users should look out for with sensitive content (ie. when working on closed-source/proprietary code), etc.

I'm good with elaborating those, but I'd need a bit clarification of your intent there. Do you mean that this disclaimer should be more either screaming or put in the very file beginning or more broad, like if you're developing some secret tool please do not use that, or all of them at once?

> Finally, I wonder if the existing CodexAI package couldn’t be improved upon with your contributions. It has seen only 28 installations so far. We try not to have “competing” packages, and push contributors to work together as a community (ie. open source as it’s meant to be, not a dumping ground or a “marketplace”). I’m not against adding this package per se, but as an outsider to me it looks like it would be better to have one really good Codex implementation rather than several somehow overlapping ones.

This one is a hard one. The plugin was as I said in the Readme, made by that OpenAI thing, like I was there just to debug the all the suggestions that it provided. The main intent of creating it was to let myself conveniently use of that OpanAI APIs in my daily workflow, and only then to bring it back to the community.

So the hard part here is that thanks to that OpenAI stuff I have the same lack of both Sublime Text a OOP Python API knowledge as I has had it before, and I'm happy with that either actually. More than that, if not Chat hype thing would have happened I'd never implement any for Sublime Text, due to lack of tutorials of its API usage. 

So to make long story shorts. I'd agree that the predecessor of that plugin are exist and even that it's actually maintained, but I myself are actually happy with my own implementation, coz it supports those flows that fits my needs. So I actually have no effort to contribute to that one because I have no effort to improve my Python OOP API knowledge which would be definitely required for such, but I definitely would be support mine despite would it be aded to Package Control or not.

braver:
> but I'd need a bit clarification of your intent there

It doesn't need to be screaming, but in many cases it's important to know that code is actually leaving the machine. It doesn't need to be "secret code", in many companies it's simply not acceptable. And it might not be expected behavior, I imagine many users will be new to the concept of using AI in this way and might not understand how the service you're using actually works. So it needs to be clearly stated in the package readme.

yaroslavyaroslav:
> So it needs to be clearly stated in the package readme.

I'm with you, like this is really not obvious thing for newcomers, but I'm needed a bit more concrete demand to the declaration of such in `Readme` above the given one. I mean, for now it's stated in two places, the **Usage** chapter and the **Note** chapter and bolded in both places, and there's no other bolded besides this one in whole readme.

So what room for improvement do you see here? Is it Note should be moved in the beginning of the file, should it be capitalized, or should it be both?

braver:
> So what room for improvement do you see here?

No, as it is it's totally ok 👍🏻

braver:
> Easy-peasy. I've deleted it just yet

The keymap file is still there, did you push your latest work?

yaroslavyaroslav:
> The keymap file is still there, did you push your latest work?

Yep, it needs a few more days to polish all changes where that deletion was just one of it.

yaroslavyaroslav:
@braver I've pushed it, now it's ready to be merged.

braver:
You need to tag a release with your latest changes. Package control will only pick up your tags.

PR #8658, opened 2022-12-11
Restore original Material Theme and remove dead packages.

equinusocio:

My package is the original Material Theme. Just restored it to remove the stolen copy published under SublimeText name.

[2]: https://semver.org

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -935,33 +935,11 @@
 		},
 		{
 			"name": "Material Theme",
-			"details": "https://github.com/SublimeText/material-theme",
+			"details": "https://github.com/material-theme/sublime-material-theme",
 			"labels": ["theme", "color scheme", "material"],
 			"releases": [
 				{
 					"sublime_text": ">=3103",
-					"tags": true
-				}
-			]
-		},
-		{
-			"name": "Material Theme - Appbar",
-			"details": "https://github.com/equinusocio/material-theme-appbar",
-			"labels": ["theme", "material design", "material theme"],
-			"releases": [
-				{
-					"sublime_text": ">=3103",
-					"tags": true
-				}
-			]
-		},
-		{
-			"name": "Material Theme - White Panels",
-			"details": "https://github.com/equinusocio/material-theme-white-panels",
-			"labels": ["theme", "color scheme", "material design"],
-			"releases": [
-				{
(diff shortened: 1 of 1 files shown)


FichteFoll:
> […] to remove the stolen copy published under SublimeText name.

This is not how open source licenses such as the MIT license work. The work is still clearly licensed to you and I merely restored it to a publicly available state, which is also explicitly mentioned at the top of the readme. I do realized that I did not consider to add an additional `author` key to the package entry to maintain the previous author name on the packagecontrol.io website, but that is a very quick thing to change. In fact, I will do it now because I don't care whether the GitHub organization name shows there or not. 

I'm not interested in maintaining the package. I simply fixed broken user setups and made a to this date very popular theme for the Sublime Text text editor installable again. However, it does not appear that you are willing to maintain the package either and from the past experience, I'd rather not run into the same or a similar situation again when the current situation works and is one that our community is in control of.

That said, I am very eager to know about the meaning and context behind the warning about "dangerous code and dependencies". I did have a quick look over the Python code in the package before reinstating it but did not notice anything relevant. I did not find anything in the repository's issues either and I wasn't able to inquire about it in the issue on the original repository where this was mentioned because it was locked.
If this is about the build toolchain used, then this is something we can document and/or remove, assuming there is an actual risk with people trying to modify the theme using these. However, before putting such a warning on the repo when it is irrelevant for most users who only want to use the package in its compiled form is probably not helpful overall.

equinusocio:
>I'm not interested in maintaining the package.
We know. But you could just contact us and fix it together instead of recreating a deleted repo and pointing the theme to a clone.

> That said, I am very eager to know about the meaning and context behind the warning about "dangerous code and dependencies"

The warning is about dependencies and cloning the repo on a local machine, is not about using the bundled version. Now, seeing the reaction to our move (which i remind is a free open-source project and we own the rights to delete it any time) we restored the repository under our scope, so thi PR points to an unmaintained (but working it seems) package.

I want to remind that Material Theme for ST was deprecated back in 2016, and did't get any update of fix since then. Is not our fault is Package Control rely only on repository URLs and doesn't handle broken themes errors. It should just disable the theme and show the error is something is wrong with the package.

equinusocio:
@jrappen done.

PR #8659, opened 2022-12-11
Change "Material Theme" author name(s)

FichteFoll:
This is a cleanup PR to address points noted in #8658.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -946,28 +946,6 @@
 			]
 		},
 		{
-			"name": "Material Theme - Appbar",
-			"details": "https://github.com/equinusocio/material-theme-appbar",
-			"labels": ["theme", "material design", "material theme"],
-			"releases": [
-				{
-					"sublime_text": ">=3103",
-					"tags": true
-				}
-			]
-		},
-		{
-			"name": "Material Theme - White Panels",
-			"details": "https://github.com/equinusocio/material-theme-white-panels",
-			"labels": ["theme", "color scheme", "material design"],
-			"releases": [
-				{
-					"sublime_text": ">=3000",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "Materialize",
 			"previous_names": ["Material Spacegray"],
 			"details": "https://github.com/saadq/Materialize",
--- a/repository/m.json
+++ b/repository/m.json
@@ -935,7 +935,7 @@
 		},
 		{
 			"name": "Material Theme",
-			"author": ["equinusocio"],
+			"author": ["equinusocio", "material-theme", "SublimeText"],
 			"details": "https://github.com/SublimeText/material-theme",
 			"labels": ["theme", "color scheme", "material"],
 			"releases": [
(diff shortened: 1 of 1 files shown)


FichteFoll:
> Packages removed:
>   - Material Theme
>   - Material Theme - Appbar
>   - Material Theme - White Panels

Well, that's a bug but doesn't matter here.

FichteFoll:
@equinusocio let me know if you want to see "equinusocio" (old repo location) or "material-theme" (new org name) here.

equinusocio:
We want it to point to [our official repository](https://github.com/material-theme/sublime-material-theme) because we may want to build new version at some point. Thanks. You can just accept my PR and if something is missing we can update the repo.

FichteFoll:
I understand that, but as I mentioned in your other PR I don't see a good reason to do that at the moment considering how you handled the previous situation. Saying you "may want to build a new version at some point" is a very vague promise if any that I do not feel inclined to put trust into. You are free to update me on this status once the new version has been worked on at which point I will reevaluate the situation and likely hand the package entry back to you, but in the current state you are not in a position to demand something like this from my point of view.

equinusocio:
This is insane. Are you telling us that anyone who published a package on Package Control is not free/the owner to move the project else where to make the repo private... my gosh.

The same would have happened if any repo is made private or moved way from github. Is a package control fail since it doesn't handle broken themes. Our next moves after deleting the repo had to be removing it from package control. What would have happened?

> but in the current state you are not in a position to demand something like this from my point of view.

This is Illicit appropriation.

FichteFoll:
> We just decided to delete OUR project from Github and Sublime Text, and you are forcing us to keep it just to no break the app (and we did it)

> This is Illicit appropriation, and we restored the deprecated repository JUST to make ST people happy. 

No, *I* did this. *I* restored your deprecated repository in accordance with its license to make ST people happy (or rather not make them unhappy because their setups broke after your repository has not been removed but *replaced*.

> This is insane. Are you telling us that anyone who published a package on Package Control is not free/the owner to move the project elsewhere or to make the repo private... my gosh.

Oh no, I'm not telling you that. What you describe here is a perfectly fine course of action that happens with packages for Sublime Text all the time, though we don't track these events. The problem here is that this is ultimately *not what you did* because you replaced the original repository with something that was incompatible with Sublime Text which is not something that any standard Package Manager could have handled differently.

> It seems you're just stealing money from our contributors and insulting the open-source and who worked for free to bring this theme to life.

I don't see how I or we are stealing money when your work was Open Source and not maintained for the past 6 years. Open Source works are available and obtainable *as is* and you're not doing anything with it (like providing support) either. Instead *I* was the one providing support for *your* theme out of my own volition and in my free time because of my appreciation of Open Source.

I should clarify that I am *not* affiliated with Sublime Hq Pty Ltd and I do not get paid for doing any of this. Considering [your blog post](https://dev.to/equinusocio/why-you-should-avoid-sublime-text-2c73), you anger against the Sublime Text authors seems misdirected.

jfcherng:
This PR causes a weird situation that `equinusocio` doesn't actually own any control to the repository but it's the only contact on the list.

FichteFoll:
I think I struck a decent balance with the last commit to just add everything.

PR #8660, opened 2022-12-14
Add WSL Build package

existentialmutt:
- [X] My package doesn't add context menu entries.
- [X] My package doesn't add key bindings.

My package is WSL Build, an ExexCommand for creating build systems for projects running in WSL2

This is my first published Sublime Package so feedback is appreciated :)

Thank you!

### DIFF
--- a/repository/w.json
+++ b/repository/w.json
@@ -1119,6 +1119,19 @@
 			]
 		},
 		{
+			"name": "WSL Build",
+			"details": "https://github.com/existentialmutt/wsl_build",
+			"description": "Create custom build systems for projects running in WSL2",
+			"labels": ["build system"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true,
+					"platforms": ["windows"]
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


repository/w.json
```diff
@@ -1118,6 +1118,19 @@
 				}
 			]
 		},
+		{
+			"name": "WSL Build",
+			"details": "https://github.com/existentialmutt/wsl_build",
+			"description": "Create custom build systems for projects running in WSL2",
```
braver, line 1124: This seems to match your GitHub repo description, so you can just remove it here (package control will find it in the repo).

braver: So, please remove that so we can wrap this up and ship your package!

deathaxe:
1. The build command works only for Ubuntu distros and files located within WSL environment in its current state. 
2. It does also not opt-in to python 3.8, which each new plugin should do to be accepted.

_Filed issues about it at the repo._

Until those are fixed it is probably not worth adding to packagecontrol index.

braver:
Please let us know if you want to re-open this PR and continue the review.

PR #8662, opened 2022-12-16
Add Plugin Logger

kapitanluffy:

My package is Plugin Logger

[2]: https://semver.org


braver:
Ok, that’s maybe a bit niche, but cool. I’d prefer it not to print anything unless in a debug mode though (on load/unload).

kapitanluffy:
Do you mean a `debug` setting? I made it so that it would use the default `_LogWriter` when the plugin is disabled in package control though.

braver:
I meant for instance [print("Plugin Logger plugin loaded")](https://github.com/kapitanluffy/sublime-plugin-logger/blob/master/plugin_logger.py#L61), that line should be removed, or put behind a debug setting. Preferably packages don’t log anything during normal operation, and ST already has its own reporting about the loading/unloading of packages.

kapitanluffy:
I see. I just removed it since it was unnecessary actually.

PR #8663, opened 2022-12-17
Add Just package

nk9:

My package is a syntax definition for the [Just task runner](https://just.systems). It has comments and a simple build system too.

[2]: https://semver.org

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -1413,7 +1413,7 @@
 			"labels": ["language syntax", "build system"],
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": ">=4075",
 					"tags": true
 				}
 			]
--- a/repository/j.json
+++ b/repository/j.json
@@ -1409,10 +1409,7 @@
 		},
 		{
 			"name": "Just",
-			"author": "nk9",
 			"details": "https://github.com/nk9/just_sublime",
-			"readme": "https://raw.githubusercontent.com/nk9/just_sublime/main/README.md",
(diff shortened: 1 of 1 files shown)


braver:
- Why do you have both uppercase and lowercase syntax files?
- The formatter functionality might be better (more flexible etc) supported by the [Fmt package](https://packagecontrol.io/packages/Fmt) (although I use similar build systems for this all the time myself). But you might be interested in documenting a setup for Fmt.
- I believe the package assumes just already being set up on the user's system. Users just starting out with just and ST might need a little help setting it up. Ie. you might want to clarify that the package doesn't ship with just and it needs to be in the user's path.

nk9:
Thanks for the review!

- I've added the `LICENSE` file back to the package. I was excluding it in the `.gitattributes` file.
- Good catch on the doubled syntax files, I've fixed that.
- I wasn't aware of Fmt, it looks useful! As it stands, though, the formatter is still unstable. There is actually [a bug at present](https://github.com/casey/just/issues/1449) which will cause it to generate code which won't parse correctly. (Fixed, but not yet released.) So I think I will wait to mention this anywhere.
-  Note that the Build System included with this package doesn't actually change the file. Rather, it produces a reformatted version in memory and compares that to the version on disk. It also parses the file and does a bit of static analysis to make sure that there are no syntax errors.
- The Build System does assume that just is installed and in the PATH, although of course the syntax works regardless. I've incldued a bit about this in the install message.

braver:
🎉

nk9:
Thanks for your help!

PR #8665, opened 2022-12-20
Add Theme Treble

53v3n3d4:

My package is ... theme for Sublime Text 4

[2]: https://semver.org

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -2113,6 +2113,17 @@
 			]
 		},
 		{
+			"name": "Theme - Treble",
+			"details": "https://github.com/53v3n3d4/Theme-Treble",
+			"labels": ["theme"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Theme - TwoDark",
 			"details": "https://github.com/erremauro/TwoDark",
 			"labels": ["theme"],


braver:
- You typically name your settings file after the package, that makes the user's customizations easier to manage. Not sure how that works for themes exactly though?
- Also, some settings like the tab style might need some additional explaining?

53v3n3d4:
Hi braver,

> * You typically name your settings file after the package, that makes the user's customizations easier to manage. Not sure how that works for themes exactly though?

You are referring to Preferences.sublime-settings file? I followed [Daneo Theme](https://github.com/SublimeText/Theme-DAneo)

> * Also, some settings like the tab style might need some additional explaining?
I will add this description below, is this good?

```
    // Tabs style options: "do", "re", "mi" and "fa".
    // Options "do" and "re" can have a background color. While "mi" and "fa" 
    // always have a transparent background. Which is set by features like
    // tabset_background or adaptive_contrast.
    // Inactive sheet dimming is supported on option "do" with a background 
    // color.
    // Option "mi" and "fa" have selected tab button that follow accent-color 
    // feature.
    // A minimal layout can be achieved with option "do" with tabset_background
    // feature false (Dark and Light themes) or adaptive_contrast feature false
    // (Adaptive theme).
```

PR #8666, opened 2022-12-20
Add Rome Formatter package

marekpiechut:

My package is adding support for Rome code formatter for Javascript and Typescript.

[2]: https://semver.org

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -1837,6 +1837,17 @@
 			]
 		},
 		{
+			"name": "Rome Formatter",
+				"labels": ["javascript", "typescript", "formatting", "format"],
+				"details": "https://github.com/marekpiechut/sublime-rome-formatter",
+				"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "RON",
 			"description": "Rusty Object Notation syntax scheme",
 			"author": "kvark",


braver:
Are you familiar with the Fmt package, and would this formatter work with that?

marekpiechut:
> Are you familiar with the Fmt package, and would this formatter work with that?

Hmm, it looks like basic functionality would work. But I think it would not support formatter config and binary auto discovery. Binary can and usually will be in project `.node_modules/.bin` folder and config file has to be exactly under `cwd` for it to be used by binary.

braver:
Yes, I can see how a specific package can have more features like that.

One thing I’d like to change is the key bindings. Because packages can override each other, it’s better if packages don’t ship with their bindings enabled. A common approach is to comment out the bindings, and provide menu and command palette entries that allow users to edit their bindings in split view. It’s then very easy to copy over the binding for your package, and maybe tweak the key combination. That also has the benefit of making the binding easier to remember.

marekpiechut:
You’re right. I think I’ll remove the key biding and just add it to Readme. Most users will probably want to remap it anyway. Give me 15 minutes to create a new release.

rchl:
Maybe also try to clarify the `Dependencies` section which says:

> Plugin needs nothing. If you don't have Rome yet installed: [install instructions]

This seems to contradict itself. Your plugin might not require *ST dependencies* but for casual user this doesn't mean anything and, had it been required, it would be handled automatically anyway.

If we are talking high-level it seems to me like the rome dependency is required.

rchl:
I also see that Rome implements LSP server so this could also be handled as a LSP plugin but we should probably not hold that against this package.

marekpiechut:
> Maybe also try to clarify the `Dependencies` section which says:
> 
> > Plugin needs nothing. If you don't have Rome yet installed: [install instructions]
> 
> This seems to contradict itself. Your plugin might not require _ST dependencies_ but for casual user this doesn't mean anything and, had it been required, it would be handled automatically anyway.
> 
> If we are talking high-level it seems to me like the rome dependency is required.

Yeah, haven't thought about it this way. But you're right, someone might think that Rome itself is included in the plugin. I've updated the readme.

rchl:
It still says "Dependencies: None, Plugin itself has no dependencies,..."

I think this still talks about ST dependencies but, as I said, this wouldn't be relevant to the user even if there would be dependencies because the user wouldn't have to do anything about that anyway -- it would be handled automatically.

I would suggest just treating "rome" as a dependency and make it explicit. Or instead you could name the section "requirements" if you think that "dependencies" is too ambiguous.

marekpiechut:
> It still says "Dependencies: None, Plugin itself has no dependencies,..."
> 
> I think this still talks about ST dependencies but, as I said, this wouldn't be relevant to the user even if there would be dependencies because the user wouldn't have to do anything about that anyway -- it would be handled automatically.
> 
> I would suggest just treating "rome" as a dependency and make it explicit. Or instead you could name the section "requirements" if you think that "dependencies" is too ambiguous.

Updated the readme, hope it makes it more obvious what has to be done. Please check it out.

braver:
Looks good to me!

PR #8667, opened 2022-12-21
Adding Pyisort to the list of repositories

returnchung:

My package is Python [isort](https://github.com/PyCQA/isort) formatter for Sublime Text.

My package is similar to `isort` and `isorted`. However it should still be added because `isort` is out of date for several years and all of them lack the flexibility for sublime settings or specific project settings. Besides, there are more supported isort command options can be configured and extended in the future.

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -2947,13 +2947,10 @@
 		{
 			"name": "Pyisort",
 			"details": "https://github.com/returnchung/pyisort",
-			"issues": "https://github.com/returnchung/pyisort/issues",
-			"author": ["returnchung"],
 			"labels": ["formatting"],
 			"releases": [
 				{
 					"sublime_text": ">=4000",
-					"platforms": "*",
 					"tags": true
 				}
 			]
--- a/repository/p.json
+++ b/repository/p.json
@@ -2945,6 +2945,20 @@
 			]
 		},
 		{
+			"name": "Pyisort",
+			"details": "https://github.com/returnchung/pyisort",
+			"issues": "https://github.com/returnchung/pyisort/issues",
+			"author": ["returnchung"],
+			"labels": ["formatting"],
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"platforms": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"details": "https://github.com/biermeester/Pylinter",
 			"labels": ["linting"],
 			"releases": [
--- a/repository/p.json
+++ b/repository/p.json
@@ -2950,7 +2950,7 @@
 			"labels": ["formatting"],
 			"releases": [
 				{
-					"sublime_text": ">=4000",
+					"sublime_text": ">=3000",
 					"tags": true
 				}
 			]


braver:
Sounds to me like we should try to find a way to upgrade the existing isort package. Can you not collaborate with @asfaltboy on that package? Or, maybe he's ok with us replacing that package with yours (assuming that for end-users it's drop-in replacement).

repository/p.json
```diff
@@ -2944,6 +2944,20 @@
 				}
 			]
 		},
+		{
+			"name": "Pyisort",
+			"details": "https://github.com/returnchung/pyisort",
+			"issues": "https://github.com/returnchung/pyisort/issues",
+			"author": ["returnchung"],
+			"labels": ["formatting"],
```
braver, line 2952: The author, issues, and platforms entries can be removed as they're just the defaults.

asfaltboy:
> Sounds to me like we should try to find a way to upgrade the existing isort package. Can you not collaborate with @asfaltboy on that package? Or, maybe he's ok with us replacing that package with yours (assuming that for end-users it's drop-in replacement).

Hi 👋 happy holiday season everyone!

The [returnchung/pyisort](https://github.com/returnchung/pyisort) package seems a lot like [rimvaliulin/isorted](https://github.com/rimvaliulin/isorted); in that it uses a subprocess approach, unlike the package I maintain which vendors in isort lib, which limits it to an older version of isort, one which ST's built-in python version supports. In general, I agree that the subprocess approach is more flexible and should probably be used for better forward compatibility support. Lastly, I'd say pyisort is most similar to isorted, and could be considered an in-place/modern replacement for isorted for ST4 users.

From a quick review of the pyisort package, I can see some cons and pros vs isorted:

Cons:
* pyisort requires ST 4, while the other two packages allow older versions of ST
* Does not work with python 3.3 which seems to be the default when I check out the repo - IIRC this can be fixed by committing [a `.python-version` file like here](https://github.com/SublimeText/UnitTesting/blob/master/.python-version) (there may be a better way, but I can't recall and struggling to find the right doc).

Pros:
* pyisort has a neater README, that does not copy paste the isort README (this is one of the first things I changed when I originally forked [thijsdezoete/sublime-text-isort-plugin](https://github.com/thijsdezoete/sublime-text-isort-plugin)).
* pyisort uses (some) type hints, and the code is easier to read in general

I think we could do a few things:
1. Leave one or both of the existing packages as they are, for users stuck on older sublime versions
2. Replace isorted with this package, and refer users to the package I maintain ([asfaltboy/sublime-text-isort-plugin](https://github.com/asfaltboy/sublime-text-isort-plugin) when they need an older version) - would love for us to also get @rimvaliulin's feedback on this one
3. Fix pyisort package to be backward compatible with older versions of ST (e.g support Python 3.3), and deprecate the 2 existing packages. Requires a bit more testing on ST3, and documenting a "migration path" for users

P.s: I'm happy to jump on a hangout/discort/etc to chat if we can organise a time that fits all our timezones (I'm in GMT)

returnchung:
Hi @asfaltboy, thank you very much for kindly reply.
About Cons and following feedbacks, I'd loved to adjust for compatible with pyhton 3.3 & 3.8 and add more support options if they're important for general usage in the future. Let me take some time to do that, and let you know when I finished. :)
Basically, unless you want to use the sublime package settings or specific project settings to overwrite, the isort always find the closet config to run.

returnchung:
Hi @asfaltboy, 
I've removed the dependency for sublime plugin 3.8 (and python 3.8 features in code) and now it's compatible with sublime plugin 3.3 too.
Besides, I've also added the `multi-line` isort options, please give another try and let me know if any further concerns. Thanks :).

https://github.com/returnchung/pyisort

rimvaliulin:
Hi @asfaltboy, sorry for the late reply. I think there no reason tor replace my package with @returnchung's,
I don't see any further benefits in that. Compatibility changes already maked to pyisort to conform Python 3.3 (removing type hinting, ... etc) takes off cons except a neater README (that can be improved).

braver:
I guess I focused on isort initially, but the same goes for isorted: if this package already exists and provides, from the end user perspective at least, identical functionality, why should pyisort be added? Seeing as maintainers of all mentioned packages are active on GitHub and in this thread, could you not make a contribution to those existing packages?

returnchung:
@braver @rimvaliulin There are some features I really need,
1. Flexible configs in sublime project and package settings can be applied to the isort command,
2. Support more isort options we can config in settings.

Initially, I don't want to replace anyone's package, just offer more help for community.
If we can do these in the `isorted`, I'm ok with you to not to add this to package control. Instead, I'll contribute to the `isolated`.

braver:
We try to push for people to contribute to each others packages. Otherwise the eco system would just fragment further and further, with one variation of a package having maybe one set of features someone needs, and another having another set of somewhat overlapping features. So if these features you mention are generally useful for people using isort in the context of Sublime Text, and isorted mostly aligns with the implementation you have in mind… let’s start with a PR there. If that doesn’t work out we can always add another package.

repository/p.json
thecontemoraryflorals: +2950

returnchung:
Thanks for your explanation, and I'm appreciated your efforts for this eco-system. I'll start an issue for `isorted` and let's go for there.

braver:
Alright, I’ll close this PR as we can’t do anything here right now. We can always re-open later if we want to.

PR #8669, opened 2022-12-30
Update Ruby Slim for Sublime Text 4107+

deathaxe:
This PR re-targets "Ruby Slim" package to SublimeText organization. The repository is a fork of original ruby-slim.tmbundle to backup its history and provide backward compatible historic data for ST2 and 3.

The main purpose however is to provide a completely rewritten syntax definition for Sublime Text build 4107+, which is in no form compatible with or related with the original package.

As such...
   _Tests and images are not excluded per design so external syntax test runners can download the archive and include tests. The preview image is included to be locally available for "Markown Image" package._

An issue to ask permission for re-targeting the package has been filed at https://github.com/slim-template/ruby-slim.tmbundle/issues/84. As the last commit in that repository dates back to 2018, I don't however expect any response. Would be glad about it though.

[2]: https://semver.org

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -2265,12 +2265,20 @@
 		},
 		{
 			"name": "Ruby Slim",
-			"details": "https://github.com/slim-template/ruby-slim.tmbundle",
+			"details": "https://github.com/SublimeText/Slim",
 			"previous_names": ["ruby-slim.tmbundle", "ruby-slim"],
 			"releases": [
 				{
-					"sublime_text": "*",
-					"branch": "master"
+					"sublime_text": "<4107",
+					"tags": "st2-"
+				},
+				{
+					"sublime_text": "4107 - 4130",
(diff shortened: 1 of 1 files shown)


braver:
👍🏻 @auvipy seems to be quite active on GitHub, so let’s give it a few weeks.

braver:
Feedback seems to be positive, I don’t think we need more permission than that. How about you?

deathaxe:
I am fine with it.

auvipy:
what you want from my side to do?

deathaxe:
Basically nothing else than agree with moving the repo link to SublimeText organization. We don't just want to overtake the repo without asking for permission. A new repo probably makes sense as the new revision is for ST only, while the original repo seems more for TextMate.

Once the move is done a hint at original repo about the new destination might help those who are used to the old location.

PR #8671, opened 2022-12-31
Renamed Cool & Clear for better discoverability

Doodpants:
This change just modifies the entry for the "Cool & Clear" package, to rename it to "Cool & Clear - Color Scheme", so that it shows up in the results when people search for "color scheme" in Package Control. It also shortens the description text a little bit.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3607,7 +3607,6 @@
 			"name": "Cool & Clear - Color Scheme",
 			"previous_names": ["Cool & Clear"],
 			"details": "https://github.com/Doodpants/Cool-and-Clear-Color-Scheme",
-			"description": "A tastefully minimal, high contrast, light-mode color scheme",
 			"author": "Karl von Laudermann",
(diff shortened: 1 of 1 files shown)


repository/c.json
```diff
@@ -3604,8 +3604,10 @@
 			]
 		},
 		{
-			"name": "Cool & Clear",
+			"name": "Cool & Clear - Color Scheme",
+			"previous_names": ["Cool & Clear"],
 			"details": "https://github.com/Doodpants/Cool-and-Clear-Color-Scheme",
+			"description": "A tastefully minimal, high contrast, light-mode color scheme",
```
braver, line 3610: We advocate keeping the description out of the repo here, so that we don't need to handle an PR every time someone wants to change it. Also in this case, sure you can strip "for Sublime Text" in this context, but it's not really a problem to leave it in.

braver (changes requested):
The name change is ok, but let's leave the description out of your repository entry.

PR #8673, opened 2023-01-06
Add Switch Window plugin

deathaxe:

The name "Switch ..." is inspired by "Quick Switch Project..." and enables a user to quickly navigate between multiple Sublime Text windows's no matter whether they contain a project or have only a simple scratch buffer open for note taking.

A similar function is provided by GotoWindow.

However it doesn't share any line of code and uses pure ST API instead of OS API hacks to perform its job. Hence it works on ST4 only. Such basic usability plugins should be organized within SublimeText organisation, IMHO.

We could also try another takeover, but well, I don't want to maintain or bother with that legacy code of the other plugin, while being able to easily and quickly perform updates on my own.

I am still a bit uncertain about the better name "Switch Window" vs. "Goto Window". In case a native english speaker may judge the latter a better choice, ... . Not sure.

[2]: https://semver.org

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -5669,6 +5669,17 @@
 			]
 		},
 		{
+			"name": "Switch Window",
+			"details": "https://github.com/SublimeText/SwitchWindow",
+			"labels": ["navigation"],
+			"releases": [
+				{
+					"sublime_text": ">=4107",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Symfony2 Override",
 			"details": "https://github.com/igormukhingmailcom/sublimetext3-symfony2-override-package",
 			"releases": [
--- a/repository/s.json
+++ b/repository/s.json
@@ -5635,6 +5635,17 @@
 			]
 		},
 		{
+			"name": "Switch Window",
+			"details": "https://github.com/SublimeText/SwitchWindow",
+			"labels": ["navigation"],
+			"releases": [
+				{
+					"sublime_text": ">=4107",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "SwitchDictionary",
 			"details": "https://github.com/Naereen/SublimeText3_SwitchDictionary",
 			"releases": [
@@ -5664,17 +5675,6 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"tags": true
-				}
-			]
-		},
-		{
-			"name": "Switch Window",
-			"details": "https://github.com/SublimeText/SwitchWindow",
-			"labels": ["navigation"],
-			"releases": [
-				{
-					"sublime_text": ">=4107",
 					"tags": true
 				}
 			]


braver:
“Switch” makes more sense to me than “Goto” for this functionality. Like you say it kinda matches the project switching feature.

Your package looks like a newer, better implementation of the exact same functionality provided by the existing package. That project seems a bit stale to me. It’s also not a super popular package. There doesn’t seem to be any reason to prefer the old package over yours. So, I’d like to consolidate the two packages in some way. Or at least discuss possibilities with the maintainers of the existing package.

deathaxe:
Basically we could create a fork or a PR to OpenWindow to rename commands and package to SwitchWindow and conserve the otherwise unchanged state for ST3, while using the content of current repo for ST4. It's probably a common strategy for such plugins. I'd however like the idea of gathering such plugins under SublimeText org.

braver:
IMO the old version could just be in a separate branch, doesn't even need to have a shared history with your work. Put that in a branch, tag a st3- release and be done with it. We would replace the entry with yours, rename it at the same time, and have the work in the org (more stuff needs to be in that org, I agree). 

Of course @ccampbell needs to agree with a takeover like that. 

In the end it's not terrible to have competing packages... but I feel our approach here is a strong USP of this eco-system over the uncontrolled chaos of dozens of abandoned duplicates in the Atom and VSCode equivalents. So IMO it's worth a bit of effort and gentle pushing to leave egos out of it and sometimes replace older projects with new ones if the rewrite is so thorough a PR won’t work.

So let’s see if we can get some feedback on that and decide in a week or so.

ccampbell:
Thanks for the ping. FWIW, I use my plugin literally hundreds of times a day lol, and I am using Sublime Text 4 as well. The only reason I never updated it to work with the native sublime APIs for switching windows was cause they never worked for me. If the new plugin can do everything mine can, and/or can do it better, then I think the branch idea is fine, but I just wanted to add that I don’t think the “bothering with that legacy code” actually would amount to that much work since it really hasn’t changed in a long time. I already have checks in the code for different sublime versions, so adding custom code for version 4 shouldn’t be much work.

May I ask why you decided to rewrite it instead of submitting a PR to the existing plugin?

Also, I just went to try out your plugin, and for some reason I can’t get it to run. I get this in the sublime text console on load when I add the `Switch.Window.sublime-package` to the Installed Packages directory. Any ideas?

```
reloading python 3.3 plugin Switch.Window.switch_window
  File "/Applications/Sublime Text.app/Contents/MacOS/Lib/python33/sublime_plugin.py", line 308, in reload_plugin
    m = importlib.import_module(modulename)
  File "./python3.3/importlib/__init__.py", line 90, in import_module
  File "<frozen importlib._bootstrap>", line 1584, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1565, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1512, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 313, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1584, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1565, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1512, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 313, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1584, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1565, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1529, in _find_and_load_unlocked
ImportError: No module named 'Switch'```

deathaxe:
> May I ask why you decided to rewrite it instead of submitting a PR to the existing plugin?

The code was written in about 10mins to answer https://forum.sublimetext.com/t/one-sublime-window-to-rule-them-all/66476/4

I rewrote it as I saw "GotoWindow" to use various OS APIs and external program calls to do its job, while just using `window.bring_to_front()` does needed. It works with fullcreeen apps and accross virtual desktops on Windows and Linux.

Hence there's no justification for external dependencies anymore. 

see also: https://github.com/sublimehq/sublime_text/issues/444#issuecomment-616082677

Known effort to prepare PRs (including investigating state of the art code) and even this post to justify my decitions wastes much more time - even days with taking all discussions into account.

I have rather given up on writing PRs for existing ST 3rd-plugins as most of them have just been ignored so far. The one way or the other - the decision is always wrong.

>  Have you tested it in macOS?

Not personally, as I don't have access to a Mac. I got some positive feedback about the plugin from users I know they use Mac however.

This feedback, was what made me try to publish that work.

> I get this in the sublime text console on load when I add the Switch.Window.sublime-package to the Installed Packages directory. Any ideas?

The plugin requires ST4. It uses python f-strings which require python 3.8 plugin_host and `bring_to_front()` API call which is available on ST4 only. 

Your console output indicates it running on python3.3 plugin host even though a valid `.python-version` file to opt-in to python 3.8 is present.

Did you try to restart ST? I've seen such kind of messages when hot-reloading plugins various times. ST's modules importer seems to have some issues when reloading plugin code from sublime-package files.

The name containing space should not be an issue as there are other packages which do so. 

Note: I just see the filename containing a `.` between `Switch` and `Window`. It should be a space. Seems to be caused by Github. That may be the reason as well.

deathaxe:
Please don't misunderstand my statement. I really appriciate your efforts and don't want to steal any ideas or dis any achievements. I am happy to contribute to the existing plugin, if it is still actively mainteined and used.

With several experiences about droped repos, I just got addicted to the idea of organizing all such kind of small useful high quality plugins, which add some kind of core functionality, within the SublimeText organization instead of personal repositories.

It is to make sure they are not deleted out of the sudden and to state out they are supported by the core community. Each author can be given full access to its repos within the org.

braver:
Thanks for all the discussion here guys. I too have seen too many PRs on packages go to waste because the maintainers aren’t actually interested anymore, so it’s great to see some actual interaction here. 
All I have to add is that I agree that more packages should be owned by the community as a whole instead of by individuals. We did a similar push in that direction for SublimeLinter and it has made it possible to keep more packages up to date on changes in the API’s, and fix bugs, with less effort. Maybe that’s besides the point of what to do with this particular package, but I applaud the suggestion at least.

ccampbell:
I have no objections to handing the plugin to the community to maintain. Honestly, my bigger concern is that there may be some regressions or things that don’t work quite the same in the new version on Mac since the plugin is too essential to my workflow right now 😅. I understand perhaps my version has some issues on Windows, but I am also happy to work with you to get those resolved. Like many of my other projects, it was never my intention to let this plugin go stale, but there are only so many hours in the day, and I don’t have the time or energy to focus a lot of attention on it (especially when I don’t have a windows machine to even test the changes on). That also lines up with the idea of giving the project to the community. 

I will try again to get your version to run in a little bit. I _am_ using Sublime Text 4 (dev channel build 4136, to be precise) so I think it should be using the proper python version, but perhaps the naming thing you pointed out is causing problems.

I guess I am not sure the proper path forward. I don’t really see the benefit in having a separate codebase for the same functionality in Sublime Text 4 (when 2, 3, and 4 _should_ already be supported in my version), but I don’t know how much work it would be to merge our changes (or implement yours on top of mine). Ideally we could consolidate them and have one plugin that works in all Sublime Versions (even if it means using hacks in some cases, as mine does). I also agree that Switch Window is a better name. I probably only called it Goto Window because that matches the other Goto menu options in ST.

image

deathaxe:
> ... regressions or things that don’t work quite the same in the new version on Mac

New code base should not be less functional. With `window.bring_to_front()` being available responsibility to bring desired windows to front should just be moved to core ST. 

> ... perhaps the naming thing ... is causing problems.

It's definitely the reason. I get the same errors on Windows, when replacing the space by period. My hope was Github using filename matches its download caption. Unfortunatelly that's not the case.

Maybe that's a compelling reason to call it SwitchWindow.

> I am using Sublime Text 4 (dev channel build 4136

I've tested the new code base with ST4107 successfully on Win/Linux.

> I don’t really see the benefit in having a separate codebase for the same functionality in Sublime Text 4

New code uses input handlers to integrate the command more seamlessly into Command Palette and kind info to display the kind/content of target windows.

It also supports empty windows, windows with unsaved files or those which only have files open. None of those windows are supported by current code base. Adding that support would be what I've called "bother with legecy code base" aka. rewrite big parts of the plugin in a hacky (ST2) and 3 compatible way while needing to test on all old ST versions.

Even if it would be possible to hack everything into a single plugin and run code paths depending on ST version, that's not what I want to do with ST2/ST3 not being actively supported anymore. The idea is just to concerve current or slighly updated code base in a separate path with the option to further improve it for ST3 users, but start from scratch for ST4.

Btw. by using `subprocess.check_output()` your plugin requires ST3 at least on MacOS. This function is not supported by python 2.6 (ST2)

I've made some efforts to 
1. prepare a possible move to SublimeText org (adjusted links in readme), 
2. rename the plugin to "Switch Window" 
3. add the working Windows related fixes from ehuss
4. and move OS related code into dedicated modules (probably requires some tests to avoid regressions).

But still, I'd use rewritten code for a dedicated ST4 branch and make sure everything works as before from there on.

In case you are not opposed to hand it over to SublimeText org, I'd rename/remove the current repo and give you full access to the handed over repo afterwards, so nothing would change for you.

> I probably only called it Goto Window because that matches the other Goto menu options in ST.

I can fully follow that and it was my first idea, too. I've now added a "Switch Window..." menu item to `Main > File` just next to `New Window` and `Close Window` command. But well, the goto menu would be as well suited.

ccampbell:
Okay, I am finally testing out your plugin. I am noticing that it doesn’t quite work correctly in some cases. I haven’t quite found the common denominator, but for a number of windows, if you type the name of the directory itself, it does not match in the list. You have to type the name of the open file in that directory. For example, check this screenshot:
image

So my directory here `foxy-c` will not complete. Whereas in my plugin it comes up by typing the name of the directory:
image

Empty here unless you type the name of the file which is not super great cause at least for me, I never know which file I have open in a given project when I want to switch to it: 
image

image

ccampbell:
Also, good call on `subprocess.check_output()`. It wasn’t actually added until recently, so this commit likely broke compatibility with Sublime Text 2 then: https://github.com/ccampbell/sublime-goto-window/commit/fe23342f896cac3b34b9a78256df9e822b52df91, but should be an easy fix

FichteFoll:
> Wouldn’t it make things a bit simpler to just take the current code and put it in a `version-2-and-3` or `legacy` branch or something rather than potentially introduce regressions by changing the file/code structure?

Yes, imho. I would move the current code into a legacy branch and push a new fresh branch with the changes from the new package as the new default branch (main or master).

> Finally, what will this change look like for existing `GotoWindow` users? Will this end up as a breaking change for anyone currently using the plugin? Will it just stop working if they already have it installed?

The exact change remains to be decided on, but in general the idea is to "update" installations on compatible ST build to the new branch and leave all other environments untouched and use the legacy version.

ccampbell:
> This legacy version can also still be updated to fix problems as it would be using a separate release channel from the ST4 version (using a different prefix for the tags).

I guess I’m a bit confused on this mainly because the version I have still works on ST4. So, for example, I’m curious if someone has the `GotoWindow` plugin installed and it uses the default keybinding of `super+shift+o`, would it suddenly stop working if the plugin were to update or would the user explicitly have to install the new `SwitchWindow` plugin which just happens to use the GotoWindow code for sublime text < 4?

I would prefer the latter behavior to prevent a breaking change for people already using my plugin. I don’t know if maybe this is an argument to keep them separate.

FichteFoll:
> So, for example, I’m curious if someone has the `GotoWindow` plugin installed and it uses the default keybinding of `super+shift+o`, would it suddenly stop working if the plugin were to update or would the user explicitly have to install the new `SwitchWindow` plugin which just happens to use the GotoWindow code for sublime text < 4?

No, if your intention is to keep the GotoWindow package as-is also for ST4, then it makes no sense to touch your package in the channel at all and SwitchWindow would just become an alternative ST4-only package that people would need to install explicitly. Unless you and @deathaxe are committed to ironing out potential discrepancies in functionality and providing backwards-compatible commands, for example, but that's something you need to figure out yourselves.

From a workflow perspective, our work as channel maintainers (or @braver's mostly, since I'm pratically retired) is done here. If you have any further technical questions, we are of course still here to assist.

deathaxe:
> So my directory here foxy-c will not complete. 

Good point. Displaying the folder name as tiltle for windows which have no project but only a folder open makes definitly more sense than displaying the active file.

> In addition, it seems you may be filtering out the current open window. 

Yes, I do. Not a big deal to revert that change. I have no strong oppinion about the one or the other being better.

> And just another minor thing, I use code to replace the $HOME path with `~` as you can see on the directory paths

Haven't really noticed that, but this is a valuable feature. I like it as it can help to shorten paths.

> This is not entirely true, if I open a new sublime text window and create an unsaved file in it, it does show up ...

Definitly not on Windows. A window doesn't show up until the file is saved to disk.

> ... why even bother rewriting/refactoring my code at all?

There is this long hanging Windows bugfix from ehuss, which I found worth merging for ST3 users. This PR was even what made me think the project got stale. Once done I found the overall code a bit noisy and loading all the windows related structures into namespace when running on Linux/MacOS is of some sort of things I personally can't avoid from trying to improve. ;-) That's just me.

>... regressions by changing the file/code structure?

I've tested the changes on Win/Linux. The callout for possible regressions is rather for someone on MacOS, but I actually don't expect any issues.


> Will this end up as a breaking change for anyone currently using the plugin? Will it just stop working if they already have it installed?

Package names would be renamed automatically upon upgrade (for ST4 users). The most noticible changes would be 
- changed command palette name
- new key binding
- required adjustments to possible custom key bindings as command name changes to `switch_window`

So yes, I'd call it a breaking change. That's why this PR to add the package was created.

--- 

I am fine to add those suggestions to the new version. The main goal of the new Package is just to use ST only API/functions without external OS related dependencies to get rid of no longer needed hacks.

braver:
If the outcome of this discussion is that everybody learns something, and we simply add a new package that sits next to the existing one (ie. nothing changes for its users)… then IMO that’s totally OK and still worth the effort. @deathaxe sounds like the way to go for you too?

deathaxe:
If breaking changes for existing users are unacceptable, yes. Maybe GotoWindow should be restricted to MacOS/Linux if no serious support for Windows can be ensured.

TerminalFi:
Just adding, I use this in Mac daily now and have had no issues.

PR #8674, opened 2023-01-07
[Chef] Move chef package to tags

onlyhavecans:
SousChefs updated the default branch, so this needs to be updated as it currently shows missing  See https://github.com/sous-chefs/sublimechef/issues/45

I am now a maintainer of this package and have updated it to the current standards including gitattributes and using tags

- [x] I'm the package's maintainer.

[2]: https://semver.org

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -864,7 +864,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"branch": "main"
+					"tags": true
 				}
 			]
 		},
--- a/repository/c.json
+++ b/repository/c.json
@@ -864,7 +864,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"branch": "master"
+					"branch": "main"
 				}
 			]
 		},


braver:
We only accept PR’s from users with access to the package repo. Additionally, branch based releases have been deprecated for years, they should switch to tags. Therefore a change of branch will not be accepted any way.

onlyhavecans:
Thanks for that information I will contact the SousChefs

FichteFoll:
I see there is already a tag pushed to the repo at https://github.com/sous-chefs/sublimechef/tags. 

In general, it's not that we absolutely require only changes made by a package maintainer but it must be acknowledged by at least one such maintainer. As such, you would be in the position to perpare the necessary changes but a maintainer must give their okay and should also be aware of the new implications of the tag-based release workflow (i.e. a new release is only installed for users when a semver tag with a higher version is found).

onlyhavecans:
I am now a maintainer of this package, have added tags, and gitattributes

onlyhavecans:
> I see there is already a tag pushed to the repo at https://github.com/sous-chefs/sublimechef/tags.

If you look at history. I am the one who added the tag after joining the org this week.
I acknowledge that any changes that I, or other members make will require creating a tag to be released

PR #8675, opened 2023-01-07
Added ChatGPT Package

eusonlito:
Related package https://github.com/eusonlito/Sublime-Text-ChatGPT

My package is ...

[2]: https://semver.org

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -766,6 +766,20 @@
 			]
 		},
 		{
+			"name": "ChatGPT",
+			"details": "https://github.com/eusonlito/Sublime-Text-ChatGPT",
+			"description" : "OpenAI ChatGPT connector",
+			"author": "Lito",
+			"labels": ["auto-complete"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true,
+					"branch": "master"
+				}
+			]
+		},
+		{
 			"name": "ChatScript Syntax",
 			"details": "https://github.com/kuzyn/chatscript-tmlanguage",
 			"labels": ["language syntax"],


eusonlito:
@packagecontrol-bot Please :)

braver:
To start with, your readme could do with some more explaining of what your package does, how to use it, maybe example use cases etc.

Also, the template for pull requests contains some points about not shipping key bindings, and about putting commands in the command palette… maybe revisit those, or expand on why you chose differently.

eusonlito:
Ok, sorry then...

I was searching by ChatGPT and doesn't return any results...

I have tried to install the OpenAI package but...

```
Package Control: Error downloading package. HTTP exception InvalidCertificateException (Host codeload.github.com returned an invalid certificate ([SSL: TLSV1_ALERT_PROTOCOL_VERSION] tlsv1 alert protocol version (_ssl.c:657))) downloading https://codeload.github.com/yaroslavyaroslav/OpenAI-sublime-text/zip/1.1.1.
error: Package Control

Unable to download OpenAI completion. Please view the console for more details.
```

It's my first package and I'm a bit confused 😅

Do you think that my package should be discarded?

Thanks!

eusonlito:
Updates:

* Removed default Key Bindings
* Added settings comments
* Updated README
* Improved `Package Settings` > `ChatGPT` > Menu

rchl:
To be fair, the other package is exactly the same thing (even if it doesn't include ChatGPT word in it). So I think per our policy we shouldn't accept it.

Also this package is lacking a bit in some ways. For example:
 - the requests are blocking (made on main thread) so the UI will be blocked until the request completes,
 - the menu entries for key bindings and settings use ST2-style where the default and user files are opened separately. This is inconvenient and not recommenced anymore since ST2 is long gone.
 - The other package has quite a bit of extra functionality, just by quick look at its settings and code.

rchl:
We could add a chatgpt tag to the other package to improve discoverability.

rchl:
> I have tried to install the OpenAI package but...

Sounds like an issue with Package Control in general in your installation and not specific to the package you were trying to install. Probably will get the same issue with most other pacakages.

eusonlito:
Ok, thanks anyway.

I will continue to use my package by manual installation because I don't want to write in the editor to use the ChatGPT.

I have updated my code to open settings and key bindings in split view.

Also added comments on settings and some other improvements.

I will add also the request worker to avoid ui blocks.

Closing PR :)

Thanks!

eusonlito:
Also added non-blocking requests :)

rchl:
Is there something about the other package that makes you not wanna use it? Curious why do you prefer maintaining your own.

eusonlito:
I have no idea as use it 😅

No Key Bindings explanation, or execute menu...

README say:

> Open the command palette and run the OpenAI Append, OpenAI Insert, or OpenAI Edit command.

I execute:

```
>>> OpenAI Append
  File "__main__", line 1
    OpenAI Append
           ^
SyntaxError: invalid syntax
```

Is the first package that I don't know how to execute 🥲

My Plugin include the `Open Command Palette` option on Package Settings Menu, also, the README link.

Maybe it's just me 🤷

Thanks for your concern.

rchl:
The command palette is what you trigger with ctrl+shift+p (or cmd+shift+p on mac).

Key bindings are not provided by default as it's not recommended but you can easily set up any key binding you want through `Preferences: Key Bindings` in the Command Palette.

eusonlito:
Ok, I get it.

I think the main difference with my plugin is that everything is requested from the input panel instead of writing the needs in the editor, for me it's much cleaner and clearer.

Thanks for your support :)

rchl:
It appears to me that the other package has 3 different modes of entering queries and at least one of them uses the same method as you (`show_input_panel`). But I don't know, haven't tried it myself. Up to you but you could contribute suggestions or features to it probably if you want to save some time. :)

yaroslavyaroslav:
I suddenly found this one, so I can answer. Yes, my package have both features: to ask question in the input view rather the sending to the server selected text and to edit selected text with a given command again in input view, though the latter gives relatively bad quality on my tasks.

And to be honest I can confirm the lack of plugin documentation that was raised by the PR author.

PR #8676, opened 2023-01-08
Unify key order in dependencies.json

FichteFoll:
Also adds `platforms` key to all items for consistency with the already
required `sublime_text` key and soon-to-be-added `python_version` key.

Follow-up on #8599 (partially cherry-picked) but additionally hard-requires the platform keys in tests.

### DIFF
--- a/repository/dependencies.json
+++ b/repository/dependencies.json
@@ -4,156 +4,164 @@
 	"dependencies": [
 		{
 			"name": "anytree",
-			"load_order": "01",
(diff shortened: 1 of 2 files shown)


FichteFoll:
I tried to attribute the first commit to @deathaxe, but had to guess an email address that GitHub would detect and apparently I failed (I thought it was `uid+username@users.noreply.github.com`).

braver:
👍🏻 now is a good time too, no open pr’s for dependencies.

PR #8678, opened 2023-01-21
Add WslBuild package

deathaxe:

My package is a custom build system target to run build commands in WSL2 on Windows OS.

It is a follow up of https://github.com/wbond/package_control_channel/pull/8660.

It ships various fixes and improvements to make sure all kinds of scripts are addressed well within WSL environment - no matter where they are located (within WSL or on a windows folder).

[2]: https://semver.org

### DIFF
--- a/repository/w.json
+++ b/repository/w.json
@@ -1119,6 +1119,17 @@
 			]
 		},
 		{
+			"name": "WslBuild",
+			"details": "https://github.com/SublimeText/WslBuild",
+			"releases": [
+				{
+					"sublime_text": ">=3176",
+					"platforms": ["windows"],
+					"tags": true
+				}
+			]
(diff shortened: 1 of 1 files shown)


braver:
Good move 👍🏻 

Maybe the nice to do would be to add the author of the original codebase access to the new repo?

Also personally I could do without the icons/emoji in the names of the build systems, one set of build systems doesn't need to be more visible or noisy than any others.

braver:
> Maybe the nice to do would be to add the author of the original codebase access to the new repo?

Well that sentence doesn't make any sense, sorry. "the nice [thing] to do" would be to "give the original author access to the new repo". Anyway, I'll let you guys figure that out in your own time, let's ship this thing.

PR #8679, opened 2023-01-22
Spin-off GitSavvy ST3 version

kaste:
GitSavvy will switch to ST4/py3.8 with the next (or so) release.  The plan is at least to only do patch releases for ST3.  We'll see.

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -852,18 +852,13 @@
 		{
 			"name": "GitSavvy",
 			"details": "https://github.com/timbrel/GitSavvy",
-			"labels": [
-				"git",
-				"dashboard",
-				"inline",
-				"stage",
-				"patch"
-			],
-			"readme": "https://raw.githubusercontent.com/timbrel/GitSavvy/master/README.md",
-			"author": "divmain",
 			"releases": [
 				{
-					"sublime_text": ">=3000",
+					"sublime_text": "<4000",
+					"tags": "st3-"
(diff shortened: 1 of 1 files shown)


TerminalFi:
Just curious, why remove the labels which are related to the package function ?

kaste:
I thought if I remove the readme key, package control will still find it because it is just the standard name of the file anyway.  I removed the author because GitSavvy does not have *one* author, and if that matters **I** have probably 2x or 3x the commits divmain did.   (Honestly, I'm just not for author *singular* for OOP.)  The labels are non-sensical, no?   'stage' , 'patch'  are no labels and yeah okay 'git', but it is in the name.

TerminalFi:
> I thought if I remove the readme key, package control will still find it because it is just the standard name of the file anyway. I removed the author because GitSavvy does not have _one_ author, and if that matters **I** have probably 2x or 3x the commits divmain did. (Honestly, I'm just not for author _singular_ for OOP.) The labels are non-sensical, no? 'stage' , 'patch' are no labels and yeah okay 'git', but it is in the name.

Yeah you are correct on the readme being auto detected. 

The authors isn’t a big deal. Just wondering. LGTM

FichteFoll:
FYI, `author` can also be an array of multiple authors (which it obviously won't do when determining it from the GitHub URL). Feel free to make a new PR with such a change.

PR #8680, opened 2023-01-23
add 42 FT COUNT LINE

DoKca42:

My package is an extension for students of 42 allowing to count the number of lines of a function according to the 'norminette' standard

[2]: https://semver.org

### DIFF
--- a/repository/0-9.json
+++ b/repository/0-9.json
@@ -79,6 +79,16 @@
 			]
 		},
 		{
+			"name": "42 Function Count Lines",
+			"details": "https://github.com/DoKca42/42_sublime_ft_count_line",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "42 Headers",
 			"details": "https://github.com/Globidev/Sublime-Text-42-Headers",
 			"releases": [


FichteFoll:
- You should use a more distinct identifier for your phantoms than `ee`.
- Performing an operation on every modification is probably not the best idea because it will prevent the grouping of modifications for the undo history and also cause some latency especially for large files. A better idea would be to debounce the update until no modification has been done for a certain amount of time, e.g. 200ms.
- Inside of a `TextCommand`, the view the command was run for is available as `self.view`.
- `view.file_name()` may be `None` for newly created, unsaved views.

DoKca42:
Thanks for your feedback.
I have corrected all the elements you gave me to modify (tag with semantic version, rename phantoms,  200ms cooldown, self.view and None file_name protection) do I need to make a new pull request now ?

And what about the "trigger-review failed" do I have to change something else or is the failure related to the elements you gave me ?

DoKca42:
I have add a LICENCE to my repo, do I have to change something else ?

DoKca42:
Can you explain me the "trigger-review (pull_request)" fail in my case?

FichteFoll:
No, but it does not matter since the automated review did eventually happen.

You should find that with your current deboincing implementation, you put an emphasis on performing operations on the first change after at least 0.2s rather than 0.2s after the last modification. Additionally, your event listener will be called for multiple views (which is unlikely to result in actual issues due to 0.2s really not being that long).

To implement proper debouncing, you should take a look at `sublime.set_timeout` which you call with a lambda or function annotated with `functools.partial`,  and compare the results of `view.change_count()` at the initial time of the event and after the timeout, e.g.

```py
class EventDump(sublime_plugin.EventListener):

    def on_modified(self, view):
        sublime.timeout(functools.partial(view, view.change_count()), 200)
        
    def debounce(self, view, last_change_count):
        if view.change_count() == last_change_count:
            pass  # do your stuff here
```

However, the package will function as-is and thus I'm merging the PR, but please try out and intergrate the above snippet regardless.

PR #8681, opened 2023-01-28
Add Dhall Syntax Highlighting

kukimik:

My package is Dhall Syntax Highligting (a `.tmLanguage` definition for the [Dhall](https://dhall-lang.org/) language).

My package is similar to https://packagecontrol.io/packages/Dhall. However it should still be added because the Dhall package is no longer available (the github repo was deleted). I have a copy of the `.sublime-syntax` file from some version of the Dhall package (obtained somewhere on github), but I have no information about its licensing, so I think I can't just redestribute it.

This also explains why I do not obey the rule included in the review guidelines:

> Syntax packages should be named after the language, i.e. "LESS", not "LESS syntax" or "LESS highlighting" because syntaxes in ST do much more than just highlighting.

The name "Dhall" is already taken, by a non-existent package (see also https://github.com/wbond/packagecontrol.io/issues/154).

[2]: https://semver.org

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -805,6 +805,17 @@
 			]
 		},
 		{
+			"name": "Dhall Syntax Highlighting",
+			"details": "https://github.com/kukimik/dhall-sublime-syntax-highlighting",
+			"labels": ["language syntax", "dhall"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Diagram",
 			"details": "https://github.com/jvantuyl/sublime_diagram_plugin",
 			"labels": ["diagrams", "plantuml"],


braver:
> The name "Dhall" is already taken, by a non-existent package

In that case it's better if you replace the existing entry with your repo. 

There is a convertor built into Sublime Text to turn your tmLanguage file into a sublime-syntax file. That will make it a lot easier to maintain going forward. Once you've done that you might be able to compare yours to the file you have from the old repo. You could use that to, if needed, complete and improve your package. I agree a 1-1 copy of the old file might be questionable, but you can definitely use it for "inspiration", it's a bit hard to actually claim authorship of these things.

braver:
> If there are still users of the original package, wouldn't this update the package they have on their disks, replacing its files with the new ones? 

Yes, that's what it would do. Usually that's a good thing if the replacement covers the same functionality and is actually somewhat actively maintained. 

This package has currently been installed just 80 times. 

> I can't say my package is superior

Well it is, because your package actually exists 😉 

Are many features missing in your package? I'm not to worried about formatting, that's probably better handled by Fmt anyway and you can point existing users to an alternative solution like that. 

> Also, technically, how should this be done?

You can create a new entry with the same name of "Dhall" and it should just work. Might need manual intervention somewhere, but that should work out just fine.  

> If you see this as a blocking issue, then I'll try to fix this and then come back here

Since your package is targeting Sublime Text specifically, we try not to add tmLangauge based packages. Also because they're criminally hard to maintain, which doesn't help for the quality of packages in the long run. If it doesn't convert, that's a problem we can investigate and solve, not something we should simply accept. I can't see the problem right away either though.

kukimik:
> Since your package is targeting Sublime Text specifically, we try not to add tmLangauge based packages. Also because they're criminally hard to maintain, which doesn't help for the quality of packages in the long run. If it doesn't convert, that's a problem we can investigate and solve, not something we should simply accept. I can't see the problem right away either though.

It's derived of a VSCode package and I don't want to get out of sync with it, so in my case it's actually easier (from the maintenance viewpoint) to keep it a tmLanguage file, because that is what VSCode is using (in the json version). 

But I would also prefer to have a sublime-syntax file (one reason is that it can be used by other tools via the https://github.com/trishume/syntect library), but only if I can obtain it in an automated (or at least semi-automated) way.

Anyway, I'll try to spend some time investigating this issue. Perhaps with a working sublime-syntax file I wouldn't be afraid of upsetting one of those 80 potential users. :wink:

For now I'm converting this PR to draft.

braver:
> It's derived of a VSCode package and I don't want to get out of sync with it

Well that's an entirely valid reason.

braver:
I'm going to close this for now. Let us know if you're done and want to re-start the review.

PR #8683, opened 2023-02-03
Add ExpandSelectionToComments package

timfjord:

The package expands the current selection to fully enclose the intersecting comment block

- works with multiple cursors
- supports soft undo
- supports comments divided with empty lines
- works with any language that has syntax highlighting

[2]: https://semver.org

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -1602,6 +1602,19 @@
 			]
 		},
 		{
+			"name": "ExpandSelectionToComments",
+			"details": "https://github.com/timfjord/ExpandSelectionToComments",
+			"labels": [
+				"expand", "selection", "comment"
+			],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


rchl:
As I potential user of this package I would wonder what "Expands the current selection to the nearest comment block" exactly means. The screencast in the repo makes it clear but by just reading the text I would think that it has to do with having cursor outside of the comment block and expanding selecting to touch the comment block from the outside? Not sure.

Would it be more precise to say "Expands the current selection to contain the nearest comment block"?

Also I wonder what happens when cursor is outside of the comment block? Nothing? If so then maybe it can be even more precise: "Expands the current selection to fully enclose the intersecting comment block"?

timfjord:
@rchl I like your suggestion. Thanks.

I've updated the description all over the places

braver:
Looks good and useful. Only thing I'd like to change is the position of your command in the Selection menu... it's a bit lost now. Nothing you can do much about I believe. So, let's ship it.

timfjord:
Woohoo 🎉 
Thanks @braver

PR #8684, opened 2023-02-04
add DefinitionPreview package

nickav:

My package is a quick preview window for function and type definitions. It does this without any LSP and works entirely with vanilla Sublime Text. This is particularly useful on Windows where installing those tools is often very difficult.

As far as I'm aware, there are no packages like it in Package Control.

[2]: https://semver.org

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -509,6 +509,16 @@
 			]
 		},
 		{
+			"name": "Definition Preview",
+			"details": "https://github.com/nickav/DefinitionPreview",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "DelayedSaveAll",
 			"details": "https://github.com/shagabutdinov/sublime-delayed-save-all",
 			"donate": "https://github.com/shagabutdinov/sublime-enhanced/blob/master/readme-donations.md",


kaste:
Interesting, it is similar in spirit to https://github.com/gladclef/HoverDocs (which btw was added last August and already is unmaintained 🙄).  

What I don't like is that it clearly refers "c" language scopes so it is very limited.  *But* it still runs on Python code and then gets it almost right.  

E.g. in your own source code

image

Almost right but the current argument is empty. 

I do like the spirit but I don't see how this is pluggable/extensible so other languages *could* be supported.  What had you in mind with your package?

nickav:
Yeah it's similar. I think the approach I'm taking is more focused about previewing function signatures and where you are in that context. Although in theory I could also add support for docstrings.

It is currently limited to C and C-like languages. My plan is to continue to add more test cases to the `public` folder as I try it out on more projects / languages. So I would manually be updating that. If you have any suggestions for how to make it more "pluggable" or more robust in general I'm definitely open to suggestions. I guess one question is how does Sublime Text build these "definition lists" itself, and could some of that be used to re-expose the full function declaration more easily?

That is interesting! I've been testing it out on itself while I develop it and I actually see that case working (de-duplicating is on the roadmap):
python_example
So I'm not exactly sure about the difference between our two setups. Maybe the scope names are different? Would need more info to fix

repository/d.json
```diff
@@ -508,6 +508,16 @@
 				}
 			]
 		},
+		{
+			"name": "Definition Preview",
+			"details": "https://github.com/nickav/DefinitionPreview",
+			"releases": [
+				{
+					"sublime_text": "*",
```
kaste, line 516: AFAICS you plugin is ST4 only

```suggestion
					"sublime_text": ">=4000",
```

kaste:
Hm, why do you have two `find_latest_scope_sub_name` references (`locs`), there is only one definition in your project.  Maybe you're already ahead of the public master?  On the version I have the `<b>` tags are somehow misaligned.  E.g. (excerpt) here you want to highlight "view":

```
<span style="color:#bbbbbb;"><b>(</span><span style="color:#b0b1b1;">view</span><span style="color:#bbbbbb;"></b>,</span>
```

I generally like it.  You should enable Python 3.8.  The plugin is ST4 only anyway.

nickav:
That's because I have multiple copies of that file open in the project (one `DefinitionPreview.py` in the repo and one that's currently installed / pasted in my Sublime Text User config folder).

Ah good catch, that's an easy fix (updated repo).

kaste:
I think the problem is that Sublime doesn't like `<b></span><span></b>` constructs.  These are invalid.  What html you get before boldifying it depends on the color scheme.  So for the source `foo(a, b, c)` you might get a single span having all the parameters but I have a scheme where the punctuation is slightly emphasized (`#bbbbbb`), or the parameter names are a little bit dimmed (`#b0b1b1`) so I have very noisy html to begin with.  I don't think your `highlight_function_argument` (which is a post-processor of Sublime's `view.export_to_html`) is prepared for that.  

*It would be sad if your plugin breaks with any random color-scheme (like mine).*

That's a bit of the problem domain here of that function in general, because we can easily have `def foo(bar: Union[str, int], zoo: int):` (Python).  Just searching for a `,` doesn't seem right.  You lost Sublime Text syntax understanding, the scopes at this point - for Python the `,` are  `punctuation.separator.parameters.python`-, and traded it for a raw (html) string.  

Let's also wait for others reviewers as technically this is mergeable.

nickav:
Yeah good point. I should do something smarter with that as well. Will add to the TODO list!

Let me know if any issues are critical to fix before submitting

braver:
Please consider using .gitattributes. This keeps your package lean, since most of the bytes are for images that are only used in the readme on GitHub.

braver:
That said I won't let that block the package further. The package looks good and well considered. If not necessarily set up for wide adoption. But it's good to have some options in the eco system. The similarity with the aforementioned HoverDocs is unfortunate. We prefer seeing people helping out on each others packages rather than publishing several different implementations of the same idea. ... and in the process adding more and more unmaintained packages. But that's also something you can figure out in your own time. It might be worth seeking out the maintainer of HoverDocs, if he's actually going to take that project anywhere. We can always consolidate the two later.

PR #8685, opened 2023-02-05
Add Easy Isort package

tiyujopite:

My package is [Easy Isort](https://github.com/tiyujopite/easy_isort), this is used to sort python files imports.

My package is similar to [isort](https://github.com/asfaltboy/sublime-text-isort-plugin) or [isorted](https://github.com/rimvaliulin/isorted). However it should still be added because these packages do not work as expected, or have a very complex configuration for the user. (I have decided to make my own package because after much testing I have not been able to get any of the two to work).

[2]: https://semver.org

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -80,6 +80,18 @@
 			]
 		},
 		{
+			"name": "Easy Isort",
+			"details": "https://github.com/tiyujopite/easy_isort",
+			"labels": ["isort", "python"],
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


kaste:
This is a rather idiosyncratic way to do this.  

Why `sys.path.append(os.path.dirname(__file__))`?  Why do you use the old `os.system()` function?  Did you asked your questions on the isorted tracker or elsewhere?

PR #8157, opened 2021-01-13
Adding Gleam package

itsgreggreg:
Hi there, this is a package for [Gleam](https://gleam.run) language support. It adds syntax highlighting and formatting on save.

Thanks! -Greg

PR #8689, opened 2023-02-13
Update/takeover Gleam package

digitalcora:
- [x] I'm the package's author.
- [x] **(N/A)** Any commands are available via the command palette.
- [x] **(N/A)** Preferences and keybindings are listed in the menu and the command palette, and open in split view.
- [x] My package is a syntax, and it doesn't also add a color scheme.
- [x] I use [.gitattributes][3] to exclude files from the package.

[2]: https://semver.org

### Context

[This package](https://github.com/digitalcora/sublime-text-gleam) supports the [Gleam](https://gleam.run/) programming language.

I'm proposing a takeover of the existing package named "Gleam", owned by @itsgreggreg ([source](https://github.com/itsgreggreg/sublime-text-gleam)), added in #8157. I installed this package as a user and reported a few bugs in it, and the owner responded by archiving the repo without a reply. I'm assuming this means they are no longer interested in maintaining it, and they don't seem to have any public contact info, so I've just tagged them here in case there is any objection.

Versus the original package, this one:

* ...has slightly better syntax support, and recognizes recent additions to the language (`use` expressions). It still doesn't do much "real" parsing, but I'd like to improve on this in the future for better meta scope coverage.

* ...doesn't have the "format on save" behavior (which was slightly broken, as the setting to turn it off didn't work, meaning lots of error pop-ups if it didn't work with your environment). Since Gleam's built-in formatter works with stdin/stdout, I reason this is better done by a tool like [Fmt](https://packagecontrol.io/packages/Fmt), and have included instructions for setting this up. Thanks to @braver in [this comment](https://github.com/wbond/package_control_channel/pull/8681#issuecomment-1410192199) for the pointer to Fmt.

* ...doesn't support Sublime Text 3, as it uses version 2 of the syntax highlighter. I don't think there is _currently_ anything in the syntax that requires this, but I'm wary of running into bugs that are unfixed in version 1, and followed the advice of the [official docs](https://www.sublimetext.com/docs/syntax.html#header) that "New syntaxes should target `2`".

Here are screenshots of the syntax in action, using some examples from the Gleam Language Tour:

* [Breakers](https://user-images.githubusercontent.com/394835/218580347-665bb929-9042-499a-80c5-2de0ed8f9fb7.png) (default light theme)
* [Mariana](https://user-images.githubusercontent.com/394835/218580361-242914a4-f560-47ad-9423-7065c218fbb8.png) (default dark theme)

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -951,12 +951,12 @@
 		},
 		{
 			"name": "Gleam",
-			"details": "https://github.com/itsgreggreg/sublime-text-gleam",
+			"details": "https://github.com/digitalcora/sublime-text-gleam",
 			"labels": ["gleam", "language syntax"],
 			"releases": [
 				{
-					"sublime_text": ">=3092",
-					"tags": true
+					"sublime_text": ">=4075",
+					"tags": "v"
 				}
 			]
 		},


braver:
> I'm assuming this means they are no longer interested in maintaining it, and they don't seem to have any public contact info, so I've just tagged them here in case there is any objection.

I agree. Let's give it 2 weeks until we proceed, just to have a little opening for complaints.

Since you're dropping ST3 support, I think it would be wise to put a not in your README to point users of that version of ST to the old repository and tell them to manually install a version from that.

digitalcora:
Thanks @braver! I will do that (hope it's not an issue if I modify tags before this gets merged). Will check back on 2/27.

digitalcora:
I've made some updates, including a section in the README about Sublime Text version compatibility. I tried it out and ST3 actually loads v2 syntaxes without complaint (it just uses the v1 engine, of course). So what I advise is that the package _can_ be installed manually in ST3, there's just no guarantee it works perfectly. This first release does appear to work fine, though.

One question I have: I used some HTML in the README to get two screenshots displaying side by side, at a smaller-than-native size. This works on GitHub, but will it also work on the Package Control site?

braver:
> This works on GitHub, but will it also work on the Package Control site?

I assume it will, HTML is totally accepted and normal inside Markdown.

digitalcora:
Hi! It has now been 2 weeks since submission and it doesn't seem we have any objections. Good to go?

braver:
Yep, let's do it!

PR #8690, opened 2023-02-14
Add Flutter-Dart-Code package

james2doyle:

My package started as a fork of [elMuso/Dartlight](https://github.com/elMuso/Dartlight) but has grown into a more complete Dart and Flutter package. It includes a native sublime-syntax file, many more snippets,  as well as a plugin.

[2]: https://semver.org

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -1288,6 +1288,17 @@
 			]
 		},
 		{
+			"name": "FlutterDartCode",
+			"details": "https://github.com/james2doyle/Flutter-Dart-Code",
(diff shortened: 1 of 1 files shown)


braver:
I think in general it would be nice to have an up to date package for dart, maybe to replace or update [the existing package](https://packagecontrol.io/packages/Dart). I don't fully understand the relationship between dart and flutter, but for future maintainability I think it makes sense to decouple the two. We also have an existing package for [flutter](https://packagecontrol.io/packages/FlutterEnhancements), seemingly overlapping with your offering here. Maybe you can contribute improvements there as well?

PR #8691, opened 2023-02-15
Add Streamlit

futureprogrammer360:

My package is [Streamlit](https://github.com/futureprogrammer360/Sublime-Streamlit), a package for [Python's Streamlit package](https://streamlit.io/) providing completion, build, and documentation.

[2]: https://semver.org

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -4038,6 +4038,17 @@
 			]
 		},
 		{
+			"name": "Streamlit",
+			"details": "https://github.com/futureprogrammer360/Sublime-Streamlit",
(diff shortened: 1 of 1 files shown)


braver:
Looks good 👍🏻 One thing though. Your package will be delivered as a zip and you cannot access files in it using os.path etc. Instead use sublime.load_resource. Example: https://github.com/braver/ColorHints/blob/master/lib/pantone.py#L37

futureprogrammer360:
@braver Thank you for the feedback and the example, the issue should be addressed now. Please let me know if there's anything else I should fix. Thanks!

braver:
Don’t forget to tag a release with that change!

PR #8693, opened 2023-02-18
[ghosttext] Update meta of existing package

fregante:

No changes were made in the following or are unrelated:

My package is https://packagecontrol.io/packages/GhostText

I just added some more fields from https://raw.githubusercontent.com/wbond/package_control/master/example-repository.json

[2]: https://semver.org

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -419,7 +419,6 @@
 			"issues": "https://github.com/fregante/GhostText/issues",
 			"donate": "https://github.com/sponsors/fregante",
 			"homepage": "https://ghosttext.fregante.com",
-			"author": "fregante",
 			"previous_names": ["ChromeTextArea"],
 			"labels": [
 				"browser integration"
--- a/repository/g.json
+++ b/repository/g.json
@@ -416,7 +416,14 @@
 		{
 			"name": "GhostText",
 			"details": "https://github.com/GhostText/GhostText-for-SublimeText",
+			"issues": "https://github.com/fregante/GhostText/issues",
+			"donate": "https://github.com/sponsors/fregante",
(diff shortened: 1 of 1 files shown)


repository/g.json
```diff
@@ -416,7 +416,14 @@
 		{
 			"name": "GhostText",
 			"details": "https://github.com/GhostText/GhostText-for-SublimeText",
+			"issues": "https://github.com/fregante/GhostText/issues",
+			"donate": "https://github.com/sponsors/fregante",
+			"homepage": "https://ghosttext.fregante.com",
+			"author": "fregante",
```
braver, line 422: Author here means "author of the package", not whatever the underlying tools are. Similarly for issues, it's typical for issues here to mean "issues with the ST package", not "issues with the underlying tool" and it's common practice to point users to other repositories for those underlying tools. Many packages are just thin wrappers around something else. I think your changes here might conflate these things in a way that's not 100% correct and might be confusing. Correct me if I'm wrong though.

fregante: You might to follow those links to better understand the setup.

- ghosttext/ghosttext-for-sublimetext is the repo for this package, but it does not have issues
- fregante/ghosttext is the main repo, where all the issues (browser extensions and ST package) are opened
- I'm the maintainer of the whole GhostText org, including the ST package, as the original programmer *ghosted* the project several years ago

braver: Thanks for explaining! I guess I was mostly reacting to the fact that you don't seem to have many actual commits on the Sublime Text package. Since you're the maintainer of the thing as a whole it's ok to have you as the "author" here, although I would argue that for the ST-specific package it might actually be fairer to have it point to the org, signaling that it's a community project more than something you've created all by yourself.

PR #8694, opened 2023-02-19
Add phpfmt8 package

driade:

My package is a clone of phpfmt but with php8 support

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -1342,29 +1342,29 @@
 			]
 		},
 		{
-            "name": "phpfmt",
-            "details": "https://github.com/nanch/phpfmt_stable",
-            "labels": ["language syntax", "formatter", "php", "formatting"],
-            "releases": [
-                {
-                    "sublime_text": ">=3000",
-                    "platforms": "*",
-                    "tags": true
-                }
-            ]
-        },
-        {
-            "name": "phpfmt8",
-            "details": "https://github.com/driade/phpfmt8",
-            "labels": ["language syntax", "formatter", "php", "formatting"],
-            "releases": [
-                {
-                    "sublime_text": ">=3000",
-                    "platforms": "*",
-                    "tags": true
-                }
-            ]
-        },
+			"name": "phpfmt",
+			"details": "https://github.com/nanch/phpfmt_stable",
+			"labels": ["language syntax", "formatter", "php", "formatting"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"platforms": "*",
+					"tags": true
+				}
+			]
+		},
+		{
+			"name": "phpfmt8",
+			"details": "https://github.com/driade/phpfmt8",
+			"labels": ["language syntax", "formatter", "php", "formatting"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"platforms": "*",
+					"tags": true
+				}
+			]
+		},
 		{
 			"name": "PHPGrammar",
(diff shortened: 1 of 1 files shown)


braver:
Won't the FMT package already support this as well? Otherwise it sounds like this would be a useful feature addition to phpfmt (it would be a bit annoying to have to change editor packages just because you're using a newer version of a language).

driade:
Hello, good morning.

The current phpfmt package, https://github.com/nanch/phpfmt_stable, it's a fork of the original package (deleted from Githtub).

It seems the fork hasn't been updated/maintained in the last seven years, and not only lacks of support for php 8 but also has many bugs for php 7.x. For example php 7.4 was released in 2019.

My package is a fork of the current phpfmt package with added php 8 support. So, it also works with php 7.x.

Thanks.

driade:
Or, the other way around, should we try to contact @nanch and ask for the transfer of the package?

braver:
Yes, if the original package looks dead, let's see if we can get you access to it, or transfer it.

driade:
Hi, good afternoon. Should we advance?

braver:
Yes, this is obviously not going anywhere. That opens up the possibility of replacing the existing package, so please adjust the repo entries to make that happen. I would still like to review the new package and make it up to date with our current requirements.

driade:
Hi, thanks, I've already done it. Please confirm if it's fine.

braver:
Alright, that looks good. Now let’s have a look at the package. I’m going to treat it like a new package, let me know if some of these suggestions are not possible because of backwards compatibility etc.

- Make sure your repo has a tags that’s newer (higher number) than the previous package.
- It might also be a good idea to have a message for this version to tell existing users what happened.
- Please update the menu entries to open settings in split view instead of having separate default/user entries.
- Is the no-sublime-package file really necessary?
- You don’t need to support python 2.
- You might want to talk about Sublime Text 4 instead of 3, or leave out the version number… ST3 is ancient history 😉
- What are the “version” and “option” default settings for? Also, please indent that JSON. It’s a bit weird to see unformatted files in a formatted package 😅

driade:
Hello, good night, let's see if I did everything right.

> Make sure your repo has a tags that’s newer (higher number) than the previous package.

Updated the tags to house a "v" before the number so they follow the original tag structure. I've checked there're no tags higher than the current one

> It might also be a good idea to have a message for this version to tell existing users what happened.

Done

> Please update the menu entries to open settings in split view instead of having separate default/user entries.

Done

> Is the no-sublime-package file really necessary?

As fas as I understand, yes, it's

> You don’t need to support python 2.

Directory deleted.

> You might want to talk about Sublime Text 4 instead of 3, or leave out the version number… ST3 is ancient history 😉

Done

> What are the “version” and “option” default settings for?

It seems it's an old setting from the original package supporting prior versions of the config. I suppose it's something I could check and delete in the future.

> Also, please indent that JSON. It’s a bit weird to see unformatted files in a formatted package 😅

👍

braver:
Thanks! Almost there 👌🏻 

> You may find an example configuration in the file driade.sublime-settings

You might want to link to the file at https://github.com/driade/phpfmt8/blob/master/driade.sublime-settings.

It looks like you may want to delete this bit of code and the version setting that is used here: https://github.com/driade/phpfmt8/blob/master/phpfmt.py#L1004-L1039
It looks like it's some way of handling version upgrades years ago, but doesn't make any sense currently. Besides, the setting it needs isn't documented in a way that users can use it.

driade:
OK, done :)

PR #8696, opened 2023-02-20
Added support for osx and windows

necarlson97:

Already an existing approved package - just looking to allow install from OSX and Windows machines, as the repo now supports them.
Let me know if there is anything else I can do to help / any other data you need for me to provide. Thanks!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,6 @@ jobs:
     steps:
       - name: Trigger package review
         # continue-on-error: true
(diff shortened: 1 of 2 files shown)


FichteFoll:
Just a note for the future: a tag like [`v1.5`](https://github.com/necarlson97/codex-ai-sublime/releases/tag/v1.5) will be ignored by PC because it does not have the three numeric parts required by https://semver.org/.

necarlson97:
Thanks! And understood

PR #8697, opened 2023-02-21
add Pretty EDN

oakmac:
https://github.com/oakmac/sublime-pretty-edn

My package is for working with data in [EDN format](https://github.com/edn-format/edn).

[2]: https://semver.org

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -2225,6 +2225,19 @@
 			]
 		},
 		{
+			"name": "Pretty EDN",
+			"details": "https://github.com/oakmac/sublime-pretty-edn",
+			"author": "Chris Oakman",
+			"labels": ["edn", "clojure", "clojurescript", "babashka", "clj", "cljs", "pretty", "lint", "minify", "validate"],
+			"releases": [
+				{
+					"platforms": ["osx", "linux"],
+					"sublime_text": ">=4000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Pretty JSON",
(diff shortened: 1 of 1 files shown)


braver:
Thanks for submitting your package. We usually need a little time to review submissions (especially if they’re not simply a color scheme or syntax).

Because packages are delivered in a zipped format, you can’t access and execute your scripts this way. Unless you can work around that in some way, you could look into using a .no-sublime-package file. 

I would also suggest using early return instead of deeply nesting your if statements.

oakmac:
> Thanks for submitting your package. We usually need a little time to review submissions (especially if they’re not simply a color scheme or syntax).

Gotcha 👍 No worries 😄 

> Because packages are delivered in a zipped format, you can’t access and execute your scripts this way. Unless you can work around that in some way, you could look into using a .no-sublime-package file.

I added a `.no-sublime-package` file to the root and bumped the version 👍 

> I would also suggest using early return instead of deeply nesting your if statements.

I am a little confused here: this seems like stylistic / preference feedback about the upstream Python code in my package.

Is this feedback more along the lines of "hey - I saw this while reading your code and just giving my opinion" or more like "I do not feel comfortable merging this package to Package Control unless you make these changes"?

It is ok either way; I am just trying to figure out the intent here.

braver:
> this seems like stylistic / preference feedback about the upstream Python code in my package

might be, not everyone does Python the same way, but typically you try to indent as little as possible (significant-white-space languages stay easier to manage that way).

totally this though:

> hey - I saw this while reading your code and just giving my opinion

PR #8699, opened 2023-02-26
Add Adaptify theme

lodev09:

My package is a theme that comes with 2 color schemes. It extends the default _Adaptive_ theme.

[2]: https://semver.org

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -939,6 +939,17 @@
 			]
 		},
 		{
+			"name": "Theme - Adaptify",
+			"details": "https://github.com/lodev09/Adaptify",
+			"labels": ["theme", "color scheme"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Theme - Afterglow",
 			"details": "https://github.com/YabataDesign/afterglow-theme",
 			"labels": ["theme"],


deathaxe:
This theme overrides default Adaptive.sublime-theme with obvious/significant changes. This shouldn't have been merged.

see: https://github.com/lodev09/Adaptify/issues/1

braver:
I thought it was pretty obvious that that’s what it does. Of course any breaking change to Adaptive might break it, but how realistic is that to ever happen, and how much of a problem is that?

deathaxe:
It just disables the use of unmodified original Adaptive theme, once this package is installed. 

It's not obvious it is doing that, probably resulting in bug reports about broken or changed visuals of Adaptive theme at sublimehq tracker. It looked like a color scheme package on the first glance.

The shipped theme adds some broken/duplicate paddings to quick panel items, for instance.

Originial themes/color schemes should just not be augmented that way.

If a package provides a modified theme, it should express it by a dedicated name, which was easily possible via `"extends": "Adaptive.sublime-theme",` key.

braver:
Oh, right, totally. Got it. I was assuming that it did exactly as you describe, by making a new theme available under a new name but under the hood extending the default theme. Didn't look closely enough I guess.

I see that a bug report is already open on the Adaptify repo and I assume the issue can be resolved there. If I can be of any assistance let me know.

lodev09:
I tried the `extends` key and works. However, the default icons are no longer working. It looks like a bug under the hood.. or am I missing something?

PR #8705, opened 2023-03-09
Add Translator (multi-engine, fast) package

dmytrovoytko:

My package is similar to [Inline Google Translate](https://packagecontrol.io/packages/Inline%20Google%20Translate) However it should still be added because that plugin is outdated - doesn't work anymore, I updated supported languages and made workflow more flexible (replace/insert mode). Additional translation engines are in testing now.
I used the same Key bindings and Contex menu entries so old users can easily upgrade their experience.

[2]: https://semver.org

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -3186,6 +3186,20 @@
 			]
 		},
 		{
+			"name": "Translator (multi-engine, fast)",
+			"details": "https://github.com/dmytrovoytko/sublimetext-translate",
+			"author": "Dmytro Voytko",
+			"homepage": "https://github.com/dmytrovoytko/sublimetext-translate",
+			"readme": "https://github.com/dmytrovoytko/SublimeText-Translate/blob/master/README.md",
+			"labels": ["translate", "text manipulation"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Translit",
 			"details": "https://github.com/malexer/SublimeTranslit",
 			"releases": [
--- a/repository/t.json
+++ b/repository/t.json
@@ -3197,7 +3197,7 @@
 			]
 		},
 		{
-			"name": "Translator (multi-engine, fast)",
+			"name": "Translator",
 			"previous_names": ["Inline Google Translate"],
 			"details": "https://github.com/dmytrovoytko/sublimetext-translate",
 			"author": "Dmytro Voytko",
--- a/repository/t.json
+++ b/repository/t.json
@@ -3201,8 +3201,6 @@
 			"previous_names": ["Inline Google Translate"],
 			"details": "https://github.com/dmytrovoytko/sublimetext-translate",
 			"author": "Dmytro Voytko",
-			"homepage": "https://github.com/dmytrovoytko/sublimetext-translate",
-			"readme": "https://github.com/dmytrovoytko/SublimeText-Translate/blob/master/README.md",
 			"labels": ["translate", "text manipulation"],
 			"releases": [
 				{
--- a/repository/t.json
+++ b/repository/t.json
@@ -3187,6 +3187,7 @@
 		},
 		{
 			"name": "Translator (multi-engine, fast)",
+			"previous_names": ["Inline Google Translate"],
 			"details": "https://github.com/dmytrovoytko/sublimetext-translate",
 			"author": "Dmytro Voytko",
 			"homepage": "https://github.com/dmytrovoytko/sublimetext-translate",
--- a/repository/i.json
+++ b/repository/i.json
@@ -726,16 +726,6 @@
 			]
 		},
 		{
-			"name": "Inline Google Translate",
-			"details": "https://github.com/MTimer/SublimeText-Google-Translate-Plugin",
-			"releases": [
-				{
-					"sublime_text": "*",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "Inline Python",
 			"details": "https://github.com/apiad/Sublime-InlinePython",
 			"releases": [


dmytrovoytko:
My first plugin for Sublime. Hope I packed it correctly.
[SublimeText-Translate-1.0.0.zip](https://github.com/wbond/package_control_channel/files/10930588/SublimeText-Translate-1.0.0.zip)

dmytrovoytko:
I made some improvements, version 1.0.1, to follow this checklist better:
- Menu and settings improvements (simplification)
- Keys are not bound by default, easy to turn suggested combination on
- Context menu removed to prevent overloaded Sublime interface

repository/t.json
```diff
@@ -3185,6 +3185,20 @@
 				}
 			]
 		},
+		{
+			"name": "Translator (multi-engine, fast)",
+			"details": "https://github.com/dmytrovoytko/sublimetext-translate",
+			"author": "Dmytro Voytko",
+			"homepage": "https://github.com/dmytrovoytko/sublimetext-translate",
+			"readme": "https://github.com/dmytrovoytko/SublimeText-Translate/blob/master/README.md",
```
braver, line 3193: You don’t need to specify homepage and readme if they’re just the defaults for GitHub.

braver:
Please put your package settings under the name of your package as it’s listed in package control.

It looks like it might be worth replacing the existing package with yours, if that one doesn’t work anymore and/or is not maintained. Or maybe you can simply contribute to that repo. Can you get in touch with any of their maintainers, see if we can work something out?

dmytrovoytko:
Thank you for accepring my package, and for the feedback.
The maintainer has no public activities since the last modification (MODIFIED 8 years ago).
Agree, replacing the existing package could be helpful for users  - some people still install it according to [package stats](https://packagecontrol.io/packages/Inline%20Google%20Translate).

On the other hand, I will add a few more translation engines soon, it wont be just "Google translate".
So I don't really know what would be better (again, it's my first experience as a package developer). As a more experienced member of this community what would you suggest? I'm open and really want to contribute better.

braver:
Well, the original package is archived so we don’t need permission to replace it. I think we could do the following:

- Remove the original package from the package control repository.
- Add its name to previous_names for your package.
- Add a message to your package explaining the transition and everything that has changed in term of settings and how to use your package. 
- Maybe link to the original package in your readme.

Since we’re replacing something that doesn’t work I wouldn’t worry too much about breaking changes like removing the context menu. All of that looks like an improvement even, and I believe you’re even compatible in terms of settings.

repository/t.json
braver: So please remove them.

dmytrovoytko: Done!

dmytrovoytko:
Oh, I see that now test fails because of Inline Google Translate:

FAIL: test_repository_package_names('./repository/i.json', ...) (package_control_channel.tests.test.DefaultRepositoryTests)
    File "./tests/test.py", line 100, in wrapper
      return method(self, *params)
    File "./tests/test.py", line 267, in _test_repository_package_names
      self.previous_package_names[pname][2]))
  AssertionError: Package names can not occur as a name and as a previous_name: Inline Google Translate, previously occured as previous_name in ./repository/t.json: Translator (multi-engine, fast)
  
  Ran 9607 tests in 3.062s
  
  FAILED (failures=1)
  
  Should I remove it from /i.json?

dmytrovoytko:
Removed, now tests OK

braver:
Alright, that sets everything up correctly. There is a little bit of code review I would like to go through before we wrap this up.

- You don't need to add `# -*- coding: utf-8 -*-` at the top of the file, Sublime Text runs python3 and utf-8 is assumed/default.
- I presume you added the [no-sublime-package](https://github.com/dmytrovoytko/SublimeText-Translate/blob/master/.no-sublime-package) file because you need to load the JSON files. There is a better way, using the API with `sublime.find_resources()` and `sublime.load_resource()`. Look here for an example: https://github.com/braver/ColorHints/blob/master/lib/pantone.py#L37
- Your package name will be used in various paths, e.g. for settings and overrides. So, your menu entries using [SublimeText-Translate](https://github.com/dmytrovoytko/SublimeText-Translate/blob/master/Main.sublime-menu#L42) won't work because the name of your package is actually "Translator (multi-engine, fast)" (might also be a reason to simply name it "Translator" as that's what you're using in commands and menus). 
- The menu entry for Settings can simply be labelled "Settings" because "Translator" is already in the parent menu entry.
- You may want to have a look at indentation in [Default.sublime-commands](https://github.com/dmytrovoytko/SublimeText-Translate/blob/master/Default.sublime-commands#L2).
- The final tag for [Inline Google Translate](https://github.com/MTMGroup/SublimeText-Google-Translate-Plugin/tags) is 2.1.0, so you might want to start numbering your package starting with 3.0.0, so that all existing users will get your latest version.

dmytrovoytko:
Wow! Thank you so much for your feedback and suggestions!
Agree, renaming to simply "Translator" in t.json, menu, commands.

dmytrovoytko:
I'd like to leave a package unpacked for a while (to be sure it will work correctly with upcoming updates as well), so I left .no-sublime-package) file. The rest is fixed/updated.

braver:
> I'd like to leave a package unpacked for a while (to be sure it will work correctly with upcoming updates as well)

That should not be necessary, and for continued development it should be beneficial to switched to a packaged release right away. You can package your code and test locally as well, see for instance https://packagecontrol.io/docs/creating_package_files

dmytrovoytko:
I need some help to understand.
I created the package file (Translator.sublime-package)
Then I deleted a folder with unpacked version, console shows
> unloading python 3.3 plugin Translator.stTranslate
...and then copied Translator.sublime-package to ~/.config/sublime-text/Installed Packages/ 
(I work in Linux)
Console shows:

reloading python 3.3 plugin Translator.stTranslate
  File "/opt/sublime_text/Lib/python33/sublime_plugin.py", line 308, in reload_plugin
    m = importlib.import_module(modulename)
  File "./python3.3/importlib/__init__.py", line 90, in import_module
  File "<frozen importlib._bootstrap>", line 1584, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1565, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1529, in _find_and_load_unlocked
ImportError: No module named 'Translator.stTranslate'

With these errors I have working Preferences ... Settings
but Tools - Translator are grey (not working), and no Translator commands via Command Palette...

What's wrong?

If package is unpacked  in Packages folder, everything works fine

dmytrovoytko:
Thank you, I updated the code, now works from packaged version as well.

braver:
Great job, thanks for acting so quickly on the feedback. Let's ship this thing 👍🏻

dmytrovoytko:
Awesome!
Thank you so much, Koen, I learned a lot from this experience, 
and hope the package will be useful to the community!

PR #8706, opened 2023-03-10
Transfer ownership of archived Bats package and add ST4 support

keith-hall:
The previous repository was archived, so I forked it to the SublimeText org and added support for ST4.


FichteFoll:
Since the package has officially been abandoned (archived), I don't think we need to ask for permission to replace the package, but I'd like to notify @sptndc just FYI.

PR #8707, opened 2023-03-11
Added MatrixCalculator package

bradybhalla:
- [ ] My package doesn't add key bindings. (see note)

My package is a matrix calculator for Sublime Text.  It can do in-text adition, mutliplication, row reduction, and inversion of matrices.  This package also formats matrices in a readable format and allows easy creation of matrices of any size.

Note: this package requires one key binding, which only runs when the user types something like `5x5` and then presses `tab`.  This is used to create an empty matrix, and because of the required text before the tab, this won't override any keybinds from other packages.

[2]: https://semver.org

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -1047,6 +1047,17 @@
 			]
 		},
 		{
+			"name": "MatrixCalculator",
+			"details": "https://github.com/bradybhalla/sublime-matrix-calculator",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": ["osx"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"details": "https://github.com/jawb/Matrixify",
 			"releases": [
 				{


braver:
Can you tag a release oh your latest changes?

braver:
If you don’t need any specific layout or styling features from minihtml, it’s more user friendly to present your help document as a markdown file. That way the file can be read on your repo, where users typically go first for more information, and users can copy text from it. It’s quite common to make your readme available via a command for instance.

repository/m.json
```diff
@@ -1046,6 +1046,17 @@
 				}
 			]
 		},
+		{
+			"name": "MatrixCalculator",
+			"details": "https://github.com/bradybhalla/sublime-matrix-calculator",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": ["osx"],
```
braver, line 1055: Is this platform limitation still relevant?

PR #8348, opened 2021-08-13
Added Sublack4

munkybutt:
Added Sublack4 package as original Sublack repository seems dead.

Forked from original Sublack plugin, the repository of which appears to no longer be active.
This has changes in it specific to ST4 so it is incompatible with ST3/2.

PR #8711, opened 2023-03-15
Re-Add Sublack4 

munkybutt:

My package is a bridge between pythons black formatter and sublime 4

This is a refactored and updated version of sublack which is no longer actively developed:
https://github.com/wbond/package_control_channel/pull/8348

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -4123,6 +4123,17 @@
 			]
 		},
 		{
+			"name": "sublack4",
+			"details": "https://github.com/munkybutt/sublack",
+			"author": "MunkyButt",
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"details": "https://github.com/yrammos/SubLilyPond",
 			"releases": [
 				{
--- a/repository/s.json
+++ b/repository/s.json
@@ -4125,6 +4125,8 @@
 		{
 			"name": "sublack4",
 			"details": "https://github.com/munkybutt/sublack",
+			"description": "Black formatting for sublime text 4, based on the sublack plugin.",
+			"labels": ["auto-format", "formatter", "black", "python"],
 			"author": "MunkyButt",
 			"releases": [
 				{


braver:
If yours is a straight fork of the existing package, we can simply replace the url with that of your repo.

munkybutt:
> If yours is a straight fork of the existing package, we can simply replace the url with that of your repo.

I had the same thought before adding this PR but I thought it would be problematic as it is not compatible with ST3

braver:
We can handle that by creating a prefixed tag on a version that still is st3 compatible. There are plenty of examples of packages that have st3 users frozen on a specific version while continuing development for st4 only.

munkybutt:
Ah cool - then I agree that would be the most elegant way forward

munkybutt:
The .no-sublime-package is there as the package includes a Python interpreter and as far as I know Python won't run from within a zip file.

rchl:
Why would you bundle a python interpreter? That seems insane. It's easy enough to tell the user to install Python on Windows.

munkybutt:
It enables the plugin to run black with no setup requirements for the user.
This was a pain point with the previous version.
The other approach would have been to have it install python and then pip install black which has way more failure potential than just encapsulating the dependencies within the package.
It is an embeddable interpreter with no unnecessary files included.
I'm not sure this reasoning can be considered insane.

rchl:
It's easy enough to tell the user to install Python by linking to the download page. The plugin could even detect if Python is present and if not trigger a dialog that would tell the user about the issue and provide a button to go the Python download page.

Once it's installed, it's added to the PATH and is available to plugins like that.

Imagine if every plugin like that would come with its own interpreter. It still feels fairly instance to me.
Not to mention that other platforms than Windows already come with Python so it's just redundant weight there.

rwols:
Who are these users that are writing Python files, but don’t have a Python interpreter installed?

munkybutt:
Bud - I am not sure what is going on for you, but your response to this is disproportionate to the reality of what is happening.

One of the main issues I had when I first used this plugin years ago was to get python and black setup and configured with the plugin.
A lot of issues raised on the initial repo are around setup issues.
Several of my colleagues from two companies I have worked at have found it complicated to manage considering the multiple ways python can be installed and their inexperience with python in general.
A lot of people only use embedded python interpreters that are not accessible outside of the host application.
Python is not added to PATH by default.
There are an infinite number of ways python can be setup on a pc and this can complicate the setup.

I am not able to query any usage scenarios beyond my anecdotal experience, but neither are you.

It's only 50mb and we aren't in the 90s anymore, people have TBs of HDD space.
Convenience is the point, not optimal disk usage.

Encapsulating all my dependencies so I have full control and reduce the chance for user error seems sensible to me, to you it is strange enough to be extremely rude about it.

I stand by my reasoning - if you wany to reject the PR because of a 50mb dependency, then reject it and let me get on with my life.

braver:
That's not how we approach packages in Sublime Text. There are dozens of packages for working with Python, JS, php, etc. and none come with the interpreter for that language. Packages that work with git don't come with git etc.. Users might want to combine various packages for a single language, and it would become an unsightly mess if all of those had different versions of runtimes and interpreters running in the background. 

Sure, installing these things is complicated, but if you want everything built in so you don't have to worry about it, there are other solutions. Sublime Text and the package control ecosystem are not that. I know from [experience](https://github.com/SublimeLinter/SublimeLinter) that managing such dependencies and having utilities appear in PATH is a challenge for some; but also that education, documentation and helping each other out is entirely possible and in the end better for everyone involved. You could ship your package as a thinner layer that interacts with black, and let me choose the specific version of black I and my team want to work with independently of your package. And you may not even have to actively maintain that package very much unless black's interface changes or you want to add features.

So, we don't typically approve packages that ship with a `.no-sublime-package` file for those reasons (although there are various other valid reasons for that file). 

Now, in this particular situation we would be reviving a dead package, that already has the "design flaw" of shipping with a fixed and specific version of a dependency. So we could gloss over that since it wouldn't be a "regression". However, although I wouldn't use the word "insane", I think you should reconsider. Package Control would be better with a good black integration package in it, and I think everyone here is just voicing, although not eloquently, the typical Sublime Text community perspective on what such a package would look like. 

Besides that, ... it is clear to you how to make the changes needed to the json files, create the proper tags, and finally replace the existing sublack package?

braver:
Closing this for now. Let us know if you want to re-start the review.

PR #8712, opened 2023-03-16
Added AssistantAI plugin

kanutron:
This is AssistantAI plugin for Sublime Text:
https://github.com/kanutron/AssistantAI

Sublime Text plugin that allows the user to query API end points and perform text operations with their responses.

My package is a plugin in Python 3.8, to allow other plugins to make API requests to endpoints, as ChatGPT.

My package is similar to [OpenAI-sublime-text](https://github.com/yaroslavyaroslav/OpenAI-sublime-text) However it should still be added because it's a more general approach and extensible with other specialized plugins. I expect this package be widely used to provide API endpoints. I'm intended to upload two more packages consuming this AssistantAI package as the base implementation. 

[2]: https://semver.org

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -1978,31 +1978,7 @@
 			"name": "AssistantAI",
 			"details": "https://github.com/kanutron/AssistantAI",
 			"author": "kanutron",
-			"labels": ["assistant", "completion", "chat", "ai", "gpt"],
-			"releases": [
-				{
-					"sublime_text": ">=3000",
-					"tags": true
-				}
-			]
-		},
-		{
-			"name": "AssistantAI-OpenAI",
-			"details": "https://github.com/kanutron/AssistantAI-OpenAI",
-			"author": "kanutron",
-			"labels": ["assistant", "assistant_ai", "openai", "chatgpt", "gpt"],
-			"releases": [
-				{
-					"sublime_text": ">=3000",
-					"tags": true
-				}
-			]
-		},
-		{
-			"name": "AssistantAI-Python",
-			"details": "https://github.com/kanutron/AssistantAI-Python",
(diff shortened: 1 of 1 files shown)


kanutron:
Now it's ready and much cleared.

I did the three changes.

@packagecontrol-bot please proceed.

kanutron:
@packagecontrol-bot please

braver:
Alright, good looking package and different enough in approach that it warrants adding it. As with the OpenAI package, please clearly place a note in the README about how or when potentially code will be sent to third parties. This is relevant because users might not be aware, and might work on code where that is not safe/allowed. See https://github.com/yaroslavyaroslav/OpenAI-sublime-text#note

repository/a.json
```diff
@@ -1974,6 +1974,18 @@
 				}
 			]
 		},
+		{
+			"name": "AssistantAI",
+			"details": "https://github.com/kanutron/AssistantAI",
+			"author": "kanutron",
```
braver, line 1980: This is simply your username, so you can omit it.

kanutron: OK. Removed the 'author' key.

kanutron:
> Alright, good looking package and different enough in approach that it warrants adding it.

Thanks for the feedback and the review.

> As with the OpenAI package, please clearly place a note in the README about how or when potentially code will be sent to third parties. This is relevant because users might not be aware, and might work on code where that is not safe/allowed. See https://github.com/yaroslavyaroslav/OpenAI-sublime-text#note

Indeed, this is an important issue to remark that I missed. I added in the README two notes:
- A "Privacy warning" at the top of the README
- A note about "paid service" in the OpenAI section

I hope this is helps.

braver:
Thanks!

PR #8714, opened 2023-03-24
Update CopyRelativePath to tag-based releases

bpicolo:
- [ ] My package doesn't add context menu entries.  (I think it does and that's the whole reason it exists)

This switches CopyRelativePath from deprecated branch-based versioning to tag-based versioning as recommended by the docs.

The goal is ultimately for this release to become live in Package Control, as I had a contributor add a fix for python versioning in ST4.
https://github.com/bpicolo/CopyRelativePath/releases/tag/1.0.0

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3768,7 +3768,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"branch": "master"
+					"tags": true
 				}
 			]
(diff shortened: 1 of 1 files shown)


braver:
FYI the python-version file is only there to opt in to the 3.8 version, it’s not a requirement.

PR #8717, opened 2023-04-01
Add AlpineJS

deathaxe:

My package is ...

[2]: https://semver.org

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -693,7 +693,7 @@
 		{
 			"name": "AlpineJS",
 			"details": "https://github.com/SublimeText/AlpineJS",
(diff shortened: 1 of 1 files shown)


repository/a.json
```diff
@@ -690,6 +690,17 @@
 				}
 			]
 		},
+		{
+			"name": "AlpineJS",
+			"details": "https://github.com/SublimeText/AlpineJS",
+			"labels": ["syntax"],
```
braver, line 696: The most commonly used label for this would be [“language syntax”](https://packagecontrol.io/browse/labels/language%20syntax).

PR #8720, opened 2023-04-07
Add a new plugin called "Saaqi Plugins"

saaqi:

My package is ...

[2]: https://semver.org

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -21,6 +21,18 @@
 			]
 		},
 		{
+			"name": "Saaqi Plugins",
+			"details": "https://github.com/saaqi/SaaqiPlugins",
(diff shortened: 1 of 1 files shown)


braver:
Thanks for your submission. What's useful for you might not be useful in the exact same combination for others. Bundling like you're doing takes away from the flexibility of the package system, that allows users to find the best fit for how they use Sublime. As it's stands, your package is too specifically personal. What that color scheme and not another? What is the output path of the minifiers and how would that work in my specific use case? 

There are several things you can do instead of publishing your bundle:

- Write a blog post about how you use Sublime and how others can be as productive.
- Write about it on the Sublime users forum.
- Contribute to one of the existing snippets packages.
- Contribute to the core default CSS and HTML packages, they might be interested in your completions and/or snippets.

PR #8722, opened 2023-04-09
Add CloseAllNoConfirm package

ddbln:

My package is "CloseAllNoConfirm": Instantly close all unsaved (and saved) files in Sublime Text, ditching all confirmation dialogs. Cautious users can opt for a single prompt, if desired.

My package is similar to some existing packages that close files or windows. However, it should still be added because it specifically focuses on closing all files (saved and unsaved) without confirmation dialogs, providing a more streamlined and efficient user experience. This can be helpful in situations where there are many or huge numbers of files open (for example after a search and replace action in multiple files / a big project folder...) that one does not want to save. Saving all is easy and native in Sublime Text, but closing all (unsaved) files requires the user to confirm a dialogue for each file, hence this plugin. Additionally, it offers an optional single confirmation prompt to make it attractive for users who are more cautious, too.

 I have reviewed the guidelines and ensured that my package:
• Has valid semantic version tags.
• Has a README that is understandable for new users and is in English.
• Does not include unnecessary files, such as package-metadata.json and *.pyc files.
• Has all commands discoverable via the Command Palette in a Default.sublime-commands file.
• Does not include unnecessary keybindings.
• Has Python plugins placed in the root folder of the repository.
• Has no dependencies on other packages.
• Does not use the .no-sublime-package file.

I believe my package adheres to the review guidelines and is ready for submission.

For updates to existing packages:
If your package isn't using tag based releases,
please switch to tags now.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1585,6 +1585,17 @@
 			]
 		},
 		{
+			"name": "Close All without Confirmation",
+			"details": "https://github.com/ddbln/CloseAllNoConfirm",
+			"labels": ["file navigation", "files", "close files", "file close", "commands", "productivity", "utilities", "refactoring"],
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Close Oldest File",
 			"details": "https://github.com/jturcotte/SublimeCloseOldestFile",
 			"releases": [
@@ -1601,17 +1612,6 @@
 				{
 					"sublime_text": "<3000",
 					"branch": "master"
-				}
-			]
-		},
-		{
-			"name": "Close All without Confirmation",
-			"details": "https://github.com/ddbln/CloseAllNoConfirm",
-			"labels": ["File Management", "Close Files", "Files", "No Confirm", "Productivity", "Utility", "Refactoring"],
-			"releases": [
-				{
-					"sublime_text": ">=3092",
-					"tags": true
 				}
 			]
 		},
--- a/repository/c.json
+++ b/repository/c.json
@@ -1605,10 +1605,8 @@
 			]
 		},
 		{
-			"name": "CloseAllNoConfirm",
 			"details": "https://github.com/ddbln/CloseAllNoConfirm",
-			"description": "Instantly close all files (saved & unsaved) without confirmations.",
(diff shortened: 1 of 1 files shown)


repository/c.json
```diff
@@ -1604,6 +1604,18 @@
 				}
 			]
 		},
+		{
+			"name": "CloseAllNoConfirm",
+			"details": "https://github.com/ddbln/CloseAllNoConfirm",
+			"description": "Instantly close all files (saved & unsaved) without confirmations.",
```
braver, line 1610: You don't need to provide a description if it's the same as your GitHub repo's description. That way you can easily change it. In fact, even the name can be omitted in your case.

braver:
Your main menu entry is not optimal. If you use the "edit_settings" command with a "base_file" instead of the "open_file" command, users can open the settings (and keybinding) with your defaults (and documentation) right next to their own settings.

I would also propose to caption your command in a more readable name. CamelCase with multiple words ends up looking like an unreadable name, and the command should be more about doing something ("Close all without confirmation") and less about the name of a package.

repository/c.json
ddbln: Thank you!

ddbln:
Thank you for your valuable feedback! I have made the suggested changes:

- I have updated the main menu entry to use the "edit_settings" command with a "base_file" instead of the "open_file" command.
- I have also updated the command caption to a more readable name: "Close All without confirmation", like you suggested (thanks again).

This is implemented in the latest release (tagged 1.2.0). I hope these improvements address your concerns. Please let me know if you have any further suggestions or feedback.

braver:
Package looks good, thanks. All that's left is to clean up your entry in [repository/c.json](https://github.com/wbond/package_control_channel/pull/8722#diff-b5eeeee20c6cd05e78128c3c469949445ba673f10ad80329baf3241b1002c27f) a bit

braver:
No problem, that's what we're here for. 2 things:

- the name change means you need to update your position in the file to keep it sorted (that's the failing test)
- labels are typically all lower case (see screenshot)

Scherm­afbeelding 2023-04-27 om 19 49 47

ddbln:
I have updated the position and changed the case. Thanks a lot for pointing that out!

PR #8723, opened 2023-04-09
add "JS Template String Converter" plugin

predragnikolic:

My package is ... 
a plugin that converts converts regular strings to template strings as soon as the users types `${`
Please see the readme for more info https://github.com/predragnikolic/JS-Template-String-Converter#readme
 
it can be used in JavaScript, TypeScript, Svelte, Vue, JSX, TSX files.

<details>
<summary>I added a keybinding.</summary>

My package does add one keybinding, and that is this:
```

[
	 { "keys": ["{"], "command": "insert_snippet", "args": {"contents": "{$0}"}, "context":
        [
            { "key": "setting.auto_match_enabled" },
            { "key": "selector", "operand": "source.js | source.jsx | source.ts | source.tsx | text.html.ngx | text.html.svelte | text.html.vue" },
            { "key": "selection_empty", "match_all": true },
            { "key": "preceding_text", "operator": "not_regex_contains", "operand": "\\\\$", "match_all": true }
        ]
    },
]
```

I added a lot of context checks to not mess with users defined keybindings.
What this keybinding does it it expands `{` to `{}` if the preceding_text text is `$`. 
I added this keybinding to fix case 1.

Usually ST is good at adding the closing bracket automatically,
but st will not add the closing bracket automatically if there is no space before the quote:
case 1 
```
"hello $|" -> type `{`
"hello ${|" -> no closing bracket because there is no space in front of " 
```
case 2
```
"hello $| " -> type `{`
"hello ${|} " -> a closing bracket because there is a space in front of " 
```

This plugin will make case 1 to behave like case 2:
```
"hello $|" -> type `{`
"hello ${|}" -> because of the keybinding from this plugin
```

</details>

[2]: https://semver.org

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -941,6 +941,7 @@
 		{
 			"name": "JS Template String Converter",
 			"details": "https://github.com/predragnikolic/JS-Template-String-Converter",
+			"labels": ["javascript", "typescript", "jsx", "vue", "svelte"],
 			"releases": [
 				{
 					"sublime_text": ">=4070",
--- a/repository/j.json
+++ b/repository/j.json
@@ -939,6 +939,16 @@
 			]
 		},
 		{
+			"name": "JS Template String Converter",
+			"details": "https://github.com/predragnikolic/JS-Template-String-Converter",
+			"releases": [
+				{
+					"sublime_text": ">=4070",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "JS Var Shortcuts",
 			"details": "https://github.com/twolfson/sublime-js-var-shortcuts",
 			"labels": ["formatting", "text manipulation", "js", "javascript", "var", "variable"],


predragnikolic:
I have added the License file

rchl:
Make your package run in 3.8 host since you are importing from `typing`.

predragnikolic:
added a `.python-version` file

FichteFoll:
The package has the same problem domain as https://packagecontrol.io/packages/JavaScriptTemplateLiteralEnhancements but uses a sufficiently different approach that it doesn't make too much sense to combine the packages.

Other than that, searching for `JavaScript` currently wouldn't find your package since it doesn't occur in either the package's name nor its description. Do you want to adjust the name or add it to the description?

predragnikolic:
I will add a description as soon as I can.
(probably tomorrow)

FichteFoll:
FYI, labels are not indexed for search, i.e. it still won't show up when searching for "JavaScript". That said, since the description is taken from your github repo, you can easily modify it even after it is merged, so I'm going ahead with it. Feel free to make follow-up PRs, as usual.

PR #8724, opened 2023-04-09
New Package 'Gitignore Extended'

Isaac-the-Man:

My package is similar to [Gitignore ](https://packagecontrol.io/packages/Gitignore) However it should still be added because:
- offers more templates includes community and global templates
- enables the user to create/edit/delete custom templates, which can be reuse afterwards

[2]: https://semver.org

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -793,6 +793,17 @@
 			]
 		},
 		{
+			"name": "Gitignore Extended",
+			"details": "https://github.com/Isaac-the-Man/Sublime-Gitignore-Extended",
+			"labels": ["vcs"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Gitignored File Excluder",
 			"details": "https://github.com/ExplodingCabbage/sublime-gitignorer",
 			"releases": [


braver:
Have you considered contributing these features to the existing package?

braver:
Alright, that's reasonable. 

- Have you looked into using `.gitattributes`?
- Why do you need [.no-sublime-package](https://github.com/Isaac-the-Man/Sublime-Gitignore-Extended/blob/master/.no-sublime-package)? You can probably access your defaults using [sublime.load_resource()](http://www.sublimetext.com/docs/api_reference.html#sublime.load_resource) etc.
- You have settings, but don't expose the settings via the main menu, nor via the command palette. 
- Please check your paths and labels: the package is called "Gitignore Extended", ie. without [the hyphen](https://github.com/Isaac-the-Man/Sublime-Gitignore-Extended/blob/master/commands.sublime-commands#L3).

Isaac-the-Man:
@braver  Appreciate the feedback. Can you point me to a example usage of `sublime.load_resource()`? The API documentation doesn't really help. Thanks.

braver:
I make use of it here: https://github.com/braver/ColorHints/blob/master/lib/pantone.py#L34

First find_resource to find the file by its filename, then load_resource to read the file.

PR #8727, opened 2023-04-12
Update Wallaby packages.json location

smcenlly:
Update Wallaby packages.json to reside on github instead of s3 to fix s3 removing support for TLS 1.0

My packages are ...

Wallaby and Quokka (existing packages, just updated the manifest location):
* https://packagecontrol.io/packages/Wallaby
* https://packagecontrol.io/packages/Quokka

### DIFF
--- a/channel.json
+++ b/channel.json
@@ -116,6 +116,7 @@
 		"https://raw.githubusercontent.com/timonwong/sublime_packages/master/packages.json",
 		"https://raw.githubusercontent.com/tomascayuelas/coolcodescheme/master/packages.json",
 		"https://raw.githubusercontent.com/vkocubinsky/sublime_packages/master/packages.json",
+		"https://raw.githubusercontent.com/wallabyjs/sublime-text-releases/main/packages.json",
 		"https://raw.githubusercontent.com/wallysalami/QuickLook/master/packages.json",
 		"https://raw.githubusercontent.com/weslly/sublime_packages/master/packages.json",
 		"https://raw.githubusercontent.com/wuub/requirementstxt/master/packages.json",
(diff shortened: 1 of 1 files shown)


smcenlly:
@braver - I had an email notification: 

> Please just add your package to [r.json](https://github.com/wbond/package_control_channel/blob/master/repository/r.json).

But I can't see any reference of your comment here. I can see that the change is failing because of unsorted repository names and have fixed that issue.

Please note, we are replacing an existing repository and _not_ creating a new one. If more changes are required, please let us know.

braver:
No, sorry, I was going through multiple reviews and put that comment in the wrong one. 

This one is fine, besides the sorting thing you seem to have fixed already.

braver:
Hm, seems we're not out of the woods yet.

smcenlly:
@braver - the `302 Found` response is normal/expected for github releases. If you paste the URL into your browser, you will see it loads correctly.

We had already updated our previous package control manifest and tested that these URLs work correctly. Besides your tests failing because of the `302`, I don't see any problem?

braver:
I'm not sure why 302 is considered an error here. It's quite common for GitHub URLs to be redirected in this way, and typically we support that just fine. @FichteFoll do you know more about this? Is the test failing for no reason, or do we need to do something about it (ie. change the URLs to the actual URLs of the package files)?

FichteFoll:
The error on 302 is likely related to the package review wrapper. If adding the repository.json file to PC's preferences and installing it manually works fine, then it shouldn't be an issue.

braver:
Everything checks out as far as I can tell, and installing certainly works just fine, so let's merge it.

PR #8728, opened 2023-04-12
Add Pluto Syntax Highlighting

Sainan:

My package is syntax highlighting for the Pluto language, a fork of Lua. 

[2]: https://semver.org

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -1750,7 +1750,7 @@
 			]
 		},
 		{
-			"name": "Pluto Syntax Highlighting",
+			"name": "Pluto",
 			"details": "https://github.com/PlutoLang/Syntax-Highlighting",
 			"labels": ["language syntax"],
 			"releases": [
--- a/repository/p.json
+++ b/repository/p.json
@@ -1750,6 +1750,17 @@
 			]
 		},
 		{
+			"name": "Pluto Syntax Highlighting",
+			"details": "https://github.com/PlutoLang/Syntax-Highlighting",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "PM2-Snippets",
 			"details": "https://github.com/riyadhalnur/PM2-Snippets",
 			"releases": [


repository/p.json
```diff
@@ -1749,6 +1749,17 @@
 				}
 			]
 		},
+		{
+			"name": "Pluto Syntax Highlighting",
```
braver, line 1753: You can simply name the package after the language itself, "Pluto" and leave out the "syntax highlighting" because it does quite a bit more than highlighting.

braver:
Always a bit of a shame to see new packages use tmLanguage, the lowest common denominator for such things. Not nice to maintain, slows down rendering, etc. Especially since it's a superset it looks like it would be nice to be able to match the default Lua highlighting (which is probably more advanced). But I understand it, not a problem.

Sainan:
tmLanguage gets too much hate; it's still the format supported by literally every text editor ever made. It doesn't mean I can't get _inspired_ by some of the things in the built-in sublime syntaxes. And at the end of the day, it's me who has to maintain it, not you. :wink:

Anyway, I think "Pluto Syntax Highlighting" is fine because:
1. Snippets are often bundled with syntax highlighting so I think it's accurate enough without being too verbose
2. This is only 1 bit of tooling for Pluto

braver:
Sure, although I think it's holding some editors back. But besides the point of this PR certainly 🤗 

Please do change the name of the package though. By convention, language support packages are simply named after the language they support, although over the years some have slipped with longer names. "Syntax highlighting" is also a misinterpretation of the features of a syntax/language package: the highlighting is only one aspect. In Sublime Text a syntax enables autocomplete, symbol lists, goto definition, and more, in addition to the colors. You're not going to put all of that in the name of your package right? 😉

Sainan:
You're really dying on a stupid hill here. This project is called "Pluto Syntax Highlighting". That's what it's called in the VS Code marketplace as well. Yes, it adds "language support" for Pluto, but it doesn't _add_ warnings, errors, completions, symbol lists, goto definitions, whatever. We have other tooling for some of these things.

braver:
> completions, symbol lists, goto definitions, whatever

Well yes it does, because in Sublime Text a syntax unlocks all of those features. In a more naive way than LSP of course, but still.

Sainan:
I see. It doesn't exactly understand what keywords are, so it doesn't seem to provide any extra completions, but I guess it does provide goto definitions.

Regardless, the project is called "Pluto Syntax Highlighting." I understand you think the "Syntax Highlighting" part is just a useless suffix, but it's clearly a convention that most every participant in this ecosystem has agreed upon, and it does provide a general overview of what the package provides — even if it doesn't disclose 100% of the unlocked functionality.

braver:
Well it's not useless, but it's also not correct and not useful. Again, by convention we leave that suffix out in package control (although I admit some hive slipped past a proper review in this sense over the decade+ of this ecosystem).  Let me give you some examples of "syntax highlighting" packages simply named after the language they support:

ASP
ActionScript
AppleScript
Batch File
C#
C++
CSS
Clojure
D
Diff
Erlang
Git Formats
Go
Graphviz
Groovy
HTML
Haskell
JSON
Java
JavaScript
LICENSE
LaTeX
Lisp
Lua
Makefile
Markdown
Matlab
OCaml
Objective-C
PHP
Pascal
Perl
Python
R
Rails
RestructuredText
Ruby
Rust
SQL
Scala
TCL
Text
Textile
XML
YAML
4GL
65816
ABAP
ABC Notation
Abella
ACTG
Active4D
Ada
Alan
AlgoLang
AngelScript
Ansible
Ant
ARM Assembly
Astro
Asymptote
AutoLISP
Autotools
Awk
Baan
Ballerina
Behat
Bicep
Bison


... Need I go on?

Sainan:
All of these have one thing in common: They were specifically made for Sublime Text. It might be wise to consider that there are external yet closely related ecosystems and as such there might be confusion with enforcing this naming convention.

Also, to be quite unfair to myself, Pluto is a rather niche language in the grand scheme of things, the name _should_ probably be more verbose, such as "PlutoLang.org Syntax Highlighting"

braver:
> They were specifically made for Sublime Text

As is package control. All users of package control are using Sublime Text. That's the context here, and that's why we have particular naming conventions here. You can name your repo any way you like, but users looking to work with `<a language>` in Sublime Text, should have to look no further than installing the package called `<a language>` in package control. Regardless of that the same package is called for another editor or whatever. 

I don't want to die on this hill, I'm not being a jerk, but I _am_ here to maintain package control and enforce certain rules. Naming conventions are important in this context wether you like it or not. We get submissions every other month from developers who have different ideas about how packages should be named, what packages should be allowed to do, what I should be doing in a review, etc. We do not bend to that, because we would lose the value of having a managed, redacted package eco system. This process is here to help developers like yourself what the Sublime Text ecosystem is about, how it is different from say VS Code, and how to play nice with everything else that users might be using ST for. I don't mind explaining. And I hope you don't mind listening and maybe learn something. 

So please don't be a jerk and leave the snark out your commit messages.

Sainan:
I understand you want to "enforce rules", but also understand what it looks like to me when you add

>  If my package is a syntax it is named after the language it supports (without suffixes like "syntax" or "highlighting").

_right after_ this PR was opened.

And maybe the fact that the PR right after this sorta just ignored that step maybe tells you how the other participants feel with this rule you just came up with on the spot. 🙃

braver:
Yeah, I can see that. That checklist is there to prevent repeated explanations of the same thing, and this topic has returned a lot recently so I thought it was worth adding it to that checklist. It gets discussed in PRs every now and then and it simply needs explaining as it's particular to Sublime Text and Package Control. So please don't take it personal, it isn't meant to be.

PR #8729, opened 2023-04-14
Added Ember & Glimmer Syntax

robclancy:

My package is syntax highlighting for Ember/Glimmer templates. https://github.com/robclancy/sublime-ember-syntax

My package is similar to Ember Syntax and Handlebars syntax however it should still be added because the Handlebars one is for an older syntax and Ember Syntax is abandoned (even the repository is gone).

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -781,6 +781,17 @@
 			]
 		},
 		{
+			"name": "Ember & Glimmer Syntax",
+			"details": "https://github.com/robclancy/sublime-ember-syntax",
+			"labels": ["language syntax", "ember.js", "glimmer", "handlebars"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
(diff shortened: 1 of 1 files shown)


repository/e.json
```diff
@@ -780,6 +780,17 @@
 				}
 			]
 		},
+		{
+			"name": "Ember & Glimmer Syntax",
```
braver, line 784: You can simply name the package after the language itself, "Ember & Glimmer" and leave out the "syntax" because syntaxes do quite a bit more than highlighting in Sublime Text. 

If you add `"previous names": ["Ember Syntax"]`, and remove their entry lower in this file. That way we can replace the existing package with yours and upgrade any existing users to your more up to date implementation.

Is your syntax a fork of handlebars in some way? How does it relate? Because your syntax still presents itself as a handlebars syntax, which might confuse people and other packages if that's not correct.

PR #8730, opened 2023-04-17
New Package: Excel Syntax

axemonk:
This is my first time releasing anything on GitHub, not to mention a package manager.

I've tried to follow all guidelines to the best of my ability. Thank you in advance for your patience if I've made mistakes anywhere.

- [X] If my package is a syntax it is named after the language it supports (without suffixes like "syntax" or "highlighting").

My package is a Sublime Syntax for Excel formulas.

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -1514,6 +1514,16 @@
 			]
 		},
 		{
+			"name": "Excel Syntax",
+			"details": "https://github.com/axemonk/Excel-Syntax-Sublime",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Exclude Paths",
 			"details": "https://github.com/kemayo/sublime-text-exclude-paths",
 			"labels": ["project", "indexing", "exclude", "paths"],
--- a/repository/e.json
+++ b/repository/e.json
@@ -1505,8 +1505,8 @@
 			]
 		},
 		{
-			"name": "Excel Syntax",
-			"details": "https://github.com/axemonk/Excel-Syntax-Sublime",
+			"name": "Excel formula",
+			"details": "https://github.com/axemonk/Excel-formula",
 			"releases": [
 				{
 					"sublime_text": "*",
--- a/repository/e.json
+++ b/repository/e.json
@@ -1505,21 +1505,21 @@
 			]
 		},
 		{
-			"details": "https://github.com/SublimeText/ExcelExec",
-			"releases": [
-				{
-					"sublime_text": "<3000",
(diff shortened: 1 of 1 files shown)


braver:
If I understand this correctly, it will highlight excel formulas in plain text files? That's a specific use case, very cool idea. I would assume a user would manually select this syntax to highlight a specific file, ie. not something that works automatically because it's not a language with a specific type of filename etc.

Some feedback

- A better name might be Excel Formula? We try to prevent having "syntax" suffixes or things like that, although here it makes sense.
- Why is the suffix for your scopes `.cs`? Wouldn't `.excel` make more sense?
- Note that syntaxes can also be selected via the command palette (might be a useful addition to your readme).

axemonk:
Thank you for the feedback, @braver!
Yes, users would select the syntax manually in a text file containing Excel formulas.

Preview:

image

1. The actual package name is simply "Excel". I'll change it to "Excel Formula", I think that's better, too.
2. I assumed that scopes needed to be defined/developed, and by using ".cs" I was simply reusing the C# scope 🤦‍♂️. I will fix this as well. 
3. Noted. Thank you.

Will get back to work on this tomorrow and follow up with details on any changes I've made.

braver:
You could read this if you have some time: https://www.sublimetext.com/docs/scope_naming.html

Basically the scopes are there to tell Sublime Text itself, other packages, and the syntax highlighting, what a sequence of characters means. So is it a function definition, a function call, a function argument, a variable, some other keyword, etc. Ideally you pick the right names (regardless of what color it would be) to describe the code semantically. That ensures the highlighting is consistent across different languages. You add `.excel` at the end to make it specific for your language, and that allows other packages and the syntax highlighting to implement a special case for your language if they need to (e.g. your functions might be slightly different from functions in other languages). 

Now with a language such as excel formulas, it might not always fit the scope naming guidelines, but it's always a good idea to keep them in mind. So don't scope your strings as keywords just to make them purple for instance (completely random example, just to illustrate).

axemonk:
Thank you so much for the feedback (again).

1. I spent some time today reworking the scopes to be in line with convention.
2. I also added an MIT license (thanks bot).
3. Also via bot, not sure if it's worth including a `.tmlanguage` fallback file or how I would go about doing that. Can't find anything relevant in the Syntax docs.
4. Specified `version: 2`
5. Renamed repo and plugin to be Excel-formula and Excel formula, respectively.
6. Tagged changes in new release `1.1.0`

New Preview:

image

braver:
Great job, that looks really good.

You can ignore that comment from the bot about the sublime-syntax support, 3092 is ancient history.

PR #8731, opened 2023-04-19
Update the Bow Tintin repository

papac:

My package is for updating the existing package

[2]: https://semver.org

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -1291,8 +1291,7 @@
 			"releases": [
 				{
 					"sublime_text": ">=3103",
-					"tags": true,
-					"platforms": "*"
+					"tags": true
 				}
 			]
 		},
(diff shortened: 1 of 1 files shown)


repository/b.json
```diff
@@ -1286,12 +1286,13 @@
...
-			"details": "https://github.com/papac/bow-tintin-sbt3",
-			"labels": ["php", "Bow", "Tintin", "language syntax", "Bow Tintin"],
+			"details": "https://github.com/bowphp/sublimetext-extension",
+			"labels": ["php", "bow", "bowphp", "Tintin", "language syntax", "Bow Tintin"],
 			"releases": [
 				{
 					"sublime_text": ">=3103",
-					"tags": true
+					"tags": true,
+					"platforms": "*"
```
braver, line 1295: if it's "*" no need to specify (ie. remove this line)

PR #8732, opened 2023-04-19
added OpenAI assistant plugin

fkuehnel:
- [ ] I use [.gitattributes][3] to exclude files from the package: images, test

My package is similar to OpenAI completion and AIassistant. However, it should still be added because
the code in those packages is horribly written by amateurs, non-maintainable and lack convenient functionality.

[2]: https://semver.org

### DIFF
--- a/repository/o.json
+++ b/repository/o.json
@@ -680,6 +680,18 @@
 			]
 		},
 		{
+			"name": "OpenAI assistant",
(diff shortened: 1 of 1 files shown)


braver:
Eh, thanks, but that's not how you convince others. Also, this is open source. Since you're such an amazing non-amateur, please enlighten the builders of said packages and contribute to their work.

PR #8733, opened 2023-04-23
Add Typst

hyrious:
      It adds similar completions for `*bold*`, `_italic_` like the markdown package, it should not affect much.

My package is basically only a sublime-syntax file that highlights http://github.com/typst/typst.

[2]: https://semver.org

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -3777,6 +3777,17 @@
 					"tags": true
 				}
 			]
+		},
+		{
+			"name": "Typst",
+			"details": "https://github.com/hyrious/typst-syntax-highlight",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"tags": true
+				}
+			]
 		}
 	]
 }


hyrious:
It does have defined 'backspace' key bindings, but all of them are scoped under 'source.typ' or 'punctuation.definition.raw.begin.typ', which should be ok.

jwortmann:
> * It has suffixes.. But I saw many other packages are doing this too?

The fact that other packages did it wrong in the past should be no reason to keep doing it wrong in the future. Someone should add a feature to packagecontrol-bot to directly emit an error if the name of a new package ends in "Syntax" or "Syntax Highlight". Then @braver doesn't have to repeat the rules over and over again for new syntax packages.

The key bindings seem fine to me and the syntax itself seems to be of pretty high quality after a quick look. For the Comments.tmPreferences, you could consider to add the block comment delimiters as well, then the default key binding for `"toggle_comment", "args": { "block": true }` would work too. See e.g. https://github.com/sublimehq/Packages/blob/56adac1c09d10e2008a836be9b2677921a0c8543/JavaScript/Comments.tmPreferences#L16-L27

hyrious:
@jwortmann Thanks! Do you mean it should be renamed to "Typst"? I was afraid to use such a short name.

Will update the comments config too.

jwortmann:
Yes, looks good now.

One other thing I noticed, is that you use `.typ` as the scope suffix in the syntax (for example the base scope `source.typ`), which was probably chosen after the file extension. There is no strict convention about how to choose the scope suffix, and there are indeed a few examples where it is the file extension, like `.cs` for C\# or `.js` for JavaScript. But it seems to be a bit more common to use the actual language name instead, for example `.python` or `.markdown`. And I found a simple [syntax definition for VSCode](https://github.com/nvarner/typst-lsp/blob/0b88dfd7494823b4eff5904265027fd81faf5916/addons/vscode/typst.tmLanguage.json), which also uses the full language name `.typst` as the scope suffix. So it might be more convenient to follow this example, just in order to use a common standard, and maybe for the (unrealistic?) case if there is a color scheme which wants to target special scopes from this syntax and is ported between editors. But I would say the final choice is up to you, as the syntax author.

Note that I'm just some random user without any executive force in this repository, but maybe my tips are a bit helpful.

hyrious:
Yes, it sounds right to have `source.typst`, will do it.

jwortmann:
Oh, and another thing that just came to mind; for this syntax it might actually be more appropriate to use the `text.typst` base scope, rather than `source.typst`. Because Typst seems to be a markup system/language quite similar to Markdown, and those syntaxes usually use the `text` base scope (e.g. Markdown, reStructuredText, LaTeX). A consequence of this would be that Sublime doesn't show the autocompletion popup with random words from the buffer by itself, which might be distracting if you write a longer markup text. And also word wrapping is enabled by default for the `text` base scope.

So, whether you want to change that scope might also be a matter of preference, and doing so would imply that it doesn't match the base scope from the VSCode syntax anymore, but I think here it would probably be reasonable.

hyrious:
Yeah that's reasonable. About the consequences:

- No autocompletion is not that bad, since you can always trigger that with <kbd>ctrl+space</kbd>. (Aha, unless you're on macOS and this shortcut is registered by default as the OS level input method switcher)
- When word wrapping is "auto" it matches what you said. From my personal experience, I'd prefer to set that to 'false' i.e. turn off word wrapping to all kind of texts or sources. Because it does not play well with CJK words and/or long urls.

So here comes the trades:

Use `text.typst`:

It matches the language self-evaluation: A markup-based typesetting system. LaTeX is doing this too. But if you want to trigger completion or turn off word wrapping, you will have to learn some configurations.

Use `source.typst`:

It can also matches the language definition: all markup syntax sugar are just calling builtin functions, the language itself is still a strict programming language. And you don't have to apply configurations in most of the time.

hyrious:
🤔 Any progress?

hyrious:
🤔 One drawback of using `source.*` is that spell check now only applies to works in `string` scope. Maybe I would just use `text.*` instead.

repository/t.json
```diff
@@ -3761,6 +3761,18 @@
 					"tags": true
 				}
 			]
+		},
+		{
+			"name": "Typst",
+			"details": "https://github.com/hyrious/typst-syntax-highlight",
+			"author": "hyrious",
```
braver, line 3768: You don't need to provide the author here, package control will pick up on your username automatically.

repository/t.json
```diff
@@ -3761,6 +3761,18 @@
...
+			"labels": ["tex", "typst"],
```
braver, line 3769: The common label for syntax, highlighting, language support, etc, packages is "language syntax".

braver:
> 🤔 One drawback of using source.* is that spell checking only applies to words in string scope. Maybe I would just use text.* instead.

Ha, you can't have both automatically. I think you made the right choice going with `text`. You can help your users set up completions, wrapping, and other stuff by putting the information or links to the Sublime Text documentation in your readme. Those things are not specific to your package, but most users won't know all the details of how to make ST do what you want. 

Looks good, a lot of effort and attention went into this 👍🏻

repository/t.json
hyrious: Removed.

repository/t.json
hyrious: Ok.

PR #8736, opened 2023-04-29
Add TimeNotify

halfmoonvic:

My package is view and remind time

[2]: https://semver.org


braver:
Your readme doesn't clearly explain how to use your package, and what that would result in. In other words, why would someone use it? What does it do?

Please use `edit_settings` also in the main menu entry, so that the settings files open in split view.

halfmoonvic:
I have updated my readme, please check

braver:
Thanks for the update. Please address my feedback regarding how settings are opened.

halfmoonvic:
> Thanks for the update. Please address my earlier feedback regarding how settings are opened.

I don't understand. Are you saying to provide a screenshot of the settings file? In that case, I have updated my readme document and listed a set of settings. Is that okay?

braver:
No, you're using open_file to open the default settings via one menu entry, and a second one for the user settings. That's the ST2 way of doing it from more than 8 years ago. You should use the edit_settings command instead, which allows you to open settings in 2 panes. There are various packages out there that you can use as [an example](https://github.com/braver/SideBarTools/blob/master/Main.sublime-menu#L17).

halfmoonvic:
> No, you're using open_file to open the default settings via one menu entry, and a second one for the user settings. That's the ST2 way of doing it from more than 8 years ago. You should use the edit_settings command instead, which allows you to open settings in 2 panes. There are various packages out there that you can use as [an example](https://github.com/braver/SideBarTools/blob/master/Main.sublime-menu#L17).

ok, I have changed

PR #8737, opened 2023-04-29
Add SymbolBalloon

matsukido:

My package is displaying current function names.


braver:
Nice work. There are packages in the same vein, but haven't seen this specific approach yet.

PR #8738, opened 2023-04-29
Add Sublitesse Color Scheme

hhofner:

My package is ...

[2]: https://semver.org

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -4999,13 +4999,11 @@
 		},
 		{
 			"name": "Sublitesse Color Scheme",
-			"description": "A calm, green-ish tint Color Scheme. Based on Ant Fu's VSCode Theme",
 			"details": "https://github.com/hhofner/sublitesse",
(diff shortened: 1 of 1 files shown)


repository/s.json
```diff
@@ -4997,6 +4997,19 @@
 				}
 			]
 		},
+		{
+			"name": "Sublitesse Color Scheme",
+			"description": "A calm, green-ish tint Color Scheme. Based on Ant Fu's VSCode Theme",
```
braver, line 5002: If you put a description on your GitHub repo, you can remove it here.
Also, you can remove the “issues” here since it’s just the default url for a GitHub repo.

braver (changes requested):

repository/s.json
hhofner: Thank you! I've added the changes. I also fixed the supported version (https://github.com/wbond/package_control_channel/pull/8738/commits/03051ed6148db9ddb666a52480457e86b0fc983e)

braver (approved):

PR #8739, opened 2023-04-29
move Terminal

braver:
now moved to https://github.com/SublimeText/Terminal


braver:
@FichteFoll this basically implements what you suggested in Discord (I think). I'd like to look at the schema version, but the key bindings I can't help without breaking changes in the package.

FichteFoll:
Yes, the schema version definitely needs to be updated (as well as the contents, such as replacing the tags details urls with `"tags": true`. I don't think the rest needs to be changed.

FichteFoll:
In fact, you might as well resolve ChannelRepositoryTools, Package Coverage and Tortoise into the main channel directly.

I should resolve my own `repositories.json` file as well …

braver:
Well that took a little doing, but at least it sort of passes now.

braver:
> In fact, you might as well resolve ChannelRepositoryTools, Package Coverage and Tortoise into the main channel directly.

I figured they were either too niche, or too dead (ChannelRepositoryTools , Tortoise) to bother. I tried to upgrade the json using ChannelRepositoryTools and it was beyond useless. 

With the packages.json in our hands we can do so whenever we feel like it. Right now I don't see the need to clutter up the Sublime Text org with repos nobody uses. Of course everyone needs to stop using these private json's, but that's a discussion for a different time 😉

FichteFoll (approved):
> I figured they were either too niche, or too dead (ChannelRepositoryTools , Tortoise) to bother. I tried to upgrade the json using ChannelRepositoryTools and it was beyond useless.

I didn't mean forking them into the ST org, just moving the repository entries into the main json files, but as you say, since it's in our hands now, we can do it any time.
In fact, it might even be a good idea to have a more generic name for this repo than `wbond-packages` so that we open up the possibility of putting further files in there for a migration period, though I don't forsee that being used in the near future. We can still do it later as well, of course.

On Discord, Will also mentioned he would be open to moving the repo, but this change doesn't depend on it and it can be done at a later point. For the sake of preserving the issues and other PRs, I'd suggest doing that sooner rather than later, however.

Resolve these notes at will. This is mergeable from my perspective.

FichteFoll:
> Ideally he transfers the repos, that’s cleanest all around.

I agree, though he won't be able to transfer to the ST org if a fork of the same name already exists. :sweat_smile: Alternatively, he could offer it directly to you or me, then we delete or rename the fork and transfer the original repo to the ST org. I suggest discussing this with him in discord/twitter DMs.

braver:
> if a fork of the same name already exists

Good thing I renamed the fork then 😉

PR #8741, opened 2023-05-03
Add M3U

freearhey:

My package is syntax highlighter for [M3U](https://en.wikipedia.org/wiki/M3U) playlists.

My package is similar to [M3U Syntax](https://packagecontrol.io/packages/M3U%20Syntax) however it should still be added because because it has support for more attributes.

[2]: https://semver.org

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -2,6 +2,17 @@
 	"schema_version": "3.0.0",
 	"packages": [
 		{
+			"name": "M3U",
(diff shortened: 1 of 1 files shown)


braver:
I'd rather not add another package, especially since the existing one is ancient. If you can't contribute there (have you tried?) then we can consider replacing it.

braver:
I'm going to close this for now. If you have no luck with contributing to the existing package for some reason, let us know and we can discuss alternatives.

PR #8743, opened 2023-05-06
Added Alter Text plugin

cibinmathew:

My package is ...

Quickly alter text in Sublime Text. 
Replace selected text with spaces or a specified character

## How to Use

From Command palette, run `Alter Text: Replace with spaces`

No default keybinding is provided. Use similar to below
```json
[
    { "keys": ["ctrl+r"], "command": "replace_text_with", "args": { "replacement": " " } },
    { "keys": ["ctrl+-"], "command": "replace_text_with", "args": { "replacement": "-" } }
]
```

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -714,6 +714,19 @@
 			]
 		},
 		{
+			"name": "Alter Text",
+			"details": "https://github.com/cibinmathew/sublime-alter-text",
+			"author": "Cibin Mathew",
+			"labels": ["text manipulation"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": ["osx", "linux"],
+					"tags": true
+				}
+			]
+		},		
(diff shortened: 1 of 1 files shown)


braver:
Thanks for your submission. Did you have a use case in mind for this package? Ie. why and how would one use it.

Also, the replacement argument doesn't seem to be implemented, therefore the commands and the replace-with-hyphen examples don't actually work.

cibinmathew:
@braver I have added the missed changes

**Current usecase**
Replace a selection with predefined character. 
Few examples:
- while working with tables(properly justified cells), replace text with space/hypen so as not to break the table alignment
- select a line and replace with hypens (to make the above line look like a heading)
- Duplicate the current line and apply it on bottom line to make it a horizontal line

I plan to add more commands similar to `replace_text_with`

PR #8748, opened 2023-05-12
Add package 'Copy Files and Folders'

toastwaffle:

My package adds a "Copy..." entry to the Side Bar context menu for copying files and folders

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3714,6 +3714,17 @@
 			]
 		},
 		{
+			"name": "Copy Files and Folders",
+			"details": "https://github.com/toastwaffle/SublimeCopyFilesAndFolders",
+			"author": "Samuel Littley",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Copy from Find Results",
 			"details": "https://github.com/NicoSantangelo/sublime-copy-from-find-results",
 			"releases": [


repository/c.json
```diff
@@ -3807,7 +3818,6 @@
 		{
 			"name": "CopyToSharePointRoot",
 			"details": "https://github.com/joladev/SublimeCopyToSharePointRoot",
-			"author": "eakron",
```
braver, line 3810: This looks like an unintentional change?

braver:
[SideBarTools](https://packagecontrol.io/packages/SideBarTools) already provides this functionality (via "duplicate"). It does a few additional things too of course, but I'm not sure we need separate packages for every thing it does. What do you think?

repository/c.json
toastwaffle: Whoops, fixed!

toastwaffle:
> [SideBarTools](https://packagecontrol.io/packages/SideBarTools) already provides this functionality (via "duplicate"). It does a few additional things too of course, but I'm not sure we need separate packages for every thing it does. What do you think?

Aha! I was sure something must exist for this already. It doesn't come up in the install menu when you search for "copy", which I guess is why I didn't see it.

braver:
>  It doesn't come up in the install menu when you search for "copy", which I guess is why I didn't see it.

That's a really good point actually. I think I need to put some more labels on that package or otherwise improve its "SEO".

PR #8749, opened 2023-05-14
Update LESS package

deathaxe:
This commit retargets LESS package to SublimeText org and adds relevant release entries to maintain current version (2.7.4) for ST3 and ship rewritten 3.0.0 to ST4107+

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -769,17 +769,21 @@
 		},
 		{
 			"name": "LESS",
-			"details": "https://github.com/danro/LESS-sublime",
+			"details": "https://github.com/SublimeText/LESS",
 			"donate": "https://paypal.me/koenlageveen",
-			"labels": ["language syntax"],
+			"labels": ["language syntax", "auto-complete"],
 			"releases": [
 				{
-					"sublime_text": "<4000",
-					"tags": "st3-"
+					"sublime_text": "3143 - 4106",
+					"tags": "st3143-"
 				},
 				{
-					"sublime_text": ">=4000",
-					"tags": true
+					"sublime_text": "4107 - 4148",
+					"tags": "st4107-"
+				},
+				{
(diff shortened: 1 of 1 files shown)


repository/l.json
```diff
@@ -769,17 +769,21 @@
 		},
 		{
 			"name": "LESS",
```
braver, line 771: Maybe we should re-name to Less for consistency. 

I wonder if we could maybe also consolidate [Less​Improved](https://github.com/braver/LessImproved) by removing it and adding it to "previous names" here?

repository/l.json
```diff
@@ -769,17 +769,21 @@
...
-			"details": "https://github.com/danro/LESS-sublime",
+			"details": "https://github.com/SublimeText/LESS",
 			"donate": "https://paypal.me/koenlageveen",
```
braver, line 773: Let's remove the donation bit. I haven't received anything in a while, and it's no longer my work.

repository/l.json
deathaxe: :( Making all the tags and releases again.

deathaxe: LessImproved isn't registered at packagecontrol.io. AFAIK, it won't be renamed by PC if it was installed manually.

Can add it to `previous_names` in case someone added the repo to PC, though.

deathaxe:
I've reset all branches on SublimeText/Less to get rid of all the merge commits and create a clean starting point.

braver:
Alright, great work. I'll try to shoot some holes in the new syntax, but LGTM right now and that doesn't need to block progress here. Just as a final check let me know if you're ready to hit merge (or just hit merge yourself) and then let's ship this thing.

deathaxe:
I can't merge, but I think any further improvement might require more user feedback. So let's go.

braver:
It is done.

PR #8750, opened 2023-05-14
Add Hadar Theme to h.json

CristianVasquez13:

[1]:https://packagecontrol.io/docs/submitting_a_package
[2]:https://semver.org/


braver:
Your theme seems to have some settings. Shouldn't they be exposed to the end user?

braver:
The bot feedback is fine, you don't need a license file. Please address my earlier feedback about the settings.

braver:
> Your theme seems to have some settings. Shouldn't they be exposed to the end user?

☝🏻

CristianVasquez13:
Y como hago eso de la configuración? 🤔

braver:
You would usually have a menu entry (and a command palette entry) that allows a user to open and edit the settings.

braver:
I'm going to close this as stale, please respond (hopefully in English, sorry about that) if you want to continue.

jwortmann:
> You would usually have a menu entry (and a command palette entry) that allows a user to open and edit the settings.

Those menu and command palette entries are already included in Sublime Text; the menu entry is under *Preferences > Settings* (might be different on macOS (?)) and the command palette entry is *Preferences: Settings*.

Note that unlike plugins, themes cannot define their own settings files, instead settings are always read from the genereal *Preferences.sublime-settings* file.

In theory, a theme could add another entry to open a predefined *Preferences.sublime-settings* file on the lefthand side, and the users setting on the righthand side, but then the [built-in settings](https://www.sublimetext.com/docs/themes.html#settings) which are expected to be implemented by themes, wouldn't be visible. And I doubt that there are many themes which do it this way. Instead, I would just describe the settings in the Readme page.

PR #8751, opened 2023-05-17
Fix Less package

deathaxe:
This PR...

1. adds "LESS" to `previous_names` due to name case changes.
2. fixes tag prefixes.


braver:
I was just looking into this. Thanks 👍🏻

PR #8753, opened 2023-05-18
Add Tryton Snippets

tiyujopite:

My package provides snippets for [Tryton](https://www.tryton.org) module development.

[2]: https://semver.org

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -3380,6 +3380,17 @@
 			]
 		},
 		{
+			"name": "Tryton Snippets",
+			"details": "https://gitlab.com/datalifeit/sublimetext-tryton-snippets",
+			"labels": ["tryton", "snippets"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "TSL Text Scraping Language",
 			"author": "Dan Borufka",
 			"details": "https://github.com/polygoat/TSL",


braver:

We can't handle your submission without one.

braver:
@FichteFoll are the tests failing because the repo is on GitLab, or am I missing something else?

FichteFoll:
It's possible but I'm not aware of any known issues with gitlab currently. Unless it's on old and already fixed issue that was just never deployed on the server that does the reviews … (I really should get this on github actions only at some point …)

braver:
allright, let's just try it then, it all looks good and we can always roll back...

@tiyujopite there seems to be a typo in your readme: 

```diff
- SublimeText Tryron Snippets
+ SublimeText Tryton Snippets
```

tiyujopite:
Thanks @braver , I'm going to fix it.
So finally, is there a problem that I have to solve for the PR to be accepted or is everything fine?

rchl:
I'm aware of missing support for repos that are not personal: https://github.com/sublimelsp/repository/pull/44#issuecomment-899479629

The case there seems to be a bit different because it's nested within two "groups" (https://gitlab.com/webhare/lsp/LSP-webhare) and here there is just one (https://gitlab.com/datalifeit/sublimetext-tryton-snippets) so maybe it's not the same case and it's just an issue with the bot.

PR #8754, opened 2023-05-19
Add NeoVintageousHighlightLine

gerardroche:
https://github.com/gerardroche/NeoVintageousHighlightLine

My package is NeoVintageousHighlightLine

https://github.com/gerardroche/NeoVintageousHighlightLine

[2]: https://semver.org

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -291,6 +291,17 @@
 			]
 		},
 		{
+			"name": "NeoVintageousHighlightLine",
+			"details": "https://github.com/gerardroche/NeoVintageousHighlightLine",
+			"labels": ["neovintageous", "vim", "vi", "vintage", "vintageous", "neovim", "nvim"],
+			"releases": [
+				{
+					"sublime_text": ">=4074",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "NeoVintageousToggle",
 			"previous_names": ["ToggleNeoVintageous"],
 			"details": "https://github.com/NeoVintageous/NeoVintageousToggle",


braver:
If you're going to publish a lot of plugins, at some point it maybe starts making sense having your own channel?

gerardroche:
I don't expect to have a crazy amount of NeoVintageous packages. Most NeoVintageous plugins are bundled with NeoVintageous. It's really only the one's that have side effects or one's that are   unique to a subset of users like keyboard layout plugins e.g. [Colemak](https://github.com/gerardroche/NeoVintageousColemak) and [Dvorak](https://github.com/gerardroche/NeoVintageousDvorak) (those are experimental in development so we'll see how it goes).

Plugins like this one are not everbody's cup of tea. But I've been using them in my own custom Subime Text plugins for a while and figured others might find them useful.

gerardroche:
@braver When you said it make sense having my own channel, did you mean
 specifically for NeoVintageous or in general for all my packages? Because if
 you meant the latter, I didn't know you could do that. I thought SublimeLinter
 and the few others were special cases.

braver:
You could do a personal channel indeed. Some do have one. I don't like doing it that way though, because the packages are out of reach of maintainers of the channel and it's a hassle if a developer stops supporting a package that we still want to keep available (or fork or replace etc). Although it's not "policy", I would say we're quietly moving away from that idea for individual developers. Even the Sublime Text org simply lists their packages in the main channel. 

I brought it up mostly indeed because of the use cases like SublimeLinter and LSP, where it's easier to manage several plugins in one place. And maintainers of those "frameworks" are better positioned to manage and review those plugins anyway. If you don't need that and don't expect a large number of plugins, let's just keep them here in the main channel. 

Especially for known active members of the community and those who already maintain multiple packages, I hope the review process isn't too much of a blocker. I do keep an eye out for the known quantities, as yourself, and move them along quickly.

gerardroche:
I don't mind the review process. It's helpful to catch some oversights. 

Plus, it's not that often you add a package. And rarely update them. I don't know why some people get upset that it could maybe take up to a few weeks for a package to be processed.

braver:
Just some folks feel they're entitled to stuff. 🤷🏻

PR #8756, opened 2023-05-20
Add GitOpen

gerardroche:
https://github.com/gerardroche/sublime-git-open

My package is Git Open

https://github.com/gerardroche/sublime-git-open

[2]: https://semver.org

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -836,6 +836,16 @@
 			]
 		},
 		{
+			"name": "GitOpen",
+			"details": "https://github.com/gerardroche/sublime-git-open",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "GitOpenChangedFiles",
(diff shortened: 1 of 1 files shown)


gerardroche:
I wasn't away of the GitSavvy command. Probably because it is prefixed with "github:" in the Command Palette instead of GitSavvy. Very useful to know though. [I've suggested adding "GitSavvy" to all those commands for discoverability](https://github.com/timbrel/GitSavvy/issues/1756).

GitLink is similar to `git-open`. One difference is that GitLink doesn't have a dependency on an external program, but then it's looking for a maintainer. GitOpen doesn't intend to handle any of the implementation details but will simply remain a wrapper around the external tool. I wouldn't be interested in maintaining something like GitLink.

I've been using `git-open` on the terminal since forever and have had this plugin in my local sublimefiles since forever too.

My bash alias: `alias o='git-open'`.
My neovintageousrc: `nnoremap <leader>o :GitOpen<CR>`

I might try changing that last one to `github: open this file on remote`.

PR #8757, opened 2023-05-20
Add ToggleDebugMode

gerardroche:
https://github.com/gerardroche/sublime-toggle-debug-mode

My package is ToggleDebugMode

https://github.com/gerardroche/sublime-toggle-debug-mode

[2]: https://semver.org

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -2570,6 +2570,17 @@
 			]
 		},
 		{
+			"name": "ToggleDebugMode",
+			"details": "https://github.com/gerardroche/sublime-toggle-debug-mode",
+			"labels": ["debug"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "ToggleExclude",
 			"details": "https://github.com/henvic/ToggleExclude",
 			"releases": [


braver:
I believe this could use a little explanation in the readme: debug mode for what? What should users expect from this package? etc.

gerardroche:
It actually add a "Developer Tools" menu item with a "Toggle Debug Mode" item. Handy to toggle console logging on and off. There is a bug in sublime that makes the indexing logger sticky on restarts. https://github.com/sublimehq/sublime_text/issues/5991

braver:
Def. sounds useful 👍🏻

gerardroche:
debugmode

braver:
Oh, I kinda missed you created a top level menu item. That's frowned upon a bit. You can imagine we don't want several packages adding items to the already crowded top level. I thought you put them in Help, which would've made some sense.

In this case your menu items could maybe go into Tools > Developer, or directly in Tools.

gerardroche:
The Tools → Developer is a bit indirect. Personally I like it right on the main menu for quick access. I'll only toggle it on and off for a few minutes or even seconds just to see commands or input. But I can see why it's a rare used thing for most users.

I'll move it and create my own custom main menu item.

I just found https://packagecontrol.io/packages/Troubleshooting. The generate bug report is something I've wanted for a while, but I really think it needs to be built-in to sublime, because users just won't install plugin to generate a bug report. 

Most users won't even install the toggle debug mode plugin. I've tried to keep the plugin as simple as possible. I want the user to be able to glance at the code and understand it in 5 seconds.

I have custom plugins like showing the build in the status bar:

```py
import sublime
import sublime_plugin

class UserShowBuild(sublime_plugin.EventListener):

    def on_activated_async(self, view):
        view.set_status('sublime_text_build', 'Build ' + sublime.version())
```

I find it useful, but for most users it's just not relevant.

Or a Dump Info command that I have in the main menu developer tools:

```py
import sys
import sublime
import sublime_plugin


class DumpInfo(sublime_plugin.WindowCommand):

    def run(self):
        print('+-------------------')
        print('| Sublime Text version               ', sublime.version())
        print('| Python version                      {}.{}.{} {}{}'.format(
            sys.version_info[0],
            sys.version_info[1],
            sys.version_info[2],
            sys.version_info[3],
            sys.version_info[4]))
        print('| %-35s %s' % ('sys.flags', sys.flags))
        print('| %-35s %s' % ('sys.abiflags', sys.abiflags))
        print('| %-35s %s' % ('sys.path', sys.path))
        print('| %-35s %s' % ('__debug__', __debug__))
        print('| sublime.platform()                 ', sublime.platform())
        print('| sublime.arch()                     ', sublime.arch())
        print('| sublime.channel()                  ', sublime.channel())
        print('| sublime.executable_path()          ', sublime.executable_path())
        print('| sublime.packages_path()            ', sublime.packages_path())
        print('| sublime.installed_packages_path()  ', sublime.installed_packages_path())
        print('| sublime.cache_path():              ', sublime.cache_path())
        print(sublime.ui_info())
        print(sublime.get_macro())
        print('+')
```

But users don't install this stuff so I'm not sure if it's worth creating a plugin. I was already on the fence with the toggle debug mode plugin. 

I think many features like this are really only useful when they're bundled with sublime or part of a built-in developer mode.

Anyways, yeah, I'll move that menu item.

braver:
> I find it useful, but for most users it's just not relevant.

Ha, now that's something I would find useful almost every week or so. But yeah, that makes two of us, and there's probably more on Discord who might as well... but that's still a small niche. 

The main problem with packages like Troubleshooting is the (lack of) visibility. Even most users in the core of the community won't know about that one. But like you say, if you're just some random user with a problem, you're going to install packages. If you're going to report the bug at all. Even if the tool were in Sublime in the first place, and that _would_ be better, hardly anyone is going to A. find it, and B. use it. But it's good that it's out there I guess 🤷🏻 

> Anyways, yeah, I'll move that menu item.

👍🏻 👍🏻

PR #8758, opened 2023-05-20
Add GotoPhpManual

gerardroche:
https://github.com/gerardroche/sublime-goto-php-manual

My package is GotoPhpManual

https://github.com/gerardroche/sublime-goto-php-manual

My package is similar to a few... However it should still be added because it adds a feature of being able to jump to tags in local downloaded manual. The jumps are also more accurate than some others.

[2]: https://semver.org

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -1713,6 +1713,16 @@
 			]
 		},
 		{
+			"name": "GotoPhpManual",
+			"details": "https://github.com/gerardroche/sublime-goto-php-manual",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "GoToPoint",
 			"details": "https://github.com/breck7/gotopoint",
 			"releases": [


gerardroche:
Put your cursor on a symbol and press <key>f1</key> for goto local manual (if it exists) or <key>shift+f1</key> for remote manual.

It probably could be added to other packages, but some packages never get updated.

It looks like the PMan package downloads the help page and displays it in a view which quiet different than this one. This plugin just opens the help page in a browser. I've added this to the readme:

---

Put your cursor on a PHP symbol and press <key>f1</key> to open a local PHP manual in your browser (if you've downloaded one) and <key>shift+f1</key> to open the remote PHP manual in your browser.

Handy if you need the PHP Manual when working offline.

braver:
I'd be surprised if Pman even still works after 10 years. Will it also show PHP4 documentation only or something? Anyway, let's just leave that where it is and add your package 👍🏻

gerardroche:
It'll show you the latest PHP manual, similar to the Goto Documentation plugin (shift+f1). But if you download the PHP Manual you goto your local manual (f1). If the local manual doesn't exist, then f1 will open the remote manual. It's great when working offline.

PR #8760, opened 2023-05-22
Remove polyfill

gerardroche:
My package is no longer maintained and has since been replaced by other plugins.

https://github.com/gerardroche/sublime-polyfill


braver:
Alright. Note that removal of a package is not typically correctly reflected on the package control website, that’s a bit of a lingering issue.

PR #8762, opened 2023-05-23
Add NeoVintageousFiles

gerardroche:

My package is NeoVintageousFiles

https://github.com/gerardroche/NeoVintageousFiles

Q: My package has two package dependencies for some functionality to work, SIdeBarTools and Origami. Is it possible to specify these dependencies to have them auto installed? Or some other solution other than instructing the user to install the dependencies?

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -291,6 +291,17 @@
 			]
 		},
 		{
+			"name": "NeoVintageousFiles",
+			"details": "https://github.com/gerardroche/NeoVintageousFiles",
+			"labels": ["vim", "neovintageous", "vintageous", "vintage"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "NeoVintageousHighlightLine",
 			"details": "https://github.com/gerardroche/NeoVintageousHighlightLine",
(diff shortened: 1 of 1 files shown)


braver:
> Q: My package has two package dependencies for some functionality to work, SIdeBarTools and Origami. Is it possible to specify these dependencies to have them auto installed? Or some other solution other than instructing the user to install the dependencies?

Not that I know of. The dependencies system is designed around python modules, not ST packages, and there is no way for one package to require another. Maybe @FichteFoll or @deathaxe can chime in, they know quite a bit more about these things.

FichteFoll:
> Not that I know of. The dependencies system is designed around python modules, not ST packages, and there is no way for one package to require another.

This is still correct, packages do not have a way  to tell PC that they depend on other packages to be installed as well. What I generally recommended to people with such requirements was to check for the import of the package they depend on at runtime (or at import time but with a try-except) and then notify the user of the missing dependency when they try to use your package, i.e. invoke a command.

FichteFoll (approved):

gerardroche:
It would be really great to have package dependencies. The import check adds a bit of cruft to packages, but it's ok. In this case I'll leave it, the documentation clearly informs about the dependencies.

Origami is the most popular dependency I use. It's just no worth trying implement that window splitting stuff when Origami does it so well.

PR #8763, opened 2023-05-24
Add ShowAsciiValue

gerardroche:

My package is ShowAsciiValue

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1591,6 +1591,17 @@
 			]
 		},
 		{
+			"name": "ShowAsciiValue",
+			"details": "https://github.com/gerardroche/sublime-show-ascii-value",
+			"labels": ["status bar"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "ShowDefinitionEx",
 			"description": "Visually showing definition",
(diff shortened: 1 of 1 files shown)


braver:
This look similar to https://packagecontrol.io/packages/Show%20Character%20Code?

gerardroche:
I didn't see that one when I searched Package Control. It's very similar.

braver:
Quite 😅 So yeah, can we consolidate in some way, or is there another reason to add your package? That project seems to get some updates still, so you could make a PR there if you have any improvements for it.

gerardroche:
Your welcome to copy it across and make a pr. The code is open sourced. I would continue to use my own, because it's not worth getting involved on another repo for something like this. I hate it when I open a pr on a project later down the road and the pr never gets merged or stays open until the cows come home and then you have maintain a fork anyway. It's a trade-off. You can't write everything yourself.

I should note that the plugins I've been submitting recently, have been part of my custom setup since forever. I've just been doing a spring clean, extracting it all out into individual packages. It was all getting a bit messy.

braver:
> because it's not worth getting involved on another repo for something like this

I totally get that. Since an existing package exists that seems to provide the same functionality, our policy is to close this PR. Hope you don't mind that.

gerardroche:
No worries.

PR #8764, opened 2023-05-25
Add ChromeExtensionExporter

iRonin:

My package is Chrome Extension Exporter.
https://github.com/iRoninIT/sublimetext-chrome-extension-exporter

[2]: https://semver.org

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1066,6 +1066,18 @@
 			]
 		},
 		{
+			"name": "Chrome Extension Exporter",
+			"description" : "Exports Chrome Extension files from the current project as ZIP for quick local distribution.",
+			"details": "https://github.com/iRoninIT/sublimetext-chrome-extension-exporter",
+			"labels": ["chrome extension"],
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


repository/c.json
```diff
@@ -1065,6 +1065,18 @@
 				}
 			]
 		},
+		{
+			"name": "Chrome Extension Exporter",
+			"description" : "Exports Chrome Extension files from the current project as ZIP for quick local distribution.",
```
braver, line 1070: It's easier to manage if you leave this out. Package control will then take the GitHub repo description (which you can easily change). However, if you want this to be different, that's also fine of course.

braver:
Thanks, looks like a useful utility. 

Your package has settings. These should be exposed to users via the main menu and command palette. Examples you can copy-paste from: [menu](https://github.com/SublimeText/Terminal/blob/master/Main.sublime-menu), [palette](https://github.com/SublimeText/Terminal/blob/master/Default.sublime-commands#L2-L9).

braver:
👋🏻

PR #8766, opened 2023-05-25
Make GitDiffView backwards compatible with ST3

predragnikolic:
Hello, 
I decided to update the GitDiffView plugin and only support Python 3.8 forward,
but I don't want to break ST3 users, so I created this PR.

I created a release tag for ST3 -> https://github.com/predragnikolic/sublime-git-diff-view/releases/tag/st3-0.5.0
I am not sure if I need to do anything else, if you spot something off, please say.

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -644,7 +644,12 @@
 			"labels": ["git", "diff"],
 			"releases": [
 				{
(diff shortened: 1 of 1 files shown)


braver:
Looks good 👍🏻

PR #8768, opened 2023-05-30
Add GotoDefinitionSplit

gerardroche:
https://github.com/gerardroche/sublime-goto-definition-split

  **A context menu item is added to Goto Definition Split**
- [ ] My package doesn't add key bindings.
  **Key bindings are added to override the goto definition keybinding. This is intentional. The point of the plugin is change the Goto Definition in split to split by group instead of by tab.**

My package is GotoDefinitionSplit. Change Goto Definition in split to split by group instead of split by tab.

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -1709,6 +1709,17 @@
 			]
 		},
 		{
+			"name": "GotoDefinitionSplit",
+			"details": "https://github.com/gerardroche/sublime-goto-definition-split",
+			"labels": ["file navigation"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


braver:
Since this depends on Origami anyway, and is by itself really tiny, wouldn't it make sense to integrate it as a feature of Origami?

I understand why you're shipping with keybindings. But I think the explanation you've provided here should be in your readme as well. It's not clear from the readme what the intent of this package is, or even that it will override any keybindings.

gerardroche:
I think this is out of scope for Origami. I expect there could be additional features added to this package and if I was maintaining Origami I wouldn't want this as part of the package.

I updated the readme: **This plugin overrides Sublime Text's Goto Definition in side-by-side key bindings so the side-by-side is in a group instead of a tab.**

PR #8770, opened 2023-05-30
Add GenerateSyntaxTestAssertions

gerardroche:
https://github.com/gerardroche/sublime-generate-syntax-test-assertions

My package is GenerateSyntaxTestAssertions. A syntax test assertion generator.

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -252,6 +252,17 @@
 			]
 		},
 		{
+			"name": "GenerateSyntaxTestAssertions",
+			"details": "https://github.com/gerardroche/sublime-generate-syntax-test-assertions",
+			"labels": ["generator", "syntax", "testing"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"details": "https://github.com/SublimeText/GenerateUUID",
 			"releases": [
 				{


repository/g.json
```diff
@@ -251,6 +251,17 @@
 				}
 			]
 		},
+		{
+			"name": "GenerateSyntaxTestAssertions",
```
braver, line 255: Again, similar to other review points in other PR's:

- consider using spaces in package names to make them easier to read (I don't suffer from dyslexia, but many programmers do, and even for me words this long look like gibberish and they all look the same)
- your readme does not explain how to use the package or what to expect of it exactly

gerardroche: The reason I generally don't use spaces is because it can cause difficulties with things like imports. I'm not sure there's a good way around that.

I updated the readme a usage example. 

## Usage

Put your cursor on a line that you want to generate syntaxes for and run the
command.

**Command Palette → Generate Syntax Test Assertions**

```php
<?php

declare(strict_types=1);
// ^^^^ embedding.php text.html.php meta.embedded.php source.php.embedded.html keyword.declaration.php
//     ^ embedding.php text.html.php meta.embedded.php source.php.embedded.html meta.group.php punctuation.section.group.begin.php
//      ^^^^^^^^^^^^ embedding.php text.html.php meta.embedded.php source.php.embedded.html meta.group.php constant.other.php
//                  ^ embedding.php text.html.php meta.embedded.php source.php.embedded.html meta.group.php keyword.operator.assignment.php
//                   ^ embedding.php text.html.php meta.embedded.php source.php.embedded.html meta.group.php meta.number.integer.decimal.php constant.numeric.value.php
//                    ^ embedding.php text.html.php meta.embedded.php source.php.embedded.html meta.group.php punctuation.section.group.end.php
//                     ^ embedding.php text.html.php meta.embedded.php source.php.embedded.html punctuation.terminator.statement.php
```

The generator inserts tests for all points on the line. You can adjust the tests
as needed.

See [Sublime Text Syntax Definition Documentation](https://www.sublimetext.com/docs/syntax.html).

braver: Excellent, thanks! I didn't know that about imports, never ran into the problem myself but that's a fair enough reason.

PR #8771, opened 2023-06-05
Add ST3 branch for TrailingSpaces

rchl:
Forking ST3 branch to be able to introduce types in the ST4 version.

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -3152,7 +3152,7 @@
 					"tags": "st2-"
 				},
 				{
-					"sublime_text": "3147 - 4107",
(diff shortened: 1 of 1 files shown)


repository/t.json
```diff
@@ -3152,7 +3152,11 @@
 					"tags": "st2-"
 				},
 				{
-					"sublime_text": ">=3148",
+					"sublime_text": "3147 - 4107",
```
rchl, line 3155: I believe this range syntax was not inclusive (correct me if I'm wrong) so it might appear like it overlaps with st2 and st4 versions.

PR #8775, opened 2023-06-15
Add Copy Relative Command

patoui:
https://github.com/patoui/CopyRelativeCommand

My package is similar to [CopyRelativePath](https://github.com/bpicolo/CopyRelativePath) however it should still be added as it provides different functionality by given the user the ability to define rules by language (source) and glob patterns. It also allows them to inject scope relevant symbols such as function, class, or function falling back to class.

As an example; if the cursor is within a [PHPUnit](https://phpunit.de/) method and the user can [define rules](https://github.com/patoui/CopyRelativeCommand/tree/main#quick-start-%EF%B8%8F) to copy a command to their clipboard to run the test in terminal.

From the [example configuration]((https://github.com/patoui/CopyRelativeCommand/tree/main#quick-start-%EF%B8%8F)) in the repo, the output would look like:
```
vendor/bin/phpunit tests/Unit/ExampleTest.php --filter=test_that_true_is_true
```

[2]: https://semver.org

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3774,6 +3774,16 @@
 			]
 		},
 		{
+			"name": "Copy Relative Command",
+			"details": "https://github.com/patoui/CopyRelativeCommand",
+			"releases": [
+				{
+					"sublime_text": ">=4107",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Copy Relative Path",
 			"details": "https://github.com/bpicolo/CopyRelativePath",
 			"releases": [


braver:
Woah, that’s hyper niche, cool 👍🏻

I think you should decide, in the command palette or elsewhere, if your package is called “Copy Relative Command” with or without the spaces.

If you want to support project specific settings you don’t need to parse the project data yourself. Instead you can access the settings for the current view, which will have the project settings applied to it.

braver:
You might also want to document how your users can set up key bindings for this.

patoui:
@braver thanks for the feedback! Sorry for the delayed response, I'll have a look at the changes you suggested within the next week or so (hopefully 🤞)

patoui:
@braver I've made the changes you suggested, let me know if there's anything else I can do to improve the package :)

braver:
🎉

PR #8776, opened 2023-06-15
Add Tweaks plugin

kapitanluffy:

My package is Tweaks

[2]: https://semver.org

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -3506,6 +3506,17 @@
 			]
 		},
 		{
+			"name": "Tweaks",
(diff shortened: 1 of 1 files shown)


braver:
Thanks for your submission. We don't typically accept collection packages like this. "Small random tweaks packed in one plugin"... Several existing packages already provide some of the functionality of your package. It's a very small chance other users will need exactly your collection of tweaks. And how (why) would someone look for and find your package? Things like these are usually better provided via a "tips & tricks" post on Medium or in the ST forums.

PR #8777, opened 2023-06-15
Adding JsonSymbolsAndColorSyntax

Monox18:

My package is ...

[2]: https://semver.org

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -1267,6 +1267,17 @@
 			]
 		},
 		{
+			"name": "JsonSymbolsAndColorSyntax",
+			"details": "https://github.com/Monox18/JsonSymbolsAndColorSyntax/",
+			"labels": ["json", "code navigation", "formatting", "color scheme", "language syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "JsonTree",
 			"details": "https://github.com/Flyclops/JsonTree",
 			"labels": ["json", "tree", "search"],
--- a/repository/j.json
+++ b/repository/j.json
@@ -1139,6 +1139,17 @@
 			]
 		},
 		{
+			"name": "Json Colors And Navigation",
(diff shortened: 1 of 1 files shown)


braver:
That's a cool concept and I like how you've implemented it. The only thing I would worry about is the name of the package. From your PR here I had no clue what to expect, and the README surprised me. That might mean that people won't be able to find your package. The functionality reminds me a little of features like Indent Rainbow and Rainbow Braces that exist for other editors, or [Rainbow Brackets](https://packagecontrol.io/packages/RainbowBrackets) for ST. May I suggest "JSON Rainbow"?

Monox18:
Thanks for the feedback. 

Not sure that `Rainbow` is a keyword people are actually searching. Seems very specific to me. But I do agree the current name might not be the most precise. Maybe we can agree `Color` is more suitable than `Rainbow` when people are thinking about syntax coloring.

Let's see, this package has 3 main purposes with JSON files: 
1) Allow symbol navigation
2) Color nested keys. Rainbow Color targeted the braces whereas here I'm targeting the string within the quotes.
3) Display idented keys in the additional Sublime Text plugin [Sublime Outline](https://github.com/warmdev/SublimeOutline)

So with those main points I can think of `JsonSymbolsColorAndNavigation`, `JsonColorKeysAndNavigation`, `JsonColorAndNavigateKeys`, etc.. Though there's a lot of keyword sutffing in the name, it provides the best chances of being discovered as Package Control search dialog in ST only uses the name and not the description.

Let me know and I will update both name and readme accordingly. Also after like 30 mins deciding on the name `JsonColorAndNavigateKeys`  seems the most descriptive and accurate.

braver:
> Also after like 30 mins deciding on the name JsonColorAndNavigateKeys seems the most descriptive and accurate.

Sounds good to me! I could also see JsonColorsAndNavigation or even "Json Colors and Navigation" (it's ok to have spaces).

Monox18:
Thanks for the feedback. I have pushed the changes.

braver:
:shipit:

PR #8778, opened 2023-06-16
Update EJS 2 Package

nwoltman:
Update [EJS 2](https://github.com/nwoltman/sublime-ejs) releases for supported versions of Sublime Text due to breaking changes in `v2.0.0`.

[2]: https://semver.org

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -524,7 +524,7 @@
 			"releases": [
 				{
 					"sublime_text": "3098 - 4074",
-					"tags": "1."
+					"tags": "st3-"
 				},
 				{
 					"sublime_text": ">=4075",
--- a/repository/e.json
+++ b/repository/e.json
@@ -523,7 +523,11 @@
 			"labels": ["language syntax", "snippets", "color scheme"],
 			"releases": [
 				{
-					"sublime_text": ">=3098",
+					"sublime_text": "3098 - 4074",
+					"tags": "1."
+				},
+				{
+					"sublime_text": ">=4075",
 					"tags": true
 				}
 			]


repository/e.json
```diff
@@ -523,7 +523,11 @@
 			"labels": ["language syntax", "snippets", "color scheme"],
 			"releases": [
 				{
-					"sublime_text": ">=3098",
+					"sublime_text": "3098 - 4074",
+					"tags": "1."
```
braver, line 527: To do this, you need to use prefixed versions. 

From the [docs](https://raw.githubusercontent.com/wbond/package_control/master/example-repository.json):

```
		// If your package is only compatible with specific builds of
		// Sublime Text, this will cause the package to be hidden from
		// users with incompatible versions.
		//
		// The "tags" key can be true for all valid semantic version tags, or
		// can be a prefix string. Only tags in the form
		// {prefix}{semantic_version} will be selected. In the example below,
		// the entry with "sublime_text": "<3000" will match tags like:
		//
		// "st2-1.0.0"
		// "st2-1.1.0"
		//
		// The release with "sublime_text": ">=3000" will match tags like:
		//
		// "st3-1.0.0"
		// "st3-1.1.0"
		{
			"details": "https://github.com/wbond/sublime_alignment",
			"releases": [
				{
					"sublime_text": "<3000",
					"tags": "st2-"
				},
				{
					"sublime_text": ">=3000",
					"tags": "st3-"
				}
			]
		},
```

rchl: I think it's still a valid prefix.

braver: `{prefix}{semantic_version}` reads to me like `1` and `.4.7` would not leave you with a valid semver in the second. It might work like that, with a major version for a specific st version, but I haven't seen it used like that and wouldn't bet on it not knowing how this actually works internally. `st3-<semver>` is safer.

rchl: In this case the prefix was `1.` so someone could do `1.4.7.0` which would result in 4.7.0 semantic version. But now that I wrote it, it doesn't look clear at all so following existing convention seems better indeed.

braver: Yes, I assume that would work. Although in this case the proposed tag was 1.4.7, which wouldn’t. We’re guessing though, prefixes aren’t part of semver spec. You’d kind of expect to use version ranges (ie. 1.X, 1.4.X, etc) to match compatibility with ST version. But this is what we’ve got and the convention works for sure.

FichteFoll: Your assumptions are correct. Prefixes are simply stripped from the tag and then the rest is parsed as a semver.

PR #8784, opened 2023-06-22
Add NpmRunner package

kennethnym:

My package is a plugin that scans all `package.json` in a workspace and makes them runnable through the Command Palette.

[2]: https://semver.org

Edit: [link to plugin repo for quicker access :)](https://github.com/kennethnym/sublime-npm-runner)

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -1133,6 +1133,18 @@
 			]
 		},
 		{
+			"name": "NpmRunner",
+			"details": "https://github.com/kennethnym/sublime-npm-runner",
+			"author": "kennethnym",
+			"labels": ["npm", "javascript", "node"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "nRepeat",
 			"details": "https://github.com/bfrascher/Sublime-nRepeat",
 			"releases": [


braver:
Thanks, that looks super useful. Even a bit surprising there isn't already something like it. 

Some minor concerns:
- you could use `.gitattributes` to exclude the [NpmRunner.sublime-project](https://github.com/kennethnym/sublime-npm-runner/blob/main/NpmRunner.sublime-project) file from the package 
- what is the sublime_lib dependency and is it not already internal to sublime? (I'm not too familiar with the existing dependency ecosystem honestly)
- it's always nice if packages also provide a command palette entry to edit settings

Ideas:
- I just tested it on a project of mine, but of course I forgot to run `npm install` first... maybe worth considering implementing some of the standard commands like install. audit, ci, fund and ls also look like useful candidates.

rchl:
You seem to be pulling for window folders changes but there is `EventListener.on_load_project_async` notification that should notify you when anything in the project changes.

kennethnym:
Thank you for everyone's feedback!

@braver 

- According to [`sublime_lib`'s repo](https://github.com/SublimeText/sublime_lib) it has to be declared in `dependencies.json` if a plugin consumes it. It may have already been included in ST4, but this plugin is also compatible with ST3, so I can't use any ST4 APIs.
- There is already [an existing plugin](https://packagecontrol.io/packages/npm) that runs other NPM commands like `install`, which is why I narrowed the scope of this plugin to just running NPM scripts. But I can implement other commands as well for completeness' sake.

@rchl 

Unfortunately `on_load_project_async` is a ST4-only method. As mentioned above NpmRunner is backward-compatible with ST3 so it is not possible for me to use the method.

rchl:
You probably wouldn't loose many users if at all if you would go with only supporting ST4+ but it's your choice.

kennethnym:
I do agree polling is not the most efficient if there's a better way to do it, so I'm happy to only support ST4. I'll make the changes when I'm free. Thanks again for the feedback!

repository/n.json
```diff
@@ -1132,6 +1132,18 @@
 				}
 			]
 		},
+		{
+			"name": "NpmRunner",
+			"details": "https://github.com/kennethnym/sublime-npm-runner",
+			"author": "kennethnym",
```
braver, line 1138: If the author is just your GitHub user name you don't need to provide it. Package Control will take that info from the repo.

kennethnym: Got it, I'll change it

braver: 👋🏻

kennethnym:
Apologies for leaving this stale as my focus was directed elsewhere. I will try to get this done over the weekend.

braver:
I'm going to close this for now. Please just drop me a note when you're ready to continue.

PR #8785, opened 2023-06-25
Add Laravel Filament

PyaeSoneAungRgn:

My package is to run [Laravel Filament](https://filamentphp.com) Artisan CLI using the Sublime Text interface, without having to open and use the command line.

My package is similar to [Laravel 5 Artisan](https://github.com/dydx/Laravel-5-Artisan). However it should still be added because my package is exclusively includes support for [Laravel Filament](https://filamentphp.com).

[2]: https://semver.org

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -299,6 +299,16 @@
 			]
 		},
 		{
+			"name": "Laravel Filament",
+			"details": "https://github.com/PyaeSoneAungRgn/sublime-laravel-filament",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Laravel Forms Bootstrap Snippets",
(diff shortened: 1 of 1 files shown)


braver:
Thanks for your submission. Some feedback:
- In Main.sublime-menu you don't need to re-set the caption of the top level items, just use the id.
- You should use edit_settings to open settings in Split View, instead of having separate entries for default/user setting.
- How should the "user commands" be used? That's not a normal usage pattern and it's not explained anywhere.

PyaeSoneAungRgn:
- used: only id at top level items
- used: edit_settings
- deleted: "user commands" which is not necessary

Thanks for your feedback.

braver:
Have a look at some example packages: you typically end up with 1 menu entry for settings: https://github.com/braver/SideBarTools/blob/master/Main.sublime-menu#L16

PyaeSoneAungRgn:
Thanks

used: single setting

PR #8788, opened 2023-07-06
Add Projectionist package

timfjord:

The package implements the logic that was originally introduced by [Tim Pope](https://github.com/tpope) in his [vim-projectionist](https://github.com/tpope/vim-projectionist) plugin. And the main idea is to provide granular project configuration using "projections".

Here are the package features:

- jump between test and implementation files
- open/jump to an alternate file from the Side Bar
- `.projections.json file` support (including JSON schema validation)
- built-in projections for Elixir, Ruby, and Sublime package development
- public API for other packages to use
- work on all platforms

[2]: https://semver.org

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -2623,6 +2623,19 @@
 			]
 		},
 		{
+			"name": "Projectionist",
+			"details": "https://github.com/timfjord/Projectionist",
+			"labels": [
+				"project"
+			],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "ProjectMaker",
 			"details": "https://github.com/bit101/ProjectMaker",
 			"labels": ["project", "template", "project-templates", "cookiecutter", "jinja2"],


braver:
That's very cool!

Some minor feedback:
- Your main menu is empty. It should at the very least give access to the settings.
- A sidebar context menu doesn't have mnemonics.

braver:
By the way, you might benefit of opting into the python3.8 package runtime.

timfjord:
@braver Thanks for taking a look.  
I've addressed your feedback.

As for Python 3.8, my plan was first to implement a reasonable amount of features and then switch to Python 3.8.
This way, Sublime Text 3 users can use the basic features of the package.
So, for now, I would keep 3.3

braver:
There aren't that many people still using ST4 though, I wouldn't expect any active users that are also looking for new package. The 3 to 4 upgrade was free IIRC, and only changed the longevity of licenses. But suit yourself of course.

timfjord:
I agree with you, but I don't want to switch the env without leveraging the full potential of Python 3.8
So I would probably stick to the original plan

FichteFoll:
If your code is compatible with both Python 3.3 and 3.8, your plugins would still benefit from running on 3.8 because it's faster.

timfjord:
Hmm, I didn't think of it from this perspective.
Will update the version then 👍🏻

PR #8790, opened 2023-07-09
Add GoTo File package

ryuukk:

My package provides an alternative to GotoAnything

When user only want to switch file, GotoAnything is distracting, because:

- it flickers the screen on each keystrone
- it preview file on each keystroke
- it hides the path you were reading on the file since it previews a different file on each keystroke
- it preview files you don't care about (why preview a.c aa.c aaa.c aaaa.c if i only want file aaaaaaa.c?)

GoTo File lists files from opened folders, caches them to avoid useless I/O and properly categorize them

Repo: https://github.com/ryuukk/goto_file

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -1651,6 +1651,16 @@
 			]
 		},
 		{
+			"name": "GoTo File",
+			"details": "https://github.com/ryuukk/goto_file",
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


braver:
I'll close this PR in the mean time. We can re-open when you're ready.

PR #8791, opened 2023-07-10
Update repo name for RailsGoToSpec

sporto:

My package is for jumping between rails code and test files.
I changed the repository name.

[2]: https://semver.org

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -205,16 +205,12 @@
 		},
 		{
 			"name": "RailsGoToSpec",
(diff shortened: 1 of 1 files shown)


braver:
Your repo description reads "Sublime Text 3 plug-in for going to Rails spec file", ST has been at version 4 for quite a number of years now. You might want to simply drop the version number there as it might lead users to think it's not compatible with recent version of Sublime.

PR #8795, opened 2023-07-18
update xonsh package

eugenesvk:
Change tag version installation scheme to separate installs based on ST versions since ST updated its native Python syntax highlighting scheme to v2 since 4148+

- [ ] ± My package doesn't add key bindings. ** 
  It overrides quotes `'`/`"` to add a nicety of autopair after "\b" in xonsh strings, so doesn't create any new __key__ bindings
- [x] (no commands) Any commands are available via the command palette.
  Don't think quote overrides need to be listed

My package is a syntax highlighting package for the Python-powered shell [xonsh](https://xon.sh) (for files with `.xsh` and `.xonshrc` extensions)

[2]: https://semver.org

### DIFF
--- a/repository/x.json
+++ b/repository/x.json
@@ -186,7 +186,11 @@
 			"labels": ["xonsh", "python", "language syntax"],
 			"releases": [
 				{
-					"sublime_text": ">=4075",
+					"sublime_text": "4075 - 4147",
+					"tags": "4075-4147_"
+				},
+				{
+					"sublime_text": ">=4148",
 					"tags": true
 				}
 			]


repository/x.json
```diff
@@ -186,8 +186,12 @@
...
 			"releases": [
 				{
-					"sublime_text": ">=4075",
-					"tags": true
+					"sublime_text": "4075 - 4147",
+					"tags": "4075-4147_"
+				},
+				{
+					"sublime_text": ">=4148",
+					"tags": "4148_"
```
braver, line 194: This looks a bit impractical to me. You should be able to continue your normal numbering for 4148+ and only serve prefixed tags to legacy versions.

eugenesvk: but then in the future if another release requires another version, I'd need to either retag all those versions or make all the future versions from that point on get another prefix? So then the benefit of non-prefixed versions is temporary anyway

braver: no, you won't have to re-tag existing versions. just introduce a prefixed release, pinning certain versions of ST on that release. Ie. you use prefixed releases to pin legacy versions while your ongoing releases are always simply ongoing numbers.

eugenesvk: Ok, I've updated the format to just `"tags":true`, but only `>=4148`, so the non-prefixed tags like `0.2.1` should be picked up just by the new version

FichteFoll (approved):

repository/x.json
eugenesvk: Hm, I see the package is updated to 0.2.1, but then this is the version that is installed even on older ST versions (checked with 4134)

PR #8796, opened 2023-07-18
Comparison plugin (convenient/soft diff usage)

BlackCatDevel0per:
- [x] My package doesn't add context menu entries. * (Can hide from settings)

My [package](https://github.com/BlackCatDevel0per/sublime-cat_compare) is a fork of the existing ["Compare Side-By-Side"](https://packagecontrol.io/packages/Compare%20Side-By-Side) package for convenient file comparison using diff, but fixing a bug with themes and color schemes for new versions of sublime text 4+ and showing tabs in new group types. I also refactored the code a bit and changed the python version to 3.8 and plan to further develop and improve the plugin (it's just that the author rarely updates it, and I often come across a comparison of different code, and similar plugins are poorly ported or they have poor support for new versions of sublime text above 3)

My package is similar to Compare Side-By-Side However it should still be added because the author rarely updates it and still doesn't have proper st4+ support

[2]: https://semver.org

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -405,6 +405,18 @@
 			]
 		},
 		{
+			"name": "Cat Compare",
+			"details": "https://github.com/BlackCatDevel0per/sublime-cat_compare",
+			"author": "bcdev",
+			"labels": ["diff", "compare", "comparison"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Cataracted Dark Color Scheme",
 			"details": "https://github.com/betraying/cataracted-dark-sublime-text",
 			"labels": ["color scheme"],


BlackCatDevel0per:
@packagecontrol-bot retest

deathaxe:
Did you try to contribute to the original plugin or ask for overtaking maintenance?

Authors may abandon their plugins for various reasons. So tranfering ownership or maintenance is not too uncommon.

If the original author does not respond, it would also be possible to replace the repository by another one.

Side-By-Side compare is a rather popular package, thus adding another one may not be sufficient fix for most users, despite having forked packages not being desireable - just to ship some fixes.

deathaxe:
We just want to avoid having more or less duplicated packages with nearly the same functionality.

Overtaking maintenance of an existing package (with agreement of the original author) doesn't mean to be limited to some fixes nor does it rule out any functional or structureal improvements.

It is also ok to move a repository to another location (repo or hoster).

It just should keep its original name, so all users already using it (and not being aware of new ones) can benefit from new changes and ongoing development. Otherwise your plugin is likely to be ignored by majority of users as Side-by-Side Compare is more popluar and more descriptive name or it creates confusion about which one to use.

The major point (and desired way) would be to:

1. ask original author, whether it's ok to overtake the package, if he or she is no longer interested in it.
2. if BitBucket is not yours, rename your repo/package to Side-by-Side-Compare and replace existing package link in your repo.

So development of Side-By-Side compare just goes on with another maintainer and repo, but without breaking users experiences.

BlackCatDevel0per:
> The major point (and desired way) would be to:
> 
> 1. ask original author, whether it's ok to overtake the package, if he or she is no longer interested in it.
> 
> So development of Side-By-Side compare just goes on with another maintainer and repo, but without breaking users experiences.

Okay, I'll discuss this with the author, but I don't think I can get a response from the developer quickly..
If I make my new plugin from scratch, but with the same basic task (comparing files with diff as in Side-By-Side), will it be added to the repository?

It's just that in this case it will be possible to avoid duplicating an existing plugin and, as I originally wanted, I can write a plugin for comparing text files based on diff

braver:
It's ok if you reimplement the package. The main thing is that for users it should be transparent. Especially for a popular package like this, users will expect the package to provide a certain functionality and you need to be careful around breakage of any of that. Being just "similar" is not enough here, it should basically be a drop-in replacement, or a clear improvement, ie. "major release". 

If you don't get any response from the current maintainer (we usually "time out" after 2 weeks), feel free to move forward with a fork.

braver:
We just merged a drop in replacement. I believe your intention is to do some more development and make changes to the package. You can now either contribute to the new comparison package, or we simply add your package as an alternative option.

PR #8797, opened 2023-07-22
Added AWS Helper plugin

cibinmathew:

My package is **AWS Helper**: https://github.com/cibinmathew/sublime-aws-helper

This plugin helps to validate IAM policy json from editor. Error findings are listed in details. The plugin makes use of `accessanalyzer` from `boto3` library.

[2]: https://semver.org

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -2756,6 +2756,7 @@
 			"name": "AWS Helper",
 			"details": "https://github.com/cibinmathew/sublime-aws-helper",
 			"labels": ["aws", "awscli", "amazon"],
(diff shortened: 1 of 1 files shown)


braver:
Nice job also on the clean and clear readme. 2 questions:

- does the script upload the target file to a service, or is the validation done locally? I didn’t get that from a quick look through the code, but if applicable I always like it if packages clearly state when and how code might leave your machine.
- any particular reason to use a new tab instead of the output panel for the output? That’s kind of what the output panel is for 😁

braver:
👋🏻

PR #8798, opened 2023-07-25
SyncedSideBar: add release for older sublime versions

TheSpyder:

My package is for synchronising the sidebar with the currently active file.

I would like to start using new Python features in the package. I've tagged the current version so anyone using older releases doesn't get a package update that breaks.

I haven't been using tag versions up to this point, I hope I've done it correctly.

[2]: https://semver.org

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -5819,8 +5819,12 @@
 			"details": "https://github.com/TheSpyder/SyncedSideBar",
 			"releases": [
 				{
-					"sublime_text": "*",
-					"branch": "master"
+					"sublime_text": "<4050",
+					"tags": "st3-"
+				},
+				{
+					"sublime_text": ">=4050",
+					"tags": true
 				}
 			]
 		},


FichteFoll:
Both [4.0.0](https://github.com/TheSpyder/SyncedSideBar/releases/tag/4.0.0) and [st3-3.0.0](https://github.com/TheSpyder/SyncedSideBar/releases/tag/st3-3.0.0) point to the same commit despit using a different version number. Was that intended? The code of this commit (the latest) appears to be compatible with ST3 but the version being different is admittedly a bit weird. I would have expected the tag for ST4 to be `3.0.0`.

TheSpyder:
It's intended but temporary.

There are changes in https://github.com/TheSpyder/SyncedSideBar/pull/64 which, as I understand it, require ST4. But because _all_ existing users are pulling from the `master` branch, if I merge it now older ST versions will break. I plan to merge it and do a new 4.0 release as soon as this channel update is published.

FichteFoll:
I'd still suggest to release the current version of the master branch as `3.0.0` instead of attempting some deletion shenanigans of the already pushed 4.0.0 tag, since that update would be missed by everyone already on the "fake" 4.0.0. Unless you mean to do it as 4.0.1 or 4.1.0 which would … work, I guess, but not exactly sit right with semantic versioning at least. Not that this would be a problem since users can't pin specific versions anyway.

TheSpyder:
The numbers aren't too important - I could call the ST3 branch `ST3-1.0.0` if I wanted, because the package hasn't had version numbers before. I just chose 3 and 4 because... well I guess because that's the sublime version they're targeting.

Let's step back a moment. I would like to publish a channel update that makes all users stop using `master` as their source of versions before I start merging breaking changes to it. Which means a "fake" release that uses the same commit for both releases. I then want to tag the `master` branch in such a way that only ST4 users will download it.

From looking at what other packages do, I thought the best way to achieve this was to prefix the tags with something like `ST3-` for older Sublime releases, and use normal semver for the current release.

When I release the changes that would break ST3 users, I think it will totally fine for those to be a patch or minor release - because the release will be functionally identical. It will just delete old code not used by ST4 and upgrade the python version.

If I've done this incorrectly I'm quite happy to change it. There's no rush on these changes, they are purely to modernise the codebase not add new features.

FichteFoll:
> I just chose 3 and 4 because... well I guess because that's the sublime version they're targeting.

You may of course do that because with the change in this PR, the two release channels for ST4 and ST3 are entirely separate. It's just a bit weird or unheard of mark the exact same commit as two different versions (when it's also the first version, essentially). The only implications that this could have is when you start adding another release channel that overlaps with ST3 and ST4, i.e. might have two source versions to update from. In that case you'd need to ensure that the new channel with ST build target overlaps is higher than both of the potential previous versions. But that is very unlikely to happen.

I'm okay with you starting at any version you wish. I just find it a bit weird and potentially misleading or unexpected for users, but most will never notice anyway. Give me a heads-up whether you want to consider this a bit more or merge it in its current state. You won't be able to go to a lower version (e.g. 2.0.0) afterwards, but that shouldn't be an issue for a package that doesn't plan on having a lot of development anyway.

TheSpyder:
Thank you for the explanation. I understand your concern now.

My current plan is to only ever have two channels - the current code (which will become the ST3 release channel) still has ST2 compatibility. If, in the future, there's a reason to branch again for a future ST5 or similar, then I'll consider what to do in order to never have overlapping channel conditions.

I agree, in an ideal state I would have switched to versioning with the ST3 compatible release some time ago and then done development for a ST4 version on `master` before splitting my release channels with this change.

As I waited until this point to have _any_ versions, and I'm unlikely to release further updates for ST3, it's unfortunately left me in the weird state where both channels have the same commit. Please merge as is.

PR #8802, opened 2023-08-01
add LogbackJsonLogViewer plugin

neno--:

My package is a viewer of log files that are stored in JSON format. It is particularly useful for Java Logback style files. It can highlight and toggle various severities etc.

It adds a context menu item only if the active document is indeed something that can be viewed as JSON formatted log.
In that case, a new tab is opened with a formatted log. Keybindings are only active in this tab.

[https://github.com/neno--/LogbackJsonLogViewer](https://github.com/neno--/LogbackJsonLogViewer)

[2]: https://semver.org

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -1561,7 +1561,7 @@
 		},
 		{
 			"name": "LogbackJsonLogViewer",
-			"author": ["Nenad Nenadović"],
+			"author": "Nenad Nenadović",
 			"details": "https://github.com/neno--/LogbackJsonLogViewer",
 			"labels": ["logback", "json", "log", "logging"],
 			"releases": [
--- a/repository/l.json
+++ b/repository/l.json
@@ -1561,7 +1561,7 @@
 		},
 		{
 			"name": "LogbackJsonLogViewer",
-			"author": ["neno--"],
+			"author": ["Nenad Nenadović"],
 			"details": "https://github.com/neno--/LogbackJsonLogViewer",
 			"labels": ["logback", "json", "log", "logging"],
 			"releases": [
--- a/repository/l.json
+++ b/repository/l.json
@@ -1551,6 +1551,18 @@
 			]
 		},
 		{
+			"name": "LogbackJsonLogViewer",
+			"author": ["neno--"],
+			"details": "https://github.com/neno--/LogbackJsonLogViewer",
+			"labels": ["logback", "json", "log", "logging"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Logger Snippets",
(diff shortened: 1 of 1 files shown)


braver:
Thanks for your submission, some feedback:
- You should create the command to open the settings file you mention in the readme, and create a main menu entry as well.
- Your processing in is_visible is potentially quite heavy. If you're trying to figure out if a file is JSON, why don't you simply read the file's scope (ie. source.json)?

braver:
👋🏻

neno--:
Hello @braver thanks for your feedback:

- I added the command to open and edit configs. It is available from Settings > Package Settings > LogbackJsonLogViewer menu.
- I am not processing is_visible anymore. It was an attempt to use heuristics to see if the file was a good candidate for processing. These files would mostly consist of JSON, but some components could add lines that are not JSON. Now users can choose whether they want to view any file with this plugin.

braver:
Please look for some examples on how to use edit_settings to open settings in Split View. Having separate menu entries for user and default settings feels like something from the nineties by now 😉 

> I am not processing is_visible anymore. It was an attempt to use heuristics to see if the file was a good candidate for processing.

Alright. We do like seeing a setting to disable context menu entries. This gives users some control over what shows up in that menu, which tends to get crowded because it's prime real-estate that all packages want to be in. If your package is not something I use every day, I don't need it there. Instead, all commands should be available via the command palette anyway (you might need to add it there still) and that's also the first place people go to look for functionality.

neno--:
I used the edit_settings command to open plugin settings. Also, added the command to the command palette.

braver:
Alright, that's better 👍🏻 

- Please have a look at the indentation in your keymap files, they look a bit wonky.
- For main menu entries where you're nesting inside existing entries (e.g. "Package Settings") please only provide the id without the caption (you would actually be overwriting the caption otherwise).
- Preferably implement a setting to disable the context menu item, since it's now unconditionally visible in all views.

braver:
👋🏻

neno--:
Hello @braver thanks for your feedback. I fixed all three issues.

braver:
👋🏻

neno--:
Hi @braver tag v2.0.0 created: https://github.com/neno--/LogbackJsonLogViewer/releases/tag/v2.0.0

repository/l.json
```diff
@@ -1559,6 +1559,18 @@
 				}
 			]
 		},
+		{
+			"name": "LogbackJsonLogViewer",
+			"author": ["neno--"],
```
braver, line 1564: Please remove the “author” field, it’s not needed since it’s just your username.

braver (changes requested):

repository/l.json
braver: Sorry for not catching that earlier!

neno--:
Hi, @braver I added my full name.

PR #8804, opened 2023-08-03
Add NeoVintageous packages Colemak and Dvorak

gerardroche:

My packages are [NeoVintageous](https://github.com/NeoVintageous/NeoVintageous) plugins that support alternate keyboard mappings: [NeoVintageousDvorak](https://github.com/gerardroche/NeoVintageousDvorak) and [NeoVintageousColemak](https://github.com/gerardroche/NeoVintageousColemak). 

I made them individual packages so users can contribute to the mappings packages without the burden of having to understand the internals of NeoVintageous. Also, as an example of how to use the source feature.

NeoVintageous v1.32 will be available soon. It eases the setup of both packages by allowing users the ability autoload the mappings packages. 

~I'm marking this as a Draft until I've released v1.32, which should be some time next week. :sparkles: :cake: :rocket:~

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -302,6 +302,28 @@
 			]
 		},
 		{
+			"name": "NeoVintageousColemak",
+			"details": "https://github.com/gerardroche/NeoVintageousColemak",
+			"labels": ["vim", "neovintageous", "vintageous", "vintage"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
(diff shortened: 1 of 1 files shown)


gerardroche:
packagecontrol-bot is requesting changes:

```
ERROR: The package does not define any file that interfaces with Sublime Text
```

This is expected. The Dvorak and Colemak packages provides alternate mappings that get sourced by NeoVintageous.

PR #8806, opened 2023-08-08
CK3 Tools Package

SnowCrystalPDS:
Added a package for Crusader Kings 3 Sublime tools used by the CK3 team, so it can be used by modders if they so wish.

- [X ] I'm the package's author and/or maintainer.
- [X] If my package is a syntax it doesn't also add a color scheme. (It does add a color scheme, but it is only used by the team for very minor specific additions. All the other stuff can be used by all color schemes)
- [X] If my package is a syntax it is named after the language it supports (without suffixes like "syntax" or "highlighting").

My package is ...

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1269,6 +1269,17 @@
 			]
 		},
 		{
+			"name": "CK3 Tools",
+			"details": "https://github.com/SnowCrystalPDS/CK3-Tools",
+			"labels": ["CK3", "Crusader Kings 3"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Claat Snippets",
 			"details": "https://github.com/ucl-casa-ce/claat-snippets-sublime",
 			"homepage": "https://www.connected-environments.org",


dementive:
Funny seeing this here, I just submitted [Imperator Rome and Victoria 3 packages](https://github.com/wbond/package_control_channel/pull/8807) as well. Jomini based paradox games will prevail! A lot of my work was actually based originally on your Imperator syntax snowcrystal and I believe my syntaxes made for imperator and victoria 3 could have lots of elements that you could copy into the ck3 package because the syntax is extremely similar between the 3 games.

braver:
Please have a look at the bot output.

braver:
👋🏻

SnowCrystalPDS:
> Please have a look at the bot output.

Hello there, sorry for the late update, I hadn't seen this before now. I have changed the files from being packaged to just being the base files now, that should solve the main issue, right?

braver:
Your package has no README and no way of understanding what it does or why. How should people who discover this package through Package Control understand what it is? If it's for a closed audience, you don't need to go through package control to efficiently distribute and update a package.

From reviewing the package, I have some concerns. It seems like you support some settings, but they're not clearly documented and you don't provide ways of opening/editing them. 

As is mentioned in comments in the code (why not in the readme?) the plugin (plugins in Sublime Text are typically called Packages, and never "ploogin" by the way ;) ), is potentially very heavy. On plugin_loaded but presumably also on_hover of basically everything? I looks to me like you should disable most behaviors unless they're enabled in a specific project where the functionality actually applies.

I'm also not sure why it ships with a syntax for yaml files.

As it stands, the package looks very naive. Which may be fine for a particular use case and you know how to use it. But for consideration for the general public via Package Control I'm afraid you still have some work to so.

braver:
👋🏻

PR #8808, opened 2023-08-19
Add Inline Fold plugin

predragnikolic:

My package is ...
meant to fold regions. This plugin can be useful when working with TailwindCSS to hide the long strings.

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -726,6 +726,16 @@
 			]
 		},
 		{
+			"name": "Inline Fold",
+			"details": "https://github.com/predragnikolic/InlineFold",
+			"releases": [
+				{
+					"sublime_text": ">=4070",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Inline Python",
 			"details": "https://github.com/apiad/Sublime-InlinePython",
 			"releases": [


braver:
Ha, don’t get me started on Tailwind and those classes 😂

Looks pretty neat and easy to use. 👍🏻 

From browsing the code, I’m not sure how well you’re guarding against views disappearing in the middle of an operation or after a debounce. In my experience anything even slightly async needs to do regular checks if views are still there (ie. the user didn’t cmd+w closed it, or maybe it was a transient view).

braver:
You might need to think about your SEO a bit here. I didn’t understand what the package does until I read the full readme. It sounds like something especially useful to developers working with css, and perhaps it could also work with doc blocks? Anyway, maybe put some choice keywords in “labels” and hint at use cases in your repo description.

predragnikolic:
Thanks!

> I’m not sure how well you’re guarding against views disappearing in the middle of an operation or after a debounce.

I pushed a [commit](https://github.com/predragnikolic/InlineFold/commit/96ba83273dcb782ec8eaaa49a4d3f004722e9ab9) for that.

Also I updated the README(with some examples) and added some labels on the Github repo.
Hopefully it is more clearer now.

braver:
Looks good 👍🏻

predragnikolic:
Thanks again :+1:

PR #8809, opened 2023-08-27
New Package: Reviw

livrasand:

My package is ...

[2]: https://semver.org

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -1568,11 +1568,9 @@
 		{
 			"name": "Reviw",
 			"details": "https://github.com/livrasand/Reviw",
-			"homepage": "https://livrasand.github.io",
 			"labels": ["jwpub", "reviw", "jw library", "jw", "snippets"],
 			"author": "Livrädo Sandoval",
 			"donate": "https://www.paypal.com/paypalme/livrados",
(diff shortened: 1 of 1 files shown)


braver:
- Please tag versions using semver notation.
- That homepage does not seem relevant to your package?
- Why are you linking to that readme while the repo already has a normal readme markdown file in it?
- Please tell us a little about your package.

braver:
- What use are the .jwtkn files in the context of this package?
- It looks like you're offering mostly snippets, one set with Spanish help texts, one with Italian? While it's fine to offer non-English functionality, that does limit your target audience. In any case, when provided like this, all users will see duplicate snippets.. that can't be what you intended (or I'm missing something).
- What is the purpose of readme.texttile and the [install.txt](https://github.com/livrasand/Reviw/blob/main/Reviw%20IT/messages/install.txt) files? Currently users will never see them. If you want to show a message on install of the package, you need to provide a messages.json file to tell Sublime where to find the text file. Also, large ASCII art is usually frowned upon here. People already installed the package, you don't need to market it anymore. Also consider that users might be installing multiple packages, and your art will take up the entire screen, hiding all otherwise useful information.

braver:
This bit is still a concern:

>  In any case, when provided like this, all users will see duplicate snippets.. that can't be what you intended (or I'm missing something).

PR #8811, opened 2023-09-01
Add Gno package

jdkato:

My package adds syntax highlighting for the [Gno programming language](https://gno.land/).

[2]: https://semver.org

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -1160,6 +1160,17 @@
 			]
 		},
 		{
+			"name": "Gno",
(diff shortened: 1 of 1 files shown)


braver:
Thanks! I understand these are like dialects of go. The pattern usually is that the extension of the main scope (`source.gno`) aligns with the specific scopes like `variable.gosum`. Currently the `n` is missing from the specific scopes, that's unexpected to me. Is that a typo, or intentional, or...?

jdkato:
Nope, thanks for the review. I corrected the scopes and pushed a new release.

PR #8813, opened 2023-09-05
add VS Code Dark Modern Color Scheme

rexarvind:

My package is to have VS Code like Dark Modern Color Scheme in Sublime Text

### DIFF
--- a/repository/v.json
+++ b/repository/v.json
@@ -597,6 +597,18 @@
 				}
 			]
 		},
+        {
+            "name": "Visual Studio Color Scheme",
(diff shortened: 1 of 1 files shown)


braver:
- You need to create a semver versioned tag (with 3 digits, e.g. 1.0.0)
- You don't need to add the issues link
- Please have a look at your indentation
- A better name might be VS Code Dark Modern, "Visual Studio Color Scheme" is a bit broad if there is only one color scheme in your package (VS Code ships with several), and Visual Studio is something quite different from VS Code.

braver:
👋🏻

PR #8814, opened 2023-09-06
add package CustomBufferName

kaxing:

My package is [CustomBufferName](https://github.com/kaxing/sublime-custom-buffer-name). It allows customization of the default "untitled" buffer name and automatically adds numbering when editing a new buffer.

[2]: https://semver.org


braver:
Is this not very similar (at least in intent) as https://packagecontrol.io/packages/Rename%20Buffer, and might it not make more sense as an additional feature of that existing package?

kaxing:
Hi @braver , this plugin is inspired by "Buffer Rename". Therefore, "Custom Buffer Name" does not provide the option to rename the current buffer in the command palette. Both plugins work well alone or with each other.

braver:
I see you do a lot of settings.set, and saving to a settings file. But you don't expose a settings file to the user. That seems a bit odd. It's a bit uncommon to only provide a wizard like interaction with a setting, and not simply exposing that setting to users to look at and edit like any other setting. 

You also don't have a default settings file, nor seem to be relying on the second argument to `settings.get`, allowing you to have defaults easily (without relying on work in plugin_loaded to ensure a setting exists). That all seems a bit roundabout, a simple (default + user) settings file, some menu and palette entries, would make it easier for users to interact with that setting, and easier for you to use that setting.

You seem to have a default with "%Y-%M-%d %A %H:%M ", but is that used? It doesn't look like that format is actually converted into an actual timestamp?

kaxing:
Hi, braver,
Thanks for the feedback. You are right, I wasn't thinking of using a setting file because there is only one thing to keep.
I've added the default setting file and exposed the settings as a palette entry to the user by adding the edit_settings command.
The default_prefix value is populated within the format function. Please give it a try and let me know if there are any other suggestions.

braver:
> and exposed the settings as a palette entry to the user by adding the edit_settings command.

Even though it's just a single setting, please also provide a main menu entry. That's where people expect to find settings, the command palette is a convenient additional way of doing it but packages should follow all normal patterns.

kaxing:
@braver [Added](https://github.com/kaxing/sublime-custom-buffer-name/commit/3fea801a600f3347be0fb2f35a0d82108410537f). Let me know if there are other things that I am missing; thank you!
Screenshot of Sublime Text with plugin settings menu entry for Custom Buffer Rename

braver:
No, I think we covered everything. Great work, let's ship it.

PR #8816, opened 2023-09-12
Diffable package added.

yaroslavyaroslav:

My package is a wrapper around sublime built in incremental diff tool, plus in addition it provides an option to pass the content to Kaleidoscope diff app.

My package is similar to [Diffy](https://github.com/zsong/diffy). However it should still be added because diffy is quite deprecated for a years, I've even tried to contribute there, and my PR even got accepted, yet it's still latest version in the channel is from 2015 year. Also it has quite ugly UI and inaccurate and slow diff also.

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -890,6 +890,16 @@
 			]
 		},
 		{
+			"name": "Diffable",
+			"details": "https://github.com/yaroslavyaroslav/Diffable",
+			"releases": [
+				{
+					"sublime_text": ">=3200",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "DiffTabs",
 			"details": "https://github.com/soandrew/DiffTabs",
 			"releases": [
--- a/repository/d.json
+++ b/repository/d.json
@@ -892,6 +892,7 @@
 		{
 			"name": "Diffable",
 			"details": "https://github.com/yaroslavyaroslav/Diffable",
+			"previous_names": ["Diffy"],
 			"releases": [
 				{
 					"sublime_text": ">=3200",
--- a/repository/d.json
+++ b/repository/d.json
@@ -892,6 +892,7 @@
 		{
 			"name": "Diffable",
 			"details": "https://github.com/yaroslavyaroslav/Diffable",
+			"donate": "https://github.com/sponsors/yaroslavyaroslav",
 			"previous_names": ["Diffy"],
 			"releases": [
 				{
--- a/repository/d.json
+++ b/repository/d.json
@@ -922,17 +922,6 @@
 			]
 		},
 		{
-			"name": "Diffy",
-			"details": "https://github.com/zsong/diffy",
-			"labels": ["diff", "compare", "difference", "comparison"],
-			"releases": [
-				{
-					"sublime_text": "*",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "Dilmes Ipsum",
 			"details": "https://github.com/thiagoh/sublime-dilmes-ipsum",
 			"author": "Thiago Andrade",


yaroslavyaroslav:
Those conflicted bindings are dismissed now. Could some one please rerun these tests?

yaroslavyaroslav:
@packagecontrol-bot please smoke tests

yaroslavyaroslav:
@braver hey, just wondering is it something wrong with this package or it's just waiting its turn to be properly reviewed?

braver:
This is a community project, we'll get to it when we get to it. It usually takes a few weeks for new submissions to float to the top of the stack.

Couple of things:

- First off: looks good 👍🏻 
- I think it would be a good idea to replace the existing Diffy package. It's well and truly dead it seems. Is yours a drop-in replacement for existing users?
- Please add a setting or something so that users can easily disable the context menu. It takes up prime real estate and it's not necessarily the easiest way of interacting with it.
- You specifically implement the Kaleidoscope app. It looks like it takes pretty standard arguments (left file + right file), and a lot of other diffing apps will work the same way. See also [SidebarTools](https://github.com/braver/SideBarTools/blob/master/SideBar.py#L53). In fact, might make sense to be able to select 2 files in the sidebar instead of specifically 2 views.
- In your Main Menu, for those standard nodes like "Preferences", just provide the id, not the caption.

yaroslavyaroslav:
Thanks for kinds words.

    - [+] I think it would be a good idea to replace the existing Diffy package. It's well and truly dead it seems. Is yours a drop-in replacement for existing users?
    - [-] Please add a setting or something so that users can easily disable the context menu. It takes up prime real estate and it's not necessarily the easiest way of interacting with it.
    - [-] You specifically implement the Kaleidoscope app. It looks like it takes pretty standard arguments (left file + right file), and a lot of other diffing apps will work the same way. See also [SidebarTools](https://github.com/braver/SideBarTools/blob/master/SideBar.py#L53). In fact, might make sense to be able to select 2 files in the sidebar instead of specifically 2 views.
    - [x] In your Main Menu, for those standard nodes like "Preferences", just provide the id, not the caption.

I'm quite confused. As long as I'm aware there's now way to enable context menu block (e.g. DIffable), GPT-4 haven't help me here also. I believe the better way would be just to drop it completely.

I'm now sure about your third point. If you are talking about to generalise third party tool comparison here? I'm up for that, but just haven't tested whether the given python code would fit other diff tools.

And about the first one: What practically outcome from that point? Do you suggest to rename the whole plugin to `Diffy` don't you? If so again I'm confused coz I believe this would lead to git repo renaming, isn't it?

braver:
> As long as I'm aware there's now way to enable context menu block

- I would get rid of the separator. Not every group of commands needs it's own delineated block, that just takes up space.
- I kind of assume that the "Diffable (2 Columns)" parent item disappears if all commands in it are unavailable. You could make an intermediate command for the context menu items, and set the is_visible for that command to return false if the context menu is disabled via a setting.

> If you are talking about to generalise third party tool comparison here

More as a suggestion for future development. Kaleidoscope is very niche, and lots of other apps can be launched in a similar manner.

>  Do you suggest to rename the whole plugin to Diffy don't you?

Package control references packages by name. If your package is mostly a drop in replacement for another abandoned package, you could change the details of that package to point to your repo. Or we remove the existing package and add its name to your package via previous_names. Some examples: https://github.com/wbond/package_control_channel/pulls?q=is%3Apr+label%3Atakeover+is%3Aclosed

yaroslavyaroslav:
> * I kind of assume that the "Diffable (2 Columns)" parent item disappears if all commands in it are unavailable.

Actually I assumed it either, but it appears that it's not. To be honest I'm completely confused with this very feature at a whole. (1) menu doesn't disappear even if it's empty (see screenshot, Diffable menu are hovered). (2) it failed to work well if I'm following examples that you've provided above: both function `is_visible` and `is_available` have different attribute signatures in the current ST version, so I've to use a workaround for those.

Anyway, I consider your issue with menu as a fair one, so I implemented it in a way as I was able it to settle.

> Package control references packages by name.

Done, should I do something else within this repo by myself to evaporate `Diffy` from a package control search?

Screenshot 2023-09-24 at 13 51 21

braver:
> should I do something else within this repo by myself to evaporate Diffy from a package control search

Yes, in fact, the existing entry for Diffy should be removed.

yaroslavyaroslav:
@braver Yep, it's already done, CI was failing otherwise.

braver:
Alright, looking good. As a matter of protocol I would like to give @zsong 2 weeks to object to the replacement and come up with an alternative approach. If nothing comes from that, we'll merge this PR.

yaroslavyaroslav:
@braver Hi, seems like the time has come.

braver:
Alright, I would say that all that's missing now is a 2.0 upgrade message (for existing users) explaining the change of name and ownership of the package and any changes that might be relevant for them.

yaroslavyaroslav:
@braver done https://github.com/yaroslavyaroslav/Diffable/blob/develop/messages/2.0.0.md

braver:
I think you need to put the full path to the 2.0.0 message in the json (ie. unlike the readme it's not in the root dir).

yaroslavyaroslav:
My bad, fixed. https://github.com/yaroslavyaroslav/Diffable/blob/master/messages.json

yaroslavyaroslav:
Occasionally revealed that there's an option to add donation page to a package, therefore added one.

PR #8817, opened 2023-09-14
Add FabricEngine Package

zamadatix:
Add FabricEngine package repository definition in f.json

My package is similar to some other switch config file syntax highlighting packages for other networking vendors but is unique in that it supports the specific syntax of the Extreme FabricEngine switch platform, which is incompatible with other syntax definitions.

Note: The repo name is more generic but overridden so the package name is set to just "FabricEngine" per the naming guidelines.

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -46,9 +46,9 @@
 		{
 			"name": "FabricEngine",
 			"details": "https://github.com/stepcg/SublimeFabricEngineSyntax",
+			"labels": ["FabricEngine", "VOSS", "language syntax"],
 			"releases": [
 				{
-					"labels": ["FabricEngine", "VOSS", "language syntax"],
 					"sublime_text": "*",
 					"tags": true
 				}
--- a/repository/f.json
+++ b/repository/f.json
@@ -46,7 +46,7 @@
 		{
 			"name": "FabricEngine",
 			"details": "https://github.com/stepcg/SublimeFabricEngineSyntax",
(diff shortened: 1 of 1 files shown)


repository/f.json
```diff
@@ -43,6 +43,16 @@
 				}
 			]
 		},
+		{
+			"name": "FabricEngine",
+			"details": "https://github.com/stepcg/SublimeFabricEngineSyntax",
```
braver, line 48: Please add the "language syntax" label to make your package easier to find, and help people understand exactly what it is.

braver:
Please have a look at the [Scope Naming](https://www.sublimetext.com/docs/scope_naming.html) and [examples](https://github.com/sublimehq/Packages). Your main scope should probably be `source.fabricengine` (to enable features for source code, not plain text which would be enabled by `text.fabricengine`), and each other scope should end with `.fabricengine` to allow features to target your syntax specifically (e.g. `variable.language.fabricengine`.

zamadatix:
> Please have a look at the [Scope Naming](https://www.sublimetext.com/docs/scope_naming.html) and [examples](https://github.com/sublimehq/Packages). Your main scope should probably be `source.fabricengine` (to enable features for source code, not plain text which would be enabled by `text.fabricengine`), and each other scope should end with `.fabricengine` to allow features to target your syntax specifically (e.g. `variable.language.fabricengine`.

Thanks for the tip! I've cleaned that up a bit, hopefully it looks a bit more correct now.

PR #8818, opened 2023-09-14
Add Gitk and Khan

gerardroche:
**Gitk**

Launch gitk from within Sublime.

**Khan**

Add, remove, and clear rulers at the current cursor using the command palette.

My packages are:

- **Gitk**: Launch gitk from within Sublime.
- **Khan**: Add, remove, and clear rulers at the current cursor using the command palette.

My rulers package, Kkan, is similar to others, but it should be added anyways. It is a simple, lightweight, command-palette only, package.

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -830,6 +830,17 @@
 			]
 		},
 		{
+			"name": "Gitk",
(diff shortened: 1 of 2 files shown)


braver:
Nice, simple, useful, good job 👍🏻

PR #8819, opened 2023-09-15
add Prettierd Format

smastrom:

Repo: https://github.com/smastrom/sublime-prettierd-format

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -2247,6 +2247,18 @@
 			]
 		},
 		{
+			"name": "Prettierd Format",
+			"details": "https://github.com/smastrom/sublime-prettierd-format",
+			"author": "Simone Mastromattei",
+			"labels": ["prettier", "prettierd", "format"],
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Pretty EDN",
 			"details": "https://github.com/oakmac/sublime-pretty-edn",
 			"author": "Chris Oakman",


braver:
Thanks!

Would the existing [Fmt](https://packagecontrol.io/packages/Fmt) package also be able to support prettierd? That might be beneficial to users because it already supports a lot of configuration and has implemented the behaviors and tested them across many use cases. It would also make maintenance a bit easier in the long run.

In any case, your settings should be accessible via the main menu.

smastrom:
Hi @braver, yes prettierd works with Fmt. This is how it can be configured:

```json
{
   "rules": [
      {
         "selector": "source.ts",
         "cmd": [
            "prettierd",
            "--stdin-filepath",
            "$file"
         ],
         "format_on_save": true,
      }
   ]
}
```

The only gotcha seems to be that rules must be entered manually for each scope.

When I recently started using Sublime, I struggled to find a way to format files faster using prettier. I had no luck searching the forum, googling or browsing issues that I ended up discovering prettierd and building this plugin.

For this reason, I doubt any novice user can discover Fmt and the above integration easily.

I think it would be great to have a straightforward solution directly in Package Control by just typing `prettier`. 

I've added the menu settings and an installation message that informs that `prettierd` should be installed on the machine in order to use this plugin.

smastrom:
Hi @braver, thank you for your feeback.

Yes prettierd works with Fmt. This is how it can be configured:

```json
{
   "rules": [
      {
         "selector": "source.ts",
         "cmd": [
            "prettierd",
            "--stdin-filepath",
            "$file"
         ],
         "format_on_save": true,
      }
   ]
}
```

The only gotcha seems to be that rules must be entered manually for each scope.

When I recently started using Sublime, I struggled to find a way to format files faster using prettier. I had no luck searching the forum, googling or browsing issues that I ended up discovering prettierd and building this plugin.

For this reason, I doubt any novice user can discover Fmt and the above integration easily.

I think it would be great to have a straightforward solution directly in Package Control by just typing `prettier`. 

I've added the menu settings and an installation message that informs that `prettierd` should be installed on the machine in order to use this plugin.

braver:
> The only gotcha seems to be that rules must be entered manually for each scope.

I guess that's how it works, but of course you can combine scopes in the selector (e.g. "source.ts, source.js" should work). 

It might be worth to add your example here to the documentation of Fmt, so that users who need multiple formatters, or are savvy with packages like that can use that. In the end it's better for users (because of overall quality of the package, consistency across multiple similar formatters, etc) if functionality is combined in to a higher level package like Fmt. Similar to SublimeLinter for instance.

That said, I'm not 100% against a specific package like yours, just trying to advocate collaborating on these things rather than creating more and more of the same (as is usually typical of open source development). 

> I've added the menu settings and an installation message that informs that prettierd should be installed on the machine in order to use this plugin.

👍🏻

PR #8823, opened 2023-09-18
Add Compass Navigator

kapitanluffy:

My package is [Compass Navigator](https://github.com/kapitanluffy/sublime-compass)

[2]: https://semver.org

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3259,6 +3259,18 @@
 			]
 		},
 		{
+			"name": "Compass Navigator",
+			"details": "https://github.com/kapitanluffy/sublime-compass",
+			"donate": "https://github.com/sponsors/kapitanluffy",
+			"labels": ["code navigation"],
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Competitive Programming",
 			"details": "https://github.com/egtoney/CompetitiveProgramming",
 			"author": "egtoneyUK",
--- a/repository/t.json
+++ b/repository/t.json
@@ -3517,17 +3517,6 @@
 			]
 		},
 		{
-			"name": "Tweaks",
-			"details": "https://github.com/kapitanluffy/sublime-tweaks",
-			"donate": "https://github.com/sponsors/kapitanluffy",
-			"releases": [
-				{
-					"sublime_text": ">=4000",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "Twee",
 			"details": "https://github.com/monospaced/sublime-twee",
 			"releases": [
--- a/repository/t.json
+++ b/repository/t.json
@@ -3506,6 +3506,17 @@
 			]
 		},
 		{
+			"name": "Tweaks",
+			"details": "https://github.com/kapitanluffy/sublime-tweaks",
+			"donate": "https://github.com/sponsors/kapitanluffy",
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Twee",
 			"details": "https://github.com/monospaced/sublime-twee",
 			"releases": [


jwortmann:
The kind icons for the different sheet types in the quick panel should not be defined as "function", "variable", "keyword", etc., because sheets are semantically completely unrelated to those predefined kinds.

Take a look how e.g. LSP does it to use the desired colors:
https://github.com/sublimelsp/LSP/blob/b7e32635ac9fefe985b9a76a6d2991153a42c54a/plugin/core/views.py#L106-L113

Or the CommandsBrowser package:
https://github.com/Sublime-Instincts/CommandsBrowser/blob/16ef299085854d1f4b61aaf8280c23c3f98a43a5/src/utils/miscellaneous_utils.py#L18-L28

kapitanluffy:
> The kind icons for the different sheet types in the quick panel should not be defined as "function", "variable", "keyword", etc., because sheets are semantically completely unrelated to those predefined kinds.
> 
> Take a look how e.g. LSP does it to use the desired colors: https://github.com/sublimelsp/LSP/blob/b7e32635ac9fefe985b9a76a6d2991153a42c54a/plugin/core/views.py#L106-L113
> 
> Or the CommandsBrowser package: https://github.com/Sublime-Instincts/CommandsBrowser/blob/16ef299085854d1f4b61aaf8280c23c3f98a43a5/src/utils/miscellaneous_utils.py#L18-L28

Thank you! Updated it. 

Curious though, where are the last two values of the tuple used? I see that you can also make them empty.

jwortmann:
> Curious though, where are the last two values of the tuple used? I see that you can also make them empty.

https://www.sublimetext.com/docs/completions.html#kind-info

The letter is shown as the icon and the third value of the tuple is shown in the tooltip when you hover with the mouse over the icon. And I think in case of autocompletion it's also shown in the details pane at the bottom of the popup when that is visible.

Btw, the reason of my suggestion from above is that the predefined (non-color) IDs like KIND_ID_FUNCTION will result in special attributes being set on the theme class rule. For example I use a little "trick" in my own theme to make the kind letter transparent for the predefined kinds, and instead show a special icon in that case.

kind_icons

But of course the icons wouldn't make sense anymore if the kind IDs are used by plugins for unrelated things, like in this example for the different sheet types. (But for the "colorish" kind IDs it will be rendered with the letter still visible, like in the default themes.)

kapitanluffy:
> I use a little "trick" in my own theme to make the kind letter transparent for the predefined kinds, and instead show a special icon in that case.

Whoa! So, I can show icons instead of letters? Can you share that? I think icons would be a more suitable visual guide.

braver:
Pretty awesome package, made even more awesome by the interaction here. 

Your keymap still has two entries for debugging purposes. I doubt a general user of the package would be able to memorize the bindings for those, might as well disable those two and instead add them to the command palette.

You're still missing main menu entries for your settings and key bindings. You might add those debugging commands there as well. 

Personally I would suggest reducing the install message a bit. You already "sold" it to the user: they installed it. If that message has lots of info, people won't read it (as far as users ever read anything...). I would suggest focussing on the fact that the package really is better with keybindings and help users set that up. Also, the version history there is not super relevant for first-time installation. 

Other than those details though, very nicely done 👌🏻

kapitanluffy:
> Pretty awesome package, made even more awesome by the interaction here.
> 
> Your keymap still has two entries for debugging purposes. I doubt a general user of the package would be able to memorize the bindings for those, might as well disable those two and instead add them to the command palette.
> 
> You're still missing main menu entries for your settings and key bindings. You might add those debugging commands there as well.
> 
> Personally I would suggest reducing the install message a bit. You already "sold" it to the user: they installed it. If that message has lots of info, people won't read it (as far as users ever read anything...). I would suggest focussing on the fact that the package really is better with keybindings and help users set that up. Also, the version history there is not super relevant for first-time installation.
> 
> Other than those details though, very nicely done 👌🏻

Thank you. I'll work on it later and let you know.

kapitanluffy:
@braver created a [new release](https://github.com/kapitanluffy/sublime-compass/releases/tag/2.1.1)

braver:
Thanks! 

Final remark: for existing menu entries please only provide the id ("package_settings") not the mnemonic or the label: you're otherwise effectively overriding those.

kapitanluffy:
> Thanks!
> 
> Final remark: for existing menu entries please only provide the id ("package_settings") not the mnemonic or the label: you're otherwise effectively overriding those.

Ah, PackageDev added that. I did not notice. Made a re-release 😅

PR #8824, opened 2023-09-29
Adds ruffix to the default channel.

drcongo:

My package is ...

A simple interface to `ruff --fix` for automatically fixing issues in Python files.

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -2425,6 +2425,16 @@
 			]
 		},
 		{
+			"name": "Ruffix",
+			"details": "https://github.com/drcongo/ruffix",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "RulerColumn",
 			"details": "https://github.com/gerardroche/sublime-ruler-column",
 			"releases": [


braver:
Thanks for your submission. It seems some JSON files need some work. From a quick look at your code it looks like Ruff runs on files on disk. Files in Sublime Editor are just a projection and might be out of sync with the file on disk. Running this command would potentially update a file that's "dirty", resulting in a conflict (the user will receive message that the file on disk has changed, asking them to reload the file). So you need to resolve that somehow. 

The way I use tools like this personally, I that I don't make a command, I make a [build system](https://www.sublimetext.com/docs/build_systems.html), simply with a json file. You can set build systems to save the file before running. And in many cases you can even set it to catch and display (inline in the code) any errors that the tool might return. Unless you want to do more than simply running a file through that utility, build systems are a simpler way of wrapping a utility like this and at the same time also more powerful. You can also set up [key bindings to trigger your specific build system](https://stackoverflow.com/questions/34340770/custom-build-key-binding-sublime-text).

drcongo:
Thanks for the feedback @braver - that makes perfect sense.

rwols:
There is also LSP-ruff

PR #8825, opened 2023-09-30
Add XdebugHelper

maliayas:

My package is [XdebugHelper](https://github.com/maliayas/SublimeText_XdebugHelper)

[2]: https://semver.org

### DIFF
--- a/repository/x.json
+++ b/repository/x.json
@@ -100,6 +100,17 @@
 			]
 		},
 		{
+			"name": "XdebugHelper",
+			"details": "https://github.com/maliayas/SublimeText_XdebugHelper",
+			"labels": ["php", "xdebug"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "xit!",
 			"details": "https://github.com/jotaen/xit-sublime",
 			"releases": [


braver:
Thanks for your submission. Sorry we kept it sitting there for a while, a package like this needs a bit more thinking about to review then say a color scheme :)

- There are multiple Xdebug packages in package control, none of which are called "SublimeTextXdebug". The repo might be called that way but the package is named differently. It might help users if you link to the specific one on package control you're interfacing with.
- As for the package itself the only thing I'm unsure about is the keybindings. All the F keys are popular, even with modifiers. You might end up masking some other package, or the other way around. That's why we recommend to not ship with bindings enabled and instead tell users how to set their own.
- Hopefully you also tested with some more recent builds of Sublime Text? The one in your readme is ancient.

braver:
👋🏻

maliayas:
Oh, I didn't notice the feedback.

1. You are right, I'll update the readme.
2. Keybindings which this plugin uses are already defined by the original Xdebug plugin. I only override them. So I think that's fair?
3. I'll do once I find time and get back to you.

Thanks for the review.

braver:
> So I think that's fair?

Certainly 👍🏻 

Let me know!

braver:
> get back to you.

let me know, we can re-open then

PR #8827, opened 2023-10-08
PyRock support only for ST4

abhishek72850:

[2]: https://semver.org

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -3094,7 +3094,7 @@
 			"labels": ["python", "import", "python-import", "import-statement", "auto-complete"],
 			"releases": [
 				{
-					"sublime_text": ">3000",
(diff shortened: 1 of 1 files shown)


braver:
If you create a new release that breaks compatibility with older versions, please create a prefixed tag (e.g. "st3-") on the last version to work < 4000. There are plenty of examples in package control of this approach, like: https://github.com/wbond/package_control_channel/blob/42ccd631714860511996a79f20fdac621c3614d5/repository/p.json#L73

abhishek72850:
Actually in my case the whole plugin (all versions) are not compatible with
ST3 , i didn't realized before that the ST3 doesn't have py3.8 support.

braver:
Oh, aha, 👍🏻

PR #8830, opened 2023-10-13
feat: add Unity Snippets package

suiyuex:

My package is `Unity Snipeets`

My package is similar to some other snippets package However it should still be added because it add different snippets with different trigger and will actively maintains.

[2]: https://semver.org

### DIFF
--- a/repository/u.json
+++ b/repository/u.json
@@ -372,6 +372,17 @@
 			]
 		},
 		{
+			"name":"Unity Snippets",
+			"details": "https://github.com/suiyuex/UnitySnippets",
+			"labels": ["snippets"],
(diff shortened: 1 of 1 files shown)


braver:
The main issue with having multiple packages in package control is that users need to start testing all of them to figure out which is any good. Any one of them might go stale eventually. Every package added here started out with the intention of being actively maintained 😉 

Would it be possible to contribute your snippets to one of the existing package, thereby also benefitting existing users? Please try that avenue first. If that fails for some reason, we can always consider adding another package, or even replacing one of the packages.

suiyuex:
Thx

PR #8833, opened 2023-10-14
Set Phpcs to use tags

benmatselby:
The project is moving to tags

My package is already in the registry, I'm just updating it to use tags.

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -1329,7 +1329,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


braver:
🙌🏻

benmatselby:
Hey there @braver

I've noticed that the [package needs a review](https://packagecontrol.io/packages/Phpcs) is that something I need to raise somewhere?

braver:
@wbond needs to have a look at that.

PR #8836, opened 2023-10-15
Added package Word Count

shadax1:

My package is ...

At least I think so, I tried searching for something that counts multiple words but couldn't find anything. If there is already such a thing, my aplogies..

[2]: https://semver.org

### DIFF
--- a/repository/w.json
+++ b/repository/w.json
@@ -730,6 +730,16 @@
 			]
 		},
 		{
+			"name": "Word Count",
+			"details": "https://github.com/shadax1/word_count",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "WordCompletion",
 			"details": "https://github.com/schmave/Sublime-Word-Completion",
 			"releases": [


FichteFoll:
- Your key binding is only active on Windows. Is there any particular reason for this? Otherwise, I'd suggest to provide no key bindings by default. Instead, provide a command for the command palette and suggest the user to add a key binding by themselves in order to reduce key binding conflicts across multiple packages.
- Searching the entire view's contents synchronously after every key press can be costly for larger views. I suggest considering a debouncing mechanism (wait until no modification has been done for a fixed time) and/or switching to the `on_modified_async` hook.
- I also suggest using the [`PhantomSet`](https://www.sublimetext.com/docs/api_reference.html#sublime.PhantomSet) helper class for managing your phantoms and replacing all phantoms at once as that will reduce the chance of flickering (when doing this on the asynchronous thread).
- I'm not going to use it myself, but personally I think I'd prefer an inline phantom after the word I have specified for counting instead of on the next line, i.e. in a block phantom.

braver:
As for similar packages, there is this one: https://packagecontrol.io/packages/WordingStatus

FichteFoll:
I thought so too, but that one counts all words in a buffer while this one counts specific words only. Though I suspect that WordingStatus is also used in the video included in WordCount's README.

shadax1:
@FichteFoll

First of all, thanks for your suggestions. This is my first time writing a ST plugin so I appreciate it. I will look into and work on some of the things you suggested as well.

Regarding some of your comments:
> * Your key binding is only active on Windows. Is there any particular reason for this?

No particular reason honestly, I thought it would make sense to include it in the repo. Based on what you said though, I will definitely remove it and include a command instead.

> * I'm not going to use it myself, but personally I think I'd prefer an inline phantom after the word I have specified for counting instead of on the next line, i.e. in a block phantom.

I thought the same intially but when using `LAYOUT_INLINE`, the counter goes on the left side of the word. Personally, I would prefer (and I think it'd make more sense) the counter to be on the right side. Is there any way to achieve that?
Moreover, is there a way to copy contents of phantoms?

---
Regarding WordingStatus, I'm aware of it but the main goal I had in mind with my own plugin was to count specific words which I don't think WordingStatus does.

> Though I suspect that WordingStatus is also used in the video included in WordCount's README.

Just to clear up any potential confusion, my plugin doesn't rely on WordingStatus or anything. I just happened to have it active in the video.

FichteFoll:
> I thought the same intially but when using `LAYOUT_INLINE`, the counter goes on the left side of the word.

Phantoms will show to the right of the "point" provided. You'll just have to use the end of the word as the point to anchor the phantom to.

> Moreover, is there a way to copy contents of phantoms?

I don't understand this question. What do you want to achieve?

shadax1:
for instance, when I select text that contains phantoms, I want their content to also be selectable and upon doing a copy (`ctrl+c`), I want the selected text + phantoms to be copied into the clipboard

FichteFoll:
Hm, in order to do that you'd probably need to intercept the copy command with an `on_text_command` handler and replace it with a custom command that inserts the (phantom) word counts for each selected word. Or you simply add a phantom with a "copy to clipboard" link that invokes a command to copy the results for all words. The latter would be easier to implement but less flexible.

shadax1:
So I was able to implement the on_modified_async method, phantom set, as well as add commands in the palette and everything works like a charm, thanks again. I also removed the key binding, as suggested.

However, an issue I'm having now is with larger views; when I type into the command palette, the `on_modified_async` method is called upon each key stroke. Is there a way to filter that out?
To explain the reason I'm asking, it's because I added a command that will sort all the words in an ascending/descending order based on their counters. Since I have to open the command palette to initiate that sort, it will enter the `on_modified_async` method after each keystroke which is not noticeable with small views but is quite time consuming with larger ones.

As for my phantom-select-copy endeavor, I decided to go with a different approach by adding a command that will add the words + counters into the clipboard. Might be considered a little cheaty as I was only able to figure out a way to do so through using a global varible.

FichteFoll:
I should probably create a forum post with this, but here you go:

```py
from functools import partial, wraps

import sublime
import sublime_plugin


def _debounced_callback(view, old_change_count, callback):
    if view.is_valid() and view.change_count() == old_change_count:
        callback()


def debounced(delay_in_ms):
    """Forward view-related event hook triggers when the method has not been called for n ms."""

    def decorator(func):
        @wraps(func)
        def wrapper(self, view, *args, **kwargs):
            if not view.is_valid():
                return
            change_count = view.change_count()
            callback = partial(func, self, view, *args, **kwargs)
            sublime.set_timeout_async(partial(_debounced_callback, view, change_count, callback), delay_in_ms)

        return wrapper

    return decorator


class DebouncedListener(sublime_plugin.EventListener):
    @debounced(500)
    def on_modified_async(self, view):
        view.window().status_message("triggered!")
```

FichteFoll:
Also, if you're using a `sublime_plugin.EventListener` (haven't checked), you should check for the result of [`view.element()`](https://www.sublimetext.com/docs/api_reference.html#sublime.View.element) (ST4 only) and exit early if it's not `None`.

shadax1:
With the risk of sounding bothersome, what does the snippet you posted do exactly? Because I don't really understand it.

On the other hand, utilizing [`view.element()`](https://www.sublimetext.com/docs/api_reference.html#sublime.View.element) was exactly what I needed, too bad it's only ST4. I'm fine with marking my plugin as functional for ST4 only `>=4000` because it's currently marked with a `*` in this request.

FichteFoll:
> With the risk of sounding bothersome, what does the snippet you posted do exactly? Because I don't really understand it.

It prevents what you mentioned earlier:

> […] the `on_modified_async` method is called upon each key stroke. Is there a way to filter that out?

The easiest way to find out would probably be to just copy the code into a user plugin and monitor the console as you type stuff and/or modify the selection.

> too bad it's only ST4

Pre-ST4, we used to get around this by checking for `view.settings().get('is_widget')`, which will be `True` for *most* irregular views, i.e. widgets.

shadax1:
Worked on this again today and checking for `view.settings().get('is_widget')` does the trick for me. Aside from counting in very large views using quite a bit of CPU, I don´t really have any other changes to make at the moment. Thanks for all the help

FichteFoll:
After taking a look at the updated recording in the video, I definitely think there should be some separation between the word and the word count, e.g. a colon and a space. Other than that, it looks pretty good. It also doesn't prevent a merge from our side.

There is still no debouncing, as you mentioned, but that's also not a blocker from my side. One last suggestion for your code would be to make use of so-called "early exits" where you check for the exceptional case and exit a function early instead of wrapping all its logic into one or multiple nested conditional if blocks.

PR #8837, opened 2023-10-19
Add UL4 syntax highlighting

doerwalter:

### DIFF
--- a/repository/u.json
+++ b/repository/u.json
@@ -93,7 +93,7 @@
 			"labels": ["language syntax"],
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": ">=4107",
 					"tags": true
 				}
 			]
--- a/repository/u.json
+++ b/repository/u.json
@@ -88,6 +88,17 @@
 			]
 		},
 		{
+			"name": "UL4",
+			"details": "https://github.com/LivingLogic/LivingLogic.SublimeText.UL4",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Ultimate Dark",
 			"details": "https://github.com/rubjo/ultimate-dark",
 			"labels": ["dark", "color scheme"],


braver:
@deathaxe could you maybe have a look at this? You have more experience with extending syntaxes for things like template languages. @doerwalter takes a different approach than Liquid or Twig. I'm not sure of the benefits or downsides of either approach.

braver:
@doerwalter you can ignore those reports from the bot. A fallback for < 3092 is not required. A license is highly recommended however, MIT is a nice and simple one for no-strings-attached open source work.

doerwalter:
OK, I've created a new version 1.0.2 that includes a `LICENSE` file (and also fixes a small bug).

What do I have to do to get my package rechecked?

deathaxe:
Creating templating syntaxes via `push ... with_prototype` often resulted in 25.000 contexts sanity limits being hit, so I'd try to avoid it in general. UL4.sublime-syntax however is simple enough to not cause it. Syntax cache of HTML-UL4 is about 7MB. That's also the size of PHP, so it's ok.

It would also be the only valid implementation in case the syntax really targets ST3.

sublime-syntax `version: 2` format however requires ST4.

So either remove `version: 2` or specify `"sublime_text": ">=4107"` in repository.

The approach taken for Liquid, PHP, ... to `extend` HTML/CSS/JS/XML would probably end in less memory usage, would be more save, allow more control about scopes (e.g.: remove `string` scope in emedded strings), but is more complicated to implement and requires some boilerplate code. It causes some more dependencies with regards to supported ST builds due to base syntaxes having changed over time. Would however be more or less a "copy and paste" task to apply it to UL4.

_Remark: I'd prefer file names to follow common naming scheme of extended syntaxes (`<Base Syntax> (UL4).sublime-syntax`) used by majority of packages, but that's just of stylistic meaning._

deathaxe:
Yeah, the docs still target ST3, mainly with that regard. It's not wrong. It is just known to have some probably unwanted side effects in certain edge cases.

Syntax files specify `version: 2`, which slightly changes behavior of syntax rules in backwared-incompatible ways, especially with regards to treatment of `meta_content_scope` and requires Sublime Text 4. You probably won't realize it due to absence of syntax tests or in best case current syntax is not effected.

The tag is ignored by ST3, but I wouldn't start mixing things up to avoid confusion.

Hence the suggestion to either remove it or restrict installation to ST4 via `"sublime_text": ">=4107"` entry in repository. Package Control will offer the package only for ST4107+ then. 

To be very perfect add `"sublime_text": ">=3092"` as this is the first build, which supports sublime-syntax format at all, but I doubt someone still using such old build.

doerwalter:
OK, I can include the `"sublime_text": ">=4107"`, but I'm not sure into which file that goes. This doesn't look like YAML syntax so probably not into `UL4.sublime-syntax`.

repository/u.json
```diff
@@ -87,6 +87,17 @@
 				}
 			]
 		},
+		{
+			"name": "UL4",
+			"details": "https://github.com/LivingLogic/LivingLogic.SublimeText.UL4",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
```
deathaxe, line 96:
```suggestion
					"sublime_text": ">=4107",
```

braver:
Thanks guys. Learn something new every day! :)

doerwalter:
The latest version of Sublime Text (Build 4189) has started to complain about the embedded UL4 syntaxes like this:

> Error loading syntax file "Packages/User/LivingLogic.SublimeText.UL4/HTML (UL4).sublime-syntax": Apparent recursion within a with_prototype action: 25000 context sanity limit hit

I guess I have to change the way the embedding is implemented. https://www.sublimetext.com/docs/syntax.html#including-other-files still seems to suggest that using `with_prototype` is the canonical approach for embedded syntaxes.

Where can I find documentation about the correct way? Do I have to dig through the PHP syntax to find out how it's done?

deathaxe:
see: https://github.com/sublimehq/sublime_text/issues/6555

PR #8839, opened 2023-10-24
B2 Jam Syntax Highlighting

mccann:
Add B2Jam project

My package is ...

[2]: https://semver.org

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -13,6 +13,17 @@
 			]
 		},
 		{
+			"name": "B2Jam",
+			"details": "https://github.com/mccann/Sublime-B2-Jam-Syntax",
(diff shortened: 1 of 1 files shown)


mccann:
the .gitattributes question is unclear, should I remove the images that are included which are used by the README.md ?

rwols:
You can omit anything that doesn’t contribute to the functioning of the sublime plugin. That probably includes images referenced in the readme.

mccann:
> You can omit anything that doesn’t contribute to the functioning of the sublime plugin. That probably includes images referenced in the readme.

Great, done. Added .gitattributes with export-ignore for 'readme-imgs/'

braver:
Very nice job sir 👍🏻

PR #8840, opened 2023-10-27
Add autodark plugin (for linux)

smac89:

My package is `Auto Dark`

[2]: https://semver.org

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -2281,6 +2281,19 @@
 			]
 		},
 		{
+			"name": "Auto Dark mode for Linux",
+			"details": "https://github.com/smac89/autodark-sublime-plugin",
+			"labels": ["color scheme", "theme", "xdg-desktop-portal"],
+			"releases": [
+				{
+					"sublime_text": ">=4050",
+					"tags": true,
+					"dependencies": ["sublime_lib"],
+					"platforms": ["linux"]
+				}
+			]
+		},
+		{
 			"name": "Auto Encoding for Ruby",
 			"details": "https://github.com/elomarns/auto-encoding-for-ruby",
 			"releases": [
--- a/repository/a.json
+++ b/repository/a.json
@@ -2208,7 +2208,6 @@
 				{
 					"sublime_text": ">=4050",
 					"tags": true,
-					"dependencies": ["sublime_lib"],
 					"platforms": ["linux"]
 				}
 			]


repository/a.json
```diff
@@ -2200,6 +2200,19 @@
...
 		},
+		{
+			"name": "Auto Dark mode for Linux",
+			"details": "https://github.com/smac89/autodark-sublime-plugin",
+			"labels": ["color scheme", "theme", "xdg-desktop-portal"],
+			"releases": [
+				{
+					"sublime_text": ">=4050",
+					"tags": true,
+					"dependencies": ["sublime_lib"],
```
braver, line 2211: Not sure you'd want to lean on that project if you develop for ST4. Having dependencies also prevents you from using python 3.8. If you're mainly using it to interface with settings, I would forego the dependency and use the native api. 

Also, better to remove the dependencies key here. You have a dependencies.json in repo, that will be used to resolve the dependencies. As the documentation states, use or other not, not both, and the json is more convenient.

braver:
Not sure you're just working around a bug in ST here? Have you reported that? 

I can imagine things being not-so-easy on Linux, maybe it's specific to some subset of configurations. As for your implementation here, maybe let's discuss a little. I'm not so sure I would go with a deamon, they have a tendency of dying or otherwise misbehaving. Or you end up spawning multiple. Although I guess it's more elegant than having event listeners. What was your approach here?

There are some print statements in your code. Would be cleaner to have those disabled unless in a debug mode.

Finally, you need to tag a semver version, so v0.1.0 (3 digits) instead of v0.1.

smac89:
@braver I believe it is a bug. I'm not exactly sure what causes it, but another user (also on Linux) had reported something similar affecting them [here](https://github.com/sublimehq/sublime_text/issues/5966).

The solution I implemented is based on [xdg-desktop-portal](https://github.com/flatpak/xdg-desktop-portal), which is the standard on Linux to detect dark/light mode preference. The plugin so far works, and I made sure to test it well in various scenarios. It does behave like an event listener via the [systemd-busctl](https://www.freedesktop.org/software/systemd/man/latest/busctl.html) client.

The only time I've noticed the thread (which is not a daemon thread btw, despite the name) lingering, is when sublime is not shutdown vs closed, i.e. when using the `File` > `Quit` menu. I could work around this by also starting another thread which detects when sublime quits and then stops the dbus client thread as well itself, but I think this something that should be implemented in sublime, not by package maintainers.

A better implementation would be to integrate the dbus support directly into sublime (assuming it doesn't do this already), or integrate a subset of the dbus api by using [libdarkman](https://godocs.io/gitlab.com/WhyNotHugo/darkman/libdarkman), which is written in go.

smac89:
I also found another [issue](https://github.com/sublimehq/sublime_text/issues/5194) which was reported.

I think I'm starting to get a clearer picture as to why dark mode switching isn't fully working on Linux. As stated here: https://github.com/sublimehq/sublime_text/issues/5194#issuecomment-1020709999:

> We grab the setting gtk-theme from `org.gnome.desktop.interface`, apply the theme to a `GtkStyleContext` and grab the selected background color. If the background color has all 3 components < 0.5 we switch to dark mode.

All I can say about this approach is that a solution that will work on "Linux", should not depend on a single desktop i.e. gnome. In the [list of desktops that support xdg-desktop-portal](https://github.com/flatpak/xdg-desktop-portal?tab=readme-ov-file#using-portals), gnome is listed as one of several. Therefore, the right solution would one that is universal.

Furthermore, the nature of the bus interface makes it so that one does not even need to know exactly which backend implements what interface. So the devs can make certain assumptions and maybe use their current solution as a backup plan.

braver:
Ah, that makes sense. At least the bug is reported, and until that's dealt with it would be good to have your package available for users. 

Be sure to tag a new release with your latest changes.

PR #8843, opened 2023-11-01
Swift Next package added.

yaroslavyaroslav:

This is the ST4 sublime-syntax support for Swift. I must clarify that I am not the author of this package. However, I firmly believe in the importance of this release, mainly due to the subpar quality of the currently available alternatives in the package control.

There are a few of them:
- https://github.com/quiqueg/Swift-Sublime-Package — this is a text-mate port that hasn't been updated for nearly a decade.
- https://github.com/colinta/decent-swift-syntax - this is the best one available. It is close to being usable, but it was created in a very simplistic way. For comprehensive language support (i.e. to provide correct entity list in a file), it needs to be redone.

I recently discovered this package quite by chance and was actually stunned by the amount of work that the developer had put into it. Unfortunately, the [main repo](https://github.com/aerobounce/Swift-Next) seems to have been neglected for half a year too. So, I merged all the pending PRs from there, created an organization, and relocated the updated repo there. In this way, I was able to invite the author of this package to the organization, granting him the owner rights (still pending) and all the recognition for this outstanding work.

This PR follows two main objectives:

1. To share this incredible package with people, because believe me, Swift developers are in dire need of it.
2. To appeal to the broader community to enhance it even more.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -5556,17 +5556,6 @@
 			]
 		},
 		{
-			"name": "Swift",
-			"details": "https://github.com/quiqueg/Swift-Sublime-Package",
-			"labels": ["language syntax"],
-			"releases": [
-				{
-					"sublime_text": "*",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "Swift Autocomplete",
 			"details": "https://github.com/Dan2552/SublimeTextSwiftAutocomplete",
 			"labels": ["auto-complete", "completions", "documentation", "swift"],
@@ -5608,6 +5597,18 @@
 			"releases": [
 				{
 					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
+			"name": "Swift Next",
+			"details": "https://github.com/Swift-Next/Swift-Next",
+			"labels": ["language syntax"],
+			"previous_names": ["Swift"],
+			"releases": [
+				{
+					"sublime_text": "4000",
 					"tags": true
 				}
 			]
--- a/repository/s.json
+++ b/repository/s.json
@@ -5556,6 +5556,17 @@
 			]
 		},
 		{
+			"name": "Swift",
+			"details": "https://github.com/Swift-Next/Swift-Next",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Swift Autocomplete",
 			"details": "https://github.com/Dan2552/SublimeTextSwiftAutocomplete",
 			"labels": ["auto-complete", "completions", "documentation", "swift"],
@@ -5597,18 +5608,6 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"tags": true
-				}
-			]
-		},
-		{
-			"name": "Swift Next",
-			"details": "https://github.com/Swift-Next/Swift-Next",
-			"labels": ["language syntax"],
-			"previous_names": ["Swift"],
-			"releases": [
-				{
-					"sublime_text": ">=4000",
 					"tags": true
 				}
 			]
--- a/repository/s.json
+++ b/repository/s.json
@@ -5608,7 +5608,7 @@
 			"previous_names": ["Swift"],
 			"releases": [
 				{
-					"sublime_text": "4000",
+					"sublime_text": ">=4000",
 					"tags": true
 				}
 			]
--- a/repository/s.json
+++ b/repository/s.json
@@ -5556,17 +5556,6 @@
 			]
 		},
 		{
-			"name": "Swift",
-			"details": "https://github.com/Swift-Next/Swift-Next",
-			"labels": ["language syntax"],
-			"releases": [
-				{
-					"sublime_text": ">=4000",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "Swift Autocomplete",
 			"details": "https://github.com/Dan2552/SublimeTextSwiftAutocomplete",
 			"labels": ["auto-complete", "completions", "documentation", "swift"],
@@ -5608,6 +5597,18 @@
 			"releases": [
 				{
 					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
+			"name": "Swift Next",
+			"details": "https://github.com/Swift-Next/Swift-Next",
+			"labels": ["language syntax"],
+			"previous_names": ["Swift"],
+			"releases": [
+				{
+					"sublime_text": ">=4000",
 					"tags": true
 				}
 			]
--- a/repository/s.json
+++ b/repository/s.json
@@ -5556,6 +5556,17 @@
 			]
 		},
 		{
+			"name": "Swift",
+			"details": "https://github.com/quiqueg/Swift-Sublime-Package",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Swift Autocomplete",
 			"details": "https://github.com/Dan2552/SublimeTextSwiftAutocomplete",
 			"labels": ["auto-complete", "completions", "documentation", "swift"],
--- a/repository/s.json
+++ b/repository/s.json
@@ -5605,7 +5605,6 @@
 			"name": "Swift Next",
 			"details": "https://github.com/Swift-Next/Swift-Next",
 			"labels": ["language syntax"],
-			"previous_names": ["Swift"],
 			"releases": [
 				{
 					"sublime_text": ">=4000",


yaroslavyaroslav:
@braver hey, could you please take a look at this one, when you'd have a chance

repository/s.json
```diff
@@ -5612,6 +5601,18 @@
 				}
 			]
 		},
+		{
+			"name": "Swift Next",
```
rchl, line 5605: A non-official drive-by review but I think the syntax package should just be called by the Language name and not include a "branding" words like "next".

yaroslavyaroslav: Apparently this is not the case for the most useful swift package for now, as it had started with the "Swift for fucking sublime" and as I believe was changed to "Decent swift" only by Apple request they released lsp support and promoted such plugin as the default option for ST. 

Although I'm seeing no problem in renaming it by a reviewer request

rchl: At least when the package is just a syntax package then I don't see a reason to have extra words in there. That can age very quickly.

This package also includes completions which I'm not sure is a good idea since those can be quite opinionated (even if just in terms of formatting).

You can wait for the channel owners to review it but those are my initial thoughts.

yaroslavyaroslav: Don't get me wrong, I didn't want it sounds like "keep out of my business". It's just the review rules are both so opaque and inconsistent for me, that I'm trying to avoid any unnecessary moves just in case. 

I see the reason in your point about the naming and snippets issues though, it's just the source of them are opaque for me as well, is it an established yet implicit practice in here, is it your personal recommendation or some another option? Whether those suggestions contradicts with @braver opinion or not?

rchl: Let's wait for braver's review but after observing reviews in this repo for a long time, I'm fairly sure that it's not just my personal opinion.

yaroslavyaroslav: Package has been renamed, it looks kinda wired though, like I'm trying to replace an existing package with the different repo.

rchl: Replacing packages normally goes through a process where the original author is notified and gets 2 weeks or so to respond.

FichteFoll:
Interesting, I didn't know someone was working on an actually reasonable `Swift.sublime-syntax`. I only took a very brief look into it, but it's probably worth exploring the package for inclusion into the default Packages. I'll raise this over the next few days in our Discord chat. There are a few things to resolve because apparently a package named "Swift" already exists on PC and I forgot how exactly PC handles that. We might need to rename the original Swift package on PC in that case.

Outside of that, while I generally agree that taking over unmaintained packages is a good course of action, the replacement does not support ST3, so we'd need to consider this in our decision. 

This is not a complete review. I just saw this in my notifications and took a quick look.

braver:
> the replacement does not support ST3, so we'd need to consider this in our decision

You can always incorporate the final commit of the replaced repository in the new repo (just as a dangling commit or branch) and tag that as the “st3-“ version.

braver:
@FichteFoll assigning to you for now, I’d like to wait for what you come up with before we do anything else. ST deserves a good Swift syntax in the default package.

deathaxe:
A clone of this package is sitting around in my Setup for months, but never got time to review it seriously.

It looks promising on a first glance.

Might probably try to make more use of named contexts.

I'd expect proper syntax tests to be present as a least pre-condition to make it a default packages candidate.

yaroslavyaroslav:
> I'd expect proper syntax tests to be present as a least pre-condition to make it a default packages candidate.

@deathaxe, just to clarify, does this contradicts with @FichteFoll request to move this PR to default packages repo, or is it nothing more than an additional task to got it approved?

yaroslavyaroslav:
@FichteFoll Yep, I believe I'd capable to manage that, I mean both, to move this PR to a default package and to resolve some bug fixes during the review process.

Although I just like to highlight that language markups isn't somehow related to my main scope of proficiency, so I'd be in my turn greatly appreciated for kind-of-dunno guidance during reviewing process.

deathaxe:
No contradiction. Just another task to be completed in order to add it to ST's default packages. Syntax tests are crucial to verify a certain level of quality and to ensure not to break things by future changes.

With regards to this PR, we in general try not to add multiple packages with nearly the same function. So if this package was to be added here, it should probably replace existing "Swift" syntax. Maybe by restricting the existing one to ST3 and use the new one for ST4.

yaroslavyaroslav:
> Syntax tests are crucial to verify a certain level of quality and to ensure not to break things by future changes.

I see a few files in the test dir, is this it or you're talking about something more advanced?

> So if this package was to be added here, it should probably replace existing "Swift" syntax. Maybe by restricting the existing one to ST3 and use the new one for ST4.

Isn't these are mutually exclusive options with moving this PR to a default packages repo? And in the case they are, which one is preferred one?

deathaxe:
Sublime Text supports automated syntax tests to make sure each token receives desired scopes. Those tests are run by CI each time changes are pushed upstream.

Example: https://github.com/sublimehq/Packages/blob/master/Java/tests/syntax_test_java.java

---

> Isn't these are mutually exclusive options ...

Sure. Just expessing expectations that solution, just in case.

yaroslavyaroslav:
I believe I have the strength to undertake the test implementation task, although it might take me a considerable amount of time. At best, I don't see it taking less than a few months. So, from my perspective, I'm completely fine not having this package within Package Control for the duration, as long as I can have it locally on my laptop. Is this acceptable to you @FichteFoll, or should we consider initially taking a shortcut by releasing it here with only a few minor bug fixes as it stands now?

As a final question for @deathaxe, I perused the Packages ci/cd config and downloaded `syntax_tests`. It seems that this binary should run as-is on a Linux system. However, I'm encountering an error when trying to run it on arm (both macOS and the Linux container). Is there a simple way to run it in those environments? It would be quite inconvenient to run the validation pipeline remotely on GitHub otherwise.

deathaxe:
You can run syntax tests directly from within Sublime Text. Just hit ctrl+shift+b to open build system while sublime-syntax file or syntax_test_... file is open and chose "Run Syntax Tests"

FichteFoll:
> So, from my perspective, I'm completely fine not having this package within Package Control for the duration, as long as I can have it locally on my laptop. Is this acceptable to you @FichteFoll, or should we consider initially taking a shortcut by releasing it here with only a few minor bug fixes as it stands now?

I don't really care, but if we intend to only have it on PC temporarily, I also vote against taking over the old Swift package name and go ahead with "Swift Next" or similar, as in the initial proposal. We don't need to disrupt users' workflow more often than necessary.

yaroslavyaroslav:
> I don't really care, but if we intend to only have it on PC temporarily

Not sure that I got this one. Whether it about to drop it from a PC version afterwards it would be covered with tests or to replace such package later on, when it'll be ready. 

But anyway I have to say that while messing with this package with my own swift project I found few more bugs and pretty sure more to come. So it definitely worth to me to revise it carefully while covering it with tests. So I consider the task of including this package in the default repo as a mid term at its best now yet still achievable with my own efforts.

FichteFoll:
I understand. Then I propose the following points of actions.

Short-term:

1. Add this package as "Swift Next" to PC to drive it on some users' machines for some time but without impacting users of the current "Swift" package (even though it is of bad quality).
1. Rename the current "Swift" package on PC to something like "Swift Legacy" in order to free the name for an eventual addition to the default packages. If the package requires changes for the new name, we will suggest a PR to the repo first and otherwise fork it into the Sublime Text organization to make the changes there and redirect the PC entry to it.

Mid-term:

1. Add the contents of this package to the default packages under a permissive (e.g. the current) license. This means that it will be available to all users of newer ST builds and that it will see some general maintenance by the various contributors but also that pull requests to it will be reviewed and merged for the forseeable future. Since none of the current collaborators uses Swift regularly, @yaroslavyaroslav will most likely be the main contributer still.
2. Deprecate the "Swift Next" package on PC and suggest users to use the default Swift package going forward using an update message and by adding it to the README.
1. Deprecate the "Swift Legacy" package and restrict installation of it to builds that don't have the default Swift syntax.

@braver, @deathaxe, what do you think?

braver:
Sounds like a plan 👍🏻

deathaxe:
It's an explicit way, which involves quite some steps and package renames though.

PC4.0 will handle built-in package overrides in sane ways (see: https://github.com/wbond/package_control/commit/3d35ece82fb1b717b7ab8531194d9ff462eb9f6c). They will behave like any other 3rd-party package.

If Swift Next is judged as being quite good already, I wonder if there's a more transparent way for end users to move on. Do we need to retain current Swift package for everyone or would it be good enough to move on as we did with other packages such as Less, Liquid, ...

1. Fork Swift into SublimeText org
2. freeze current state for installation in ST2-3
3. create a branch on top for further development
4. merge in content of "Swift Next" to it (while maintaining snippets or some other useful parts of current "Swift" package).
5. Register the fork as "Swift".

This way:

1. all ST3 users stay on current "Swift" syntax
2. all ST4 users are automatically updated to new syntax and we can use ST4 syntax features when needed.
3. once "Swift" has been added to built-in packages, its installation can be restricted to older builds.
4. we could advice users of newer ST builds to run "Package Control: Revert built-in package".

The downside might be ST4 users not being able to actively choose the old syntax anymore.

braver:
> The downside might be ST4 users not being able to actively choose the old syntax anymore.

Not easily via package control perhaps, but users can always be pointed to a commit of a given repo that can be installed manually. I've handled breaking changes in packages that way before, it's fine.

FichteFoll:
We could also do what @deathaxe proposed. I won't really have time for managing any of this for the next 1.5 weeks at least, but that's probably not much of a problem since you should have the necessary permissions as well.

It's good to know that PC 4.0 will have some code for handling a situation like this and it's not like ST4 users would lose out on many of the improvements done to Swift in the default package since it'd be basically overriding itself. From a user perspective, I presume this to be way with the least effort required to get the latest Swift syntax definition available to you.

yaroslavyaroslav:
@FichteFoll 

Well it took a while, but here we are. I've fixed not all but most of issues provided in your feedback. 

> - [x] As I mentioned, the completions file actually contains several snippets. These must be converted to actual snippet files (because snippets are treated differently in ST regarding their usage and how they can be disabled) or removed entirely, because we don't really add (controversial) snippets in default packages anymore.
> - [x] The `constant.numeirc` scope isn't set on the punctuation of floating point numbers, i.e. `1.0`.
> - [x] Always matching `default` as a keyword seems a bit over-eager. 
> - [ ] Not sure how I feel about matching all non-mutable (i.e. let) ...
> - [x] The number in implicit parameters, i.e. `delegates.invoke { _ = $0.calibrationChanged(Device.Origin(device, event: event), location: location, thresholds: sensor.thresholds) }`, should not be a `keyword.operator` and `constant.numeric` because it's a variable. `variable.parameter` seems appropriate imo.
> - [x] Not sure why these have `variable.other` (blue) (and only sometimes?).
> - [ ] `sensors` in `let foreOut = sensors[Sensor.Location.foreOut.rawValue]` is marked as `variable.function`.
> - [x] Generics are marked as `support.generic`…

As you can notice I've skipped `sensor` and using `variable.other.constant` for let and other one constants issues. This is because as I can see there's a plenty amount of effort required to fix those one. More than that I would have need to put some extra mile here as I still quite lame in all this syntax definition stuff. 

Just to clarify the last one, it doesn't mean that I refuse to fix it completely, but it is mean that I won't do it anytime soon.

in addition to that I reverted the renaming commit in here, so the Package is called **Swift-Next** yet again.

yaroslavyaroslav:
@FichteFoll @braver 
It's not that I'm rushing you in any way, just checking if there's anything left on me here or should I just relax and keep waiting when the magic happens?

repository/s.json
```diff
@@ -5555,17 +5555,6 @@
 				}
 			]
 		},
-		{
-			"name": "Swift",
```
braver, line 5559: Please revert this removal of the Swift package, since we now intend to leave it where it is and just add your package as a new and separate one.

braver (changes requested):

repository/s.json
yaroslavyaroslav: Done

braver: @yaroslavyaroslav you removed the previous_names, but the removal of Swift from the index is still in the PR.

yaroslavyaroslav: @braver yep, my bad, now everything should be fine

braver (approved):

braver:
Alright, let's get this package out there!

@deathaxe @FichteFoll we can do an iteration later, where we do basically what @deathaxe [suggested](https://github.com/wbond/package_control_channel/pull/8843#issuecomment-1817817372). For now, let this new package prove itself in the wild for a bit, there is no rush.

FichteFoll:
> It's not that I'm rushing you in any way, just checking if there's anything left on me here or should I just relax and keep waiting when the magic happens?

I saw your reply but was just busy with other stuff over the past weeks. Reviewing a package (and especially a huge syntax definition) takes some time, even when you're just iterating over stuff you had previously seen, and I want to give it the attention it deserves instead of doing a rushed analysis, if you could even call it that.

That said, we were in agreement over how to handle this package going forward (get it onto PC first, think about making it a default package later) and the first step was just completed, so let's see how the package fares in the near future. I personally do not have any interest in working with Swift any time soon, so I don't have much interest in a hands-on review, but I'm confident we'll eventually manage something when it comes to aligning it more with the default packages eventually, which is an ongoing process even for the syntaxes that we already have.

---

> So I decided to made it nested (i.e. `constant.numeric.value.swift punctuation.separator.decimal.swift`) which still breaks syntax highlight though. Therefore fell free to force drop `punctuation` definition it if you think it'd be better,

No, it's quite alright. This is what you should be doing.

As for the rest, they are too minor to discuss them in detail here. I'll just add a refence to this package in the latest Swift issue/PR on the default packages and move on for a while until we get back to it.

PR #8844, opened 2023-11-04
Remove dead packages

deathaxe:
All removed packages' repositories error out with 404.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -1084,19 +1084,6 @@
 			]
 		},
 		{
-			"name": "Angular CLI",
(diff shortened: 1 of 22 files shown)


FichteFoll (approved):
I have written a shell one-liner to verify this and they indeed all returned 404:
```sh
curl -s https://patch-diff.githubusercontent.com/raw/wbond/package_control_channel/pull/8844.diff | rg -o '^-.*"details": "(.*?)".*' -r '$1' | xargs curl -sSLI --write-out '%{stderr}%{http_code} %{url}\n' >/dev/null
```

PR #8846, opened 2023-11-07
Added Quote With Marker Package.

dennykorsukewitz:
- [x] My package does not add key bindings. ** 

My package is **[Quote With Marker](https://github.com/dennykorsukewitz/Sublime-QuoteWithMarker)** 

quotewithmarker

[2]: https://semver.org

### DIFF
--- a/repository/q.json
+++ b/repository/q.json
@@ -447,7 +447,6 @@
 		{
 			"name": "Quote With Marker",
 			"author": "Denny Korsukéwitz",
-			"description": "This package quotes the selected area and adds a custom marker to it.",
 			"details": "https://github.com/dennykorsukewitz/Sublime-QuoteWithMarker",
 			"labels": ["quote", "workflow", "codemarker", "marker", "markers", "comment", "commentline", "patch", "patcher", "diff", "compare", "helper"],
 			"releases": [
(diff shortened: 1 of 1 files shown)


dennykorsukewitz:
I already used tags: https://github.com/dennykorsukewitz/Sublime-QuoteWithMarker/tags

repository/q.json
```diff
@@ -444,6 +444,19 @@
 				}
 			]
 		},
+		{
+			"name": "Quote With Marker",
+			"author": "Denny Korsukéwitz",
+			"description": "This package quotes the selected area and adds a custom marker to it.",
```
braver, line 450: It's easier if you just put a description on your GitHub repo. Package control will know to find and use that.

braver:
> QuoteWithMarker is a Sublime Text 2/3 package 

🤔 That's a bit of a weird audience since ST4 has been out for years already. 

> My package does add key bindings. ** But so hard to trigger.

Haha, yeah, that takes some gymnastics. Good reason to not ship with any at all and let users set something easier to do and to remember 😉 

> There are no packages like it in Package Control

Indeed, I don't think I ever saw something like it. Nice 👍🏻

repository/q.json
dennykorsukewitz: Ah, thanks for that hint.
Fixed with: https://github.com/wbond/package_control_channel/pull/8846/commits/c8946b2eeebe69494f09f57f191eeeff9c561b84

PR #8847, opened 2023-11-10
Add AutoProjects

deathaxe:

My package is a tiny helper to automatically open folders as projects, to mimic behavior of VS Code's `.vscode` directory.

Folder2Project seems to have somewhat a comparable scope, but only opens existing projects for a file already open in ST.

My package however targets workflows where users open folders in ST via context menu but actually want them to be treated as ST project with all its advantages.

[2]: https://semver.org

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -2514,6 +2514,17 @@
 			]
 		},
 		{
+			"name": "AutoProjects",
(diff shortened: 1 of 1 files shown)


braver:
Amazing!

PR #8848, opened 2023-11-13
Added GitHubFileFetcher Package.

dennykorsukewitz:

My package is [GitHubFileFetcher](https://github.com/dennykorsukewitz/Sublime-GitHubFileFetcher).
This package searches and fetches files from GitHub.

githubfilefetcher

[2]: https://semver.org

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -701,6 +701,19 @@
 			]
 		},
 		{
+			"name": "GitHub File Fetcher",
+			"author": "Denny Korsukéwitz",
+			"description": "This package searches and fetches files from GitHub.",
+			"details": "https://github.com/dennykorsukewitz/Sublime-GitHubFileFetcher",
+			"labels": ["workflow", "helper", "github", "file", "files", "fetch", "fetcher", "grab", "download", "get"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


braver:
Is the list of possible repositories a "white list", ie. only repos in that list can be searched?

dennykorsukewitz:
> In the main menu file, for any existing entries like "Preferences", "Package Settings", etc. please only provide the id. Otherwise you're effectively overriding the menu entry name/caption and mnemonic.
> 
> The Default.sublime-keymap file doesn't seem to exist in your repo?

Ah good to know.
I fixed that all with https://github.com/dennykorsukewitz/Sublime-GitHubFileFetcher/commit/622d267ff995aedd6e8dee7fa03aa091e618cc2b

dennykorsukewitz:
> Is the list of possible repositories a "white list", ie. only repos in that list can be searched?

It is rather a list of already saved repos (from already searched repos). 

You are looking for a repo. Either by the owner or directly by name.

Then you can get the desired file in the desired branch.

Finally, you are asked whether you want to save this repo. 
This list of saved repos is then displayed each time.

This allows you to use your frequently used repos quickly.

braver:
Alright, let's ship this, but don't forget to tag a new release with your latest changes. Package Control only sees the tags on your repo, general commits won't deploy to end users.

PR #8849, opened 2023-11-14
Added ToggleSidebarFocus plugin

educbraga:
- [x] I have have read [the docs](https://packagecontrol.io/docs/submitting_a_package)

My package is [ToggleSidebarFocus](https://github.com/educbraga/ToggleSidebarFocus)
A plugin for Sublime Text that allows you to open, focus, and close the sidebar using the same keyboard shortcut.

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -2717,6 +2717,16 @@
 			]
 		},
 		{
+			"name": "Toggle Sidebar Focus",
+			"details": "https://github.com/educbraga/ToggleSidebarFocus",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},	
+		{
 			"name": "Toggle Symbol to String",
 			"details": "https://github.com/zoomix/SublimeToggleSymbol",
 			"releases": [


braver:
Maybe it’s just on MacOS, but ctrl+0 already runs the focus_sidebar command, which opens the sidebar if it’s closed. I don’t understand what your package adds to what’s already built into ST?

educbraga:
> Maybe it’s just on MacOS, but ctrl+0 already runs the focus_sidebar command, which opens the sidebar if it’s closed. I don’t understand what your package adds to what’s already built into ST?

Hey, @braver 👋🏻
Absolutely! On macOS, the standard process to open and focus on the sidebar in Sublime Text involves a two-step keyboard shortcut sequence (cmd + k, followed by cmd + b) and then ctrl + 0. To close the sidebar, the same cmd + k cmd + b sequence is repeated. This plugin simplifies this by allowing the user to define a single command that can both open the sidebar with focus and close it, streamlining the workflow significantly to those who use sublime with the sidebar closed.

PR #8850, opened 2023-11-19
Add a syntax for FASTA format

SilentGene:

My package is a syntax for FASTA format

[2]: https://semver.org


braver:
- you need to tag a semver style version with 3 digits (1.0.0 instead of 1.0)
- please make sure all your scope have the `.fasta` suffix

SilentGene (commented):
Ok. Thanks!

PR #8853, opened 2023-11-30
Heta language support

metelkin:

My package is the highlighter for the Heta modeling language, see https://hetalang.github.io/#/ 

[2]: https://semver.org

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -411,6 +411,16 @@
 			]
 		},
 		{
+			"name": "Heta Highlight",
+			"details": "https://github.com/hetalang/heta-highlight-sublime",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Hex to HSL Color Converter",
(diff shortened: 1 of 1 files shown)


metelkin:
@packagecontrol-bot

braver:
Unless you're targeting multiple editors with that tmLanguage file I highly recommend switching to the new yaml based sublime-syntax format. That's a lot easier to maintain than the ancient text-mate grammar. 
Please also have a look at the [scope naming guide](https://www.sublimetext.com/docs/scope_naming.html), especially `space.heta` looks like it should be wrapper in a higher level scope. `invalid.illegal` is also missing the `.heta` suffix.

Please add `"labels": ["language syntax"]` to your entry here to indicate the type of package.

metelkin:
Thank you for your feedback.
Yes, we are using tmLanguage to support multiple editors. I guess it would be easier to proceed with the old approach.
I fixed the scopes you mentioned in projects master. I guess we will revise all of them in future releases.

metelkin:
I've found the problem. This is because of exception in conversion from tmLanguage. Hope I'll fix it in the new release 0.1.9

PR #8861, opened 2024-01-03
Add EasyCommit

educbraga:

My package simplifies the process of making Git commits. After pressing a single key binding the plugin performs a series of Git operations: fetch, pull, and add all or add the current file. After staging changes, it prompts for a commit message. It commits with the provided message and then pushes the changes to the remote repository on the current branch. 

In case of any issues during the Git operations it stops and displays the error messages. Allows cancelling the commit process by pressing ESC. It provides ongoing feedback in the Sublime Text status bar, keeping the user informed at every step.

[2]: https://semver.org

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -147,6 +147,7 @@
 		{
 			"name": "EasyCommit",
 			"details": "https://github.com/educbraga/EasyCommit",
+					"labels": ["git", "commit", "push", "easy"],
 			"releases": [
 				{
 					"sublime_text": "*",
--- a/repository/e.json
+++ b/repository/e.json
@@ -145,6 +145,16 @@
 			]
 		},
 		{
+			"name": "EasyCommit",
+			"details": "https://github.com/educbraga/EasyCommit",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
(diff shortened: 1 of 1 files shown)


braver:
Thanks for your submission and sorry for taking a while to review it. 

Looks really nice and useful 👍🏻 Some remarks:

- The commands are not exposed via the command palette. Especially for new users that makes them hard to find (it's only after seeing the package does nothing, then finding the README, that you figure out it needs more setup). Key bindings are nice if you can remember them, but they should never be the only way to trigger commands.
- You have quite a bit of code duplication going on here. Getting that cleaned up will help you maintain it and others contribute to it over time.

repository/e.json
```diff
@@ -144,6 +144,16 @@
 				}
 			]
 		},
+		{
+			"name": "EasyCommit",
+			"details": "https://github.com/educbraga/EasyCommit",
```
braver, line 149: Might be useful to add some labels for findability, eg. "git".

educbraga:
> Please respond to the feedback to continue the review.

Thank you for the reminder @braver!
I made some changes based on your suggestions.

If there is any other one, just let me know! 😄✌🏻

PR #8862, opened 2024-01-03
Add TreeSitter package

kylebebak:
- [x] I have tagged a release with a [semver][2] version number. **[See 1.0.0 here](https://github.com/sublime-treesitter/TreeSitter/releases/tag/1.0.0)**
- [ ] I use [.gitattributes][3] to exclude files from the package: images, test files, sublime-project/workspace. **There are no such files in this repo**

This package provides other Sublime Text packages with a performant and flexible interface to [Tree-sitter](https://tree-sitter.github.io/tree-sitter/).

It also provides commands to:

- Get a Tree-sitter `Tree` by its buffer id, or get trees for all tracked buffers
- Subscribe to tree changes in any buffer in real time using `sublime_plugin.EventListener`
- Get a tree from a string of code
- Get a node from a point or selection
- Select ancestor, descendant, sibling, or "cousin" nodes
- Query a tree, walk a tree
- Goto symbols returned by tree queries

There are no packages like it in Package Control. See more in the package README: https://github.com/sublime-treesitter/treesitter

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -3296,6 +3296,24 @@
 			]
 		},
 		{
+			"name": "TreeSitter",
+			"details": "https://github.com/sublime-treesitter/TreeSitter",
+			"labels": [
+				"tree-sitter",
+				"navigation",
+				"selection",
+				"symbols",
+				"parsers"
+			],
+			"author": "kylebebak",
+			"releases": [
+				{
+					"sublime_text": ">=4107",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "TreeTop",
 			"details": "https://github.com/royvandewater/Sublime2-TreeTop",
 			"releases": [


kylebebak:
Note that `.no-sublime-package` is defined to ensure dependencies for package can be installed in subdirectories under the main package directory.

These deps are as follows:

- The [Python bindings to Tree-sitter](https://github.com/tree-sitter/py-tree-sitter) (a pip package)
- The various Tree-sitter grammar repos, [e.g. the grammar for Python](https://github.com/tree-sitter/tree-sitter-python)
- The `.so` (shared object / dynamically linked library) files for the parsers, which are compiled by the user's Python from the code in the grammar repos

braver:
Thanks for your submission. Looks like a well thought out package. Of course, as you noticed, this is a bit on the edge of what Sublime Text and package control are set up to manage or what we consider normal for a package. But that can't be helped and seems to be handled as well as it could.

I only have a minor remark, and that is that the settings should be reachable via the main menu (in addition to the command palette).

kaste:
I think it should be added at least for visibility.  The installer notes should be crystal clear though what steps are necessary to make this work.  (It actually works after you install everything.)   If this is not added how should people find out about it.  (I actually searched for a plugin that bridges to TreeSitter as it is very interesting.)  This (the visibility) is much needed as the plugin itself doesn't even have much functionality; it is in this early phase basically the bridge/the adapter plus some demo commands.  

On the argument of dependencies, GitSavvy doesn't work without the external git.  (Although git is much easier to setup.)  SublimeLinter doesn't ship without other plugins (the adapters) *and* external executables (the linters).  

The plugin needs a tiny step more (at least for Windows) to be not too scary imo.

braver:
My thinking aligns with @kaste here. It has its limitations and challenges, but it's well made and not malicious. If we add it to package control, it's easy to find and "easy enough" to get it to work. It's not like users will flock to it immediately, it doesn't readily provide any functionality by itself. The community can then start figuring out how to make use of the package. Packages like SublimeLinter and LSP have had similar starts where everyone just needs time to figure it out. 

The only reason I would hold back for a while is to see if the author(s) of the package are fully willing to join this discussion and to put in the work to make the package a success. This obviously isn't a "rainy Sunday afternoon" package, but that check is one of the primary reasons for the review process.

deathaxe:
I gave TreeSitter package another try today, but even with explicitly setting python paths settings, it fails to compile treesitters languages on Windows 11.

As the target of each language is an `.so` file, I even wonder whether it's compatible with Windows.

deathaxe:
`tree_sitter` seems to expect a compiler to be present in order to compile languages. see: https://github.com/sublime-treesitter/TreeSitter/issues/3

Unless syntaxes are not shipped in a pre-compiled state, I wouldn't call this very end-user friendly. I wouldn't want to install Visual Studio build tools, just to compile syntaxes.

In fact I have them installed, but they don't seem to be found out of the box.

kaste:
That's basically the reason that we need a full (external) python3.8 here because we need to compile.  That typically works with the toolchain to compile python itself (or extensions or wheels for it).  (That is to say the compiler used to compile python and the one to build extensions must match if I'm not mistaken.)  That would be Visual Studio 2017 Build Tools https://aka.ms/vs/15/release/vs_buildtools.exe  (You can probably also use `winget`?)  

You could also pre-compile (but not at this stage of the project) or use `pip install tree_sitter_languages` which installs just *all* languages plus tree_sitter. This usually comes at the price that the compiled languages might be behind "master" of the published language repo (as usual). 

This is still promising.  Look at how I just cycle through function arguments.

Unbenanntes Video – Mit Clipchamp erstellt

Or select multi-line statement and if-blocks:

Unbenanntes Video – Mit Clipchamp erstellt (1)

deathaxe:
A full python for windows does not automatically include visual studio build tools. Thesa are another couple of gigs to install, if we include Windows SDK and other dependencies, needed to compile package. I find it tedious to expect end-users to install those in order to use a package. Actually I only have VS2022 build tools installed for other reasons, but normally wouldn't want to do so as an ordinary user.

Ok, this package is not harmful, but also not very useful with all those external dependencies. Another building block for those who complain about ST being tedious and complicated to use with lots of hazzle.

kaste:
You actually added tree_sitter to Package Control dependency channel (? how is that named)
(https://github.com/sublime-treesitter/TreeSitter/issues/2).  That itself is of limited use as the 
user still has to compile or somehow "get" language files.

I pointed above to the python package `tree_sitter_languages` which a user
also can _pip install_.  If you could add this package to PC then this plugin here
could become plug'n'play as `tree_sitter_languages` installs tree_sitter and *compiled* 
languages.  

Other than that, I think you're too negative here.  The README of the plugin doesn't even mention
any useful "end-user" commands.  It does advertise that you get a tree from source
and can implement an ordinary `EventListener` to subscribe to tree changes.  So the "end-user" 
of the plugin is currently not necessarily what you imagine.  It is also the typical "scary" Windows 
experience.  (But I'm also on Windows.)

--- 

For future reference here is what I did/used on Windows to install the dependencies.

1. Install rye https://rye-up.com/guide/installation/ as it is a simple tool to maintain multiple python versions
2. Create a directory somewhere and in it execute

```
rye pin 3.8
rye init
rye add pip
rye sync
rye shell
where python
where pip
```
3. Fill in the location of python and pip in the plugin's settings
4. Install the Microsoft compiler https://aka.ms/vs/15/release/vs_buildtools.exe
5. Restart Sublime Text

deathaxe:
> You actually added tree_sitter to Package Control dependency channel (? how is that named)

All libraries are just normal python packages and thus they are named alike. Actually Package Control just downloads the wheel from pypi.org and extracts it to $data/Lib/python38.

So the name is just `tree_sitter` and is meant to be a replacement for the pip call invoked by the package to install it.

> ... add this package to PC then this plugin here could become plug'n'play as `tree_sitter_languages` installs tree_sitter and compiled languages.

Sorry, oversaw this hint. That's the solution I was looking for. 

It would probably require some tweaks in TreeSitter to make use of this library as it currently looksup compiled languages from its own build path and expects `.so` files, while `tree_sitter_languages` ships "languages.dll" on Windows.

kylebebak:
Hey @deathaxe @braver @kaste 

Thank you for reviewing this. Based on your suggestions, I'm going to do the following:

- Remove the `pip_path` setting, and install the `tree_sitter` [package as an ST dependency](https://github.com/sublime-treesitter/TreeSitter/issues/2) (thank you deathaxe)
- Improve the README by making the dependency on an external Python more clear
    - For now I think I'd like to keep the dependency on the external Python
        - [I wrote about this more in this issue](https://github.com/sublime-treesitter/TreeSitter/issues/2#issuecomment-1894122724) that deathaxe opened
        - If we want to discuss this more, maybe we should use that issue, which is in the STS repo
    - Add some instructions on how to install a Python 3.8 that can compile Tree-sitter languages
        - [pyenv](https://github.com/pyenv/pyenv) for Linux/MacOS
        - [rye](https://github.com/mitsuhiko/rye) (or maybe [pyenv-win](https://github.com/pyenv-win/pyenv-win)?) for Windows
        - I don't develop on Windows, maybe someone who's done this, like kaste, can open a PR adding these instructions to the README
- Improve the README by documenting some of the built-in STS commands
    - `tree_sitter_select_ancestor/sibling/cousins/descendant`
        - I'm guessing @kaste you're using these in the GIF you pasted above
    - `tree_sitter_goto_symbol`
        - Explain how `queries` files work
    - Debug commands: `tree_sitter_print_tree`, `tree_sitter_show_node_under_selection`
- Make package settings reachable via the main menu (in addition to the command palette)

Also, excuse me for taking a while to respond to issues, and make code changes. I don't have as much time as I'd like to work on this (job + newborn daughter), but I would like to do what's needed to make the package useful to other ST users.

If anyone wants to develop a package that depends on this one, and have it live under the [sublime-treesitter](https://github.com/sublime-treesitter) org, I can add you to this org. I was thinking that packages under this org could use Tree-sitter to do useful things, e.g. build a live-updating "symbol map" for the text in a buffer. Likewise let me know if you want to collaborate on the core STS package.

braver:
@kylebebak Thanks for responding. It's ok if a an editor plugin is not your life's priority, it's probably even a good sign 😉 But it's good to see your response and intent for the package. As I mentioned, if effort is being made to improve the out of the box experience of the package, and it otherwise plays nice within the eco system, I have no problem with adding it. So, for now, just get back to us once you've addressed some of those issues and we can move this PR forward.

kaste:
Shameless plug:  https://github.com/kaste/TreeSitter-calls-and-callers as a first try.  It is not completely trivial to use tree_sitter. Of course the keywords/names for "things" are not completely standardized between the languages (yet?) (like the syntaxes here for Sublime);  it also moves very fast;  and it is not as performant as I would have hoped.  Part of the plugin I advertise here I had for Python only, thus idiosyncratic, built on scopes and such things.  But it by a magnitude outperforms/-ed the TreeSitter implementation.   FWIW ☘️

kylebebak:
Hey all, I made the changes discussed above, pushed them to https://github.com/sublime-treesitter/TreeSitter, and cut a new release: https://github.com/sublime-treesitter/TreeSitter/releases/tag/1.1.0

- Install `tree_sitter` from `dependencies.json` instead of using external `pip`
- Make `python_path` optional, using binaries bundled with `tree_sitter_languages` by default
  - Thanks @deathaxe for making both `tree_sitter` and `tree_sitter_languages` installable as dependencies
- Package settings file reachable via main menu
- Add reload command by @kaste in https://github.com/sublime-treesitter/TreeSitter/pull/5

deathaxe:
LGTM!

Added two PRs to fix some minor stylistic issues:

- https://github.com/sublime-treesitter/TreeSitter/pull/7
- https://github.com/sublime-treesitter/TreeSitter/pull/8

As a general note: It is not required to ship a package with `.no-sublime-package`, just because it want's to download extra files.

Content from _Installed Packages/<package>.sublime-package_ and _Packages/<package>_ is merged and appears like a single directory in ST. It would theoretically be possible to ship TreeSitter as an ordinary TreeSitter.sublime-package and still download and compile extra languages into _Packages/TreeSitter_. The only requirement would be to ensure target directory is created on demand.

kylebebak:
Hey @deathaxe , I merged both of your PRs, and left a comment regarding this issue here: https://github.com/sublime-treesitter/TreeSitter/issues/9

Here's the new TreeSitter release: https://github.com/sublime-treesitter/TreeSitter/releases/tag/1.2.0

deathaxe (approved):

braver:
@kylebebak 2 remaining points of review I'd like to cover before we move forward with merging this:

- As @deathaxe explains, there does not seem to be a reason to ship with `.no-sublime-package`. Typically the main reason for it is to run scripts outside sublime's python environment. I won't try to grasp everything your package does, but would prefer to ship without the `.no-sublime-package` if at all possible.
- If you open your settings via the menu, you get a ["python_path" setting](https://github.com/sublime-treesitter/TreeSitter/blob/fa9266c9f0ec50652eca099be048146345eaa18c/Main.sublime-menu#L20C38-L20C49) by default that's not covered in your default settings file, and is also missing if you open the settings via the command palette... I think this is just something that got away during a refactor.

kylebebak:
@braver 

> If you open your settings via the menu, you get a ["python_path" setting](https://github.com/sublime-treesitter/TreeSitter/blob/fa9266c9f0ec50652eca099be048146345eaa18c/Main.sublime-menu#L20C38-L20C49) by default that's not covered in your default settings file, and is also missing if you open the settings via the command palette... I think this is just something that got away during a refactor.

Thanks for catching this, this indeed just got lost in a refactor. [I just cut a release that fixes this](https://github.com/sublime-treesitter/TreeSitter/releases/tag/1.3.1).

---

I included `.no-sublime-package` to ensure the `build` directory always exists. For users using the `python_path` setting to manage language binaries themselves, the `build` directory is where language repos are downloaded and language binaries are generated.

Even after reading https://packagecontrol.io/docs/submitting_a_package#Step_5, I still don't understand how `.no-sublime-package` works well enough to know whether it's necessary. It sounded to me like it was meant for these types of use cases:

> If including executables or shared libraries, add a .no-sublime-package file

I saw what deathaxe wrote in an earlier comment:

> Content from Installed Packages/.sublime-package and Packages/ is merged and appears like a single directory in ST. It would theoretically be possible to ship TreeSitter as an ordinary TreeSitter.sublime-package and still download and compile extra languages into Packages/TreeSitter. The only requirement would be to ensure target directory is created on demand.

Maybe I'd need to create the `build` directory on demand? Right now, I've got a `.gitignore` file in the build directory to ensure the directory exists: https://github.com/sublime-treesitter/TreeSitter/blob/main/build/.gitignore. deathaxe added a `.gitattributes` file with `.gitignore              export-ignore`. Would that prevent this `.gitignore` file and the empty `build` directory from being included in the package?

kylebebak:
And while we're on the subject of things I don't know how to do...

Until `TreeSitter` is in package control, I don't know how to test the "installation via PC". I just know things work in local development. Basically, I don't know how to test the effect of including/excluding `.no-sublime-package` without installing the package from PC. But if you think we can just get rid of it, I'll do that. Should I just add some code that creates the `build` directory dynamically if it doesn't exist?

kaste:
TBH to have or not have a `.no-sublime-package` in the context of this plugin is just micro-management.  I hate that.  There are clear cases of where you *don't* need the extracted package and then you should of course say so.  But this case is not a clear case as you actually need some place and code to manage a "build" directory.  You can also put stuff in the `cache_path()`, doesn't make stuff easier as well.  🤔

braver:
@kylebebak basically follow this guide: https://packagecontrol.io/docs/creating_package_files

- run the "Create package file" command
- move the generated sublime-package file into the "Installed Packages" directory
- move your original package directory out of the "Packages" directory (just anywhere outside the ST directory structure)
- restart ST
- check if everything still works as expected

I appreciate this creates a bit of overhead for this package, but it's useful to have a good handle on how shipping your package in this way affects your package. Everything is always easier if you ship the package just as you develop it, but there are downsides to that as well (which is of course why the sublime-package concept was introduced in the first place). 

> Would that prevent this .gitignore file and the empty build directory from being included in the package?

Shipping as a package basically creates a clean split between the code you deliver to the end user's machine, and any files generated during runtime. The build directory with the .gitignore file in it now lives inside the sublime-package (in "Installed Packages"), and is no longer a writable location. So if you need to output to a build directory, create it inside "Packages/TreeSitter".

kaste:
Just putting the build dir somewhere is the actual easy part here, and the cache dir or system wide temporary dir are better choices than the packages dir probably.  But the plugin ships a file/script which gets executed by the external python.  `str(PROJECT_ROOT / "src" / "build.py")`. And that's the ugly part because you must copy that around at Sublime startup.  You cannot execute that out of the installed zip file.  It's doable of course.

kylebebak:
Hey @braver @kaste 

[I tried removing](https://github.com/sublime-treesitter/TreeSitter/commit/be5aa751464094c037b289168c3de78565e37da3) `.no-sublime-package` and installing everything under `sublime.cache_path() / "TreeSitter"`. I like installing language repos and binaries in this path more than in the package dir. However, running the plugin from within `TreeSitter.sublime-package` created a few problems:

- Can't read files in `queries` without unzipping package
- Can't read contents of `build.py`, e.g. to copy it into the cache dir so it can be executed by the system Python, without unzipping package

I could do something like this, using `zipfile` from the stdlib:

```py
from zipfile import ZipFile

with ZipFile("TreeSitter.sublime-package") as zf:
    with zf.open("src/build.py") as f:
        # Read build.py and write it to the build dir
    # Do the same for all files in the queries dir
```

But this feels a little weird, especially with the `queries` files. If I copy them over to the cache path every time ST starts up, do I clobber existing files, or not? Both ways have tradeoffs. It seems simpler to just keep them in the package dir.

So, [I pushed another commit](https://github.com/sublime-treesitter/TreeSitter/commit/04166871f87433243a13518b912ef643b9231ec0) that reverted some of the changes. It keeps `.no-sublime-package`, but installs language repos and binaries in `cache_path()`.

What do you think? I'd rather not move this around any more unless you feel strongly about it.

deathaxe:
We shouldn't overcomplicate things. So it's probably more safe to keep `.no-sublime-packages` then.

braver:
Alright, that covers it here, let’s roll.

kylebebak:
Thank you for all the feedback, the package has improved a lot because of it =)

The next issue I want to tackle is this one: https://github.com/sublime-treesitter/TreeSitter/issues/9

FichteFoll:
> But this feels a little weird, especially with the `queries` files. If I copy them over to the cache path every time ST starts up, do I clobber existing files or not? Both ways have tradeoffs. It seems simpler to just keep them in the package dir.

For convenience (and your information), sublime_lib actually provides a function for that: [`ResourcePath.copytree`](https://sublimetext.github.io/sublime_lib/#sublime_lib.ResourcePath.copytree).

PR #8863, opened 2024-01-03
Add Yollama

mwwwa:

My package is ...

This package adds a command to query an Ollama instance (local LLM inference runner) for assistance or chit chat.

[2]: https://semver.org

### DIFF
--- a/repository/y.json
+++ b/repository/y.json
@@ -220,6 +220,20 @@
 			]
 		},
 		{
+			"name": "Yollama",
+			"details": "https://gitlab.com/mwwaa/yollama",
+			"labels": ["ai", "assistant"],
+			"author": "mwa",
+			"homepage": "https://maxime.walzberg.fr/articles/yollama",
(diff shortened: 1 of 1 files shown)


braver:
Thanks for your submission. This looks relatively straight forward. The only thing to add here is that since your package has settings, they should be exposed via the menu and command palette (otherwise users won't know about the settings and won't be able to easily edit them).

braver:
Excellent, thanks. Note that for the "Package Settings" you should also just provide the id, instead of effectively resetting the caption and mnemonic.

FichteFoll:
> Note that for the "Package Settings" you should also just provide the id, instead of effectively resetting the caption and mnemonic.

Actually, the name and mnemonic *should* be provided for this option since it't not part of the Default menu and it's not guaranteed that some other package has already defined the same menu item, e.g. when this package is the only one installed (besides PC).

PR #8865, opened 2024-01-11
Add the "evscript" package

ISSOtm:
- [x] Any commands are available via the command palette. (None.)
- [x] I use [.gitattributes][3] to exclude files from the package: images, test files, sublime-project/workspace. (There are none.)

My package is a new syntax definition, with no accessories for now. (I'll add some goodies in a later release, e.g. "shell config"; but I want to move on for a while.)

There are no packages like it in Package Control. (The language is really niche.)

- [x] For bonus points also considered how the review guidelines apply to your package:
https://github.com/wbond/package_control_channel/wiki#reviewing-a-package-addition

[2]: https://semver.org

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -1440,6 +1440,17 @@
 			]
 		},
 		{
+			"name": "evscript",
+			"details": "https://github.com/ISSOtm/sublime-evscript",
+			"labels": ["language syntax", "game boy", "RGBDS"],
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Exact Quick Find",
 			"details": "https://github.com/aafulei/sublime-exact-quick-find",
 			"author": "Aaron Fu Lei",


repository/e.json
```diff
@@ -1439,6 +1439,17 @@
 				}
 			]
 		},
+		{
+			"name": "evscript",
+			"details": "https://github.com/ISSOtm/sublime-evscript",
+			"labels": ["language syntax", "game boy", "RGBDS"],
+			"releases": [
+				{
+					"sublime_text": ">=3092",
```
deathaxe, line 1448: Syntax definition specifies `version: 2`, which requires at least Sublime Text 4.

The first stable release is:

```suggestion
					"sublime_text": ">=4107",
```

ISSOtm: Oh, I wasn't aware of that. Since those are simple changes, I'd like to prioritise supporting earlier ST versions. Are there other caveats I should be aware of?

(And, for that matter, are there automated checks for this? I haven't found any.)

deathaxe: v2 applies meta scopes a bit different in detail.

Your syntax seems not to be affected by that change, so removing `version: 2` and reverting to `pop: true` should be all.

braver: @ISSOtm indeed, supporting versions of ST much older than 4107 sounds like a wasted effort. Please apply that suggestion so we can continue the review.

ISSOtm: Whoops, sorry, this fell off my radar... I pushed 1.0.1, which fixes deathaxe's review.

My apologies for the delay!

ISSOtm:
Thank you!

PR #8873, opened 2024-02-18
Add ERB Formatter

gobijan:

My package is a formatter for ERB files often used in Ruby on Rails applications.

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -1173,6 +1173,16 @@
 			]
 		},
 		{
+			"name": "ERB Formatter",
+			"details": "https://github.com/gobijan/ERBFormatter",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "master"
+				}
+			]
+		},
+		{
 			"name": "ERB Snippets",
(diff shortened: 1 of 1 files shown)


gobijan:
@packagecontrol-bot can you review?

gobijan:
Not sure what is failing... Can anybody help?

FichteFoll:
I took a quick look at it since it was posted on discord.

1. Per convention, your menu entry to open the settings should open the User key bindings (platform-specific).
2. I generally recommend to always create command palette entries for everything that a package can do and that you would define menu entries for, since the command palette is much easier to access using a keyboard.
3. This functionality is also generically provided by the [Fmt package](https://packagecontrol.io/packages/Fmt), which aims to replace 100 packages for different formatters with a unified package design based on configuration. Imo it makes more sense to add multiple known formatter configuration to that package's readme or wiki instead of yet another package to the list, but YMMV. For searchability, a generic package might be harder to find than a specific one.

gobijan:
Hey, I agree with you @FichteFoll.
I'll add to the wiki for the FMT package. Any idea how to promote the FMT package? I didn't know about it before you told me via Discord.

PR #8875, opened 2024-02-19
Add Elevator

kapitanluffy:

My package is Elevator

[2]: https://semver.org

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -607,6 +607,17 @@
 			]
 		},
 		{
+			"name": "Elevator",
+			"details": "https://github.com/kapitanluffy/sublime-elevator",
+			"donate": "https://github.com/sponsors/kapitanluffy",
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Elixir",
 			"details": "https://github.com/elixir-editors/elixir-tmbundle",
 			"releases": [


braver:
Looks like you're basically providing what [LineJumper](https://packagecontrol.io/packages/LineJumper) already does. That's a pretty stale package though. Please find out if you can contribute to that package instead of adding a new package. Maybe you can even take over that repo. Otherwise, if that does not work out, we can consider removing that existing package with yours, and bring the updated and improved functionality to more users.

braver:
Please let us know if things worked out with LineJumper.

kapitanluffy:
> Please let us know if things worked out with LineJumper.

Will do. I contacted the author 👍

braver:
Looks like the answer to [your question](https://github.com/hypebeast/LineJumper/issues/1) is "no" 😄. Feel free to proceed with a replacement of the existing entry with yours.

kapitanluffy:
> Looks like the answer to [your question](https://github.com/hypebeast/LineJumper/issues/1) is "no" 😄. Feel free to proceed with a replacement of the existing entry with yours.

Oh, this has been out of my radar. Yeah, tried reaching out but he did not reply. What should I do to proceed?

braver:
The cleanest thing to do is remove the existing entry for LineJumper, and add "LineJumper" to the previous_names of your package (example: https://github.com/wbond/package_control_channel/pull/8875/files#diff-2b7dc45bc94d46ab4847d3281d04dac17d5f228e762a8491333857d6ca7cc6d2R455). Also ensure your package has a version higher than the last one for LineJumper (1.0.0). This effectively replaces that with yours and will upgrade existing users to the latest version of your package. Bonus points for adding an upgrade message ([docs](https://packagecontrol.io/docs/messaging)) telling those users what happened.

kapitanluffy:
Ok. Got it. I will close this for now. I will add the highlighting feature from linejumper and reopen this PR.

PR #8877, opened 2024-02-22
Take-over Compare Side-By-Side

kaste:
- [x] I have have read [the docs][1].  Yeah, at some point but it's been a while.

My package is https://github.com/kaste/Compare-Side-By-Side forked from https://bitbucket.org/dougty/sublime-compare-side-by-side/

The original repo has not been updated to work with ST4.  The PRs were ignored.  There is no interaction at all.  The hosted screenshots (now removed from the README) have been deleted.  

- This is a minimal take-over, just making it work.  The repo history is intact.
+ (Change:) Doesn't add global key bindings, ie. it uses contextualized bindings 🚀 
+ (Change:) Uses the standard `edit_settings` in the main menu for editing the settings in split view fashion
+ (Change:) Runs on py3.8

Actually I'm running this (my) version for years now and it should be on PC.  I just hoped initially someone else would make it.  Now I do.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3150,7 +3150,11 @@
 			"labels": ["diff", "compare", "comparison"],
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": "<4000",
+					"tags": "st3-"
+				},
+				{
+					"sublime_text": ">=4000",
 					"tags": true
 				}
 			]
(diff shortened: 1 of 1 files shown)


braver:
The proper process here would be to open an issue on their repo (if you haven’t already) and wait for two weeks. If there is no response, we can merge this PR.

kaste:
Gone.  Issues are closed.  PR without comments since ST4 rolled out.  Also the only project by dougty on bitbucket.  🤷‍♀️

braver:
Well in that case, 🚀

PR #8878, opened 2024-02-22
Add Explain PostgreSQL

MGorkov:

My package is **Explain PostgreSQL** https://github.com/MGorkov/explain-postgresql-sublime

It adds context menu entries, but only in sql context and their visibility is conditional.

[2]: https://semver.org

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -1724,6 +1724,17 @@
 			]
 		},
 		{
+			"name": "Explain PostgreSQL",
+			"details": "https://github.com/MGorkov/explain-postgresql-sublime",
+			"labels": ["sql", "postgresql", "formatting", "formatter", "explain", "tools"],
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


MGorkov:
@braver Thanks for you comments!
Could you tell how the context menu items can be switched off ?
Plugin commands have the is_visible() so they show in context menu only if syntax is sql.

braver:
Sure! In the class that defines your command, you can implement an is_visible method. In that method you can check a setting, or the current scope, etc. If you want to apply a setting only to the context menu entry, you need that to be a separate command (a different class), so you can implement the is_visible method specifically for this case. But you can of course call the same command implementation.

Hope that answers your question!

MGorkov:
@braver Thanks for you answer
This already has been implemented - all commands in the plugin have is_visible method that checks current scope.
So, the PR can be merged?

braver:
All set 👍🏻

PR #8881, opened 2024-02-27
New plugin: Ipynb preview as Markdown entry

valentinbarral:

My package is Ipynb preview as Markdown. There is almost no package that handles Jupyter notebooks in ST. This one generates a human-readable (non editable) preview of the notebook, in case you want to read it inside ST or copy the text or source code without having to launch the Jupyter environment.

[2]: https://semver.org

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -1094,6 +1094,18 @@
 			]
 		},
 		{
+			"name": "Ipynb Preview as Markdown",
+			"details": "https://github.com/valentinbarral/IpynbPreviewAsMarkdown",
+			"labels": ["text manipulation", "formatting"],
+			"releases": [
+				{
+					"platforms": "*",
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "IPython Notebook",
 			"details": "https://github.com/maximsch2/SublimeIPythonNotebook",
 			"releases": [


braver:
Nice, looks good 👍🏻 

You say it opens in split view, which in Sublime Text usually means one pane on the left, one on the right. But from your code it looks like it will actually open an output panel. Which would also be better IMO, as it won't affect the user's current layout :) Although your readme does imply left/right layout. Maybe I'm just reading it incorrectly (I don't have any `ipynb` files)?

repository/i.json
```diff
@@ -1093,6 +1093,18 @@
 				}
 			]
 		},
+		{
+			"name": "Ipynb Preview as Markdown",
+			"details": "https://github.com/valentinbarral/IpynbPreviewAsMarkdown",
+			"labels": ["text manipulation", "formatting"],
+			"releases": [
+				{
+					"platforms": "*",
```
braver, line 1102: If it's `*` you can remove the platforms key.

valentinbarral: Hi, 
you are right, it makes more sense to open a separate window to show the preview. I have modified the code so that it now behaves that way.
I have also removed the platforms key as you suggested.

jwortmann:
Hello, just a few suggestions (not required for accepting the package to Package Control, but maybe they would be possible improvements for the package):

- This part of the code
  ```python
        else:
            sublime.error_message("The current file is not an .ipynb file.")
  ```
  could be replaced for example by using a `sublime_plugin.TextCommand` and adding
  ```python
  def is_enabled(self, **kwargs):
      file_name = active_view.file_name()
      return file_name and file_name.endswith('.ipynb')
  ```
  Then the command would only be active for relevant files.

- The `.ipynb` file format (Jupyter Notebook) is not exclusively for Python code. For more flexibility you could for example add an optional command argument which would allow to set the language, like
  ```python
  class PreviewIpynbAsMarkdownCommand(sublime_plugin.TextCommand):
      def run(self, edit, language='python'):
          # ...
          markdown_content += '```{}\n'.format(language) + ''.join(cell['source']) + '\n```\n\n'
  ```

- If you would be okay with dropping support for Sublime Text 2/3 (`"sublime_text": ">=4000"`), you could experiment to use the `mdpopups` dependency in order to render the notebook content as a HtmlSheet, instead of a normal Markdown file. This may look a bit nicer for displaying the Markdown cells as Html and mdpopups also has direct support for the fenced code blocks, including syntax highlighting.
https://facelessuser.github.io/sublime-markdown-popups/api/#new-html-sheet

valentinbarral:
Hi,
I have adapted part of your comments @jwortmann . Now the plugin uses TextCommands that are only activated if the file format is correct. Now it also checks the language used in the source code cells, to use the same in the preview. 

I have also added some extra functionality, such as the possibility to hide the source code cells or the cells containing the output after a run. Regarding the cells with results, when these have images they are saved in the notebook in Base64 format, which messed up the preview if they were added that way directly. So I have made those images to be saved to files in the temp directory and referenced in the markdown as any image.

braver:
Nice work 👍🏻

PR #8887, opened 2024-03-10
Add the "Redo" package

ISSOtm:
- [x] Any commands are available via the command palette. (It's a build system.)

My package adds support for [the Redo build system](https://redo.rtfd.io/), adding a build system and syntax definition.

For bonus points, also considered how the review guidelines apply to this package:
https://github.com/wbond/package_control_channel/wiki#reviewing-a-package-addition

[2]: https://semver.org

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -863,6 +863,17 @@
 			]
 		},
 		{
+			"name": "Redo",
+			"details": "https://github.com/ISSOtm/sublime-redo",
+			"labels": ["redo", "language syntax", "build system"],
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Redux Snippets",
 			"details": "https://github.com/reduxframework/snippets_sublime",
 			"releases": [


braver:
Please use a consistent suffix to your scopes, even in the hidden syntax. e.g. `redo_ouput`.

I think there might be a better way to create your syntax, by simply extending the default bash syntax. Try this:


```yaml
%YAML 1.2
# https://www.sublimetext.com/docs/syntax.html
---
name: Redo
scope: source.redo
version: 2
extends: Packages/ShellScript/Bash.sublime-syntax

file_extensions:
  - do

```

braver:

ISSOtm:
> Please use a consistent suffix to your scopes, even in the hidden syntax. e.g. `redo_ouput`.

I've done it that way for consistency with other syntaxes, like the built-in `Makefile`, or e.g. `Rust Enhanced`. Not sure which way should be privileged.

> I think there might be a better way to create your syntax, by simply extending the default bash syntax. Try this:
> 
> [...]
> 
> This might mean you'd have to bump your requirement to something more recent though, which should not be a problem anyway. I would suggest going with 4142 as a minimum (that's already more than a year old).

Okay, fair enough. I can't perform the modification right now, but I will. [Documentation indicates that the requirement is 4075](https://www.sublimetext.com/docs/syntax.html#inheritance), FWIW.

braver:
> I can't perform the modification right now, but I will

Alright, just report back when you're ready and we can continue this PR. I will close it in the mean time.

PR #8891, opened 2024-03-16
Sass: v3 tags, prep for 4169+

braver:
cc @deathaxe

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -175,7 +175,7 @@
 			"name": "Sass",
 			"details": "https://github.com/SublimeText/Sass",
 			"labels": ["language syntax", "scss", "sass", "completions", "snippets"],
-			"releases":  [
+			"releases": [
 				{
 					"sublime_text": "<4107",
 					"tags": "st3-"
--- a/repository/s.json
+++ b/repository/s.json
@@ -173,8 +173,7 @@
 		},
 		{
 			"name": "Sass",
-			"details": "https://github.com/braver/Sass",
-			"donate": "https://paypal.me/koenlageveen",
+			"details": "https://github.com/SublimeText/Sass",
 			"labels": ["language syntax", "scss", "sass", "completions", "snippets"],
 			"releases": [
 				{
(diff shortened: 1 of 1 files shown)


deathaxe:
How about using least required ST build number as prefix?

LSP started to use it and it has proven useful as it clearly indicates which version is required.

braver:
Yeah that’s smart, we’ll probably need to freeze and branch off some more for future versions

deathaxe:
Appears bundling CSS is an easy task.

Hence I suggest creating a `4107-` prefix which points to `st4107` branch, which includes bundled CSS from ST4169 and a `4169-` prefix, which points to current master (without the CSS file).

The `v2-` and `v3-` prefixes are obsolete then and all stable ST4 releases are served with latest fixes.

deathaxe:
With the tags and branch created in Sass Repo, we could go with this.

```json
		{
			"name": "Sass",
			"details": "https://github.com/SublimeText/Sass",
			"labels": ["language syntax", "scss", "sass", "completions", "snippets"],
			"releases": [
				{
					"sublime_text": "<4107",
					"tags": "st3-"
				},
				{
					"sublime_text": "4107 - 4168",
					"tags": "4107-"
				},
				{
					"sublime_text": ">=4169",
					"tags": "4169-"
				}
			]
		},
```

braver:
@deathaxe looks like we're all set. Let me know if you're also ready to hit merge here.

deathaxe:
Let's go.

deathaxe:
We propably should merge ASAP. I wasn't aware of Package Control already picking up the 4.0.0 tag.

deathaxe:
It might break SCSS for older ST4 releases.

braver:
Oh, right, of course it would do that. I added a message, moved some tags around. Should be all good now 🤞🏻

deathaxe:
Backported it to st4107 and st4134 branches and updated their tags.

4.0.0 is currently set onto st4134 branch so everyone upgrading until next crawler cycle will get it with bundled CSS to avoid failures.

PR #8897, opened 2024-03-26
Add ImperatorTools package

dementive:
- [x] Preferences are listed in the menu and the command palette, and open in split view.

My package is ImperatorTools — A plugin for Imperator: Rome scripting that has a syntax, auto completion, validation, go to definition, and a ton more features.

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -348,6 +348,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


braver:
Much of the same feedback applies here as well, so I'd like to at least for now keep a single discussion going in #8898

PR #8902, opened 2024-03-30
Add Markdown File Links Plugin

kevinfiol:

My package is [Markdown File Links](https://github.com/kevinfiol/markdown_file_links)

[2]: https://semver.org

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -491,6 +491,17 @@
 			]
 		},
 		{
+			"name": "Markdown File Links",
+			"details": "https://github.com/kevinfiol/markdown_file_links",
+			"labels": ["markdown"],
+			"releases": [
+				{
+					"sublime_text": ">=3211",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Markdown HTML Preview",
 			"details": "https://github.com/zeyon/MarkdownHtmlPreview",
 			"labels": ["markdown", "preview", "build", "html"],


braver:
Interesting, looks useful 👍🏻 

One thing I'd like to see is a command palette entry in addition to the context menu entry, to make it easy to use the functionality from the keyboard (and also to consistently have all functionality exposed via the palette).

Your is_markdown_file method looks for file extensions. That means that files without extension (that are configured to open as markdown, for instance some README files) are not recognized and the feature won't work there. Or maybe the user is using files with the md extension with a different syntax (I seem to remember more languages use that extension, although markdown is definitely more common). A more reliable way of checking the filetype or syntax of a file is to use [match_selector()](https://www.sublimetext.com/docs/api_reference.html#sublime.View.match_selector), passing it a 0,0 point (ie. the first position in the file) and matching for the `text.html.markdown` scope.

kevinfiol:
Thanks for the feedback @braver 

> One thing I'd like to see is a command palette entry in addition to the context menu entry, to make it easy to use the functionality from the keyboard (and also to consistently have all functionality exposed via the palette).

How would you imagine this would work? The best thing that comes to mind is this flow:

1. User left-clicks on a link so the cursor is currently "on" a link (like imagine the pipe is the cursor here `[[hel|lo]]`)
2. Open command pallete, and there should be an option to open the link

> A more reliable way of checking the filetype or syntax of a file is to use [match_selector()](https://www.sublimetext.com/docs/api_reference.html#sublime.View.match_selector), passing it a 0,0 point (ie. the first position in the file) and matching for the text.html.markdown scope.

Can you give me an example of doing this? It's been pretty difficult for me in general to grok the Sublime plugin API :sweat_smile:

braver:
>  It's been pretty difficult for me in general to grok the Sublime plugin API

I can imagine, it's huge! 

> How would you imagine this would work

You could just, once the command is triggered, get the user's current [selection](https://www.sublimetext.com/docs/api_reference.html#sublime.Selection), which is a set of [regions](https://www.sublimetext.com/docs/api_reference.html#sublime.Region), where there start and end of each region is a ["point"](https://www.sublimetext.com/docs/api_reference.html#sublime.Point) similar to the `cursor_point` you're currently using. You could simply and naively take the first region and the starting point of that region. 

> Can you give me an example of doing this

- https://github.com/braver/ColorHints/blob/master/ColorHints.py#L53
- https://github.com/search?q=repo%3ASublimeText-Markdown%2FMarkdownEditing%20match_selector&type=code
- https://github.com/MGorkov/explain-postgresql-sublime/blob/master/utils/utils.py#L30

deathaxe:
The functionality of this plugin is more or less already part of MarkdownEditing.

kevinfiol:
Thanks @deathaxe . I've been using your plugin for so long and had no idea this functionality was included already. I noticed it was buried at the [bottom of the Usage page!](https://sublimetext-markdown.github.io/MarkdownEditing/usage/#wiki). 

Anyway, if it's preferable to close this PR, please proceed @braver

braver:
Well we can always reconsider if the demand arises, but for now let’s close and not add something new to package control.

PR #8903, opened 2024-04-02
OpenAI minimal version bump.

yaroslavyaroslav:

It's nothing more but a bump of a minimal version of the plugin. Actually it never was compatible with ST3 due to python 3.8 usage right from the start. So just making things right finally.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -548,19 +548,6 @@
 			]
 		},
 		{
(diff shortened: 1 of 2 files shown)


yaroslavyaroslav:
I would've ask to not rush this one. Since now it includes renaming, which I have to prepare for on the plugins side.

PR #8905, opened 2024-04-06
Move Formatter from repo to channel

bitst0rm:

It is just convenient for me to manage package infos by myself using channel. Thanks!

My package is ... Formatter

[2]: https://semver.org

### DIFF
--- a/channel.json
+++ b/channel.json
@@ -15,7 +15,7 @@
 		"https://raw.githubusercontent.com/Andr3as/Sublime-SurroundWith/master/packages.json",
 		"https://raw.githubusercontent.com/apophys/sublime-packages/master/packages.json",
 		"https://raw.githubusercontent.com/BenjaminSchaaf/sbnf/master/package_control/packages.json",
-		"https://raw.githubusercontent.com/bitst0rm-stx/meta/master/packages.json",
+		"https://raw.githubusercontent.com/bitst0rm-pub/Formatter/master/packages.json",
 		"https://raw.githubusercontent.com/blachniet/sublime-mimosa/master/packages.json",
 		"https://raw.githubusercontent.com/blake-regalia/linked-data.syntaxes/master/channels/sublime/package-control.json",
 		"https://raw.githubusercontent.com/blueplanet/sublime-text-2-octopress/master/packages.json",


deathaxe:
What's the specific reason for this change?

I'd rather prefer to move packages from external repos to the internal one as each external repo means a risk due to uncontrollable changes.

FichteFoll:
The primary motivations to have packages specified directly inside of this repository instead of an external `repository.json` file are that we can 

1. batch-update package specifications in case of a schema version update (4.0.0 exists already but the crawler code currently deployed on packagecontrol.io is behind so we haven't been able to update the repo)
2. review newly added packages

FichteFoll:
The location of the `repository.json` file (this is PC terminology) does not matter here. It's the fact that a repository can contain multiple packages and is outside of our control that is relevant. The only way to address our concerns would be by effectively reverting this PR (i.e. rejecting it). That said, while we heavily encourage people to specify their packages inside the json files inside of this git repository directly, I don't think we have a stance on requiring it when there is no technical need to do otherwise and we should discuss that. I'll raise that discussion elsewhere and report back with our conclusion.

<sub>Regardless of malicious intents (which are rare), there are benefits to a manual review process, such as giving direct feedback to package authors, making them aware of shortcomings with their package and thus also maintaining a minimum level of quality for packages available for installation. This is in my opinion still worth it (assuming we have the manpower to perform such reviews) becuase it will be beneficial to the ecosystem in the long run. 

Also, the change to `channel.json` last week was actually [a removal of an entry](https://github.com/wbond/package_control_channel/commit/b0287ef6dfb035c4600f14f5160abdb95e05b11c).</sub>

FichteFoll:
> I am almost sure this plugin, also its creator, are at least not less qualified than the most of them.

I'm afraid I don't understand what you mean with this.

> Mir ist wichtig, dass es gleiche Rechte für alle gibt.

(Replying in english so that others may also follow the conversation:) We do have a goal of treating everyone the same in our various judgements during a package review. There are some decisions that are subjective because no objective ruleset has been established yet and we haven't codified every single step in this process (even though we codified a lot already). If there is a particular reason why you believe you're being treated unfairly, please address the issue directly so that we can look into it in more detail and perhaps explain why certain things are (still) in a certain way even though they go against our policies.

PR #8911, opened 2024-04-19
Spin-off PyTest ST3 version

kaste:
Prepare for a ST4 only release for PyTest (https://github.com/kaste/PyTest) as it is broken on Apple Silicon on py33.  
Bug report:  https://forum.sublimetext.com/t/sublime-text-pytest-package/71958/17

https://github.com/kaste/PyTest/tags


kaste:
Thanks @braver

PR #8912, opened 2024-04-23
Update dependencies.json

TerminalFi:
- [ x I use [.gitattributes][3] to exclude files from the package: images, test files, sublime-project/workspace.

My package is vision. A lib for making minihtml generation easier by implementing functions for support css and tags

### DIFF
--- a/repository/dependencies.json
+++ b/repository/dependencies.json
@@ -498,22 +498,6 @@
 			]
 		},
 		{
-			"name": "vision",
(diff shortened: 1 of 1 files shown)


predragnikolic:
While it can make sense that the dep is named "minihtml",
It can cause confusion in the future.

If someone reports an issue about "minihtml"
The name will be ambiguous, 
and you would first need to deduce of they are taking about ST minihtml concept, 
Or your dependency.

FichteFoll:
Superseded by https://github.com/packagecontrol/channel/pull/10.

PR #8913, opened 2024-04-28
Add MCNP syntax

deathaxe:

My package provides MCNP syntax definition and snippets for ST4.

Syntax is primarily sourced from https://github.com/repositony/vscode_mcnp and has been converted to sublime-syntax format with some optimizations added.

[2]: https://semver.org

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -1141,6 +1141,17 @@
 			]
 		},
 		{
+			"name": "MCNP",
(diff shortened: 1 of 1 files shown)


FichteFoll (approved):
Happened to look at this already briefly when it was pushed (since I get emails for that), so this was a quick review.

PR #8915, opened 2024-05-05
Add new color scheme: "Wakame"

toshimaru:

My package is ...

I newly create a color scheme names "**Wakame**".

[2]: https://semver.org

### DIFF
--- a/repository/w.json
+++ b/repository/w.json
@@ -24,6 +24,17 @@
 			]
 		},
 		{
+			"name": "Wakame Color Scheme",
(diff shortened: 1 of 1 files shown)


repository/w.json
```diff
@@ -23,6 +23,17 @@
 				}
 			]
 		},
+		{
+			"name": "Wakame",
```
braver, line 27: Many color schemes have "Color Scheme" in their name, makes them easy to identify. You might want to consider doing the same.

toshimaru: @braver 
I added "Color Scheme" as a suffix.

PR #8916, opened 2024-05-09
Add Starlark syntax

Vasfed:

My package is syntax highlighting for Starlark language.

[2]: https://semver.org

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -3712,6 +3712,17 @@
 			]
 		},
 		{
+			"name": "Starlark",
+			"details": "https://github.com/Vasfed/sublime_starlark",
+			"labels": ["language syntax"],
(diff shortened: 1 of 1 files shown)


Vasfed:
@packagecontrol-bot please review

braver:
Well done 👍🏻

deathaxe:
What's the main difference between Starlark and the syntax provided by https://github.com/niosus/BazelSyntax?

braver:
Oh, nice find… one thing this new one has going for it is that it has the correct name 😁. Starlark is not limited to the Bazel project, it’s a language all on its own. .. Although I guess we did duplicate a bit now, the new one is probably much more findable (the old one doesn’t even mention Starlark).

PR #8919, opened 2024-05-14
Take over Materialize development

zyphlar:
The existing repo has been archived for 4 years and lacking activity for 7; I've merged an open PR enabling colored title bars and published to Github.

- [x] I'm the package's ***new*** author and/or maintainer.

My package is a theme

My package is a fork of Materialize, however it should still be added because the existing Materialize has been abandoned and marked as archived for a long time.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -924,8 +924,7 @@
 		},
 		{
 			"name": "Materialize",
-			"previous_names": ["Material Spacegray"],
-			"details": "https://github.com/saadq/Materialize",
+			"details": "https://github.com/zyphlar/Materialize",
 			"labels": ["theme", "color scheme", "material"],
 			"releases": [
 				{


FichteFoll:
Regarding your changes, I suggest to re-use the existing `themed_title_bar` setting for your theme instead of inventing your own (`material_theme_title_bar`) as it is encouraged to be used for this purpose. Furthermore, I don't really understand your [comment about "On OSX"](https://github.com/saadq/Materialize/compare/master...zyphlar:Materialize:master#diff-123226ec5ca72a09fb432c327f3299478e3031d2e53588b31748003451fe1ba9R7) because it will be active on all platforms.

Finally, I'm also pinging @saadq to inform them of this takeover. The old repo is archived, so I don't expect any objections, but even for archived repos it makes sense to respect out 2 week grace period for the previous maintainer to raise concerns.

zyphlar:
I'll investigate, thanks -- so far this is just a merge of upstreams so I'll have to develop and test improvements.

Also happy to make a new/renamed project like `Materialize2024` or something out of respect for @saadq's hard work and rights to the project, obviously keeping the same package name will help existing users with updates etc

PR #8922, opened 2024-05-17
Add Interactivity package

ichichikin:

My package is Interactivity.
This plugin allows you to run shell commands and scripts directly within the editor, providing their output right alongside your written content, making your workflow more dynamic and interactive.

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -1299,6 +1299,17 @@
 					"tags": true
 				}
 			]
+		},
+		{
+			"name": "Interactivity",
+			"details": "https://github.com/ichichikin/sublime-plugin-interactivity",
+			"releases": [
+				{
+					"platforms": "*",
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
 		}
 	]
 }
--- a/repository/i.json
+++ b/repository/i.json
@@ -922,6 +922,16 @@
 			]
 		},
 		{
+			"name": "Interactivity",
+			"details": "https://github.com/ichichikin/sublime-plugin-interactivity",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "International Keymap",
 			"details": "https://bitbucket.org/hxss/internationalkeymap",
 			"releases": [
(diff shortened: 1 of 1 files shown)


kaste:
Not gonna lie, that looks good (lgtm).   I think the configuration is a bit awkward. E.g. you have
```
"text_shortcuts": "@ -> ##param##\n@@ -> chat4(r\"\"\"##param##\"\"\")",
```
and I wonder why that isn't just a mapping `prefix: command`.  E.g. `{"@": "##param##"}`.  I'm also not  sure how I would define a command that runs Node instead of python.  Or how I could add my own function, e.g. instead of `chat4`.  Also `##param##` looks unfamiliar (instead of, e.g., `$param` or `${param}`) but that maybe makes sense in your context.

kaste:
Ah, the entry here, add it sorted case-insensitive.

ichichikin:
@kaste I'm glad that you liked the package. I made changes to settings management as you suggested. 

Regarding `##param##`, I intentionally picked this unusual syntax to reduce the chance of interference with any possible REPLs.

For interacting with Node, please try these settings:

```json
{
    "append_output": "",
    "environment_variables": {},
    "lines_to_suppress": 2,
    "output_filter": "^>\\s(?:\\.\\.\\.\\s)*",
    "prepend_output": " ",
    "shell": "node",
    "shell_params": [
        "-i"
    ],
    "shutdown_commands": ".exit",
    "startup_commands": "",
    "text_shortcuts": {
        "@": "##param##"
    }
}
```

If you want to add custom functions in Python, please follow these [steps](https://github.com/ichichikin/sublime-plugin-interactivity/?tab=readme-ov-file#custom-functions).

I'd be glad if you commit any useful modules for Node as well.

ichichikin:
@kaste thank you for sharing your ideas!

I didn't want to overload the plugin too much, but I like your idea of accessing different shells simultaneously. I will consider this functionality for future releases.

I've added all the necessary files to my repository.

kaste:
Check the sorting https://github.com/wbond/package_control_channel/pull/8922#issuecomment-2174334550

ichichikin:
@kaste, it looks like I'm missing something here. Could you please tell me more about what exactly should be sorted case-insensitively?

kaste:
Then entries in `i.json` should be sorted by name.

ichichikin:
@kaste, got it, thanks! Done.

kaste:
@braver I think that LGTM.

braver:
@kaste , thanks, your help here is really appreciated 👍🏻

PR #8923, opened 2024-05-18
Update b.json

RohanVashisht1234:

My package is a syntax highlighter for the Bend programming language

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -515,6 +515,17 @@
 			]
 		},
 		{
+			"name": "Bend programming language",
+			"details": "https://github.com/RohanVashisht1234/bend-syntax-highlighter-sublime-text",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Berry Color Scheme",
 			"details": "https://github.com/patoui/berry",
 			"labels": ["color scheme"],


RohanVashisht1234:
there might be mistakes, so if there are mistakes, please tell me, I'll create the release once all the issues are solved.

deathaxe:
The repository doesn't provide a valid Sublime Text syntax definition.

RohanVashisht1234:
Ok, I'll update the syntax soon

braver:
```
error parsing lexer: Packages/User/bend.sublime-syntax: Error trying to parse sublime-syntax: %YAML indicator is required in 
```

Let us know when your package is ready to review. I'll close it in the mean time, but we can re-open when you're ready.

RohanVashisht1234:
fine np

RohanVashisht1234:
> Let us know when your package is ready to review.

I have updated the bend.sublime-text

RohanVashisht1234:
ok, so there must be a fallback file named .tmLanguage, is there a way to convert .sublime-syntax to .tmLanguage?

deathaxe:
Not necessary, because ST builds before 3143 are unlikely to be in use these days. 

Package Control 4 requires 3143+ as well.

You could restrict the package to ST builds larger or requal than 3092, to disable the message.

deathaxe:
Would it make sense to call the package just "Bend"?

RohanVashisht1234:
fine

> Would it make sense to call the package just "Bend"?

RohanVashisht1234:
ok, is it ready to merge?

PR #8924, opened 2024-05-20
Add WASM Interface Type syntax definition package

deathaxe:

My package is a syntax definition package for WIT aka. WebAssembly Interface Type.

It implements the syntax according to:

- https://component-model.bytecodealliance.org/design/wit.html
- https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md

[2]: https://semver.org

### DIFF
--- a/repository/w.json
+++ b/repository/w.json
@@ -68,6 +68,21 @@
 			]
 		},
 		{
+			"name": "WASM Interface Type",
+			"details": "https://github.com/SublimeText/WasmInterfaceType",
+			"labels": ["language syntax", "wasm", "wit"],
+			"releases": [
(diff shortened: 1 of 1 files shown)


braver:
Is `wit` in any way related to [wast, for which there is a package](https://packagecontrol.io/packages/WebAssembly%20Text%20Syntax)?

deathaxe:
WAST is a text representation of WebAssembly byte code. WIT is an interface specification language to create interoperable WASM modules and reduce amount of so called "glue code" to create language bindings. It is part of the WASI project. WAST and WIT are independent syntaxes though.

PR #8925, opened 2024-05-22
Add OpenFileOverSSH plugin for remote file editing

AncientRoman:

My package is a lightweight plugin for opening, editing, and saving of files on a remote machine using only ssh.
It adds the Open File Over SSH command which is available in the command plate, as a menu item, and as an optional keybinding that must be enabled.

My package seems similar to [SSH-Panel](https://packagecontrol.io/packages/SSH-Panel), however that package requires manually installing dependences (by editing the sublime installation) and doesn't claim support for MacOS.

[2]: https://semver.org

### DIFF
--- a/repository/o.json
+++ b/repository/o.json
@@ -783,6 +783,16 @@
 			]
 		},
 		{
+			"name": "OpenFileOverSSH",
+			"details": "https://github.com/AncientRoman/SublimeOpenFileOverSSH",
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


braver:
Nice work! Looks like you've been building for quite a while now. And looks like a good addition to the eco system. There are similar packages out there, but sometimes outdated in their approach, or simply never left "WIP status". So, let's roll 🚀

PR #8929, opened 2024-06-03
Add Ruby Gems viewer package

ZainIftikhar7vals:

My package is Ruby Gem viewer.

[2]: https://semver.org

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -1921,6 +1921,19 @@
 			]
 		},
 		{
+			"name": "GotoRubyGem",
+			"details": "https://github.com/ZainIftikhar7vals/GotoRubyGem",
+			"readme": "https://raw.githubusercontent.com/ZainIftikhar7vals/GotoRubyGem/main/README.md",
+			"labels": ["code navigation", "ruby gems"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": ["osx", "linux"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "GotoTab",
 			"details": "https://github.com/SublimeText/GotoTab",
 			"releases": [
--- a/repository/g.json
+++ b/repository/g.json
@@ -1923,8 +1923,7 @@
 		{
 			"name": "GotoRubyGem",
 			"details": "https://github.com/ZainIftikhar7vals/GotoRubyGem",
-			"readme": "https://raw.githubusercontent.com/ZainIftikhar7vals/GotoRubyGem/main/README.md",
-			"labels": ["code navigation", "ruby gems"],
+			"labels": ["code navigation", "ruby gems", "ruby"],
 			"releases": [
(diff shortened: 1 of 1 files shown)


kaste:
This looks good to me but I do wonder

```
      gem_home = os.popen("echo $GEM_HOME").read().split("\n")[0]
      gems_directories = [
        os.path.expanduser("%s/gems/" % gem_home),
        os.path.expanduser("%s/bundler/gems/" % gem_home)
      ]
```

Isn't that readable just from within Sublime Text using `os.environ["GEM_HOME"]`.  Would be nice to make it available for Windows too.  Is that impossible to achieve?

The latest tag right now is "0.2.0-alpha".  Not sure if a suffix (`-alpha`) counts as release version.  braver will know that, or change it to just `0.3.0`.

kaste:
* Please lift the plugin to run on the 3.8 host. For that, create a file .python-version with its only contents 3.8\n in the root of your package.

ZainIftikhar7vals:
> * Please lift the plugin to run on the 3.8 host. For that, create a file .python-version with its only contents 3.8\n in the root of your package.

Can you please explain it a little bit more, as I am new to Sublime package development?

ZainIftikhar7vals:
> Would be nice to make it available for Windows too. Is that impossible to achieve?

It's definitely possible, but since I don't have access to a Windows machine right now, I can't test it. However, I plan to do so in the future.

kaste:
Yeah, to "lift" the package (or better worded: choose the environment) create a file just like https://github.com/kaste/InlineOutline/blob/master/.python-version.

ZainIftikhar7vals:
@kaste All comments resolved.

kaste:
We need @braver for the merge.

repository/g.json
```diff
@@ -1920,6 +1920,19 @@
 				}
 			]
 		},
+		{
+			"name": "GotoRubyGem",
+			"details": "https://github.com/ZainIftikhar7vals/GotoRubyGem",
+			"readme": "https://raw.githubusercontent.com/ZainIftikhar7vals/GotoRubyGem/main/README.md",
```
koen-at-peppered, line 1926: Your readme is in the default location so no need to reference it here.
```suggestion
```

repository/g.json
```diff
@@ -1920,6 +1920,19 @@
...
+			"labels": ["code navigation", "ruby gems"],
```
koen-at-peppered, line 1927:
```suggestion
			"labels": ["code navigation", "ruby gems", "ruby"],
```
A general "ruby" label will be more useful actually.

koen-at-peppered (changes requested):
Just some final tweaks to the entry here, looks good otherwise 👍

ZainIftikhar7vals:
@koen-at-peppered changes have been incorporated, please review.

PR #8931, opened 2024-06-04
Migrate to `tags` based releases

gillibrand:
Add `tags: true` for `Expand Selection to Function (JavaScript)` package for future updates.

Just a switch to tags based versioning. The command, keys, etc. are unchanged since... a long time. Only a recent bug fix.

- [ ] My package doesn't add key bindings. ** (It DOES add, but is an old package that always did, so not a change.)

My package is ...

[2]: https://semver.org

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -1639,7 +1639,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"branch": "master"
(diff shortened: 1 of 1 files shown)


braver:
Please remember to stick to semver tags, so 1.1.0 instead of 1.1.

PR #8932, opened 2024-06-05
Add `Alt + E` open currt folder And Fast Print Variable Plug

mucpsing:

This is a basic plugin with the following main features:

When using the shortcut key alt+e, it will invoke the file manager to quickly open the directory where the current file is located.
A right-click menu item Cps Print Currt Param is added, which allows for quickly printing the variable selected by the current cursor (currently only supports Python and JavaScript).
It integrates some commonly used shell invocation methods (under Windows) for quick testing and debugging.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -5032,6 +5032,18 @@
 					"tags": true
 				}
 			]
+		},
+		{
+			"name": "cps plugins",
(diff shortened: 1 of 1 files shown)


braver:
Please respond to continue the review.

PR #8934, opened 2024-06-06
feat: add Tact package

novusnota:
- [x] Any commands are available via the command palette. It doesn't add new commands, only a new build option, which gets into the command palette automagically.
- ~~[ ] Preferences and keybindings (if any) are listed in the menu and the command palette, and open in split view.~~ This package only sets indentation settings (4 spaces) for `.tact` files and doesn't add any keybindings, so I'm not sure this point applies here.

My package is adding various support for [Tact language](https://tact-lang.org/): syntax highlighting, auto-completions, snippets, simple build integration and more. Full list of features is specified on the README.

Link: https://github.com/tact-lang/tact-sublime

There are no packages like it in Package Control. Hope everything's correct here :)

[2]: https://semver.org

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -243,6 +243,19 @@
 			]
 		},
 		{
+			"name": "Tact",
+			"details": "https://github.com/tact-lang/tact-sublime",
+			"description": "All-in-one package for Tact programming language",
+			"author": "Novus Nota",
+			"labels": ["tact", "auto-complete", "build system", "code navigation", "language syntax", "snippets"],
+			"releases": [
+				{
+					"sublime_text": ">=3210",
+					"tags": "v"
+				}
+			]
+		},
+		{
 			"name": "TADS3",
 			"details": "https://github.com/VoidPhantom/sublime-tads3",
 			"labels": ["language syntax"],


novusnota:
@packagecontrol-bot

braver:
If you're not sharing the tmLanguage syntax with packages for other editors, I would suggest upgrading it to the sublime-syntax. 

You might also want to consider contributing an [LSP-tact package](https://lsp.sublimetext.io/language_servers/) to help users install and set that up. 

You're using some settings here that you don't expose to the user, that looks intentional to me? You can't guarantee users won't change settings afterwards though. But, it should help users start off with the "correct" or canonical defaults.

Anyway, looks good, thanks!

novusnota:
> If you're not sharing the tmLanguage syntax with packages for other editors, I would suggest upgrading it to the sublime-syntax.

Good suggestion, I've looked into it and that syntax was much sweeter when specifying nested constructs. However, I indeed use tmLanguage syntax in bunch of other: it's used then in VSCode, Monaco (editor powering VSCode) and Shiki highlighting (for docs). And we've decided for the root of truth of the tmLanguage syntax to be in Sublime, because you have neat testing system in place for highlighting!

> You might also want to consider contributing an [LSP-tact package](https://lsp.sublimetext.io/language_servers/) to help users install and set that up.

Sure thing! However, the current LSP is rather soon will be superseded by a newer one, so I'll be contributing it then :)

> You're using some settings here that you don't expose to the user, that looks intentional to me? You can't guarantee users won't change settings afterwards though. But, it should help users start off with the "correct" or canonical defaults.

Yup, that's precisely the reason to set the indentation (and only it)

> Anyway, looks good, thanks!

:heart:

braver:
👍🏻👍🏻

PR #8935, opened 2024-06-10
Add InlineOutline

kaste:

My package is https://github.com/kaste/InlineOutline

** key bindings can be configured.  It ships contextual bindings; these can be turned off via settings.  

Unbenanntes Video – Mit Clipchamp erstellt (3)

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -736,6 +736,16 @@
 			]
 		},
 		{
+			"name": "InlineOutline",
+			"details": "https://github.com/kaste/InlineOutline",
+			"releases": [
+				{
+					"sublime_text": ">=4132",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Inno Setup",
 			"details": "https://github.com/idleberg/sublime-innosetup",
 			"labels": ["language syntax", "build system", "inno setup", "innosetup"],


braver:
Nifty 👍🏻 

That's an interesting approach to the keybindings. I'm not sure that's where I'd go looking for changing the keybindings (over settings > package > keybindings). I'm also a bit confused about your example keybinding with the additional argument, from the readme... do I put that in the settings file, or in the keybindings file? And if I put it in the keybindings file, what does the setting in the settings file do precisely, since the keybind now defines the keys? And is `write_key_bindings` ever going to mess with bindings I set up manually?

kaste:
Yeah, hi!

The example binding in the README is basically an additional binding. I don't use it, but I *could* imagine it being useful for filtering the list of symbols to just functions, just classes, etc. It is just an additional example, as you could use the command to define a command for the Command Palette or some context menu. The example itself is not generic enough to work as a global key binding, in my opinion, as it would only work for languages that use "class" verbatim. As such, it is *not* a replacement (again, in my opinion) for the other, automatic bindings.

But, as it is in the README, I make an API promise that this usage is something I will stay true to. And if you just add the example to your user key bindings using the *same* chord (e.g., `ctrl+shift+o`) as the main key in the settings file, it will work as you expect. It takes precedence, and everything else (the toggling of the mode, etc.) will work.

Other than that, the key binding approach is novel because right now, I need three bindings to make the "mode" work as I want it. And, of course, I need context constraints for basically all of them, too. That is a bit complicated by itself, but the point is: I don't want that as my API. My gut tells me I didn't get it right the first time, so I expect a changing API, i.e., changing key bindings, especially key contexts.

Of course, `write_key_bindings`, or however this will be named after refactoring, will *never* interfere with your key bindings. That would be a non-starter to begin with. (Like the old days when plugins added rules to the color themes/schemes.)

*EDIT:* I update the README.

kaste:
I thought about an "leave-my-keys-alone" setting, one thing you can toggle so the plugin doesn't grab anything from your muscle memory key layout.  But I have that as well, just set `never` as the binding.  From that perspective, it is a matter of the right default.  Ship with "never" or with `primary+shift+o` and I choose the later because a plugin that does nothing and is effectively dysfunctional after installing, I don't like that.   *TODO*:  Add an *install.txt* file.

braver:
Thanks for all that detail. I knew you’d have put a lot of thought in it and it’s a totally reasonable approach too. Again: nifty 👍🏻

PR #8936, opened 2024-06-18
Add Dark Castle Color Scheme

scottgriv:

My package is similar to color schemes in general... However it should still be added because ...
I'm creating a new multi-platform theme, that's unlike any theme I've personally come across.
I've built it up for a [number of platforms](https://darkcastle.netlify.app/), and I would love to have Sublime included in my list.

Thanks!

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -117,7 +117,7 @@
 			"name": "Dark Castle Color Scheme",
 			"author": "Scott Grivner",
 			"details": "https://github.com/scottgriv/Dark-Castle-Sublime",
-			"labels": ["theme", "color scheme", "dark theme", "dark castle", "dark"],
+			"labels": ["color scheme"],
 			"releases": [
 				{
 					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


repository/d.json
```diff
@@ -113,6 +113,18 @@
 				}
 			]
 		},
+		{
+			"name": "Dark Castle Color Scheme",
+			"author": "Scott Grivner",
+			"details": "https://github.com/scottgriv/Dark-Castle-Sublime",
+			"labels": ["theme", "color scheme", "dark theme", "dark castle", "dark"],
```
braver, line 120: In Sublime Text parlance this is a color scheme and _not_ a theme, so it should also not have that label. Repeating your theme’s name as a label also feels a bit superfluous.

braver (changes requested):
Also have a look at the bot’s feedback please!

repository/d.json
scottgriv: Good points. Thank you for your feedback. I'll make the appropriate changes.

braver:
Package Control only looks at your tags, so if you made an improvement please also tag a new release.

scottgriv:
Thanks for your help.

PR #8937, opened 2024-06-18
Blitz Search 

Natestah:
- [ ✔️] I'm the package's author and/or maintainer.
- [✔️ ] I have have read [the docs][1].
- [ ✔️] I have tagged a release with a [semver][2] version number.
- [✔️ ] My package repo has a description and a README describing what it's for and how to use it.
- [✔️ ] My package doesn't add context menu entries. *
- [✔️ ] My package doesn't add key bindings. **
- [✔️ ] Any commands are available via the command palette.
- [ ✔️] Preferences and keybindings (if any) are listed in the menu and the command palette, and open in split view.

My package is ...

There are no packages like it in Package Control.  Blitz Search Provides a simple interface to control what is being search in the external Blitz Search Tool that I'm developing at natestah.com.

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -1067,6 +1067,16 @@
 			]
 		},
 		{
+			"name": "BlitzSearch",
+			"details": "https://github.com/Natestah/BlitzSt",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},		
+		{
 			"name": "Block Cursor Everywhere",
 			"details": "https://github.com/karlhorky/BlockCursorEverywhere",
 			"releases": [
--- a/repository/b.json
+++ b/repository/b.json
@@ -1072,6 +1072,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
+					"platforms": ["windows"],
 					"tags": true
 				}
 			]


braver:
Thanks for your addition here. I'm not going to necessarily test your package, but from a quick read.. are you sure this is going to work on someone else's machine?

```py
ipcPath = os.path.join(appdata, 'NathanSilvers','POORMANS_IPC')
ipcFile = os.path.join(ipcPath,'SET_SEARCH.txt' )
```

Natestah:
Good catch.. I've updated and added a licence.md to the root of that repo.

kaste:
Please lift the plugin to run on the 3.8 host. For that, create a file .python-version with its only contents 3.8 in the root of your package.

The video on your website shows a way fancier live/on-the-fly search thing using a typical text/input panel/box.  Could you expand this demo plugin to that?  Would that be the goal?

Natestah:
> Please lift the plugin to run on the 3.8 host. For that, create a file .python-version with its only contents `3.8\n` in the root of your package.
> 
> The video on your website shows a way fancier live/on-the-fly search thing using a typical text/input panel/box. Could you expand this demo plugin to that? Would that be the goal?

This plugin requires a download from the website of the external tool.   The goal with the plugin is that I'll never really have to change anything plugin-wise. It's a very thin layer so that I can have Blitz Search Running in more than one text editor / IDE. 

You can run the "Blitz Search This" command which will highlight the text in the box, from there you can type over that with whatever you want and search for anything.

Natestah:
There's a number of things to do with the tool itself. Perhaps the worst part of being independent is that people are going to have to click through scary "Windows Defender messages" to get it installed.  Which is unfortunate for me, I've only had a few friends testing it so far.  There's a discord channel if you're feeling bold and brave enough to click through those. Updated that python-version file for you

kaste:
Yeah, I thought it would be real IPC in the long run; you send the contents of an input field to a Blitz-server, and display the response in an editor view.  A real-time, as-you-type search is missing in Sublime.

Natestah:
The real IPC thing was a thought early on but I realized I would have to do UI in Many Unfamiliar languages to get a Uniform support, I decided to stick with my strength C# WPF (now Avalonia) where I could do awesome things and then implement this super simple plugin for as many Editors as possible,  I plan to do this same thing for a chrome/any browser extension where we can find-locally from within web-browser ( Github, Perforce Swarm Reviews, etc ).

Not to say that it won't ever happen, I have actually framed it up to be "Client/Server" internally to where the Query is a MessagePack object that is Serialized/Deserialized for every requests even though it's in-process, and even tracking the process/identification of each request. The predecessor to this had a Real IPC and Multiple instances of the In-house IDE could access the search server so I kind of have an idea of what that looks like.

It's good that you mentioned it.. Maybe I'll build up some sort of API in the future and delegate that to who ever wants to have a true built in Search in their editor of choice.

PR #8938, opened 2024-06-22
Add SmPL

jtojnar:

My package is syntax for SmPL (as used by [Coccinelle](https://coccinelle.gitlabpages.inria.fr/website/).

There is already a similarly named syntax package for SMPL but that is some other language (it is uclear which one). And the name is actually different.

[2]: https://semver.org

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -2564,6 +2564,17 @@
 			]
 		},
 		{
+			"name": "SmPL (Coccinelle)",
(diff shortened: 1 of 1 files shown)


braver:
Package Control won't accept that with the existing package (the different case doesn't differentiate enough). Maybe we can name your package "Coccinelle SmPL"?

braver:
👋🏻

braver:
Works for me 👍🏻

PR #8939, opened 2024-06-23
Add CSS-Intellisense plugin

id-sakurata:

My package is CSS-Intellisense

My package is similar to "IntelliSense for CSS class names in HTML" from vscode However it should still be added because very useful when developing html page.

[2]: https://semver.org

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1,36 +1,29 @@
 {
 	"schema_version": "3.0.0",
-	"packages": [
-		{
+	"packages": [{
 			"name": "C Improved",
 			"details": "https://github.com/abusalimov/SublimeCImproved",
 			"labels": ["language syntax"],
-			"releases": [
-				{
-					"sublime_text": "*",
-					"tags": true
-				}
-			]
+			"releases": [{
+				"sublime_text": "*",
+				"tags": true
+			}]
 		},
 		{
 			"name": "C# Compile & Run",
 			"details": "https://github.com/chrokh/csharp-build-singlefile-sublime-text-2",
-			"releases": [
-				{
-					"sublime_text": "*",
-					"branch": "master"
-				}
-			]
+			"releases": [{
+				"sublime_text": "*",
+				"branch": "master"
+			}]
 		},
 		{
 			"name": "C# Snippets",
 			"details": "https://github.com/etic/CSharpSnippets",
-			"releases": [
-				{
-					"sublime_text": "*",
-					"tags": true
-				}
-			]
+			"releases": [{
+				"sublime_text": "*",
+				"tags": true
+			}]
 		},
 		{
 			"name": "C++ & C Single File Builder - Minghang Yang",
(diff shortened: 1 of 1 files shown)


repository/c.json
```diff
@@ -4400,6 +4400,17 @@
 				}
 			]
 		},
+		{
+	        "name": "CSS-Intellisense",
+	        "details": "https://github.com/id-sakurata/CSS-Intellisense",
+			"labels": ["css", "autocomplete", "intellisense"],
```
braver, line 4406:
```suggestion
		"labels": ["css", "autocomplete", "intellisense"],
```

braver (changes requested):
Some problems:
- sort your package correctly 
- remove the package_control.json and sublime-package.json files from the package, package control will take care of those
- you shouldn't read settings on plugin load: they're always served from memory can be read at will. that way you also ensure you get the latest version of the setting as users edit them. 
- similarly you will want to update your package on any changes to the settings, also when those don't happen through your specific commands - and not through polling as you're doing with with the cache refresh how. there should be event listeners you could use.

In general I understand the utility (although I wonder if a language service wouldn't be a better approach). However, it seems to be very much set up for a single project. You might want to investigate the use of project settings to allow users to set up "css_folders" for different projects. If you working on many projects those "global" settings will just continue to grow over time. 

In its current state I would say the package is a bit too naive to be generally useful (although I can totally see how it works well for _you_). Maybe you can get others to contribute insights and maybe even code, for instance by talking about it on the Sublime forums and Discord.

repository/c.json
id-sakurata: I made this plugin so that I can select class files according to what I need and don't have to have all class files in the autocomplete cache to reduce RAM usage. maybe this is useful for some people. In the settings I also include whether the user wants to activate the feature to search for classes automatically or not, there is also a quick action to add a folder to the cache by right clicking on the folder list sidebar. sorry my bad english, i can only use google translate.

id-sakurata: This is my first time doing a plugin publication request, so I don't know what to do.

braver:
Looks like you re-formatted the entire JSON file. Please don't do that 😉

braver:
If you have all your settings covered in your default settings file, it's better to not have those default duplicated in both settings.get calls and the defaults for the class properties. You end up effectively defining your defaults 4 times and they get out of sync (have a close look at your scopes setting for instance). 

I wonder if the way you handle settings now, where they're both user settings and temporary caches, doesn't cause issues somewhere down the line. You miss out on Sublime's default handling of automatically updating settings internally whenever users change settings. You seem to end up going deeper and deeper into that rabbit hole, having to call your load_settings method every time your package wants to do anything. 

Maybe it makes sense to separate "pure" settings from the caches. That way you can simply use Sublime's api to get and set settings like `enabled` and `auto_refresh_interval`. The `css_folders` and `css_files` have a more transient aspect to it. But even there I think you might benefit from separating the user setting from the current runtime's caches, and merge the two whenever you want to read those values.

braver:

PR #8942, opened 2024-06-29
Add LSP-Spyglass package

MulverineX:

My package is [Spyglass](https://github.com/SpyglassMC/LSP-Spyglass) which encaptualizes an npm package [here](https://github.com/SpyglassMC/Spyglass/tree/main/packages/language-server)

My package is similar to [MCFunction](https://github.com/AjaxGb/Sublime-MCFunction), however, Spyglass should still be added because the old package is no longer maintained, has fewer features, and is several years of Minecraft updates behind.

[2]: https://semver.org

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -1810,6 +1810,27 @@
 			]
 		},
 		{
+			"name": "LSP-Spyglass",
+			"details": "https://github.com/SpyglassMC/LSP-Spyglass",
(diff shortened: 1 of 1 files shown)


braver:
LSP packages are managed by the sublimelsp org, and you can submit your package there: https://github.com/sublimelsp/repository

PR #8945, opened 2024-07-01
STFileFilter v1.0.0

vsc-cnst:

My package is meant to filter a file content without any changes to the file content

[2]: https://semver.org

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -622,6 +622,21 @@
 			]
 		},
 		{
+			"name": "File Filter",
+			"details": "https://github.com/Vasco1/STFileFilter",
+			"readme": "https://github.com/Vasco1/STFileFilter/blob/main/README.md",
+			"issues": "https://github.com/Vasco1/STFileFilter/issues",
+			"author": "vsc",
+			"labels": [ "file content search", "file content query", "filer", "search", "query", "content", "logs", "regex" ],
+			"releases": [
+				{
+					"sublime_text": ">3000",
+					"tags": true,
+					"python_versions": [ "3.3", "3.8" ]
+				}
+			]
+		},
+		{
 			"name": "FileHeader",
 			"details": "https://github.com/shiyanhui/FileHeader",
 			"labels": ["file creation", "templating", "file header"],
--- a/repository/f.json
+++ b/repository/f.json
@@ -528,7 +528,7 @@
 			"labels": [ "file filter", "file content filter", "logs filter", "file", "filter", "regex" ],
 			"releases": [
 				{
-					"sublime_text": ">3000",
+					"sublime_text": ">4000",
 					"tags": true
 				}
 			]
--- a/repository/f.json
+++ b/repository/f.json
@@ -524,8 +524,8 @@
 		{
 			"name": "File Filter",
 			"details": "https://github.com/vsc-cnst/SublimeTextFileFilter",
-			"author": "vsc",
-			"labels": [ "file filter", "file content filter", "logs filter", "lines filter", "file", "line", "filter", "regex" ],
+			"author": "vsc-cnst",
+			"labels": [ "search", "logs", "utilities", "code navigation" ],
 			"releases": [
 				{
 					"sublime_text": ">4000",
--- a/repository/f.json
+++ b/repository/f.json
@@ -522,6 +522,20 @@
 			]
 		},
 		{
+			"name": "File Filter",
+			"details": "https://github.com/Vasco1/STFileFilter",
+			"readme": "https://github.com/Vasco1/STFileFilter/blob/main/README.md",
+			"issues": "https://github.com/Vasco1/STFileFilter/issues",
+			"author": "vsc",
+			"labels": [ "file content search", "file content query", "filer", "search", "query", "content", "logs", "regex" ],
+			"releases": [
+				{
+					"sublime_text": ">3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "File Rename",
 			"details": "https://github.com/brianlow/FileRename",
 			"labels": ["rename", "file rename"],
@@ -618,21 +632,6 @@
 				{
 					"sublime_text": "*",
 					"tags": true
-				}
-			]
-		},
-		{
-			"name": "File Filter",
-			"details": "https://github.com/Vasco1/STFileFilter",
-			"readme": "https://github.com/Vasco1/STFileFilter/blob/main/README.md",
-			"issues": "https://github.com/Vasco1/STFileFilter/issues",
-			"author": "vsc",
-			"labels": [ "file content search", "file content query", "filer", "search", "query", "content", "logs", "regex" ],
-			"releases": [
-				{
-					"sublime_text": ">3000",
-					"tags": true,
-					"python_versions": [ "3.3", "3.8" ]
 				}
 			]
 		},
--- a/repository/f.json
+++ b/repository/f.json
@@ -523,7 +523,7 @@
 		},
 		{
 			"name": "File Filter",
-			"details": "https://github.com/Vasco1/STFileFilter",
+			"details": "https://github.com/vsc-cnst/SublimeTextFileFilter",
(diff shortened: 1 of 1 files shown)


kaste:
Just to let you know: You really screwed up the PR here, 😉.  You committed files that don't belong here and re-formatted the whole `f.json`.

kaste:
Please opt into Python3.8 by creating a file `.python-version`.  (E.g. https://github.com/ichichikin/sublime-plugin-interactivity/blob/main/.python-version).  Then use the built-in `enum` (the one that ships with py38), and don't change `sys.path`.

Vasco1:
ok. tnks for the nice feedbok and sory for the mess. all done.  let me know if it's all ok

kaste:
Some notes:

- https://github.com/Vasco1/STFileFilter/blob/823a95731de4395e24bc0a41236cc88339741745/file_filter.py#L1-L9 can be removed.  You're not allowed to alter `sys.path` and luckily you don't need to and then this is unused code.

- Usage of `settings-internal-file-filter.sublime-settings` is strange as settings are per definition not "internal" or "private".  They can be overridden by the user and so it is a bit confusing.  You really can make that a global in `file_filter.py`, e.g. 

```py
ADDITIONAL_MENU_ENTRIES = [
    ["prompt", "___prompt___"],
    ["clear", "___clear___"],
]
```

- There should be an "edit" command for the (external) settings.  E.g. just like the
first entry in https://github.com/kaste/InlineOutline/blob/master/InlineOutline.sublime-commands  *But* please use the standard name for the settings file, instead of `settings-external-file-filter.sublime-settings` just `File Filter.sublime-settings`.  (a. these are always "external" in the sense of overwritable by the user; b. these are settings by their file-extension; c. the name mirrors the name of your extension)

- In `https://github.com/Vasco1/STFileFilter/blob/main/file_filter.sublime-commands` don't prefix the commands with your name ("vsc : ").  The standard prefix would be `File Filter: ` (note the spaces) again as that's the name of your extension.  You used "Filter File" that would be okay too but then name your extension also that way.  I personally think "Filter File" is the better wording actually so I would welcome if you would like to rename everything.  It's not too late to do that.  

- In the entry here in this repository, omit the "readme" and "issues" keys as these are set automatically for the github repo.

Happy coding

Vasco1:
i took you notes into consideration (tank you so much for the very welcomed and excelente and feedback)

the changes were
- removed sys.path references
- single settings file, settings managements review; 
- added edit settings command; 
- standardize package name - I've named files in snake_case as i do not feel comfortable with spaces in file names. hope that's not a problem

tnks

kaste:
Holy, you rewrote the package already and also changed your name  😁.  It would be good to go, _but_ you introduced a root logger https://github.com/vsc-cnst/SublimeTextFileFilter/blob/d5293c1f80651f6a2a03abb7bcf8ef7feb392fb1/file_filter.py#L71-L76.  This is a no-go as you're in a shared environment.  E.g. your root logger immediately emits for other packages/plugins:

```
[sublack:utils.py.get_log-77](INFO) Loglevel set to INFO
[INFO][FileFilter][root_logger.plugin_loaded():113]  plugin loaded with settings Settings(123)
[INFO][FileFilter][FileFilter.__init__():136]  Command Has Started
[INFO][FileFilter][FileFilterClearCommand.__init__():136]  Command Has Started
[INFO][FileFilter][FileFilterPromptRegexCommand.__init__():136]  Command Has Started
[INFO][FileFilter][FileFilterQuickPanelCommand.__init__():136]  Command Has Started
[INFO][FileFilter][FileFilterSetFoldingTypeCommand.__init__():136]  Command Has Started
[INFO][FileFilter][FileFilterSetHighlightTypeCommand.__init__():136]  Command Has Started
[WARNING][FileFilter][SublimeLinter.lint.backend.excess_warning():216]  'plugin.py' puts in total 5(!) tasks on the queue:  1x ruff-lsp, 1x codespell, 1x ruff, 1x flake8, 1x mypy.
```

That is an example of what gets logged on startup.  Generally, you can't use `basicConfig` as that installs the root logger.  `LOGGER = logging.getLogger(f'root_logger')` is also suspicious, the *actual* root logger is the one without a name, t.i. `getLogger()`.  Unlikely you want that, instead use your package name, e.g. `getLogger("FileFilter")`.  If you want sub-loggers, you would need to namespace them using dots, e.g. `FileFilter.QuickPanelCommand` etc.  because they really form a hierarchy.  

What you want to do is more like:

```python
LOGGING_LEVEL = logging.WARN
LOGGING_FORMAT = "[%(levelname)3s][FileFilter][%(name)s.%(funcName)s():%(lineno)s]  %(message)s"

_handler = logging.StreamHandler()
_formatter = logging.Formatter(LOGGING_FORMAT)
_handler.setFormatter(_formatter)
LOGGER = logging.getLogger('FileFilter')
LOGGER.setLevel(LOGGING_LEVEL)
LOGGER.addHandler(_handler)
```

Lastly, one tip: since all keybinding files have the same content, you could also just ship one file named `Default.sublime-keymap`, t.i. without specifying a platform in the name.  Then this file is used on all platforms.

vsc-cnst:
Hey,
Yes, i do have as a really old account, and i felt it needed a refresh 😄 

yes, it was a really early version and this is a log of new things for me, but i think it is ready now.
those logs you mention were a mistake, they are for dev only.  I'll set the level to error as i do not yet have any error nor warnings messages 

one last thing, for packages update, every release i make ir automatically update our do i have to submit a PR somewhere? 

one again, thank you for your excellent feedback. I hope this is the one 🚀 😄

kaste:
Once it's on Package Control you just make new tags/releases and they get shipped automatically.  Don't forget to tag the latest changes as `2.0.1` or so.

@braver  This LGTM!

vsc-cnst:
thank so much for your help!

kaste:
@vsc-cnst Don't forget to tag/release the lastest changes!

vsc-cnst:
done ;)

braver:
Nice! Thanks @kaste for helping review this. Only thing I would change it add a `.gitattributes` file that excludes the tests. That way they won't ship to end users (saving a few bytes here and there), and I guess it also makes sure ST doesn't try to load any of those by accident or whatever. Otherwise, looks good :)

repository/f.json
```diff
@@ -521,6 +521,18 @@
 				}
 			]
 		},
+		{
+			"name": "File Filter",
+			"details": "https://github.com/vsc-cnst/SublimeTextFileFilter",
+			"author": "vsc",
+			"labels": [ "file filter", "file content filter", "logs filter", "lines filter", "file", "line", "filter", "regex" ],
```
braver, line 528: The author is fine, but wouldn't you prefer that to be simply your GitHub username?

The labels won't help that might I think... they're just all random combinations of file and filter etc. Reads as an attempt to hack the SEO :D None of them will mean much to users. Just use a few of the regularly used ones, e.g. from this list: https://packagecontrol.io/browse/labels

braver:

Well, it looks like a different approach to https://github.com/davidpeckham/sublime-filterlines. A smarter approach too, so it's all good. But just FYI 😉

vsc-cnst:
Hi,

thank you very much for the positive feedback.

yes, I was trying to optimize my package reachability, don´t blame for that :P
but that's fine, author name changed, labels changed and .gitattributes added

latest version with all updates has been released.

let me know if there is any optimization left to do ;)

tnks

repository/f.json
```diff
@@ -521,6 +521,18 @@
 				}
 			]
 		},
+		{
+			"name": "File Filter",
+			"details": "https://github.com/vsc-cnst/SublimeTextFileFilter",
+			"author": "vsc-cnst",
```
braver, line 527:
```suggestion
```

Just one final thing: if we're going with your GitHub username here, we don't even need it here, it's automatic.

braver (changes requested):

vsc-cnst:
author name removed ;)

PR #8947, opened 2024-07-06
Adds SSHubl

HorlogeSkynet:
- [x] @d3vyce and I are the package's authors and/or maintainers.
- [x] We have have read [the docs][1].
- [x] We have tagged a release with a [semver][2] version number.
- [x] Our package repo has a description and a README describing what it's for and how to use it.
- [x] Our package doesn't add context menu entries. *
- [x] Our package doesn't add key bindings. **
- If our package is a syntax it doesn't also add a color scheme. ***
- If our package is a syntax it is named after the language it supports (without suffixes like "syntax" or "highlighting").
- [x] We use [.gitattributes][3] to exclude files from the package: images, test files, sublime-project/workspace.

Our package is a proper (Open)SSH integration for Sublime Text, inspired by Visual Studio Code's Remote - SSH plugin.
You can mount remote folders through sshfs (FUSE), open or close (reverse) port forwards using commands.
If [Terminus](https://github.com/randy3k/Terminus) is installed, you can even open a remote terminal directly in a sheet.
When re-opening a Sublime project, previous connections are restarted and mounts/forwards are setup back.

Our package may look similar to "SSH-Panel" (#8487), but :

* it doesn't depend on a weird .DLL to download for Windows
* it doesn't depend on libffi for Linux
* it doesn't vendor (!) Paramiko
* it doesn't uses static servers configuration in settings (SSHubl allows you to dynamically connect to whatever host you want)
* it takes advantage of sshfs to mount remote folders
* it allows (reverse) port forwards management to ease development of remote services

---

Many thanks for your time, see you :wave:

[2]: https://semver.org

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -3596,6 +3596,17 @@
 			]
 		},
 		{
+			"name": "SSHubl",
+			"details": "https://github.com/HorlogeSkynet/SSHubl",
+			"labels": ["forward", "remote", "ssh", "terminal"],
+			"releases": [
+				{
+					"sublime_text": ">=4081",
+					"tags": true
+				}
+			]
+		},
+		{
 			"details": "https://github.com/bmc/ST2SyntaxFromFileName",
 			"releases": [
 				{


braver:
Thanks for your submission. Always good to see a modern implementation of what I'm sure is a very useful feature.

I haven't tested your package, but browsed to some of the code and will throw you some things that stand out to me. So far, just minor stuff 👍🏻 

I wonder if, when loading settings at load of the package, your package will see changes made to user's settings later on? You don't have to worry about performance of loading and reading settings, they're served from memory usually. For convenience you can wrap load+get in a method so that settings are loaded every time you want to retrieve a setting. You seem to have settings that require restarting ST if changed? Not sure, but if that's true that should be in the README.

It's a bit common practice to include a command palette entry for changing settings. Makes them even easier to access. 

Maybe we can get some more eyes on this later, but that's what I have for now :)

HorlogeSkynet:
Many thanks for your message and quick feedback, we'll address this ASAP :bow:

HorlogeSkynet:
Hello @braver ! Little up please 🙏

braver:
You know what, LGTM. I bet there is a lot we can talk about on the nitty gritty of the code, but it looks well considered, so let's roll.

PR #8948, opened 2024-07-08
Add Zukan Icon Theme

53v3n3d4:

My package is ...

My package is similar to A File Icon and FileIcons. It is different if we consider it uses mostly original icons. And, also there are file icons that do not have on other packages, as of today.

It is part of my original theme, which I split in 3 parts: theme, color-scheme and file icons.

Sorry for reposting, I deleted my fork thinking that it would not affect here. I was wrong.

[2]: https://semver.org

### DIFF
--- a/repository/z.json
+++ b/repository/z.json
@@ -264,6 +264,17 @@
 			]
 		},
 		{
+			"name": "Zukan Icon Theme",
+			"details": "https://github.com/53v3n3d4/Zukan-Icon-Theme",
+			"labels": ["theme", "file", "icons"],
+			"releases": [
+				{
+					"sublime_text": ">=4075",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Zurb Foundation",
 			"details": "https://github.com/ronydebnath/Sublime-Foundation",
 			"labels": ["automation", "utils"],


braver:
I'm getting an error, even after restarting ST a few times:
```
No such file or directory: '/Users/.../Sublime Text/Packages/Zukan Icon Theme/icons_preferences'
```

braver:
Otherwise, looks like a nice addition :)

53v3n3d4:
Hi Braver,

> I'm getting an error, even after restarting ST a few times:
> 
> ```
> No such file or directory: '/Users/.../Sublime Text/Packages/Zukan Icon Theme/icons_preferences'
> ```

I will check this. I think you test the main repo.

But I will double check and confirm when fix it.

I changed data file because of some settings options that I am still doing, and this folder now needs to be created. Sorry.

Thanks.

braver:
> I think you test the main repo.

Oh, yeah, I just checked out your repo (to the correct directory though). Let me know what you find out, it might have just been my mistake!

53v3n3d4:
Hi Braver,

I checked and fixed the error, so main repo should work. I released a new version with this change.

Thanks.

PR #8949, opened 2024-07-10
Take over Github Tools

braver:
And a slight rename to GitHubTools/

The context menu was there in the original package. How to disable the menu is documented in the README. 

I've had a [pull request](https://github.com/temochka/sublime-text-2-github-tools/pull/41) open on the original repo for months with no communication. The package has been [stale for about 7 years](https://packagecontrol.io/packages/Github%20Tools).

@temochka FYI, this PR is just to keep the package alive and update it again. If you object to the takeover let me know within 2 weeks.

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -814,7 +814,7 @@
 		},
 		{
 			"name": "GitHubTools",
-			"previous_names": ["GitHub Tools"],
+			"previous_names": ["Github Tools"],
 			"details": "https://github.com/braver/GitHubTools",
 			"releases": [
 				{
--- a/repository/g.json
+++ b/repository/g.json
@@ -782,16 +782,6 @@
 			]
 		},
 		{
-			"name": "Github Tools",
-			"details": "https://github.com/temochka/sublime-text-2-github-tools",
-			"releases": [
-				{
-					"sublime_text": "*",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "GithubEmoji",
 			"details": "https://github.com/akatopo/GithubEmoji",
 			"labels": ["emoji", "github", "markdown"],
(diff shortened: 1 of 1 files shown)


temochka:
@braver I do not intend to continue to support Github Tools as I’m no longer a Sublime Text user myself. I don’t mind if the package finds a new home and I’m happy to learn that people still find it useful. With regards to handing off the package control entry, I strongly suggest using a proper GitHub fork instead of a copied repo to avoid any confusion for existing users. Your README should also clearly state that the project maintainers have changed. Once you do that, I will have no objections and I’ll even update my repo’s README with the project status a link to your fork. Thank you!

braver:
@temochka thanks for your reply! We can do all of that, no problem.
What also works is adding me as a maintainer to the existing repo, then we don’t need to change anything here.

temochka:
@braver if I’m not going to be involved in the development (I will not), I’d rather not have it under my name. If you don’t want the repo to appear as a fork then as long as you keep the same license, preserve Git history, and have a note in the readme linking to the original repo, I’m also fine with that.

braver:
@temochka that's cool. The license and the history are not going anywhere. I'll add a link back to the OG in the repo: https://github.com/braver/GitHubTools/blob/master/README.md#credits
We do typically "detach" the fork in these cases, because otherwise you (and especially other new contributors) keep fighting with GitHub to point PR's to the new repo instead of the old one. It's purely a practical consideration, never with the intention to obscure the history of the repo or take ownership of pre-existing work.
So, if you're ok with it as it stands, I'd like to continue this "takeover" with the new repo. I could even add a message to users about the change, but I wasn't really planning on any breaking changes or anything that would otherwise be too visual or interrupting for existing users.

temochka:
@braver alright, then I have no objections about updating the Package Control entry. Feel free to proceed. I’ll update the repo README once this is merged. Good luck!

braver:
Fantastic, thanks!

PR #8953, opened 2024-07-17
Takeover wowtoc

p3lim:

The original package has been dead for 8 years and archived for ~6, and I've been given permission by @AndrielChaoti to take over the name for this new package that provides the same functionality but modernized (to use sublime-syntax) and updated to the most recent spec.

[2]: https://semver.org

### DIFF
--- a/repository/w.json
+++ b/repository/w.json
@@ -902,6 +902,7 @@
 		},
 		{
 			"name": "World of Warcraft TOC",
+			"previous_names": ["World of Warcraft TOC file Syntax"],
 			"details": "https://github.com/p3lim/sublime-toc",
 			"labels": ["language syntax"],
 			"releases": [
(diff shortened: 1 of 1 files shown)


repository/w.json
```diff
@@ -901,13 +901,13 @@
 			]
 		},
 		{
-			"name": "World of Warcraft TOC file Syntax",
-			"details": "https://github.com/AndrielChaoti/wowtoc-syntax",
+			"name": "World of Warcraft TOC",
+			"details": "https://github.com/p3lim/sublime-toc",
```
braver, line 905: Since you're changing the name (good 👍🏻 ) you need to add previous_names:

```suggestion
			"previous_names": ["World of Warcraft TOC file Syntax"],
			"details": "https://github.com/p3lim/sublime-toc",
```

braver (changes requested):
Excellent. Just one minor change to make to the entry here.

braver (approved):

PR #8956, opened 2024-08-06
eloquence syntax

macMikey:
syntax for both eloquence program files and eloquence dialog files
https://eloquence.marxmeier.com/

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -736,7 +736,7 @@
 			]
 		},
 		{
-			"name": "Eloquence Syntax",
+			"name": "Eloquence",
 			"details": "https://github.com/macMikey/eloquence-sublime-text-syntax",
 			"labels": ["eloquence","language syntax"],
 			"releases": [
--- a/repository/e.json
+++ b/repository/e.json
@@ -736,6 +736,17 @@
 			]
 		},
 		{
+			"name": "Eloquence Syntax",
+			"details": "https://github.com/macMikey/eloquence-sublime-text-syntax",
+			"labels": ["eloquence","language syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Eloquent Snippets",
 			"details": "https://github.com/m2quared/Sublime-Eloquent-Snippets",
 			"labels": ["snippets", "laravel", "eloquent"],


braver:
Please use the checklist in the PR template to tell us about your package.

macMikey:
is this what you mean? i did not see the template, so i copied the template from another PR

* [x]  I'm the package's author and/or maintainer.
* [x]  I have have read [the docs](https://packagecontrol.io/docs/submitting_a_package).
* [x]  I have tagged a release with a [semver](https://semver.org) version number.
* [x]  My package repo has a description and a README describing what it's for and how to use it.
* [x]  My package doesn't add context menu entries. *
* [x]  My package doesn't add key bindings. **
* [x]  Any commands are available via the command palette.
* [x]  Preferences and keybindings (if any) are listed in the menu and the command palette, and open in split view.
* [x]  If my package is a syntax it doesn't also add a color scheme. ***
* [x]  If my package is a syntax it is named after the language it supports (without suffixes like "syntax" or "highlighting").
* [ ]  I use [.gitattributes](https://www.git-scm.com/docs/gitattributes#_export_ignore) to exclude files from the package: images, test files, sublime-project/workspace.


there are no extraneous files or directories in the repo, so, there is no need for .gitattributes

braver:
Was this intended to be a syntax file too? https://github.com/macMikey/eloquence-sublime-text-syntax/blob/main/dlg-sublime-syntax.yaml

braver:
In captures, you don't need to put quotes around the indexes. Also please ensure all scopes end with `.eq` (matching "source`.eq`"). Also please double check [the guidelines](https://www.sublimetext.com/docs/scope_naming.html), you might be able to specify some of those keywords. E.g.:

```yml
captures:
        1: keyword.control.eq
        2: entity.name.function.eq
```

repository/e.json
```diff
@@ -735,6 +735,17 @@
 				}
 			]
 		},
+		{
+			"name": "Eloquence Syntax",
```
braver, line 739: > If my package is a syntax it is named after the language it supports (without suffixes like "syntax" or "highlighting").

.... 🧐 


```suggestion
			"name": "Eloquence",
```

braver (changes requested):

macMikey:
> Was this intended to be a syntax file too? https://github.com/macMikey/eloquence-sublime-text-syntax/blob/main/dlg-sublime-syntax.yaml

there are two types of eloquence files: eloquence program files (.eq), and eloquence dialog files (.dlg). they are supposed to be included.

macMikey:
thanks for your patience, i'm a package n00b

braver:
You'll probably want to rename [dlg-sublime-syntax.yaml](https://github.com/macMikey/eloquence-sublime-text-syntax/blob/main/dlg-sublime-syntax.yaml) to `dlg.sublime-syntax` then. 

> thanks for your patience, i'm a package n00b

No worries, that's what we're here for 😉

macMikey:
> You'll probably want to rename [dlg-sublime-syntax.yaml](https://github.com/macMikey/eloquence-sublime-text-syntax/blob/main/dlg-sublime-syntax.yaml) to `dlg.sublime-syntax` then.
> 
> > thanks for your patience, i'm a package n00b
> 
> No worries, that's what we're here for 😉

done

braver:
Alright, main thing then remains that all scopes need to end with the identifier of the syntax, ie. `keyword.dlg` and `keyword.eq`, etc. This makes it possible for package developers to target your specific syntax.

macMikey:
so: in the .dlg syntax file, all of the scopes end in .dlg and in the .eq file, all the scopes end in .eq?

braver:
Yes. Basically your base scope for a syntax is source.XX and then you make sure all the scopes in that syntax also end with .XX

braver:
Almost!

source.dlg.eq should still be source.dlg

Also remember that your captures also assign scopes, so you need to cover those as well.

Have a look at some example syntaxes like https://github.com/sublimehq/Packages/blob/master/CSS/CSS.sublime-syntax#L3216C1-L3224C43

macMikey:
* source.dlg updated
* captures updated

thanks for all the help.

braver:
Looks good!

macMikey:
thanks for all the help.
package-control n00b...out

braver:
🫡

PR #8960, opened 2024-08-14
Update t.json

stuartelimu:
Add Toggle Sidebar and Reveal Current File package

My package is Toggle Sidebar and Reveal Current File

My package is similar to FocusFileOnSidebar. However it should still be added because they work differently

[2]: https://semver.org

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -2756,6 +2756,17 @@
 			]
 		},
 		{
+			"name": "Toggle Sidebar and Reveal Current File",
+			"details": "https://github.com/stuartelimu/toggle-sidebar-and-reveal-file",
+			"labels": ["sidebar", "toggle"],
+			"releases": [
+				{
+					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


braver:

No, you [didn't](https://github.com/stuartelimu/toggle-sidebar-and-reveal-file/tags) 😉

stuartelimu:
> > I have tagged a release with a [semver](https://semver.org/) version number.
> 
> No, you [didn't](https://github.com/stuartelimu/toggle-sidebar-and-reveal-file/tags) 😉

Sorted!🙈 
https://github.com/stuartelimu/toggle-sidebar-and-reveal-file/releases/tag/v1.0.0

braver:
The normal “toggle sidebar” command is also available via the command palette. It would also make sense for your command, I think.

stuartelimu:
Makes sense, I'll add it in the next release. Thank you

PR #8961, opened 2024-08-24
Rename Better CoffeeScript to CoffeeScript

deathaxe:
This PR 

1. renames `Better CoffeeScript` package to `CoffeeScript`, replacing the old ST2-only package.
   The target repository has already been renamed to [SublimeText/CoffeeScript](https://github.com/SublimeText/CoffeeScript).

2. registers a ST4-only (ST4143+) tag prefix as _CoffeeScript Literate_ is planned to be extended from bundled Markdown syntax.

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -581,21 +581,6 @@
 			]
 		},
 		{
-			"name": "Better CoffeeScript",
-			"details": "https://github.com/SublimeText/BetterCoffeeScript",
-			"labels": ["language syntax", "snippets", "linting", "watch", "coffeescript"],
-			"releases": [
-				{
-					"sublime_text": ">=3000",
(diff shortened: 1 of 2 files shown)


FichteFoll (approved):

FichteFoll:
Per our usual procedure, the current maintainer, @sustained, would be allowed a two-week period to object the replacement of their package on the channel. However, because of https://github.com/sustained/CoffeeScript-Sublime-Plugin/pull/63 I assume that there is no interest in maintaining the package (last change was 12 years ago) and replacing the package with a more recent version is appreciated.

I'm also assuming that @Xavura and @sustained are the same person because otherwise the change to simply add a license seems rather questionable.

PR #8962, opened 2024-08-26
Fix typo in CoffeeScript build number

deathaxe:
Fix a typo in build number. Latest releases should target 4143 instead of 3143.


FichteFoll (approved):
Ah well, I only checked the tags part but not the st build selector :sweat_smile:

PR #8964, opened 2024-08-28
Update l.json

JohhannasReyn:
Add Lensort package

My package is ...
    Unique, I haven't found anything that sorts lines by length, which is why I wrote it.

[2]: https://semver.org

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -788,6 +788,17 @@
 			]
 		},
 		{
+    			"name": "Lensort",
+    			"details": "https://github.com/JohhannasReyn/Lensort",
+    			"labels": ["sorting", "lines", "formatting"],
+    			"releases": [
+        			{
+            				"sublime_text": "*",
+            				"tags": true
+        			}
+    			]
+		},
+		{
 			"name": "Less",
 			"details": "https://github.com/SublimeText/Less",
 			"labels": ["language syntax", "auto-complete"],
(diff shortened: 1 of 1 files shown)


braver:
[SortBy](https://packagecontrol.io/packages/SortBy) can "Sort by the length of lines". I'm not against a package that only does sorting by length though, I kind of like that focus actually. 

- You still need to tag a release (e.g. 1.0.0), that's also why the tests fail.
- Your preferences menu still has separate entries for User and Default settings... that went out of vogue ages ago ;) Use the `edit_settings` command there just like you do in sublime-commands.
- Since you ship with example key bindings, it would be a good idea to include keybindings in the preferences menu, and in the command palette.
- In my experience, users don't really like update messages unless they mention something important. If a user just installed your package, a 1.0.0 message might do a lot other than confirmed it worked (which package control also does).
- There should not be a `package-metadata.json` file in your repo. The metadata is in this entry here in package control.

JohhannasReyn:
I've made the corrections you have listed; I believe it's ready to be reviewed again. Thank you.

braver:
@JohhannasReyn please tag a release with your latest work.

JohhannasReyn:
> @JohhannasReyn please tag a release with your latest work.

I believe I completed this already, no?

braver:
Looks like!

PR #8966, opened 2024-08-31
Update Theme - Spacegray

deathaxe:
Spacegray 1.4.0+ is no longer compatible with ST3.

Hence shipping ST3 version via `st3-1.3.5` tag is required.

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -2071,11 +2071,15 @@
 		{
 			"name": "Theme - Spacegray",
 			"details": "https://github.com/SublimeText/Spacegray",
-			"labels": ["theme"],
+			"labels": ["theme", "color scheme"],
 			"releases": [
 				{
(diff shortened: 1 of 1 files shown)


repository/t.json
```diff
@@ -2071,11 +2071,15 @@
...
+			"labels": ["theme", "color scheme"],
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": ">=4000",
 					"tags": true
+				},
+				{
+					"sublime_text": "<4000",
+					"tags": "st3-"
```
braver, line 2082: I may be blind, but I'm not seeing a prefixed release on the repo yet.

deathaxe: Oh, maybe forgot to push.

Just did so, as well as a ST4-only 2.0.0 release.

PR #8970, opened 2024-09-10
Updates p.josn, Added Piranha syntax highlighter.

Kostasn2:

My package is a syntax highlighter for Piranha scripting language library.

[2]: https://semver.org

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -1592,7 +1592,7 @@
 		{
 			"name": "Piranha",
 			"details": "https://github.com/Kostasn2/piranha-syntax-highlighting-package",
-			"labels": ["piranha", "syntax highlight"],
+			"labels": ["piranha", "language syntax"],
 			"releases": [
 				{
 					"sublime_text": "*",
--- a/repository/p.json
+++ b/repository/p.json
@@ -1590,8 +1590,9 @@
 			]
 		},
 		{
-			"name": "Piranha Syntax Hightlighter",
+			"name": "Piranha",
 			"details": "https://github.com/Kostasn2/piranha-syntax-highlighting-package",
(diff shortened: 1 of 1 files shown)


repository/p.json
```diff
@@ -1589,6 +1589,17 @@
 				}
 			]
 		},
+		{
+			"name": "Piranha",
+			"details": "https://github.com/Kostasn2/piranha-syntax-highlighting-package",
+			"labels": ["piranha", "syntax highlight"],
```
braver, line 1595:
```suggestion
			"labels": ["piranha", "language syntax"],
```

"language syntax" is the canonical label for a package like this.

braver (changes requested):
Looks good, just some minor things:

- `source.mr` feels odd to me, I would've expected `source.piranha`? also since your reference that [elsewhere](https://github.com/Kostasn2/piranha-syntax-highlighting-package/blob/main/Piranha.tmPreferences#L7)
- I would recommend adding tests. It's [easy](https://www.sublimetext.com/docs/syntax.html#testing) and makes it a lot easier to make changes later

repository/p.json
Kostasn2: Yea sure, you are right! Just changed it!

Kostasn2:
> Looks good, just some minor things:
> 
> * `source.mr` feels odd to me, I would've expected `source.piranha`? also since your reference that [elsewhere](https://github.com/Kostasn2/piranha-syntax-highlighting-package/blob/main/Piranha.tmPreferences#L7)
> * I would recommend adding tests. It's [easy](https://www.sublimetext.com/docs/syntax.html#testing) and makes it a lot easier to make changes later

Ok ok, i also changed that!

braver (approved):

PR #8972, opened 2024-09-11
Add TestExplorer plugin to `t.json`

IPWright83:

My package is TestExplorer, designed to allow navigation around files containing unit tests. It displays a list of groups and test file names, providing a live navigation drop-down. https://github.com/IPWright83/sublime-TestExplorer

[2]: https://semver.org

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -788,6 +788,16 @@
 			]
 		},
 		{
+			"name": "TestExplorer",
+			"details": "https://github.com/IPWright83/sublime-TestExplorer",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


braver:
I wonder how broadly applicable this is outside the language and test suite you built this for. You might want to specify what that is a bit more clearly in your readme, also because you ship with a pre-configured setup that assumes a testing framework. But otherwise, the package looks good 👍🏻

IPWright83:
Thanks for merging (and the feedback @braver). I did indeed make this currently for use with Jest based tests, but I'm hoping with the regex config that it's easy enough to tweak for other languages. I'll probably try adding support for Go soon locally to give that a try. I might share common configs too in the README going forward.

braver:
Sounds like a plan! 👍🏻

PR #8973, opened 2024-09-12
Initial submission of Render View.

cepthomas:

Renders the text in a view to html.
- Wysiwyg printing in full color: first render to html, then print from the browser.
- Renders markdown also.
- Line numbers optional.
- Supports all scheme colors including companions [Highlight Token](https://github.com/cepthomas/SbotHighlight)
  and [Notr](https://github.com/cepthomas/Notr).

I was unable to find a current plugin that supplies these features.
I did find https://packagecontrol.io/packages/Print%20to%20HTML but it appears unsupported
and doesn't support scopes.

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -1269,6 +1269,18 @@
 			]
 		},
 		{
+			"name": "Render View",
+			"details": "https://github.com/cepthomas/SbotRender",
+			"labels": ["render", "print", "markdown"],
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"platforms": ["windows", "linux"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Renpy Language",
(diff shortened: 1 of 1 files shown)


braver:
Thanks for your submissions. I'll (slowly) get to each one, but let's start with this one. Overall, very nicely done. I have only one remark, being that the package name is not used consistently in files. For for the command names that's entirely fine, users customizing at that level, adding menus and shortcuts, your readme is also very clear. It's just the setting file, "[SbotRender.sublime-settings](https://github.com/cepthomas/SbotRender/blob/main/SbotRender.sublime-settings)" that for most users won't clearly linked to the "Render View" package: that's the name of your package according to package control, and what users see in the menu. Of course it works whatever name you give it, but something to consider.

cepthomas:
Good suggestion - I will update all my projects similarly. Probably best to do this before they go live.

Thanks very much for this work you do.

PR #8975, opened 2024-09-12
Initial submission of Notr.

cepthomas:

Notr is a Sublime Text application plugin for displaying and managing collections of text notes.
It is a plain text format with some similarity to markdown. However there is no rendering-for-publication step: the text itself is the whole point.

Some key features:
- Sections with tags and simple (non-hierarchal) folding.
- References to section, file (image or other), uri.
- Navigation to targets via quick panel. Has MRU and sticky entries.
- Lists with several bullet types.
- Tables with insert/delete column, fit, sort.
- Multiple notr projects.
- Search in all project notr files.
- Works well with [Render View](https://github.com/cepthomas/SbotRender) to generate static output.

I couldn't find anything in current Package Control that did what I wanted because:
- Too simplistic.
- Too complicated, usually with onerous configuration and dependencies.
- Poor code quality -> unmaintainable.
- Abandoned.

A few that I looked at:
- https://packagecontrol.io/packages/Notes
- https://packagecontrol.io/packages/NotePad
- https://packagecontrol.io/packages/PlainNotes
- https://packagecontrol.io/packages/Outline%20Notes%20Publisher

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -1055,6 +1055,18 @@
 			]
 		},
 		{
+			"name": "Notr",
+			"details": "https://github.com/cepthomas/Notr",
+			"labels": ["note", "note taking", "markdown"],
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"platforms": ["windows", "linux"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Nova Framework Snippets",
 			"details": "https://github.com/nova-framework/Nova-Framework-Snippets",
 			"labels": ["snippets"],
--- a/repository/n.json
+++ b/repository/n.json
@@ -1061,7 +1061,7 @@
 			"releases": [
 				{
 					"sublime_text": ">=4000",
-					"platforms": ["windows", "linux"],
+					"platforms": "*",
 					"tags": true
 				}
 			]


braver:
> However there is no rendering-for-publication step: the text itself is the whole point.

Well you could say the same for markdown 😅 But for sure, this is an interesting project! 

We already covered some topics in your other packages, we don't need to go through that again. I have only tiny nit picks or questions:

-  `markup.bold` etc should probably also get the `.notr` suffix?
- `view.syntax().name` is fine, especially since you ship with the syntax and it's unlikely to get an alternative. But normally I'd match on scope (I've always found the "syntax" api interface weird on that point). 🤷🏻

repository/n.json
```diff
@@ -1054,6 +1054,18 @@
...
 			]
 		},
+		{
+			"name": "Notr",
+			"details": "https://github.com/cepthomas/Notr",
+			"labels": ["note", "note taking", "markdown"],
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"platforms": ["windows", "linux"],
```
braver, line 1064: Is this not Mac-compatible now too? If you can't test that's fine, most package devs can't test more than one platform. But if the intention or expectation is that it _could_ work _mostly_, that's good enough 😄

cepthomas: Yes. Done.

cepthomas:
> Don’t forget to tag your latest work!

Yup, just waiting for the merge in case there were other odds and ends.

PR #8976, opened 2024-09-12
Initial submission of Plugin Pdb

cepthomas:

Sublime Text plugin for debugging plugins.

- Uses pdb remotely over a plain-text socket.
- Optional colorizing of pdb output.
- Optional timeout to unblock ST if client doesn't connect.
- Smart client manages the connection to make the edit-run-edit cycle much smoother.

There appears to be only one similar package and it has been dead for 10 years and has no
valid repository: https://packagecontrol.io/packages/Plugin%20Debugger.

There are several remote debugging packages but none specifically for ST plugin development.
I've borrowed liberally from [remote-db](https://github.com/ionelmc/python-remote-pdb).

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -1756,6 +1756,18 @@
 			]
 		},
 		{
+			"name": "Plugin Pdb",
+			"details": "https://github.com/cepthomas/SbotPdb",
+			"labels": ["debug", "plugin", "remote"],
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"platforms": ["windows", "linux"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "PluginLogger",
 			"details": "https://github.com/kapitanluffy/sublime-plugin-logger",
 			"donate": "https://github.com/sponsors/kapitanluffy",


braver:
From reading your readme, this might not work as a Sublime Package. For one, they're shipped as a zip that users can't, or aren't supposed, to open. So 
> Copy sbot_pdb.py to the directory of the plugin you are debugging.

Users would have to copy the file from your GitHub repo or something. It's unclear to me what a workflow would look like. 

Also
> Note that sublime is blocked while running the debugger so you can't edit files using ST.

...  makes me take pause. 

So I think I get it as a utility you can use if you know how. And having it in package control would give it some visibility. But as a package delivered that way, it maybe doesn't fit what users would expect of it. You'd expect the package to be loaded, and then available for interaction using menus, command, etc. Your readme seems to indicate users should set up their own commands? 

Maybe, in short, I don't get it 😅 Maybe it's better distributed as a set of useful scripts that package developers can use, ie outside of package control?

cepthomas:
Yes, I see your point and maybe this doesn't really fit well with the package model. I will cancel this PR. It's still useful (along with a bunch of other utilities I have) - what would be a good way to announce things like this? Forum post? Thanks for your feedback.

cepthomas:
Just to clarify, this tool runs pdb in the ST python environment. When it hits the breakpoint, the ST python is blocked. The client can then interact with it like any other pdb implementation. Since ST is stopped you can't edit anything. It's a minor problem with the edit/breakpoint workflow but the value outweighs the restriction. IMO.

braver:
Sure, get some eyes on it via a forum post, but definitely also the package developer channel in our Discord. 

Maybe this could work as a dependency? Or maybe there is another way to “packagize” it? More debugging tools for de community would definitely be welcome for sure! Folks on the Discord might have ideas there too.

PR #8977, opened 2024-09-12
Initial submission for Signet Bookmarks

cepthomas:

ST's built-in bookmark system is commonly known to be deficient. There are several packages
that claim to improve on it but those I've used tend to be over-featured and/or brittle.
e.g. https://packagecontrol.io/packages/Better%20Bookmarks which is over-complicated
and hasn't been updated in 6 years.

- The basics toggle/next/previous/clear.
- Persisted per ST project.
- Option to traverse just the current file or all files in project.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1729,6 +1729,18 @@
 			]
 		},
 		{
+			"name": "Signet Bookmarks",
+			"details": "https://github.com/cepthomas/SbotSignet",
+			"labels": ["bookmark", "navigation"],
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"platforms": ["windows", "linux"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"details": "https://github.com/321hendrik/SikuliTools",
 			"author": "Hendrik Elsner (321hendrik@gmail.com)",
 			"labels": ["sikuli", "gui", "test", "automation", "app", "web", "image", "capture"],


cepthomas:
It's been almost two months. If there's something I've missed in my submissions (five), please advise. Otherwise carry on.

repository/s.json
```diff
@@ -1728,6 +1728,18 @@
...
 			]
 		},
+		{
+			"name": "Signet Bookmarks",
+			"details": "https://github.com/cepthomas/SbotSignet",
+			"labels": ["bookmark", "navigation"],
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"platforms": ["windows", "linux"],
```
braver, line 1738:
```suggestion
					"platforms": "*",
```
> Linux and OSX should be ok but are minimally tested 


😉

cepthomas:
> I appreciate you didn't add entries to the context menu. Adding a Main menu entry with some child items would be totally fine though. That's less "contested space".

It's so annoying to have a plugin clog up the context menus. There's a certain sidebar plugin...

> > // Scope name for gutter icon color.
> > "scope": "keyword",
> 
> Could be a good use case for the [colorish scopes](http://www.sublimelinter.com/en/master/settings.html#styles-colors)

They are supported already. I have made region.redish the default so should be more obvious.

> I would like to be able to toggle signets etc from the command palette though?

OK., added. I never use the palette so tend to overlook. I don't see its value except for a few things that are clumsy from the context menu and key bindings. But that's me.

braver:
>  There's a certain sidebar plugin...

Hah, wouldn't know what you're talking about there 😉

PR #8978, opened 2024-09-22
Added Crackboard support

nafey:

My package is a client for supporting crackboard.dev which is a leaderboard for programmers. This allows Sublime Text users to compete on crackboard.dev leaderboard.

[2]: https://semver.org

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -4046,6 +4046,16 @@
 			]
 		},
 		{
+			"name": "Crackboard",
+			"details": "https://github.com/nafey/crackboard",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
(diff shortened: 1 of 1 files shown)


braver:
Some remarks:

- Access to the settings file should be possible via the package settings menu.
- Since this adds global event listeners, I would like to see users have some more control over when that runs. You want to be  able to turn that off situationally or per project. The only way right now is to turn the package off or uninstall it, and reload ST. This could be as simple as an on/off boolean in the settings: such settings can be easily toggled using commands, keybindings, and even project settings. 
- Calling the API on every text change might be a bit overkill for what crackboard needs. You might want to look into optimizing, or debouncing that.

braver:
👋🏻 please respond to continue the review!

PR #8982, opened 2024-09-30
re: request add css intellisense

id-sakurata:
Add CSS Intellisense

My package is CSS Intellisense

My package is similar to "IntelliSense for CSS class names in HTML" from vscode plugin. However it should still be added because it can make it easier to develop html without remembering the complete css class. you can use autocomplete from Css Intellisense.

[2]: https://semver.org

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -4410,6 +4410,17 @@
 			]
 		},
 		{
+			"name": "CSS Intellisense",
+			"details": "https://github.com/id-sakurata/CSS-Intellisense",
+			"labels": ["css", "snippets", "auto-complete"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "CSS Less(ish)",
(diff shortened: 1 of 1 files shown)


braver:
There are various important points in that checklist still open. Please ensure you covered all of that and then report back.

braver:
I see you addressed most feedback from the [previous PR](https://github.com/wbond/package_control_channel/pull/8939). It's looking pretty good now, let's just see how it does out in the wild 😉

id-sakurata:
> I see you addressed most feedback from the [previous PR](https://github.com/wbond/package_control_channel/pull/8939). It's looking pretty good now, let's just see how it does out in the wild 😉

I really thank you for accepting my request. I can now recommend this plugin to my work friends 😆.

and now how do i update my plugin, i just found a problem when updating the settings, and now i have fixed the problem in the new update.

braver:
> how do i update my plugin

Any time you create a tag (with a higher version number), that will get pushed to users (typically when they start ST, or when they manually check for updates).

PR #8983, opened 2024-10-01
Drop non-existing Charcoal package

deathaxe:
The package no longer exists and the owning user Dlnd is marked as "joined 2 weeks ago".

Maybe a new actor.


braver:
I believe deletions still aren’t processed properly, so it will probably still appear on the site after we merge this.

PR #8985, opened 2024-10-03
TokensCounter package added

yaroslavyaroslav:

My [package](https://github.com/yaroslavyaroslav/SublimeTokenCounter) is simply counting selected text as token, which is useful for folks that are working with llm. Is uses phantoms to present the calculations.

There's no package alike that, because there were lack of dependencies to perform this operation. Now they are available in PC.

[2]: https://semver.org

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -2900,7 +2900,6 @@
 		{
 			"name": "TokensCounter",
 			"details": "https://github.com/yaroslavyaroslav/SublimeTokenCounter",
-			"issues": "https://github.com/yaroslavyaroslav/SublimeTokenCounter/issues",
 			"donate": "https://github.com/sponsors/yaroslavyaroslav",
 			"labels": ["nlp", "tokens", "llm", "openai"],
 			"releases": [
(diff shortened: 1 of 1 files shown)


yaroslavyaroslav:
Looks like tests failing not by my fault.

repository/t.json
```diff
@@ -2897,6 +2897,19 @@
 				}
 			]
 		},
+		{
+			"name": "TokensCounter",
+			"details": "https://github.com/yaroslavyaroslav/SublimeTokenCounter",
+			"issues": "https://github.com/yaroslavyaroslav/SublimeTokenCounter/issues",
```
braver, line 2903: That's the default URL for issues so no need to add it here, PackageControl will find that automatically.

```suggestion
```

braver (changes requested):
Took me a while to realize what kind of tokens 😅 Package looks good though. Just one minor remark about your entry here.

yaroslavyaroslav:
@braver could you please check it once again, as I applied fixes that have been requested.

PR #8986, opened 2024-10-05
Merge "Syntax Highlighting for Sass" into Sass

deathaxe:
This commit merges former "Syntax Highlighting for Sass" package into "Sass"
as "Sass" is a fork and old ST2 versions are still present.

A st2-1.0.0 tag was created in "Sass" repository to keep providing ST2
compatible packages, even though that's probably not important anymore.

Intention is to cleanup packages and get rid of heavy out-dated packages.


braver:
Oh that one was archived even. Sure, good move 👍🏻

PR #8990, opened 2024-10-13
Migrate SCSS users to Sass

deathaxe:
This commit removes heavily out-dated and unmaintained SCSS package, migrating its users to Sass package.

Alternatively, we could split current Sass package into 2 (Sass and SCSS) and continue providing both as separate dedicated syntax packages.

Not sure what a perfect strategy is, but it appears many users still relying on out-dated SCSS package or even having installed both packages, not being aware they basically intend to provide same value.

SCSS saw its last update 10 years ago and it is time to sunset it.

related with https://github.com/SublimeText/Sass/issues/111

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -175,7 +175,7 @@
 			"name": "Sass",
 			"details": "https://github.com/SublimeText/Sass",
 			"labels": ["language syntax", "scss", "sass", "completions", "snippets"],
-			"previous_names": ["Syntax Highlighting for Sass"],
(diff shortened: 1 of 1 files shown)


deathaxe:
I wonder why review bot says both packages SCSS and Sass are removed ??

braver:
> I wonder why review bot says both packages SCSS and Sass are removed

I wondered the same last time. The ways of the bot are deep and mysterious.

braver:
@MarioRicalde are you ok with replacing your (by now 10 years old) SCSS package with the (actively maintained and up to date) Sass package?

PR #8993, opened 2024-10-24
Takeover SublimeFileBrowser

kaste:
SublimeFileBrowser (forked from dired) is actually good but has bugs and is unmaintained 😢.
I have a few PR's open without any response.  

The goal is to get this going again (but it is almost there) and to move it to python 3.8.
The update to 0.8.5 should be not too disruptive to existing users.  There is just one change in
look and feel: I prefer to use the user's color scheme for the dired view.  I still ship the custom
scheme so users can opt-in or revert to the old look.  This, of course, also closes an issue because
some users would have liked it to behave like that from the start.  I.e. the custom scheme was a
light one, and there was no dark one.    

My package is https://github.com/kaste/SublimeFileBrowser

This is also a fork of the unmaintained dired package.  All commands are named "dired_x".

@aziz and @vovkkk to notice and to comment.  

[2]: https://semver.org

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -579,13 +579,16 @@
 		},
 		{
 			"name": "FileBrowser",
-			"details": "https://github.com/aziz/SublimeFileBrowser",
+			"details": "https://github.com/kaste/SublimeFileBrowser",
 			"labels": ["file navigation", "file creation", "file open", "sidebar"],
 			"releases": [
 				{
-					"sublime_text": "*",
-					"tags": true,
-					"dependencies": ["package_events", "pathtools", "watchdog"]
+					"sublime_text": "<4000",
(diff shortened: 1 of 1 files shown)


FichteFoll:
You should add a note about the primary fork source to your readme (in addition to the now-secondary and deleted one).

I also initially though that the open PRs (of which the oldest is 3 weeks old) haven't been around that long, but I also see that the other PRs haven't seen any replies from a maintainer either in the recent years, not to mention the issues that have been created since then, so it seems quite clear that this repository is unmaintained.

@aziz, as per ususal, you have 2 weeks to object to this replacement.

kaste:
Updated the README and switched to py38 in a new release.

braver:
🎉

PR #8996, opened 2024-10-26
Migrate PHP-Twig users to Twig

deathaxe:
PHP-Twig ...

1. is marked missing for a reasonable time and thus can't be installed.
2. is a tmLanguage based syntax
3. that hasn't seen any updates for about 8 years
4. seems to fail on ST4 according to its issue trackers

This PR therefore proposes to merge PHP-Twig into Twig.

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -1229,16 +1229,6 @@
 			]
 		},
 		{
-			"name": "PHP-Twig",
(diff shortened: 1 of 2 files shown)


braver:
Just to follow proper process here, @uhnomoli are you ok with your package being basically folded into the (well maintained) Twig package?

PR #8997, opened 2024-11-03
Added VHS syntax package

patopesto:

My package is a syntax highlighter for charmbracelet's [VHS](https://github.com/charmbracelet/vhs) tape files

### DIFF
--- a/repository/v.json
+++ b/repository/v.json
@@ -321,7 +321,6 @@
 			"name": "VHS",
 			"details": "https://github.com/patopesto/Sublime-Text-VHS-Syntax",
 			"labels": ["language syntax"],
-			"author": "patopesto",
 			"releases": [
 				{
 					"sublime_text": ">=4107",
--- a/repository/v.json
+++ b/repository/v.json
@@ -318,6 +318,18 @@
 			]
 		},
 		{
+			"name": "VHS",
+			"details": "https://github.com/patopesto/Sublime-Text-VHS-Syntax",
+			"labels": ["language syntax"],
+			"author": "patopesto",
+			"releases": [
+				{
+					"sublime_text": ">=4107",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


patopesto:
The repository actually contains a top level LICENSE file. Not entirely sure how to handle the bot's warning

braver:
That's a bit of a weird test result. Maybe the script just got confused, because your repo does seem to be missing a syntax file? 🧐

patopesto:
Oops terribly sorry. I've now committed the forgotten syntax file.

braver:
Not sure what the bot is on about… anyway

- Your variable.annotation scope is missing the tape suffix
- Perhaps you could specify those keywords a bit?
- In your PR here you don’t have to specify the author, it will pick up your GitHub username automatically

patopesto:
Thanks for the feedback. 
I updated the missing suffix and the tag.
However I'm not sure I totally understand what you mean by "specifying the keywords a bit". The language is very simple and only contains a specific set of commands/keywords which are parsed. Are you suggesting parsing each individual commands?

braver:
> The language is very simple and only contains a specific set of commands/keywords which are parsed

It seemed like you have two sets of keywords, keys and instructions and they might map to one of the [common groups of keywords](https://www.sublimetext.com/docs/scope_naming.html#keyword).. or not, also fine 🙂 

In any case something you can refine later.

repository/v.json
```diff
@@ -317,6 +317,18 @@
 				}
 			]
 		},
+		{
+			"name": "VHS",
+			"details": "https://github.com/patopesto/Sublime-Text-VHS-Syntax",
+			"labels": ["language syntax"],
+			"author": "patopesto",
```
braver, line 324: Since this is just your GitHub username you can omit it. 

```suggestion
```

braver (changes requested):

braver (approved):

PR #5016, opened 2015-11-30
Fixed broken repository url for Base Encoder

rwieckowski:
Configuration is pointing to non existing repository

PR #9001, opened 2024-11-11
Restore Base Encoder package

poke:
This restores the [Base Encoder package](https://packagecontrol.io/packages/Base%20Encoder) which has been marked as “missing“ on the website and has been removed from the channel in #8844.

The original repository is gone but the fork by @socsieng is still available and [the original author agreed](https://github.com/wbond/package_control_channel/pull/5016#issuecomment-2462770647) to have the package point to the fork as part of the discussion in #5016.

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -190,6 +190,16 @@
 			]
 		},
 		{
+			"name": "Base Encoder",
+			"details": "https://github.com/socsieng/base-encoder",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"branch": "master"
+				}
+			]
+		},
+		{
 			"name": "Base Snippets",
 			"details": "https://github.com/base/sublime-text-base-snippets",
 			"labels": ["snippets", "JavaScript", "base", "node", "node-base"],


deathaxe:
https://packagecontrol.io/packages/StringUtilities at least provide base64 decoder and encoder.

I wonder if it would make more sense to add missing base16 and base32 converters to that package instead.

That said code of existing "Base Encoder" looks rather dated.

If we want to re-add it, I'd suggest to also revise the code base.

1. make use of `edit_settings`
2. opt-in to python 3.8 for ST4
3. replace the threaded `set_status()` call by `window.status_message()`
4. maybe drop default key bindings as it is easier to add them than to get rid off.

deathaxe:
Also found https://packagecontrol.io/packages/StringEncode with base64 encoder/decoder.

braver:
"missing" just means "deleted" most of the time, PC's website doesn't handle those very well. Since there are more modern implementations of the same idea already in PC, I don't see a particular reason to bring this one back from the dead?

FichteFoll:
From the discussion in the other issue/PR https://github.com/wbond/package_control_channel/pull/5016, it appears that there are at least 3 people actively communicating that they are using the package and would like to see it reappear.

@poke, @euphwes any comments on @deathaxe's review above?

deathaxe:
FWIW, I have a rewritten "Base Encoder" with all my review comments applied.

I just wonder, whether we should instead takeover also unmaintained StringUtilities package to SublimeText org and add missing base encoder commands from "Base Encoder" package instead of restoring this one.

braver:
@deathaxe that sounds like a great plan 👍🏻

deathaxe:
StringUtilities shares most conversion functions with [StringEncode](https://github.com/colinta/SublimeStringEncode). As I found myself dropping various obsolete functions from StringUtilities I got the impression of StringEncode having been gone the same route a decade before.

As StringEncode also implements an "Paste Encoded..." command to solve the "convert in-place" vs. "insert converted string" challenge without relying on separate commands or package settings and author still seems to be quite active, I decided to create some PRs to push it to python 3.8 and add missing base12 and base16 commands.

deathaxe:
StringEncode 2.7.0 with all base encoding/decoding commands was released.

braver:
So StringUtilities is pretty much [dead](https://github.com/akalongman/sublimetext-stringutilities/pull/51), but it's there and I guess it's not hurting anyone. How about instead of bringing Base Encoder back, we now add that to the "previous_names" for StringEncode? Not sure what that will do for people who still have Base Encoder installed, ie. if that would be annoying, but at least it cleans up the registry (by now I assume deletions will never be handled properly by the PC site). @deathaxe?

As for this PR, it seems the original package has been superseded in various ways, and it doesn't make a lot of sense to bring it back. Also considering that there are dozens of ways of installing packages, and even have them auto-update, without them being in package control. If there are 2 users of the package, or maybe even more 😉, they already have options.

braver:
Since there have been no objections or new insights, I will close this PR.

PR #9002, opened 2024-11-11
Updates p.josn, Added PLI10 syntax highlighter

Kostasn2:
Added pli10 syntax highlighting.

My package is a syntax highlighter for PLI10 pseudo-language.
A greek "programming language", with its own compiler, that is being teached to EAP (Elliniko Anoixto Panepisthmio, Hellenic Open Uniersity) in Greece, to the students for them to learn fundamentals of programming, programming logic and basic algorithmic structures.

[2]: https://semver.org

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -1716,6 +1716,17 @@
 			]
 		},
 		{
+			"name": "PLI10",
+			"details": "https://github.com/Kostasn2/pli10_eap-syntax-highlighting-package",
+			"labels": ["pli10", "language syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "plist",
 			"details": "https://bitbucket.org/fschwehn/sublime_plist",
 			"labels": ["language syntax"],


Kostasn2:
Its a week now, should i make any changes? Do we have any update about my package?

Kostasn2:
Hello, its almost a month now, is my package ok? Is anything i should fix?
Any update i should make?

Thanks in advance!

Kostasn2:
@FichteFoll , Sorry for the ping and the taf, i just saw you are one of the collaborators.
Its almost a month now (3 weeks), and i have no update about my package.

Sorry for the tag, i appreciate all the effort you put in.

Thanks in advance

braver:
This is a community project. Sometimes we're pretty quick, sometimes other things in life have more priority. Pinging people doesn't help usually, we get to it when we get to it. 

Two remarks:
- you might want to use a [look ahead here](https://github.com/Kostasn2/pli10_eap-syntax-highlighting-package/blob/main/PLI10.sublime-syntax#L139-L146) so that the brackets aren't captured when all you want to do is pop out of the context.
- since your snippets are only active in the scope of your syntax, I would personally drop the`_pli10` suffix from the triggers.

Kostasn2:
> This is a community project. Sometimes we're pretty quick, sometimes other things in life have more priority. Pinging people doesn't help usually, we get to it when we get to it.
> 
> Two remarks:
> 
> * you might want to use a [look ahead here](https://github.com/Kostasn2/pli10_eap-syntax-highlighting-package/blob/main/PLI10.sublime-syntax#L139-L146) so that the brackets aren't captured when all you want to do is pop out of the context.
> * since your snippets are only active in the scope of your syntax, I would personally drop the`_pli10` suffix from the triggers.

I fixed the `_pli10` suffix, and i got a little confused about what you said for the brackets. 
The meaning is that when there is a bracket that is "alone" i want to be marked as invalid. And if there are brackets inside brackets then if the user tries to enters ')' for example while theρε is already a bracket like this, i want to pop "like going to the next one outside" until there is nothing else, and then it will be placed a ')' but it will be invalid if there is nothing to be paired. 
For example:

image

Inside the '{}' if i enter the '}' it will see that there is a already a closing curly bracket so the cursor will go to the next closing bracket.

And again sorry for the previous πινγ, i understand there are priorities in life of caurse, its all my fault, iam really sorry for rushing things out. I understad that you are doing an amazing job and yes you are right having us pinging you, make things even worse.
It wont happen again.

Thanks again!

Have a nivce month!

braver:
Excellent, no problem. Let's get this package out there then :)

PR #9005, opened 2024-11-13
Add MarkdownCodeExporter

maliayas:

My package is [MarkdownCodeExporter](https://github.com/maliayas/SublimeText_MarkdownCodeExporter)

[2]: https://semver.org

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -608,7 +608,7 @@
 		},
 		{
 			"name": "MarkdownCodeExporter",
-			"details": "https://github.com/maliayas/SublimeText_MarkdownCodeExporter",
+			"details": "https://github.com/SublimeText-Markdown/MarkdownCodeExporter",
 			"labels": ["markdown"],
 			"releases": [
 				{
--- a/repository/m.json
+++ b/repository/m.json
@@ -607,6 +607,17 @@
 			]
 		},
 		{
+			"name": "MarkdownCodeExporter",
+			"details": "https://github.com/maliayas/SublimeText_MarkdownCodeExporter",
+			"labels": ["markdown"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "MarkdownComplements",
 			"details": "https://github.com/junShimoji/MarkdownComplements",
 			"releases": [


maliayas:
Thanks for the valuable review!

> You're using an event listener so we need to worry about performance. Your `on_modified` handler for instance is synchronous, that looks like a risk.

I gave it a try and used async version. But this time long texts caused a lag. I'm not an expert on async stuff. Would you like to test it on your side, such that replace "on_modified" with "on_modified_async" and create a md document which have 20 times of this code block:

    ```
    lorem ipsum
    ```

Then try to insert any char to anywhere in the document. On my ST version (3125) unacceptable lags occur with the async version. Non-async function works with almost zero lag. I can only notice a small lag when the document has more than 500 code blocks.

maliayas:
> Finding out if the current file is markdown via the view's syntax is tricky (as you seemingly noticed) due to all the different syntax packages for markdown. It might be easier to check the file's scope, it should be text.html.markdown regardless of the package used.

Good idea but there is a limitation with the ST API. In the `is_applicable()` method I don't have access to the `view` object, so I can't do a `if sublime.score_selector(self.view.scope_name(0), "text.html.markdown") == 0` check there. I only have `settings` object at that point, which gives me only the active syntax file's path.

Instead I can move that check to `on_modified()` like below:

```diff
--- main.py
+++ main.py
@@ -5,16 +5,16 @@
 from html import escape
 
 class MarkdownCodeExporter(sublime_plugin.ViewEventListener):
-    @classmethod
-    def is_applicable(cls, settings):
-        return settings.get('syntax', '').lower().find('markdown') != -1
-
     def __init__(self, view):
         super().__init__(view)
         self.phantom_set = sublime.PhantomSet(view)
         self.update_phantoms()
 
     def on_modified(self):
+        # Limit the event listener to Markdown.
+        if sublime.score_selector(self.view.scope_name(0), "text.html.markdown") == 0:
+            return
+
         self.update_phantoms()
 
     def on_load(self):
```

But this time I loose the performance benefit of `is_applicable()` which is only run upon settings change (e.g. syntax changed) and then cached. On the other hand if I do the check in `on_modified()` every non-markdown file gets triggered on every modification which is a performance penalty even if the trigger does little job.

I think we can safely assume that almost any Markdown syntax file has a "markdown" phrase in its file name. Thus keep the existing approach.

What is your opinion?

braver:
> Good idea but there is a limitation with the ST API

Oh, you're right of course. Well, at least we considered it :) Your existing approach is certainly just as valid.

> On my ST version (3125) unacceptable lags occur with the async version

Hm, indeed. Since you seem to re-render all phantoms on every update, when doing it async will make the entire document jump around. I wonder if you could leave any unchanged phantoms alone.... It's been a while since I went deep into phantoms, but I don't think I ever had that issue with ColorHints. I do notice your PhantomSet doesn't have a key, that might help?

By the way, something else. I noticed the border radius on the buttons is not equal. It looks like the tops are getting clipped.

Scherm­afbeelding 2024-12-01 om 14 57 26

braver:
Excellent, have fun developing your package!

PR #9011, opened 2024-11-27
Update d.json

JohhannasReyn:
Added DuplExpellio to the plugin repository

- [x] I have tagged a release with a [server][2] version number.

My package is ...

There are no packages like it in Package Control.  <--- This one

[2]: https://semver.org

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -1830,6 +1830,16 @@
 			]
 		},
 		{
+			"name": "DuplExpellio",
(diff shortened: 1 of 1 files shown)


braver:
Thanks for your submission. Some remarks:

- If I understand correctly, the behavior is toggled via a setting. Wouldn't it be easier if there were two commands: one to select all duplicates, one to delete them? 
- I also don't fully understand the use case yet. What is considered a duplicate? A character, a word, a line? The threshold setting is also a bit mystifying to me.

PR #9012, opened 2024-11-29
Add DuplExpellio package to Package Control Channel

JohhannasReyn:

My package is unique.

[2]: https://semver.org

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -1830,6 +1830,16 @@
 			]
 		},
 		{
+			"name": "DuplExpellio",
+			"details": "https://github.com/JohhannasReyn/DuplExpellio",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Duplicate And Increment",
 			"details": "https://github.com/112KA/duplicate_and_increment",
 			"releases": [
--- a/repository/d.json
+++ b/repository/d.json
@@ -1830,16 +1830,6 @@
 			]
 		},
 		{
-			"name": "DuplExpellio",
-			"details": "https://github.com/JohhannasReyn/DuplExpellio",
-			"releases": [
-				{
-					"sublime_text": "*",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "Duplicate And Increment",
 			"details": "https://github.com/112KA/duplicate_and_increment",
 			"releases": [
--- a/repository/m.json
+++ b/repository/m.json
@@ -576,6 +576,17 @@
 			]
 		},
 		{
+			"name": "Markdown2Dir",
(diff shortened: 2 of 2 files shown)


braver:
Thanks for your submission. Some remarks:

- If I understand correctly, the behavior is toggled via a setting. Wouldn't it be easier if there were two commands: one to select all duplicates, one to delete them?
- I also don't fully understand the use case yet. What is considered a duplicate? A character, a word, a line? The threshold setting is also a bit mystifying to me.

JohhannasReyn:
By duplicate I mean a duplicate entry in a delimited file, whether it's delimited by a new line (targeting duplicate lines), by comma, by semi-colon, or by a space (targeting duplicate words) the delimiter can be specified by the user in the settings as well as how many of them it takes to qualify as a delimiter so that the plugin can make its way down the list until a delimiting string or character satisfies the criterium (i.e. to not accidentally use commas for a semi-colon separated file.)

I like your suggestion of making this into two separate commands, I will implement this now and create a new pull request, if you could reject this one, assuming it's not too late, as I have encountered an error with my latest release (I apologize) I noticed just in the last hour that the plugin is failing to work whilst in the zipped .sublime-package form, and is only functional as an unzipped package. I know where I messed up, and will have it corrected and should have it followed with a new pull request in the next couple hours. Again, I apologize for my oversight. Please don't fire me, haha.

braver:
> Please don't fire me, haha.

🤣

Sure, no worries. 

Thanks for the explanation, I can definitely see use cases for that!

The change in behavior when your code gets packaged is one we all run into every now and then. Let me know, in this PR, when you’ve completed the changes you wanted to make. We’ll merge ASAP then.

braver:
> By duplicate I mean a duplicate entry in a delimited file

I also think you might want to expand on that a bit in the readme, perhaps add some examples. Because if I open any random file and then run your command, I still don't fully understand what to expect.

JohhannasReyn:
Thank you so much for your feedback, I will make those adjustments as soon
as I have time. I've already began working on another plugin, this one way
more complex than the last, and your feedback helped clear some things up
for me. I very much appreciate it! So, just so there's no confusion, if my
package fails the automated plugin checker, and I fix the issue inside the
repo, I can request another review without failing again due to a version
tag not reflecting the latest changes made? I'm not sure if that makes
sense, I've not worked with repo's until now, and maybe confused on how the
process works, but an issue I (thought, but just as well could've been
something entirely different) was having was after made an adjustment to
the package, if that change made to the repo wasn't reflected in the latest
version tag, it would get flagged and labeled invalid by the checker when
an automated review was requested, which is why I had been creating new
pull requests. My bad 🥴🫣

braver:
Package Control and our tools for this review only see your latest tag. So if you’re hoping for a code change to help pass the automated test, you do need to create a new tag first (and then request a new review from the bot).

braver:
Hm, I think that was not your intended effect?

PR #9013, opened 2024-11-30
adds codemp plugin package

ghyatzo:

My package is the plugin for the codemp library for remote collaborative editing.

[2]: https://semver.org

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2241,7 +2241,6 @@
 			]
 		},
 		{
-		{
 			"name": "CodePresenter",
 			"details": "https://github.com/fwph/codepresenter",
 			"labels":["presentation"],
--- a/repository/c.json
+++ b/repository/c.json
@@ -2230,6 +2230,18 @@
 			]
 		},
 		{
+			"name": "Codemp",
+			"details": "https://github.com/hexedtech/codemp-sublime",
+			"labels": ["collaborative", "crdt", "editing"],
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"tags": true
+				}
+			]
+		},
+		{
+		{
 			"name": "CodePresenter",
 			"details": "https://github.com/fwph/codepresenter",
 			"labels":["presentation"],


ghyatzo:
@packagecontrol-bot

braver:
Thanks for your submission. I'm not going to test through the entirety of your package 😅  Overall it looks well considered. Just some general remarks:

- The settings and keybindings should be available via the Package Settings menu, not just via the command palette (although the entries there are definitely appreciated).
- Your [code](https://github.com/hexedtech/codemp-sublime/blob/dev/Codemp.sublime-commands#L12) assumes the package resides in a `CodempClient` directory, but since the package is named [Codemp](https://github.com/wbond/package_control_channel/pull/9013/files#diff-b5eeeee20c6cd05e78128c3c469949445ba673f10ad80329baf3241b1002c27fR2233) here, that will also be its directory name, as well as its name in [package control](https://github.com/hexedtech/codemp-sublime/blob/dev/README.md?plain=1#L27). Note that when manually installed, or indeed when developing the package, that should also be the directory name (ie. a simple git clone in to Packages will break references to the settings file etc.).

ghyatzo:
Ah good catch regarding the CodempClient name. It was the name when I created the package initially but forgot to update it.

Thanks for the review, much appreciated. I'll fix the naming issue And add entries in the package settings menu entry.

One small question, the package depends on a library `codemp` (lowercase, already registered).
I assumed there won't be conflict issues given they are in two separate registries, can you confirm this please?

braver:
> I assumed there won't be conflict issues given they are in two separate registries, can you confirm this please?

That's a good point actually, I didn't think about that. They're in different registries, but all packages, dependency or not, end up in the same namespace. I'm not sure we can rely on case-sensitivity to help here, perhaps for clarity's sake we shouldn't. "Codemp Client" is not a bad name for this package either?

ghyatzo:
Ok thanks! I'll think a bit about it. I'll also consider to change the library name instead.

ghyatzo:
Hey sorry for the delay.
I've decided to keep `Codemp` and I'll change the library entry instead.
So you can go ahead for now!
I'll make a PR to change the name of the library.

PR #9014, opened 2024-12-01
Add DuplExpellio plugin v1.0.1 to Package Control channel.

JohhannasReyn:

My package is ... one of a kind!

There are no packages like it in Package Control. <-- Yup, what he said.

[2]: https://semver.org

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -1830,6 +1830,16 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


braver:
You don't need to create a new PR when you make changes to your package. Let's just continue [here](https://github.com/wbond/package_control_channel/pull/9012).

PR #9016, opened 2024-12-05
Add "FindInFiles-addon"

kaste:
- [x] My package *does* add key bindings with specific contexts.
- [x] There are no new commands.
- [x] There are no preferences (yet?)

My package is FindInFiles-addon.  https://github.com/kaste/FindInFiles-addon

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -971,6 +971,17 @@
 			]
 		},
 		{
+			"name": "FindInFiles-addon",
+			"details": "https://github.com/kaste/FindInFiles-addon",
+			"labels": ["find", "add-on"],
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


braver:
Wow, nice work. As usual 😉 This looks really useful! 

I have basically just one remark, being that F5 isn’t normally a “refresh” shortcut on MacOS. In Sublime F5 sorts lines. In most apps, browsers etc. CMD+R refreshes the page. Most users will understand why F5 was chosen, but it’s not as natural and discoverable as it would be on Win/Linux.

kaste:
I added `primary+r`.  😼

braver:
👍🏻

PR #9018, opened 2024-12-10
Adds the Memorize package

ckm2k1:

My package is a visual memory aid designed to help explore call stacks in code.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -1303,7 +1303,7 @@
 				{
 					"sublime_text": ">=4180",
 					"tags": true,
-					"platforms": ["osx", "linux"]
+					"platforms": ["osx", "linux", "windows"]
 				}
 			]
 		},
--- a/repository/m.json
+++ b/repository/m.json
@@ -1285,6 +1285,18 @@
 			]
 		},
 		{
+			"name": "Memorize",
+			"details": "https://github.com/ckm2k1/sublime_memorize",
+			"labels": ["call stack", "memory aid"],
+			"releases": [
+				{
+					"sublime_text": ">=4180",
+					"tags": true,
+					"platforms": ["osx", "linux"]
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


kaste:
LGTM from me.  👍   You should improve your README though!  

Btw, I don't see it tightly coupled to the idea of call stacks, as it remembers and visualizes any file loc.  E.g.

image

Maybe there should be a way to re-order these bits, and remove some frames.  I also wish the stack-view would be visible while navigating around.  (Maybe an output panel at the bottom would be a good fit?)  But these are just ideas...

ckm2k1:
Thanks @kaste! All of your suggestions are already on my todo list :)
A better README is coming shortly!

kaste:
You restrict this to Mac/Linux, but the current version runs on Windows too.  (I'm on Windows.)  Maybe change that?

ckm2k1:
> You restrict this to Mac/Linux, but the current version runs on Windows too. (I'm on Windows.) Maybe change that?

I didn't have a Windows machine to test this on so I chose to be conservative about claiming to support it, but given you ran it successfully, I'll tag it as supporting.

repository/m.json
```diff
@@ -1295,6 +1295,18 @@
...
 		},
+		{
+			"name": "Memorize",
+			"details": "https://github.com/ckm2k1/sublime_memorize",
+			"labels": ["call stack", "memory aid"],
+			"releases": [
+				{
+					"sublime_text": ">=4180",
+					"tags": true,
+					"platforms": ["osx", "linux", "windows"]
```
braver, line 1306: Since you now support all known platforms, just remove that line 😉

ckm2k1: Done

braver:
One last thing (sorry, should have mentioned that yesterday): if you add a main menu entry for the keybindings, that makes them easier to find.

ckm2k1:
@braver - done, added a menu entry for keybindings

PR #9019, opened 2024-12-11
Add LCOV plugin to default channel.

thraex-aquator:

My package is showing covered/missed lines based on an lcov.info file.

[2]: https://semver.org

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -676,7 +676,6 @@
 		{
 			"name": "LCOV",
 			"details": "https://gitlab.com/aquator/sublime-lcov",
-			"readme": "https://gitlab.com/aquator/sublime-lcov/-/raw/master/README.md",
 			"labels": ["lcov", "code coverage"],
 			"releases": [
 				{
--- a/repository/l.json
+++ b/repository/l.json
@@ -680,7 +680,7 @@
 			"labels": ["lcov", "code coverage"],
 			"releases": [
 				{
-					"sublime_text": ">=3000",
+					"sublime_text": ">=4000",
 					"tags": true
 				}
 			]
--- a/repository/l.json
+++ b/repository/l.json
@@ -674,6 +674,18 @@
 			]
 		},
 		{
+			"name": "LCOV",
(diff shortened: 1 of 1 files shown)


kaste:
LGTM from here.  I think it requires Sublime Text `>=4000` though as you target (rightfully) python 3.8.  Note that you don't need to instantiate `Listener()`; Sublime does it for you automatically.

thraex-aquator:
Thanks for the heads-up! Bumped the version to 4000, also removed the extra initialization and reduced the scope of the observer while I was already there. Moved the 1.0.0 tag to the current head as it was never really released before.

kaste (approved):

repository/l.json
```diff
@@ -673,6 +673,18 @@
 				}
 			]
 		},
+		{
+			"name": "LCOV",
+			"details": "https://gitlab.com/aquator/sublime-lcov",
+			"readme": "https://gitlab.com/aquator/sublime-lcov/-/raw/master/README.md",
```
braver, line 679:
```suggestion
```

Since that's just the default readme location, you can omit it.

braver:
Ugh, our tools don't want to run on GitLab repos. It looks good enough to just run with it and see what happens though 👍🏻

repository/l.json
thraex-aquator: Thanks, did not know that, I copied from another gitlab based plugin.
Removed the readme line.

thraex-aquator:
Do I need to do anything else regarding being on gitlab, or you will handle it? (The suggested change is there)

braver:
Ok, let’s see if it works. Maybe we need to add some metadata here, but we just need to check in a couple of hours.

kaste:
✋ packagecontrol.io doesn't pick it up

braver:
@kaste thanks reporting that. It seems to be a more widespread issue than just this package. I have no idea what might cause it. Of 5516 packages, 6 are on GitLab, half of which are missing. 🤷🏻 

Will look into it: https://github.com/wbond/package_control_channel/issues/9020

PR #9021, opened 2024-12-18
Update Ngx HTML package

deathaxe:
1. ship ST4 syntax as of first stable release 4107.
2. add a way point for releases, requiring ST 4152.

~~requires https://github.com/princemaple/ngx-html-syntax/issues/35~~

EDIT: required tag was created.

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -624,16 +624,19 @@
 		},
 		{
 			"name": "Ngx HTML",
-			"previous_names": ["Angular2 HTML Syntax"],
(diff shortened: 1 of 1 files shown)


repository/n.json
```diff
@@ -624,16 +624,19 @@
 		},
 		{
 			"name": "Ngx HTML",
-			"previous_names": ["Angular2 HTML Syntax"],
```
braver, line 627: Is this removal intentional?

princemaple: Happy for it to be removed.

PR #9023, opened 2024-12-21
feat: add SettingSwitcher

halfmoonvic:

My package is [SettingSwitcher](https://github.com/halfmoonvic/SettingSwitcher)

[2]: https://semver.org

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1255,6 +1255,16 @@
 			]
 		},
 		{
+			"name": "SettingSwitcher",
+			"details": "https://github.com/halfmoonvic/SettingSwitcher",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "SetWindowTitle",
 			"details": "https://github.com/gwenzek/SublimeSetWindowTitle",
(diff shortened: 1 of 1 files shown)


braver:
Ooh, that's interesting. 

I wonder if you  need to worry about `find_resource` potentially not returning your package's file. I think it might search across all available resources and it's a pretty generic name.. Haven't tested this in a while. Perhaps your base commands.json file could just exist as a python constant instead of a JSON file? 

Just cause I'm curious I kind of wonder if it would be possible to add custom key bindings to switch to a given settings group?

halfmoonvic:
> Ooh, that's interesting.
> 
> I wonder if you need to worry about `find_resource` potentially not returning your package's file. I think it might search across all available resources and it's a pretty generic name.. Haven't tested this in a while. Perhaps your base commands.json file could just exist as a python constant instead of a JSON file?
> 
> Just cause I'm curious I kind of wonder if it would be possible for a user to add custom key bindings to switch to a given settings group?

Command.json seems really unnecessary, and I have removed it. Custom commands can be added to shortcuts, etc. I added instructions in README. Because it is a user-defined commands, there is no need to add default shortcuts.

PR #9027, opened 2024-12-31
Update Nunjucks Releases to Fix Syntax Loading Error

alsolovyev:
### Changes:
- update the `releases` field to include version-specific details

### Notes:
Please confirm if I’ve configured everything correctly: all new releases are for version >4188, while tags: "2.0.1" remains for <=4188. This is necessary to resolve the [error](https://github.com/alsolovyev/Nunjucks/issues/27):

> Error loading syntax file "Packages/Nunjucks/Syntaxes/Nunjucks.sublime-syntax": Apparent recursion within a with_prototype action: 25000 context sanity limit hit.

### Details:

[2]: https://semver.org

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -1325,12 +1325,12 @@
 			"previous_names": ["Nunjucks Syntax"],
 			"releases": [
 				{
-					"sublime_text": ">4188",
-					"tags": true
+					"sublime_text": "<=4180",
+					"tags": "st3-"
 				},
 				{
-					"sublime_text": "<=4188",
-					"tags": "2.0.1"
+					"sublime_text": ">=4186",
+					"tags": true
 				}
 			]
 		},
--- a/repository/n.json
+++ b/repository/n.json
@@ -1325,8 +1325,12 @@
 			"previous_names": ["Nunjucks Syntax"],
 			"releases": [
 				{
-					"sublime_text": ">=3092",
+					"sublime_text": ">4188",
 					"tags": true
+				},
+				{
+					"sublime_text": "<=4188",
+					"tags": "2.0.1"
 				}
 			]
 		},


repository/n.json
```diff
@@ -1325,8 +1325,12 @@
 			"previous_names": ["Nunjucks Syntax"],
 			"releases": [
 				{
-					"sublime_text": ">=3092",
+					"sublime_text": ">4188",
 					"tags": true
+				},
+				{
+					"sublime_text": "<=4188",
+					"tags": "2.0.1"
```
braver, line 1333: I believe valid values for "tags" is either `true` or a prefix string, ie. not specific versions. The typical way to handle this situation is to tag ref `2.0.1` also with, e.g. `st3-2.0.1`, and then to specify `st3-` here. This allows you to create new tags to fix issues in that branch of your project, without having to open new PR's. 
See also https://raw.githubusercontent.com/wbond/package_control/master/example-repository.json and search for `The "tags" key`.

braver (changes requested):

repository/n.json
alsolovyev: thank you for the fast response and detailed explanation! I appreciate your guidance.

I’ve updated the tags as per your suggestion - please check the new version. Additionally, I wanted to ask: is it possible to rename the 2.0.2 tag to st3-2.0.2, or should i create a new one with the correct name?

braver: > I’ve updated the tags as per your suggestion - please check the new version

Looks good 👍🏻 

> is it possible to rename the 2.0.2 tag to st3-2.0.2, or should i create a new one with the correct name?

Renaming doesn't really exist in git. But you can add as many new tags to any version as you like, and you can remove those you don't need anymore.

One thing to think about though is that Package Control will already have seen your 2.0.3 tag, and will have delivered to users already (assuming most of them have used ST in the days since that tag was created). And it will not actually go and deliver versions with a lower number to any of those users. So, even once we merge this PR, versions < 2.0.3, prefixed or not, will probably not go anywhere. I'm not 100% on this and how this works out in practice.. but I would advise to make sure that the version you want to have delivered to st3 users is tagged with a high enough number so that it's going to get rolled out to all users. Just think of the prefixed versions as a "branch" of releases for your package: 2.0.2 without the prefix does not need to be the same as 2.0.2 _with_ the prefix.

PR #9028, opened 2025-01-02
Add TextTree syntax highlighter

mdoering:

My package is ...

[2]: https://semver.org

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -999,6 +999,7 @@
 		{
 			"name": "TextTree",
 			"details": "https://github.com/gbif/text-tree-sublime",
+			"labels": ["language syntax"],
 			"releases": [
 				{
 					"sublime_text": "*",
--- a/repository/t.json
+++ b/repository/t.json
@@ -997,6 +997,16 @@
 			]
 		},
 		{
+			"name": "TextTree",
+			"details": "https://github.com/gbif/text-tree-sublime",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Theme - Adaptify",
 			"details": "https://github.com/lodev09/Adaptify",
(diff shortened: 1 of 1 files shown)


braver:
Your syntax uses uncommon scope names (please refer to [the guide](https://www.sublimetext.com/docs/scope_naming.html)). I'm sure that with some effort you can find scope names that will allow users to use their preferred color scheme instead of the one your package ships with. Users care very deeply about color schemes, and you can't expect them to use yours just because of your syntax.

braver:
That's better. Please don't force your color scheme on users though. Because you set it in the language specific settings it will be hard for users to change, and it will definitely be unexpected for most. Consider for instance [toml](https://github.com/jasonwilliams/sublime_toml_highlighting) or [apache conf](https://github.com/colinta/ApacheConf.tmLanguage), all are more data formats than truly language, yet don't need or even ship with custom color schemes.

repository/t.json
```diff
@@ -996,6 +996,16 @@
 				}
 			]
 		},
+		{
+			"name": "TextTree",
+			"details": "https://github.com/gbif/text-tree-sublime",
+			"releases": [
```
braver, line 1002: Even though this is not a "language" in the traditional sense it would still make sense to add the "[language syntax](https://packagecontrol.io/browse/labels/language%20syntax)" label.

mdoering: Would it be ok to ship with the color scheme, but remove the default use of it in `TextTree.sublime-settings`?

braver: Sure, that’s totally fine.

mdoering: Remove in https://github.com/gbif/text-tree-sublime/releases/tag/1.0.3

braver: Don’t forget to add the label hete though!

mdoering: added

PR #9029, opened 2025-01-02
Add co-author to existing package (Update e.json)

axemonk:
Added co-author @michaelblyons 

My package is an existing package adding support for syntax highlighting for Excel formulas (with aspirational support for GSheets and others).

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -1530,6 +1530,7 @@
 		{
 			"name": "Excel formula",
 			"details": "https://github.com/axemonk/Excel-formula",
+			"author": ["axemonk", "michaelblyons"],
 			"releases": [
 				{
 					"sublime_text": "*",


michaelblyons:
I endorse this. I'm not sure if that was clear.

axemonk:
thanks for the heads up @michaelblyons. Done.

axemonk:
Update: eventually we will want to rename this package to be "Spreadsheet formula" instead of "Excel formula", with "Excel formula" being added to the `"previous_names"` key instead. Since that will change the proper position of this package in the Package Control repository according to alphabetization, I'm not sure what the best way to go about that is. Please let me know if you have any specific guidance. Thanks.

braver:
> Since that will change the proper position of this package in the Package Control repository according to alphabetization

A package's name is its unique identifier. If you rename a package, move the entry to the new alphabetical position (ie. delete the old one, create a new one). The "previous_names" key will ensure package control knows to treat it as the same package as before but under a different name. 

> should I close this pull request before reopening it with both the original change and the move/rename committed instead?

Yes please.

PR #9031, opened 2025-01-05
Add Claudette

barryceelen:

My package is https://github.com/barryceelen/Claudette

There are no packages specific to interacting with the Claude API like it in Package Control.

[2]: https://semver.org

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1335,6 +1335,7 @@
 		{
 			"name": "Claudette",
 			"details": "https://github.com/barryceelen/Claudette",
+			"labels": ["ai", "claude"],
 			"releases": [
 				{
 					"sublime_text": ">=4075",
--- a/repository/c.json
+++ b/repository/c.json
@@ -1333,6 +1333,16 @@
 			]
 		},
 		{
+			"name": "Claudette",
+			"details": "https://github.com/barryceelen/Claudette",
+			"releases": [
+				{
+					"sublime_text": ">=4075",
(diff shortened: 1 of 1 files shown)


braver:
- Your menu entry for settings don't specify a "user" file. It's also recommended to expose the settings via the command palette.
- As always for packages like this, the readme must be clear and unambiguous as to how and when code might leave the user's machine (if that happens at all, not sure in this particular case): that might not always be obvious or intuitive, and might be a problem in certain scenarios.
- We don't have recommended labels for these kinds of packages, but you might want to add [ai](https://packagecontrol.io/browse/labels/ai).
- The package name is entirely up to you, but you could consider something more explicitly linked to Claude. "Claude Client" for instance would be less clever, but more immediately understood, and might make more people find your package more easily. ...  maybe it says more about me than about your package, but I didn't immediately get it 😅

braver:
> but I might be misunderstanding your request

Sorry, my fault. The edit_settings command takes both a base_file and a user_file argument, and I was unsure if the user_file (or a default value) were required... but it's not, so just ignore me 😄

> Appreciate your feedback and concerns, let's keep it as-is.

Sure! With the added labels and a good repo description you should be good.

barryceelen:
> The edit_settings command takes both a base_file and a user_file argument

Gotcha, yup after looking into this I saw that other packages were passing a default value, added that as well for good measure.

Thanks again @braver for your time!

PR #9032, opened 2025-01-07
Add NinetyFive

juaoose:

My package is ...

My package is similar to other code assistants, however, it:
1. Provides a free mode.
2. Uses a different backend which we're optimizing for speed.

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -753,7 +753,7 @@
 		{
 			"name": "NinetyFive",
 			"details": "https://github.com/ninetyfive-gg/sublime-plugin",
-			"labels": ["completions"],
+			"labels": ["completions", "ai"],
 			"releases": [
 				{
 					"sublime_text": ">=4000",
--- a/repository/n.json
+++ b/repository/n.json
@@ -751,6 +751,17 @@
 			]
 		},
 		{
+			"name": "NinetyFive",
+			"details": "https://github.com/ninetyfive-gg/sublime-plugin",
+			"labels": ["completions"],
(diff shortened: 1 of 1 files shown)


juaoose:
Any feedback is appreciated

braver:
- Your settings aren't exposed in the Package Settings menu, nor in the command palette.
- As always for packages like this, the readme must be clear and unambiguous as to how and when code might leave the user's machine (for training or otherwise). Ideally the users control this from project to project and explicitly opt-in to that.

repository/n.json
```diff
@@ -750,6 +750,17 @@
 				}
 			]
 		},
+		{
+			"name": "NinetyFive",
+			"details": "https://github.com/ninetyfive-gg/sublime-plugin",
+			"labels": ["completions"],
```
braver, line 756: ["ai"](https://packagecontrol.io/browse/labels/ai) is becoming popular as a label for packages like this.

braver:
Be careful that you don't run into potential case sensitivity issues with your package name. Especially in file paths like those [in your main menu file](https://packagecontrol.io/browse/labels/ai). But also elsewhere you're mixing Ninety**F**ive and Ninety**f**ive.

juaoose:
> Be careful that you don't run into potential case sensitivity issues with your package name. Especially in file paths like those [in your main menu file](https://packagecontrol.io/browse/labels/ai). But also elsewhere you're mixing Ninety**F**ive and Ninety**f**ive.

Thanks for catching this! I've made the updates.

PR #9033, opened 2025-01-10
Update PowerShell package

deathaxe:
PowerShell v4.0 requires ST4 as it uses sublime-syntax "version 2".

This commit therefore introduces a release based on `st2-` tags for ST2 and ST3, continuing with ST4-only releases without prefixed tags.

@michaelblyons

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -2140,12 +2140,12 @@
 			"labels": ["language syntax"],
 			"releases": [
 				{
-					"sublime_text": "<3156",
-					"tags": "version/st3155/"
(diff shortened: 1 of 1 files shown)


michaelblyons:
Okay. I liked how the tag prefixes folded in SM, but I know nobody else does it that way. :sweat_smile:

deathaxe:
There are only 2 "tag-branches", effectively. So that folding doesn't help too much, does it? Once merged, I'd even suggest to remove those noisy ones ;-)

PR #9034, opened 2025-01-11
Update m.json with plugin Markdown2Dir

JohhannasReyn:
Add Markdown2Dir to Package Control

My package is ... is unique in the way it not only converts markdown to files and vise versa, but also generates a directory tree in the generated markdown to see what files have been copied into the markdown. Also, the plugin has a fully customizable configuration allowing you to filter by explicit names, directories, extensions, and partial file names.

[2]: https://semver.org

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -576,6 +576,17 @@
 			]
 		},
 		{
+			"name": "Markdown2Dir",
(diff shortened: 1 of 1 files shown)


braver:
To keep your PR's separate, make 2 different branches in your fork of this repo, and then make a PR for each branch.

PR #9043, opened 2025-01-24
Add Ollama Plugin

dapepe:
- [x ] I'm the package's author and/or maintainer.

My package is adding Ollama support inside Sublime text.

My package is similar to Yollama. However it should still be added because it has more commands, e.g. let's me switch models on the fly, adds a palette of prompts, shows my prompt history, etc.

### DIFF
--- a/repository/o.json
+++ b/repository/o.json
@@ -301,6 +301,17 @@
 			]
 		},
 		{
+			"name": "OllamaSublime",
+			"details": "https://github.com/dapepe/OllamaSublime",
+			"labels": ["ollama","llm","ai"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "OmniDocs",
 			"details": "https://github.com/bordaigorl/sublime-omnidocs",
 			"labels": ["search", "documentation"],


braver:
Can't those features be contributed to the existing Yollama package?

dapepe:
Hi @braver ! No, the Yollama package has a very different feature set. OllamaSublime focuses on prompting and switching between different models - mainly for text files. Yollama focuses very much on code completion and code assisting. Different use case.

braver:
Alright, please then:

- check the first point in [the docs about naming your package](https://packagecontrol.io/docs/submitting_a_package#Step_2), perhaps you could simply name your package "Ollama"?
- reread the notes about key bindings in the [PR template](https://raw.githubusercontent.com/wbond/package_control_channel/refs/heads/master/.github/PULL_REQUEST_TEMPLATE.md): it's more user friendly to provide examples for the keybindings than hardcoding them in your package, especially if there is no context specified. 

I wonder what the bot is tripping over 🤔

braver:
It seems to fail on the download, but you do already have a proper semver version tagged. Sadly no other output that's of any use... could it be the dependencies file? Perhaps try that without the schema?

dapepe:
@braver , I will close this merge request and open another pull request with the updated package name.

PR #9045, opened 2025-01-28
Update m.json

JohhannasReyn:
Added Markdown2Dir to repository m.json

My package is ...

[2]: https://semver.org

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -576,6 +576,17 @@
 			]
 		},
 		{
+			"name": "Markdown2Dir",
+			"details": "https://github.com/JohhannasReyn/Markdown2Dir_and_Dir2Md",
+			"labels": ["markdown", "code extraction", "directory management"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "MarkdownAssistant",
 			"details": "https://github.com/code-reaper08/MarkdownAssistant",
 			"labels": ["markdown","snippets","snippet"],


braver:
- Please remove [[sublime-package.json](https://github.com/JohhannasReyn/Markdown2Dir_and_Dir2Md/blob/main/sublime-package.json)], your package metadata is now here in this PR.
- You should opt in to python 3.8 ([example](https://github.com/braver/SideBarTools/blob/master/.python-version))
- When I drop your package into the Packages dir it doesn't load. How did you work on and test the package? Any python files in the root directory of a package are immediately loaded by Sublime. I looks like you should move things like markdown_base_command.py into a subdirectory.
- I've never seen someone use `context` that way to enable or disable menu items. That's [not documented](https://www.sublimetext.com/docs/menus.html) either and should not work. I believe you might have made the assumption it works like keybindings do, although I don't believe you can [access settings](https://www.sublimetext.com/docs/key_bindings.html#context-key) that way there either.
- Your directions for manually installing the package are a bit convoluted. It's much easier for users to simply clone or download your repo into the Package directory. Leave the "installed packages" directory for package control.

Don't take this the wrong way, but I see a lot of mistakes in here that are common for AI generated code. Did you test and verify all the features of your package?

braver:

JohhannasReyn:
I’m so sorry for not responding sooner, I have been meaning to make the
needed corrections you had mentioned but I’ve been so busy with work I
haven’t had the time yet. I will try getting these corrections made
within the next few days, I very much appreciate you taking the time to
help me with this, and also your patience with me. Please don’t fire me! 😅

Very Respectfully,
John Reyn

braver:
We can always re-open the PR when you're ready.

PR #9046, opened 2025-01-28
Add OpenInCursor

spencerchristensen:

(The final four checkboxes are non-applicable to my package)

My package is a way to quickly open your current Sublime file + project in the Cursor editor. I love Sublime Text as a text editor and purposely choose not to turn it into an IDE or AI-powered experience. However, sometimes I do want to quickly hop into Cursor to hash out a quick problem, task, etc.

### DIFF
--- a/repository/o.json
+++ b/repository/o.json
@@ -497,6 +497,18 @@
 			]
 		},
 		{
+			"name": "Open In Cursor",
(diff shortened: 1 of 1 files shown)


braver:
Nice one, my favorite kind of package 😄

PR #9048, opened 2025-02-02
Ollama Support

dapepe:

My package is adding Ollama support inside Sublime text.

### DIFF
--- a/repository/o.json
+++ b/repository/o.json
@@ -300,6 +300,18 @@
 				}
 			]
 		},
+
+		{
+			"name": "Ollama Integration",
+			"details": "https://github.com/dapepe/OllamaSublime",
+			"labels": ["AI", "Ollama", "LLM", "GPT"],
+			"releases": [
+				{
+					"sublime_text": ">4000",
+					"tags": true
+				}
+			]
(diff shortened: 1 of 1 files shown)


braver:
Thanks! Now the bot’s also happy 😊 

Please have a look at this for what I tried to explain about providing commented-out key bindings: 
- the bindings themselves https://github.com/ckm2k1/sublime_memorize/blob/master/Default.sublime-keymap
- making them easy to find via the menu: https://github.com/ckm2k1/sublime_memorize/blob/master/Main.sublime-menu#L12

repository/o.json
```diff
@@ -300,6 +300,17 @@
 				}
 			]
 		},
+		{
+			"name": "Ollama Integration",
+			"details": "https://github.com/dapepe/OllamaSublime",
+			"labels": ["AI", "Ollama", "LLM", "GPT"],
```
braver, line 306: This is really minor, but the convention is mostly for the labels to be lowercase. I don’t think they’re case sensitive, but you know, since we’re here I did want to mention that.

braver:

PR #9051, opened 2025-02-04
feat: adding nekonight color scheme

bgcicca:

My package is an a beautiful and inspiring color scheme with the potential to become a theme pack with regular updates.

As it is a color scheme package, it is expected that there will be packages similar to this one but without the same colors...

[2]: https://semver.org

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -237,6 +237,17 @@
 			]
 		},
 		{
+			"name": "neko night",
+			"details": "https://github.com/neko-night/sublime",
+			"labels": ["nekonight", "dark theme", "theme", "color scheme"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Nelua",
 			"details": "https://github.com/edubart/nelua-sublime",
 			"labels": ["nelua", "language syntax"],


BrunoCiccarino:
I'm going to add installation instructions to the original repository so that the pull request meets the necessary standards to be accepted

BrunoCiccarino:
ready

braver:
The convention is to put “color scheme” in the name of the package. Makes them easier to find. That’s not a hard requirement per se though.

You still need to tag the latest version of your package.

If you’re targeting Sublime Text specifically, you’ll want to convert it to the sublime-color-scheme format. That uses variables, and is much easier to edit and maintain. A tool is available in ST to do the conversion for you.

bgcicca:
> The convention is to put “color scheme” in the name of the package. Makes them easier to find. That’s not a hard requirement per se though.
> 
> You still need to tag the latest version of your package.
> 
> If you’re targeting Sublime Text specifically, you’ll want to convert it to the sublime-color-scheme format. That uses variables, and is much easier to edit and maintain. A tool is available in ST to do the conversion for you.

Ohh, ok, I'm on the street, but when I get home I'll switch to the sublime colorscheme format and refactor the pr so I don't get this lint error anymore, thanks for the feedback

bgcicca:
Sorry for the delay, it's taking me a while because I'm moving house and everything is a mess, but I'll fix it as soon as I have some time left.

braver:
No problem, thanks for letting me know

bgcicca:
I'm trying to convert to spoken format but I can't even find a guide explaining +- how to do it, can you give me some direction please?

braver:
Look for this command:

Screenshot 2025-02-12 at 21 15 13

It doesn't seem to work on all your color schemes, you might want to check if the files are all correctly formatted. And perhaps by using the variables in a smart way you can use one to base all the others on?

PR #9055, opened 2025-02-05
Add MoonBit

hyrious:

My package is a sublime-syntax definition for highlighting the [MoonBit](https://github.com/moonbitlang/moonbit-docs) language.

[2]: https://semver.org

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -2399,6 +2399,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


braver:
Nice!

Note that your screenshot doesn’t always render, guess it depends a bit on the environment. You might want to take a simpler approach there.

PR #9056, opened 2025-02-06
Add AskPerplexity

spencerchristensen:

Last 4 don't apply to this.

Ask a question to [Perplexity](https://perplexity.ai/) and have it answered in your default browser (not Sublime). Facilitate a seamless transition between Sublime and Perplexity while maintaining a clear and defined boundary: no AI in your text editor.

Workflow:

- lovely coding in Sublime
- hey, I have a question to ask AI
- run "Ask Perplexity" command
- type question
- submit, opening a browser window with Perplexity

Makes it easier to context switch directly from Sublime to Perplexity. Doesn't insert Perplexity output into Sublime itself.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -1935,6 +1935,17 @@
 			]
 		},
 		{
+			"name": "AskPerplexity",
+			"details": "https://github.com/spencerchristensen/sublime-ask-perplexity",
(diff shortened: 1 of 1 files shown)


braver:
Makes sense to me! 👍🏻

PR #9058, opened 2025-02-06
New ST4 tag series for SSH Config

michaelblyons:

My package is already in PC. This just allows a new supported version series.

[2]: https://semver.org

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -3656,7 +3656,7 @@
 					"tags": "st3-"
 				},
 				{
-					"sublime_text": ">=4074",
+					"sublime_text": ">=4075",
 					"tags": "st4-"
 				}
 			]
--- a/repository/s.json
+++ b/repository/s.json
@@ -3652,8 +3652,12 @@
 					"tags": "st3.0-"
 				},
 				{
-					"sublime_text": ">=3156",
+					"sublime_text": "3156 - 4074",
 					"tags": "st3-"
+				},
+				{
+					"sublime_text": ">=4074",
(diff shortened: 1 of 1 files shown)


keith-hall (approved):

michaelblyons:
Ah, phooey. I put in the wrong version.

repository/s.json
```diff
@@ -3652,8 +3652,12 @@
 					"tags": "st3.0-"
 				},
 				{
-					"sublime_text": ">=3156",
+					"sublime_text": "3156 - 4074",
```
deathaxe, line 3655:
```suggestion
					"sublime_text": "3156 - 4073",
```

michaelblyons: Haha. Wrong one. The ST4 one has to be 4075. I'll fix it.

repository/s.json
```diff
@@ -3652,8 +3652,12 @@
...
 					"tags": "st3-"
+				},
+				{
+					"sublime_text": ">=4074",
```
michaelblyons, line 3659:
```suggestion
					"sublime_text": ">=4075",
```

deathaxe:
I'd probably even start with first official stable release 4107

FichteFoll:
There is currently no tag with a `st4-` prefix on this repo, meaning installation would fail for the specified build range until it is pushed. I suggest to push a tag ahead of time since its presence doesn't break anything.

michaelblyons:
Tag pushed.

PR #9059, opened 2025-02-08
New ST4 tag series for HTML (C#)

michaelblyons:
    - (Will in the future)

This is just a tag series update for an existing package.

My package is unique in its support for ASP.NET WebForms.

[2]: https://semver.org

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -1079,8 +1079,12 @@
 			"labels": ["language syntax", "completions"],
 			"releases": [
 				{
-					"sublime_text": ">=3153",
-					"tags": true
+					"sublime_text": "3153 - 4133",
+					"tags": "version/st3153/"
+				},
+				{
+					"sublime_text": ">=4134",
+					"tags": "version/st4134/"
 				}
 			]
 		},


braver:
First time I’ve seen “namespacing” in tags like that, are you sure that’s valid semver?

deathaxe:
Tag prefix can be arbitrary text, so it's technically ok.

Using `/` is interesting idea as it allows folding tags in Sublime Merge.

IMHO, `version/` part is however redundant. I'd go with `st4153/` or even just `4153/`

braver:
I thought you’d still need an hyphen (or a plus) to separate the prefix from the actual version? Didn’t make a full study of it but that’s what I got from the spec.

deathaxe:
Not required. A given tag is just stripped by given tag-prefix. It can be anything.

braver:
👍🏻

michaelblyons:
> IMHO, `version/` part is however redundant. I'd go with `st4153/` or even just `4153/`

🤷 I use other top-level tag namespaces for things other than versions.

PR #9060, opened 2025-02-09
Rename Laravel Blade and fix tag prefix

deathaxe:
This PR...

1. fixes tag prefix of ST4 releases (needs to be `4143-` instead of `st4143-`)
2. finally renames package to _Laravel Blade_ again.

The overall reason for recent PRs not being successful was the wrong tag prefix. 
So crawler wasn't able to pick up valid releases for ST4.

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -220,19 +220,10 @@
 			]
 		},
 		{
-			"name": "Laravel Blade AutoComplete",
-			"details": "https://github.com/ahmedash95/sublime-laravel-blade-autocomplete",
-			"releases": [
-				{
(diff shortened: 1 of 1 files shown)


repository/l.json
```diff
@@ -244,7 +235,17 @@
 				},
 				{
 					"sublime_text": ">=4143",
-					"tags": "st4143-"
+					"tags": "4143-"
```
deathaxe, line 238: This was the culprit.

deathaxe:
Appears package is currently marked "missing" due to invalid tag prefixes.

PR #9061, opened 2025-02-11
Fix: Correct indentation and alphabetical order for AutoTagMate entry…

baslie:
… in a.json

My package is AutoTagMate, a unique Sublime Text plugin that automatically generates closing tags in plain text and .txt files. There are no packages like it in Package Control.

[2]: https://semver.org

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -2621,6 +2621,16 @@
 			]
 		},
 		{
+			"name": "AutoTagMate",
+			"details": "https://github.com/baslie/AutoTagMate",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Autotools",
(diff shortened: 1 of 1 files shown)


baslie:
@packagecontrol-bot

I've updated AutoTagMate to version 1.0.2. This release fixes the invalid JSON error in the auto_tag_mate.sublime-keymap file (it was empty previously). The keymap file now contains valid JSON, which should resolve the issue reported by packagecontrol-bot.

Thanks for reviewing!

baslie:
Hi @braver,

Thank you for your detailed feedback! I have implemented your suggestions:

- Default settings are now defined directly in the auto_tag_mate.sublime-settings file rather than being set programmatically.
- The plugin now uses scope matching via view.match_selector instead of relying on file extensions.
- I’ve added commands to expose the plugin settings both through the Command Palette and the main menu.

Thanks again for helping!

baslie:
Package was updated. Thx!

PR #9063, opened 2025-02-16
Drop and replace WakaTime

deathaxe:
This commit therefore removes WakaTime and replaces existing entry by a common "Banned" package name on packagecontrol.io to vanish all its presence and force removal from all clients, which still have it installed.

This drastic step is a response of WakaTime having repetitively attacked packagecontrol.io by insane high (400 per second) fake install stat requests even though it hasn't been installed by anybody for years or within attacks.

#### Stats from Feb/16/2025

grafik

grafik

related with: https://github.com/wakatime/sublime-wakatime/issues/118

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -146,23 +146,23 @@
 			]
 		},
 		{
-			"name": "Banned",
-			"details": "https://github.com/SublimeText/Banned",
-			"previous_names": ["WakaTime"],
-			"releases": [
-				{
-					"sublime_text": "*",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "Bang Search",
 			"details": "https://github.com/bsoun/bang-search",
 			"labels": ["search"],
 			"releases": [
 				{
 					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
+			"name": "Banned",
+			"details": "https://github.com/SublimeText/Banned",
+			"previous_names": ["WakaTime"],
+			"releases": [
+				{
+					"sublime_text": "*",
 					"tags": true
 				}
 			]
--- a/repository/b.json
+++ b/repository/b.json
@@ -146,6 +146,17 @@
 			]
 		},
 		{
+			"name": "Banned",
+			"details": "https://github.com/SublimeText/Banned",
+			"previous_names": ["WakaTime"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Bang Search",
 			"details": "https://github.com/bsoun/bang-search",
 			"labels": ["search"],
--- a/repository/w.json
+++ b/repository/w.json
@@ -35,17 +35,6 @@
 			]
 		},
 		{
-			"name": "WakaTime",
-			"details": "https://github.com/wakatime/sublime-wakatime",
-			"labels": ["time tracking", "coding", "utilities", "productivity"],
-			"releases": [
-				{
-					"sublime_text": "*",
-					"tags": true
-				}
-			]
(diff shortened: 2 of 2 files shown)


alanhamlett:
I'm the creator of this package, and we're (anyone working for WakaTime) not the ones attacking package control. I don't think banning the package is the right response... people still do use it. We have about 10k people using the package per year according to our [public stats](https://wakatime.com/api/v1/stats/2024).

Is there some way to block those requests without banning the package? Banning our package because someone is attacking package control just means they will start using another package instead for their requests.

alanhamlett:
WakaTime definitely doesn't have 600k installs per day even for our most popular editor (vscode). I noticed the requests are all for Mac, 603k in one day. The Linux and Windows numbers could be legit but the Mac count is definitely not. Even if this was counting upgrades, we haven't updated the package recently and we would probably only see about 10k updates max. This should be pretty easy to block without banning our package?

Screenshot 2025-02-16 at 10 38 12 AM

deathaxe:
> Is there some way to block those requests without banning the package? 

I am not aware of any.

> Banning our package because someone is attacking package control just means they will start using another package instead for their requests.

WakaTime has been the only one showing such suspicious behavior so far:

1. Not any install over years until update v11.1.1, when this behavior started.
2. Not even any install stat between assumed attacks.
3. Insane high pseudo install counts (way higher than Package Control itself).

It is currently, unclear what thread actors are attempting:
a) possible DDOS 
b) pushing WakaTime stats.
c) just a large scale scheduled deployment _(rather not)_

alanhamlett:
> 1. Not any install over years until update v11.1.1, when this behavior started.

v11.1.1 was a simple [bugfix change](https://github.com/wakatime/sublime-wakatime/commit/9a1d1e447c1502d6fcb52d2e0cd22e2133bf7ace) I don't think there's any way it could have triggered mass installs/updates unless there's a bug in package control itself

Is the reason for this ban because of fake stats or high server load. If it's high server load, just block the IP making the requests. If it's because of fake stats, just delete all the install stats for Feb 16th for the WakaTime package.

I want to re-iterate: **nobody affiliated with WakaTime make those requests and we definitely shouldn't be banned for it**

deathaxe:
As packagecontrol.io only counts installs from unique IPs, thus blocking a single IP won't work. Hence it rather looks like multiple actors - maybe a botnet being involved.

The interesting part is install counts having been constantly at 10 -> 20 -> 40k per day over a long period, while recently switching to 200k / 400k and now 600k installs for single days.

IMHO, one reason for recent service downtimes.

Well, any action beyound removing misbehaving packages can be taken by service owner anyway. At this point community can't do anything beyond waiting for next outage.

alanhamlett:
Sounds like package control needs DDoS protection. I'm all about data quality (WakaTime's a data product) so it would also be nice to delete these fake installs. Who's the service owner, @wbond?

braver:
Let's for now close this PR. Whatever happened WakaTime is unlikely to have cause it themselves, and removing it won't solve anything. I'll open a new issue for maybe hopefully @wbond to look into the situation and correct the install count for this package.

deathaxe:
FWIW, an update on WakaTime incidences.

Just 1M hits on Apr 30 - a new record.

grafik

alanhamlett:
Damn that's a lot. I bet it's someone who doesn't like WakaTime trying to get us banned, or they would have mixed up the target plugin if they were only trying to DDoS packagecontrol?

deathaxe:
Hard to say. It could also be a side effect of something going wrong internally. But pretty obvious is, in the past when not reacting on it, we saw those traffic for multiple days.

It therefore looks like a bad actor is probing packagecontrol.io

PR #9067, opened 2025-02-28
Add Dhall

kukimik:

My package is named [Dhall](https://github.com/kukimik/dhall-sublime-syntax-highlighting). It is a syntax highlighting package for the [Dhall configuration language](https://dhall-lang.org).

It replaces the original [Dhall](https://packagecontrol.io/packages/Dhall) package that was removed from GitHub in 2020. The present package is based on the original `.sublime-syntax` file, republished with the original author's permission.

This is a second attempt at adding this package, see [#8681](https://github.com/wbond/package_control_channel/pull/8681) for the first one, which failed because back then I only had a `.tmLanguage` syntax definition.

[2]: https://semver.org

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -817,6 +817,17 @@
 			]
 		},
 		{
+			"name": "Dhall",
+			"details": "https://github.com/kukimik/dhall-sublime-syntax-highlighting",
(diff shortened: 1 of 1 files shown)


braver:
Excellent, 10 points, no remarks 😉

PR #9069, opened 2025-03-07
Adds GoogleSQL Syntax Highlighting

pratik-m:
- [x ] I'm the package's author and/or maintainer.
- [ x] I have have read [the docs][1].
- [ x] I have tagged a release with a [semver][2] version number.
- [ x] My package repo has a description and a README describing what it's for and how to use it.
- [ x] If my package is a syntax it doesn't also add a color scheme. ***
- [ x] If my package is a syntax it is named after the language it supports (without suffixes like "syntax" or "highlighting").

My package is GoogleSQL Syntax Highlighting

[2]: https://semver.org

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -1591,6 +1591,18 @@
 			]
 		},
 		{
+			"name": "GoogleSQL Syntax Highlighting",
+			"details": "https://github.com/pratik-m/sublime-googlesql-syntax",
+			"author": "Pratik Munot",
+			"labels": ["syntax", "GoogleSQL", "BigQuery"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "main"
+				}
+			]
+		},
+		{
 			"name": "GoOracle",
 			"details": "https://github.com/waigani/GoOracle",
 			"labels": ["go"],


repository/g.json
```diff
@@ -1590,6 +1590,18 @@
 				}
 			]
 		},
+		{
+			"name": "GoogleSQL Syntax Highlighting",
```
braver, line 1594:
```suggestion
			"name": "GoogleSQL",
```

> If my package is a syntax it is named after the language it supports (without suffixes like "syntax" or "highlighting").

repository/g.json
```diff
@@ -1590,6 +1590,18 @@
...
+			"details": "https://github.com/pratik-m/sublime-googlesql-syntax",
+			"author": "Pratik Munot",
+			"labels": ["syntax", "GoogleSQL", "BigQuery"],
```
braver, line 1597:
```suggestion
			"labels": ["language syntax", "GoogleSQL", "BigQuery"],
```

https://packagecontrol.io/browse/labels/language%20syntax

repository/g.json
```diff
@@ -1590,6 +1590,18 @@
...
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "main"
```
braver, line 1601: Branch based releases are ancient history.

Also

.. you didn't actually.

braver (changes requested):

repository/g.json
pratik-m: Added tags and created a release with semver version number.

pratik-m:
Thanks @braver ! I've made the changes based on your feedback!

braver:
Excellent, thanks 👍🏻

braver:
I have some remarks about the syntax itself, will get back to you with that!

braver:
- Since you register your syntax to handle `.sql` files, you'll probably interfere with the existing SQL language package. It's better to leave that more generic extension alone, focusing on a specific extension for your syntax (hopefully there is one) and then tell users how to switch if they do want to use GoogleSQL syntax for all sql files.
- Similarly you use the `source.sql` main scope and `.sql` suffix on each scope. Since your syntax if probably very much _like_ SQL but not _exactly_ SQL, it's better to have a suffix of your own. An example would be Twig, which is mostly just html, but with its own specifics on top, so it uses the `text.html.twig` scope (ie `.twig` on top of `text.html`). That way selectors targeting HTML would match, but the specific Twig language things can also be targeted individually. You might not want that if GoogleSQL is too different from SQL, in which case `source.googlesql` might be better (ie. a completely different suffix).
- I wonder if you could extend the default SQL syntax and add GoogleSQL-specific adjustments on top? That might be easier to maintain over time. I do say that without knowing anything about GoogleSQL though 😅

pratik-m:
Hi @braver 
As you mentioned, GoogleSQL is very similar to standard SQL, with a few additional keywords and syntax variations. While Google refers to the SQL dialect used on their platform as "GoogleSQL," I don't believe there's a dedicated extension specifically for it.
- I’ve created a syntax file to support the GoogleSQL-specific syntax, using standard SQL as the base.
- I’m not sure introducing a separate file extension would be helpful, since most users will likely continue saving their files with the .sql extension.
- Could you guide me on how to update the default SQL syntax to incorporate the GoogleSQL additions? I tried looking through Sublime’s GitHub repository but couldn’t find a clear approach.

That said, I still think it might be worth keeping the GoogleSQL syntax highlighter separate, so we have flexibility to accommodate any future changes.

Would love to hear your thoughts or suggestions.

braver:
A reason to not specify the .sql extension is to make sure you don't "steal" it.. but then again, If someone is installing your package, that's likely to be what they want it to do. It might not work anyway, I'm not 100% how Sublime decides which syntax to apply in that case where there are multiple for same extension... it might pick the first match, or it might pick the last match. Anyway, that's fine by me either way.

So, the idea would be to introduce your new syntax alongside the existing SQL syntax. But you can use that syntax as a basis (ie. extend it) so you only have to add what's different in GoogleSQL: https://www.sublimetext.com/docs/syntax.html#inheritance

pratik-m:
Hi @braver - I think I can extend the default SQL syntax and override the elements for BigQuery. I've started working on it and will share the updates in few days. I'll retain the extension to '.sql'..
Thanks!

pratik-m:
Hi @braver  - As per your suggestion, I've extended the delivered templates and have overriden the tags specific to BigQuery. 
- I've simplified the yaml as well by consolidating all the functions into one, instead of having categories for them. 
- Added a three GoogleSQL related syntax  
 Can you pls review it and let me know your thoughts? 
 
 Thanks Pratik

braver:
Nice job!

pratik-m:
Hi @braver - My package is not coming up on https://packagecontrol.io/ 
I tried to some digging around but couldn't pinpoint the issue. Can you help me point to the issue and I'll get it fixed? 
Thanks!

braver:
It’s probably just the index not updating properly, we’ve been having intermittent issues with that for a while. Nothing you or I can do something about, people who can are already working on it. Just give it some time 🤞🏻

pratik-m:
Thanks for the update @braver. I'll keep an eye out!

PR #9070, opened 2025-03-08
added rails-migraiton-boy

haroon26:

My package is [Rails Migration Boy](https://github.com/haroon26/rails-migration-boy-subl). This provides basic rails migration commands (Up/Down/Redo) to be performed via command palette and code lens. 

[2]: https://semver.org

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -108,6 +108,7 @@
 		{
 			"name": "Rails Migration Boy",
 			"details": "https://github.com/haroon26/rails-migration-boy-subl",
+			"labels": ["rails", "ruby on rails"],
 			"releases": [
 				{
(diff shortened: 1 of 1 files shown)


braver:
Thanks for your submission. Your readme speaks specifically about SublimeText 3, which is no longer supported and has been superseded by ST 4 roughly 4 years ago... surely your brand new package works with ST4 as well?

haroon26:
Hey @braver, thanks for pointing that out. The plugin surely works for ST4, and I'll update the README accordingly

braver:
Also don't forget to set a description on your GitHub repo. That will show up as quick explanation of your package in various places inside package control.

PR #9072, opened 2025-03-13
Add new plugin to indent any code

zharkomi:

My package is SimpleIndent

[2]: https://semver.org

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1894,6 +1894,18 @@
 			]
 		},
 		{
+			"name": "Simple Indent",
+			"details": "https://github.com/zharkomi/SimpleIndent",
+			"author": "zharkomi",
+			"labels": ["indent", "bracket"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Simple JSX",
 			"details": "https://github.com/ccampbell/sublime-jsx",
 			"labels": ["JavaScript", "JSX", "React"],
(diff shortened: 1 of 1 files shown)


repository/s.json
```diff
@@ -1966,6 +1966,15 @@
 				}
 			]
 		},
+		{
+			"details": "https://github.com/zharkomi/SimpleIndent",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "main"
```
braver, line 1974: Branch based releases are not supported anymore. Please refer to the docs, check off the tasks in the PR template, and report back.

braver: You might also want to specify labels.

braver (commented):
The package looks interesting enough... you'd kind of expect something like this to exist already, but I can't find it.

zharkomi:
Fixed package config and created tag in the repo.

braver:
You'll need 3 digits in that tag to comply with semver (so 1.0.0 or 0.1.0).

zharkomi:
Created release:
https://github.com/zharkomi/SimpleIndent/releases/tag/1.0.0

zharkomi:
> You'll need 3 digits in that tag to comply with semver (so 1.0.0 or 0.1.0).

Created release:
https://github.com/zharkomi/SimpleIndent/releases/tag/1.0.0

braver:
Thanks, that's better already.

One thing I didn't catch before is that your menu entry still uses open_file instead of edit_settings, which will open the setting in split view ([example](https://github.com/braver/SideBarTools/blob/master/Main.sublime-menu#L17)). Please adjust that and we're ready to ship. 

While you're at it, it's recommended to also expose your settings via the command palette.

zharkomi:
> Thanks, that's better already.
> 
> One thing I didn't catch before is that your menu entry still uses open_file instead of edit_settings, which will open the setting in split view ([example](https://github.com/braver/SideBarTools/blob/master/Main.sublime-menu#L17)). Please adjust that and we're ready to ship.
> 
> While you're at it, it's recommended to also expose your settings via the command palette.

Fixed settings

PR #9073, opened 2025-03-13
Add Sone - Theme & Color Scheme

predragnikolic:

My package `Sone` is a Theme & Color Scheme.

`Sone` is similar to other themes and color schemes.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -2885,6 +2885,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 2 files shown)


predragnikolic:
I've added the MIT License and tagged a new release.

braver:
Nice 👍🏻

predragnikolic:
Thanks for the review 🙂

PR #9076, opened 2025-03-16
Add TestManager

cschreib:
- [ ] My package doesn't add key bindings. ** -- It does, but they are limited to a specific scope and won't interfere with other packages)
- [ ] Preferences and keybindings (if any) are listed in the menu and the command palette, and open in split view. -- As far as I can tell, they don't open in split view. Not sure how to do that?
- [ ] If my package is a syntax it doesn't also add a color scheme. *** -- Not applicable.
- [ ] If my package is a syntax it is named after the language it supports (without suffixes like "syntax" or "highlighting"). -- Not applicable.

My package provides the ability to discover, list, run, and inspect tests from any number of languages and frameworks in a unified way. It comes with support for a limited set of languages/frameworks by default, but provides an interface for supporting additional languages/frameworks with third-party packages. The interface aims to provide an equivalent to the Visual Studio "test explorer" (sadly that package name was already taken!), and is heavily inspired from the interface of [SublimeGit](https://github.com/SublimeGit/SublimeGit) (I am not the author). The package is cross-platform, though has not been tested on OSX.

There are no packages like it in Package Control; a number of packages exist for running tests, but none provide a dedicated user interface for listing all tests and easily running a custom subset of the tests.

[2]: https://semver.org

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -880,6 +880,23 @@
 			]
 		},
 		{
+			"name": "TestManager",
+			"details": "https://github.com/cschreib/sublime-TestManager",
+			"author": "cshreib (Corentin Schreiber)",
+			"labels": [
+				"testing",
+				"c++",
+				"python",
+				"rust"
+			],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "testNameGenerator",
 			"details": "https://github.com/testNameGenerator/SublimeText-plugin",
 			"releases": [


braver:
That's impressive, thanks for your submission. If only you supported phpunit, I'd be able to really try it 😅 

Some remarks:
- Your menu items for settings etc. still use `open_file` instead of `edit_settings`.
- It's typically also useful to be able to open the preferences and keybindings via the command palette.
- In your syntax, the `.test-list` should usually go at the end of a scope as the final specifier.... although I guess everything after that are made up scopes and I can kind of see how this would also make sense?

A larger package like yours typically takes a bit to go through and review properly. Hope you don't mind, I'll try and have a good look at it later and will update this PR when I get around to that.

cschreib:
Hi and thanks for the kind words! I have pushed (and tagged) the changes you suggested, thanks for the pointers. I have modeled the menu and commands on the LSP plugin, which seems like it would be a gold standard these days. Hopefully that's OK.

As for the syntax and the `.test-list` scope; indeed whatever is after that is made up. I don't really mind either way, though perhaps some of these scoped aren't totally made up (e.g. `string.other.test-list.time-delta` -- that could match a more generic scope). Is there somewhere a list of commonly-used scopes, to check what I could reuse?

As for PHP; if that helps with the review, I added a quick experimental implementation for PHPUnit. I had a look at the documentation, and it was pretty easy since it already supports test reports in the TeamCity format, for which I already have a parser. Unfortunately there are a few limitations:
 - I have found the `--list-tests` option for test discovery, however I cannot seem to get it to give me the file and line at which each test is defined; this isn't mandatory, but without it you cannot press 'Enter' on a test and directly go to the corresponding source code, which is very handy. 
 - I haven't been able to find a way to make PHPUnit run a selected list of tests (as opposed to all tests, or just a single test), so for now the test runner is running tests one by one, each in its own invocation of PHPUnit. This is not efficient, particularly on Windows where process creation is expensive.

The implementation is functional, except for 'Enter' to view the test source code, but I have tested it only a very small sample of tests, so let me know if it breaks ;)

braver:
>  I have modeled the menu and commands on the LSP plugin

Thanks, yeah, that's definitely a leading package that sees a lot of active maintenance and follows best practices all around.

> Is there somewhere a list of commonly-used scopes, to check what I could reuse?

Certainly! https://www.sublimetext.com/docs/scope_naming.html

Thanks for making the effort of looking into phpunit. I'll give it a spin (hopefully later this if nobody messes up my schedule at work) and might be able to give you some input on those things you mentioned. I'm no expert, but maybe I can figure something out.

braver:
So, I couldn't get it set up to work for phpunit. I'm really too much of a novice of that system (or unit tests in general to be honest) to be able to debug that since it seems to be phpunit itself that fails. 
I did get a feel for your UI choices and how well everything is documented and presented to the user. And I must say, job well done.

So, to keep this short, what this package needs is more users and feedback, and perhaps contributions. So let's ship it!

PR #9079, opened 2025-03-17
Add Color Scheme - Dracula

facelessuser:

My package is ...

Color Scheme - Dracula Neue

Adds a non-official variant of Dracula that takes a different approach to syntax highlighting and that makes available various Pro variants as well.

Merging this with the official Dracula color scheme is something that will not be happening due to the stricter management of the official one and how we are deviating by adding Pro variants. Due to many requests, I am attempting to make this available alongside the official one despite the similarities.

[2]: https://semver.org

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2641,7 +2641,7 @@
 			]
 		},
 		{
-			"name": "Color Scheme - Dracula",
+			"name": "Color Scheme - Dracula Neue",
 			"details": "https://github.com/facelessuser/sublime-dracula-scheme",
 			"labels": ["color scheme"],
 			"releases": [
--- a/repository/c.json
+++ b/repository/c.json
@@ -2641,8 +2641,8 @@
 			]
 		},
 		{
-			"name": "Color Scheme - Dusk",
-			"details": "https://github.com/geekpradd/Dusk-Color-Scheme",
+			"name": "Color Scheme - Dracula",
+			"details": "https://github.com/facelessuser/sublime-dracula-scheme",
 			"labels": ["color scheme"],
 			"releases": [
 				{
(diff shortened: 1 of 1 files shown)


braver:
Interesting, we've do have a handful packages following the "Color Scheme - X" naming convention, but forgot all about that. "X Color Scheme" is the more common one. And which the current Dracula package follows. I wonder if, for users just searching for ["Dracula"](https://packagecontrol.io/search/Dracula) it's apparent which is which and why. One says "Dracula color scheme for Sublime", the other "🧛🏻‍♂️ Dark theme for Sublime Text" ... not very helpful. One has 250k installs, the other definitely doesn't 😉 

I know you don't necessarily care about the marketing of your packages, but I wonder if we can put something in the repo description or name of the package to describe why this one exists and should at least be tried if you care about Dracula. I've always liked the [Monokai Neue](https://packagecontrol.io/packages/Monokai%20Neue) name for that reason (perhaps also because I'm a typography nerd). It tells you it's newer, but without making it sound like it's better per se. ... Suggestion:

**Dracula Neue Color Scheme**
A bespoke implementation of Dracula for Sublime Text, based on Mariana.

facelessuser:
Hmm, that is a possibility. I think I could just rename it to Dracula Pro (even though it also as the classic one as well)...I'd need to re-tag again, but maybe that could work.

I'm not sure if I care about whether it is Color Scheme - X or X Color Scheme and all of that, but I could certainly add a qualifier.

facelessuser:
If my suggestion is acceptable, I can go about updating with the qualifier "Pro".

braver:
That does perhaps suggest that you're delivering what's behind this button? 

Screenshot 2025-03-17 at 21 20 43

I've also seen several packages in package control where the "pro" denotes a paid for package (thinking about Monokai Pro specifically). I'd want to avoid that suggestion and go with "Next", "Neue", "Improved", "++" or something in that vein.

This borders on "who am I to tell you what to do" and if you want Pro I'm not going to stop you, but naming things is both hard and important, so I don't mind taking a second to bike shed over this.

braver:
.... Well you are in a way delivering an interpretation of what's behind that button. Hm. 🤔

facelessuser:
I don't know, I'm more just trying to find a path forward. I'm not attached to "Pro", I just wanted to get it up 🙃. I guess I'll just use "Neue".

braver:
🎉

PR #9082, opened 2025-03-24
Add HandyLLM

atomiechen:
  - Response: my package bundles support for the python PyPI project [HandyLLM][4] which defines a special format, Handy Prompt (`.hprompt` files); my package contains Hprompt syntax support, but it is named after the python project.

My package is named HandyLLM and bundles support for the python PyPI project [HandyLLM][4], which is a handy toolkit for using LLMs and defines Handy Prompt, a self-containing prompt in a human-friendly mark-up format `.hprompt`.

This package adds syntax highlight and fold support, snippets, commands, and the build system to run hprompt files. 

<div align="middle">
screenshot_example
</div>

[2]: https://semver.org
[4]: https://pypi.org/project/HandyLLM/

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -197,7 +197,7 @@
 			"labels": ["language syntax", "snippets", "build system"],
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": ">=3092",
 					"tags": true
 				}
 			]
--- a/repository/h.json
+++ b/repository/h.json
@@ -191,6 +191,18 @@
 			]
 		},
 		{
+			"name": "HandyLLM",
+			"details": "https://github.com/atomiechen/Sublime-HandyLLM",
+			"author": "Atomie CHEN",
+			"labels": ["language syntax", "snippets"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Hansl-Gretl-Language",
 			"details": "https://github.com/atecon/Hansl-Gretl-Language",
 			"labels": ["language syntax", "snippets", "completions"],
--- a/repository/h.json
+++ b/repository/h.json
@@ -194,7 +194,7 @@
 			"name": "HandyLLM",
 			"details": "https://github.com/atomiechen/Sublime-HandyLLM",
 			"author": "Atomie CHEN",
-			"labels": ["language syntax", "snippets"],
+			"labels": ["language syntax", "snippets", "build system"],
 			"releases": [
 				{
 					"sublime_text": "*",


braver:
- I’d leave the “(hprompt)” out of the syntax name.
- Each scope in the syntaxes should end with .hprompt to allow easy targeting of syntax-specific cases.
- Since your settings look to be basically syntax specific setting, I would let users change them via the existing “edit syntax settings” menu and command, instead of introducing a new one. Although I’m not 100% sure on this and don’t know how that would work out with your needs to refresh views when that setting changes.

repository/h.json
```diff
@@ -190,6 +190,18 @@
...
 			]
 		},
+		{
+			"name": "HandyLLM",
+			"details": "https://github.com/atomiechen/Sublime-HandyLLM",
+			"author": "Atomie CHEN",
+			"labels": ["language syntax", "snippets", "build system"],
+			"releases": [
+				{
+					"sublime_text": "*",
```
braver, line 200: You can choose to ignore the bot report, but it is correct in that “>= 3092” would be more accurate.

atomiechen: Thanks for the correction. I have made the change just now.

atomiechen:
> * I’d leave the “(hprompt)” out of the syntax name.
> * Each scope in the syntaxes should end with .hprompt to allow easy targeting of syntax-specific cases.
> * Since your settings look to be basically syntax specific setting, I would let users change them via the existing “edit syntax settings” menu and command, instead of introducing a new one. Although I’m not 100% sure on this and don’t know how that would work out with your needs to refresh views when that setting changes.

I really appreciate your kind suggestions. I have made according changes for the first and second point in my package repo. 

As for the third one, I have one concern: when users open "HandyLLM Settings", they can refer to the default settings on the left and made customizations. However, if I move all settings to the syntax specific setting (`Hprompt.sublime-settings`), they will not see it anymore. 

How can I resolve this issue? Or, my current setup is just fine? I would like your suggestion. Thanks a lot!

braver:
> they will not see it anymore

That’s true. I guess the question is whether the settings are purely settings, or defaults, for the syntax, or if they have any other significance for the functionality of the package. If you’re just saying “I want wrapping to be the default for this syntax”, then I would say the user doesn’t need to see _your_ settings as a reference. Instead, the user will see all global settings as a reference. 

What you can perhaps do though in that case, is put word_wrap in the “default” that you pass to the edit_settings command, since that’s the important setting here.

atomiechen:
Thanks for your reply. 

I'm sorry for not making myself clear. In my package, there are currently **two** setting files and they are for different reasons:
1. `HandyLLM.sublime-settings`: this one configures the realtime decorations of `.hprompt` files (and possibly more in the future) using Python plugins. These functions **cannot** be achieved with normal syntax definitions (like scanning the file twice to determine invalid frontmatter area):
   ```sublime-settings
   {
     // enable the decoration of invalid frontmatter area
     "enable_decor_frontmatter": true,
     // enable underline of message roles (e.g. $user$)
     "enable_underline_role": true,
     // the underline style of roles if enabled (solid, stippled, squiggly)
     "role_underline_style": "stippled",
   }
    ```
   These settings are **unique** to my package, and I need users to see the original `HandyLLM.sublime-settings` and customize (like the screenshot below).
  
   image

2. `Hprompt.sublime-settings`: the normal language specific setting, and currently it makes word wrapping the default for `.hprompt` files. Like what you said, I don't need users to see defaults in my package. When users open the language specific settings on a `.hprompt` file, the behavior is expected (`Preferences.sublime-settings` is opened, like the screenshot below).
  
   image

My question is: **Is this separate setting setup acceptable? Because they serve different purposes, I am unable to find a better way to merge them.** If this is okay, we can safely move on.

Thanks a lot for your advice!

atomiechen:
Fixed! Thanks for your quick reply!

PR #9085, opened 2025-04-10
Add Pyperclip to dependencies

mkruselj:
- [x] I have have read [the docs]

This package is required as a dependency for another package I am maintaining, in order to put the transpiled code onto system clipboard, when the transpiler is executed from CLI.

Thanks!

### DIFF
--- a/repository/dependencies.json
+++ b/repository/dependencies.json
@@ -825,6 +825,21 @@
 			]
 		},
 		{
+			"name": "pyperclip",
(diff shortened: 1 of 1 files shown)


mkruselj:
@packagecontrol-bot Please verify!

FichteFoll:
New libraries should be added to https://github.com/packagecontrol/channel/ using the new 4.0 schema format exclusively.

That said, do you really *need* pyperclip or is the ST-provided clipboard API enough for your use case?

PR #9088, opened 2025-04-14
Reduce duplicate tags of "language syntax"

michaelblyons:
This is not a new package. I am attempting to simplify the [Labels page](https://packagecontrol.io/browse/labels) on PC.

I didn't know what to do with @malexer's syntax history package, which is closer to AutoSetSyntax and ApplySyntax, so I left tags for that one alone.


braver:
💪🏻

PR #9089, opened 2025-04-15
Add Jumper

Mister7F:
https://github.com/Mister7F/sublime-jumper

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -1388,6 +1388,20 @@
 			]
 		},
 		{
+			"name": "Jumper",
+			"details": "https://github.com/Mister7F/sublime-jumper",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			],
+			"labels": [
+				"text manipulation",
+				"code navigation"
+			]
+		},
+		{
 			"name": "JumpTo",
 			"details": "https://github.com/SublimeText/JumpTo",
 			"author": ["JonasPf", "FichteFoll"],


repository/j.json
```diff
@@ -1387,6 +1387,25 @@
...
 		},
+		{
+			"name": "Jumper",
+			"author": [
+				"Mister7F"
+			],
+			"details": "https://github.com/Mister7F/sublime-jumper",
+			"readme": "https://raw.githubusercontent.com/Mister7F/sublime-jumper/refs/heads/master/readme.md",
+			"issues": "https://github.com/Mister7F/sublime-jumper/issues",
+			"releases": [
```
kaste, line 1398: Don't fill out author, readme, and issues as these are derived from details.

kaste:
You need to fill out the issue template.

The onboarding is kinda difficult for me.  

E.g.

```
{
    "keys": ["find", "enter", "<character>"],
    "command": "jumper_go_to_anywhere",
},
```

Is "find" really a key on your keyboard, and maybe on mine too?  

You do bind quite some keys by Default and without specifying a context.  That is strictly not allowed.  Can you imagine a providing command that helps configuring the key bindings?  Or a detailed README section otherwise that walks-through this process.  

I know this plugin started tinier.  Is this really a good bundle of functionality or should this rather be split up into multiple packages so users can choose à la carte what they want.  (This for example makes toggling/disabling parts of the plugin easier as the user just needs to disable the plugin.)

Mister7F:
Or I can just comment all keybind and let the use choose ? (Btw there’s a menu item so it should be easy to find)

kaste:
Yeah, you basically see it as a Jump-suite.  Technically, there is no code share between the enhanced modification history and the core acejumpy feature.  The select-* family is also something different. 

That aside, the shipped key-bindings are the problem here I think.  That file is huge and ships global bindings.  That means it can override user ones.   

From what I see, `jumper_go_to_anywhere` and `jumper_quick_scope`  form the core, novelty feature.  And we have 4 and then 6 bindings all based on "find" to make them work.   So I just thought: why not let the user specify this main shortcut -- "find" in your example --  and you write the keymap out for that?

Please fill out the PR/issue template, you just deleted it.  And let's see what others have to say as well.

Mister7F:
I added the issues / pr template

I also commented all keybinds (so we see example of keybinds when we click on the keybind menu item)

>  So I just thought: why not let the user specify this main shortcut -- "find" in your example -- and you write the keymap out for that?

What's the good way to do it ? A command that write on `Default (os)` in the package I guess, but how to ask for the shortcut ?
(it can be a combination of key,s a list, etc)

kaste:
I'm a bit confused why the bot stumbles about the key bindings although they're all commented out.  🤔 But also you changed that since then and now use `alt+i` for the example bindings.  I think a close/reopen dance will trigger the bot immediately.

While we wait for braver...  On these bindings, you use key chords or sequences like `<shortcut>, <character>` which makes the configuration maybe a bit longer.  As I counted before, a user needs 10 bindings for the core functionality.  What are the pro's here to use that sequence?  Very often plugins open an input box on the first shortcut.  And then just wait, while the input box is open, for the first next char, read it and proceed.  (Usually ESC in that editor mode escapes the mode and returns back to the editing mode.)  Using that conventional approach, you only have one binding -- that makes the configuration for the users simple -- and total control as you're basically in a custom editor "mode".  What are the reasons for your approach?

Mister7F:
>  What are the reasons for your approach?

The reason is that I don't really like that input box, it takes some space on the screen and can hide text

Ideally, it should never be visible, but I didn't manage to not use it at all

It's still annoying if the text you want to jump to is behind the input panel, but I don't have solution for it

So delaying the panel is a bit better than opening immediately imo, so if you press by error, you don't have it and you can just press escape 

Note that the panel is filled with the key you pressed when it was not yet opened (you can still press backspace if you want to jump to an other character)

Mister7F:
or if we have an input panel can can be opened in the top right corner, I can open it immediately, but I don't think it's possible

kaste:
~~How do you "escape" with your approach~~ That's probably built-in; Sublime escapes key sequence with "escape".

Mister7F:
@kaste Also what's weird is that `find` is a valid recognized key by sublime, but I understand is niche

kaste:
Take a picture of the key.  The name isn't listed at  https://www.sublimetext.com/docs/key_bindings.html#key-names and maybe that's all the "linter" recognizes.

Mister7F:
@kaste Probably

image

kaste:
A picture from your physical keyboard, lol.  Is that one the media keys?

Mister7F:
@kaste It look like this :) (but you can choose which key you want for each physical key in [vial](https://get.vial.today/))

image

Mister7F:
@braver Hi, I think it should be fine

braver:
🎉

PR #9092, opened 2025-04-18
New dark color scheme

loverde-co:
For personal use, but all can download, use and modify

My package is LC Dark Color Scheme

[2]: https://semver.org

### DIFF
--- a/channel.json
+++ b/channel.json
@@ -123,6 +123,7 @@
 		"https://raw.githubusercontent.com/x3ns/x3-alpha-theme/master/packages.json",
 		"https://raw.githubusercontent.com/xgenvn/InputHelper/master/packages.json",
 		"https://raw.githubusercontent.com/zfkun/sublime-kissy-snippets/master/packages.json",
+		"https://git.loverde.com.br/git/LC-Dark-Color-Scheme/src/main/package.json",
 		"https://tinkerwell.app/plugins/sublime/packages.json"
 	]
 }
--- a/repository/l.json
+++ b/repository/l.json
@@ -663,6 +663,18 @@
 			]
 		},
 		{
+			"name": "LC Dark Color Scheme",
+			"details": "https://gi.loverde.com.br/git/LC-Dark-Color-Scheme",
+			"author": "Loverde Company",
+			"labels": ["theme", "color scheme", "dark color", "dark"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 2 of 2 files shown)


loverde-co:
Hi, fixed the conflict : )

braver:
Looks like you're out of alphabetical order there.

channel.json
```diff
@@ -120,6 +120,7 @@
 		"https://raw.githubusercontent.com/x3ns/x3-alpha-theme/master/packages.json",
 		"https://raw.githubusercontent.com/xgenvn/InputHelper/master/packages.json",
 		"https://raw.githubusercontent.com/zfkun/sublime-kissy-snippets/master/packages.json",
+		"https://git.loverde.com.br/git/LC-Dark-Color-Scheme/src/main/package.json",
```
braver, line 123: If you're adding to the channel, you don't need to also add to the repository (IIRC, admittedly we don't get this often), since that package json should describe the entire package.

I'm getting a "this connection is note private" warning on that URL though, not sure that will work.

loverde-co: It will work, it is a private repository, on my VPS server. But this repo is public

braver: Right. The format is not correct though. Have a look at this for example: https://raw.githubusercontent.com/zfkun/sublime-kissy-snippets/master/packages.json

Since you're using something other than the public services we can crawl, you need to provide a pre-packages sublime package (zip), and a json describing the package according this document:
https://raw.githubusercontent.com/wbond/package_control/master/example-repository.json

braver: I'd recommend getting a certificate on that server so https will actually work properly.

loverde-co: i have a certificate, it is a Lets Encrypt certificate. You need to trust to open ( wired, i know ).

I will fix those things and update this PR. Thanks

braver: Well that's not really weird, the cert is not valid for that domain:

```
Web sites prove their identity via certificates. Firefox does not trust this site because it uses a certificate that is not valid for git.loverde.com.br. The certificate is only valid for [loverde.com.br](https://loverde.com.br/).
 
Error code: [SSL_ERROR_BAD_CERT_DOMAIN](about:certerror?e=nssBadCert&u=https%3A//git.loverde.com.br/git/LC-Dark-Color-Scheme/src/main/package.json&c=UTF-8&d=%20#certificateErrorDebugInformation)
```

braver:
> I will fix those things and update this PR. Thanks

Let me know when you're ready to continue the review. The cert issue might be a sticky thing, the crawler might simply not accept that.

PR #9094, opened 2025-04-18
Reduce duplicate tags of "utilities"

michaelblyons:
Similar to https://github.com/wbond/package_control_channel/pull/9088

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1961,7 +1961,7 @@
 		{
 			"name": "Code Champion",
 			"details": "https://github.com/Atbox/CodeChampion",
-			"labels": ["utilities", "utils", "utility"],
+			"labels": ["utilities"],
 			"releases": [
 				{
 					"sublime_text": "*",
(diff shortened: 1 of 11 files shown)


braver:
Perhaps we should add a style guide for this to the README or the PR template...

repository/c.json
```diff
@@ -3157,7 +3157,7 @@
 			"name": "CommentFold",
 			"description" : "Gives an overview of your comments by folding all other code",
 			"details": "https://github.com/mapsiter/CommentFold",
-			"labels": ["utilily", "code navigation"],
```
braver, line 3160: 🌷

braver:
Do we need to discuss singular vs plural here 😅

deathaxe:
FWIW, with regards to how labels are presented on website and number of provided features (utilities, completions, ...) not being specified, plural might be the more common/suitable choice to avoid multiple terms for same things.

michaelblyons:
I don't mind if you want this to be plural or not, but I picked the one with 4x the existing usage.

It's also unclear what "utilities" even means... commands? anything not covered by linting, ... , formatting, etc.?

braver:
Yeah I had the same idea… if a package doesn’t provide some utility, then why does it exists? I guess all non color scheme/theme packages are utilities.

Let’s not get too deep into the semantics, consistency is key and everything else is secondary or less.

deathaxe:
If this label is too generic, why not removing?

PR #9095, opened 2025-04-18
Reduce label duplication

michaelblyons:
Similar to #9088, except WIP since more work needs to be done:

- [x] Fix `snippets` duplicates.
- [x] Fix `completions` duplicates.
- [x] Pick a canonical label for Python-provided completions, probably from among `auto-complete` (263 uses), `autocomplete` (27 uses), and `autocompletion` (8 uses).
    - There is some talk on Discord about using `auto-completions` for these to mirror `completions` for static completions. 
- [ ] ??? Sort packages among the three options above. (A package could have more than one.)


michaelblyons:
PC bot working hard on this one. 😜

braver:
Looks good, although I didn't check everything. I personally don't care in what order labels appear. The order has no functional implication or meaning. But now that you've reordered many of them, no need to revert that either.

PR #9100, opened 2025-04-24
Remove ST2 packages

deathaxe:
This PR proposes to remove all ST2-only packages.

Time for ST2 is finally over. It fails to establish connections to Package Control on most modern OSs and is end-of-life for over a decade.


braver:
For those from the channel.json, how do you decide which to drop?

Also, removals don't get handled by the site as far as I'm aware, so they'll be discoverable there and I'm not sure it might create other problems?

deathaxe:
> For those from the channel.json, how do you decide which to drop?

PC3+ assumes all packages in repositories of schema_version < 2.0 to support ST2 only. It assigns `sublime_text: <=3000` selector implicitly.

> Also, removals don't get handled by the site as far as I'm aware, so they'll be discoverable

1. They won't be crawled any longer, reducing risk of hitting rate limits again.
2. They won't be part of channel_v3.json anymore, reducing download size.

_If we really want them to no longer be discoverable, we could create a dummy package "Removed" (or reuse https://github.com/SublimeText/Banned) and add all of them to its `previous_name` field. But that required some efforts for packages provided by external repositories._

FichteFoll (approved):

braver:
Well it parses them and does various related checks like alphabetical order, which all presumably doesn’t work if they’re not valid. I don’t know the inner workings enough to say if it does a separate validity check or with what kind of parser.

FichteFoll:
It parses the changed files of git merge base and the PR head and then diffs the parsed data.

deathaxe:
JSON iteself is already validated by unnittests. see: https://github.com/wbond/package_control_channel/blob/3edf24fff3afcb87a847b0e091e343a97d178e0b/.github/workflows/ci.yml#L16-L17

They are valid.

It's just the changeset, any part of packagecontrol chokes on. Maybe there are too many changes.

Crawler crawls a small subset of packages per run. If each package of such run was removed, it may return an error compared to what we see in Package Control on network errors. "No packages found for update". It might be mis-interpreted as network error aborting any real database update, thus failing again and again on the same set of packages.

That's just a guess with some knowledge about some design flaws, which seem possible. Real debugging would require insights into what's going wrong though.

PR #9103, opened 2025-05-01
switch colinta to 'main' branch

colinta:
The packages.json file that was being pulled down was out of date - I've pushed to `master` in the meantime, but also figured I would update this to the correct branch name.


colinta:
The CI check is failing, but I'm not getting any helpful output from the server
```
curl -i --http1.1 -X POST https://packagecontrol.io/test_pr/9103.json
HTTP/1.1 500 Internal Server Error
Server: nginx
...
```

colinta:
Rerunning on an older PR there are no issues:
```
curl -i --http1.1 -X POST https://packagecontrol.io/test_pr/9098.json
HTTP/1.1 200 OK
```

PR #9106, opened 2025-05-05
Add WGSL Enhance Package

iahu:
- [ x If my package is a syntax it doesn't also add a color scheme. ***

My package is WGSL Enhance

My package is similar to WGSL However it should still be added because my package include snippets and supports JS template syntax.

[2]: https://semver.org

### DIFF
--- a/repository/w.json
+++ b/repository/w.json
@@ -360,6 +360,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


deathaxe:
We try to avoid duplicate or too similar packages.

You should attempt to file a PR on original package's repository to ask for addition of desired features first.

PR #9111, opened 2025-05-08
Add FTP Backup plugin

H4ckMM3:
- [x] I have read [the docs](https://packagecontrol.io/docs/submitting_a_package).
- [x] I use [.gitattributes](https://git-scm.com/docs/gitattributes) to exclude files from the package: images, test files, sublime-project/workspace.

My package is FTP Backup and it adds automatic file backup functionality for Sublime Text with support for FTP servers.

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -2052,6 +2052,17 @@
 					"tags": true
 				}
 			]
+		},
+		{
+		    "name": "FTP Backup",
+		    "details": "https://github.com/H4ckMM3/FTP-Backup",
+		    "releases": [
+		        {
+		            "sublime_text": ">=3000",
+		            "tags": true
+		        }
+		    ],
+		    "labels": ["file backup", "backup", "ftp"]
 		}
 	]
 }


braver:
@H4ckMM3 Please use the [PR template](https://github.com/wbond/package_control_channel/blob/master/.github/PULL_REQUEST_TEMPLATE.md) (available right when you create the PR) to tell me various things about the package and help speed up the review process.

braver:
Please also explain the need for the [.no-sublime-package](https://github.com/H4ckMM3/FTP-Backup/blob/main/.no-sublime-package) file, and I'm also curious about your use of font-awesome here.

H4ckMM3:
.no-sublime-package is required to allow the plugin to write backup files to the filesystem, which isn't possible if it's installed as a .sublime-package (read-only).

Font Awesome is used for icons in the backup panel UI. I can replace or remove it if needed.

H4ckMM3:
> Please also explain the need for the [.no-sublime-package](https://github.com/H4ckMM3/FTP-Backup/blob/main/.no-sublime-package) file, and I'm also curious about your use of font-awesome here.

.no-sublime-package is required to allow the plugin to write backup files to the filesystem, which isn't possible if it's installed as a .sublime-package (read-only).

Font Awesome is used for icons in the backup panel UI. I can replace or remove it if needed.

H4ckMM3:
> @H4ckMM3 Please use the [PR template](https://github.com/wbond/package_control_channel/blob/master/.github/PULL_REQUEST_TEMPLATE.md) (available right when you create the PR) to tell me various things about the package and help speed up the review process.

I've completed the PR template as requested

braver:
> allow the plugin to write backup files to the filesystem

That should only be true if you’re writing to the package directory of your package. And one should never do that, except for some edge cases. But especially not for backups, because when uninstalling, and potentially even when upgrading, the package that directory and all the backups would get deleted.

braver:
Please note that your package is not added in alphabetical order (which is what the bot is failing over).

H4ckMM3:
> > allow the plugin to write backup files to the filesystem
> 
> That should only be true if you’re writing to the package directory of your package. And one should never do that, except for some edge cases. But especially not for backups, because when uninstalling, and potentially even when upgrading, the package that directory and all the backups would get deleted.

Understood. I've removed the .no-sublime-package file, since the plugin doesn't write to its own package directory. Backups are stored in a separate location.

H4ckMM3:
> Please note that your package is not added in alphabetical order (which is what the bot is failing over).

Placed the plugin entry in f.json in the correct alphabetical position.

Let me know if anything else needs to be changed.

braver:
>  My package doesn't add key bindings

... why did you check that box?

braver:
Anyway, have a look at the new output from the bot.

I can't read your README, so that makes it hard to evaluate the package. I also notice many of the menu entries are in Russian. Which isn't a problem per se, but severely limits the reach of your package as most users (like me) won't be able to understand your package. 

By the way, I notice you introduce a new top level main menu. Please don't do that. I think your functionality could live in the tools menu.

> Font Awesome is used for icons in the backup panel UI. I can replace or remove it if needed.

I didn't know you could do that. Without ever referencing the js file even. Can you tell me how that works? It looks like you want to render a full featured HTML page in ftp_backup_interface.py, but most of the code (like prepare_html_with_api, and starting/stopping the http server) is not referenced it seems. 

Your default backup_root has your own username and desktop backup path it in. That's not appropriate for a public package. You also seem to mention a create_month_folder setting, shouldn't that be in the default settings file too?

H4ckMM3:
> Anyway, have a look at the new output from the bot.
> 
> I can't read your README, so that makes it hard to evaluate the package. I also notice many of the menu entries are in Russian. Which isn't a problem per se, but severely limits the reach of your package as most users (like me) won't be able to understand your package.
> 
> By the way, I notice you introduce a new top level main menu. Please don't do that. I think your functionality could live in the tools menu.
> 
> > Font Awesome is used for icons in the backup panel UI. I can replace or remove it if needed.
> 
> I didn't know you could do that. Without ever referencing the js file even. Can you tell me how that works? It looks like you want to render a full featured HTML page in ftp_backup_interface.py, but most of the code (like prepare_html_with_api, and starting/stopping the http server) is not referenced it seems.
> 
> Your default backup_root has your own username and desktop backup path it in. That's not appropriate for a public package. You also seem to mention a create_month_folder setting, shouldn't that be in the default settings file too?

The web interface code is part of a planned future feature to provide a more user-friendly way to manage backups. Currently, all interaction happens through the editor, but I’ve included the base structure for upcoming integration.
Thanks for pointing that out. I've removed unused interface-related functions for now, fixed the default backup_root to avoid using a hardcoded local path, and added the missing create_month_folder option to the default settings file.
Let me know if there's anything else to adjust.

H4ckMM3:
> > My package doesn't add key bindings
> 
> ... why did you check that box?



> * My package doesn't add key bindings. **

I accidentally checked the box saying "My package doesn’t add key bindings." The package does include key bindings in the appropriate file.

H4ckMM3:
I've made all the requested changes. Could you please review it again?

H4ckMM3:
@braver Everything is fine?

kaste:
The Default (Windows).sublime-keymap is strange.  a) you normally aren't allowed to grab or overwrite shortcuts *unconditionally*/without using a *context*.  b) these only apply on Windows, why is that? c) Why do you define "noop" bindings here for what is typically the "save"-action?

IMO since this a public tracker/registry with a very limited community and with limited resources I *personally* would require this to be English.  Esp. as it is in fact possible to advertise the package elsewhere and then to require to add the repository using Package Control.

H4ckMM3:
> The Default (Windows).sublime-keymap is strange. a) you normally aren't allowed to grab or overwrite shortcuts _unconditionally_/without using a _context_. b) these only apply on Windows, why is that? c) Why do you define "noop" bindings here for what is typically the "save"-action?
> 
> IMO since this a public tracker/registry with a very limited community and with limited resources I _personally_ would require this to be English. Esp. as it is in fact possible to advertise the package elsewhere and then to require to add the repository using Package Control.



> The Default (Windows).sublime-keymap is strange. a) you normally aren't allowed to grab or overwrite shortcuts _unconditionally_/without using a _context_. b) these only apply on Windows, why is that? c) Why do you define "noop" bindings here for what is typically the "save"-action?
> 
> IMO since this a public tracker/registry with a very limited community and with limited resources I _personally_ would require this to be English. Esp. as it is in fact possible to advertise the package elsewhere and then to require to add the repository using Package Control.

Thank you for your review and feedback regarding my FTP Backup package.

After considering your concerns, I've decided to take a different approach for distribution. Since this plugin is primarily intended for use by my small team of 5 people and contains Russian-language UI elements, I'll be distributing it via a direct GitHub repository instead of through the main Package Control channel.

My team members will add the repository manually through the "Package Control: Add Repository" command and install it from there. This seems like the most appropriate solution given:

1. The localized nature of the plugin (Russian interface)
2. The specialized keyboard bindings that intentionally override standard save commands
3. The limited target audience (our development team)

I appreciate your time and the valuable feedback you've provided. This approach will allow us to use the plugin as needed while respecting Package Control's standards.

PR #9116, opened 2025-05-14
Add LLM Prompter plugin

sashabaranov:
- [] Preferences and keybindings (if any) are listed in the menu and the command palette, and open in split view.

My package is ...

[2]: https://semver.org

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -1458,6 +1458,16 @@
 			]
 		},
 		{
+			"name": "LLM Prompter",
+			"details": "https://github.com/sashabaranov/sublime-llm-prompter",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "master"
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


repository/l.json
```diff
@@ -1457,6 +1457,16 @@
 				}
 			]
 		},
+		{
+			"name": "LLM Prompter",
+			"details": "https://github.com/sashabaranov/sublime-llm-prompter",
+			"releases": [
+				{
+					"sublime_text": "*",
```
kaste, line 1465: I guess must be >4000 as you use python 38 syntax

repository/l.json
```diff
@@ -1457,6 +1457,16 @@
...
+					"branch": "master"
```
kaste, line 1466: "branch" is deprecated, use `"tags": true`

braver:
👋🏻 please respond to the feedback to continue the review.

PR #9117, opened 2025-05-14
Fix duplicate BBCode Syntax registration

deathaxe:
Issue: "BBCode Syntax" is defined in local b.json and an external repository.

This commit moves packages into local repositories and removes reference to the external one to resolve the conflict, given that local BBCode syntax package is newer.

It was found by https://github.com/wbond/package_control/pull/1698#issuecomment-2878149450

### DIFF
--- a/channel.json
+++ b/channel.json
@@ -94,7 +94,6 @@
 		"https://raw.githubusercontent.com/seanliang/ConvertToUTF8/master/packages.json",
 		"https://raw.githubusercontent.com/sentience/DokuWiki/master/packages.json",
 		"https://raw.githubusercontent.com/sentience/HyperlinkHelper/master/packages.json",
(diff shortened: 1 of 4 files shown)


deathaxe:
Well, given those packages registrations are just moved and already shipped in their current state, we have 2 options to move on (given there are no plans to overtake maintenance):

1. ignore review errors and just merge
2. remove IPS BBCode package as "non-complient"

FichteFoll:
Well, the package obviously provides a bad UX, but we don't have a policy on removing packages that do not meet our current standards, so we should leave it in for now. I am strongly in favor of ignoring.

That said, we should inform the GitHub account affected by this so they don't end up wondering why nothing happens after updating the channel, unlikely as it seems since the file hasn't been touched in 7 years.

@Skullmonkey FYI

FichteFoll (approved):

PR #9119, opened 2025-05-15
Drop sublime rubocop

pderichs:
No longer maintained.

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -2193,16 +2193,6 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


braver:
Let's leave it in. It's still seeing installs and it presumably still works. If you're desperate to get rid of the repo it can be transferred elsewhere. If we leave it in a new maintainer might come up, or not. If it's no longer in there, it's really end of the line probably. Deletion is always an option at any later point in time.

PR #9120, opened 2025-05-17
Added LaTeXOutline

odapg:
   Suggested keybindings are commented in sublime-keymaps files
   It is not a syntax, but has a specific (hidden) one, with specific (hidden) color-scheme

My package is similar to [Outline](https://packagecontrol.io/packages/Outline) at first glance.
However i still believe it should be added because of the following reasons.
Outline is a general tool for markdown/latex languages that puts the symbols list in a side view in a clickable way.
By specializing to LaTeX files, LaTeXOutline can:
- show a table of contents of the document, with suitable indentation and color highlighting
- show the sections and labelled environment numbers (by parsing the .aux file)
- show the environment name corresponding to a label (by looking into the .tex file)
- propose to easily copy a label/have a look at it without moving the selection (something I believe is quite helpful, at least to me!)

I don't think that this "LaTeX table of contents" feature (that one can see for instance on overleaf) presently exists in Package Control.

[2]: https://semver.org

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -554,7 +554,7 @@
 		{
 			"name": "LaTeXOutline",
 			"details": "https://github.com/odapg/LaTeXOutline",
-			"labels": ["LaTeX"],
+			"labels": ["latex"],
 			"releases": [
 				{
 					"sublime_text": ">=4085",
--- a/repository/l.json
+++ b/repository/l.json
@@ -557,7 +557,7 @@
 			"labels": ["latex"],
 			"releases": [
 				{
-					"sublime_text": ">=4085",
+					"sublime_text": ">=4149",
 					"tags": true
 				}
 			]
--- a/repository/l.json
+++ b/repository/l.json
@@ -552,6 +552,17 @@
 			]
 		},
 		{
+			"name": "LaTeXOutline",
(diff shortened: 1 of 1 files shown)


repository/l.json
```diff
@@ -551,6 +551,17 @@
 				}
 			]
 		},
+		{
+			"name": "LaTeXOutline",
+			"details": "https://github.com/odapg/LaTeXOutline",
+			"labels": ["LaTeX"],
```
braver: We've recently added a bit of a style guid: https://github.com/wbond/package_control_channel?tab=readme-ov-file#style-guide
> Labels are always in lowercase.


```suggestion
			"labels": ["latex"],
```

odapg:
- For the `messages.json` issue, I am not sure to understand what caused the problem. Perhaps the trailing comma? I modified the file.
- For the concerns about .sublime-mousemap files, I added a context to them. Correspondingly, I have set 4149 as the minimal version for ST from now on.

braver:
>  I am not sure to understand what caused the problem

Perhaps the key "1.0", since that's never a supported version anyway (that would be 1.0.0).

🤷🏻 I don't necessarily know the inner workings of the bot or how to interpret some of the output...

PR #9128, opened 2025-06-05
Remove "previous_names" entries older than 2 years

deathaxe:
This commit cleans up previous names entries, which haven't been modified for at least 2 years.


braver:
Oh... damn, conflicts.

No worries that the old crawler will bork over this?

deathaxe:
Those settings actually seem to cause quite some work to check renames and update database when crawling. The crawler primarily seems to get stuck with too many existing packages being removed.

PR #9132, opened 2025-06-06
Replace external repositories

deathaxe:
This PR moves package registration to internal repository and removes corresponding external repositories from channel to reduce external dependencies.

External repositories remaining are...

1. ST2-only repositories which no longer exist in `four-point-oh` branch
2. repositories pointing to explicit download assets, which can't be migrated without preventing automated package updates
3. trusted repositories with many packages (e.g.: LSP, SublimeLinter, ...)

1 covers 59 repositories

2 and 3 cover 23 repositores

_With ST2-only channels removed and this PR merged, only 23 external repositories remain._

### DIFF
--- a/channel.json
+++ b/channel.json
@@ -7,7 +7,6 @@
 		"https://emmetio.github.io/sublime-text-plugin/registry.json",
 		"https://packages.monokai.pro/packages.json",
 		"https://raw.githubusercontent.com/20Tauri/DoxyDoxygen/master/DoxyDoxygen.json",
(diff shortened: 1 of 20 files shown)


braver:
As you'll have noticed I'm behind on reviews and this is a big one. Can we perhaps do them in batches so we can also more clearly see where the bot chokes?

deathaxe:
I barely remember to have found some candidates, but wasn't able to find the reason. 

Anyone has a script to create a PR per commit?

PR #9135, opened 2025-06-08
Add UIkitCompletions package

dlzi:

My package is ...

[2]: https://semver.org

### DIFF
--- a/repository/u.json
+++ b/repository/u.json
@@ -88,6 +88,17 @@
 			]
 		},
 		{
+			"name": "UIkitCompletions",
+			"details": "https://github.com/dlzi/UIkitCompletions",
+			"labels": ["auto-complete"],
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "UL4",
 			"details": "https://github.com/LivingLogic/LivingLogic.SublimeText.UL4",
 			"labels": ["language syntax"],


braver:

Well, that's not really true 😅 https://packages.sublimetext.io/packages/UIkit%20autocomplete/ exists... it's truly ancient too. Perhaps you can contribute to that project?

dlzi:
> > There are no packages like it in Package Control.
> 
> Well, that's not really true 😅 https://packages.sublimetext.io/packages/UIkit%20autocomplete/ exists... it's truly ancient too. Perhaps you can contribute to that project?

Hey. You are right, I missed that part, my bad.

Yeah, it seems they don’t update their own plugin. I asked about the status of the project here: https://github.com/uikit/uikit-sublime/issues/20. But feel free to decline or reject this merge — it’s just a side plugin I made for myself, and I might not have time to keep up with it.

braver:
Thanks! The thing to do in this case is to give them a couple of weeks to respond. If they don't, we can replace their package with yours. That way, packages stay fresh and maintained. But ideally they do and allow you to contribute. We'll see :) Let's check back in a week or two or so.

dlzi:
Hi @braver, it seems that UIkit people are not into fixing the plugin. What is the policy now?

braver:
There are 2 possible courses of action:

- You update the existing entry to point at your repo
- You remove the existing entry and add its name to `previous_names` on your entry

Both have the same end result, but with option 2 you get to rename and change whatever else you like about the package entry.

dlzi:
Sorry mate, I didn't understand. What do you recommend me to do?

repository/u.json
```diff
@@ -87,6 +87,17 @@
 				}
 			]
 		},
+		{
+			"name": "UIkitCompletions",
+			"details": "https://github.com/dlzi/UIkitCompletions",
+			"labels": ["auto-complete"],
```
braver, line 82:
```suggestion
			"labels": ["auto-complete"],
			"previous_names": ["UIkit autocomplete"],
```

And then remove the "UIkit autocomplete" entry from this file. That tells Package Control that yours is now leading and can be used as an upgrade from "UIkit autocomplete".

You'll also need to tag a new version of your package with a higher number than 2.0.1 (the final version of "UIkit autocomplete").

braver:
Thanks, you're not ready to remove the entry around https://github.com/wbond/package_control_channel/pull/9135/files#diff-7d271effacde3e2304cc742f38c9df214c2f3f659f19117635d89177c6b99192R79

dlzi:
Hi,

No worries. But I'm lost, what do you want me to do?

Cheers,

braver:
Basically, you're taking over the old package so that entry needs to get removed. You could apply this diff:

```diff
diff --git a/repository/u.json b/repository/u.json
index a4807607..a0b61871 100644
--- a/repository/u.json
+++ b/repository/u.json
@@ -76,17 +76,6 @@
                                }
                        ]
                },
-               {
-                       "name": "UIkit autocomplete",
-                       "details": "https://github.com/uikit/uikit-sublime",
-                       "labels": ["auto-complete"],
-                       "releases": [
-                               {
-                                       "sublime_text": "*",
-                                       "tags": true
-                               }
-                       ]
-               },
                {
                        "name": "UL4",
                        "details": "https://github.com/LivingLogic/LivingLogic.SublimeText.UL4",
```

PR #9143, opened 2025-06-24
Xeto plugin

mgiannini:
- [X ] I have have read [the docs][1].

My package is jsut to add syntax highlighting support for the Xeto data specification language. This is the only package of its sort in Sublime Text right now

### DIFF
--- a/repository/x.json
+++ b/repository/x.json
@@ -101,6 +101,17 @@
 			]
 		},
 		{
+			"name": "Xeto",
+			"details": "https://github.com/mgiannini/sublime-xeto",
+			"labels": ["xeto", "project haystack"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "xit!",
 			"details": "https://github.com/jotaen/xit-sublime",
 			"releases": [


repository/x.json
```diff
@@ -100,6 +100,17 @@
 				}
 			]
 		},
+		{
+			"name": "Xeto",
+			"details": "https://github.com/mgiannini/sublime-xeto",
+			"labels": ["xeto", "project haystack"],
```
braver:
```suggestion
			"labels": ["language syntax", "xeto"],
```

Since this is a language support package, the language syntax label should be added. The "project haystack" label is unique and therefore kind of meaningless in this context.

braver:
Your package README says "A Sublime Text 3 plugin" a lot... Sublime Text 4 has been out for years now, very few users of version 3 remain that are looking for new packages. Maybe just leave out version numbers from the readme as it might suggest it won't work with newer versions.

mgiannini:
> Your package README says "A Sublime Text 3 plugin" a lot... Sublime Text 4 has been out for years now, very few users of version 3 remain that are looking for new packages. Maybe just leave out version numbers from the readme as it might suggest it won't work with newer versions.

I pushed an update to my README to remove references to sublime text 3 as you suggested. thanks!

PR #9146, opened 2025-07-01
IssueBadge

issuebadge:

My package is ...

[2]: https://semver.org


braver:
Ok, thanks. 

The settings should be exposed via the Main menu, and ideally also via the command palette.

Your entry in this repository needs to go into the `i.json` file.

PR #9147, opened 2025-07-01
Add Penrose package

LexouDuck:

My package is "Penrose" (language syntax)

[2]: https://semver.org

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -749,7 +749,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"branch": "master"
+					"tags": true
 				}
 			]
 		},
--- a/repository/p.json
+++ b/repository/p.json
@@ -744,6 +744,16 @@
 			]
 		},
 		{
+			"name": "Penrose",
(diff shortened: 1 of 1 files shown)


LexouDuck:
Indeed, good points. I have fixed these issues and thus pushed a new "1.0.1" tag.

> Also I wonder what is the purpose of the [Penrose.sublime-settings](https://github.com/LexouDuck/SublimeText-Penrose/blob/e05c2e040244402ced95accce8ee2b937a68069c/Penrose.sublime-settings) file?

My thought was that it would automatically associate the relevant file extensions with this syntax definition, but it seems this is accomplished by the `file_extensions:` field in the .sublime-syntax file itself - and, upon further reflection, it is silly to have all 3 different syntaxes in one such settings file, that makes no sense as it doesn't determine which of the 3 file extensions ought to go with which respective syntax definition.
In any case, I've cleaned up this mistake - thank you for bringing it up.

braver:
You've got a couple of pretty advanced regexes in there, you might want to look into adding some [tests](https://www.sublimetext.com/docs/syntax.html#testing).

Other than that, looks good!

PR #9150, opened 2025-07-03
Language syntax tag (A-Z)

michaelblyons:
This PR adds missing `language syntax` tags to packages that provide a syntax definition. I skipped packages whose syntax was meant only to highlight its own build output or other never-saved output.

See https://github.com/packagecontrol/thecrawl/issues/40 for some additional context.

### DIFF
--- a/repository/x.json
+++ b/repository/x.json
@@ -232,7 +232,6 @@
 			"name": "XQuery",
 			"details": "https://github.com/PogiNate/XQuery-Sublime",
 			"labels": ["language syntax"],
-			"labels": ["language syntax"],
 			"releases": [
 				{
 					"sublime_text": "*",


repository/x.json
```diff
@@ -228,6 +231,8 @@
 		{
 			"name": "XQuery",
 			"details": "https://github.com/PogiNate/XQuery-Sublime",
+			"labels": ["language syntax"],
+			"labels": ["language syntax"],
```
FichteFoll: duplicate key

FichteFoll (changes requested):
Thanks for these updates. I spotted one error but otherwise LGTM.

FichteFoll (approved):

PR #9151, opened 2025-07-04
Fix editorconfig typos

michaelblyons:
Fixes typos added in #3545 and never changed. I don't use editorconfig, but we should use the right key names if we're going to keep this file.


FichteFoll (approved):
Nice find. (I don't use editorconfig either.)

PR #9152, opened 2025-07-05
Adds RoadBookmarks plugin

stevengpa:

My package is Sublimet Text plugin that enhance the native Sublime Text bookmarks feature by persisting the information and making them accesible in a global scope through a Window Panel for easy navigation.

My package is similar to Signet Bookmarks However it should still be added because it has a more convenient navigation window panel, it also stores bookmarks in global scope (not only project or open files scope), allowing devs to access files from other projects from the window panel.

[2]: https://semver.org

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -1737,25 +1737,23 @@
 			]
 		},
 		{
+			"name": "RoadBookmarks",
+			"details": "https://github.com/stevengpa/RoadBookmarks",
+			"labels": ["bookmark", "navigation"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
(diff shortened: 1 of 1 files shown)


repository/r.json
```diff
@@ -1735,6 +1735,19 @@
 				}
 			]
 		},
+		{
+			"name": "RoadBookmarks",
+			"details": "https://github.com/stevengpa/RoadBookmarks",
+			"readme": "https://raw.githubusercontent.com/stevengpa/RoadBookmarks/refs/heads/main/README.md",
+			"issues": "https://github.com/stevengpa/RoadBookmarks/issues",
```
deathaxe: Those lines are redundant and can be removed. Related information is automatically scraped by the crawler.

PR #9165, opened 2025-08-05
Drop dead packages

deathaxe:
The repo errors with 404.


braver:
Thanks! 🧹 This feels like spring cleaning 😄

PR #9166, opened 2025-08-05
Drop "muukii"

deathaxe:
The repo behind just contains a README and doesn't provide any tags.

Given its nonsense content, this commit drops it.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -2876,16 +2876,6 @@
 			]
 		},
 		{
-			"name": "muukii",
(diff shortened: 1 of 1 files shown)


braver:
That's odd. That repo is from 2020, but the package was added in 2014: https://github.com/wbond/package_control_channel/pull/2939/files. The user no longer seems to have any other repos that look like they might have a color scheme in it. I think it just got lost in housekeeping at some point.

PR #9173, opened 2025-08-08
Add Nabla - Theme

aelmoghit:

My package is a theme and color scheme for ST4

[2]: https://semver.org

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -2,7 +2,7 @@
 	"schema_version": "3.0.0",
 	"packages": [
 		{
-			"name": "Nabla - Theme",
+			"name": "Nabla Theme",
 			"details": "https://github.com/aelmoghit/Nabla-Theme",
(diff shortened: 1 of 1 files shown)


braver:
Oh, that makes sense, you’re changing two settings at once that way. I’ll double check a couple more of things later today and will probably merge it then.

repository/n.json
```diff
@@ -1,6 +1,17 @@
 {
 	"schema_version": "3.0.0",
 	"packages": [
+		{
+			"name": "Nabla - Theme",
```
braver: For consistency in the eco system, the naming would be "Theme - Nabla", or "Nabla Theme" (without the dash) or simply "Nabla". There is no policy here, just something to be aware of.

aelmoghit: Thanks for the clarification. I have made a few changes for consistency in the ecosystem.

PR #9174, opened 2025-08-09
Add Telescope

Mister7F:
- [ ] I use [.gitattributes](https://www.git-scm.com/docs/gitattributes#_export_ignore) to exclude files from the package: images, test files, sublime-project/workspace.

My package is similar to the "live grep" feature of Telescope (the VIM plugin)

https://github.com/Mister7F/sublime-telescope

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -545,6 +545,19 @@
 			]
 		},
 		{
+			"name": "Telescope",
+			"details": "https://github.com/Mister7F/sublime-telescope",
+			"releases": [
+				{
+					"sublime_text": ">=4198",
+					"tags": true
+				}
+			],
+			"labels": [
+				"code navigation"
+			]
+		},
+		{
 			"name": "Templ (go)",
 			"details": "https://github.com/papierkorp/sublime-templ",
 			"labels": ["templ", "golang", "go"],


Mister7F:
A similar package already exists

kaste:
Which package do you refer to?

kaste:
Oh, thanks, you mean my package.  Yeah, Find-in-Files is generally so fast, I use rg only to search on whole trees, supersets, anything that is not a project and already indexed.  

To have a live re-load would be great.  Techically, one input_panel, on typing you close that and initiate a find-in-files search like my plugin, then immediately show the input panel again.  If you use the InputPanel at the top of the view, maybe even easier, if a FiF search does not close such an input "box".  Actually, sounds like fun to try, now that I'm writing this. 

(1) Mhm, for the quick panel.  You can just close the current and open a new panel.  Unless you really have a large amount of items, Sublime itself is fast at rendering.   Serializing the items (Python->C++) likely takes more time than the drawing.

(2) For the search flags.  If you start the Find panel with a clear defined state, you know their values.  That might be even enough, but you could also listen for the commands (eg. alt+r).  When mouse actions do not fire, mousers are out.

Mister7F:
> (1) Mhm, for the quick panel. You can just close the current and open a new panel

That's what I call duct tape 😅 eg, if your cursor is not at the end, then it will move to the end when it refresh.

>  If you use the InputPanel at the top of the view

Is that possible ?

>  If you start the Find panel with a clear defined state, you know their values.

But then, I'll reset the value each time the sublime text restart, etc, or save the state somewhere, etc. It's not impossible, but the solution won't be clean. I prefer waiting for the API changes to have something clean

kaste:
Sometimes it is advisable to have a very good Python implementation that rocks and they improve the performance afterwards.  That would be not very different to other jobs, first the drop-in and the clean implementation.

> That's what I call duct tape 😅 eg, if your cursor is not at the end, then it will move to the end when it refresh.

Isn't that just that you *also* listen for on_selection_modified. So on_modified plus on_selection_modified...

> ... InputPanel ... at the top
That's called differently.  TextInpuHandler.  Con is their API is very, very unfriendly.  I deeply hate these two handlers.

Mister7F:
@kaste This is very frustrating, we can see how good it can be with that new component...

https://drive.google.com/file/d/1mRrG7TSo3zaJxQreSLrFPgnOa1PzHK-O/view?usp=sharing

But it's terrible to use with all those hacks

kaste:
Interesting.  For that flow, I would rather say, one input panel at the bottom (the old api), two views using a temporary split (forgot how that is named in Sublime... selected sheets?), left for the search results, right one for the preview.  Then the window layout doesn't change until you exit this mode.

Mister7F:
That’s what I did just before the poc with the quick panel. Not with a simple panel but using the « find in files » panel (and I also have a commit with an io panel, so no splited view for that third poc), that way you can enable regex mode using that, etc.
It work well, but sublime is not fast enough for a quick « live reload » on large project. So I need ripgrep with aggressive filter, and so I need to know if the regex button is on or off (that’s why I created the second github issue I listed above).

However, I think if we are ok with a slow update, it should be doable without ripgrep, and with a very minimal code (abusing the existing « find in files sidebar ». But then it’s super similar to your addon (except the live reload, which could be an option) hence my conclusion to close the pr.

Mister7F:
@kaste Hi, they did a similar hack for "LSP go to symbols in workspace", so I rework the plugin to use the same code

https://github.com/sublimelsp/LSP/blob/main/plugin/core/input_handlers.py

The LSP component has the "selection goes to the end when it refresh" bug, that I fixed in https://github.com/Mister7F/sublime-telescope/blob/master/telescope.py#L106

I'm open for a review / suggestions

I really hope the will implement the component as a core feature, they are still flickering, which are very annoying :/ and it's not as responsive as other part of the UI, but at least, the code I have to maintain is smaller, so that's ok for now

Cheers

kaste:
Very good.  I needed to make a quick and dirty patch to run it on Windows.  `head` is not available, the quoting of the path was actually too much, just "" is enough, and of course the absolute path names have the form "C:\some\path:15:thecontent", so we have one ":" more than you expect.

```diff
 telescope.py | 8 ++++----
 1 file changed, 4 insertions(+), 4 deletions(-)

diff --git a/telescope.py b/telescope.py
index 61eab81..75b6ec7 100644
--- a/telescope.py
+++ b/telescope.py
@@ -212,11 +212,10 @@ def _live_search(window, search_query, extension):
         """
         rg -t %(extension)s --no-heading --max-filesize 100M --max-count 10000 --follow --line-number --fixed-strings %(rg_search)s %(base_dir)s
         | fzf --filter %(search_query)s
-        | head -n50
         """.strip().replace("\n", " ")
         % {
             "extension": shlex.quote(extension),
-            "base_dir": " ".join(shlex.quote(d) for d in window.folders()),
+            "base_dir": " ".join(f'"{d}"' for d in window.folders()),
             "rg_search": " ".join(f"-e {shlex.quote(p)}" for p in rg_search),
             "search_query": shlex.quote(search_query),
         }
@@ -226,7 +225,8 @@ def _live_search(window, search_query, extension):
     for i, line in enumerate(os.popen(cmd).readlines()):
         if not line.strip():
             continue
-        path, line_number, content = line.split(":", 2)
+        drive, path, line_number, content = line.split(":", 3)
+        path = drive + ":" + path
         to_trim = next((i for i, s in enumerate(content) if s.strip()), 0)
         content = content.strip()
 
@@ -337,7 +337,7 @@ def _get_valid_file_extensions_update_cache(window):
 
     glob = " ".join(f"--iglob {shlex.quote('!' + e)}" for e in exclude_patterns)
     cmd = (
-        """
+        r"""
         rg  --no-heading --follow --max-filesize 100M --files --fixed-strings %(glob)s %(base_dir)s | sed -n 's/.*\(\.[^/]*\)$/\\1/p' | sort -u
         """.strip().replace("\n", " ")
         % {
```

Mister7F:
I thought Python would be slower than that; after some benchmarking, it's even sometimes faster than spawning a new process, so I pushed a new commit to remove the usage of `sed` and `head`, and I did it manually (120 ms vs 110 ms on my large project, when removing head eg)

kaste:
This is all *very* good.  I would

```
diff --git a/telescope.py b/telescope.py
index e948b05..2d7863f 100644
--- a/telescope.py
+++ b/telescope.py
@@ -71,6 +71,7 @@ class GlobsInputHandler(sublime_plugin.TextInputHandler):
             return current_globs[self.window]
 
         if (file_name := self.window.active_view().file_name()) and "." in file_name:
+            sublime.set_timeout(self._select_and_reset)
             return file_name.rsplit(".", 1)[-1]
 
     def _select_and_reset(self) -> None:
```

IMO a better first start.  

I don't know how to unset the glob, however.  *rg* has good filters by default, e.g. it ignores git and .sublime-workspace and so on.  And I would prefer that *initially* over any filter.

Mister7F:
>  it ignores git and .sublime-workspace and so on. And I would prefer that initially over any filter

Good idea :) 

So now by default, it will use no filter (except the ripgrep default one)

> I don't know how to unset the glob, however. 

You can just press enter with the input empty to have no filter, I updated the readme

repository/t.json
```diff
@@ -544,6 +544,19 @@
...
+			"details": "https://github.com/Mister7F/sublime-telescope",
+			"releases": [
+				{
+					"sublime_text": ">=4198",
+					"tags": true
+				}
+			],
+			"labels": [
+				"code navigation"
+			]
```
braver, line 558:
```suggestion
			"labels": ["code navigation"],
			"releases": [
				{
					"sublime_text": ">=4198",
					"tags": true
				}
			]
```

Just to follow conventional formatting here. Otherwise looks good 👍🏻

braver:
It's somewhat common practice to also expose keybinding setting via the command palette (in addition to the Package Settings menu entry you already have). Something to consider. Package looks good otherwise 👍🏻

PR #9178, opened 2025-08-11
add ColorConverter

braver:

My package is [ColorConverter](https://github.com/braver/ColorConverter). It is a modern converter of (CSS) colors, supporting the most commonly used formats, including relatively new ones like `lab()` and `color()`, and uses the current recommended notation for familiar formats like `rgb()`. Under the hood it uses the [coloraide](https://facelessuser.github.io/coloraide/) library by @facelessuser.

This new package intends to consolidate several legacy packages currently available in Package Control. All of them attempt to implement similar feature sets. However, they are roughly between 7 and 12 years old and don't support modern formats. Furthermore they all lack either a feature ("convert all"), don't support a certain format, have outstanding bugs, and/or simply no longer install or work correctly in recent versions of ST4/py3.8. As it stands, there currently is no reliably working and future-proof alternative out there (excluding some larger packages like ColorHelper of course, but their scope and complexity is in many cases unnecessary), which led me to build something new to replace them all.  
This is also why the package comes with a context menu, as that's convenient and will be expected by users of the replaced packages. There is a section in the README about how to customize or completely remove the menu.  
Since upgrading to this package will break key bindings anyway, the package does not ship with any enabled. But there is an example in place that can serve as a template for users to set up their own.

~The version requirement is relatively steep because at least some API features are used that were introduced in 4181, and I don't want to go back and test against older versions. You'll need a "this year" version of ST to use this. I assume this means that users of old versions of ST, with existing versions of the replaced packages, get to keep those.~
Edit: I was able to drop the version requirement to 4143, from 3 years ago. Should be good for pretty much anyone who still actually uses ST.

Pinging authors of existing packages for approval of the replacement: @obetame, @TheDutchCoder, @vitorleal, @aroscoe, @atadams.

@FichteFoll or @kaste perhaps one of you can serve as reviewers in this instance 😄

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2849,7 +2849,7 @@
 			],
 			"releases": [
 				{
-					"sublime_text": ">=4151",
+					"sublime_text": ">=4143",
 					"tags": true
 				}
 			]
--- a/repository/c.json
+++ b/repository/c.json
@@ -2849,7 +2849,7 @@
 			],
 			"releases": [
 				{
-					"sublime_text": ">=4189",
+					"sublime_text": ">=4151",
 					"tags": true
 				}
 			]
--- a/repository/c.json
+++ b/repository/c.json
@@ -2558,16 +2558,6 @@
 			]
 		},
 		{
-			"name": "Color Convert",
-			"details": "https://github.com/zhouyuexie/ColorConvert",
-			"releases": [
-				{
-					"sublime_text": "*",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "Color Highlight",
 			"labels": ["color", "highlight", "highlighter", "hex", "rgb", "hsl"],
 			"details": "https://github.com/Kronuz/ColorHighlight",
@@ -2847,6 +2837,24 @@
 			]
 		},
 		{
+			"details": "https://github.com/braver/ColorConverter",
+			"donate": "https://paypal.me/koenlageveen",
+			"labels": ["css", "colors", "utilities"],
+			"previous_names": [
+				"Color Convert",
+				"CSS Color Converter",
+				"Hex to HSL Color Converter",
+				"Hex to RGB Converter",
+				"Hex-to-RGBA"
+			],
+			"releases": [
+				{
+					"sublime_text": ">=4189",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Colored Comments",
 			"details": "https://github.com/TheSecEng/ColoredComments",
 			"author": "Terminal / TheSecEng",
@@ -2873,6 +2881,8 @@
 		},
 		{
 			"details": "https://github.com/braver/ColorHints",
+			"donate": "https://paypal.me/koenlageveen",
+			"labels": ["css", "colors", "utilities"],
 			"releases": [
 				{
 					"sublime_text": ">=3200",
@@ -4340,16 +4350,6 @@
 			"releases": [
 				{
 					"sublime_text": "<3000",
-					"branch": "master"
-				}
-			]
-		},
-		{
-			"name": "CSS Color Converter",
-			"details": "https://github.com/TheDutchCoder/ColorConvert",
-			"releases": [
-				{
-					"sublime_text": "*",
 					"branch": "master"
 				}
 			]
--- a/repository/h.json
+++ b/repository/h.json
@@ -450,16 +450,6 @@
 			]
 		},
 		{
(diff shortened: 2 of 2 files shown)


braver:
Don't ask me why the bot doesn't see the addition 🤦🏻

braver:
> Can we just "update" old users to a "random" other package?

No, not to a random package. When we review for these kinds of replacements, we require that the scope and approach of the new package is in line with the original package. In other words, does this feel, for the end user, as a logical continuation of what they were using before. I believe that is the case in this situation. It would be something else to introduce a massive UI driven package like ColorHelper as a replacement of packages that weren't much more than a keyboard shortcut before. The features added here, compared to the more limited packages I'm intending to replace, are all on the wish list for those packages: support a few more color formats (but not _all of them_), convert all colors in a document or selection. 

> Should the registry be ever-growing -- in particular, should the website be only growing

That's not been our policy no. One reason is that it's very hard to fight the "SEO" of a pre-existing package. So if the old package is effectively abandoned, and the new package has a fresh take on the same concept, a replacement is better for everyone in the end. I believe in this case all packages to be replaced are if not completely abandoned at least somewhat forgotten. The most recently updated one (Color Convert) does not work under ST4 + py3.8, and the others are very very old and have open issues that have not been responded to in years. 

I'll say it's very rare that one new package absorbs several old ones. In this case, I tried all of them. Several of them seem to be copies of each other, as the implementation is very similar. None of them really work anymore. So, to prevent future users looking for a similar solution from having to try 5 packages to find one (or none) that works, this is a positive move IMO. But of course I'm biased here 😅 

> Is there such a right for a package name

There is, but only up to a point. If any of the authors of the packages I'm replacing objects, they'll certainly get to keep their package in the registry. They got there first, they can keep the name if they want it. We do ask they respond here in a few weeks, because we see quite often that users are simply not active on GitHub anymore. So you have a right, but that comes with a little responsibility as well.

> What do I actually get to install, when you remove all these packages?

That I don't know for sure how that plays out. I guess once these ancestor packages are removed, you don't get to install them. I don't know what the cut-off point is, or the circumstances, where they stop working. So on older versions of ST, you do get a bit of a limbo state... perhaps these packages are no longer available to you, but good chance they didn't really work in the first place either.
Another thing I checked was the install charts on pc.io, and these all see very few new installations if any at all. Color Convert does, but again, it doesn't actually work. Perhaps that's a py3.8 thing.
If that's a problem, I can try to make a version that's compatible with an older version of ST... but I'm not getting payed here and there is only so much I can actually test for. I never test compatibility with old ST versions for any package I make. Here I just noticed a 4181 api feature I was using in the api docs.

repository/c.json
```diff
@@ -2846,6 +2836,24 @@
...
+			"previous_names": [
+				"Color Convert",
+				"CSS Color Converter",
+				"Hex to HSL Color Converter",
+				"Hex to RGB Converter",
+				"Hex-to-RGBA"
+			],
+			"releases": [
+				{
+					"sublime_text": ">=4189",
```
kaste: What else do you use actually.  Isn't that just `find_all` which is in 4181?  `dependecies.json` doesn't need the version spec, just set "*", as the package is installed on the correct version anyway.

braver: I added a version check since the "within" thing is just a minor feature. That leaves 4132 as a (known) minimum, the next public build for which is 4142, is almost 3 years old now. I took the 4143 bug fix build that came right after. Licenses expire after 3 years anyway.. seems like a fair deal.

braver:
```
Packages removed:
  - CSS Color Converter
  - Color Convert
  - ColorHints
  - Hex to HSL Color Converter
  - Hex to RGB Converter
  - Hex-to-RGBA
  ```
  
Ugh that bot... I'm obviously _not_ removing ColorHints.

repository/c.json
kaste: Ah, is that another unwritten rule that we refer only final versions and not dev builds?

braver: Not really no, no rule. But I'm being (overly) careful and assume the new feature might not work correctly in non-final builds. I also assume only "5 random cooks" use dev builds ;)

kaste: 🤣

kaste:
LGTM

braver:
Hmmmm, that’s smart… so the context menu command triggers the base command, and the base command being in the command palette can use an InputHandler? Hmmmmm!

Thanks!

Maybe you should be doing more of these package submission reviews, you’re good at it! 😉😄

PR #9179, opened 2025-08-11
del RAML Syntax Highlighter

braver:
Crawler fail says "your IP address is not permitted to access this resource", plus the repo has been archived ... they really don't want us to have it anymore?

@kaste ?

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -325,17 +325,6 @@
 			]
 		},
 		{
-			"name": "RAML Syntax Highlighter",
-			"details": "https://github.com/mulesoft-labs/raml-sublime-plugin",
-			"labels": ["language syntax"],
-			"releases": [
-				{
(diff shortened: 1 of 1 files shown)


kaste:
Nah, that failing is a recent change.  They just applied an org wide setting, and we can't tell them because they closed the issues.  Here "archived" really means "frozen": we don't work on it anymore!  Fork it!

braver:
Ah, but because the crawler can't access it, the archived bit doesn't flip.

kaste:
Reading `archived_at` is a recent addition.  In theory, we could patch the workspace to fix that.  

BTW the other package with the same message also was seen last around the same date, a day before or after.  Likely, GitHub sent a mass mail and two companies flipped a flag org-wide.

braver:
Gotcha. So this is maybe a crawler todo.

PR #9180, opened 2025-08-12
Add Tmux 3 syntax hightlight

Edditoria:

My package is Tmux 3 Syntax Highlight. It supports Tmux v3.

My package is similar to "Tmux Syntax Highlight". However I still decided to submit my own one because:

1. My package uses `.sublime-syntax` rather than `.tmLanguage` (ST3+).
2. My package follows the official Tmux man-page, and deprecates v2 syntax.
3. I will maintain my own repo anyway because I need to share it in my dotfiles using MIT license.
4. I would like to control releases (tags) myself.

I understand the instruction about package name. However I still decided to use the name "Tmux 3 Syntax Highlight" because:

- There is an existing package named "Tmux" (but not related to syntax).
- There is an existing package named "Tmux Syntax Highlight". Adding "3" behind Tmux is better than adding "ST3+" (confusing).

Thank you for your review.

[2]: https://semver.org

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -2655,12 +2655,12 @@
 		},
 		{
 			"name": "Tmux Syntax Highlight",
-			"details": "https://github.com/kei-q/sublime-tmux-syntax-highlight",
+			"details": "https://github.com/Edditoria/tmux-sublime",
 			"labels": ["language syntax"],
 			"releases": [
 				{
-					"sublime_text": "*",
-					"branch": "master"
+					"sublime_text": ">=3000",
+					"tags": true
 				}
 			]
 		},


repository/t.json
```diff
@@ -2653,6 +2653,18 @@
 				}
 			]
 		},
+		{
+			"name": "Tmux 3 Syntax Highlight",
+			"details": "https://github.com/Edditoria/tmux-sublime",
+			"author": "Edditoria",
```
braver:
```suggestion
```

Since this is your GitHub username, it's better not to specify it. Package Control will pick it up automatically.

Edditoria:
Thanks for your review.

In this repo:

- My name is removed as requested.
- Renamed the package to "Tmux Syntax" like other packages I'd seen in Package Control.
  - I think it is nicer for more features in future.
  - If you decide to keep the existing one from awesome @kei-q . I would pick the name "Tmux 3 Syntax Highlight" again, so users can search and see the difference on screen easier.

In my repo:

- As requested, the file `package-control.json` is added to `export-ignore`.
- Unfortunately, I couldn't copy the `.tmLanguage` from that repo (that does not have a license). The policy in my company is quite strict :-(
- For fallback file, may be do-it-later but depends on user feedback.
- Version is dumping to v1.0.1

Thank you again. Let's go!

braver:
Awesome, thanks. Let's give it two weeks. If there is no response from @kei-q I would like to go ahead and replace their package with yours.

repository/t.json
```diff
@@ -2653,6 +2653,17 @@
 				}
 			]
 		},
+		{
+			"name": "Tmux Syntax",
```
braver:
```suggestion
			"name": "Tmux Syntax Highlight",
```

Now that the 2 weeks are over and we've heard nothing, let's go ahead with the replacement. You can rename your package and then remove the existing entry for "Tmux Syntax Highlight".

Edditoria:
I updated `t.json` according to your last review.
Then, I rebased to latest commit of your master and squashed 3 commits into 1.

I think now it is okay to merge, hopefully.

Thank you!! Let's go!!

PR #9182, opened 2025-08-13
new fork for CoolFormat

edokan:

My package is similar to CoolFormat; however, it should still be added because I added support for Intel Silicon (ARM64).

[2]: https://semver.org

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3662,6 +3662,17 @@
 			]
 		},
 		{
+			"name": "CoolFormat (OSX-ARM64 Support)",
+			"details": "https://github.com/edokan/Sublime-CoolFormat",
+			"labels": ["formatting", "tidy"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Copy as HTML",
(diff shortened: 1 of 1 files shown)


edokan:
can @packagecontrol-bot review?

edokan:
That sounds viable. I do not want to *steal* their work, just want to make it work. I put the necessary files into my clone of the repository and updated the relevant issue on the original repository for a workaround, so two weeks should not be an intermediate issue. Thank you for your time.

PS: I think, I also need to update the version so people can update.

braver:
> I do not want to steal their work

Sure, I get that. Just make sure credits are given to the original authors. Ideally you can help continue the work in the original repo, but if they don't respond to issues and questions, sometimes that's just not an option. We believe pointing the entry in Package Control to the latest implementation of a package is the best way of keeping things up to date for all users, and prevents the duplication that is so much of an issue in many similar package ecosystems.

braver:
@akof1314 we're about to replace your CoolFormat entry in Package Control with the form by @edokan. Since you have some unattended issues and PR's on the project, we kind of assume that's fine with you, but please let us know if otherwise.

braver:
Looks like it! 👍🏻

PR #9191, opened 2025-08-26
Add package: MoveSyntax (language syntax)

Haalioa:
 - [ ] I use [.gitattributes](https://www.git-scm.com/docs/gitattributes#_export_ignore) to exclude files from the package: images, test files, sublime-project/workspace.

My package is a language syntax and comment definition for NestedText.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -2568,6 +2568,16 @@
 			]
 		},
 		{
+			"name": "MoveSyntax",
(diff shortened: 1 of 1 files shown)


braver:
Did you pick the name MoveSyntax because "Move" as just the name for the language would not be clear enough? You need to sort it correctly though, that's what the bot's failing over.

braver:
Also note that we already have a package for NestedText: https://packages.sublimetext.io/packages/NestedText/

PR #9193, opened 2025-08-28
Update t.json

vivekgohel2004:
Fix: update Text-Mate Theme entry (sublime_text version)

[2]: https://semver.org

My package is ...

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -3921,6 +3921,18 @@
 			]
 		},
 		{
+			"name": "Text-Mate Theme",
+  			"details": "https://github.com/vivekgohel2004/Text-Mate-Theme",
+  			"author": "vivekgohel2004",
+  			"labels": ["theme"],
+  			"releases": [
+    			{
+      		 	"sublime_text": ">=3000",
+      			"tags": true
+    			}
+  			]
+		},
+		{
 			"name": "Typst",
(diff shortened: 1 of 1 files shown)


repository/t.json
```diff
@@ -3920,6 +3920,18 @@
 				}
 			]
 		},
+		{
+			"name": "Text-Mate Theme",
+  			"details": "https://github.com/vivekgohel2004/Text-Mate-Theme",
+  			"author": "vivekgohel2004",
+  			"labels": ["theme"],
```
braver, line 3927:
```suggestion
			"name": "Text-Mate Color Scheme",
  			"details": "https://github.com/vivekgohel2004/Text-Mate-Theme",
  			"labels": ["color scheme"],
```

As the file extension indicates, in Sublime Text parlance your package is not a "theme" (which is for UI elements only) but a "color scheme" (syntax highlighting).

You don't need to specify the author if it's your GitHub username.

Finally, Text-Mate is a bit of an unfortunate name because [TextMate](https://macromates.com) is an editor that ST (and VS Code) borrows a lot of config file formats from. Like the `tmTheme` file format ("Text Mate Theme")... using the words "Text Mate" in your package name without it having a relation to the TextMate editor is going to be really confusing for a lot of people 😅

PR #9195, opened 2025-09-02
Add ShareToS3 package

jamesacklin:
Adds a Sublime Text package which allows the user to upload whatever is in their current buffer to a .txt file in an S3 bucket, then copy the resulting URL to their clipboard.

My package is similar to the following packages:

- https://github.com/benanders/Rust-Share
- https://github.com/devdinu/ShareFile
- https://github.com/geekpradd/sublime-sourcesharer-plugin
- https://packagecontrol.io/packages/Cloudup
- https://packagecontrol.io/packages/Just%20Paste
- https://github.com/jonasdp/Snipplr
- https://github.com/camperdave/EC2Upload
- https://src.grytoyr.io/kim/npaste (404)

However, it should still be added because these packages all interact with proprietary or single-party file uploading/sharing services. My plugin uses any S3-compatible endpoint (MinIO, AWS S3, DigitalOcean storage, etc).

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1384,6 +1384,17 @@
 			]
 		},
 		{
+			"name": "ShareToS3",
+			"details": "https://github.com/jamesacklin/ShareToS3",
+			"labels": ["share", "snippets", "S3", "upload"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Shebang",
 			"details": "https://github.com/samizdatco/sublime-text-shebang",
 			"releases": [


jamesacklin:
LICENSE file added.

braver:
Please use edit_settings for your preferences, like so: https://github.com/braver/ColorConverter/blob/main/Main.sublime-menu#L17

It's common practice to expose settings via the command palette as well ([example](https://github.com/braver/ColorConverter/blob/main/Commands.sublime-commands#L3)).

braver:
Package and entry here look good by the way 👍🏻

jamesacklin:
@braver Apologies, just updated. Thanks for the review!

braver:
Thanks! Don't forget to tag your latest version.

PR #9196, opened 2025-09-02
Add LUC package

isma1062:

[2]: https://semver.org

My package is ...


repository.json
```diff
@@ -33,3 +33,18 @@
 		"./repository/dependencies.json"
 	]
 }
+        ,
+        {
+            "name": "LUC",
+            "description": "Soporte completo para los lenguajes LUC y AtomScript (Coloreado, Snippets, Ejecución)",
+            "author": "TuNombreDeUsuario",
+            "homepage": "https://github.com/TuNombreDeUsuario/sublime-luc-package",
```
braver, line 41:
```suggestion
```

These lines are not needed, they'll get resolved from the `details`.

PR #9202, opened 2025-09-13
 s.json

abharkhalif711-hue:
- [ x ] I'm the package's author and/or maintainer.
- [ x ] I have have read [the docs][1].
- [ x ] I have tagged a release with a [semver][2] version number.
- [ x ] My package repo has a description and a README describing what it's for and how to use it.
- [ x ] My package doesn't add context menu entries. *
- [ x ] My package doesn't add key bindings. **
- [  ] Any commands are available via the command palette.
- [  ] Preferences and keybindings (if any) are listed in the menu and the command palette, and open in split view.
- [ x ] If my package is a syntax it doesn't also add a color scheme. ***
- [ x ] I use [.gitattributes][3] to exclude files from the package: images, test files, sublime-project/workspace.

[2]: https://semver.org

My package is ...

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -6120,6 +6120,18 @@
 					"tags": "st3-"
 				}
 			]
-		}
+		},
+		{
+			"name": "superclouds",
+			"details": "https://github.com/abharkhalif711-hue/supercloud-dark-theme",
+			"description": "a supercloud is soft dark theme",
+			"labels": ["color scheme"],
+			"releases": [
+			{
+				"sublime_text": "*",
+				"tags": true
+			}
+		]
+	  }
 	]
 }


repository/s.json
```diff
@@ -6120,6 +6120,18 @@
 					"tags": "st3-"
 				}
 			]
-		}
+		},
+		{
+			"name": "superclouds",
+			"details": "https://github.com/abharkhalif711-hue/supercloud-dark-theme",
+			"description": "a supercloud is soft dark theme",
```
braver, line 6127: Just some minor changes are needed:

instead of adding your entry to the end, add it in chronological order around line [5337](https://github.com/abharkhalif711-hue/package_control_channel/blob/d6268e57124c9c638efec48e8352503287f8d80d/repository/s.json#L5337).

Instead of adding a description here, add a description on your GitHub repo:

Image

Your readme refers to it as "Supercloud", ie. singular and capitalized, so match there here as well.

```suggestion
			"name": "Supercloud",
			"details": "https://github.com/abharkhalif711-hue/supercloud-dark-theme",
```

braver (changes requested):

repository/s.json
abharkhalif711-hue: ok,

braver: Let me know when you've been able to handle the feedback, or if you need any help.

braver:
Let me know if you want to restart the review.

PR #9210, opened 2025-10-03
Update t.json for Transpy

cumulus13:
add "Transpy" - Sublime Text Translation Plugin

My package is a translation plugin for Sublime Text.  
It allows users to translate selected text between languages using online translation services.

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -3370,7 +3370,7 @@
 		{
 		    "name": "Transpy",
 		    "details": "https://github.com/cumulus13/Transpy",
-		    "labels": ["translation", "utility"],
+		    "labels": ["translate", "utilities"],
 		    "releases": [
 		        {
 		            "sublime_text": "*",


repository/t.json
```diff
@@ -3367,6 +3367,18 @@
 				}
 			]
 		},
+		{
+		    "name": "Transpy",
+		    "details": "https://github.com/cumulus13/Transpy",
+		    "labels": ["translation", "utility"],
```
braver: [translate](https://packages.sublimetext.io/?q=label%3Atranslate) is actually the more common label, you can search on our labels [overview](https://packages.sublimetext.io/labels/).

braver:
Please use the `edit_settings` command for settings and keybinding files ([example](https://github.com/braver/ColorConverter/blob/main/Commands.sublime-commands#L4)), and add your menu entries to the correct section in the main menu ([example](https://github.com/braver/ColorConverter/blob/main/Main.sublime-menu#L13)). If you made your package with help of an LLM make sure all features actually work as intended. 

Finally, please follow the [PR template](https://github.com/wbond/package_control_channel/blob/master/.github/PULL_REQUEST_TEMPLATE.md) and make sure you check all boxes properly (like using semver numbering for releases).

braver:
Let me know if you need any help processing the feedback. Especially your `Preferences.sublime-menu` is currently broken.

PR #9212, opened 2025-10-05
Migrate packages from my private repository.json

FichteFoll:
This has not been necessary since schema version 3.0 and all my new
packages have been added to the channel directly, but I never migrated
my old ones until now.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -4561,6 +4561,21 @@
 			]
 		},
 		{
+			"name": "CSScheme",
+			"labels": ["color scheme editing"],
+			"details": "https://github.com/FichteFoll/CSScheme",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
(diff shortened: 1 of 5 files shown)


repository/c.json
```diff
@@ -4560,6 +4560,21 @@
...
 			]
 		},
+		{
+			"name": "CSScheme",
+			"labels": ["color scheme editing"],
+			"details": "https://github.com/FichteFoll/CSScheme",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": "v"
```
braver, line 4570: I thought the v prefix was part of semver and as such not required to specify (ie. you could just get away with `true` here)? Not that it needs changing, just curious.

FichteFoll: That is correct. The `v` prefix here is optional, but since I am using it I figured I might be explicit.

PR #9213, opened 2025-10-08
Update c.json

tonynupe:
Add CiscoCollab package to Package Control

This package provides custom syntax highlighting, color schemes, and mouse bindings for Cisco-style configurations in Sublime Text. Includes dark/light variants and auto theme switching.

it's a work in progress by Cisco TAC engineers

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1158,6 +1158,17 @@
 			]
 		},
 		{
+			"name": "CiscoCollab",
+  			"details": "https://github.com/tonynupe/CiscoCollab",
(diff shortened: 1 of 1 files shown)


braver:
Oh, you named the release correctly, but the important thing here is the tag and that just says “version”.

tonynupe:
Ah ok I see it. I thought it was a tag to label the version. any
suggestions to what should it be here and for future reference?

tonynupe:
Apologies again, I have corrected it to match the version 1.0.0

PR #9215, opened 2025-10-12
Add XmlTransformer v1.0.0

haloway13:
Adds XmlTransformer v1.0.0 to the main Sublime Text package channel

- [x] I'm the package's author and maintainer (haloway13).
- [x] I have read https://docs.sublimetext.io/guide/package-control/submitting.html.
- [x] Tagged release v1.0.0 with semantic versioning (https://semver.org).
- [x] Repo (https://github.com/haloway13/XmlTransformer) has a description and README.md detailing usage, installation, and features.
- [x] No context menu entries; uses build command (Cmd+B/Ctrl+B).
- [x] No custom key bindings; relies on default build key.
- [x] Commands (XmlTransformer: Build) available via command palette.
- [x] Settings (XmlTransformer.sublime-settings) accessible via command palette/menu, opens in split view.
- [x] Not a syntax package; no color scheme included.
- [x] Added .gitattributes to exclude test_xslt/, *.sublime-project, *.sublime-workspace from ZIP.

**Uniqueness**: No existing XSLT transformation plugins with Saxon-HE integration in Package Control.

**Details**:
Tested on Windows, Linux, and macOS with Sublime Text 3/4. MIT License. No bundled dependencies (external via setup scripts).

Features:
- XSLT transformations using Saxon-HE 12.9 and xmlresolver 6.0.6.
- Quick panel for XSL and parameter file selection.
- Manual parameter entry or XML files with validation.
- Cross-platform setup scripts and error handling.

Repo: https://github.com/haloway13/XmlTransformer
Tag: v1.0.0
ZIP: https://github.com/haloway13/XmlTransformer/archive/v1.0.0.zip (verified)

### DIFF
--- a/repository/x.json
+++ b/repository/x.json
@@ -172,15 +172,6 @@
 			]
 		},
 		{
-			"details": "https://github.com/haloway13/XmlTransformer",
-			"releases": [
-				{
-					"sublime_text": ">=3000",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "XmlDocs",
 			"details": "https://github.com/irrationalistic/sublime-xmldocs",
 			"labels": ["documentation", "c-sharp"],
@@ -192,13 +183,13 @@
 			]
 		},
 		{
-		    "details": "https://github.com/haloway13/XmlTransformer",
-		    "releases": [
-		        {
-		            "sublime_text": ">=3000",
-		            "tags": true
-		        }
-		    ]
+			"details": "https://github.com/haloway13/XmlTransformer",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
 		},
 		{
 			"name": "XMPFilter",


repository/x.json
```diff
@@ -182,6 +182,21 @@
...
+		    "releases": [
+		        {
+		            "sublime_text": ">=3000",
+		            "platforms": ["windows", "linux", "osx"],
+		            "version": "1.0.0",
+		            "url": "https://github.com/haloway13/XmlTransformer/archive/v1.0.0.zip",
+		            "date": "2025-10-12"
+		        }
+		    ]
+		},
```
braver: You only need to do it this way if you're actively building the release asset. Doesn't look like you do, this is just the 1.0.0 tag. You can drastically simplify, and that way you also don't need to come in here for any future versions:

```suggestion
		    "details": "https://github.com/haloway13/XmlTransformer",
		    "releases": [
		        {
		            "sublime_text": ">=3000",
		            "tags": true
		        }
		    ]
		},
```

braver (changes requested):

repository/x.json
```diff
@@ -182,6 +182,18 @@
 				}
 			]
 		},
+		{
+		    "name": "XmlTransformer",
+		    "description": "Sublime Text package for XSLT transformations using Saxon-HE.",
+		    "author": "haloway13",
+		    "details": "https://github.com/haloway13/XmlTransformer",
```
braver: Just so you know, we can go even further. The description is on your GitHub repo, the author is your GitHub username, and the package name matches the repo name.
```suggestion
		    "details": "https://github.com/haloway13/XmlTransformer",
```

haloway13: Thanks @braver
I am new to this!

haloway13:
I removed the file.

My bad, I was reading it as an internal dependency for my package. The setup scripts handle all external dependencies.

My apologies!

repository/x.json
```diff
@@ -182,6 +191,15 @@
 				}
 			]
 		},
+		{
+		    "details": "https://github.com/haloway13/XmlTransformer",
```
braver: Hey, now it's in here twice 😅

braver:
Ok, got it. That should be fine then.

Now that I look at it again though, users won't be able to run the install scripts because they'll be in your zipped packages. I.e. the script won't be at `~/.config/sublime-text/Packages/XmlTransformer/setup_XmlTransformer_ubuntu.sh`. They're not runnable.

braver:
No that makes sense. A package like this simply needs some setup to work. You might want to add an install message, so you can notify users that installed your package about some extra steps. You can send them to the right section in your readme and then since you've described it pretty clearly they should be able to figure that out. 

See: https://docs.sublimetext.io/guide/package-control/utilities.html#messages

haloway13:
> I think you're missing the messages.json that tells package control where your message lives.

Great catch!

Added messages.json and moved install.txt to messages/install.txt to enable post-install setup instructions for copying/running setup scripts.

Please let me know if you find anything else!

haloway13:
> Looks good, but you need a new tag. Package Control, as well as this bot here, only looks at your tags.

Done

haloway13:
> Well shit, I don't know why the bot doesn't see the messages.json.
> 
> Meanwhile, I ran some tests...
> 
> * instead of `platform.system()` you can also use `sublime.platform()` ([docs](https://www.sublimetext.com/docs/api_reference.html#sublime.platform))
> * you're running `sublime.load_settings()` right at the top, but the api might not be fully ready at that point. instead, consider using a getter (e.g. `def is_debug()`) instead of a constant

XmlTransformer v1.0.3
- Fixed platform.system() to sublime.platform() for platform detection.
- Centralized settings load with debug getter in XmlTransformer_build.py

braver:
For things like menu options inside Sublime you mean? I actually never looked into that to be honest.

PR #9211, opened 2025-10-04
Standardize some labels

braver:
Some one-off labels were found (https://github.com/packagecontrol/thecrawl/pull/137) that can be improved (typos, capitalization, etc), and some minor standardization like `minification` ➡️ `minify`.

PR #9097, opened 2025-04-21
Lowercase labels & a few label tweaks

michaelblyons:
Following the footsteps #4277

PR #9217, opened 2025-10-13
Add some extra labels

michaelblyons:
About half the label changes are for my own stuff, but I happened to notice some obvious improvements in a couple of the files.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -4,7 +4,7 @@
 		{
 			"name": "C Improved",
 			"details": "https://github.com/abusalimov/SublimeCImproved",
-			"labels": ["language syntax"],
+			"labels": ["language syntax", "c"],
 			"releases": [
 				{
 					"sublime_text": "*",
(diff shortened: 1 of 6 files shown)


michaelblyons:
> Like, if it's not obvious from the name your package is for python, put "python" in a label.

LSP-Terraform is tagged {`lsp` `terraform`} and nothing else.

> And it's good to avoid the most obvious typos and case issues (https://github.com/wbond/package_control_channel/pull/9211).

I think #9088, #9093, #9094, #9095, and #9097 demonstrate that I agree. Though I see some new duplicates for `auto-complete` that either I missed or were added since then.

> But nobody's going to hunt through 5k packages to correctly label all the ones that support a certain language.

You're probably right. But who knows?

> So, put between zero and say 5 labels on your package to help people understand what the package is about. Use the common labels if you can. We'll do a few interesting things with it on the site. But that's it IMO. This isn't metadata on the ecosystem, it's just tags.

Okay.

braver:
> But who knows

I certainly don't! 

Maybe instead of expectations there should be hope 😄 

I'm ok with merging this, it's fine, it's an improvement. Maybe what I'm saying is just, you making these changes and me reviewing them is just not the most useful and fulfilling use of our time and energy.

PR #9218, opened 2025-10-15
use st-package-reviewer-action

braver:
Context: https://github.com/packagecontrol/thecrawl/issues/66
Fixes: #8050

If it fails you need to click through on the failed test to get details, but it's there and mostly readable due to annotations formatting. 

Scherm­afbeelding 2025-10-15 om 13 57 26

Scherm­afbeelding 2025-10-15 om 13 57 38

### DIFF
--- a/.github/workflows/schema.yml
+++ b/.github/workflows/schema.yml
@@ -11,6 +11,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v5
-      - uses: packagecontrol/st-schema-reviewer-action@cad5e3795afc23ed6e2ee298d5c27b5a6a621b8d
+      - uses: packagecontrol/st-schema-reviewer-action@f0ed8819bd9058dfa9921e4a061a6a1b643339a2
         with:
           test_repositories: true


braver:
📦

.github/workflows/package.yml
```diff
@@ -0,0 +1,34 @@
+name: Package test
+
+on:
+  pull_request:
+  issue_comment:
+    types: [created]  # run when comment added with just "📦"
```
FichteFoll, line 6: I think the old trigger of requesting a review from the review bot was pretty neat, personally, and also less annoying by not having to create a full comment and then triggering (email) notifications as a result.

In that case, I would suggest to reduce this to `pull_request: [opened, synchronize, reopened, ready_for_review, review_requested]` and remove the `issue_comment` trigger entirely.

(Checking for the requested user like in the old pipeline would also be neat but not necessary since a new pipeline run should be less noisy compared to adding a comment like before. I also assume that github will include all pipeline runs referencing the latest commit of PR in that commit's "Checks" or however that should be phrased, but we will see that once the change is live.)

.github/workflows/package.yml
```diff
@@ -0,0 +1,34 @@
...
+          script: |
+            const pr = await github.rest.pulls.get({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: context.payload.issue.number
+            });
+            core.setOutput('base_sha', pr.data.base.sha);
+            core.setOutput('head_sha', pr.data.head.sha);
+
+      - uses: packagecontrol/st-package-reviewer-action@de58167cfca93775fcbf5061cce6e452b51ce7b6
```
FichteFoll, line 33: Please add a comment here that this commit hash refers to a specific branch and is not accessible from the default branch (yet). I can totally see this becoming a point of confusion in the future.

.github/workflows/package.yml
michaelblyons: So if there are changes in the repo of the package being audited, to re-trigger this scanner in PCC, you convert the PR to draft and back to ready?

FichteFoll: Requesting a review from @packagecontrol-bot (or anyone, depending on the `if` condition) would be easier, but yes, converting to draft and undraft would also work this way.

braver: My assumption was that the bot won't be in the picture anymore. I have no idea of how the bot works, have no access to the bot account, or any control over settings around that, so I figured we might as well abandon it. But if the bot still works, that works for me as well. I need to be able to trigger it as a reviewer, and the draft-undraft dance seems a bit cumbersome and roundabout to me. And will probably results in confusing emails/notifications for the package authors as well. Hence the idea of running on a comment.

braver: Also, by the way, what usually happens is authors need to tell me they're ready for it to be reviewed. Nobody can figure out how to make that happen (people sometimes try random comments asking for the bot to do something). So it's me who ends up having to trigger the review. Then, most of the time the review doesn't run, because running any workflow needs to be approved by a maintainer (for first time contributor). 

So, I'm involved anyway before anything happens, 9 times out of 10, and it will cause emails anyway. Maybe just the emoji is a little obscure, perhaps I'll come up with something better that's easy to document.

braver:
For now I'll keep the old review bot thing alongside the new one. That way they can be compared. And hopefully they'll say the same thing. That might be a way to carefully transition over to a new approach. That way we can also see if the "comment to run tests" thing is annoying, while the "ask the bot" option is also still there.

.github/workflows/package.yml
FichteFoll: > I need to be able to trigger it as a reviewer

The "bot" is effectively a normal github user that is a collaborator in this repo, which means that it will be suggested as a reviewer. My proposal would just be abusing this fact. The account does not actually need to do anything since we're only using it a as a filter as our action trigger, just like the previous action did.

braver: Hm, I'm not sure that approach will continue to work. With this new approach the bot will stop actually reviewing anything, and will stop posting comments. It might become a hassle to try and re-ask it for a review, since it might forever be stuck in limbo (ie. you'll see a yellow dot, not a 🔄 icon). We'll have to dismiss the request for review and then ask again. I foresee lots of clicking around. Also, with the bot no longer reviewing, it becomes conceptually less logical to ask it for a review. 

But we'll see. We can try a few things.

braver: [example of annoying limbo state](https://github.com/wbond/package_control_channel/pull/9213)

Screenshot 2025-10-20 at 10 50 51

There is no clear and easy way to get the bot to review that PR again. Maybe remove it as a review and add it again?

FichteFoll: True, if the bot does not post a review anymore, we might need to remove and re-request that way. I say we try out both ways, i.e. the request review and the comment, and see what fares better in practice, though if we keep the old review flow active for the time being, the review flow will probably work better.

PR #9220, opened 2025-10-17
add Task List plugin

noahcoad:
- [na] If my package is a syntax it doesn't also add a color scheme. ***
- [na] I use [.gitattributes][3] to exclude files from the package: images, test files, sublime-project/workspace.

My package is ...

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -2248,6 +2248,17 @@
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


noahcoad:
I meant to add another plugin (I wrote these like 6 yrs ago, finally getting around to publishing), but wanted to keep them as seperate pull requests

PR #9221, opened 2025-10-18
add Auto Open Notes package

noahcoad:

My package is ...

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -2248,6 +2248,17 @@
 			]
 		},
 		{
+			"name": "Auto Open Notes",
+			"details": "https://github.com/noahcoad/SublimeAutoOpenNotes",
+			"labels": ["notes.txt", "readme.txt", "readme.md", "readme"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Auto Refresh",
 			"details": "https://github.com/Wouterdek/AutoRefresh",
 			"releases": [


repository/a.json
```diff
@@ -2247,6 +2247,17 @@
 				}
 			]
 		},
+		{
+			"name": "Auto Open Notes",
+			"details": "https://github.com/noahcoad/SublimeAutoOpenNotes",
+			"labels": ["notes.txt", "readme.txt", "readme.md", "readme"],
```
braver, line 2253: Those aren't very common [labels](https://packages.sublimetext.io/labels/). Might I suggest

```suggestion
			"labels": ["notes", "note taking", "documentation", "readme"],
```

noahcoad: Excellent point, thanks! Will update.

noahcoad:
> This is great, I love the idea of opening a directory and immediately seeing the README.

aww thanks 😊

Curious if I may ask... I like the "Auto Open Notes" name bc I use notes and they could be different type of file names, but it really can open any files on opening in a folder and README is probably the most common form of notes, so I was debating with the names "Auto Open Files" or "Auto Open Readme" ... Do you think any of those would be easiest to find the package and understand for users?

braver:
This is just my opinion, but "auto open notes" didn't immediately mean anything to me, whereas "auto open readme" kind of does. But I have now seen what your package does and that already colors my opinion. That said, if you know it can auto-open a readme, you might immediately wonder "but how does it do that" and "perhaps it can auto-open something else too"? 
... I don't know, naming things is hard and definitely not my speciality 😅

repository/a.json
braver: Let me know when you've had the chance to look at this!

PR #9222, opened 2025-10-18
add Task List package

noahcoad:

My package is ...

BTW, thank you for this checklist, it really helped me clean things up

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -416,6 +416,17 @@
 			]
 		},
 		{
+			"name": "Task List",
+			"details": "https://github.com/noahcoad/SublimeTaskList",
+			"labels": ["todo", "gtd", "tasks"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Task Timer",
 			"details": "https://github.com/jcowgar/sublime-task-timer",
 			"releases": [


noahcoad:
> * Your example keymap is not referenced anywhere. Note that you can leverage the edit_settings command to make it easy for users to find your suggested bindings and copy them over.

Thank you! I'll add "Key Mapping" to user settings with the key mapped there.

> * Might I suggest using an icon that doesn't have a religious connotation for "to do"? E.g. ☑️ or ⚛️ if you want one from the purple set? If it's "just" and icon the start of David feels like an odd choice.

Thanks for calling that out! YES, will update. (had used the previous bc it was the closet 'empty check box' look that stood out, but great point on it being a religious symbol)

> * As someone doesn't give a hoot about ones preferred style of indentation, the tone in that comment reads a bit off... I'd edit that out before "going public"

Oh boy, thank you, gosh I wrote that like 10 yrs ago and just now getting around to publishing. Definitely will update, thank you.

Thank you for the close review! Really appreciate the excellent feedback.

noahcoad:
@braver Finally circled back on this — sorry for the long gap. Tagged `v0.1.1` with all three items addressed:

1. **Keymap** — dropped the unreferenced `Example.sublime-keymap` for a real `Default.sublime-keymap` shipping the suggested binding commented out, following the sublime-telescope pattern you pointed me at. Added a **Key Bindings** entry under Preferences > Package Settings > Task List via `edit_settings`, and both Settings and Key Bindings are in the command palette now.
2. **Icon** — the Star of David is gone; default "to do" is now ⚛️. Trying to keep it in a similar shape as other green/red/yellow/orange emojis. A few alternates are documented in the settings file and readme so users can swap easily.
3. **Tone** — that indentation comment is deleted.

Also untracked a stray `.DS_Store` and added a `messages/0.1.1.txt` so existing users know the default icon changed. Would you mind re-opening when you get a chance? Happy to open a fresh PR instead if that's easier.

braver:
Cool, thanks. I can't re-open from here so perhaps best if you can start a fresh PR.

noahcoad:
will do.

noahcoad:
Opened a fresh PR as suggested: #9567. Tagged `v0.2.0` there, with everything from this thread addressed plus a couple of new commands since. Thanks for the patience on this one.

PR #9223, opened 2025-10-18
add Insert Markdown Hyperlink package

noahcoad:

My package is similar to ... a few other Markdown packages that also have a command to insert a hyperlink

However it should still be added because ... Those other markdown packages insert just the '[]()' base template, as they do with any other markdown shortcut, like '' for images, '#' for headers etc. This package inserts hyperlinks much more intelligently by using the selected text/url placing it the right spot, or using a url in the clipboard, then grabs the title from the url if available, and sets the selected text and cursor in the spot most needed (title or url).

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -825,6 +825,17 @@
 			]
 		},
 		{
+			"name": "Insert Markdown Hyperlink",
+			"details": "https://github.com/noahcoad/SublimeInsertMarkdownHyperlink",
+			"labels": ["markdown", "hyperlink", "url", "link"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Insert Nums",
 			"author": ["jbrooksuk", "FichteFoll"],
 			"details": "https://github.com/SublimeText/InsertNums",


repository/i.json
```diff
@@ -824,6 +824,17 @@
 				}
 			]
 		},
+		{
+			"name": "Insert Markdown Hyperlink",
+			"details": "https://github.com/noahcoad/SublimeInsertMarkdownHyperlink",
+			"labels": ["markdown", "hyperlink", "url", "link"],
```
braver, line 830:
```suggestion
			"labels": ["markdown", "url", "link"],
```

`hyperlink` is not a label that already exists, and doesn't feel like it really adds much. It feels very 1990's to be talking about "hyperlinks" 😅

braver:
It sounds a little bit like what [MarkdownLink](https://packages.sublimetext.io/packages/MarkdownLink/) tries do to. But different enough to be a good addition to the ecosystem.

Only thing I see in the package itself is that the example keymap isn't used. You could link it in Package Settings (except you don't have any settings). For both settings and key bindings it's common practice to have them in the command palette as well, so you could consider that too.

repository/i.json
noahcoad: Thank you for the feedback. Will address.

noahcoad:
> Only thing I see in the package itself is that the example keymap isn't used. You could link it in Package Settings (except you don't have any settings). For both settings and key bindings it's common practice to have them in the command palette as well, so you could consider that too.

Can you say more? Thought the best practice was to not use a keymap by default (hence I had recommended keymap in there), but I'd certainly like to include a default package keymap settings w the one key mapped if that's okay.

And good point about having the keycap in the settings. I'll add that.

braver:
Let me know when you've had a chance to look at the feedback!

repository/i.json
braver: I'd love to publish your packages, but all 3 PR's have feedback on them that needs to be handled first.

PR #9225, opened 2025-10-18
Add package: Vue Tools Kit

Maung90:
- [✅] I'm the package's author and/or maintainer.
- [✅] I have have read [the docs][1].
- [✅] I have tagged a release with a [semver][2] version number.
- [✅] My package repo has a description and a README describing what it's for and how to use it.
- [✅] My package doesn't add context menu entries. *
- [✅] My package doesn't add key bindings. **
- [✅] Any commands are available via the command palette.
- [✅] Preferences and keybindings (if any) are listed in the menu and the command palette, and open in split view.
- [✅] If my package is a syntax it doesn't also add a color scheme. ***

[2]: https://semver.org

My package is **Vue Tools Kit**, an all-in-one toolkit for Vue 3 development in Sublime Text. It bundles a custom syntax, a vibrant neon color scheme, essential snippets, and a command to create new components, providing a complete and modern development environment.

My package is similar to other Vue syntax packages, but it should still be added because it provides an integrated "out-of-the-box" experience with a matching theme and time-saving tools specifically tailored for the Vue 3 Composition API and `<script setup>` workflow.

### DIFF
--- a/repository/v.json
+++ b/repository/v.json
@@ -851,6 +851,17 @@
 			]
 		},
 		{
+            "name": "Vue Tools Kit",
+            "details": "https://github.com/Maung90/vue-toolkit",
+            "labels": ["vue", "javascript", "theme", "snippets", "syntax"],
+            "releases": [
+                {
+                    "sublime_text": "*",
+                    "tags": true
+                }
+            ]
+        },
+		{
 			"name": "Vuln",
 			"details": "https://github.com/Shimmy/vuln",
 			"author": "Shimmy",


braver:
In any case the color scheme should be shipped separately. Color schemes should not be language-specific and syntaxes should not expect features from color schemes. Besides, people will want to use their preferred color scheme. Or someone might like your color scheme but perhaps they don't write any Vue.

As for the features you're adding for Vue, why can't they be added to the [Vue](https://packages.sublimetext.io/packages/Vue/) package for instance?

Maung90:
Hi Braver,

Thank you so much for the detailed feedback. I really appreciate you taking the time to review this.
The reasoning makes perfect sense. I understand the importance of separating the color scheme from the language features to keep packages focused and modular.
Here is my plan of action:
	I will close this pull request.
	I will create a new, separate repository and package submission just for the color scheme.
	I will find the main Vue package and prepare a new pull request there to contribute my language-specific features.

Thanks again for the guidance!

> In any case the color scheme should be shipped separately. Color schemes should not be language-specific and syntaxes should not expect features from color schemes. Besides, people will want to use their preferred color scheme. Or someone might like your color scheme but perhaps they don't write any Vue.
> 
> As for the features you're adding for Vue, why can't they be added to the [Vue](https://packages.sublimetext.io/packages/Vue/) package for instance?

braver:
Excellent 👍🏻

PR #9228, opened 2025-10-20
Added changes to add SelectionsToFiles Sublime Text package.

dtoader:

[2]: https://semver.org

My package is credited to Tito Bouzout.
It is released under the GNU General Public License
I've gotten permission to maintain it.

Text Selection is saved to new file using shortcut key.

Usage
1. Select some text..
2. Main menubar -> Selection -> Copy To New Files
3. CTRL/CMD+P Selections: Copy To New Files

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -959,8 +959,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"tags": true,
-					"details": "https://github.com/dtoader/SelectionsToFiles/tree/master"
+					"tags": true
 				}
 			]
 		},


kaste:
Hi! 👋  If that is a takeover you should actually point us to the original package.  Likely we would just update the details URL or so.  It should also be clear that you forked it from Tito.  GitHub would tell us in their UI, "forked from ...".  Tito might also mark it archived.  Is there an issue where he says that you can have it?  Changes to channel.json are not allowed and not necessary either.  Look how the other packages are registered.  "tags: true" is the option you likely want.

repository/s.json
```diff
@@ -954,6 +954,16 @@
 				}
 			]
 		},
+		{
+			"name": "SelectionsToFiles",
+			"details": "https://github.com/dtoader/SelectionsToFiles",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"details": "https://github.com/dtoader/SelectionsToFiles/tree/master"
```
braver:
```suggestion
			"details": "https://github.com/dtoader/SelectionsToFiles",
			"releases": [
				{
					"sublime_text": "*",
					"tags": true
```

Your details points us to the repo. Then set `tags` to tell us that you'll be tagging each change you want to publish. The package name is not needed as it's just the repo name.

channel.json
```diff
@@ -30,6 +30,7 @@
 		"https://raw.githubusercontent.com/damccull/sublimetext-SolarizedToggle/master/packages.json",
 		"https://raw.githubusercontent.com/danielmagnussons/orgmode/master/packages.json",
 		"https://raw.githubusercontent.com/danielobrien/sublime-DLX-syntax/master/packages.json",
+		"https://raw.githubusercontent.com/dtoader/SelectionsToFiles/master/packages.json",
```
braver:
```suggestion
```

As already mentioned, this is not needed.

braver:
It looks like Tito never published it as a package but just dropped the code on the forum. Most of his packages are retracted and repos are gone by. I can't find anything related in the channel. So it's fine to publish this way as far as I can tell. For completeness sake you could set [SelectionsToFiles](https://packagecontrol.io/packages/SelectionsToFiles) in `previous_names` though, that would be nice.

However, some changes are needed on the repository:
- packages.json has to be removed, your "metadata" is here in this repo and everything else will be handled by the crawler
- I'm a little bit concerned about a potentially decade-plus old script from Tito aimed at ST2 and 3 (4 has been out for ages)... I doubt it's still necessary given the latest state of the API. You're just inserting text into a new view, so perhaps use a TextCommand and simply [insert](https://www.sublimetext.com/docs/api_reference.html#sublime.View.insert).

kaste:
I'm fooled also because Github's search isn't working: https://github.com/search?q=repo%3Awbond%2Fpackage_control_channel+SelectionsToFiles&type=commits  <-- doesn't show when the package got removed.   But git does locally

```
  [a]ll: true  -S'SelectionsToFiles'

...                               
| ● | | | | | | | | | | | | | | | 6940c5e5 Remove some packages                     ​ Mar 17 2016, Tito Bouzout
.. ● 5abb8504 add vraySublime                                                       ​ Jul 1 2015, unknown
| ● | | | | | | | | | | | | | | | | | | 00fd8a55 Upgraded to schema_version 3.0.0.. ​ Dec 10 2014, wbond
.. ● 44996366 SelectionsToFiles Attempt 1 :P                                        ​ May 9 2014, Tito
```
and then 6940c5e5 points to https://github.com/wbond/package_control_channel/pull/5335

Do not set previous_names as you re-use the same name!  

The code is out-dated and wouldn't pass for a *new* package.  Can you actually  -- in the sense of: are you capable of doing it -- just make it a TextCommand and drop the Edit abstraction?

dtoader:
I'm used to working off of JIRA issue numbers, creating an issue and having that flow into the bitbucket space.  Any code change commit log message is prefixed with an issue string.  Any way to do this in github?
I've made the requested changes and committed (with a nice descriptive log message) to the main branch of
https://github.com/dtoader/SelectionsToFiles
I've also made changes to the AddSelectionsToFiles branch of my package_control_channel fork
https://github.com/dtoader/package_control_channel/tree/AddSelectionsToFiles
Do I need to attempt to merge 
https://github.com/dtoader/package_control_channel/tree/AddSelectionsToFiles
back into the main repo?
I'm assuming after making the requested changes, this package will get approved, merged and I'll be able to use the ST Package Control instead of local files.

repository/s.json
```diff
@@ -955,11 +955,11 @@
 			]
 		},
 		{
-			"name": "SelectionsToFiles",
 			"details": "https://github.com/dtoader/SelectionsToFiles",
 			"releases": [
 				{
 					"sublime_text": "*",
+					"tags": true,
 					"details": "https://github.com/dtoader/SelectionsToFiles/tree/master"
```
braver:
```suggestion
```

dtoader: Removed keyval from releases object.

braver:
> Any code change commit log message is prefixed with an issue string. Any way to do this in github?

You typically put the issue number in the commit message, e.g. `#1234 fix problem`. 

> Do I need to attempt to merge

No, any changes there appear in this PR, and by completing the PR and merging it here you'll be able to install the package via Package Control.

braver:
Don't forget to tag your latest version. Since you're bringing an existing package back, and put in the effort to rewrite it already, I'm inclined to go ahead and ship it. But the latest tag still points at an old version.

PR #9233, opened 2025-10-25
added EKL syntax

kaessajidi:

[2]: https://semver.org

My package is a specific 3DEXPERIENCE language syntax

### DIFF
--- a/channel.json
+++ b/channel.json
@@ -115,7 +115,6 @@
 		"https://raw.githubusercontent.com/x3ns/x3-alpha-theme/master/packages.json",
 		"https://raw.githubusercontent.com/xgenvn/InputHelper/master/packages.json",
 		"https://raw.githubusercontent.com/zfkun/sublime-kissy-snippets/master/packages.json",
-		"https://raw.githubusercontent.com/kaessajidi/ekl-sublim-text/refs/heads/main/package.json",
 		"https://tinkerwell.app/plugins/sublime/packages.json",
 		"https://tobygiacometti.com/sublime/packages.json"
 	]


channel.json
```diff
@@ -115,6 +115,7 @@
 		"https://raw.githubusercontent.com/x3ns/x3-alpha-theme/master/packages.json",
 		"https://raw.githubusercontent.com/xgenvn/InputHelper/master/packages.json",
 		"https://raw.githubusercontent.com/zfkun/sublime-kissy-snippets/master/packages.json",
+		"https://raw.githubusercontent.com/kaessajidi/ekl-sublim-text/refs/heads/main/package.json",
```
braver: Please remove the package.json from your repository, and then this line. It's not necessary, your entry is fully described in the e.json file.

As for that file, make sure you add your entry in the correct alphabetical order.

kaessajidi: fixed version is 3.0.0.

repository/e.json
```diff
@@ -1,6 +1,18 @@
 {
 	"schema_version": "3.0.0",
 	"packages": [
+		{
+			"details": "https://github.com/kaessajidi/ekl-sublim-text",
+			"name": "Dassault Systemes EKL Syntax",
```
braver, line 6:
```suggestion
			"name": "Dassault Systemes EKL",
```

Language support packages don't need to have "syntax" or something like it in the name, since typically it does more than just the syntax: in your case it also provides completions (which you might want to specify in the labels). 

If you want to call it Dassault Systemes EKL and not just EKL, it should go into the d.json file instead, an in alphabetical order there.

braver: > If you want to call it Dassault Systemes EKL and not just EKL, it should go into the d.json file instead, an in alphabetical order there.

You missed this point I'm afraid.

braver: 👋🏻 your PR is not mergeable in its current state.

PR #9235, opened 2025-10-27
Add Breadcrumb Panel (ST4+, GitHub tags)

sarev:

My package is similar to the Breadcrumbs package except it is a cleanroom implementation that outputs to a (semi-persistent) panel, similar to how the linters often work, meaning you can always have a view of your breadcrumb context for the active view while you're editing. I've taken steps to ensure it isn't too much of a performance drain and been very generous with the comments to hopefully help other plugin developers to understand how it works.

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -1456,6 +1456,17 @@
 			]
 		},
 		{
+			"name": "Breadcrumb Panel",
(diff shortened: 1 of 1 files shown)


braver:
Please have a look at the checks, your package is out of alphabetical order and the automated test on the package itself also fails.

PR #9242, opened 2025-11-12
Add Nilesoft-Shell-Script Package

abdalmoniem:
Change-Id: I9640d3267b070f2622f91b3e64837e3e6f804977

[2]: https://semver.org

My package is A Sublime Text syntax definition for Nilesoft Shell Script files.

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -758,7 +758,6 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"platforms": "*",
 					"tags": true
 				}
 			]


repository/n.json
```diff
@@ -751,6 +751,18 @@
 				}
 			]
 		},
+		{
+			"name": "Nilesoft Shell Script",
+			"details": "https://github.com/abdalmoniem/Nilesoft-Shell-Script",
+			"labels": ["syntax highlighting", "nilesoft shell", "nss"],
```
braver:
```suggestion
			"labels": ["language syntax", "nilesoft shell", "nss"],
```

"language syntax" is the canonical label for a package like this.

repository/n.json
```diff
@@ -751,6 +751,18 @@
...
+			"releases": [
+				{
+					"sublime_text": "*",
+					"platforms": "*",
```
braver:
```suggestion
```

"platforms" can be omitted if you support all.

braver:
Your package looks good, but we can't test it because the 1.0.0 tag points to the version that just has the LICENSE file in it. Tag the latest version of your repo so we can test and then ship it!

abdalmoniem:
> Your package looks good, but we can't test it because the 1.0.0 tag points to the version that just has the LICENSE file in it. Tag the latest version of your repo so we can test and then ship it!

done ☺️

PR #9245, opened 2025-11-16
add Agentic

alecGraves:
- [x] I have tagged a release with a [semver][2] version number. (1.0.0)
    - This package uses keybindings that only trigger on files created by the package for a minimal chat interface.
    - Scope of keybindings are limited using a `"context"` key that looks for file settings set by the plugin.

[2]: https://semver.org

My package is Agentic, a minimal OpenAI API `v1/chat/completions` interface for launching multiple custom LLM 'actions' on a file or code snippet in a project and streaming results.

My package is similar to others such as [OpenAI completion](https://packagecontrol.io/packages/OpenAI%20completion) however it should still be added because it focuses on achieving a minimal interface for rapidly launching repetitive user-defined custom agentic actions like 'simplify this code' or 'comment this code'. My package further focuses on streaming from multiple LLMs concurrently to enable multi-LLM workflows, and it supports configuration of pools of models for specific actions through package settings.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -482,7 +482,7 @@
 			]
 		},
 		{
-			"author": "Alec Graves",
+			"name": "Agentic",
 			"details": "https://github.com/alecGraves/Agentic",
 			"labels": ["ai", "llm", "chat"],
 			"releases": [


alecGraves:
I also tested the plugin with sublime 3211 and it runs fine. I think Agentic should mostly work for any >= 3000. I am only using python 3.3 features and Python standard library, with basic sublime features for file manipulation.

repository/a.json
```diff
@@ -481,6 +481,18 @@
 				}
 			]
 		},
+		{
+			"name": "Agentic",
+			"author": "Alec Graves",
+			"details": "https://github.com/alecGraves/Agentic",
```
braver:
```suggestion
			"author": "Alec Graves",
			"details": "https://github.com/alecGraves/Agentic",
```

Since the name matches the repo name, it's optional.

braver:
Note by the way about the version of python: https://www.sublimetext.com/blog/articles/sublime-text-4200. You'll really want to be targeting something newer than 3.3 by now. See for more info also here: https://www.sublimetext.com/docs/api_environments.html#python-version

Finally, make sure you remember to tag your latest version. Package Control only deploys tags to users.

alecGraves:
Oh, thanks for the info! I have been running with 3.8 for a couple days and everything works still. I pushed a new tag with the updated `.python-version` file for 3.8.

PR #9246, opened 2025-11-16
use the_crawl-based package reviewer

braver:


kaste:
It becomes easier to decipher when you just have one job(?) per action(?) / file.  

I may just merge into the forked main which is easier to use and to see and remember.

braver:
📦

PR #9250, opened 2025-11-23
Add autoTOC entry with details and releases

mynl:

[2]: https://semver.org

My package is autoTOC. It creates a new tab with a table of contents for the doc. Works for 
py, md, qmd, tex, css, and yaml files. 

My package is similar to an old ST3 toc (that I used all the time), but that never 
got updated for ST4.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -2626,6 +2626,16 @@
 			]
 		},
 		{
+		    "name": "autoTOC",
+		    "details": "https://github.com/mynl/autoTOC",
+		    "releases": [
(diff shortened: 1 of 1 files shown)


repository/a.json
```diff
@@ -2625,6 +2625,16 @@
 				}
 			]
 		},
+		{
+		    "name": "autoTOC",
+		    "details": "https://github.com/mynl/autoTOC",
```
braver, line 2630: This URL returns a 404?

PR #9253, opened 2025-12-03
Update facelessuser packages for Python 3.13 support

facelessuser:
Additionally, keep ColorHelper on older tag for ST versions without Python 3.8

[2]: https://semver.org

My package is ...

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2891,8 +2891,16 @@
 			"details": "https://github.com/facelessuser/ColorHelper",
 			"releases": [
 				{
-					"sublime_text": ">=3170",
+					"sublime_text": "3170 - 3999",
+					"tags": "st3a-"
+				},
+				{
+					"sublime_text": "4000 - 4200",
 					"tags": "st3-"
+				},
+				{
+					"sublime_text": ">=4201",
+					"tags": "st4-"
 				}
 			]
 		},
--- a/repository/a.json
+++ b/repository/a.json
@@ -1620,8 +1620,12 @@
 			"details": "https://github.com/facelessuser/ApplySyntax",
 			"releases": [
 				{
-					"sublime_text": ">=3000",
+					"sublime_text": "3000 - 4200",
 					"tags": "st3-"
+				},
+				{
+					"sublime_text": ">=4201",
+					"tags": "st4-"
 				}
 			]
 		},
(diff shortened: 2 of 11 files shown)


repository/c.json
```diff
@@ -2891,8 +2891,16 @@
 			"details": "https://github.com/facelessuser/ColorHelper",
 			"releases": [
 				{
-					"sublime_text": ">=3170",
+					"sublime_text": "3170 - 3999",
+					"tags": "st3a-"
+				},
+				{
+					"sublime_text": "4000 - 4200",
```
deathaxe, line 2898: Just a side note:

This is an example of when and why using ST build numbers as tag prefixes would be more straight forward. A ST4-only release tagged `st3-` is somewhat confusing.

Using `st3170-`, `st4000-` and `st4201-` or only `3170-`... would be unambiguious and future-proof tagging style.

facelessuser: For me, `st4-` represents when I consciously dropped explicit Sublime Text 3 support. The ColorHelper situation is very specific.

I updated ColorHelper to support Python 3.8 (so I didn't have to keep backporting the coloraide dependency to Py 3.3), but I had had a lapse in judgment and had not realized that absolutely none of the ST 3xxx builds supported 3.8. I've already made multiple releases on the `st3` tag and didn't want to roll the changes back.  The choice was to either split the ST3 or the ST4 releases. I chose to go with the mental model that st4 is the new stuff and st3 is the old stuff. I wanted to keep it consistent across all my packages.

The plan is to only release on the `st4` tag moving forward, except in rare cases where I may need a bugfix on `st3` before Python 3.13 support is in a stable release.

Most people don't know or care what tags are used, only that they want a working version on Sublime Text. `st4-` is easy for me to remember for the "new stuff" across all my packages. I imagine it is likely that I will switch to `py314-`, `py315-` prefixes in the future when a new Python version is available and I need to drop support for the old (unless a major Sublime architectural change forces me to rethink that).

Anyway, I'm more than happy to leave things as they are unless this is not a suggestion and is instead blocking the merge. If this is blocking, I'll just retag the older releases (before my `st4-` tags) and lock them both to a specific release. My intent is to focus feature development on the `st4-` tags moving forward, except for bugfixes on `st3-` until a stable branch is available.

deathaxe: Finally, tag prefixes are package author responsibility. Don't expect this comment being a blocker.

facelessuser: Thanks for the clarification!

PR #9254, opened 2025-12-06
Drop dead packages

deathaxe:
This commit drops packages, which fail with 404.

see: https://github.com/packagecontrol/thecrawl/releases/tag/crawler-status

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -1769,8 +1769,9 @@
 			]
 		},
 		{
-			"name": "DrupalContribSearch",
-			"details": "https://github.com/seregatte/DrupalContribSearch",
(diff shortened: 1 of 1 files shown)


repository/d.json
```diff
@@ -1768,17 +1768,6 @@
 				}
 			]
 		},
-		{
-			"name": "Drupal Snippets",
-			"details": "https://github.com/juhasz/drupal_sublime-snippets",
```
braver, line 1773: This one doesn't 404: https://packages.sublimetext.io/packages/Drupal%20Snippets

deathaxe: Replaced with failing DrupalContribSearch.

PR #9257, opened 2025-12-09
Add Umbral syntax package

hersac:
- [x] I have read the docs.
- [x] Any commands are available via the command palette. (Not applicable: syntax-only package)
- [x] Preferences (if any) open properly in split view.
- [x] This syntax package doesn't add a color scheme.
- [x] I use .gitattributes to exclude non-package files.

My package is **the official syntax highlighting package for the Umbral programming language**.

Umbral is a general-purpose programming language created by the author of this package, with its own syntax, modules, examples, and tooling.  
This package provides:

- Syntax highlighting (`tmLanguage`)
- File extension support for `.um`
- Example programs demonstrating the language
- A clean, minimal structure following Package Control guidelines

There are no other packages in Package Control that support the Umbral language.

### DIFF
--- a/repository/u.json
+++ b/repository/u.json
@@ -145,7 +145,7 @@
 		},
 		{
 			"name": "Umbral",
-			"details": "https://github.com/heridev/umbral-sublime",
+			"details": "https://github.com/hersac/umbral-sublime",
 			"releases": [
 				{
 					"sublime_text": ">=4000",


repository/u.json
```diff
@@ -143,6 +143,16 @@
 				}
 			]
 		},
+		{
+			"name": "Umbral",
+			"details": "https://github.com/heridev/umbral-sublime",
```
braver: You mean https://github.com/hersac/umbral-sublime?

```suggestion
			"details": "https://github.com/hersac/umbral-sublime",
```

braver: Hm, that doesn't actually have a syntax (just the json), and it mentions a 1.1.5 version that doesn't exist (there are not tags on that repo) 🤔

hersac: The repository link is adjusted and the Tag is generated with the corresponding version.

braver: The repo still doesn't have a syntax though. You'll need to convert the json to a tmLanguage, or even better: a sublime-syntax.

deathaxe: Yeah, VS Code syntaxes or syntax templates, don't work in ST.

And a proper syntax package shouldn't ship sublime-settings file to assign file extensions, as those should be defined in the syntax file itself.

hersac: The previous VS Code configuration file was removed and the correct file was added. I tested it to verify it worked in Sublime Text, and the release file was also generated again.
image

deathaxe: Thanks.

1. Still, Umbral.sublime-settings is not required, as file extension is already assigned via https://github.com/hersac/umbral-sublime/blob/83647435757218c7542ee67a9e78ce3f3a52ab57/syntaxes/Umbral.sublime-syntax#L4-L5.

   Especially, https://github.com/hersac/umbral-sublime/blob/83647435757218c7542ee67a9e78ce3f3a52ab57/Umbral.sublime-settings#L2 may have other unwanted side effects, when specified by packages. Such settings should be dedicated to user settings.

2. FWIW, With a single syntax file being the primary value of a package, it feels rather over the top to burry it in a sub directory.

3. `examples` folder should probably excluded from sublime-package files via .gitattributes directives.

4. Any intents to add preferences to support `toggle_comment` command? Maybe also indentation rules?

hersac: The requested adjustment has been made by removing unnecessary files and simplifying the project architecture.

image

braver:
Looking good. Please tag your latest version, 1.1.5 still points to an older version.

PR #9259, opened 2025-12-09
Add BuildSock

iccir:

[2]: https://semver.org

My package is a "build system in reverse" - it creates a local socket and allows Sublime to receive status/issues from a continuously-running build server. Unlike LSP or build systems, Sublime is not the initiator of the build - it's just another listener.

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -1697,16 +1697,16 @@
 				}
 			]
 		},
-        {
-            "name": "BuildSock",
-            "details": "https://github.com/iccir/BuildSock",
-            "releases": [
-                {
-                    "sublime_text": ">=4200",
-                    "tags": true
-                }
-            ]
-        },
+		{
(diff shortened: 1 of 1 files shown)


iccir:
Thanks for the feedback! I'll investigate the customary menu and command palette entries later today.

My implementation of settings is due to type safety - I wanted to fall back to the default value if the wrong type is used (similar to how Sublime handles its internal settings) but also wanted to avoid duplicating the defaults in both a `.sublime-settings` file and in Python. Let me see if I can come up with a better solution for that. (It would be so much easier if `sublime.Settings.get()` allowed you to enforce a type).

braver:
Nice one. Please tag your latest version (0.1.0 points to an earlier version), and we can ship it!

iccir:
Done! Thanks again for all your guidance!

PR #9260, opened 2025-12-13
Migrate packages from external wbond repository.

deathaxe:
This PR... moves packages registered at [SublimeText/wbond-packages](https://github.com/SublimeText/wbond-packages) into this repository, to reduce amount of externally maintained repositories.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -548,6 +548,19 @@
 				}
 			]
 		},
+		{	
+			"name": "AiComplete",
+			"description": "Auto code completion using Gemini API",
+			"author": "Chinmay Agarwal",
+			"details": "https://gitlab.com/chinag32/AiComplete",
+			"labels": ["completions", "auto-complete","gemini","ai"],
+			"releases": [
+				{
+					"sublime_text": ">=4143",
+					"tags": true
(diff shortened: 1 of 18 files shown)


deathaxe:
## Package Pontrol

Errors about using `sublime.error_message()` may be legit, despite having been there since day 1.

Complains about existing `package-metadata.json` are wrong as it is included by intend. It is required to treate PC as "managed" after it was installed by ST's installer (or just copying Package Control.sublime-package) into Installed Packages folder. There's no other mechanism, which would otherwise create that file for PC.

Warnings about platform key are nonsense. Yes it uses platform specific code, but it doesn't automatically mean to run only on a subset of them. oscrypto may fail on arm64 platforms however, but that's optional dependency, used on python 3.3 (ST3) only.

# Tortoise

Fails most likely due to being just a python 2.6 (ST2) plugin. So it is of no interest.

braver:
The package tests are useful, but not bulletproof. Most cases it checks for are hardly ever touched. And since the packages were already in the registry as is, I'm not going to lose sleep over it. Thanks for the explanation though, all seems to make sense.

PR #9261, opened 2025-12-14
Four point oh

deathaxe:
This PR prepares package_control_channel for Package Control 4.0 by...

1. upgrading scheme version of channel and repositories to 4.0.0
2. opt-in compatible libraries to python 3.8
3. drop all ST2-only repositories/packages/libraries

This PR is frequently synced with this repo's master branch and https://github.com/packagecontrol/channel.

It contains all current ST3+ packages and libraries.

As many as possible packages have been migrated from external repositories to this one, to reduce attack surface of uncontrolled 3rd-party registries.

> **Warning**
>
> This PR must not be merged before packagecontrol.io has been upgraded
> see: 
> - https://github.com/wbond/packagecontrol.io/pull/155
> - https://github.com/wbond/packagecontrol.io/pull/156
> - https://github.com/wbond/packagecontrol.io/pull/157


deathaxe:
A relevant merge commit is 9e808d553bf51d291ea778524dc72a79cf35abc5. It merges all/many commits with package migrations into four-point-oh.

braver:
These are not overly popular packages and it wouldn't be the first time packages have been in the registry for years that are simply impossible to install or use on current versions of Sublime Text and PC. In a way I'm not even so worried about breaking them: if someone complains there is an opportunity to get it fixed.

If I'm going to help in reviewing this I'm going to have to find a way to break this down... GitHub's PR UI won't work on a changeset this large. 

I guess the first worry has to be to get the schema tests to pass?

kaste:
Usually, such large changesets require that we see and review the scripts that generated them.  One after the other, e.g. "drop all packages that have a version <3000".  If we then agree that this is useful step, @braver -- the people with commit rights -- runs the script locally and merges. 

Esp. dropping is problematic as we actually want to keep track of the used names.  (Which we already have to do in the workspace.json.)  Instead marking them tombstoned/removed: datetime is in the long run preferrable.  There is always a risk to loose the workspace.json at some point, relatively unlikely that we use the registry history (this repo) as a whole.

deathaxe:
The deletions itself are not the problem, but the conflicts they cause with master. I therefore don't intend to put any effort in separating them.

FWIW, this PR should be merged not before old packagecontrl.io site is dropped. Hence there's probably no reason to hurry.

kaste:
How and where should we backup the old tombstoned names?

repository/e.json
```diff
@@ -591,6 +569,20 @@
...
+		{
+			"name": "Element Finder",
+			"details": "https://github.com/thinkv/sublime-elfinder",
+			"description": "Search HTML files using CSS selectors",
+			"author": ["Keegan Street", "thinkV"],
+			"labels": ["file navigation", "search"],
+			"releases": [
+				{
+					"base": "https://github.com/thinkv/sublime-elfinder",
+					"sublime_text": ">=3000",
```
braver:
```suggestion
					"sublime_text": "*",
```

repository/e.json
```diff
@@ -1202,17 +1186,20 @@
...
+			"name": "Erlyman",
+			"details": "https://github.com/Mendor/sublime-erlyman",
+			"labels": ["documentation"],
 			"releases": [
 				{
-					"sublime_text": "<3000",
-					"branch": "master"
+					"sublime_text": "*",
+					"platforms": ["osx", "linux"],
+					"tags": true
```
braver, line 1082: There are no semver-valid [tags](https://github.com/archydragon/sublime-erlyman/tags) on this repo though. Not sure that matters.

deathaxe: They are all valid PEP440 versions, which PC4 fully supports.

PC3 also supported these style of versions, regardless SemVer having been adverticed.

repository/0-9.json
```diff
@@ -29,28 +28,16 @@
...
 			"releases": [
 				{
-					"sublime_text": "*",
 					"tags": true
 				}
 			]
 		},
-		{
-			"name": "2pdf",
-			"details": "https://github.com/fraoustin/Sublime2pdf",
```
braver, line 39: This one is archived but not deleted. So that's still valid as a package?

FichteFoll:
FYI, I am unsubscribing from this PR. If any new developments happen, feel free to mention me again.

michaelblyons:
🔜

deathaxe:
To verify this PR producing same set of packages as current master branch:

1. clone https://github.com/packagecontrol/thecrawl
2. checkout https://github.com/packagecontrol/thecrawl/tree/feat/compare-registries
3. run

   ```bash
   # create registry from sublimehq/package_control_channel/master
   uv run -m scripts.generate_registry --output master/registry.json
   # create registry from deathaxe/package_control_channel/four-point-oh
   uv run -m scripts.generate_registry --output four-point-oh/registry.json --channel https://raw.githubusercontent.com/deathaxe/package_control_channel/refs/heads/four-point-oh/channel.json
   # compare
   uv run -m scripts.compare_registries --registry1=master/registry.json --registry2=four-point-oh/registry.json
   ```

It compares all relevant packages, which are installable with Package Control 4. All packages not compatible with ST3143+ are ignored.

PR #9263, opened 2025-12-15
Add Text Decode + JSON Format

callzhang:
  - [x] I have tagged a release with a semver version number (v1.0.0).
  - [ ] I use .gitattributes to exclude files from the package: images, test files, sublime-project/workspace. (Not
  needed for current minimal payload; can add if requested.)

  My package is Sublime Text commands to auto-decode messy text (URL/HTML/unicode/hex/base64/gzip) and fix/pretty-
  print JSON, with full-buffer fallback when nothing is selected.

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -990,6 +990,17 @@
 			]
 		},
 		{
+			"name": "Text Decode + JSON Format",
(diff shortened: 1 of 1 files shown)


braver:
👋🏻

PR #9264, opened 2025-12-17
Add CiscoSyntax BadWifi package (Alt to existing) 

mburknoe:
- [YES ] I'm the package's author and/or maintainer.
- [YES ] I have have read [the docs][1].
- [ YES, though it's the first] I have tagged a release with a [semver][2] version number.
- [ YES] My package repo has a description and a README describing what it's for and how to use it.
- [ YES] My package doesn't add context menu entries. *
- [ YES] My package doesn't add key bindings. **
- [N/A] Any commands are available via the command palette.
- [N/A] Preferences and keybindings (if any) are listed in the menu and the command palette, and open in split view.
- [ TRUE ] If my package is a syntax it doesn't also add a color scheme. ***
- [ DONE IN PATCH 1.0.1, I didn't know this was a thing! thx.] I use [.gitattributes][3] to exclude files from the package: images, test files, sublime-project/workspace.

My package is an alternative to the existing "Cisco Syntax Highlighter". The existing one has not been updated in 7 years, and really doesn't do anything so I have never used it as a CCIE Cisco Engineer and over my career been frustrated by the lack of Cisco Syntax in Sublime

Mine will be active and I'll address recomendations from users.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1205,6 +1205,16 @@
 			]
 		},
 		{
+			"name": "CiscoSyntax BadWifi",
+			"details": "https://github.com/badwifi/CiscoSyntax-Badwifi-Sublime-syntax",
+			"releases": [
+				{
+				"sublime_text": ">=3000",
+				"tags": true
+				}
+			]
+		},
+		{
 			"name": "CiteBibtex",
 			"details": "https://github.com/sjpfenninger/citebibtex",
 			"releases": [


repository/c.json
```diff
@@ -1204,6 +1204,16 @@
 				}
 			]
 		},
+		{
+			"name": "CiscoSyntax BadWifi",
+			"details": "https://github.com/badwifi/CiscoSyntax-Badwifi-Sublime-syntax",
```
braver, line 1209: Your repo does not seem to exist at that URL (looks like this might be the one: https://github.com/mburknoe/CiscoSyntax-Badwifi-Sublime-syntax).

mburknoe: Yes, you have the right one, Not sure how I made that mistake. I appreciate you fixing it. 

Did you think that the proposed name is appropriate or should I name it something else?

braver: I think we can actually just replace the existing package and use its name.

mburknoe: I like that alot, would you let me to try the pull request again, or do you have it covered?

If we replace the existing one using the same name, what does that do to all of the users that already have the existing one installed?

braver: I'll make a PR and let you double check, that'll be quicker.

Your package will have a higher version than the existing one, so all users will be upgraded to your new version. 

Ideally, you have a [message](https://docs.sublimetext.io/guide/package-control/utilities.html#messages) explaining what happened and linking back to the original package (so that users who want that can still find it and manually install).

mburknoe: Very good, I have added the message, message.json and changed the package name to "Cisco". 

Is there anything else you need from me? Thank you, this is a great upgrade!

PR #9265, opened 2025-12-19
Add WorkspaceMirroring package

nickname13:

[2]: https://semver.org

My package helps automatically save files to a remote host.
The idea is to keep the code locally so you can use fast file search and other editor features, while syncing changes to the development machine where the code is compiled.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -2101,11 +2101,12 @@
 			]
 		},
 		{
-			"details": "https://github.com/hydralien/SimpleSync",
+			"name": "SimpleSync",
+			"details": "https://github.com/nickname13/SimpleSync",
 			"releases": [
 				{
 					"sublime_text": "*",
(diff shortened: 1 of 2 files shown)


repository/s.json
```diff
@@ -2101,11 +2101,12 @@
 			]
 		},
 		{
-			"details": "https://github.com/hydralien/SimpleSync",
+			"name": "SimpleSync",
+			"details": "https://github.com/nickname13/SimpleSync",
```
braver, line 2105: Adding the name is technically not necessary since it's the name of the GitHub repo, but I don't mind having it there.

nickname13:
I changed a bit the "popup" configuration as an upload_error_notificationthat you can choose between a popup,dialog or nothing at all. 

Addressed all the other comments.

Thanks, again, a lot for your time

Issue #9270, opened 2025-12-22
Add Android Syntax Package


ismell:
I want to add the following new plugin: https://github.com/google/sublime-text-android-syntax

PR #9271, opened 2025-12-22
Add Android Syntax Package (#9270)

ismell:
This CL adds the Android Syntax package.

- [N/A] Any commands are available via the command palette.
- [N/A] Preferences and keybindings (if any) are listed in the menu and the command palette, and open in split view.

[2]: https://semver.org

My package is [Android Syntax](https://github.com/google/sublime-text-android-syntax)

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -1013,6 +1013,17 @@
 			]
 		},
 		{
+			"name": "Android Syntax",
+			"details": "https://github.com/google/sublime-text-android-syntax",
+			"labels": ["android", "language syntax"],
+			"releases": [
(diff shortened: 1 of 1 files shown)


braver:

ismell:
I pushed a new version with the fixes: https://github.com/google/sublime-text-android-syntax/releases/tag/v1.0.1

> However, shouldn't it be "syntaxes" plural?

I guess technically since it does include multiple different file formats. It just doesn't sound as nice though. I would prefer to stick with `Android Syntax` if that's ok with you..

Thanks again

PR #9274, opened 2025-12-24
feat(MarkerStack):  add MarkerStack package

vwheeler63:
## MarkerStack Package:

Efficiently manage interruptions (and sub-tasks) via PUSH/POP of cursor (caret) position in View with single keystroke; marks position in left gutter; saves state between sessions.  This feature enables you to interrupt a task to change something ELSE in the same file, and then return to a prior position and viewport position.  This interruption can be nested to any depth, thus the STACK architecture.

MarkerStack provides a much-loved-and-used function from my prior editor:  Multi-Edit.  No other Sublime Text package provides this service.  

- [n/a] I use [.gitattributes][3] to exclude files from the package: images, test files, sublime-project/workspace.

[2]: https://semver.org

My package is ...

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -744,7 +744,7 @@
 		},
 		{
 			"details": "https://github.com/vwheeler63/MarkerStack",
-			"labels": ["marker stack", "cursor position stack"],
+			"labels": ["bookmark", "navigation", "marker"],
 			"donate": "https://buymeacoffee.com/vwheeler63",
 			"releases": [
 				{


repository/m.json
```diff
@@ -742,6 +742,19 @@
 				}
 			]
 		},
+		{
+			"description": "Push/pop cursor (caret) position in View with single keystroke; saves state between sessions.",
+			"details": "https://github.com/vwheeler63/MarkerStack",
+			"readme": "https://github.com/vwheeler63/MarkerStack/blob/master/README.rst",
```
braver:
```suggestion
			"details": "https://github.com/vwheeler63/MarkerStack",
```

You can put the description on your GitHub repo and we'll pull it from there. That way you can also change it without having to open a PR here.
The readme is in a default location so the crawler will pick that up automatically as well.

vwheeler63: Hi, @braver !  Thank you for your input!  Makes sense.

Exactly where does the description get pulled from?  Is it the repository description itself?  Or one of the contained files?

braver:
> am I understanding correctly that the package description will be fetched from the "About > Description" of the repository?

Correct!

> under View Settings, that the names would all be used in the same dictionary

Package settings are not all automatically accessible via the view settings. Global user preferences (and all their defaults) are, but not all of them. If we look at [project settings](https://www.sublimetext.com/docs/projects.html#settings-key), (some/most) global settings are supported there. If you want to support settings on the view or project level you would need to do a little bit of work. SublimeLinter for instance supports changes to just the linters settings on the project/view level (see [documentation](https://github.com/SublimeLinter/SublimeLinter/blob/master/docs/settings.rst#project-settings)). And there you would indeed have to namespace your settings.

repository/m.json
```diff
@@ -742,6 +742,17 @@
 				}
 			]
 		},
+		{
+			"details": "https://github.com/vwheeler63/MarkerStack",
+			"labels": ["marker stack", "cursor position stack"],
```
braver: These labels are unique to this package, so that won't really help people find it. Some suggestions: bookmark, code navigation. Maybe have a look at all labels currently being used for some ideas: https://packages.sublimetext.io/labels?q=nav

PR #9275, opened 2025-12-27
Add NeoCodeMap

klmp200:

[2]: https://semver.org

My package adds an interactive code map that helps visualise and navigate through symbols on files.

My package is similar to [sublime-codemap](https://github.com/oleg-shilo/sublime-codemap/tree/master) However it should still be added because it uses modern sublime API. It's design is very different and allows for live update while avoiding navigation bugs.

### DIFF
--- a/repository/n.json
+++ b/repository/n.json
@@ -283,6 +283,17 @@
 			]
 		},
 		{
+			"name": "NeoCodeMap",
+			"details": "https://github.com/klmp200/NeoCodeMap",
+			"labels": ["workflow", "code navigation"],
+			"releases": [
+				{
+					"sublime_text": ">=4200",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Neon Intrinsics",
 			"details": "https://github.com/ilya-lavrenov/sublime-neon",
 			"author": "Ilya Lavrenov",


klmp200:
> Cool, so this is like a modern re-interpretation of the [CodeMap package](https://packages.sublimetext.io/packages/CodeMap)?
> 
> I'm going to copy over some comments about loading settings that I posted for a different PR:
> 
> > In general, avoid loading the settings on package load. You want to catch any changes a user makes to the setting, and now you need to add an on change listener, and and unload method... If you just load the settings whenever you need them, that will be instant anyway. There is no performance concern to worry about (in fact your listener is probably detrimental), settings are served from memory, and Sublime already handles changes in the background seamlessly. Just load settings on the fly ([example](https://github.com/braver/ColorConverter/blob/main/convert.py#L185)) and your code can be simplified and made more reliable and performant.
> 
> In short, unless you have very specific needs load settings on the fly as needed.
> 
> Beyond that though, looks really well done 👌🏻

I've checked to apply the suggested changes. This would cause issue with the codemap positioning :/
It's intended for the whole codemap to automatically redraw if you change it's settings, that avoid some stale state and weird behaviour when moving inside the code after changing settings.

For example, if you change the position of the codemap from right to left and don't apply changes immediately on all windows, the moment you move inside one of those windows that uses the codemap, you'll have a really unpleasant full UI refresh that is really annoying.

I already have a settings listener for that reason, I tried to avoid  it but I really don't like how it ends up. I think it's the right way to go, at least for that plugin.

PR #9276, opened 2025-12-29
Move Zxx packages to GitLab

michaelblyons:
The two packages owned by ZxxLang are marked as "archived" in GitHub and have a "moved to GitLab" message in their repo descriptions. They don't appear to have any changes in the GitLab version, but it will drop the MIA tag.


braver:
Oh whatever, let's just go.

PR #9277, opened 2025-12-30
CodeContinue: LLM powered code completion tool

kXborg:

[2]: https://semver.org

My package is codeContinue. 
 - It's build to provide LLM powered code completion.
 - Just start with 1-2 line of comments and continue coding, it suggest next line of code instantly
 - Use any OpenAI compatible endpoint to serve a model and use the same here. I use Qwen2.5-Coder-1.5B-Instruct on a jetson orin nano from vLLM. It works decent.
 - The goal is to have control over the models we use, and keep data local.

Disclaimer: I am aware of potential key binding conflicts, however `Tab` key rarely will be creating issue. Also it is only triggered when suggested code phantom appears while coding. We have tested it rigorously, and we beleive it should not create issues. Having said that, the instructions have been added to change it in the first run -- in case any conflicts exists. I hope the package will not be declined. 

*)   If you do need a context menu, make sure the menu applies to the cursor
     context, and the commands are conditional. Space in this menu is limited!
**)  There aren't enough keys for all packages, so you risk overriding those
     of other packages. You can put commented out suggestions in a keymap file, 
     and/or explain how to create bindings in your README.
***) Syntaxes should work in any color scheme the user chooses.

For bonus points also consider how the review guidelines apply to your package:
https://docs.sublimetext.io/reference/package-control/reviewing.html

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2082,6 +2082,16 @@
 			]
 		},
 		{
+			"name": "Codechef",
+			"details": "https://github.com/prongs/SublimeCodechef",
+			"releases": [
+				{
+					"sublime_text": "<3000",
+					"branch": "master"
+				}
+			]
+		},
+		{
 			"name": "CodeContinue",
 			"details": "https://github.com/kXborg/codeContinue",
 			"labels": ["auto-complete", "ai", "llm", "code completion"],
@@ -2089,16 +2099,6 @@
 				{
 					"sublime_text": ">=4000",
 					"tags": true
-				}
-			]
-		},
-		{
-			"name": "Codechef",
-			"details": "https://github.com/prongs/SublimeCodechef",
-			"releases": [
-				{
-					"sublime_text": "<3000",
-					"branch": "master"
 				}
 			]
 		},


repository/c.json
```diff
@@ -2081,6 +2081,19 @@
 				}
 			]
 		},
+		{
+			"name": "CodeContinue",
+			"details": "https://github.com/kXborg/codeContinue",
+			"author": "Kukil Kashyap Borgohain",
+			"description": "AI Code Completion for Sublime Text",
```
braver:
```suggestion
```

You can leave this out, the GitHub repository description will be taken instead, making it easier for you to make changes.

kXborg: Thanks, updated the config

repository/c.json
```diff
@@ -4,7 +4,10 @@
 		{
 			"name": "C Improved",
 			"details": "https://github.com/abusalimov/SublimeCImproved",
-			"labels": ["language syntax", "c"],
+			"labels": [
```
braver: Looks like you formatted the entire file, please restrict that to just your package.

kXborg:
@braver,
Thanks again. Made the following changes.
- Added Package Settings menu with Settings, Key Bindings, and Configure Wizard entries
- Added 2 new commands: `Preferences: CodeContinue Settings` and `Preferences: CodeContinue Key Bindings`
- Added `"debug": false and "api_key": ""` with documentation comments
- `log()` now only prints when debug setting is true

Please check and let me know.

FichteFoll:
> I'm actually not sure if `sublime.save_settings` writes to the User package automatically?

It does.

braver:
> Please check and let me know.

For keybindings it's usually preferable to open the user's global (ie. not package-specific) keybindings file on the right. You usually want to manage all your bindings in one place, because you've only got one keyboard and typically want to have a clear overview of what you've set to which key combo. [example](https://github.com/sublimelsp/LSP/blob/main/Main.sublime-menu#L301)

Also please revert the formatting of the repository file in this PR.

repository/c.json
kXborg: Yeah, sorry got carried away with Tab completions. Let me fix.

repository/c.json
```diff
@@ -2081,6 +2081,22 @@
...
 		},
+		{
+			"name": "CodeContinue",
+			"details": "https://github.com/kXborg/codeContinue",
+			"labels": [
+				"auto-complete",
+				"ai",
+				"llm",
+				"code completion"
+			],
```
braver:
```suggestion
			"labels": ["auto-complete", "ai", "llm", "code completion"],
```

We tend to labels on one line. File get's long enough as it is :)

kXborg: Sure, updated.

PR #9279, opened 2026-01-03
Add StickyLines

klmp200:

[2]: https://semver.org

My package shows the starting lines of currently visible nested scopes at the top of the editor.
It also provides an on demand popup that does it on demand under the cursor.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -4070,6 +4070,17 @@
 			]
 		},
 		{
+			"name": "StickyLines",
+			"details": "https://github.com/klmp200/StickyLines",
+			"labels": ["workflow", "code navigation"],
+			"releases": [
+				{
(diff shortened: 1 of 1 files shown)


braver:
They're not in the main menu though ([example](https://github.com/braver/ColorConverter/blob/main/Main.sublime-menu#L16)). Many times that's the first place people go to look for these things (e.g. some people are more mouse-oriented).

klmp200:
> They're not in the main menu though ([example](https://github.com/braver/ColorConverter/blob/main/Main.sublime-menu#L16)). Many times that's the first place people go to look for these things (e.g. some people are more mouse-oriented).

Oh ! I see ! You're right, it's missing. I've added it with some other improvements.
Thanks for the heads up.

PR #9280, opened 2026-01-03
Add LiveScript package

deathaxe:
This commit adds a rewritten LiveScript package for Sublime Text 4.

[2]: https://semver.org


deathaxe:
Maybe someone should fix the package review bot to not add false positive warnings. The package is restricted to ST4107. Hence printed warning is invalid.

PR #9281, opened 2026-01-07
ADD json-with-comment package

naget:
My package is json-with-comment

My package is similar to Pretty JSON, However it should still be added because  Pretty JSON cannot format JSON strings that contain comments, json-with-comment solves this problem.

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -1511,6 +1511,18 @@
 					"tags": true
 				}
 			]
+		},
+		{
+			"name": "json-with-comments",
+			"details" : "https://github.com/naget/sublime-json-with-comment",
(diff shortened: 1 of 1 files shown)


deathaxe:
With regards to https://github.com/dzhibas/SublimePrettyJson/issues/187, would it make sense to propose an update to _Pretty JSON_ to enable it handling JSONC? 

Otherwise the package name doesn't really give any hints about its functionality.

braver:
Please let us know when you've come around to addressing the feedback. 

Also, please remove the description from this PR and instead just put it on the GitHub repo. Same end result with the big difference that you can change it at will.

PR #9290, opened 2026-01-10
Restrict TOML

deathaxe:
This commit restricts external TOML package to ST builds before 4198.

Installing it on later builds overrides bundled TOML syntax and causes compatibility issues with syntaxes extending it. At least Python throws errors due to different sublime-syntax versions. Jinja2 and other packages may also be affected.

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -3069,7 +3069,7 @@
 			"labels": ["language syntax", "toml"],
 			"releases": [
 				{
-					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


braver:
Many language packages exist that are alternatives to default languages. Even python and JS have several alternatives. Perhaps a version of this TOML package can be made that doesn't interfere so much, e.g. by a change to its scope name? 

@jasonwilliams perhaps you can help find a solution and avoid having to effectively delist the package?

PR #9292, opened 2026-01-13
Add InlineShell package

randolph555:
- [x] I have read the docs.
- [x] I have tagged a release with a semver version number. (v1.0.2)
- [x] My package doesn't add key bindings. Any commands are available via the command palette. Preferences and keybindings (if any) are listed in the menu and the command palette, and open in split view.

Context menu: Adds one entry "Inline Shell: Run Command" for running selected text as a shell command.

Package name: InlineShell (renamed from sublime-run-command). Entry moved to repository/i.json.

Summary:
- Lightweight inline shell runner (no terminal panel)
- Inline/streaming output
- Pipe support (selection + |command)
- History autocomplete
- Dangerous command confirmation
- macOS/Linux only

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -6151,17 +6151,6 @@
 					"tags": "st3-"
 				}
 			]
-		},
-		{
-		    "name": "sublime-run-command",
-		    "details": "https://github.com/randolph555/sublime-run-command",
-		    "labels": ["shell", "terminal", "command", "execute", "run"],
-		    "releases": [
-		        {
-		            "sublime_text": ">=4000",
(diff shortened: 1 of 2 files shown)


repository/s.json
```diff
@@ -6151,6 +6151,17 @@
 					"tags": "st3-"
 				}
 			]
+		},
+		{
+		    "name": "sublime-run-command",
```
braver: Try to give your package a meaningful name, and one that doesn't start with "Sublime" since all package in PackageControl are for Sublime.

PR #9294, opened 2026-01-18
Added EXIF Syntax Highlighting

alberti42:
- [x] I have read the docs.

My package is **EXIF Syntax**, a Sublime Text syntax definition for highlighting `exiftool` output.

It provides readable highlighting for EXIF metadata key/value output (including grouped output such as `-G`), without adding commands, key bindings, menus, or color schemes.

There are no existing packages in Package Control that provide a dedicated syntax for `exiftool` / EXIF metadata output.

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -1601,6 +1601,17 @@
 			]
 		},
 		{
+			"name": "EXIF Syntax",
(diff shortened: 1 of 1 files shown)


alberti42:
Thanks for the feedback and making it upstream!

PR #9299, opened 2026-01-25
Add Jujutsu package

urschrei:

[2]: https://semver.org

My package is a package for using the [jujutsu VCS](https://www.jj-vcs.dev/latest/) in Sublime Text

### DIFF
--- a/repository/j.json
+++ b/repository/j.json
@@ -1356,6 +1356,16 @@
 			]
 		},
 		{
+			"name": "Jujutsu",
+			"details": "https://github.com/urschrei/jujutsu-sublime-plugin",
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"tags": true
+				}
+			]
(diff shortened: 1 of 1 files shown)


repository/s.json
```diff
@@ -4809,6 +4797,16 @@
 				}
 			]
 		},
+		{
+			"name": "SublimeJJ",
```
braver: We try to avoid using Sublime in package names, since everything is for Sublime in this context anyway. Here, since the thing you're supporting is called Jujutsu, maybe just call the package Jujutsu too?

urschrei: @braver Done: Package renamed in its repo and a new release (v0.4.3) created, and this PR updated.

braver: Cool!

urschrei:
> You should be sorted before Julia.

@braver Done.

PR #9302, opened 2026-02-06
Add BookLeaf package

JoeFerrucci:

[2]: https://semver.org

My package is a scratch notes manager that stores notes as individual files with fuzzy search across filenames and file contents via the Quick Panel.

There are no packages like it in Package Control. The closest is Scratchpad, but it uses a single file with timestamps appended. BookLeaf stores each note as a separate file, allowing for better organization and full-text search.

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -1194,6 +1194,17 @@
 			]
 		},
 		{
+			"name": "BookLeaf",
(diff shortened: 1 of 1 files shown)


braver:
Excellent, thanks!

PR #9303, opened 2026-02-07
Add Buildifier package

mfilippov:

[2]: https://semver.org

My package is support call Buildifier tool when you edit Bazel files.

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -1686,6 +1686,17 @@
 			]
 		},
 		{
+			"name": "Buildifier",
+			"details": "https://github.com/mfilippov/sublime-buildifier",
+			"labels": ["build system", "bazel"],
+			"releases": [
+				{
+					"sublime_text": ">=4200",
+					"tags": true
+				}
+			]
+		},
+		{
 			"details": "https://github.com/misotomlam/BuildParts",
 			"releases": [
 				{


repository/b.json
```diff
@@ -1685,6 +1685,17 @@
 				}
 			]
 		},
+		{
+			"name": "Buildifier",
+			"details": "https://github.com/mfilippov/sublime-buildifier",
+			"releases": [
```
braver, line 1692: You might want to add some labels here, e.g. "[build system](https://packages.sublimetext.io/?q=label%3A%22build+system%22)", "[bazel](https://packages.sublimetext.io/labels?q=bazel)".

braver:
A bazel syntax package already exists, so better to point users to that instead of introducing a potentially conflicting and empty syntax.

repository/b.json
```diff
@@ -1707,6 +1685,18 @@
...
 			]
 		},
+		{
+			"name": "Buildifier",
+			"details": "https://github.com/mfilippov/sublime-buildifier",
+			"labels": ["build system", "bazel"],
+			"releases": [
+				{
+					"sublime_text": ">=4200",
+					"python_versions": ["3.8"],
```
braver: If you want to run on the 3.8 runtime, you do that with a file on your package. PackageControl has no say over that, that's just a delivery mechanism. [Example](https://github.com/braver/ColorConverter/blob/main/.python-version).

braver: Ah, and you already have that. So please remove this line.

mfilippov: Done

PR #9308, opened 2026-02-12
MDEE (Markdown) & MDSVX (Svelte + Markdown) Language Syntax

Autr:

[2]: https://semver.org

My package is similar to _Markdown Extended_ However it should still be added because _Fork of Unmaintained Package & Adds New Things_

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -1217,6 +1217,17 @@
 			]
 		},
 		{
+			"name": "MDEE (Markdown) & MDSVX (Svelte + Markdown)",
+			"details": "https://github.com/autr/MDEE-MDSV",
+			"labels": ["language syntax", "markdown", "svelte", "mdsv", "mdsv", "mdsvx"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "main"
+				}
+			]
+		},
+		{
 			"name": "MDL Language",
 			"details": "https://github.com/ardenpm/sublime-mdl-language",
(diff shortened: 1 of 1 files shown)


repository/m.json
```diff
@@ -1216,6 +1216,17 @@
 				}
 			]
 		},
+		{
+			"name": "MDEE (Markdown) & MDSVX (Svelte + Markdown)",
```
deathaxe, line 1220: The "name" will be the folder name of the package, locally. Hence it should be carefully choosen. Even though most filesystems handle it, `&` and co. should be avoided. It's overall a bit too verbose.

deathaxe (commented):
Further notes:
 
1. The syntaxes shouldn't override default markdown or svelte file extensions. Those should stay reserved for default markdown and Svelte templates, by default.
2. I don't think the way fenced code blocks are implemented, works well with current syntaxes. It is VERY likely those never get popped for various languages.
3. The overall content of MDEE.sublime-syntax looks rather dated using lot's of lookbehinds. Maybe those syntaxes should better extend ST's default Markdown and only extend relevant contexts to add svelte specific template tags. You may want to have a look at Astro, Liquid or Jinja to learn how to do so.

In general it might be better to add a "Markdown (Svelte)" syntax to Svelte package, in case its template tags are to be re-used.

deathaxe:
see: https://github.com/corneliusio/svelte-sublime/pull/41

braver:

PR #9313, opened 2026-02-16
Add Kodi Log Highlighter package

bossanova808:

[2]: https://semver.org

My package is (sort of) similar to [KodiDevKit](https://packages.sublimetext.io/packages/KodiDevKit) ... However it should still be added because that has not been developed for years, is far more complicated, and very buggy.


repository/k.json
```diff
@@ -412,6 +412,19 @@
...
+			"details": "https://github.com/bossanova808/kodi-log-highlighter",
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"tags": true
+				}
+			],
+			"labels": ["language syntax", "logs"],
+			"readme": "https://raw.githubusercontent.com/bossanova808/kodi-log-highlighter/main/README.md",
+			"issues": "https://github.com/bossanova808/kodi-log-highlighter/issues"
```
braver, line 426:
```suggestion
			"name": "Kodi Log Highlighter",
			"details": "https://github.com/bossanova808/kodi-log-highlighter",
			"labels": ["language syntax", "logs"],
			"releases": [
				{
					"sublime_text": ">=4000",
					"tags": true
				}
			]
```

The readme and issues URLs are not needed, they're default GitHub URLs that our crawler will resolve automatically.

repository/k.json
```diff
@@ -412,6 +412,19 @@
 				}
 			]
 		},
+		{
+			"name": "Kodi Log Highlighter",
```
braver, line 416: Would "Kodi Log" also work? That would indicate this is the package to use for those kinds of files. A syntax package is typically not limited to highlighting.

PR #9314, opened 2026-02-19
Add HandeeFramer plugin details to h.json

JohhannasReyn:
- [*] My package doesn't add context menu entries. *

[2]: https://semver.org

My package is HandeeFramer.

It parses a simple “tree” scaffold from a document (or current selection) and generates the corresponding folders/files. It also supports inline path shorthand (e.g. project/src/foo.txt) and can populate generated files using nearby code fences.

The build command is context-aware: if text is selected it builds from the selection; otherwise it builds from the full document. File collisions are handled non-destructively by creating (n) variants rather than overwriting.

All suggested key bindings are commented out by default and can be enabled or customized by the user per the inline comments in the key binding/menu files.

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -189,7 +189,7 @@
 			"releases": [
 				{
 					"sublime_text": ">=3000",
-					"branch": "master"
+					"branch": "main"
 				}
 			]
 		},


repository/h.json
```diff
@@ -181,6 +181,18 @@
...
 		},
+		{
+			"name": "HandeeFramer",
+			"details": "https://github.com/JohhannasReyn/HandeeFramer",
+			"author" : "Johhannas Reyn",
+			"labels": ["productivity", "developer-tools", "scaffolding", "markdown", "codefences", "file-folder-creation"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"branch": "main"
```
braver, line 192: Please refer to the docs: https://docs.sublimetext.io/reference/package-control/repository.html#releases

PR #9322, opened 2026-02-24
Add WooCommerce package

renzojohnson:

My package is a collection of 172 curated WooCommerce code snippets for Sublime Text, covering hooks, CRUD, REST API, HPOS, Block Checkout, Store API, Action Scheduler, shipping, emails, and more.

### DIFF
--- a/repository/w.json
+++ b/repository/w.json
@@ -717,7 +717,7 @@
 		{
 			"name": "WooCommerce",
 			"details": "https://github.com/renzojohnson/woocommerce",
-			"labels": ["snippets", "completions", "woocommerce", "wordpress", "php"],
+			"labels": ["snippets", "woocommerce", "wordpress", "php"],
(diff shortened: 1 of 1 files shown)


repository/w.json
```diff
@@ -714,6 +714,17 @@
 				}
 			]
 		},
+		{
+			"name": "WooCommerce",
+			"details": "https://github.com/renzojohnson/woocommerce",
+			"labels": ["snippets", "completions", "woocommerce", "wordpress", "php"],
```
braver:
```suggestion
			"labels": ["snippets", "woocommerce", "wordpress", "php"],
```

Thanks for handling the previous feedback. Final change to make is that this package does not actually ship "completions" as that feature is understood in Sublime Text. See also https://docs.sublimetext.io/guide/package-control/submitting.html#what-labels-to-use

PR #9323, opened 2026-02-27
Rename Django package

deathaxe:
This commit renames "Django" package to "Django Snippets" and fixes labels to reflect its purpose more accurately. It actually just provides snippets.

The package doesn't include package name related resources such as settings or commands. Hence renaming is transparent for end users.

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -1144,6 +1144,20 @@
 			]
 		},
 		{
+			"name": "Django Snippets",
+			"details": "https://github.com/mattseymour/sublime-django",
+			"labels": ["snippets", "django"],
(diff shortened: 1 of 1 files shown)


FichteFoll:
Order is not correct.
```
+  'Django Snippets',
   'Django Starter',
   'Django Syntax',
   'Django Tags',
   'Django-DocsSearch',
   'Django-Model-Converter',
   'Django-Rest-Snippets',
-  'Django Snippets',
```

FichteFoll:
Also cc @mattseymour, though this renaming is fine in my book and does not necessarily require approval. We could wait a week or so regardless.

PR #9324, opened 2026-03-01
Add HostSettings package

Taffer:
- [x] **N/A** My package doesn't add context menu entries. *
- [x] **N/A** My package doesn't add key bindings. **
- [x] **N/A** Any commands are available via the command palette.
- [x] **N/A** Preferences and keybindings (if any) are listed in the menu and the command palette, and open in split view.
- [x] **N/A** If my package is a syntax it doesn't also add a color scheme. ***

[2]: https://semver.org

My package is for adding specific settings overrides based on the computer it's running on, keyed by the computer's host name. This is useful if you have system-specific settings that can't be controlled by platform (for example, I'm running on multiple Linux platforms, and need to tweak `font_size` different on my laptop than on the system with a 4k monitor).

My package is similar to various other Settings control packages. However it should still be added because none of them allow per-system settings changes, and they generally don't let you change settings without changing the on-disk `.sublime-settings` file(s), which will cause unnecessary churn for people keeping their Sublime setup in version control.

**Note:** I wasn't able to test this change; I couldn't get ChannelRepositoryTools to stop telling me to open the `package_control_channel` folder, even when it was open as the only folder. :shrug: It's valid JSON at least:

```py
>>> with open("repository/h.json") as fp:
...     j = json.load(fp)
...
>>> j
{'schema_version': '3.0.0', 'packages': [{'name': 'Hacker News Reader', 'details': 'https://github.com/Dimillian/Sublime-Hacker-News-Reader',
```

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -858,17 +858,6 @@
 			]
 		},
 		{
-			"name": "HostSettings",
-			"details": "https://github.com/Taffer/HostSettings",
-			"labels": ["utilities"],
-			"releases": [
-				{
-					"sublime_text": ">=4050",
-					"tags": true
-				}
-			]
-		},
-		{
 			"details": "https://github.com/samueljohn/Homebrew-formula-syntax",
 			"labels": ["language syntax"],
 			"releases": [
@@ -985,6 +974,17 @@
 			"releases": [
 				{
 					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
+			"name": "HostSettings",
+			"details": "https://github.com/Taffer/HostSettings",
+			"labels": ["utilities"],
+			"releases": [
+				{
+					"sublime_text": ">=4050",
 					"tags": true
 				}
 			]


repository/h.json
```diff
@@ -857,6 +857,17 @@
 				}
 			]
 		},
+		{
+			"name": "HostSettings",
+			"details": "https://worktree.ca/taffer/HostSettings",
```
braver: We don't support worktree right now, just GitHub, GitLab and Bitbucket. Would you be able to move the project to any of those?

Taffer: That's unfortunate. I'll set up a mirror repo and update the URL here.

Thanks for looking!

repository/h.json
```diff
@@ -857,6 +857,17 @@
 				}
 			]
 		},
+		{
+			"name": "HostSettings",
```
braver: Looks like you need to be sorted between 'HostsEdit' and 'Hot Dog Stand'.

Taffer: :facepalm: Wow, not sure what happened there.

PR #9326, opened 2026-03-05
add ConvCalculate

convictss:

[2]: https://semver.org

My package is come from the needs of my work.

When I want to quickly calculate the addition, subtraction, multiplication and division of high-precision decimals, none of the calculation plugins I found in the sublime text plugin market are based on decimal calculations, so I wrote this widget to quickly support high-precision decimal calculations.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1590,17 +1590,6 @@
 			]
 		},
 		{
-			"name": "Close All without Confirmation",
-			"details": "https://github.com/ddbln/CloseAllNoConfirm",
-			"labels": ["file navigation", "files", "close files", "file close", "commands", "productivity", "utilities", "refactoring"],
-			"releases": [
-				{
-					"sublime_text": ">=3092",
-					"tags": true
-				}
-			]
-		},
-		{
 			"name": "Close Oldest File",
 			"details": "https://github.com/jturcotte/SublimeCloseOldestFile",
 			"releases": [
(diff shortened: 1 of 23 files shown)


repository/c.json
```diff
@@ -3619,6 +3609,20 @@
...
+			"name": "ConvCalculate",
+			"details": "https://github.com/convictss/ConvSublimeCalculate",
+			"description": "A simple calculator with high accuracy.",
+			"author": "Convict Yellow (Kailian.Huang)",
+			"labels": ["calculator"],
+			"releases": [
+				{
+				  "sublime_text": ">=3000",
+				  "tags": true,
+				  "branch": "master"
```
braver:
```suggestion
			"author": "Convict Yellow (Kailian.Huang)",
			"labels": ["calculator"],
			"releases": [
				{
				  "sublime_text": ">=3000",
				  "tags": true
```

It's better if you put the description on your GitHub repo, then you can change it whenever you like. Also the "branch" for release is no longer allowed (and not needed if you have tags).

braver:
Some feedback on the package:

- Please remove [package.json](https://github.com/convictss/ConvSublimeCalculate/blob/master/package.json) from your repo, that information is in your entry here. 
- Expose your keybinding in the package settings menu, but comment it out so it won't conflict with any existing bindings. Also would allow users to pick a binding that makes sense for them, and make it work on their OS.
- Don't print to the console unless in a debug mode. 
- Note that you can (and should) opt in to the 3.8 python runtime ([like so](https://github.com/braver/SideBarTools/blob/master/.python-version)).

repository/c.json
convictss: OK, I have updated. (c.json)

braver:
Ok, let's ship it. 

Have a look at this though, there might be a better way to do the keymap files... I wasn't even aware of that really: https://github.com/kaste/st_package_reviewer/issues/6#issuecomment-4173368169. You can do that in your own time though and just make sure to tag any changes you make to the package so they actually ship to end users.

PR #9333, opened 2026-03-12
Resurrect Modelines package

michaelblyons:
- [x] I ~~am~~ _[represent]_ the package's author and/or maintainer. cc @Frizlab

[2]: https://semver.org

This package is back from the dead! Credit to Frizlab.

I can't merge https://github.com/SublimeText/Modelines/pull/9 to master without this change.[^1] Some of the checkmarks above are dependent on that merge. I'll tag an ST4 release when things are settled.[^2]

[^1]: That's based on the very unlikely assumption that someone is using ST2, has this package installed, and has automatic updates from Package Control. :laughing: But we'll pretend.
[^2]: **Edit:** I've tagged a prerelease for ST4. I'd *like* to save the big 2.0.0 tag for the merge commit or a `messages.json` update, but I can tag it now if it's necessary for the automated tests.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -2090,6 +2090,10 @@
 				{
 					"sublime_text": "<3000",
 					"tags": "st2/"
+				},
+				{
+					"sublime_text": ">=4135",
+					"tags": "st4135/"
 				}
 			]
 		},


michaelblyons:
> Is there anything I can do to move this forward?

My understanding is that package changes with lots of Python functionality (i.e. not a color scheme, syntax, or snippet collection) take reviewer time, which is in short supply.

We could probably get a quicker merge on a PR that limits ST2 to tags rather than `master`. Then I could merge https://github.com/SublimeText/Modelines/pull/9 safely and a subsequent PR for the ST4 tag series would at hopefully :crossed_fingers:  get the automatic reviewer to look at the right tags.

But after Some Incidents :tm: in the past, packages get at least a security check to make sure they don't do anything nefarious. And sometimes additional code review.

braver:
The package reviewer script mostly looks at the tip of the default branch. That's in almost all cases the version we want to review. See also https://github.com/sublimehq/package_control_channel/pull/9330#issuecomment-4055769068. The repo layout is also a bit confusing for contributors if that's not the case... so while I'm inclined to take @michaelblyons 's word for it that the package is good, but I would love to run the bot on it anyway :)

As for my own very quick review, some notes:
- The package is called "Modelines", but refers to itself in menus as "Sublime Modelines", where being inside Sublime the Sublime prefix is extra superfluous :wink: 
- ... although maybe the settings file people might still have is already called "Sublime Modelines.sublime-settings" that can't be helped? Do we expect anyone to actually have used this all these years and are old settings still compatible even?

Frizlab:
> The package reviewer script mostly looks at the tip of the default branch.

Not sure if that helps (probably not), but I have updated the default branch of my fork to be the one I worked on, and that will be merged.

> The package is called "Modelines", but refers to itself in menus as "Sublime Modelines"

The proper naming is indeed Modelines now, because I have added support for common modelines (VIM and Emacs). I can rename everything.
(The naming previously made sense because it was only supporting _Sublime_ modelines.)

> maybe the settings file people might still have is already called "Sublime Modelines.sublime-settings"

I think there were no settings file prior to my rewrite, so that will not be an issue.

PR #9334, opened 2026-03-12
Normalize indentation label

michaelblyons:
Normalize indentation label from "indent" and "text indentation". Also remove "language syntax" from some indentation packages that do not provide a language syntax.

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -548,7 +548,7 @@
 			"name": "Indent Respectful Sort",
 			"author": "Mustafa Nural",
 			"details": "https://github.com/mvnural/sublime-indent-respectful-sort",
-			"labels": ["sort"],
+			"labels": ["sort", "indentation"],
 			"releases": [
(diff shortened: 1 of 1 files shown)


repository/i.json
```diff
@@ -548,7 +548,7 @@
 			"name": "Indent Respectful Sort",
 			"author": "Mustafa Nural",
 			"details": "https://github.com/mvnural/sublime-indent-respectful-sort",
-			"labels": ["sort", "indent"],
+			"labels": ["sort"],
```
braver: This package works with indentation in a way that I think warrants the label. It doesn't "do" indentation, but it's "about" indentation anyway.

michaelblyons: Seems reasonable. I'm adding the "suggestion" diff below so I can patch from my phone.

```suggestion
			"labels": ["sort", "indentation"],
```

PR #9335, opened 2026-03-13
Add Checklist+ syntax package

bluejack:
- [x] I have tagged a release with a [semver][2] version number. {{ v1.1.0 }}
- [x] My package repo has a description and a README describing what it's for and how to use it. See: [Checklist+ Repo](https://github.com/bluejack/ChecklistPlus)
- [x] My package doesn't add key bindings. ** {{ but does add a click binding, scoped to this file type }}
- [ ] (1) Any commands are available via the command palette. {{ there is a command, but it needs the click event, so it doesn't make sense in the command palette: see note 1. }}
- [ ] (2) If my package is a syntax it doesn't also add a color scheme. *** {{ because this makes extensive use of markdown scope, which very few color schemes implement in whole or in part, this *does* come with a color scheme that is scoped to this syntax, and which will not interfere with the user's global style they may use for editing code. See note 2. }}

[2]: https://semver.org

My package is "Checklist+" 

I have not found any package like this in Package Control, although there are plenty of simple TODO list syntaxes. This is much more extensive and interesting: it is intended to make Sublime a fully functional Obsidian-comparable local wiki/knowledgebase and time tracker.

Notes on the exceptions above:

1. This package offers clickable links: local files & web uris. Those click events to require a command, which takes a click event. It would not make sense for those commands to be in the UI.

2. This package includes a color scheme, contrary to your instructions. The reason for this is that I use sublime for coding; and I have not found any coding-centric color scheme that includes the kind of heavily extensive markup color scheme that would work for this app. Accordingly, locking the included color scheme to this file extension allows me to have both my coding color scheme and my Checklist+ color scheme. If this is absolutely banned, let me know and I will remove it, and package them separately with instructions on how to set up color-specific syntaxes ... but that is a worse user experience, honestly.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -805,6 +805,19 @@
 			]
 		},
 		{
+			"name": "Checklist+",
+			"author": "bluejack",
+			"description": "Syntax, color, and extended functionality for checklists, todo-lists, note-taking, time-management, task-logging, and wiki-like behavior. When you like what Obsidian does, but not how it looks, and you'd rather be in Sublime.",
+			"labels": ["syntax highlighting", "note-taking", "todo", "checklist", "productivity", "wiki", "markdown", "time-management", "work log"],
+			"details": "https://github.com/bluejack/ChecklistPlus",
+			"releases": [
+				{
+					"sublime_text": ">=4050",
+					"tags": true
+				}
+			]
+		},
+		{
 			"details": "https://github.com/vaisaghvt/CheckTypos",
 			"releases": [
 				{


repository/c.json
```diff
@@ -804,6 +804,19 @@
 				}
 			]
 		},
+		{
+			"name": "Checklist+",
+			"author": "bluejack",
+			"description": "Syntax, color, and extended functionality for checklists, todo-lists, note-taking, time-management, task-logging, and wiki-like behavior. When you like what Obsidian does, but not how it looks, and you'd rather be in Sublime.",
+			"labels": ["syntax highlighting", "note-taking", "todo", "checklist", "productivity", "wiki", "markdown", "time-management", "work log"],
+			"details": "https://github.com/bluejack/ChecklistPlus",
```
braver, line 812:
```suggestion
			"labels": ["syntax highlighting", "note-taking", "todo", "checklist", "productivity", "wiki", "markdown", "time-management", "work log"],
			"details": "https://github.com/bluejack/ChecklistPlus",
```

The author and description can be taken from the GitHub repo so you don't need them here. Especially the description is easier to manage that way.

Please have a look at [commonly used labels](https://packages.sublimetext.io/labels). Introducing labels that exist only on your package doesn't help users find it and adds noise.

PR #9336, opened 2026-03-16
Add SessionTimeMachine

JCrun:
Add a description:

[2]: https://semver.org

My package is SessionTimeMachine, a Sublime Text 3 plugin that provides local Session snapshots, rollback, history search, and optional Git sync.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -196,6 +196,17 @@
 			]
 		},
 		{
+			"name": "SessionTimeMachine",
+			"description": "Session snapshots, rollback, history search, and optional Git sync.",
+			"details": "https://github.com/JCrun/SessionTimeMachine",
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Sass",
 			"details": "https://github.com/SublimeText/Sass",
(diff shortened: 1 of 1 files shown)


kaste:
Hej.  (Would have been easier if I would speak Chinese.)

I just write down some notes, in a random way.

* the setting git_pull_on_startup is not referenced in code (but sync_pull_on_startup is)
* Would be actually cool/cooler if it would (at atleast a bit) restore the windows if the user had multiple windows open
* All .sublime-keymap files are empty, so better remove them.  Also no need to reference them in the https://github.com/JCrun/SessionTimeMachine/blob/main/Main.sublime-menu#L21 then
* The README should be clearer and maybe multi-lingual.  Instead of just an enumeration of commands you provide, maybe explain first what the plugin does, esp. if I'm reading this correct, it esp. auto-saves *unnamed* buffers; optionally pushing everything to a server (using git).  Maybe explain it from the perspective of a use-case/workflow.
* Likely people want to know what data you save and where; esp. when you also eventually push the data out of the local computer.  (That's a privacy matter.)

braver:
@JCrun please respond to the feedback to continue the review!

braver:
Let us know if you want to continue, we can re-open the PR.

PR #9351, opened 2026-03-27
Debian Syntax: add ST4 release channel with unprefixed tags

barnumbirr:

[2]: https://semver.org

Adds a separate release entry for Sublime Text 4+ using unprefixed tags, while preserving the existing ST3 release channel with `st3-` prefixed tags.

This allows [Debian Syntax](https://github.com/barnumbirr/sublime-debian) v2.0.0+ (which uses ST4-only features like `set:`) to be served to ST4+ users, while ST3 users continue to receive v1.4.2.

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -417,7 +417,7 @@
 				},
 				{
 					"sublime_text": ">=4000",
-					"tags": "st4-"
+					"tags": true
 				}
 			]
 		},


repository/d.json
```diff
@@ -413,7 +413,11 @@
 			"releases": [
 				{
 					"sublime_text": ">=3092",
```
michaelblyons:
```suggestion
					"sublime_text": "3092 - 3999",
```
This is the normal format, but I'm not sure that the version support cut-off is necessarily at 4000. It may be good enough, though. 😉

barnumbirr: There doesn't seem to be any consensus (at least that I could find) in any of the files in this repo 😕

michaelblyons: https://docs.sublimetext.io/reference/package-control/repository.html#sublime-text

barnumbirr:
@braver Anything I can do to get this merged?

rchl:
It's probably more practical to keep st4+ version on the main branch and use unprefixed tags for that. Only fork into prefixed tags when there is a breaking change.
But up to you.

barnumbirr:
@rchl Nice suggestion (and the smart thing to do). Updated, thank you!

barnumbirr:
I've released v2.4.1 to fix the two review warnings:
- removed syntax-specific view settings
- added .gitattributes to exclude non-essential files from the installed package

PR #9354, opened 2026-03-30
feat: add sublime-breadcrumb plugin

huyhoang8398:

[2]: https://semver.org

A clean  breadcrumb that shows both the file path and the current symbol hierarchy at the top of the editor. Powered by the LSP plugin for accurate
image


kaste:
Yeah.

```
    "sublime_text": ">3084"
```

is probably unrealistic as you require LSP.  

We don't have best in class solution for this.  You could name the plugin LSP-addon-... to make that more clear.  But you can also really write the code a bit more pessimistic:

Maybe do a `try import LSP` at the global scope and catch the `ImportError`; and if you catched it let the event handler do nothing.  Or even just print a message to the console   and don't register any of the extension names, no Command no EventListener.

Register an install.txt message that clearly says:  "This is an addon to LSP"/"Attention: this requires LSP <link to LSP> and does this...".

braver:
@huyhoang8398 can you perhaps rebase on master so we can run the latest tests? I think they should pass. 

It's not super clear that this package depends on LSP. I think prefixing it with LSP- in some way would indeed help. Or catch the situation where someone installs it without also having LSP active, and then explain that somehow. 

Package itself looks very interesting at least 👌🏻

rchl:
BTW. There is this rule in the docs:

> Try not to use the word "Sublime" in your package name. Every package available via Package Control is for Sublime Text. Using the word Sublime just adds noise to the list when trying to find packages. You can use the word Sublime in your marketing material (README etc.) though of course.

huyhoang8398:
@braver @rchl @kaste Hey guys, thanks so much! Just got back from holidays so sorry for the slow response — really glad you found the plugin interesting!

A couple of things have changed on my end in the meantime:

**Repo rename:** I've renamed the repo to `LSP-addon-breadcrumb`. The old `LSP-` prefix convention is reserved for actual language server clients, so I went with `LSP-addon-` to make it clear this is a companion/consumer of LSP rather than a server itself.

**On the code side**, I've landed two things based on your feedback:

- `try/except ImportError` at the global scope for the LSP import, if LSP isn't installed the plugin still loads and shows the file path breadcrumb, it just skips the symbol context gracefully.
- Added a `messages.json` + `messages/install.txt` so Package Control shows a note on install pointing users toward LSP for the full experience.

Let me know if you have any other thoughts, happy to review further!

repository/l.json
```diff
@@ -1833,6 +1833,17 @@
 				}
 			]
 		},
+		{
+			"name": "LSP-addon-breadcrumb",
+			"details": "https://github.com/huyhoang8398/LSP-addon-breadcrumb",
+			"labels": ["file", "utilities", "tooltips", "breadcrumb"],
+			"releases": [
+				{
+					"sublime_text": ">3084",
```
kaste, line 1842: Really?

kaste (commented):
The plugin itself LGTM!

repository/l.json
huyhoang8398: `"sublime_text": ">3084"`  for this I dont have a solution for this, maybe just remove this part

kaste: The current LSP is `>=4132`.

huyhoang8398: thanks, i guess i will use this tag then

braver: In a way this package depends more on LSP than on Sublime, so if you've managed to satisfy the LSP dependency then you should be good. There are version ranges for LSP that support older versions of Sublime. You're welcome to try and figure out the support matrix between that and any Sublime API's this package might be using, but that's usually beyond the call of duty for a simple package. 
I'm ok with being optimistic and setting a low requirement and let that shake out in practice. You're welcome to come back and raise the requirement if this becomes a problem.

PR #9355, opened 2026-03-30
Integrate claude code and codex cli as agentic coding tool

synote:

[2]: https://semver.org

My package is agentic coding tool named TermMate, bringing claude code and codex cli directly into ST. 

The entire agent workflow is driven by the Command Palette. The package create a chat view that directly uses native ST Edit view to create a seamless agent chat and command palette experience. It built a clean, generic abstraction layer [genfoundry](https://github.com/flashmodel/termmate/tree/main/genfoundry) which take a conversational, long-running agent cli(codex, claude) process. It may support more agent cli tool in the future.

The plugin add context menu of Chat with TermMate which can send current context(file name, lines) to chat. Much like vscode copilot, the solution is far more easy to use.

There are similar packages [Codex](https://packagecontrol.io/packages/Codex) in Package Control. The main difference is that TermMate provides a native, universal multi-agent Interface. This package uses a standard Sublime Text edit tab (ChatView), allowing you to use all native editor features. It's designed to orchestrate multiple agent CLI.

This package may implement most of the features of vscode Agent HQ in the future.

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -741,6 +741,17 @@
 			]
 		},
 		{
+			"name": "TermMate",
+			"details": "https://github.com/flashmodel/termmate",
+			"labels": ["agent", "chat", "ai", "llm"],
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "termX",
 			"homepage": "https://malinowski.be/termX",
 			"details": "https://github.com/afterdesign/termX",


repository/t.json
```diff
@@ -770,6 +770,17 @@
 				}
 			]
 		},
+		{
+			"name": "TermMate",
+			"details": "https://github.com/flashmodel/termmate",
+			"labels": ["agent", "chat", "codex", "claude", "skills"],
```
braver: Maybe add [ai](https://packages.sublimetext.io/?q=label%3A%22ai%22)? I'm not sure "codex", "claude", "skills" are super meaningful, especially codex and skills are ambiguous (most "codex" packages are something to do with Wordpress).

braver:
I'll probably try to do a more thorough code review later, but here's some details at least:

Small typo in your readme:
> While TermMate automatically detects most **agnet** cli installation paths

For any packages like this we always ask to make it crystal clear from the readme when, how, and what data might leave the user's system to be sent to the llm/agent/service.

You probably want your ChatMD syntax to be hidden, ie. only automatically assigned for your package's features and not selectable by users.

You probably want to have specific keybindings for each OS, as in Win/Linux you'd want to bind CTRL+Enter instead, and on macOS CMND+Enter.

synote:
Now the keybindings include Default.sublime-keymap for Windows/Linux and Default (OSX).sublime-keymap for mac.
New label "ai" "llm" was added to package which can be more clear for user.

braver:
@kaste seems you enjoy reviewing these kinds of packages, wanna have a look? 🙂 Nothing really jumps out at me. 

Just an extremely minor nitpick: you typically don't need a [separator](https://github.com/flashmodel/termmate/blob/main/Main.sublime-menu#L24) if there are just 2 items in a submenu, example:
Scherm­afbeelding 2026-05-09 om 12 43 27

Any interest in supporting other services? Let's say someone wants to introduce support for the tool they like to use, could you give some guidance on how to contribute that?

kaste:
Yeah, sorry.  

This all looks good but the architecture/file layout isn't quite there.  You define multiple python files at the root.  Each python file at the root is observed by ST for changes and is called a plugin.

It is allowed to have multiple plugins in a package, as entry points, but you must not import them from each other.  E.g.

```
from . import plugin
from .genfoundry import (
    ClaudeCodeAgent, CodexAgent, AgentOptions, AssistantMessage, TextBlock,
    PermissionResultAllow, PermissionResultDeny)
from .genfoundry.claude_agent import find_claude_cli
from .genfoundry.codex_agent import find_codex_cli
from .chatprocessor import ClaudeMessageProcessor, CodexMessageProcessor
```
is not allowed. 

All modules/files you import must come from a subdir because we cannot guarantee that all plugins are atomically on the same version/revision (at reload/import time).  

That is why you often see a single entrypoint `plugin.py` and everything else in `core/`.  And `plugin.py` then basically only does e.g. a `from .core import *` and has reloader code in it.  E.g.

```
import sys

# kiss-reloader:
prefix = __spec__.parent + "."  # don't clear the base package
for module_name in [
    module_name
    for module_name in sys.modules
    if module_name.startswith(prefix) and module_name != __name__
]:
    del sys.modules[module_name]
 ```

(https://github.com/kaste/KissReloader)

That's a bit of a stupid work that copy pasting but hopefully not that complicated.

EDIT: Otherwise LGTM!

synote:
The core sublime command files have been moved into a dedicated [chatview](https://github.com/flashmodel/termmate/tree/main/chatview) sub-dir
The package now uses a clean entry point (termmate.py) at the root directory.

braver:
Don't forget to tag your latest work!

synote:
New tag now support latest package layout. 
I'll try to provide unified agent message for genfoundry module next phase, thus adding new agent could be more easy. Currently, the core functionality is focused on getting codex and claude up.

PR #9363, opened 2026-04-05
added bitbake syntax for sublime 4

wahajmurtaza:

[2]: https://semver.org
[4]: https://github.com/Driim/bitbake-syntax/
My package is similar to [bitbake syntax][4] However it should still be added because the old one doesn't support sublime text 4. I needed to create my own to work with bitbake. Also I added version to be below 4 for old one

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -902,7 +902,18 @@
 			"labels": ["language syntax"],
 			"releases": [
 				{
-					"sublime_text": ">3084",
+					"sublime_text": ">3084 <4000",
+					"tags": true
+				}
+			]
+		},
+		{
+			"name": "Bitbake Syntax ST4",
+			"details": "https://github.com/wahajmurtaza/sublime-text-4-bitbake-syntax",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": ">=4000",
 					"tags": true
 				}
 			]
--- a/repository/b.json
+++ b/repository/b.json
@@ -902,7 +902,7 @@
 			"labels": ["language syntax"],
 			"releases": [
 				{
-					"sublime_text": ">3084 <4000",
+					"sublime_text": "<4000",
 					"tags": true
 				}
 			]


wahajmurtaza:
I fixed the `">3084 <4000"` . 
Regarding the PR to the original repo, this syntax doesn't work well with ST3. So have to keep 2 different packages.

rchl:
It can be the same repo with two entries, one using tag with a prefix like `st3-`.

It would be better to not create duplicate syntax as it creates confusion and the existing one is already referenced in some places, for example https://github.com/sublimelsp/LSP-bitbake/blob/823046c2896ce03341b181cf474a5661a77b334b/README.md#L8-L8

The author of original syntax is active in the community. I'm sure that a deal can be made here to update it if yours is better.

rchl:
Actually I was a bit confused - I was thinking of another syntax. It appears there are 3 now:
- one on package control - https://github.com/Driim/bitbake-syntax/
- one referenced from LSP-bitbake by @huyhoang8398 - https://github.com/huyhoang8398/BitBake-Syntax
- yours - https://github.com/wahajmurtaza/sublime-text-4-bitbake-syntax

I think that a one good syntax should come out of this.

rchl:
Also, what makes you say that the original syntax is only for ST3? I haven't tried it but it looks like a modern syntax that should work in ST4.

(The package mentioning ST3 is probably just a short-sighted comment made at the time of creation because that was the latest version then)

wahajmurtaza:
@rchl the driim package crashes when a bb file is opened. Please note that I have found that it is only occurring in my ubuntu machine and this works on my windows system.
I was using this package on ST3 on ubuntu before I updated and needed fixing.

Regardless @huyhoang8398 package is not in the package-control, will needs testing on linux machine.


324df294-62b1-40c8-b422-b1054de3f92c

huyhoang8398:
hello, im the maintainer of LSP-bitbake and also made my own Bitbake syntax due to the original Bitbake syntax cannot working properly currently due to `with_prototype` , as @rchl suggested, i also think best way is to open PR to the original repo, or ask if he could transfered that project to us, that's why i closed my PR https://github.com/sublimehq/package_control_channel/pull/9325.

braver:
Since @Driim hasn't responded yet, let's say we replace the original package. The new repo should have one tag on it for ST3 as @rchl suggests, and a new tag range for ST4+. And then we can move all of this forward. Make sense @wahajmurtaza?

kaste:
A canonical way is

```
        {
            "name": "Bitbake Syntax",
            "details": "https://github.com/wahajmurtaza/sublime-text-4-bitbake-syntax",
            "author": ["Driim", "wahajmurtaza"],
            "releases": [
                {
                    "sublime_text": ">=4000",
                    "tags": true,
                },
                {
                    "sublime_text": "<4000",
                    "tags": true,
                    "base": "https://github.com/Driim/bitbake-syntax"
                }
            ]
        },
```

if I'm reading @braver correctly and we want @Driim's version for older ST builds.  If so, this is the transparent change.

braver:
@wahajmurtaza so you still want to proceed?

PR #9366, opened 2026-04-07
Add Base64AutoCharset package

emotion:
- [x] If my package is a syntax it doesn't also add a color scheme. *** (N/A - not a syntax)

[2]: https://semver.org

My package is a Base64 encode/decode tool with automatic charset detection. It supports decoding Base64 with automatic detection of the original charset (UTF-8, GB18030/GBK/GB2312, Big5, EUC-JP, Shift_JIS, EUC-KR, etc.), and encoding text to Base64 with either GB18030 or UTF-8. This is particularly useful for debugging CJK email content (MIME headers often use Base64 with non-UTF-8 encodings) and API data inspection.

There are no packages like it in Package Control. Existing Base64 packages (e.g. `Base64 Fold`) only handle raw byte encode/decode and do not perform charset detection, which makes them unusable for GB18030/GBK encoded content common in Chinese enterprise email systems.

**Repository**: https://github.com/emotion/base64-decode-autocharset
**Latest release**: [1.0.2](https://github.com/emotion/base64-decode-autocharset/releases/tag/1.0.2) (uses `"tags": true` so Package Control tracks all semver tags automatically)

Verified with `st_package_reviewer`: 0 failures, 0 warnings across all 19 checkers.

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -230,6 +230,17 @@
 			]
 		},
 		{
+			"name": "Base64AutoCharset",
+			"details": "https://github.com/emotion/base64-decode-autocharset",
+			"labels": ["base64", "encoding", "charset", "gb18030", "decode", "cjk"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"branch": "main"
+				}
+			]
+		},
+		{
 			"name": "Bash Build System",
 			"details": "https://github.com/macite/sublimebashbuildsystem",
 			"labels": ["build system"],
--- a/repository/b.json
+++ b/repository/b.json
@@ -232,7 +232,7 @@
 		{
 			"name": "Base64AutoCharset",
 			"details": "https://github.com/emotion/base64-decode-autocharset",
-			"labels": ["base64", "encoding", "charset", "gb18030", "decode", "cjk"],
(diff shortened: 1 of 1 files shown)


braver:
Please tell your agent to use the PR template.

emotion:
Thanks for the feedback @braver! I've updated the PR to follow the template, and made the following compliance changes to the package:

- Removed `Context.sublime-menu` (no context menu entries)
- Removed `Default.sublime-keymap` (no key bindings; suggested keymap moved to README)
- Added `.gitattributes` to exclude test files and docs from the package
- Translated README to English
- Renamed commands file to `Default.sublime-commands`
- Tagged releases up to [v1.0.2](https://github.com/emotion/base64-decode-autocharset/releases/tag/1.0.2)
- Updated this entry to use `"tags": true` instead of tracking the main branch

Verified with `st_package_reviewer`: 0 failures, 0 warnings across all 19 checkers.

Please take another look when you have time.

braver:
Looks good! All I have to add is to recommend opting into the python 3.8 runtime ([example](https://github.com/braver/ColorConverter/blob/main/.python-version)).

PR #9369, opened 2026-04-08
fix(MarkerStack):  fix required sublime version

vwheeler63:
This PR fixes an erroneous ST version requirement in my previous `MarkerStack` Package submission.

[2]: https://semver.org

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -759,7 +759,7 @@
 			"donate": "https://buymeacoffee.com/vwheeler63",
 			"releases": [
 				{
-					"sublime_text": ">=3000",
+					"sublime_text": ">=4050",
 					"tags": true
 				}
 			]


vwheeler63:
Issue with Package is now fixed -- checks can be re-initiated.

braver:
Let's ignore that about the keymap file, it's fine :)

vwheeler63:
> Let's ignore that about the keymap file, it's fine :)

Thank you for the input....  I was going to say that it might be a challenge to get the reviewer software to tell when it's an empty JSON object filled with comments that are in alignment with the guidance on submitting packages, namely:
```
We strongly advice against adding keybindings by default. There aren't 
enough keys for all packages, so you risk overriding those of other packages. 
Instead, provide commented out suggestions in a keymap file, and/or explain 
how to create bindings in your README.
```

braver:
Yeah, we noted that on the repo for the test script, just haven't found the right way to implement it I guess. 🤷🏻‍♂️

PR #9371, opened 2026-04-08
Add Viv (DSL for games and simulations)

james-owen-ryan:
- [x] I have read the docs.
- [ ] If my package is a syntax it doesn't also add a color scheme.

My package is a syntax definition and compiler integration for [Viv](https://viv.sifty.studio), a domain-specific language (DSL) for emergent narrative in games and simulations. It provides syntax highlighting, six optional color themes, autocompletion, and integration with the Viv DSL compiler via the Sublime build system.

There are no packages for Viv in Package Control.

Notes on unchecked items:
- **Preferences/split view:** The package has no user-facing preferences or keybindings.
- **Color schemes:** The six bundled Viv themes are optional, and the syntax works with any color scheme. The themes provide finer-grained semantic distinctions specific to the Viv language.
- **gitattributes:** The package is self-hosted. The CD workflow builds the `.sublime-package` zip with an explicit file list, so `.gitattributes` export–ignore is not applicable.

Notes with regard to the reviewer guidelines:

* **`.no-sublime-package`:** Present in the zip. The package's build system invokes a Python bridge script (`bridge/compiler_bridge.py`) via the host system's Python, which requires the file to be on disk. This falls under "running scripts."
* **`repository.json` includes a custom `compilerVersion` field:** This is read by the bridge script at runtime to check compiler compatibility. Package Control ignores it.
* **No online services or data transmission.** The Viv compiler runs locally.
* **Tests:** `ChannelRepositoryTools` would not recognize the cloned repo (`"Please open the package_control_channel folder"`), so I couldn't run the tests. But I have confirmed that the package installs and functions correctly via the self-hosted `repository.json`.

### DIFF
--- a/channel.json
+++ b/channel.json
@@ -88,6 +88,7 @@
 		"https://raw.githubusercontent.com/seanliang/ConvertToUTF8/master/packages.json",
 		"https://raw.githubusercontent.com/sentience/DokuWiki/master/packages.json",
 		"https://raw.githubusercontent.com/sentience/HyperlinkHelper/master/packages.json",
+		"https://raw.githubusercontent.com/siftystudio/viv/main/plugins/sublime/repository.json",
 		"https://raw.githubusercontent.com/sokolovstas/SublimeWebInspector/master/packages.json",
 		"https://raw.githubusercontent.com/soncy/AutoComments-for-Sublime-Text-2/master/packages.json",
 		"https://raw.githubusercontent.com/spectacles/CodeComplice/master/packages.json",


braver:
Just curious, does the `.no-sublime-package` still do anything if the package is actually already supplied in this way as a `.sublime-package` file? I.e. would PC unzip it upon installation (cc @deathaxe)?

Providing the package via a repository json just because the source is inside a monorepo is a bit unfortunate. But I can see how you'd want to do it that way. In any case it seems to be valid and it's good you're using v4. 

Note that your readme mentions "Color Themes", which is not a thing in Sublime Text. It's either a "color scheme" or a "theme" and in this case they're all color schemes.

> Note that selecting a Viv color scheme will replace your current Sublime Text color scheme globally.

It is actually possible to set the color scheme as a syntax specific setting. That way a Viv color scheme would only affect Viv files.

james-owen-ryan:
First off, thanks for taking the time to review!

> Just curious, does the `.no-sublime-package` still do anything if the package is actually already supplied in this way as a `.sublime-package` file? I.e. would PC unzip it upon installation (cc [[@deathaxe](https://github.com/deathaxe)](https://github.com/deathaxe))?

Yep, my testing indicates that PC does still unzip it on install, and without `.no-sublime-package` it does not. That extraction matters here because `Viv.sublime-build` invokes a Python bridge script (`bridge/compiler_bridge.py`) as an external `python3` subprocess, and that subprocess can't read from inside a zipped `.sublime-package`.

> Providing the package via a repository json just because the source is inside a monorepo is a bit unfortunate. But I can see how you'd want to do it that way. In any case it seems to be valid and it's good you're using v4.

Yeah, it's a bit arcane, but I couldn't find a better way given our monorepo structure. And I'd rather not spin off a dedicated repo for this package, because its dependency on the DSL compiler means they tend to change together, which is easier to manage in a single repo.

> Note that your readme mentions "Color Themes", which is not a thing in Sublime Text. It's either a "color scheme" or a "theme" and in this case they're all color schemes.

Oh, shoot, thank you for the correction. I had totally missed that distinction, but it's now rectified across the README and package code.

> > Note that selecting a Viv color scheme will replace your current Sublime Text color scheme globally.
>
> It is actually possible to set the color scheme as a syntax specific setting. That way a Viv color scheme would only affect Viv files.

Neat! I looked into it, and then updated the README to explain this alternative.

deathaxe:
Predefining color schemes in "syntax specific" settings is not that cool as it may cause significant hard to disable visual disruption. Earlier versions of MarkdownEditing did so and it was always terrible to get a white MD file with otherwise dark color scheme being set, globally. Don't do that.

james-owen-ryan:
@deathaxe So just to be clear, I'm not shipping a `.sublime-settings` file with a predefined color scheme, but rather the README now contains instructions ([here](https://github.com/siftystudio/viv/blob/main/plugins/sublime/README.md#color-schemes)) for manually scoping a Viv color scheme to `.viv` files, using `Preferences: Settings - Syntax Specific`. This is presented as an alternative to setting a Viv color scheme globally.

That said, I do see what you're saying. I can imagine a user scoping to `.viv` files, forgetting that they did this, and getting annoyed when the editor e.g. switches from dark to light for those files.

Do you feel like syntax-specific color schemes are a bad idea in general, or do you still see value for users who manually opt in?

braver:
> Do you feel like syntax-specific color schemes are a bad idea in general, or do you still see value for users who manually opt in?

I guess that for the user to figure out. As long as it's opt-in it's fine. We've seen cases where authors will want to push a color scheme, and that's not great. It's fine if your syntax has specific needs, but you can't be too opinionated about that.

In fact, I'm not sure what your needs are with the syntax, but perhaps users could extend the one they like with some features you could suggest. Or perhaps between the several options you provide there is already plenty to choose from. 

In any case, nice job on the package and thanks for responding to the feedback.

PR #9372, opened 2026-04-12
Add LSP-Path-Scoping package

skovsboll:
Adds [LSP-Path-Scoping](https://github.com/skovsboll/Sublime-LSP-Path-Scoping), a Sublime Text 4 package that enables directory-based LSP server scoping.

This allows routing different LSP servers to different project directories — for example, directing LSP-Deno to `backend/` and LSP-typescript to `frontend/` in a mixed Deno/Node monorepo.

- **Repository:** https://github.com/skovsboll/Sublime-LSP-Path-Scoping
- **Tag:** 1.0.0
- **License:** MIT
- **Requires:** Sublime Text >= 4000, LSP package

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -1856,6 +1856,17 @@
 			]
 		},
 		{
+			"name": "LSP-Path-Scoping",
+			"details": "https://github.com/skovsboll/Sublime-LSP-Path-Scoping",
+			"labels": ["lsp"],
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Lua Dev",
 			"details": "https://github.com/rorydriscoll/LuaSublime",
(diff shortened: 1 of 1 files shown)


rchl:
It's not a good approach to overwrite internal code of the package.

In fact one of existing PR already will break your code (or your code will break LSP) because the signature of that method changes - https://github.com/sublimelsp/LSP/blob/569e2bf2186869be675e3d882994f441bb272dd5/plugin/core/types.py#L1034-L1036

You should instead make a feature request to LSP, discuss it and potentially provide a PR to implement something similar.

Or keep monkey patching but don't release it to the public as it's just asking for trouble.

braver:
Indeed, while I have no immediate opinion the contents of this package, this either goes into LSP directly, or in their repository for plugins. Please see https://github.com/sublimelsp for more information.

PR #9375, opened 2026-04-13
add Kamailio Syntax plugin

igorolhovskiy:

[2]: https://semver.org

My package is a syntax highlighter for [Kamailio](https://www.kamailio.org/w/) SIP Proxy config.

### DIFF
--- a/repository/k.json
+++ b/repository/k.json
@@ -36,6 +36,17 @@
 			]
 		},
 		{
+			"name": "Kamailio Config",
+			"details": "https://github.com/igorolhovskiy/kamailio-sublime",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "KanaKana",
 			"details": "https://github.com/mitsu-ksgr/KanaKana",
 			"labels": ["japanese", "text manipulation"],


huyhoang8398:
I would suggest to rename it as kamailio-syntax to follow with other plugin

igorolhovskiy:
@huyhoang8398 You means repo (kamailio-sublime -> kamailio-syntax) or something else?

braver:
The package and name are fine. Don't forget to tag a version (e.g. 1.0.0) on your repo, Package Control only picks up on tags.

igorolhovskiy:
Maybe I'm doing something wrong, asking for help here.
I have a tag - https://github.com/igorolhovskiy/kamailio-sublime/tags
But not sure if it's "semantic" or no.
If I'm trying to run 'git tag 0.1', I'm getting "tag exists".
Sorry for dumb questions, I'm not really familiar with a git tags.

braver:
"semantic" means following these specs: semver.org. In short, you need 3 bits, not 2. So 0.1.0 instead of 0.1 should work.

igorolhovskiy:
Thanks, will do a new MR

braver:
That's really not necessary. Just publish a tag on your repo and let us know in this PR, and then we can merge it.

PR #9385, opened 2026-04-22
Add Better Go To File package

Suor:
- [x] I have read the docs.
- [x] I have tagged a release with a semver version number. Latest tag: 1.0.2.

My package is a small command that opens a file path under the cursor. It supports project-relative paths, absolute paths, line numbers, multiple matches via a quick panel, and falls back to goto_definition when no path resolves.

It is similar to GoToFile. However it should still be added because that package is ST2-only in Package Control, uses a broad regex search over project files, and exposes a different command. Better Go To File is a separate ST3+ implementation focused on deterministic path-under-cursor resolution.

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -614,6 +614,17 @@
 			]
 		},
 		{
+			"name": "Better Go To File",
+			"details": "https://github.com/Suor/GoToFile",
+			"labels": ["file open", "code navigation"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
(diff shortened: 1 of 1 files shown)


braver:
Your package is called "Better Go To File" so all file paths (e.g. in edit_settings) will use that, with spaces. You should also use that to refer to your package, and not sometimes "Go To File". The command palette command for your settings then becomes, canonically, "Preferences: Better Go To File Settings".

Suor:
Ok, renamed to BetterGoToFile, no spaces. Updated everythere to be consistent

PR #9386, opened 2026-04-22
Add RememberSyntax package

Suor:
- [x] I have read the docs.
- [x] I have tagged a release with a semver version number. Latest tag: 1.0.0.

My package remembers manually selected syntax for files and restores it when those files are opened again. It also supports basename glob rules for files with generic names or no extensions.

There are no packages like it in Package Control that I am aware of.

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -1091,6 +1091,17 @@
 			]
 		},
 		{
+			"name": "RememberSyntax",
+			"details": "https://github.com/Suor/RememberSyntax",
+			"labels": ["utilities"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Remify",
 			"details": "https://github.com/brousalis/remify",
 			"releases": [


guille:
There is https://github.com/facelessuser/ApplySyntax

Also note some of the behaviour is available through standard ST with open extension as...

michaelblyons:
> There is https://github.com/facelessuser/ApplySyntax

Also AutoSetSyntax. Neither of them has quite this UX, but the end result is the same.

Suor:
The rules based thing is secondary to my plugin, the core workflow - open a file, it has some weird name like ”service” and not matched in any way, so you just set its syntax manually and it's remembered. That's it. 

The rules based approach are covered by the above mods better.

braver:
Indeed, seems like a new approach to this, and I must say I really like how it's simple and easy to understand. 

Only one small remark: as with your other submission the command palette entry for settings typically follows the "Preferences: Package Name Settings" pattern.

Suor:
Fixed

braver:
Your settings say "Per-file saved syntaxes (managed automatically, do not edit manually).". Wouldn't it actually be ok to edit this? As long as users don't mess up the structure of that data, it should actually just work right?

braver:
Anyway, that's just a question, let's ship it :)

Suor:
> Your settings say "Per-file saved syntaxes (managed automatically, do not edit manually).". Wouldn't it actually be ok to edit this? As long as users don't mess up the structure of that data, it should actually just work right?

Yeah, it would work as long as you don't mess it up. I guess I should rephrase this.

michaelblyons:
Oof, there was indeed already "Syntax History."

PR #9387, opened 2026-04-22
Update Sublime Text version requirement for CodeMap

oleg-shilo:

[2]: https://semver.org

My package is [CodeMap](https://packagecontrol.io/packages/CodeMap). It is an existing package that is already published on [package_control_channel](https://github.com/sublimehq/package_control_channel) and has been working properly since 2017. I only updated the Update Sublime Text version compatibility information in the `c.json` file.  I have also tested the package with the latest ST4.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2266,7 +2266,7 @@
 			"details": "https://github.com/oleg-shilo/sublime-codemap",
 			"releases": [
 				{
-					"sublime_text": ">=3000",
+					"sublime_text": ">=4000",
 					"tags": true
 				}
 			]


rchl:
There are many references to ST3 and Sublime Text 3 in your repo.
If it now works only with ST4 then those should be updated. Ideally don't mention version - just say Sublime Text.

oleg-shilo:
Thank you @rchl,

Indeed, the codebase was developed a long time ago and has not needed any updates since then. And only the direct request from one of the plugin users has triggered this PR. Otherwise, I wouldn't even bother with the update.

Would you mind elaborating on your comment?

- The plugin works with both ST3 and ST4. Does it mean that this PR is not required?
- `> don't mention version - just say Sublime Text.`
   Are you referring to the `"sublime_text": ">=4000",` or something else? 
   
Thank you

rchl:
I thought you've updated the package to only work from >=4000. If not and it still works with >=3000 then that also covers >=4000 so there is no need to change anything.

I think it's just a matter of user being confused because the package states that it supports ST3 but doesn't mention ST4. Not referencing the number anywhere should be less confusing.

oleg-shilo:
OK, then it was my mistake. I will close the PR then.
Thank you for your help.

PR #9388, opened 2026-04-22
C Family Rewrite (as a standalone package)

camila314:
- [ x ] I'm the package's author and/or maintainer.
- [ x ] I have read [the docs][1].
- [ x ] I have tagged a release with a [semver][2] version number.
- [ x ] My package repo has a description and a README describing what it's for and how to use it.
- [ x ] My package doesn't add context menu entries. *
- [ x ] My package doesn't add key bindings. **
- [ x ] Any commands are available via the command palette.
- [ x ] Preferences and keybindings (if any) are listed in the menu and the command palette, and open in split view.
- [ x ] If my package is a syntax it doesn't also add a color scheme. ***
- [ x ] I use [.gitattributes][3] to exclude files from the package: images, test files, sublime-project/workspace.

[2]: https://semver.org

My package is ...

The [C Family Rewrite](https://github.com/sublimehq/Packages/pull/4147) as a standalone package, in all its glory. C, C++, Objective-C, and Objective-C++. Modern, bugs fixed, and with 2023-level features for each language.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2,7 +2,7 @@
 	"schema_version": "3.0.0",
 	"packages": [
 		{
-			"name": "C Family Rewrite (C23/C++23)",
+			"name": "C Family Rewrite (C23 and C++23)",
 			"details": "https://github.com/camila314/C-Family-Rewrite",
 			"labels": ["language syntax", "c", "c++", "objc", "objc++", "objective-c"],
 			"author": "camila314",
--- a/repository/c.json
+++ b/repository/c.json
@@ -6,7 +6,7 @@
 			"details": "https://github.com/camila314/C-Family-Rewrite",
 			"labels": ["language syntax", "c", "c++", "objc", "objc++", "objective-c"],
 			"author": "camila314",
-			"description": "Complete modern rewrite of C/C++/Objc/ObjC++ syntax",
+			"description": "Complete modern rewrite of C/C++/ObjC/ObjC++ syntax",
 			"releases": [
 				{
 					"sublime_text": ">=4199",
--- a/repository/c.json
+++ b/repository/c.json
@@ -5,7 +5,6 @@
 			"name": "C Family Rewrite (C23 and C++23)",
 			"details": "https://github.com/camila314/C-Family-Rewrite",
 			"labels": ["language syntax", "c", "c++", "objc", "objc++", "objective-c"],
-			"author": "camila314",
 			"description": "Complete modern rewrite of C/C++/ObjC/ObjC++ syntax",
 			"releases": [
 				{
--- a/repository/c.json
+++ b/repository/c.json
@@ -9,7 +9,7 @@
 			"description": "Complete modern rewrite of C/C++/Objc/ObjC++ syntax",
 			"releases": [
 				{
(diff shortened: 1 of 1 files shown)


repository/c.json
```diff
@@ -1,6 +1,19 @@
 {
 	"schema_version": "3.0.0",
 	"packages": [
+		{
+			"name": "C Family Rewrite (C23 and C++23)",
+			"details": "https://github.com/camila314/C-Family-Rewrite",
+			"labels": ["language syntax", "c", "c++", "objc", "objc++", "objective-c"],
+			"author": "camila314",
```
michaelblyons:
```suggestion
```
Author is automatically derived from the GitHub user.

repository/c.json
```diff
@@ -1,6 +1,19 @@
...
+			"description": "Complete modern rewrite of C/C++/Objc/ObjC++ syntax",
```
michaelblyons:
```suggestion
			"description": "Complete modern rewrite of C/C++/ObjC/ObjC++ syntax",
```
Fix capitalization, OR


```suggestion
```
Let description come from the Repository description (the sidebar blurb).

repository/c.json
```diff
@@ -1,6 +1,19 @@
...
+			"releases": [
+				{
+					"sublime_text": ">4200",
```
michaelblyons:
```suggestion
					"sublime_text": ">=4199",
```
Do you really need the "secret" alpha builds to run this?

michaelblyons (changes requested):

michaelblyons (approved):
LGTM. I hope you will be willing to upstream if/when this matures.

braver:
Wow, what a project. As someone who's dabbled in syntaxes from time to time I have a lot of respect for the depth of the rewrite. The default packages, and some in the SublimeText org, are of an amazingly high standard. I would never dream of attempting getting close for any language, let alone this set. So, you're looking to let this sit out in the wild for a while to for increased visibility and perhaps someone else might want to contribute?

Just to double check a few things with you that will be hard to change later:
- You're happy with the name of the package? It's long-ish, but fine by me.
- You want to have the description on the entry here, instead of just taking it from the repo? They're different, so I can how you'd want that. But if you remove it here, it will be take from the repo and that way you can change it any time you like.
- If someone were to install this, how would that play out vs the default package? Not sure if there is a lot of inheritance between or dependencies upon the default syntaxes. But the readme could perhaps guide the user a bit in how to have Sublime pick these syntaxes vs any other?

camila314:
Yes exactly, I put a lot of effort into this and it would be quite a shame if it was all for nothing, even if it's not default package material.

I chose the package name because I noticed when searching for packages myself that the search tool doesn't take description into account, so I wanted to ensure that anyone who was potentially searching for modern C/C++ syntaxes would have it pop up.

I chose to have the description differ from the repository because I wanted to clarify in the repo that it was a syntax specifically for Sublime Text and I figured that was self-evident on the package index.

The way I designed the packages there is zero dependency on any of the default syntaxes, only dependency within themselves. I did just update the README to have some basic instructions on setting them as default.

braver:
Alright, let's go then!

PR #9391, opened 2026-04-24
Move Kotlin package

guille:

[2]: https://semver.org

Repo: https://github.com/guille/sublime-kotlin

My package is a Kotlin syntax. Rather than have a "Better Kotlin" package I figured this made more sense.

Two disclaimers:
- The syntax is AI generated. It does fix every open issue I saw on the old package's (and then some) but it is probably not perfect. I am not good at writing syntaxes so any maintaining will also be done with this approach, although syntax tests seem to make it output decent stuff.
- I have not tried to submit patches [the old package](https://github.com/vkostyukov/kotlin-sublime-package). The last commit was 8 years ago, there have been open issues for longer, and I don't feel comfortable dumping a 1000-line syntax file on someone else.

I would understand if the change is rejected. I initially wasn't going to submit it because I don't think a lot of people use ST for Kotlin but it's been pointed out that Sublime Merge would also benefit from a better syntax.

The only "breaking" change beyond better highlighting is the scope name changes from `source.Kotlin` to `source.kotlin`. The only places I can imagine that would cause issues is in [LSP-Kotlin](https://github.com/search?q=repo%3Asublimelsp%2FLSP-kotlin%20source.kotlin&type=code) and [AFileIcon](https://github.com/search?q=repo%3ASublimeText%2FAFileIcon%20source.kotlin&type=code). If this is accepted I'd submit patches there too.

## Comparison

**Before**

---

**After:**

### DIFF
--- a/repository/k.json
+++ b/repository/k.json
@@ -484,7 +484,12 @@
 			"labels": ["language syntax"],
 			"releases": [
 				{
-					"sublime_text": "*",
+					"sublime_text": "<4075",
+					"base": "https://github.com/vkostyukov/kotlin-sublime-package",
+					"branch": "master"
+				},
+				{
+					"sublime_text": ">=4075",
 					"tags": true
 				}
 			]
--- a/repository/k.json
+++ b/repository/k.json
@@ -480,12 +480,12 @@
 		},
 		{
 			"name": "Kotlin",
-			"details": "https://github.com/vkostyukov/kotlin-sublime-package",
+			"details": "https://github.com/guille/sublime-kotlin",
 			"labels": ["language syntax"],
 			"releases": [
 				{
 					"sublime_text": "*",
-					"branch": "master"
+					"tags": true
 				}
 			]
 		},


guille:
(Alternatively, @vkostyukov if you want to maintain it and cut a new tag in your repo please let me know)

repository/k.json
braver:
```suggestion
					"tags": true
```

braver:
I'm totally ok with transferring the package, although @vkostyukov is of course welcome to object. Let's give that about 2-3 weeks.

If we're doing that, we should als move to tag-based releases and that will also ensure your work gets picked up by existing installations. 

I would also propose to make some changes to the readme. Maybe reference the previous package so that people can go back if they need to. And typically a link to language documentation, and stuff like how-to-contribute etc. Like, what are the yaml tests for and how does one work with those? I get the sense that LLMS should be pretty good at this stuff, especially if you start from good syntax tests, and give it good context. Which might also be something you'd want to put in the readme so others can contribute fixes in a way that fit your approach here.

braver:
Oh, once we're ready to merge, it would be good to _first_ update the depending packages to be forwards-compatible.

repository/k.json
guille: Good catch, pushed!

FichteFoll:
Since this package provides a syntax definition with `version: 2`, it should target builds 4075+. Any older (notably ST3) builds could be redirected to the old repo still by specifying it as the release channel's `base`.

guille:
Updated the json to restrict to ST4075, I think I did it correctly but please give it a lookover.

Also added some meat to the [README](https://github.com/guille/sublime-kotlin#kotlin-syntax-for-sublime-text), esp. around the contributing part

As for the other packages with `source.Kotlin`, both PRs raised.
- The LSP-Kotlin was forward-compatible and has already been merged (just need to make sure there's a release including the commit before we merge here).
- AFI has a PR raised but I don't think it can be completely forward/backwards compatible, so I've put it in draft and hopefully that gets in afterwards.

guille:
Is it time @braver

braver:
It is time!

PR #9392, opened 2026-04-25
Add A6s

Imchrisrivers:

My package is **A6s** — a Sublime Text client for the local A6s CLI daemon. It provides intelligent multi-agent orchestration inside Sublime: invoke AI agents, stream RIGOR-phase progress, and review/apply generated refactors and tests. All traffic flows over `ws://localhost:9876/ws` to a locally-installed daemon (`a6s` Homebrew formula); the package itself never holds API credentials or speaks to a remote orchestrator.

There are no packages like it in Package Control today — most AI-assist plugins target a single hosted vendor; A6s is a thin client over a local daemon that owns the upstream call. Sibling clients exist for VS Code, JetBrains, Neovim, and Emacs.

Notes on the checklist:
- **Context menu**: A6s adds a single nested submenu under right-click ("A6s") with selection-conditional commands (`selection_empty == false`), so it only surfaces when the user has highlighted code.
- **Key bindings**: A6s ships **no default key bindings**. `Default.sublime-keymap` contains commented-out suggestions only — users copy what they want into their own keymap.

Repo: https://github.com/The-Autonoma/a6s-sublime
Latest release: v2.0.1

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -14,6 +14,17 @@
 			]
 		},
 		{
+			"name": "A6s",
+			"details": "https://github.com/The-Autonoma/a6s-sublime",
+			"labels": ["ai", "tools", "code", "websocket"],
+			"releases": [
+				{
+					"sublime_text": ">=3211",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "ABAP",
 			"details": "https://github.com/flaiker/abap-st3",
 			"labels": ["language syntax"],


braver:
Haven't dug into the code yet, just some superficial stuff from an initial quick look:

- Since notes about installing stuff seems to mostly reference homebrew, does any of this actually run on something other than MacOS?
- Your [dependencies](https://github.com/The-Autonoma/a6s-sublime/blob/main/dependencies.json) file seems to not list any?
- Typically settings are also listed in the command palette (as "Preferences: A6s Settings").

braver:

PR #9394, opened 2026-04-25
Update kaste/st_package_reviewer

kaste:
Update both package review workflow phases to use the current st_package_reviewer action revision.

Also limit package tests to actual registry changes.


braver:
Thanks!!

PR #9395, opened 2026-04-26
Drop dead packages

deathaxe:
This PR drops dead packages, which error out with 404 for at least a couple of days.


kaste:
Hi @deathaxe, I have prepared https://github.com/sublimehq/package_control_channel/pull/9390 to make that automatic in the future.

michaelblyons (approved):

PR #9397, opened 2026-04-29
Add StyleTokenHighlighter

leoshome:

[2]: https://semver.org

My package is similar to **StyleToken** and **TextMarker**, both of which are years old and no longer maintained. However it should still be added because:

1. It combines the features of them, one is right-click context menu of **StyleToken** and one is easy mark selected text feature of **TextMarker**.
2. It is designed to mimic **Notepad++'s "Style all occurrences of token"** via the right-click context menu, which is more intuitive and familiar for users coming from Notepad++.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -4297,7 +4297,6 @@
 			]
 		},
 		{
-			"name": "StyleTokenHighlighter",
 			"details": "https://github.com/leoshome/StyleTokenHighlighter",
 			"releases": [
 				{
--- a/repository/s.json
+++ b/repository/s.json
@@ -4297,6 +4297,16 @@
 			]
 		},
 		{
+			"name": "StyleTokenHighlighter",
+			"details": "https://github.com/leoshome/StyleTokenHighlighter",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "StylishHaskell",
 			"details": "https://github.com/hairyhum/SublimeStylishHaskell",
 			"releases": [


kaste:
LGTM.  If you ship a Context menu which is maybe okay for such a plugin, add to the README how to get rid of it.  Still preferred is maybe to *not* ship a Context but just a copyable example Context and tell the user in the README or even install message (messages/install.txt; likely your AI will know how to do that) that you think that this is a good idea.  Personally, I would not use it from the Context menu as that seems slow but bind either some colors to key chords, e.g. `ctrl+k, ctrl+num1 ...`, or just use the `style_token_toggle_highlight` on one binding (if I'm not mistaken) for light use.  

An improvement is also to add a settings file which basically holds the list of scopes, e.g.

```
{
    "colors": ["string", "keyword", "my_custom_name"]
} 
```

It is easy to add custom colors/scopes to Sublime's themes/schemes (which one, never remember the names).  Just defining 3 means you cycle just the three, *or* you fill in your defaults.  Still, that would mean the static Context menu *could* have too many entries.  🤔 Ask the AI to generate the menus on the fly.  LoL. 

I started with a LGTM.  🥂

leoshome:
@kaste Thanks for the review. Since Notepad++'s "Style all occurrences of token" doesn't have a key binding, I also prefer the Context menu 😅 over shortcut keys, so I didn't add a key binding. If some users request it, I can add it later.

braver:
Yeah, don't ship keybindings ;) They get problematic real quick. But it's good to have some guidance in the readme to help users do it themselves. Same goes for the context menu. You can't make them toggleable via settings, but that menu gets crowded a lot so it's good to give users control. And example of a readme that covers these topics other than help users customize: https://github.com/braver/ColorConverter#key-bindings

braver:

leoshome:
> Please have a look at the feedback.

Thanks for the feedback, and sorry for the late reply.

I've made the following changes based on your suggestions:

- Added `StyleTokenHighlighter.sublime-settings` — users can customize color scopes 
- Added README sections for context menu customization and key bindings setup 
- Added `Main.sublime-menu` for Preferences > Package Settings access
- Key bindings remain commented-out examples only, no active bindings shipped

Thanks again for the review!

kaste:
I just re-triggered the reviewer because it needed a style tweak.  Btw. instead of `super` vs `ctrl` I think you can use `primary` and hence reduce the three key bindings files to one platform-agnostic one.  E.g. Example.sublime-keymap or so.

leoshome:
> I just re-triggered the reviewer because it needed a style tweak. Btw. instead of `super` vs `ctrl` I think you can use `primary` and hence reduce the three key bindings files to one platform-agnostic one. E.g. Example.sublime-keymap or so.

Thanks for the tip! Consolidated into a single `Example.sublime-keymap.

kaste:
> 'Example.sublime-keymap' only contains commented examples. Consider defining 'Example.sublime-keymap' and linking it from 'Main.sublime-menu'.

LoL for the error message.  But indeed, the example is not loaded by Sublime Text so you don't need to comment the bindings.  That's the main (and only?) advantage, users can just copy/paste.

leoshome:
Hi, just added user_file to the Key Bindings entries in Main.sublime-menu and Default.sublime-commands, and uncommented the example bindings in Example.sublime-keymap.

kaste:
You already changed the entry here.  I tested against the old sha.  

LGTM! @braver

braver:
w00t!

PR #9399, opened 2026-04-30
Add syntax highlighting for the Praia language

viggou:

[2]: https://semver.org

My package is syntax highlighting for the Praia programming language.

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -2234,9 +2234,8 @@
 			]
 		},
 		{
-			"name": "Praia Syntax Highlighting",
 			"details": "https://github.com/praia-lang/praia-sublime",
-			"labels": ["praia", "language syntax"],
+			"labels": ["language syntax"],
 			"releases": [
 				{
 					"sublime_text": "*",
--- a/repository/p.json
+++ b/repository/p.json
@@ -2234,6 +2234,7 @@
 			]
 		},
 		{
+			"name": "Praia Syntax Highlighting",
 			"details": "https://github.com/praia-lang/praia-sublime",
 			"labels": ["language syntax"],
 			"releases": [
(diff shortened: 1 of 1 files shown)


braver:
Are the settings necessary? Typically these lean a bit opinionated, in which case it's better to recommend them in the readme rather than force them on users.

The syntax itself seems ok to me. I would recommend adding tests, as that will make it a lot easier to change things later without breaking nuances of the syntax: https://www.sublimetext.com/docs/syntax.html#testing

repository/p.json
```diff
@@ -2233,6 +2233,17 @@
 				}
 			]
 		},
+		{
+			"name": "Praia Syntax Highlighting",
+			"details": "https://github.com/praia-lang/praia-sublime",
+			"labels": ["praia", "language syntax"],
```
braver: From the [docs](https://docs.sublimetext.io/guide/package-control/submitting.html#pick-a-name):

> Language support (aka "syntax" or "grammar") packages are named after the language it supports, without suffixes like "syntax" or "highlighting".

```suggestion
			"details": "https://github.com/praia-lang/praia-sublime",
			"labels": ["language syntax"],
```

This would be the first package for "praia" support, so the label isn't meaningful at this point.

viggou:
Thanks, I've moved the settings to the readme and added a test.

repository/p.json
braver: Oops, looks like my suggestion also took out the package name 🤦🏻

viggou: whoops, added it back in

braver: Sorry, it should be "Praia" without the "Syntax Highlighting". ([docs](https://docs.sublimetext.io/guide/package-control/submitting.html#pick-a-name))

viggou: right, done

PR #9405, opened 2026-05-04
Update package reviewer action

kaste:
Point the package review workflows at the latest st_package_reviewer action commit.

This esp. adds a clickable link to the (GitHub) repo in the PR comment.

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Diff and review changed/added packages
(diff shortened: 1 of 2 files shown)


michaelblyons:
> This esp. adds a clickable link to the (GitHub) repo in the PR comment.

Thanks. I was considering adding that to the PR template, but your way is foolproof.

michaelblyons (approved):

braver:
> This esp. adds a clickable link to the (GitHub) repo in the PR comment.

Nice. I always wonder why more browser don't have the thing that Safari does where I can just right-click anything that looks like a link and then select "open in a new tab". With all the AI everywhere it's Apple of all companies that figured out how to parse a link and use it for stuff? 😉 😅 This is nice though, CMD-clicking is even easier.

PR #9408, opened 2026-05-06
Add Makinda Icons

makindajack:
- [x] I'm the package's author and/or maintainer.
- [x] I have read [the docs][1].
- [x] I have tagged a release with a [semver][2] version number.
- [x] My package repo has a description and a README describing what it's for and how to use it.
- [x] My package doesn't add context menu entries.
- [x] My package doesn't add key bindings.
- [x] Any commands are available via the command palette. (n/a — no commands)
- [x] Preferences and keybindings (if any) are listed in the menu and the command palette, and open in split view. (n/a — none)
- [x] If my package is a syntax it doesn't also add a color scheme. (n/a — not a syntax)
- [x] I use [.gitattributes][3] to exclude files from the package: images, test files, sublime-project/workspace.

[1]: https://docs.sublimetext.io/guide/package-control/submitting.html
[2]: https://semver.org
[3]: https://www.git-scm.com/docs/gitattributes#_export_ignore

My package is a file-icon theme for Sublime Text. It ships 16/32/48 px PNG icons and `.tmPreferences` files that bind TextMate scopes (e.g. `source.js`, `source.python`) to those icons so files get the right glyph in the sidebar and tabs.

My package is similar to A File Icon. However it should still be added because it follows a different visual language — warm, single-fill, optimized for legibility at 16 px — and is the icon half of the Makinda design system (the color scheme half is in #9409).

Repo: https://github.com/makinda-jackson/makinda-icons-sublime — tag `v1.0.13`, MIT.

Submission tidied up after some earlier formatting mistakes. I'll be handling responses on this PR personally.

PR #9409, opened 2026-05-06
Add Makinda Themes

makindajack:
- [x] Any commands are available via the command palette. (n/a — no commands)
- [x] Preferences and keybindings (if any) are listed in the menu and the command palette, and open in split view. (n/a — none)
- [x] If my package is a syntax it doesn't also add a color scheme. (n/a — this package *is* color schemes, not a syntax)

[2]: https://semver.org

My package is a pair of color schemes — Makinda Light and Makinda Dark — for Sublime Text 3 / 4, with warm orange accents and a focus on contrast and readability.

My package is similar to many existing color schemes in Package Control. However it should still be added because it's the canonical Sublime distribution of the Makinda design system, intended to pair with the Makinda Icons package (#9408).

Repo: https://github.com/makinda-jackson/makinda-sublime — tag `v1.0.4`, MIT.

Submission tidied up after some earlier formatting mistakes (an outdated README reference, a noisy channel diff, missing top-level LICENSE — all corrected). I'll be handling responses on this PR personally.


braver:
Your repo talks about a `source/palette.json` file that does not exist in your codebase.

It also looks like you formatted the entire JSON file here and submitted both packages in one PR. 

Don't take this the wrong way, but it looks like whatever agent you're using needs a little more babysitting.

makindajack:
Thanks for the review, @braver! You're right on all counts.

I'll fix the `source/palette.json` reference in the repo README, revert the formatting changes to `m.json` so the diff only includes my additions, and keep this PR scoped to Makinda Themes only.

Will push the fixes shortly.

PR #9410, opened 2026-05-06
Add InlineGitBlame to Package Control

akostoth02:

[2]: https://semver.org

My package is similar to Git Blame, however I would like to add it, because it shows the blame inline without any direct trigger. It works just like in VS Code for example. Similar, but still different :)
Also planning to add more useful features, maybe the exact change made for the line with the last commit. However i will try to keep it lean, as all things should be.

### DIFF
--- a/repository/i.json
+++ b/repository/i.json
@@ -757,15 +757,15 @@
 		},
 		{
 			"name": "InlineGitBlame",
-		    "details": "https://github.com/akostoth02/ST-InlineGitBlame",
-		  	"labels": ["git", "blame", "inline", "annotation"],
-	  		"releases": [
-		  		{
-			  		"sublime_text": ">=4000",
-			  		"tags": true
-			  	}
-	  		]
-	  	},
+			"details": "https://github.com/akostoth02/ST-InlineGitBlame",
+			"labels": ["git", "blame", "inline", "annotation"],
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"tags": true
+				}
+			]
+		},
 		{
 			"name": "InlineOutline",
 			"details": "https://github.com/kaste/InlineOutline",
--- a/repository/i.json
+++ b/repository/i.json
@@ -756,6 +756,17 @@
 			]
 		},
 		{
+			"name": "InlineGitBlame",
+		    "details": "https://github.com/akostoth02/ST-InlineGitBlame",
+		  	"labels": ["git", "blame", "inline", "annotation"],
+	  		"releases": [
+		  		{
+			  		"sublime_text": ">=4000",
+			  		"tags": true
+			  	}
+	  		]
+	  	},
+		{
 			"name": "InlineOutline",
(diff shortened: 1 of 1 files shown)


deathaxe:
FWIW, GitGutter optionally shows blame info inline and in statusbar.

akostoth02:
Yes, i just created such a package, because i do not use most of it's features. I want this to be a standalone package for inline blames.
Also this runs a bit smoother, as the blame is loaded for the full file, when it's opened. The phantom is served by a cache.

There are still things to work on, but I thought it's in a shape already that could be published.

kaste:
That looks fantastic!

Let's see what the reviewer bot says.

LGTM! @braver 

(braver is not the bot.)

kaste:
* `"base_file": "${packages}/InlineGitBlame/Default (InlineGitBlame).sublime-keymap",`

That's not a standard name, and then you must set ... the key ... "user_file"?  The reviewer will know that.  

Also add them to "commands", the `ctrl+shift+p` thing, Command Palette.  `Preferences: ...`

* Because you prepared for that, add at BOF in your root.py

```
import sys

# kiss-reloader:
prefix = __spec__.parent + "."  # don't clear the base package
for module_name in [
    module_name
    for module_name in sys.modules
    if module_name.startswith(prefix) and module_name != __name__
]:
    del sys.modules[module_name]
``` 
(https://github.com/kaste/KissReloader)

so that the package hot-reloads if you ship updates.

* There is bug 😭 .  In md files, e.g. a README, I cannot reliably use arrow-down.  It just jumps at EOL.

braver:
> (braver is not the bot.)

🤣

braver:
- Consider adding keybindings and settings entries to the command palette [(example)](https://github.com/braver/ColorConverter/blob/main/Commands.sublime-commands#L3-L15).
- Not sure what the bot it complaining about with the main menu entries, but have a look (also check @kaste's comments).
- Your entry in this PR has mixed indentation.
- While messages about updates are good, people do tend to find them annoying if there are a lot of them. So I would advise to limit it to changes in UI or settings that users need to know about.

kaste:
> Not sure what the bot it complaining about with the main menu entries

I update the reviewer.  The culprit is that "Inline Git Blame" is used but the package name is "InlineGitBlame".

akostoth02:
thanks for the feedback everyone, seems like i'm the bot here!
i'll get back to this soon, sometimes work just comes hard at the poor.

kaste:
The bug is with Word Wrap: on

off:


https://github.com/user-attachments/assets/4329b619-f867-4514-82b1-ad16e4f6c4a9

on:

https://github.com/user-attachments/assets/5398153c-c881-46cd-ae52-48472c0372a3

braver:

braver:
Ideally we don't leave PR's open for months. If you don't have time to address the feedback that's ok. We can close it so this is off our list of things to keep an eye on, and then when you're ready we can re-open or start anew.

PR #9411, opened 2026-05-06
Add ST4 Word Count package

dgdiddy:
Adds **ST4 Word Count**, a simple Sublime Text 4 plugin that shows a live word count in the status bar.

- Shows total word count of the document at all times
- When text is selected, also shows the word count of just the selection
- Works correctly with split panes
- ST4 only (`>=4000`)

Repository: https://github.com/dgdiddy/st4-wordcount

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -3741,6 +3741,18 @@
 			]
 		},
 		{
+			"name": "ST4 Word Count",
+			"description": "Live word count in the status bar, with selection count when text is selected",
+			"details": "https://github.com/dgdiddy/st4-wordcount",
+			"labels": ["word count"],
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "ST_AWK",
 			"details": "https://github.com/qiuxiafei/st_awk",
 			"releases": [


repository/s.json
```diff
@@ -3740,6 +3740,18 @@
 				}
 			]
 		},
+		{
+			"name": "ST4 Word Count",
```
kaste, line 3744: Names should not reference Sublime Text or ST because every plugin is a ST plugin.  Word Count is taken so you need to be creative here.

repository/s.json
```diff
@@ -3740,6 +3740,18 @@
...
+			"description": "Live word count in the status bar, with selection count when text is selected",
```
kaste, line 3745: description is not needed. If omitted we show the description from here: 

Image

So none, right now LoL, but you could of course copy paste that to there.  Easier to change for you in the future.

kaste:
Works in its naivety and innocence.  I *really* think that should be only on for specific files, in Sublime Text `selectors`, like md or rst or txt.  For code or logs or huge files it doesn't seem to be a good fit and it can be too slow for big files as well.  

For text files, it feels nice and useful.  👍

braver:
Please have a look at the feedback, and redo the OP by using the PR template.

PR #9413, opened 2026-05-06
Take over Smart Title Case

braver:

[2]: https://semver.org

My package is an updated fork of Smart Title Case originally by @mattstevens. That repo hasn't been updated in 14 years and has various outstanding issues and PR's, including mine which updates it for python 3.14: https://github.com/mattstevens/sublime-titlecase/pull/6

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -2535,6 +2535,7 @@
 		{
 			"name": "Smart Title Case",
 			"details": "https://github.com/braver/SmartTitleCase",
+			"author": ["mattstevens", "braver"],
 			"releases": [
 				{
 					"sublime_text": "*",
--- a/repository/s.json
+++ b/repository/s.json
@@ -2534,7 +2534,7 @@
 		},
 		{
 			"name": "Smart Title Case",
-			"details": "https://github.com/mattstevens/sublime-titlecase",
+			"details": "https://github.com/braver/SmartTitleCase",
 			"releases": [
 				{
 					"sublime_text": "*",


braver:
@mattstevens You're welcome to object to this take-over of your package. Your package does not have a license file, but since it's a pretty thin wrapper around ppannuto/python-titlecase, which is MIT-licensed, my assumption is that it's could be considered to be similarly open.

mattstevens:
I've merged the mentioned PR, I don't think a takeover is necessary.

braver:
Indeed, let's continue with the existing package!

braver:
While my contributions were merged they haven't been tagged and are therefore still, after repeated nudging, unavailable to users. This is a trivial package and should be trivial to maintain. I intend to do so a bit more proactively than has currently been the case.

mattstevens:
I will make some time for it next week. Since it supports older versions of ST I wanted to verify it still works with those.

braver:
The API's the package itself uses (there are barely any) haven't changed in forever. That should work all the way back to SublimeText 2, apart from the fact you can't run ST2 on a modern operating system of course 😉. The only touchpoint is [the library](https://github.com/ppannuto/python-titlecase/blob/main/titlecase/__init__.py) and whether it would run on python 3.3 (because it does not itself interface with any ST api's) and there is no indication at all that it won't. 

Really, this is not a concern at all. My contributions don't change that: that package didn't change, only the library did. In fact it's the other way around: without them it stops working on upcoming versions of ST. 

This is what copilot says about that topic:

> Short answer: Yes — the script is compatible with Python 3.3.
> 
> What I checked and why it’s OK
> 
> Syntax: no Python‑3.6+ features (no f-strings, no variable annotations, no async/await syntax that would be incompatible with 3.3).
> Unicode source: the file declares # -- coding: utf-8 -- so the fancy quotation marks in constants are allowed in Python 3.3.
> Exception/try/except/else, function defs, class definitions, subclassing bytes/str — all supported in 3.3.
> Standard-library use (argparse, logging, os, string, sys, re) is fine in 3.3.
> map() is used and joined immediately ("/".join(map(...))) — in Py3 map returns an iterator, which join accepts; this works in 3.3.
> The code falls back to the stdlib re (import re as regex) when the third‑party regex module is not available, and alternative regex patterns are provided in that case.

braver:
There is also the concern of ongoing support of the package. We ask authors of package that they at least somewhat actively maintain their package. Your repository has open issues and pull requests going back to 2018, 2015 even. These could have easily been addressed by updating the library at any point in the past decade. Even if now you do update the package that doesn't bode well for the future. It's not something that needs attention all the time, but ST moves forward, Python moves forward, you can't let things sit for a decade plus. Let me help you out with this one. I maintain [several packages](https://packages.sublimetext.com/?q=author%3Abraver) and use ST daily both professionally and for various hobby projects. I know what I'm doing and don't needlessly break things for users.


Edit: All that to say this isn't about taking something from you in any way, I don't want credit for anything, I just want for good packages to exist for Sublime Text, for people to be able to contribute to them, and prefer improving existing packages over duplicating them and thereby making it harder for users to find the good package. It doesn't even need to be via this fork, you can just let me work on your repo, make small updates, respond to issues and pr's, etc.

mattstevens:
I'd like to actually try it just to double check. Once set up to do that again it will be simple to verify any future PRs.

mattstevens:
Tested and new release tagged, I don't think there's any need for a takeover.

braver:
Thanks, all good!

PR #9416, opened 2026-05-11
Update package reviewer action

kaste:
Point Package Control Channel review workflows at the latest st_package_reviewer commit.

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Diff and review changed/added packages
-        uses: kaste/st_package_reviewer/gh_action@81e964270472dae33e904a15a354976b99d7b2f2
+        uses: kaste/st_package_reviewer/gh_action@ad89510c574e97d92cfd4d9e0f5c9029394f5c94
         with:
           pr: ${{ github.event.pull_request.html_url }}
           file: repository.json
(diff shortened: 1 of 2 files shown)


braver:
Scrolled through your changelog for this, awesome stuff. Really appreciated for that. And of course for the help in doing the code reviews, it really helps a lot!

kaste:
Let's see if we get a better signal to noise, slowly but steadily.  👨‍⚕️

kaste:
Would be good if the reviewer would run immediately.  Or can I start these manually?  Maybe there is a triage role on GitHub to allow that.

braver:
> Would be good if the reviewer would run immediately.

That the CI actions no longer need approval you mean? That's a setting on the repository that @BenjaminSchaaf should be able to change. The approval is indeed not necessary at all, just that GitHub introduced it as the default behavior and Will was never around to do anything about it.

braver:
FYI I've rolled back to before this PR. e4e205ded09a710aab7d35bf603b3f10bfd763eb which we updated to #9420 is not (or no longer) an actual version of the action ([error](https://github.com/sublimehq/package_control_channel/actions/runs/26184518171/job/77036263826)). The tip of main of the action (573b332e25687618809f6ed33c40d7f0949b1c9c) also seems to [fail](https://github.com/sublimehq/package_control_channel/actions/runs/26184037777/job/77034538487).

kaste:
I never pushed e4e205.  Let's try #9426

PR #9420, opened 2026-05-17
Update reviewer action ref

kaste:
Point package review workflows at the latest st_package_reviewer commit, which fixes the composite action project context used by uv.


kaste:
🚒

braver:
🚀

PR #9421, opened 2026-05-18
Add Atlas Ragnarok color scheme

AyoubTadlaoui:

  [2]: https://semver.org

  My package is a dark color scheme for Sublime Tex. It is similar to other dark color schemes, but should still be added because it offers a distinct thunder-blue/crimson palette not currently in the channel.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -2152,6 +2152,17 @@
 			]
 		},
 		{
+			"name": "Atlas Ragnarok Color Scheme",
+			"details": "https://github.com/AyoubTadlaoui/atlas-ragnarok",
+			"labels": ["color scheme"],
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"tags": "st-"
+				}
+			]
+		},
+		{
 			"name": "Atomic Color Scheme",
 			"details": "https://github.com/gerardbm/Sublime-Atomic-Scheme",
 			"labels": ["color scheme"],


braver:
Please tell your agents to follow the PR template, see also https://github.com/sublimehq/package_control_channel/blob/master/AGENTS.md. Or create the PR yourself...

AyoubTadlaoui:
@braver  
 
 I redid the description properly against your template.

I added a .gitattributes and recut the tag as st-1.0.1 so the archive Package Control pulls is now just editors/sublime/ plus the readme and license, checked the tarball to make sure...

 One thing unrelated to my PR: the package check seems to be failing on every open PR right now with ModuleNotFoundError: No module named  'st_package_reviewer'. Looks like a CI env issue on your side rather than the submissions, figured I'd flag it in case it's you are not aware of.

  Let me know if there's anything else you want changed, i will proceed to fix it.

braver:
>  Looks like a CI env issue on your side rather than the submissions, figured I'd flag it in case it's you are not aware of.

Indeed, it's the test itself that was broken. There is fix for it now, so if you rebase you should get the latest and it should run properly.

repository/a.json
```diff
@@ -2151,6 +2151,17 @@
 				}
 			]
 		},
+		{
+			"name": "Atlas Ragnarok",
```
braver: It's not required but I would recommend putting "Color Scheme" in the name (like [many](https://packages.sublimetext.io/?q=label%3A%22color+scheme%22) packages do), because if a user sees the package name without the label "Atlas Ragnarok" isn't immediately recognizable as to what the package does.

braver:
Oh, never mind the tests then 😅 Sorry, was hoping the latest version was actually stable. I'm working on that though. 

Had a look at your package, it's A-OK 👌🏻 So, this PR is approved, I'll get the tests to run and pass and merge all of it.

braver (approved):

AyoubTadlaoui:
@braver No rush on the tests, really appreciate you reviewing it manually, thanks you!

braver:
merged in 94785f0e5a12a1f83c56f3ea82f230ea87338d60

PR #9426, opened 2026-05-20
Bump reviewer action ref

kaste:
Point the package review workflows at the reviewer commit that sets a setuptools-scm fallback version before uv builds the action checkout.

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Diff and review changed/added packages
-        uses: kaste/st_package_reviewer/gh_action@81e964270472dae33e904a15a354976b99d7b2f2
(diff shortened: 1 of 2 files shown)


braver:
We could also go with @v1 now right?

This time around I'll make a branch to run some actual package/pr tests before merging :D

kaste:
Yes you can do the switch to v1.  I tried a canary/smoke test yesterday using my forks.  That worked but is also borderline confusing as I made fake PRs and then you juggle with 6 browser tabs, get 3 reds and 5 greens.  

Maybe I can automate that and tell an agent to run that and read the output.

kaste:
v1 is already on the next commit / v.1.1.1 michaelblyons reported an error

braver:
Looking good! Handled via 42c2b1cdd610408041ea45adc14eb08d9cf157ee

PR #9427, opened 2026-05-20
Add note on using formatter in tools directory

kaste:
Added a note about using the formatter in tools directory.


braver:
Nice one!

PR #9432, opened 2026-05-22
add ClickableLinks

pakhromov:

[2]: https://semver.org

My package is similar to Clickable URLs. However it should still be added because Clickable URLs requires a keyboard shortcut (Ctrl+Alt+Enter), you can create a mousemap to use it on mouse click but that still requires holding some modifier key. This plugin opens links with just a mouse click.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1479,6 +1479,16 @@
 			]
 		},
 		{
+			"name": "ClickableLinks",
+			"details": "https://github.com/pakhromov/ClickableLinks",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Clickable WSDL",
 			"details": "https://github.com/lwilli/ClickableWSDL",
 			"releases": [
--- a/repository/c.json
+++ b/repository/c.json
@@ -1470,17 +1470,7 @@
 		},
 		{
 			"name": "Clickable URLs",
-			"details": "https://github.com/leonid-shevtsov/ClickableUrls_SublimeText",
-			"releases": [
-				{
-					"sublime_text": "*",
-					"tags": true
-				}
-			]
-		},
-		{
-			"name": "ClickableLinks",
-			"details": "https://github.com/pakhromov/ClickableLinks",
+			"details": "https://github.com/pakhromov/ClickableUrls",
 			"releases": [
 				{
 					"sublime_text": "*",


braver:
Clickable URLs has been archived. It would be nice to replace that with a maintained version of the same package. Especially if that has new features. It sounds like your package is effectively a drop-in replacement? Can the one-click opening of links be disabled? That way existing users can continue to use it the way they're used to.

pakhromov:
> Clickable URLs has been archived. It would be nice to replace that with a maintained version of the same package. Especially if that has new features. It sounds like your package is effectively a drop-in replacement? Can the one-click opening of links be disabled? That way existing users can continue to use it the way they're used to.

Yes, I think I can do that, can also add some other options, like customizing the underline style and color. Should I try contacting the Clickable URLs author to transfer the repo or are you OK with a fork?

braver:
> Should I try contacting the Clickable URLs author to transfer the repo or are you OK with a fork?

The first would be more elegant. The second is also OK. In either case we should reach out to them and give the original author an opportunity to object (for a week or two). So maybe try the transfer first and if that goes nowhere we can go with the fork.

pakhromov:
> > Should I try contacting the Clickable URLs author to transfer the repo or are you OK with a fork?
> 
> The first would be more elegant. The second is also OK. In either case we should reach out to them and give the original author an opportunity to object (for a week or two). So maybe try the transfer first and if that goes nowhere we can go with the fork.

Contacted the original author a week ago via email (original repo is archived so could not open an issue there), but did not get a response unfortunately. So I created a [fork](https://github.com/pakhromov/ClickableUrls) with a few added settings, while keeping the default behavior unchanged. What is the best way forward, do I close this PR and open a new one replacing the original link with my fork?

braver:
> original repo is archived so could not open an issue there

Oh indeed, in that case we can als ignore the normal 2 week grace period and just go ahead with the replacement.

I would prefer if you just change this PR to update the existing entry to point at your fork. That's easier to keep track of :)

pakhromov:
> This PR changes Clickable URLs.
> ## Review for Clickable URLs master-230ad3b-2026.06.08.00.32.29
> 
>     * Repository is at https://github.com/pakhromov/ClickableUrls
> 
>     * Common used command prefix is: open.
> 
>     * Tip of master is tagged with 2.0.1. ✅
> 
> 
> 4 failures:
> 
>     * The binding ['ctrl+alt+enter'] has no context. Packages should only ship key bindings that use a specific context such as 'selector', 'setting.*', or a custom context key. If the binding defines a main entry-point to your package, move it to an example keymap instead so users can decide on their own.
>       File: Default (Windows).sublime-keymap
> 
>     * The binding ['super+alt+enter'] has no context. Packages should only ship key bindings that use a specific context such as 'selector', 'setting.*', or a custom context key. If the binding defines a main entry-point to your package, move it to an example keymap instead so users can decide on their own.
>       File: Default (OSX).sublime-keymap
> 
>     * The binding ['ctrl+alt+enter'] has no context. Packages should only ship key bindings that use a specific context such as 'selector', 'setting.*', or a custom context key. If the binding defines a main entry-point to your package, move it to an example keymap instead so users can decide on their own.
>       File: Default (Linux).sublime-keymap
> 
>     * 'Main.sublime-menu' has a 'Key Bindings' entry whose 'args.base_file' does not exist: ${packages}/Clickable URLs/Default ($platform).sublime-keymap
> 
> 
> 1 warning:
> 
>     * You define 'ClickableUrls.sublime-settings' but 'ClickableUrls' is neither your package name nor the name of a syntax you ship. Did you mean 'Clickable URLs.sublime-settings'?
>       File: ClickableUrls.sublime-settings

I left the keybind as it was in the original plugin, and since I am taking over the existing package, I thought that removing it (moving it to an example keymap) would break the behavior for people already using it.
In regards to the warning, the original plugin uses SETTINGS_FILENAME = 'ClickableUrls.sublime-settings' (no space) and I also did not change it.
Should I keep it as is for existing users or follow the new requirements?

braver:
> Should I keep it as is for existing users or follow the new requirements?

No you're right, in this case backwards compatibility for existing users is more important than adhering to the guidelines. It's neat if you can find a way to change the package to do both, but it's not essential to this PR.

PR #9433, opened 2026-05-22
Add ConfigHighlight

pakhromov:

[2]: https://semver.org

My package provides highlighting for generic configuration files like .conf or .config, etc. There are no packages like it in Package Control. The closest I could find is Splunk Conf File Syntax Highlighting but that one works only with specific .conf files, my plugin tries to be more general.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3493,6 +3493,16 @@
 			]
 		},
 		{
+			"name": "ConfigHighlight",
+			"details": "https://github.com/pakhromov/ConfigHighlight",
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Confluence",
 			"details": "https://github.com/mlf4aiur/SublimeConfluence",
 			"author": "Kevin Li",


braver:
> There are no packages like it in Package Control

There are a few more actually:
- https://packages.sublimetext.io/packages/Generic%20Config
- https://packages.sublimetext.io/packages/ApacheConf
- the one for EditorConfig also highlights those files: https://packages.sublimetext.io/packages/EditorConfig

pakhromov:
> > There are no packages like it in Package Control
> 
> There are a few more actually:
> 
>     * https://packages.sublimetext.io/packages/Generic%20Config
> 
>     * https://packages.sublimetext.io/packages/ApacheConf
> 
>     * the one for EditorConfig also highlights those files: https://packages.sublimetext.io/packages/EditorConfig

Yes, Generic Config is very similar actually, idk how I did not see it, thanks for pointing out and sorry for the trouble.

braver:
A completely new approach that sets your package apart from what's there would be accepted, but I think what's there already does a good job and you could perhaps contribute to it :)

PR #9437, opened 2026-05-26
Added a new package for the BoxLang Language

lmajano:
- [√ ] I'm the package's author and/or maintainer.
- [ √] I have read [the docs][1].
- [ √] I have tagged a release with a [semver][2] version number.
- [ √] My package repo has a description and a README describing what it's for and how to use it.
- [ √] My package doesn't add context menu entries. *
- [√ ] My package doesn't add key bindings. **
- [ √] Any commands are available via the command palette.
- [ √] Preferences and keybindings (if any) are listed in the menu and the command palette, and open in split view.
- [ √] If my package is a syntax it doesn't also add a color scheme. ***
- [ √] I use [.gitattributes][3] to exclude files from the package: images, test files, sublime-project/workspace.

My package is a comprehensive package for support for the BoxLang JVM Language. 

There are no packages like it in Package Control. We are the authors of the language itself.

### DIFF
--- a/repository/b.json
+++ b/repository/b.json
@@ -1415,7 +1415,7 @@
         {
 			"name": "BoxLang",
 			"details": "https://github.com/ortus-boxlang/sublimetext-boxlang",
-            "labels": [ "boxlang", "cfml", "server", "jvm", "java", "serverless", "desktop", "spring-boot" ],
+            "labels": [ "boxlang", "cfml", "server", "jvm", "matchbox", "miniserver", "java", "serverless", "desktop", "spring-boot" ],
 			"releases": [
 				{
 					"sublime_text": ">=3000",
--- a/repository/b.json
+++ b/repository/b.json
@@ -1415,7 +1415,7 @@
         {
 			"name": "BoxLang",
 			"details": "https://github.com/ortus-boxlang/sublimetext-boxlang",
-            "labels": ["boxlang","cfml","server","jvm","java","serverless","desktop"],
+            "labels": [ "boxlang", "cfml", "server", "jvm", "java", "serverless", "desktop", "spring-boot" ],
 			"releases": [
 				{
 					"sublime_text": ">=3000",
--- a/repository/b.json
+++ b/repository/b.json
@@ -1412,6 +1412,17 @@
 				}
 			]
 		},
+        {
+			"name": "BoxLang",
+			"details": "https://github.com/ortus-boxlang/sublimetext-boxlang",
+            "labels": ["boxlang","cfml","server","jvm","java","serverless","desktop"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
 		{
 			"name": "BracketFlasher",
 			"details": "https://github.com/jaytiar/SublimeBracketFlasher",
--- a/repository/b.json
+++ b/repository/b.json
@@ -1415,7 +1415,7 @@
         {
 			"name": "BoxLang",
 			"details": "https://github.com/ortus-boxlang/sublimetext-boxlang",
-            "labels": [ "boxlang", "cfml", "server", "jvm", "matchbox", "miniserver", "java", "serverless", "desktop", "spring-boot" ],
+            "labels": [ "boxlang", "cfml", "jvm", "formatting", "language syntax", "java", "build system", "completions", "auto-complete", "utilities" ],
 			"releases": [
 				{
 					"sublime_text": ">=3000",
--- a/repository/b.json
+++ b/repository/b.json
@@ -1415,7 +1415,7 @@
         {
 			"name": "BoxLang",
 			"details": "https://github.com/ortus-boxlang/sublimetext-boxlang",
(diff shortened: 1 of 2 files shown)


lmajano:
All issues have been addressed

lmajano:
this is using the old tag of 1.0.2, 1.0.3 is the latest and everything is now fixed

repository/b.json
```diff
@@ -1412,6 +1412,17 @@
 				}
 			]
 		},
+        {
+			"name": "BoxLang",
+			"details": "https://github.com/ortus-boxlang/sublimetext-boxlang",
+            "labels": [ "boxlang", "cfml", "jvm", "formatting", "language syntax", "java", "build system", "completions", "auto-complete", "utilities" ],
```
braver: You've got mixed indentation in your entry, I believe mostly on this line.

braver:
Just some superficial stuff till I get around to a more thorough review:
- It would probably be a good idea to opt-in to the newer python runtime: [example](https://github.com/braver/ColorConverter/blob/main/.python-version).
- You can use `.gitattributes` to exclude tests, agent specs, etc from the package as it's delivered to users.
- Note that messages are served as plain text. Raw markdown is pretty readable, but you have a lot of markup in there that would be better if simplified.
  - Related: users will see this only once and probably won't read all of it (especially if it's this much). Consider that users already installed the package so you don't have to sell it on them here. So focus on "getting started" stuff and then link to the README for more details.

lmajano:
Here is an update

✅ - fixed mixed spacing on package control
✅ -  It would probably be a good idea to opt-in to the newer python runtime
✅ - You can use `.gitattributes` to exclude tests, agent specs, etc from the package as it's delivered to users.
✅ - Note that messages are served as plain text. Raw markdown is pretty readable, but you have a lot of markup in there that would be better if simplified.
✅ - Related: users will see this only once and probably won't read all of it (especially if it's this much). Consider that users already installed the package so you don't have to sell it on them here. So focus on "getting started" stuff and then link to the README for more details.

Done!
Thanks for the feedback and the great work that you do!

repository/b.json
```diff
@@ -1412,6 +1412,17 @@
 				}
 			]
 		},
+        {
+			"name": "BoxLang",
+			"details": "https://github.com/ortus-boxlang/sublimetext-boxlang",
+            "labels": ["boxlang","cfml","jvm","formatting","language syntax","java","build system","completions","auto-complete","utilities"],
```
braver:
```suggestion
			"labels": ["boxlang","cfml","jvm","formatting","language syntax","java","build system","completions","auto-complete","utilities"],
```

Some mixed indentation here still.

braver:
How are these settings used? I don't find any references in the codebase, but was wondering if scopes or colorish colors could be used instead of hardcoding values that might not work in the user's color scheme.

```
boxlang_tag_style
boxlang_tag_attribute_style
```

braver:
Some more questions:

What's the reason for the LEGACY_SETTINGS_FILE? This package hasn't been published yet, at least not on package control. Or is that just the remains from a refactor?

The way you're shipping the file icons basically means you're forcing users to use A File Icon as well, otherwise the BoxLang files will not have any icon at all. I would recommend not touching the icons, that's really the responsibility of the various Theme packages, and packages like A File Icon or FileIcons (ie. open PR's on those kinds of packages to have them support your file types). 

Finally, more out of curiosity that anything else, I wonder if what you're providing here wouldn't have made more sense as an LSP?

lmajano:
✅ Removed unused color styles, left over from initial build.
boxlang_tag_style
boxlang_tag_attribute_style

✅ This was left from the refactor
> What's the reason for the LEGACY_SETTINGS_FILE? This package hasn't been published yet, at least not on package control. Or is that just the remains from a refactor?

✅ Agreed. I guess since this is realtively a young language, we wanted to offer something.
> The way you're shipping the file icons basically means you're forcing users to use A File Icon as well, otherwise the BoxLang files will not have any icon at all. I would recommend not touching the icons, that's really the responsibility of the various Theme packages, and packages like A File Icon or FileIcons (ie. open PR's on those kinds of packages to have them support your file types).

>Finally, more out of curiosity that anything else, I wonder if what you're providing here wouldn't have made more sense as an LSP?

Yes, we are trying to see how to package our LSP outside of our VSCode extension.  So this will eventually come instead of our AST parser.

braver:
I think we're mostly looking good here. Please address the whitespace issues in the b.sjon and revert the change to c.json.

lmajano:
done

braver:
You have [base file](https://github.com/ortus-boxlang/sublimetext-boxlang/blob/main/Default.sublime-commands#L14C13-L14C93) set to `"base_file": "${packages}/BoxLang/inputmaps/Default ($platform).sublime-keymap"`, which does not exist (the actual file [is here](https://github.com/ortus-boxlang/sublimetext-boxlang/blob/main/inputmaps/Default.sublime-keymap)).

braver:
@lmajano please have a look at the last bit of feedback.

lmajano:
Will look into this

lmajano:
Fixed now and pushed

braver:
@lmajano there is still a problem there ☝🏻

lmajano:
Done

PR #9442, opened 2026-06-02
Add MarkdownImagePaste

vjeantet:
- [x] My package doesn't add key bindings. ** (my package overide Cmd+V and it is its purpose)

With my package, in a markdown view, Cmd+V saves a clipboard image next to the current .md file and inserts a `` reference; otherwise it falls back to a normal paste.

My package is similar to [Image​Paste](https://packagecontrol.io/packages/ImagePaste) However it should still be added because it works on ST4, macOS and will work on LINUX and WIN.

- Repository: https://github.com/vjeantet/sublime-image-paste
- One package, repository root.
- Semver tag `1.0.0` is published (`tags: true`).
- Sublime Text 4 only (`>=4000`); the package runs under the Python 3.8 plugin host.
- macOS only (`platforms: ["osx"]`): it ships a universal arm64+x86_64 helper binary that reads images from NSPasteboard. A `.no-sublime-package` file forces on-disk extraction so the helper is runnable.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -685,6 +685,18 @@
 			]
 		},
 		{
+			"name": "MarkdownImagePaste",
+			"details": "https://github.com/vjeantet/sublime-image-paste",
+			"labels": ["markdown", "image", "clipboard", "paste"],
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"platforms": ["osx"],
+					"tags": true
+				}
+			]
+		},
+		{
 			"details": "https://github.com/sekogan/MarkdownLight",
 			"labels": ["language syntax", "markdown"],
 			"releases": [


kaste:
Please note ImagePaste's running fork at https://github.com/kaste/imagepaste  

Maybe you should or could translate the messages in the code to English.  Over there we just use the Python package/dependency *pillow*.  Isn't that easier than writing a go extension?

vjeantet:
Thanks for pointing this out, @kaste. I genuinely was not aware of your fork when I started this project. I did look for active alternatives before going ahead, but your repo doesn’t surface easily (no description, no topics, 0 stars at the time). That's a bit frustrating in retrospect, but fair enough.

On the Go helper vs. Pillow question: you're right that it's an odd choice for a Sublime Text plugin. I went with Go because I am more comfortable with it and it gave me a self-contained binary for pasteboaed access on all os. But using Pillow as a dependency through Package Control is clearly the more idiomatic approach : no binary to ship, no .no-sublime-package constraint, and cross-platform for free.

I'm going to take a proper look at your code. Depending on how much overlap there is, it might make more sense to contribute there than to maintain a separate package. I'll report back.

kaste:
>  I genuinely was not aware of your fork

Of course, it is just a fork so nobody knows.  I *think* my package is pretty good at least on Windows because Pillow calls into Windows-API's directly.  Yours does one or two popen spawns per essentially `ctrl+v`.  But Pillow also does call out to sub processes on Mac and Linux.  That's why I always thought I should propose mine only for Windows.  (But the un-forked, golden version of ImagePaste does use Popen even more 🙄 so that wouldn't help old ImagePaste users at all.)

(I do wonder if compilation to a shared lib/DLL from Go is easy enough so that you could use that from Python directly.)

vjeantet:
Hi @kaste, and thanks again for the thoughtful exchange.

After reviewing your fork, I'm going to **close this PR and step back in favor of your `imagepaste`**. It already does everything mine does and more: cross-platform support, Pillow as a managed Package Control dependency (no binary to ship and sign myself), reStructuredText support, filename confirmation, and a project-relative target folder. Shipping a macOS-only package that bundles a Go binary would just fragment a problem you've already solved more cleanly.

One small nuance for the record, since it might be useful to you: on macOS my Go helper reads `NSPasteboard` directly, with no subprocess spawn, whereas Pillow's `ImageGrab.grabclipboard()` shells out on macOS/Linux. It's a marginal difference and almost certainly imperceptible on a `Cmd+V`, so I don't think it justifies maintaining a separate binary, but if you ever want a no-subprocess macOS fast path, the approach is there in this repo.

Closing this. Thanks for maintaining `imagepaste` - I'll point people your way.

kaste:
Thanks for the kind words.  Then I should polish and suggest it for this official registry perhaps, and for all platforms.  

For Mac, we -- in theory, but I'm not a Mac developer -- could add a pyobjc variant as dependency 

```
$ uv run -m scripts.crawl_libraries --try "base: pypi:pyobjc-framework-Cocoa;"
pyobjc-framework-Cocoa release matrix; -v to see the raw JSON output
Source: pypi:cache
Latest version: 12.2

              py38  py314
-----------+-------------
osx-x64       A     B
osx-arm64     A     B

A = 10.3.2
B = 12.2
``` 

and use that directly from Python.  (Basically what I would do and Pillow actually already does for Windows.)

PR #9443, opened 2026-06-03
Adding an up-to-date Cypher syntax package

Fulguritude:

[2]: https://semver.org

My package is [Cypher ST3+](https://github.com/Fulguritude/cypher-sublime-syntax). It is a syntax coloring package for the Neo4J Cypher query language.

My package is similar to [Cypher](https://github.com/kollhof/sublime-cypher). However it should still be added because the repo has been archived for almost a decade, and the plugin only works for ST versions `<3000`.

Concerning the name of the package, since mine is set to versions `>=3092`, there shouldn't be a conflict if it was just named `Cypher`, but I named it this way to avoid any conflict. Feel free to change it if so desired.

I'm available to discuss, improve or fix anything you'd like.

The syntax coloring in action:
image

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -5121,17 +5121,6 @@
 		},
 		{
 			"name": "Cypher",
-			"details": "https://github.com/kollhof/sublime-cypher",
-			"labels": ["language syntax"],
-			"releases": [
-				{
-					"sublime_text": "<3000",
-					"branch": "master"
-				}
-			]
-		},
-		{
-			"name": "Cypher ST3+",
 			"details": "https://github.com/fulguritude/cypher-sublime-syntax",
 			"labels": ["language syntax"],
 			"description": "Syntax highlighting for the Neo4j Cypher query language for ST3 and above.",
--- a/repository/c.json
+++ b/repository/c.json
@@ -5131,6 +5131,18 @@
 			]
 		},
 		{
+			"name": "Cypher ST3+",
+			"details": "https://github.com/fulguritude/cypher-sublime-syntax",
+			"labels": ["language syntax"],
+			"description": "Syntax highlighting for the Neo4j Cypher query language for ST3 and above.",
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Cython",
 			"details": "https://github.com/NotSqrt/sublime-cython",
 			"labels": ["language syntax"],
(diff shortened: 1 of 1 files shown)


braver:
Since the original package, we can update the existing entry to point to your repo instead.

Fulguritude:
Hello, thanks.
That sounds good to me, though it would break Cypher for ST2, wouldn't it ? Not sure how keen you are / your policy on maintaining very old versions. In any case, let me know if you want me to change the code PR to enable that switch of repo reference instead.

braver:
> though it would break Cypher for ST2, wouldn't it 

Not really, since current versions of package control don't work on ST2 and ST2 can't work with the current package index anyway. In other words, ST2 is ancient history :)

Fulguritude:
Duly noted. I merged upstream master and replaced the old package in `c.json` then.

Concerning "Consider adding an entry in 'Main.sublime-menu' and 'Default.sublime-commands' to help users find or customize them.": I took a look at https://docs.sublimetext.io/guide/customization/menus.html and https://docs.sublimetext.io/guide/extensibility/command_palette.html, but not I'm really sure how to test things or make a design that makes sense. If this is unnecessary, I'll ignore it for now.

braver:
An example for the command palette:
https://github.com/braver/ColorConverter/blob/main/Commands.sublime-commands#L2-L8

An example for the main menu:
https://github.com/braver/ColorConverter/blob/main/Main.sublime-menu#L13-L21

repository/c.json
```diff
@@ -5121,12 +5121,13 @@
 		},
 		{
 			"name": "Cypher",
-			"details": "https://github.com/kollhof/sublime-cypher",
+			"details": "https://github.com/fulguritude/cypher-sublime-syntax",
 			"labels": ["language syntax"],
+			"description": "Syntax highlighting for the Neo4j Cypher query language for ST3 and above.",
```
braver: Please put this on the GitHub repo instead. That way you can change it any time you like and it's consistent for users as well.

braver:
I'd like to get this merged, but please do address the feedback about settings. Adding menu entries for settings isn't rocket science, it's just a matter of adding the right JSON files and there are a lot of examples out there to copy from.
Alternatively remove the settings file. Syntax settings like indentation are typically personal preference more than a rule, so you shouldn't always enforce them on users even if they make a good default.

repository/c.json
Fulguritude: You mean just removing the "description" line so that it can infer it from my github repo's description? Or is there something I can/should do with the labels line as well?

braver: > You mean just removing the "description" line so that it can infer it from my github repo's description? 

Indeed, just that :)

Fulguritude:
I removed the settings file from my repo and the description line from c.json in my fork; also pulled the upstream (this) repo and merged.

PR #9444, opened 2026-06-04
Add ES|QL syntax highlighting package

bartoval:

My package is a TextMate grammar (.tmLanguage) that adds syntax highlighting for ES|QL, Elasticsearch Query Language — Elastic's query language for the Elasticsearch platform.

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -1335,6 +1335,17 @@
 			]
 		},
 		{
+			"name": "ES|QL",
+			"details": "https://github.com/elastic/esql-syntax",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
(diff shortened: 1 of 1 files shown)


repository/e.json
```diff
@@ -1334,6 +1334,17 @@
...
 			]
 		},
+		{
+			"name": "ESQL",
+			"details": "https://github.com/elastic/esql-syntax",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "main"
```
deathaxe: Please use tag based releases for package control to deploy from defined checkpoints, only.

```suggestion
					"tags": true
```

braver:
How have we gone all these years without this? 😄

PR #9447, opened 2026-06-10
Add sublime-mcp

dpc00:

[2]: https://semver.org

My package is similar to MCPHelper and MCPServer. However it should still be added because the direction is inverted: sublime-mcp turns Sublime Text into an MCP *server* that an external AI agent (e.g. Claude Code CLI) can directly control. The agent reads cursor context, makes edits, navigates files, runs builds, and evaluates Python in ST's runtime — while the user watches the changes appear in real time. The AI is the client; ST is the controlled environment. Neither MCPHelper nor MCPServer expose ST's internal API at this level of granularity to an external agent.

I built this to solve my own workflow and have been using it daily. I intend to maintain it.

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -1204,6 +1204,17 @@
 			]
 		},
 		{
+			"name": "MCP Commander",
+			"details": "https://github.com/dpc00/sublime-mcp",
+			"labels": ["ai", "mcp"],
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "MCPHelper",
 			"description": "Connects to a local Model Context Protocol (MCP) server for AI-powered code generation, review, refactoring, and translation using OpenAI or Gemini.",
 			"author": "David Donohue",
--- a/repository/s.json
+++ b/repository/s.json
@@ -4661,6 +4661,17 @@
 			]
 		},
 		{
+			"name": "sublime-mcp",
+			"details": "https://github.com/dpc00/sublime-mcp",
+			"labels": ["ai", "mcp"],
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"details": "https://github.com/ccreutzig/sublime-MuPAD",
 			"labels": ["language syntax"],
 			"releases": [
--- a/repository/s.json
+++ b/repository/s.json
@@ -4661,17 +4661,6 @@
 			]
 		},
 		{
-			"name": "sublime-mcp",
-			"details": "https://github.com/dpc00/sublime-mcp",
-			"labels": ["ai", "mcp"],
-			"releases": [
-				{
-					"sublime_text": ">=4000",
-					"tags": true
-				}
-			]
-		},
-		{
 			"details": "https://github.com/ccreutzig/sublime-MuPAD",
 			"labels": ["language syntax"],
 			"releases": [


kaste:
This needs some work.  It looks like it is only for Windows so define the "platforms" key in the entry here.

Overall the repo/package would be shipped as source only and *we* package it.  E.g. remove .egginfo, dist/, requirements.txt.  However, your requirements are not available in Sublime Text (maybe in your personal Sublime Text installation) but not generally.  The install.txt refers that users should run `pip install sublime-mcp` to get the package working; but that's like a loop-link as it refers to the same repo.  There also (stray?) .js files in the repo which I think require standard node/npm tooling. TBH this one looks like very unfinished business for other users.  It may perfectly work for you of course, but maybe it is tailored to your computer/programming environment.

Note we also have https://packages.sublimetext.io/packages/AI%20Bridge

dpc00:
Thanks for the detailed review — here's what's been addressed:

Platforms: The plugin now detects the OS at startup and listens on 127.0.0.1:9500 (Windows) or 127.0.0.1:9501
(Mac/Linux), overridable via SUBLIME_MCP_PORT. Works on all three platforms.

Stray JS files / Node.js dependency: The .gitattributes now marks index.js, package.json, and node_modules/ as
export-ignore, so Package Control never installs them. The Node.js server is an optional alternative for users who  ▎ prefer npm.

pip install sublime-mcp loop-link: Package Control installs the ST plugin; users then separately install the MCP
server via pip install sublime-mcp or npx sublime-mcp. The install.txt has been rewritten to make this clear.

dist/ and egg-info: Removed from git tracking entirely.

Python requirements in ST: mcp and httpx are only required by mcp_server.py (the external MCP server process). The  ▎ ST plugin itself has no external dependencies — stdlib only.

Happy to make any further changes you need.

kaste:
Ok, I see. Overall, that is confusing.  You basically have 3 projects in one repo, and hide the files then using .gitattributes.  Obviously, a reviewer/programmer cannot hold this state in their head.  

I also don't see a reason to have separate dependencies - be it via `pip` or `npx` - as you're already running inside a Python and can just implement the server and point your agent at it.

dpc00:
The reason for the separate server process is that Sublime Text 4 ships with Python 3.8, and the `mcp` package requires Python 3.10+. It is not possible to run the MCP server inside the ST plugin environment — the embedded Python is too old for the dependency. The external process (pip or npx) is the only viable approach given that constraint.

kaste:
Claude/Codex can write that; you don't need the dependency.  It isn't even a lot of code.

dpc00:
You're right — and it's done. v1.3.0 implements the MCP 2024-11-05 SSE transport directly in `sublime_mcp.py` using only Python 3.8 stdlib (no pip, no npx required). The server listens on port 9502 (Windows) / 9503 (Mac/Linux). Agents connect with a single config line:

\\\json
{ type: sse, url: http://127.0.0.1:9502/sse }
\\\

The plugin now has zero external dependencies. The Node.js server (`index.js`) and Python package (`mcp_server.py`) remain in the repo for users who prefer them, but they're excluded from the Package Control archive via `.gitattributes` and are not needed for the ST plugin to function.

kaste:
I said, it's not that much code.  IMO you should actually remove the other files; that's just confusing and you'll never touch them anyway.   🤷‍♀️ 

Update, the install.txt and README.  Also the docstrings in `sublime_mcp.py`.  Really, your agent should own these instructions and keep them up-to-date.  You confuse us both. 🤣 

You don't have any settings or commands; you just start the server on `load`.  People might hate that.  Tell them in the README.  Maybe just add two commands: "Start/Stop server".  Or a simple QuickPanel, or [ListInputHandler](https://www.sublimetext.com/docs/api_reference.html#sublime_plugin.ListInputHandler) thingy where you tell the people using the preview, if the server is running, and if so on which port; then to start it if stopped, and to stop it if already running...  A bit onboarding.  

You don't get a LGTM from me (because of the "looks-good") but I think it is merge-able.  👍

dpc00:
Thanks for the detailed feedback - all addressed in v1.3.1:

**Server Status command**
Added `MCP Commander: Server Status` to the Command Palette. Uses a ListInputHandler that shows whether the server is running and on which port, and lets the user start or stop it with a single pick.

**README and install.txt**
Rewritten to reflect the pure-Python SSE architecture. Old pip/npx steps removed. New install.txt is a single-step setup.

**Module docstring**
Updated in sublime_mcp.py to match the current architecture.

Tip of main is tagged v1.3.1.

kaste:
LGTM! @braver

PR #9450, opened 2026-06-13
Add Paste Markdown

Wuvist:
- [x] Preferences and keybindings are listed in the menu and the command palette, and open in split view.
- [x] This package is not a syntax and does not add a color scheme.
- [x] I use [.gitattributes](https://www.git-scm.com/docs/gitattributes#_export_ignore) to exclude test and repository-only files from the package.

Paste Markdown converts rich HTML copied from Chrome into Markdown when pasted into Sublime Text on macOS. It supports headings, links, images, lists, blockquotes, code blocks, and tables, with plain-text fallback.

There are no packages like it in Package Control. The existing Markdown to Clipboard package handles the opposite direction.

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -504,6 +504,18 @@
 			]
 		},
 		{
+			"name": "Paste Markdown",
+			"details": "https://github.com/Wuvist/subl_paste_md",
+			"labels": ["text manipulation", "clipboard", "paste", "markdown"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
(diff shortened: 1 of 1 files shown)


kaste:
That should work.  LGTM.  @braver 

@Wuvist You claim that you support ST v3 (`>=3000`).  That may be totally fine but I can't compute the API diff between 3 and current Sublime Text in my head.  I just trust you on this one.  👍

braver:
Nice!

PR #9451, opened 2026-06-16
Add CompareHistory entry to c.json

xalteropsx:
- [x] My package add context menu entries. *
- [x] My package add key bindings. **

[2]: https://semver.org

My package is ...

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -3304,14 +3304,14 @@
 			]
 		},
 		{
-		    "details": "https://github.com/xalteropsx/CompareHistory",
-		    "labels": ["diff", "compare", "comparison", "version-history", "side-by-side"],
-		    "releases": [
-		        {
-		            "sublime_text": ">=3143",
-		            "tags": true
-		        }
-		    ]
+			"details": "https://github.com/xalteropsx/CompareHistory",
+			"labels": ["diff", "compare", "comparison", "version-history", "side-by-side"],
+			"releases": [
+				{
+					"sublime_text": ">=3143",
+					"tags": true
+				}
+			]
 		},
 		{
 			"name": "Compass",
--- a/repository/c.json
+++ b/repository/c.json
@@ -3278,6 +3278,18 @@
 			]
 		},
 		{
+		    "name": "CompareHistory",
+		    "details": "https://github.com/xalteropsx/CompareHistory",
+		    "labels": ["diff", "compare", "comparison", "version-history", "side-by-side"],
+		    "releases": [
+		        {
+		            "sublime_text": ">=3143",
+		            "tags": true
+		        }
+		    ],
+		    "platforms": ["*"]
+		},
+		{
 			"name": "Compare Side-By-Side",
 			"details": "https://github.com/kaste/Compare-Side-By-Side",
 			"labels": ["diff", "compare", "comparison"],
--- a/repository/c.json
+++ b/repository/c.json
@@ -3304,7 +3304,6 @@
 			]
 		},
 		{
-		    "name": "CompareHistory",
 		    "details": "https://github.com/xalteropsx/CompareHistory",
 		    "labels": ["diff", "compare", "comparison", "version-history", "side-by-side"],
 		    "releases": [
--- a/repository/c.json
+++ b/repository/c.json
@@ -3278,18 +3278,6 @@
 			]
 		},
 		{
-		    "name": "CompareHistory",
-		    "details": "https://github.com/xalteropsx/CompareHistory",
-		    "labels": ["diff", "compare", "comparison", "version-history", "side-by-side"],
-		    "releases": [
-		        {
-		            "sublime_text": ">=3143",
-		            "tags": true
-		        }
-		    ],
-		    "platforms": ["*"]
-		},
-		{
 			"name": "Compare Side-By-Side",
 			"details": "https://github.com/kaste/Compare-Side-By-Side",
 			"labels": ["diff", "compare", "comparison"],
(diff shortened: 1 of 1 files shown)


kaste:
Actually `sbs_*` is a prefix of one of my plugins, and `sbs_compare` actually *is* the main command from that package.  :grin:

There are enough prefixes for all of us.  

Btw there is no "caption" for key bindings.

xalteropsx:
> Actually `sbs_*` is a prefix of one of my plugins, and `sbs_compare` actually _is_ the main command from that package. 😁
> 
> There are enough prefixes for all of us.
> 
> Btw there is no "caption" for key bindings.

many thanks for your plugin i will mention your plugin soon in description yah it gave me alot of inspiration my apologies i dont know how i forget such important things

xalteropsx:
ah sorry i didnt notice 3 failures rather i just fix warning brb

xalteropsx:
@kaste done sorry for trouble havent really notice failure one instead fix warning one please let me know if their any other issue thank you >.</

kaste:
Take your time, dude.  You now have `sbs_comparex`, that's still "my" prefix.  The prefix is the `sbs_`.  Your package is CompareHistory, so you could use `ch_` or `cohi_` or `cohis_` ...  😀

xalteropsx:
> Take your time, dude. You now have `sbs_comparex`, that's still "my" prefix. The prefix is the `sbs_`. Your package is CompareHistory, so you could use `ch_` or `cohi_` or `cohis_` ... 😀

ah i though it messup so i endup using x  give me 30min after dinner i will check it bro

xalteropsx:
all done bro https://github.com/xalteropsx/CompareHistory/releases/tag/v1.0.0

xalteropsx:
sorry kaste brother i trouble you alot today i hope when i next submit another package i hope that time i wont do same mistake and save your time my humble apology give me 20-30min i will check

kaste:
Now you have no warnings and failures but also removed too much.  You likely want to add key bindings, just not global ones, but specific to *your* view.  Since you took Compare Side-by-side as the reference, you also have `is_ch_compare` set on such a view.  And you can use that for your bindings: https://github.com/kaste/Compare-Side-By-Side/blob/69ffa759632e083281944ed7e5b5ca5fa4e9c2d4/Default.sublime-keymap#L10-L10 e.g. your next/previous commands.  

Also your default settings file is empty, but in the README you list many settings.  And you also add a command to edit them.  The defaults go in the setting file.  Look how I just use 

```
def sbs_settings():
    return sublime.load_settings('SBSCompare.sublime-settings')
```

and then  `sbs_settings().get('hide_sidebar', False)` everywhere.  This is a bit old-school implementation-wise but works, and is better than having settings in the code (https://github.com/xalteropsx/CompareHistory/blob/main/CompareHistory.py#L15); except of course you want hidden constants/settings.

xalteropsx:
i left it intentionally since  not want them to modify unnecessary since in future update it to v1.0.1 so people can directly modify changes inside compare and effect the original file  but although can do alt n and alt p may be can set sbscompare sublime setting rather hardcoded may be you right  i will take a small break will check give me 30min

kaste:
I think that LGTM @braver

braver:
Some small stuff to wrap it up then:
- The entry is not sorted correctly in the c.json file.
- What are the "commented out by default" key bindings that the README speak about?
- Consider exposing the keybindings via the menu, and/or more completely documenting the keybindings. Your README has a "Key Bindings" section that doesn't mention the commands they bind, and "How to Enable CompareHistory Key Settings" that does but doesn't explain what any of the commands or arguments are for (ch_comparex? last_selections?). A well commented example bindings file is typically at least as useful or more so as a README.

xalteropsx:
give me 30minute i will check this readme seems not been updated comparex has been replace with compare

xalteropsx:
done recheck

repository/c.json
```diff
@@ -3303,6 +3303,18 @@
...
+		    "details": "https://github.com/xalteropsx/CompareHistory",
+		    "labels": ["diff", "compare", "comparison", "version-history", "side-by-side"],
+		    "releases": [
+		        {
+		            "sublime_text": ">=3143",
+		            "tags": true
+		        }
+		    ],
+		    "platforms": ["*"]
+		},
```
braver: If you want to specify platforms it goes into releases. But here we should remove it since it covers all/\*
```suggestion
		    ]
		},
```

repository/c.json
```diff
@@ -3303,6 +3303,18 @@
 				}
 			]
 		},
+		{
+		    "name": "CompareHistory",
+		    "details": "https://github.com/xalteropsx/CompareHistory",
```
braver: If your package name matches the repo, you don't even need to specify that.

```suggestion
		    "details": "https://github.com/xalteropsx/CompareHistory",
```

braver:
Why does `ch_versioned_hx_settings` exist? It looks like there are now 2 ways of opening settings, as a user I won't understand which is which. The README also seems to suggest that `Ctrl+Shift+,` opens the settings but by default that binding is not bound.

xalteropsx:
i will do it after 1hr break i think i will make a constant setting rather two separate setting

braver:
I'm just going to merge it now, but do please have a look at the duplicate entries for settings in the command palette, that's a bit confusing still.

xalteropsx:
i though awhile will do it on sunday release add new feature that allow to modify original file also put all setting to preference package rather separating that file path and limit

PR #9452, opened 2026-06-16
Add KeyBindingReport package

vwheeler63:
- [x] (n/a)  If my package is a syntax it doesn't also add a color scheme. ***

## KeyBindingReport

...is a Sublime Text Package that produces a wide variety of reports about the current state of Sublime Text key bindings on the system it is running on, and has the capacity to simulate running on other platforms for the sake of reporting Key Binding information for that platform.

### Comparison to Other Packages

There are no packages like it in Package Control.  2 built-in reports begin with the name "Key Binding Overrides", which sounds a lot like the "FindKeyConflicts" Package.  The difference between the two is:

- Only 2 of the built-in reports deal with "key conflicts".

- "FindKeyConflicts" only reports on when it finds the same keypresses used in 2 or more key bindings in all packages on a system, and importantly, *it has no consideration for the "context" entries of these key bindings* thus making it report key bindings that are normal to have many of (e.g. for `"`, `'`, `enter`, `tab` and other keys that mean different things at different times), and it tends to cause *real key conflicts* to hide among the large amount of data reported.

- The 2 "Key Binding Overrides" reports DO take the key-bindings' "context" entries into account, and thus are far more effective at isolating real binding overrides.

- KeyBindingReport has 57 other reports that have nothing to do with "key conflicts".

Other than the above, it is different than any other Package managed under PackageControl.
 
### Note:

**st_package_reviewer** produces a warning on this Package:

    It looks like you're using platform-dependent code ...

due to platform-specific code in `platform.py`.  Indeed, this module detects the platform it is running on, and normally reports about *that platform*.  However, it also provides facilities (optional argument for all reports) to "simulate" being on platforms other than the one it is running on.  In any case, the Package is indeed designed to run on all three platforms:  Windows, Linux and OSX.

### DIFF
--- a/repository/k.json
+++ b/repository/k.json
@@ -190,6 +190,17 @@
 			]
 		},
 		{
+			"details": "https://github.com/vwheeler63/KeyBindingReport",
+			"labels": ["key", "binding", "key binding", "utilities"],
+			"donate": "https://buymeacoffee.com/vwheeler63",
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "KeyboardNavigation",
 			"details": "https://github.com/robertcollier4/KeyboardNavigation",
 			"releases": [
--- a/repository/k.json
+++ b/repository/k.json
@@ -190,7 +190,7 @@
 			]
 		},
 		{
-			"details": "https://github.com/vwheeler63/KeyBindingReport",
+			"details" : "https://github.com/vwheeler63/KeyBindingReport",
 			"labels": ["key", "binding", "key binding", "utilities"],
 			"donate": "https://buymeacoffee.com/vwheeler63",
 			"releases": [
--- a/repository/k.json
+++ b/repository/k.json
@@ -195,7 +195,7 @@
 			"donate": "https://buymeacoffee.com/vwheeler63",
 			"releases": [
 				{
-					"sublime_text": ">=4171",
+					"sublime_text": ">=4000",
 					"tags": true
 				}
 			]
--- a/repository/k.json
+++ b/repository/k.json
@@ -190,7 +190,7 @@
 			]
 		},
 		{
-			"details" : "https://github.com/vwheeler63/KeyBindingReport",
+			"details": "https://github.com/vwheeler63/KeyBindingReport",
 			"labels": ["key", "binding", "key binding", "utilities"],
 			"donate": "https://buymeacoffee.com/vwheeler63",
 			"releases": [
--- a/repository/k.json
+++ b/repository/k.json
@@ -195,7 +195,7 @@
 			"donate": "https://buymeacoffee.com/vwheeler63",
 			"releases": [
 				{
-					"sublime_text": ">=4000",
+					"sublime_text": ">=4171",
 					"tags": true
 				}
 			]


vwheeler63:
I will work on moving my `reload()` function to a lower level....  (not in root).

kaste:
Hi!   About the reloader: https://github.com/kaste/KissReloader#add-a-reloader-to-your-package has some recipes.  These are easier to use as they only require code in the top level root module.

vwheeler63:
> Hi! About the reloader: https://github.com/kaste/KissReloader#add-a-reloader-to-your-package has some recipes. These are easier to use as they only require code in the top level root module.

Thank you, kind sir!!

vwheeler63:
Hi, @kaste !  So if I am understanding that correctly, Recipe #3 goes into the top-level plugin module and the bit at the bottom (`with InPlaceReloader(): ...`) encapsulates (and intercepts) the imports so that they become reloads if they've already been loaded once?

Does   https://github.com/kaste/KissReloader/blob/master/plugin.py   play a role in that?

Edit:

I have studied and worked with importing and the `importlib` module, but if there are 8 levels of understanding possible, I'm probably at level 4 or 5, and have never worked with "finders" vs "loaders" vs "importers" etc. at that level yet.

kaste:
All of this correct.  I made that a contextmanager because, beside the side effect of importing modules, it also installs import machinery.  And that needs to be uninstalled in a finally clause.  

All three recipes go into the root plugin.  Note that recipe 3 is new and fresh and a simplification (yes indeed!) of a battle tested hot-reloader.  Right now it is used here: https://github.com/timbrel/GitSavvy/blob/master/git_savvy.py#L62  

The plugin itself, KissReloader, is a small convenience for devs so we can quickly reload packages while working on them.  E.g. as a build system (`ctrl+b`)

```
{
	"build_systems":
	[
		{
			"name": "Reload GitSavvy",
			"target": "kiss_reloader_reload_current_package",
		},
``` 

But it also adds a command to the Command Palette.  Feedback welcome; not registered here as well.  https://packages.sublimetext.io/packages/AutomaticPackageReloader is *the" similar plugin, and its reloader is the previous GitSavvy reloader actually, extracted.  Now we're full circle. But it never had recipes.

vwheeler63:
Good afternoon, Herr Kaste!  ( @kaste )

Thank you for the orientation!!  Great tools!  Looks like it's time for me (in the near future) to dive deeper into (to better understand) the import machinery!  

It also helped me discover that `__package__` was deprecated, and so I switched all my Packages that have sub-packages to using `__spec__.parent`, as well as simplified code related to importing!

vwheeler63:
Hi, Herr Kaste!  ( @kaste )

Regarding [the 2 warnings above](https://github.com/sublimehq/package_control_channel/pull/9452#issuecomment-4742159095):

I thought the first one was handled in `st_package_reviewer` around January 2026?  The Package has:
```
./resources/commands/KeyBindingReport.sublime-commands
```

Regarding the 2nd warning:  `./resources/menus/Main.sublime-menu` does indeed have a `Settings` entry and it works, pointing to `${packages}/KeyBindingReport/resources/settings/KeyBindingReport.sublime-settings`.  (I modeled this directory structure similar to the `OverrideAudit` Package.)  It's at the bottom of the file.  The override file for local settings is `User/KeyBindingReport.sublime-settings`.  Is there is a requirement that the `.sublime-settings` file be in the root of the Package that I am unaware of?

kaste:
> ....hat I am unaware of?

Don't think so.  If the package doesn't work, we need to change the reviewer.  For now, I patched the reviewer to accept it.  (It just looked for files at the root for this rule.)

vwheeler63:
> > ....hat I am unaware of?
> > Don't think so.  If the package doesn't work, we need to change the reviewer.  For now, I patched the reviewer to accept it.  (It just looked for files at the root for this rule.)

Excellent.  Thank you!  🙂

PR #9453, opened 2026-06-17
feat: add harpoon package

huyhoang8398:

[2]: https://semver.org

A Sublime Text plugin for marking files and jumping back to them instantly.
Marks are scoped per project, stored inside your .sublime-project file, so each project keeps its own independent list.

Features
- Mark/unmark the current file with one hotkey
- Jump to any mark directly by slot number (1, 2, 3, 4...)
- Browse all marks in a quick panel
- Cycle forward/backward through marks
- Marks persist across restarts (saved in the .sublime-project file)
- Dead marks (deleted or moved files) are pruned automatically
- Remember the line 

**NOTE**: dependency: `AutoProject` plugin or save folder as project

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -300,6 +300,17 @@
 			]
 		},
 		{
+			"name": "Harpoon",
+			"details": "Faster Navigate inside Sublime Text with your favourite list of file",
+			"labels": ["file navigation", "utilities"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "master"
+				}
+			]
+		},
+		{
 			"name": "Harvey",
 			"details": "https://github.com/rayje/harvey-sublime",
 			"labels": ["language syntax"],
--- a/repository/h.json
+++ b/repository/h.json
@@ -301,7 +301,7 @@
 		},
 		{
 			"name": "Harpoon",
-			"details": "Faster Navigate inside Sublime Text with your favourite list of file",
+			"details": "https://github.com/huyhoang8398/harpoon",
 			"labels": ["file navigation", "utilities"],
 			"releases": [
 				{


repository/h.json
```diff
@@ -299,6 +299,17 @@
 				}
 			]
 		},
+		{
+			"name": "Harpoon",
+			"details": "Faster Navigate inside Sublime Text with your favourite list of file",
```
braver: Erm... this needs to be a URL...

huyhoang8398: oops hehe my bad, fixed in latest

huyhoang8398: hmm, it is weird that CI found no release

braver:
Well huh... not sure why the bot doesn't pick up on it now. Maybe try making a new PR. 

I did have a quick look at the package. It's unlikely to pass tests, or get accepted, with those keybindings in place. They're too common and they're global. I understand the desire to have them, but they should ship as examples the user can set up and customize themselves. And menu and command palette entries have to exist so the user can find and modify them.

I love the idea though! Would be neat if it could also remember the line number to make it even more useful.

huyhoang8398:
> Well huh... not sure why the bot doesn't pick up on it now. Maybe try making a new PR.
> 
> I did have a quick look at the package. It's unlikely to pass tests, or get accepted, with those keybindings in place. They're too common and they're global. I understand the desire to have them, but they should ship as examples the user can set up and customize themselves. And menu and command palette entries have to exist so the user can find and modify them.

thanks i will remove the default hotkey, 

> I love the idea though! Would be neat if it could also remember the line number to make it even more useful.

i could make them remember the line at the time you add the file to the list, or i can track and update the line every time you leave/close a harpooned file. :thinking: which case you think it might be better :thinking:

braver:
> i could make them remember the line at the time you add the file to the list, or i can track and update the line every time you leave/close a harpooned file.

Hmmm, I can see how that get's complicated quickly... I was personally thinking of using this to store reference material and documentation, ie. stuff that doesn't change all the time.

huyhoang8398:
> > i could make them remember the line at the time you add the file to the list, or i can track and update the line every time you leave/close a harpooned file.
> 
> Hmmm, I can see how that get's complicated quickly... I was personally thinking of using this to store reference material and documentation, ie. stuff that doesn't change all the time.

oh if you i need to tweak it a little bit, instead of using `.sublime-project` we have kind of "global" file to store the filepath so you can access the file at any project. but it is not what i meant to develop this plugin i think, it is more for people that want to have a list of file in a project that they usually open

PR #9454, opened 2026-06-18
add: Harpoon package

huyhoang8398:

[2]: https://semver.org

Marks are scoped per project, stored inside your .sublime-project file, so each project keeps its own independent list.

Features
- Mark/unmark the current file with one hotkey
- Jump to any mark directly by slot number (1, 2, 3, 4...)
- Browse all marks in a quick panel
- Cycle forward/backward through marks
> - Marks persist across restarts (saved in the .sublime-project file) ( deprecated )
- Now using window.settings() to store the marks, thanks to @FichteFoll 
- Dead marks (deleted or moved files) are pruned automatically
- Remember the line

**NICE TO HAVE**: dependency: `AutoProject` plugin or save folder as project

### DIFF
--- a/repository/h.json
+++ b/repository/h.json
@@ -306,7 +306,7 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"branch": "master"
+					"tags": true
 				}
 			]
 		},
--- a/repository/h.json
+++ b/repository/h.json
@@ -300,6 +300,17 @@
 			]
 		},
 		{
+			"name": "Harpoon",
+			"details": "https://github.com/huyhoang8398/Harpoon",
+			"labels": ["file navigation", "utilities"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "master"
+				}
+			]
+		},
+		{
 			"name": "Harvey",
 			"details": "https://github.com/rayje/harvey-sublime",
 			"labels": ["language syntax"],


huyhoang8398:
create duplicate MR because of https://github.com/sublimehq/package_control_channel/pull/9453#issuecomment-4732956690

huyhoang8398:
@braver look like the CI is broken :thinking:

repository/h.json
```diff
@@ -299,6 +299,17 @@
...
 			]
 		},
+		{
+			"name": "Harpoon",
+			"details": "https://github.com/huyhoang8398/Harpoon",
+			"labels": ["file navigation", "utilities"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "master"
```
FichteFoll:
```suggestion
					"tags": true
```

kaste:
Hej! `setting.is_widget` is not a context esp. `false` that would not be considered "global".  You basically just found a limitation or bug in the reviewer.  You also define platform specific and unspecific key binding files.  Just move all of this to an Example.sublime-keymap.

If you don't have any settings.  Why not drop the file and the menu entry?

huyhoang8398:
> Hej! `setting.is_widget` is not a context esp. `false` that would not be considered "global". You basically just found a limitation or bug in the reviewer. You also define platform specific and unspecific key binding files. Just move all of this to an Example.sublime-keymap.
> 
> If you don't have any settings. Why not drop the file and the menu entry?

Is there is a way to show user the Example.sublime-keymap inside ST or they have to see it in the repo :thinking: if not then i think i will keep the Default keymap but comment all the entry

kaste:
Example for a Default.sublime-command file

```
    {
        "caption": "Preferences: Paste Markdown Key Bindings",
        "command": "edit_settings",
        "args":
        {
            "base_file": "${packages}/Paste Markdown/Example.sublime-keymap",
            "user_file": "${packages}/User/Default (${platform}).sublime-keymap",
            "default": "[\n    $0\n]\n"
        }
    }
```

In a Main.sublime-menu file this becomes:

```
                            {
                                "command": "edit_settings",
                                "args":
                                {
                                    "base_file": "${packages}/Paste Markdown/Example.sublime-keymap",
                                    "user_file": "${packages}/User/Default (${platform}).sublime-keymap",
                                    "default":
                                    "[\n    $0\n]\n"
                                },
                                "caption": "Key Bindings"
                            }
```

huyhoang8398:
> Example for a Default.sublime-command file
> 
> ```
>     {
>         "caption": "Preferences: Paste Markdown Key Bindings",
>         "command": "edit_settings",
>         "args":
>         {
>             "base_file": "${packages}/Paste Markdown/Example.sublime-keymap",
>             "user_file": "${packages}/User/Default (${platform}).sublime-keymap",
>             "default": "[\n    $0\n]\n"
>         }
>     }
> ```
> 
> In a Main.sublime-menu file this becomes:
> 
> ```
>                             {
>                                 "command": "edit_settings",
>                                 "args":
>                                 {
>                                     "base_file": "${packages}/Paste Markdown/Example.sublime-keymap",
>                                     "user_file": "${packages}/User/Default (${platform}).sublime-keymap",
>                                     "default":
>                                     "[\n    $0\n]\n"
>                                 },
>                                 "caption": "Key Bindings"
>                             }
> ```

https://github.com/huyhoang8398/Harpoon/pull/3
thanks, it works well for me

huyhoang8398:
> Marks persist across restarts (saved in the .sublime-project file) ( deprecated )

Now using window.settings() to store the marks, thanks to @FichteFoll

kaste:
LGTM!   That's from the youtuber, right?

What is missing or I didn't see is, editing/changing the order of the harpoon-list.  In vim this is (IIRC) just a buffer where you swap lines.  

One typical approach is, e.g. `ctrl+1` selects list item 1, `ctrl+shift+1` sets the current file as list item 1.  But harpoon did not work like that, it just has add (which is also remove, hence a toggle (iirc)); that is maybe the UX advantage; maybe you can have that list (quick menu thing), but then bind e.g. `alt+up/down` plus some key binding on your home row (Lol) to swap or reprogram the order.

But just making it a buffer where I can swap lines ...  That would only be surprising when the buffer comes up the first time, but then swapping is muscle memory, whereby swapping functionality for a quick list is unusual in Sublime. ... Hm.

LGTM! @braver 

Ah, P.S., if you set `marks.append({"path": file_name, "row": row, "col": col})`, *maybe* also store/restore the viewport position ... but `show_at_center` is actually not bad and useful, makes it predictable where to look at.

huyhoang8398:
> LGTM! That's from the youtuber, right?
>
> Ah, P.S., if you set `marks.append({"path": file_name, "row": row, "col": col})`, _maybe_ also store/restore the viewport position ... but `show_at_center` is actually not bad and useful, makes it predictable where to look at.

yes the idea is similar to Harpoon Neovim
Thanks, i found `show_at_center` is more ease to the eye

> What is missing or I didn't see is, editing/changing the order of the harpoon-list. In vim this is (IIRC) just a buffer where you swap lines.
> 
> One typical approach is, e.g. `ctrl+1` selects list item 1, `ctrl+shift+1` sets the current file as list item 1. But harpoon did not work like that, it just has add (which is also remove, hence a toggle (iirc)); that is maybe the UX advantage; maybe you can have that list (quick menu thing), but then bind e.g. `alt+up/down` plus some key binding on your home row (Lol) to swap or reprogram the order.
> 
> But just making it a buffer where I can swap lines ... That would only be surprising when the buffer comes up the first time, but then swapping is muscle memory, whereby swapping functionality for a quick list is unusual in Sublime. ... Hm.
> 
> LGTM! @braver
> 

Yeah the harpoon list in Neovim is a buffer so you can easily interact with that list ( like edit/delete/paste ). I really like that since Im also Neovintageous user but we cannot have that in ST

But you are right, we should have a way to edit the list position,  my best idea is to have a hotkey, to open the list -> then you select which entry to move, then there will be a button to move up / down, i think it is better since if you use keybind, user will be more overhead to remember ( also i dont think quick panel support hotkey )

see: https://github.com/huyhoang8398/Harpoon/pull/5/changes

image

kaste:
> but we cannot have that in ST

Both solutions would work in ST.  You can just create a buffer/new (scratch) view, populate it with filenames, and then update the in-memory state e.g. `on_close`.  (This allows deletion, addition, changing, and reordering.) (And also maybe update the buffer on state change, in case users forgot to close the view.)

The quick panel also works.  You just need to tag the input box of it to bind it.

```
        show_quick_panel(
            ... )
        window.run_command("harpoon_grab_quick_panel_view")


class harpoon_grab_quick_panel_view(TextCommand):
    def run(self, edit):
        view = self.view
        if view.element() != "quick_panel:input":
            print("Didn't work; Damdn; Help; ...")
            return
        view.settings().set("harpoon_list", True)
```

huyhoang8398:
> You can just create a buffer/new (scratch) view, populate it with filenames, and then update the in-memory state e.g. `on_close`. (This allows deletion, addition, changing, and reordering.) (And also maybe update the buffer on state change, in case users forgot to close the view.)

oh i like the buffer idea, but another contributor already submit another way to handle the harpoon list which is interesting and similar to what you propose that assign by hotkey, if you want to try https://github.com/huyhoang8398/Harpoon/pull/6

kaste:
image

Obvious Claude slop without really telling the agent what environment we're running in. 

Generally, that doubles the key bindings ... since this is all about UX, you maybe should trust prime that he got it right.  You'd need to battle test these, not because of possible bugs but UX differences.

Do you know the exact behavior/rules of the original.  I need to look that up again.

huyhoang8398:
> image
> Obvious Claude slop without really telling the agent what environment we're running in.
> 
> Generally, that doubles the key bindings ... since this is all about UX, you maybe should trust prime that he got it right. You'd need to battle test these, not because of possible bugs but UX differences.
> 
> Do you know the exact behavior/rules of the original. I need to look that up again.

i didnt have time to test yet, i will take a look tonight. the idea is fine i do like it, but i also like the buffer idea, also my previous [PR](https://github.com/huyhoang8398/Harpoon/pull/5) also is fine in somecase, i will see

kaste:
If you add tests, no pytest, but https://packages.sublimetext.io/packages/UnitTesting.  Point your agent at https://packages.sublimetext.io/packages/UnitTesting#headless-container-runner

Sibling modules that import each other are not allowed in the root dir (but in sub dirs they were), because the root dir is special in Sublime Text.

stadja:
> image
> 
> Obvious Claude slop without really telling the agent what environment we're running in.
> 
> Generally, that doubles the key bindings ... since this is all about UX, you maybe should trust prime that he got it right. You'd need to battle test these, not because of possible bugs but UX differences.
> 
> Do you know the exact behavior/rules of the original. I need to look that up again.

Well, it is maybe Claude slope but the +730 are because :

1. Added [a dev plan](https://github.com/stadja/Harpoon/blob/452adf4da366cc3e3d691f565d55f1b76ad842f5/feat/rts_style_grouping.md) (feat/rts_style_grouping.md) explaining step by step what I did (in order to be able to explain exactly what I did and why I did it)
2. I then extracted mark manipulations function in order to be able to test them
3. Added a test framework because nothing was tested (and if there were tests, I am sorry, I missed that and didn't found them)
4. Tried to make the documentation a bit tighter/more focus
5. Added the new key shortcuts

It is a too big PR, and at first I had 9 commits in it. But I made a very bad git move in the end (I had a github authentification problem because while working on something else I had to change my SSH key and forgot about it). And in the end I had to squash everything because I just left with the code without any commit. Sorry about that. I understand if it is not okay.

But just read my dev plan (inside feat/ folder). I like what I did.

kaste:
That was easy to misread, sorry, I smiled about Claude which never hesitates to pull more stuff in (e.g. pytest where it should know that inside Sublime Text it should use the framework/runner I linked above), or multiple root plugins that import each other.  Claude could pause for a moment and imitate what the other packages do.  

```
Thinking:  Ok, I need to calm down a moment as user said that I'm not running on 
the CLI but inside Sublime Text which might be a special environment with its special 
restrictions and best-practices.  ... I know LSP, SublimeLinter and GitSavvy. Let's quickly 
look up how they approach testing to be clear...
```

:grin:

stadja:
> That was easy to misread, sorry, I smiled about Claude which never hesitates to pull more stuff in (e.g. pytest where it should know that inside Sublime Text it should use the framework/runner I linked above), or multiple root plugins that import each other. Claude could pause for a moment and imitate what the other packages do.
> 
> ```
> Thinking:  Ok, I need to calm down a moment as user said that I'm not running on 
> the CLI but inside Sublime Text which might be a special environment with its special 
> restrictions and best-practices.  ... I know LSP, SublimeLinter and GitSavvy. Let's quickly 
> look up how they approach testing to be clear...
> ```
> 
> 😁

It is my very first attempt at sublime text packaging and PRing (after +15 years of use !!!!)
I should have read a bit more before going on ... I even didn't knew that there was a specific test framework.

Right now I can't but I'll try to update my PR.

huyhoang8398:
@stadja  @kaste  what do you think about this, i find it more easier to edit the list ( maybe im bias because i use neovim with harpoon before ), 

> create a buffer/new (scratch) view, populate it with filenames, and then update the in-memory state e.g. on_close. (This allows deletion, addition, changing, and reordering.) (And also maybe update the buffer on state change, in case users forgot to close the view.)

was taken from this idea, and also very similar to Harpoon OG.

see https://github.com/huyhoang8398/Harpoon/pull/7

[Screencast_20260624_230954.webm](https://github.com/user-attachments/assets/9962631b-26b8-4bbb-a32c-91dd0731a753)

kaste:
I like both ways tbh.  @stadja's is very Sublime Texty, e.g. alt+1 to switch to it, alt+shift+1 to program it.  

Implementation note on the buffer implementation; I wouldn't go with a temp file, just a scratch buffer.  

```
view = window.new_file(); 
view.set_scratch(True); 
view.settings().set("harpoon_editor", True); 
( on refresh, delete the contents first
view.run_command("select_all"); 
view.run_command("left_delete"); 
)
view.run_command("append", content)
```

I would say go with both, but not sure they are compatible.  @stadja probably allows holes in the list, and the buffer implementation not.

huyhoang8398:
> I like both ways tbh. @stadja's is very Sublime Texty, e.g. alt+1 to switch to it, alt+shift+1 to program it.
> 
> Implementation note on the buffer implementation; I wouldn't go with a temp file, just a scratch buffer.
> 
> ```
> view = window.new_file(); 
> view.set_scratch(True); 
> view.settings().set("harpoon_editor", True); 
> ( on refresh, delete the contents first
> view.run_command("select_all"); 
> view.run_command("left_delete"); 
> )
> view.run_command("append", content)
> ```
> 

i have implemented as you suggested and it works better than the tmp file, thanks alot :smiley: 

> I would say go with both, but not sure they are compatible. @stadja probably allows holes in the list, and the buffer implementation not.

I will merge my scratch buffer implementation and I could help you re-fractor your feature since i do like having these two feature together, is it ok for you @stadja. I have only a doubt because I sometime having more than 9 marks, and end up using `Harpoon_next`, and the buffer re-order seems suitable me

kaste:
Anyway, you're on a good way.  

LGTM! @braver

PR #9455, opened 2026-06-19
Place the submission guide first in the README

kaste:
That is mostly ordering, and then to have as checklist the simple, clear steps to the goal.


braver:
Definitely makes more sense that way!

PR #9457, opened 2026-06-23
Remove MCP Commander

dpc00:
Removing MCP Commander at the author's request. The package will no longer be maintained.


dpc00:
Closing to fix a UTF-8 BOM encoding error introduced by my edit. Will resubmit with correct encoding.

PR #9462, opened 2026-06-30
Add labels to PCR's "Remarkable" packages

michaelblyons:
I took a quick run-through of the packages in [PCR](https://packages.sublimetext.io/)'s "Remarkable" section and added tags to the ones in this registry where they were missing.

The "dark" labels may be mediocre value, but a couple of the packages really needed some tagging. LaTeXTools does *even more things,* but I was reluctant to add too many labels.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -457,7 +457,7 @@
 			"name": "Catppuccin color schemes",
 			"details": "https://github.com/catppuccin/sublime-text",
 			"author": ["Catppuccin"],
-			"labels": ["color scheme"],
+			"labels": ["color scheme", "dark", "light"],
 			"releases": [
 				{
 					"sublime_text": ">=3100",
@@ -2924,6 +2924,7 @@
 		},
 		{
 			"details": "https://github.com/facelessuser/ColorHelper",
+			"labels": ["color"],
 			"releases": [
 				{
 					"sublime_text": "3170 - 3999",
--- a/repository/c.json
+++ b/repository/c.json
@@ -2924,7 +2924,7 @@
 		},
 		{
 			"details": "https://github.com/facelessuser/ColorHelper",
-			"labels": ["color"],
+			"labels": ["color", "preview", "utilities"],
 			"releases": [
 				{
 					"sublime_text": "3170 - 3999",
--- a/repository/t.json
+++ b/repository/t.json
@@ -2134,7 +2134,7 @@
 		{
 			"name": "Theme - Spacegray",
 			"details": "https://github.com/SublimeText/Spacegray",
-			"labels": ["theme", "color scheme"],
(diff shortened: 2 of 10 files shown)


repository/c.json
```diff
@@ -2924,6 +2924,7 @@
 		},
 		{
 			"details": "https://github.com/facelessuser/ColorHelper",
+			"labels": ["color"],
```
braver: “utilities”?

repository/t.json
```diff
@@ -2134,7 +2134,7 @@
 		{
 			"name": "Theme - Spacegray",
 			"details": "https://github.com/SublimeText/Spacegray",
-			"labels": ["theme", "color scheme"],
+			"labels": ["theme", "color scheme", "dark", "light"],
```
braver, line 2137: If you’re doing this… you’d have to now go and check all the color schemes. Which you’re welcome to I guess, it would even have value (I prefer light color schemes, makes them easier to find)… I’m certainly not going to police this on new color schemes though. So that makes this a bit random.

repository/c.json
michaelblyons: Sure. I went looking for more specific things, but the best I could find was `preview`, and that's wasn't very satisfying.

repository/t.json
michaelblyons: I don't _have to_ check all the color schemes. I just added a couple more to the pattern some people had already been doing. :wink: 

But I might someday check all the color schemes anyway. Don't tempt me with a ~~time-consuming low-value project~~ *[good time]*!

braver: Well you don’t really have to of course, but something tells me that you will 😄

repository/c.json
michaelblyons:
```suggestion
			"labels": ["color", "preview", "utilities"],
```

braver: Works for me.

We should actually perhaps consolidate color and colors? The singular form is indeed most common though, so this is good 👍🏻

kaste:
I'm against such changes almost in general; in this case the "Remarkable" section looked better IMO with that variance of not all packages having a sub-heading (it isn't styled as label so people actually shouldn't know immediately that this is in fact the first label that acts as secondary text).  And "brackets" for "BracketHighlighter" ... rly?

michaelblyons:
> And "brackets" for "BracketHighlighter" ... rly?

The brackets label existed already. Would you want someone to load the packages tagged with it and not include the most significant package for brackets?

kaste:
I guess it would appear if you start searching for "brack".  I'm just strange here I guess as I usually don't touch the db as the primary artifact; at least for open databases; closed dbs of course have maintenance, normalization and consistency and people doing that all the time;  visually I did like the rhythm of the three packages not having a sub-heading.

PR #9464, opened 2026-07-03
Label syntax management packages with `syntax` and `utilities`

michaelblyons:
If accepted, I will also PR

- ColinTA's registry for SetSyntax.
- Jack Cherng's registry for AutoSetSyntax.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -1661,6 +1661,7 @@
 		},
 		{
 			"details": "https://github.com/facelessuser/ApplySyntax",
+			"labels": ["syntax", "utilities"],
 			"releases": [
 				{
 					"sublime_text": "3000 - 4200",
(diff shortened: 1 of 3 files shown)


braver:
I don't think `syntax` is a very strong label. It's ambiguous and easily confused with `language syntax`: _is_ the package a syntax or does it "do something" with syntax? If I look at current usage of it, it's a mixed and mostly confusing bag. And since syntax is already in the names of those packages it doesn't add very much to repeat it in the label. `utilities` I can get behind, but I would like to move away from using `syntax` completely.

michaelblyons:
In the Discord I asked about using "syntax management" and this dual tag was the counter proposal.

I don't have a problem with "utilities," but I think something else should be tagged as well.

braver:
Alright, sure, it's not like I have super strong opinions here.

michaelblyons:
Yeah, I don't have super strong opinions about what the tags *are.* I just want to be able to link someone the "list of packages that do advanced syntax choosing" when extension+first-line are not sufficient.

PR #9465, opened 2026-07-03
Consolidate some Git labels

michaelblyons:
Add and dedup some Git package labels.

Note: GitGutter not included here because it's already in https://github.com/sublimehq/package_control_channel/pull/9462

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -562,7 +562,7 @@
 		{
 			"name": "Git Conflict Resolver",
 			"details": "https://github.com/sascha-wolf/sublime-GitConflictResolver",
-			"labels": ["git", "conflicts", "gutter"],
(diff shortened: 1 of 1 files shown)


repository/g.json
```diff
@@ -561,6 +562,7 @@
 		{
 			"name": "Git Conflict Resolver",
 			"details": "https://github.com/sascha-wolf/sublime-GitConflictResolver",
+			"labels": ["git", "conflicts", "gutter"],
```
braver: There are no live package with "[conflicts](https://packages.sublimetext.io/?q=label%3A%22conflicts%22)", so I wouldn't introduce that one. 
```suggestion
			"labels": ["git", "gutter"],
```

PR #9468, opened 2026-07-04
Add Sema

HelgeSverre:

[2]: https://semver.org

My package is language support for Sema, a Lisp with first-class LLM primitives.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1045,7 +1045,7 @@
 		{
 			"name": "Sema",
 			"details": "https://github.com/sema-lisp/sublime-sema",
-			"labels": ["language syntax"],
+			"labels": ["language syntax", "build system"],
 			"releases": [
 				{
 					"sublime_text": ">=4075",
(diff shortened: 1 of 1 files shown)


braver:
I'm going to assume you're using and agent and apparently agents are incapable of following the [PR template](https://github.com/sublimehq/package_control_channel/blob/master/.github/PULL_REQUEST_TEMPLATE.md) (as a human it's not that easy to skip the template). All the same, we don't review PR's that don't follow it. Please do try again, look like a useful package.

HelgeSverre:
Sorry, my bad! 🫡

braver:
Does the `comment_token` setting actually do anything? I don't believe that's a thing.
As for the other settings, if these are a strong convention for all users of the language that's fine, but don't ship your preferences. It's usually better to put those in the README and let users figure that out for themselves.

kaste:
> subprocess.Popen is used in a Windows-supported package ...

was a false positive.  Fixed in the next bot version.

/review

PR #9470, opened 2026-07-07
Add EDIBlime

encefalopiatto:

[2]: https://semver.org

My package is EDIBlime, an Electronic Data Interchange (EDI) helper tool focused on syntax for the most common EDI formats (EDIFACT, UN, TRADACOMS, and others). It does more than syntax highlighting, and provides help with understanding every segment, through detailed breakdowns, does beautification, envelope integrity checks, envelope repair, normalization to standard EDI ISO JSON, and simple conversion to XML, JSON and JSONC.

My package is somewhat similar to EDIFACT Highlighter (https://packagecontrol.io/packages/EdifactHighlighter) however it should still be added because this is a helper tool that goes beyond its limited highlighting purpose and adds more dialects than only EDIFACT. Also, I work in the field, and this will be maintained regularly, as it will be our main working tool for EDI file inspection in the whole company. Without this tool, our workforce is using VSCode and a dedicated plugin (https://marketplace.visualstudio.com/items?itemName=DericLee.edi-edifact-support). Overkill and resource hungry for a simple file inspection and basic edits.

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -395,7 +395,7 @@
 		{
 			"name": "EDIBlime",
 			"details": "https://github.com/encefalopiatto/EDIBlime",
-			"labels": ["language syntax"],
+			"labels": ["language syntax", "utilities", "minify", "convert"],
 			"releases": [
 				{
 					"sublime_text": ">=4095",
--- a/repository/e.json
+++ b/repository/e.json
@@ -393,6 +393,17 @@
 			]
 		},
 		{
+			"name": "EDIBlime",
+			"details": "https://github.com/encefalopiatto/EDIBlime",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": ">=4095",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "EdifactHighlighter",
 			"details": "https://github.com/Frogli/sublime-edifact",
 			"labels": ["language syntax"],


braver:
> Do not import root-level plugin module '.edi_core'. Sublime Text loads root-level Python files as independent plugins; move shared code into a subpackage.

This is the main architectural issue we need to address here. Python files in the root directory are all loaded as plugins themselves. Which can be what you want, but you then can't really use one as your entry point and load others into it. Instead move them to a subdirectory and load them from there. For development you'll need something to help reload those files, for which solutions exist. There is a bit of discussion about just this topic in this PR, that might set you on the right track: https://github.com/sublimehq/package_control_channel/pull/9355#issuecomment-4412729153

braver:
Also please consider adding a command palette entry for the settings ([example](https://github.com/braver/ColorConverter/blob/main/Commands.sublime-commands#L2-L9)).

repository/e.json
```diff
@@ -392,6 +392,17 @@
 				}
 			]
 		},
+		{
+			"name": "EDIBlime",
+			"details": "https://github.com/encefalopiatto/EDIBlime",
+			"labels": ["language syntax"],
```
braver: Maybe these are also fitting:

```suggestion
			"labels": ["language syntax", "utilities", "minify", "convert"],
```

encefalopiatto:
@braver I am working on applying the fixes you proposed. Once resolved, I will post an update here. It won't take much.

encefalopiatto:
@braver It's done.

PR #9471, opened 2026-07-08
Add Spintax

investblog:
- [x] Any commands are available via the command palette. — the package adds no commands.
- [x] Preferences and keybindings (if any) are listed in the menu and the command palette, and open in split view. — none of either.

[2]: https://semver.org

My package is a syntax definition for **spintax** — the template language used to generate
text variations (`{a|b|c}` enumerations, `[<config>a|b]` permutations, `%vars%`,
`#set`/`#def` directives, `#include`, `{?VAR?…}` conditionals, `{plural N: …}` plural
agreement, `/# … #/` comments). It highlights `.spintax` and `.gtw` files.

The grammar mirrors the `@spintax/core` engine contract rather than approximating it, so the
tricky discriminators behave the way the engine does — a leading `<and>` inside `[ … ]` is a
separator while `<li>…</li>` is content, `{plural ` requires the trailing space and a colon,
`{?VAR?` requires a valid identifier. Those cases are covered by the syntax tests, which run
in CI on Sublime Text `latest`, `stable` and build `3210`.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -3380,7 +3380,7 @@
 			"labels": ["language syntax"],
 			"releases": [
 				{
-					"sublime_text": ">=4000",
+					"sublime_text": ">=3092",
 					"tags": true
 				}
 			]
--- a/repository/s.json
+++ b/repository/s.json
@@ -3375,6 +3375,17 @@
 			]
 		},
 		{
+			"name": "Spintax",
+			"details": "https://github.com/investblog/sublime-spintax",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "SPIP",
 			"details": "https://github.com/phenix-factory/Sublime-SPIP",
 			"labels": ["language syntax"],


deathaxe:
1. The syntax in its current form is compatible with ST3092+. No need to restrict it to ST4.
2. I'd recommend proper syntax test files instead of proprietary python scripts. 

   To learn how to write tests, have a look into sublimehq/Packages.

   For CI syntax tests then maybe consider https://github.com/sublimetext/syntax-test-action

investblog:
Thanks for the review! Both points addressed:

1. **ST version** — dropped the ST4 restriction; the release selector is now `sublime_text: ">=3092"`. The syntax uses no ST4-only features.
2. **Tests** — replaced the Python script with proper syntax test files (`tests/syntax_test_spintax.spintax`) covering every construct and the tricky discriminators. CI now runs them via `SublimeText/syntax-test-action` on ST `latest`, `stable`, and ST3 build `3210`, so ST3092+ compatibility is verified on every push.

Ready for another look when you have a moment.

deathaxe:
LGTM.

braver:
I'm going to assume you're using and agent and apparently agents are incapable of following the [PR template](https://github.com/sublimehq/package_control_channel/blob/master/.github/PULL_REQUEST_TEMPLATE.md) (as a human it's not that easy to skip the template). All the same, we don't merge PR's that don't follow it. Please do try again, look like a useful package.

investblog:
Sorry, the abbreviation LGTM threw me off. I’d already thought it had been posted and I’d forgotten :)))
You're right, and sorry — the checklist was missing from the description. I've rewritten it to follow the template.

Filling it in honestly turned up one item I couldn't tick: .gitattributes was
only normalising line endings, so tests/ and .github/ were shipping inside the
installed package. Fixed in 1.1.1 — the tag archive now contains just the
syntax, Comments.tmPreferences, README, CHANGELOG and LICENSE.

Two details in the old description were also stale: the selector is
sublime_text: ">=3092" (per @deathaxe's review, already in the diff), and the
current tag is 1.1.1, not 1.0.0. Since 1.0.0 the syntax also gained the #def
directive, which the engine added alongside #set.

braver:
> Sorry, the abbreviation LGTM threw me off. I’d already thought it had been posted and I’d forgotten

No worries, we have a few community members going through PR's and flagging whether stuff looks good to them and I fully trust their judgement. But I've got my own bits of process and my own hang-ups I guess 😁 All good, thanks for the update!

PR #9472, opened 2026-07-08
Add TabStack package

FichteFoll:

[2]: https://semver.org

My package is a different take on MRU-based tab switching than the available alternatives. Its primary selling factor is that it mimics the behavior found in other places, such as in browsers or on Windows' windows where releasing the control key causes the current selection (in a quick panel) to be committed

- https://packages.sublimetext.com/packages/TabTeleport uses a text-based representation of the tab lists
- https://packages.sublimetext.com/packages/Compass%20Navigator uses a quick panel but follows a different key binding philosophy

Additionally, TabStack properly supports HTML and image sheets by using the newly added `Sheet.name` API of 4205. This was implemented in an optional way so that it supports ST 4 builds back until 4100, which includes the first stable build of ST~~4~~.

Key bindings override default bindings because this package replaces the default functionality on these keys. However, a special context is in place that will cause the bindings not to trigger if the necessary system-level libraries (for ctypes calls) cannot be loaded. This way, the default bindings can still take effect even  when the plugin fails to load properly.

There is no configuration and the commands are not meant to be invoked via the command palette. In fact, it won't even work because the ctrl key must be held at all times.

### DIFF
--- a/repository/t.json
+++ b/repository/t.json
@@ -182,6 +182,17 @@
 			]
 		},
 		{
+			"details": "https://github.com/SublimeText/TabStack",
+			"author": ["FichteFoll"],
+			"labels": ["tab navigation"],
+			"releases": [
+				{
+					"sublime_text": ">=4100",
+					"tags": "st-4100/"
+				}
+			]
+		},
+		{
 			"details": "https://github.com/jbrooksuk/TabsToTable",
 			"releases": [
 				{
--- a/repository/t.json
+++ b/repository/t.json
@@ -183,7 +183,7 @@
 		},
 		{
 			"details": "https://github.com/SublimeText/TabStack",
-			"author": ["FichteFoll"],
+			"author": "FichteFoll",
 			"labels": ["tab navigation"],
 			"releases": [
 				{


repository/t.json
```diff
@@ -181,6 +181,17 @@
...
 			]
 		},
+		{
+			"details": "https://github.com/SublimeText/TabStack",
+			"author": "FichteFoll",
+			"labels": ["tab navigation"],
+			"releases": [
+				{
+					"sublime_text": ">=4100",
+					"tags": "st-4100/"
```
FichteFoll, line 191: The tag is prefixed with this string. I suppse the reviewer does not handle prefixed tags?
https://github.com/SublimeText/TabStack/tags

braver: I believe the package review tests HEAD of the package, and the current state of handling prefixes or not is confusing to me 😅

braver: Where the schema test gets `Removed: 4721` from is beyond me though, it has a tendency to be really 😵‍💫

FichteFoll: I was referring to this line in the first review:

> No semantic version tags found

PR #9475, opened 2026-07-10
Add Extractor

bogdancstrike:

[2]: https://semver.org

My package, Extractor, scrapes structured data out of the current file (or the selection) and writes the de-duplicated matches into a new tab. It ships 22 built-in extractors — emails, URLs, domains, hostnames, IPv4/IPv6, ports, phone numbers, dates, times, timestamps, file paths, UUIDs, MAC addresses, hex colors, numbers, hashtags, mentions, Luhn-validated credit-card numbers, secrets/API keys and Markdown link URLs — plus a custom-regex extractor. Everything is reachable from the Command Palette and the Tools menu; it has no dependencies and runs on ST3 and ST4.

My package is loosely similar to Extract Text to File and Extract Lines, which write the current selection or matching lines to a file. However it should still be added because it works the other way around: instead of the user selecting or searching first, it recognises 22 kinds of tokens (or any user-supplied regex) anywhere in the file, de-duplicates them, and collects them into a new file in one step.

### DIFF
--- a/repository/e.json
+++ b/repository/e.json
@@ -1981,6 +1981,17 @@
 			]
 		},
 		{
+			"name": "Extractor",
+			"details": "https://github.com/bogdancstrike/sublime_plugin_extractor",
+			"labels": ["text manipulation", "utilities"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "eZ Publish Syntax",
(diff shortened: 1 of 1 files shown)


braver:
Interesting! Just that one remark from the bot about labelling the command palette entry, but looks good otherwise!

bogdancstrike:
I have implemented the requested changes (command palette entry for editing settings should start with 'Preferences: Extractor') and tagged a new release 1.0.1

PR #9476, opened 2026-07-10
Add Open All in Browser

bogdancstrike:

[2]: https://semver.org

My package, Open All in Browser, opens every URL found in the current selection in the default browser, each in its own tab. It recognises http://, https:// and www. links (bare www. opens over https), opens each unique URL only once, and trims trailing punctuation so links embedded in prose still resolve. It's available from the Command Palette, the Tools menu and the right-click menu, has no dependencies, and does nothing when the selection has no URLs.

My package is similar in name to Open In Browser / OpenInBrowser (which render the current file in a browser) and to Open URL (which opens a single URL under the cursor). However it should still be added because it solves a different task: it opens all URLs in a multi-line selection at once — each in its own ab, de-duplicated — rather than opening one link or previewing a file. This makes it useful for logs, Markdown, config files and link lists.

### DIFF
--- a/repository/o.json
+++ b/repository/o.json
@@ -435,6 +435,17 @@
 			]
 		},
 		{
+			"name": "Open All in Browser",
+			"details": "https://github.com/bogdancstrike/sublime_plugin_open_all_in_browser",
+			"labels": ["productivity", "utilities"],
+			"releases": [
+				{
+					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Open DOI",
 			"details": "https://github.com/crazytexer/open_doi",
 			"releases": [


braver:
Looks good and I understand how you're adding something to what's already there. We do try to avoid fragmentation though. Have you considered proposing this feature on either OpenUri or Open URL? Especially OpenUri seems already very close to what you're proposing here.

braver:
If we were to continue a separate package, please add a setting to disable the context menu item.

bogdancstrike:
I have implemented the requested changes (added a setting to disable the context menu) and tagged a new release 1.0.1

braver:
Either you or Claude made a mistake in those last edits. Your package name, and therefore also the path to the package, will be as specified in this PR: "Open All in Browser". So when you work on it locally make sure to clone into a directory with that name.

braver:
Please have a look at the last bits of feedback.

PR #9479, opened 2026-07-13
Add Claude Code IDE

Daichi-Kudo:
- [x] My package doesn't add key bindings. (Suggested bindings ship commented out in the keymap; the diff-review UI is driven by in-view Accept/Reject buttons and the command palette.)
- [x] If my package is a syntax it doesn't also add a color scheme. (Not a syntax package.)

[2]: https://semver.org

My package is a native integration of Claude Code (Anthropic's CLI coding agent) for Sublime Text 4. It implements the same WebSocket/MCP IDE-integration protocol that the official VS Code and JetBrains extensions speak (as documented by the claudecode.nvim project), so the agent gets editor context and the editor gets review control:

- in-editor side-by-side diff review of the agent's proposed edits — accept, reject, or hand-edit before accepting
- live selection / open-tabs / workspace context sharing, and `@`-mention of the current selection
- multiple concurrent Claude Code sessions against one Sublime instance

Protocol core is dependency-free Python 3.8 (unit-tested outside Sublime); the server binds to 127.0.0.1 only, with token auth. Verified end-to-end against the real Claude Code client (0.1.1). It is an unofficial community plugin and the README states the trademark relationship prominently.

My package is similar to Claudette in that both talk to Anthropic models. However it should still be added because Claudette is a chat client inside the editor, while this package integrates the Claude Code agent CLI running in a terminal via its IDE protocol — different tool, different workflow (agentic edits with in-editor review rather than chat).

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1376,6 +1376,17 @@
 			]
 		},
 		{
+			"name": "Claude Code IDE",
+			"details": "https://github.com/Cognisant-llc/sublime-claude-code",
+			"labels": ["ai", "claude", "diff"],
+			"releases": [
+				{
+					"sublime_text": ">=4107",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Claudette",
 			"details": "https://github.com/barryceelen/Claudette",
 			"labels": ["ai", "claude"],
--- a/repository/c.json
+++ b/repository/c.json
@@ -1398,6 +1398,17 @@
 			]
 		},
 		{
+			"name": "Claude Code IDE",
+			"details": "https://github.com/Cognisant-llc/sublime-claude-code",
+			"labels": ["ai", "claude", "diff"],
+			"releases": [
+				{
+					"sublime_text": ">=4107",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "ClaudeSublime",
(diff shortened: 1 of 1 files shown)


braver:
Please have a look at the review output and resolve the conflict.

Daichi-Kudo:
Conflict resolved — rebased onto current master (ClaudeSublime landed at the same spot in c.json).

The review output was addressed in release 0.2.0:
- `args.user_file` added to the Key Bindings entries in Default.sublime-commands and Main.sublime-menu
- settings file renamed to `Claude Code IDE.sublime-settings`
- commented keymap examples moved to an uncommented `Example.sublime-keymap`, linked from the menu
- the release tag is now at the tip of main

/review

braver:
While I don't think it's necessarily healthy for us to have an ever increasing number of mostly vibe-coded Claude integrations in the ecosystem, probably the way forward for now is to let it shake out in practice (I'm not convinced the Claudette project is still alive for instance). So, not to be unfair to your effort here, but instead of everyone whipping up their own short lived whatever, it would be better if some people stuck their heads together to build the LSP or SublimeLinter of coding agents. 

My grumblings aside, this looks fine. I can't tell if this is the one has the juice to carry it over time, but let's hope it does :)

PR #9481, opened 2026-07-15
Add Recent Explorer plugin details to r.json

kushaagr:

[2]: https://semver.org

My package is **Recent Explorer**, a lightweight, project-dependent, keyboard-first dashboard view. It dynamically crawls your open sidebar project directories to discover and display source code files, sorting them instantly using intrinsic system timestamps (creation or modification times). 

### Regarding Key Bindings:
I left the key bindings box unchecked because the package includes `Default.sublime-keymap` and `Default.sublime-mousemap` files. However, **these bindings are completely safe and do not override global keys**. They are strictly restricted via Sublime Text context constraints to only trigger when the user is actively inside the custom `recent_explorer_dashboard` scratch view buffer:
```json
"context": [
    { "key": "setting.recent_explorer_dashboard", "operator": "equal", "operand": true }
]
```

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -787,6 +787,17 @@
 			]
 		},
 		{
+			"name": "Recent Explorer",
+			"details": "https://github.com/kushaagr/recent-explorer-sublime",
+			"labels": ["dashboard", "recent files", "project navigation", "explorer", "productivity"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"branch": "main"
+				}
+			]
+		},
+		{
 			"name": "RecentTabSort",
 			"details": "https://github.com/Stuk/sublime-recent-tab-sort",
 			"releases": [
--- a/repository/r.json
+++ b/repository/r.json
@@ -793,7 +793,6 @@
 			"releases": [
 				{
 					"sublime_text": "*",
-					"branch": "main",
 					"tags": true
 				}
 			]
--- a/repository/r.json
+++ b/repository/r.json
@@ -789,7 +789,7 @@
 		{
 			"name": "RecentExplorer",
 			"details": "https://github.com/kushaagr/recent-explorer-sublime",
-			"labels": ["project", "open-files", "file-navigation", "productivity"],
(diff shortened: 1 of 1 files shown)


repository/r.json
```diff
@@ -786,6 +786,17 @@
 				}
 			]
 		},
+		{
+			"name": "RecentExplorer",
+			"details": "https://github.com/kushaagr/recent-explorer-sublime",
+			"labels": ["dashboard", "recent files", "project navigation", "explorer", "productivity"],
```
braver: Please have a look at currently used labels and see if you can match some that are already in there: https://packages.sublimetext.com/labels. Introducing new labels unique to your package adds more noise than helping anyone really, at some point that kind of information is better put in the package description.

braver:
Looks like a nice package! Your point about the keybindings is totally fair. Just some minor remarks from the automated tests that need addressing.

repository/r.json
kushaagr: Ok, I've looked at the mentioned URL and updated the labels lists.

repository/r.json
```diff
@@ -786,6 +786,17 @@
 				}
 			]
 		},
+		{
+			"name": "RecentExplorer",
+			"details": "https://github.com/kushaagr/recent-explorer-sublime",
+			"labels": ["project", "open-files", "file-navigation", "productivity"],
```
braver:
```suggestion
			"labels": ["project", "open files", "file navigation", "productivity"],
```

- [file navigation](https://packages.sublimetext.io/?q=label%3A%22file+navigation%22)
- [open files](https://packages.sublimetext.io/?q=label%3A%22open+files%22)

kushaagr:
- Removed `Default.sublime-mousemap` from the repo
- Added suggestion in README to update mousemap file in User directory

braver:
Excellent, looking good. Please tag your latest work and we're ready to ship it!

kushaagr:
Done! Updated tag v1.0.0 to latest main commit

PR #9482, opened 2026-07-15
Add Carve

dereuromark:

[2]: https://semver.org

My package is a syntax highlighting package for [Carve](https://github.com/markup-carve/carve), a post-Markdown lightweight markup language (`.crv` files). It ships a single `sublime-syntax` (v1, so Sublime Text 3 and 4 are both supported) whose scope names mirror the canonical `text.carve` TextMate grammar, plus syntax-specific editing defaults (soft tabs, word wrap, spell check) that apply automatically to Carve files. There are no commands, menus, key bindings, or color schemes; the settings file is the syntax-specific kind that Sublime Text picks up by name, so there is nothing to surface in the command palette.

There are no packages like it in Package Control. Carve is a distinct language (not Markdown or Djot), and no existing package highlights it.

Sorry for skipping the template the first time - that was my mistake in the submission flow, not lack of intent to maintain this. I am the author of both the language and this package and will support it long term.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -405,7 +405,6 @@
 			"name": "Carve",
 			"details": "https://github.com/markup-carve/sublime-carve",
 			"labels": ["language syntax", "carve", "markup", "djot"],
-			"description": "Syntax highlighting for Carve, a post-Markdown lightweight markup language",
 			"releases": [
 				{
 					"sublime_text": ">=3092",
--- a/repository/c.json
+++ b/repository/c.json
@@ -402,6 +402,18 @@
 			]
 		},
 		{
+			"name": "Carve",
+			"details": "https://github.com/markup-carve/sublime-carve",
+			"labels": ["language syntax", "carve", "markup", "djot"],
+			"description": "Syntax highlighting for Carve, a post-Markdown lightweight markup language",
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Case Conversion",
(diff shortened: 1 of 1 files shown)


braver:
I'm going to assume you're using and agent and apparently agents are incapable of following the [PR template](https://github.com/sublimehq/package_control_channel/blob/master/.github/PULL_REQUEST_TEMPLATE.md) (as a human it's not that easy to skip the template). All the same, we don't review PR's that don't follow it. Please do try again, look like a useful package.

dereuromark:
Fair point, and sorry about that. I've rewritten the description to follow the template properly.

While going through the checklist I also noticed the repo was missing a `.gitattributes`, so I added one (export-ignore for the test file, `.github/`, and `.gitattributes` itself) and tagged `0.1.1` so the shipped package only contains the package files.

repository/c.json
```diff
@@ -401,6 +401,18 @@
 				}
 			]
 		},
+		{
+			"name": "Carve",
+			"details": "https://github.com/markup-carve/sublime-carve",
+			"labels": ["language syntax", "carve", "markup", "djot"],
+			"description": "Syntax highlighting for Carve, a post-Markdown lightweight markup language",
```
braver: You can remove the "description", it will then be taken from your GitHub repo, also meaning you can update it at will.

```suggestion
```

braver:
As for the settings... settings like this on a syntax aren't usually recommended. A lot of it is personal (or team) preference, and might be unexpected for users. If you do want to provide them, expose them in the Package Settings menu. But alternatively you could also just mention them in the README and let users set and tweak their own.

PR #9483, opened 2026-07-16
Add Vibescript

mgomes:

[2]: https://semver.org

My package is language support for [Vibescript](https://github.com/mgomes/vibescript), a Ruby-inspired embeddable scripting language: syntax highlighting for `.vibe` files (strings with interpolation, regex literals, percent arrays, enums, typed signatures), indentation rules, symbol indexing for classes/methods/enums, and LSP integration (diagnostics, hover documentation, completions) through the `vibes lsp` server and the LSP package. Syntax tests run in CI against real Sublime builds.

### DIFF
--- a/repository/v.json
+++ b/repository/v.json
@@ -334,6 +334,17 @@
 			]
 		},
 		{
+			"name": "Vibescript",
+			"details": "https://github.com/mgomes/sublime-vibescript",
+			"labels": ["language syntax", "lsp"],
+			"releases": [
+				{
+					"sublime_text": ">=4107",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "View Bookmarks",
 			"details": "https://github.com/adam-zethraeus/SublimeViewBookmarks",
 			"labels": ["bookmarks", "file navigation"],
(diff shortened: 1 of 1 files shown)


rchl:
The convention and probably just better option is to have the LSP helper be a separate package from the syntax. Then people can decide if they want to use it. As it is now, the user is forced to install LSP or else the code will trigger exceptions when importing LSP.

Also, new LSP helpers should use the new LspPlugin class - https://lsp.sublimetext.io/migrating_to_lsp_plugin/

mgomes:
Thanks! done on both counts.

The package is split: [Vibescript](https://github.com/mgomes/sublime-vibescript) (tagged 1.1.0) is now syntax-only with no LSP import anywhere, and the helper moved to a new [LSP-vibescript](https://github.com/mgomes/LSP-vibescript) package built on the new `LspPlugin` class per the migration guide, submitted separately to sublimelsp/repository#162.

braver:
Looks good, thanks for splitting the packages. If you want you can add the settings also to the command palette, but I won't let that block the merge.

PR #9484, opened 2026-07-16
Add Gold Rose

zacharytruong:
- [✅] I'm the package's author and/or maintainer.
- [✅] I have read [the docs][1].
- [✅] I have tagged a release with a [semver][2] version number.
- [✅] My package repo has a description and a README describing what it's for and how to use it.
- [✅] My package doesn't add context menu entries. *
- [✅] My package doesn't add key bindings. **
- [✅] Any commands are available via the command palette.
- [✅] Preferences and keybindings (if any) are listed in the menu and the command palette, and open in split view.
- [✅] If my package is a syntax it doesn't also add a color scheme. ***
- [✅] I use [.gitattributes][3] to exclude files from the package: images, test files, sublime-project/workspace.

[2]: https://semver.org

My package is ...

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -1406,6 +1406,17 @@
 			]
 		},
 		{
+			"name": "Gold Rose",
+			"details": "https://github.com/zacharytruong/gold-rose",
+			"labels": ["color scheme", "theme"],
+			"releases": [
+				{
+					"sublime_text": ">=4050",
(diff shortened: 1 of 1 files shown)


braver:
I would recommend adding a screenshot to your readme so people get a better idea what it would look like before they install it.

PR #9485, opened 2026-07-16
Master odoo

Mr7F:

### DIFF
--- a/repository/o.json
+++ b/repository/o.json
@@ -216,6 +216,17 @@
 			]
 		},
 		{
+			"name": "Odoo",
+			"details": "https://github.com/Mr7F/sublime-odoo",
+			"labels": ["odoo", "language syntax", "snippets", "auto-complete"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Odoo Snippets",
 			"details": "https://github.com/Jarsa/Sublime-Odoo",
 			"labels": ["snippets", "odoo"],


braver:
Have a look at the test results and please tell us some more about the package. I couldn't really grasp from the readme what we're looking at here :)

Mr7F:
@braver It's a package to help developing Odoo module

https://www.odoo.com/documentation/19.0/developer/reference/frontend/owl_components.html
https://www.odoo.com/documentation/19.0/developer/reference/user_interface/view_architectures.html

With different syntax (condition are controlled by XML attributes, some for loop, etc)

There's an existing package, but very outdated, and with only snippet (no color highlighting auto-complete, etc: https://packages.sublimetext.io/packages/Odoo%20Snippets)

braver:
Alright, thanks for that :) Package looks good. Only thing to watch out for is that extending default syntaxes is a bit risky. It can get very hard to maintain compatibility across multiple versions of Sublime Text and/or to to keep up with those syntaxes. I guess it works as long as it works and it's more elegant than having a copy of that default syntax, but that's something I did end up doing myself in similar cases. So, good for now, just something to be mindful of.

repository/o.json
```diff
@@ -215,6 +215,17 @@
 				}
 			]
 		},
+		{
+			"name": "Odoo",
+			"details": "https://github.com/Mr7F/sublime-odoo",
+			"labels": ["odoo"],
```
braver:
```suggestion
			"labels": ["odoo", "language syntax", "snippets", "auto-complete"],
```

I believe these might also apply and give a bit more insight into what the package provides.

Mr7F:
@braver Thanks 🙂 I did the changes 🙂

PR #9491, opened 2026-07-20
Add LSP-carve

dereuromark:

[2]: https://semver.org

My package is an LSP client for [Carve](https://github.com/markup-carve/carve), a post-Markdown lightweight markup language. It runs the language server (`carve-lsp`) through `lsp_utils`, which installs it and provides the Node runtime, so users need no global Node. It pairs with the Carve syntax package, which was added to the channel in #9482.

The repository is named `sublime-carve-lsp` to sit alongside the syntax repo (`sublime-carve`) and to avoid confusion with `carve-lsp`, the language server itself; the package installs under the conventional `LSP-carve` name.

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -1857,6 +1857,17 @@
 			]
 		},
 		{
+			"name": "LSP-carve",
+			"details": "https://github.com/markup-carve/sublime-carve-lsp",
+			"labels": ["language server", "lsp", "carve", "markup"],
+			"releases": [
+				{
+					"sublime_text": ">=4096",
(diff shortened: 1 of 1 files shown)


rchl:
LSP-* packages are by convention added in a separate channel maintained in https://github.com/sublimelsp/repository

dereuromark:
Thanks for the pointer, I didn't know LSP-* helpers had their own channel. Re-opened in the right place: sublimelsp/repository#165.

Closing this one. Sorry for the noise.

PR #9493, opened 2026-07-22
Add Candela Color Schemes

schovi:
- [x] I have read the docs.

My package is a family of color schemes (light and dark) tuned for eye-strain comfort: low-glare backgrounds, desaturated accents, and WCAG-AA contrast throughout. It ships only `.sublime-color-scheme` files (no commands, menus, key bindings, or syntaxes).

My package is similar to other color scheme packages in Package Control. However it should still be added because it's a single coordinated multi-scheme family generated from one source of truth, with a specific low-glare / contrast-floor design shared across editors and terminals, rather than a single standalone scheme.

The color schemes are generated from a source-of-truth repo (https://github.com/schovi/candela-themes) and published to the tagged distribution repo linked below; I maintain both.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -269,6 +269,17 @@
 			]
 		},
 		{
+			"name": "Candela Themes",
+			"details": "https://github.com/schovi/candela-themes-sublime",
+			"labels": ["color scheme", "theme"],
+			"releases": [
+				{
+					"sublime_text": ">=3149",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Candy Dark Color Scheme",
 			"details": "https://github.com/rahulsharmagg/Candy-Dark",
 			"labels": ["color scheme"],
--- a/repository/c.json
+++ b/repository/c.json
@@ -269,9 +269,9 @@
 			]
 		},
 		{
-			"name": "Candela Themes",
+			"name": "Candela Color Schemes",
 			"details": "https://github.com/schovi/candela-themes-sublime",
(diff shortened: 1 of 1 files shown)


repository/c.json
```diff
@@ -268,6 +268,17 @@
 				}
 			]
 		},
+		{
+			"name": "Candela Themes",
+			"details": "https://github.com/schovi/candela-themes-sublime",
+			"labels": ["color scheme", "theme"],
```
braver: In Sublime Text "themes" are for the UI, "color schemes" are for syntax highlighting. These are "color schemes", not "themes".

```suggestion
			"name": "Candela Color Schemes",
			"details": "https://github.com/schovi/candela-themes-sublime",
			"labels": ["color scheme"],
```

braver:
Nice! I'm definitely going to try these. Always love seeing a good light color scheme that maintains readable and consistent contrast throughout. 

As far as the review goes there is just the naming convention thing. Please don't refer to these as "themes" in the context of Sublime Text, that will confuse users.

schovi:
@braver Applied, ready for another look.

braver:
The change here looks basically fine but please also make that change on the readme. Users will see that too, especially on Package Control and it sends confusing messaging as to what you're providing. It's fine if you use the word "theme" in various places, but especially the title of the readme and the repo description should not.

schovi:
@braver Done, both are live on the package repo (`schovi/candela-themes-sublime`, tagged v0.2.6):

PR #9494, opened 2026-07-23
Add ClaudeSublime

petrjakub:
- [x] If my package is a syntax it doesn't also add a color scheme. *** (not a syntax)

[2]: https://semver.org

My package, **ClaudeSublime**, runs Claude Code (Anthropic's CLI coding agent) in a
Terminus output panel rooted at the project, and live-highlights the files Claude
edits by tailing its session log. It also offers optional rsync-over-SSH sync of the
project to/from a remote host (macOS/Linux). Docs: https://gitlab.com/petr.jakub/claudesublime

There are no packages like it in Package Control — the closest, *Claudette*, is
unrelated (a color scheme). Existing terminal/sync packages (Terminus, rsync helpers)
each cover only one piece; ClaudeSublime combines an AI-agent panel with change
highlighting derived from the agent's own edit log.

A couple of notes on the checklist:

- **Key bindings:** none are bound by default. The platform keymaps ship every
  shortcut *commented out* as a suggestion, and a "Key Bindings" entry under
  Preferences → Package Settings opens them in split view. Every command is also in
  the Command Palette.
- **Context menu:** I do add a single, nested "ClaudeSublime" submenu to the editor
  and sidebar context menus (not multiple top-level items). All of its commands are
  conditional (`is_enabled`), so they grey out where they don't apply. I kept these
  because the sidebar sync / "add file to context" actions are the natural
  right-click workflow, but I'm happy to adjust if you'd prefer them removed.

Terminus is a runtime dependency; since Package Control can't declare a package (vs.
library) dependency, the plugin detects when Terminus is missing and offers to
install it via Package Control.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -1387,6 +1387,16 @@
 			]
 		},
 		{
+			"name": "ClaudeSublime",
+			"details": "https://gitlab.com/petr.jakub/claudesublime",
+			"releases": [
+				{
+					"sublime_text": ">=4107",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Claudette",
 			"details": "https://github.com/barryceelen/Claudette",
 			"labels": ["ai", "claude"],


petrjakub:
Thanks for the review! Addressed in **1.0.2**:

- Moved the key bindings into a single `Example.sublime-keymap` (left uncommented) and wired the *Preferences → Package Settings → ClaudeSublime → Key Bindings* menu entry to it via `edit_settings` (with `user_file` pointing at the user keymap), plus a `Preferences: ClaudeSublime Key Bindings` palette command. Nothing is bound by default.
- Added a `Preferences: ClaudeSublime Settings` palette command using `edit_settings`.
- The notification-banner subprocesses now pass `STARTF_USESHOWWINDOW` / `CREATE_NO_WINDOW` on Windows (via a `_no_window_kwargs()` helper), so no console window flashes. (The other flagged `Popen` is the rsync call, which runs inside a `pty` and is gated off on Windows entirely, but it now passes the same kwargs for uniformity.)

On the two remaining warnings:

- **`.no-sublime-package`** is intentional and necessary: the plugin generates its gutter icons at runtime into `Packages/ClaudeSublime/icons/`, which isn't writable inside a zipped `.sublime-package`, so it has to install unpacked.
- **Package name** — I'd prefer to keep `ClaudeSublime`. It's the established name of the project and how current users already refer to it. Happy to discuss if it's a hard blocker.

Tip of `main` is now tagged **1.0.2**. Thanks again!

kaste:
> the plugin generates its gutter icons

I didn't look at what you generate *exactly*, but it sounds like `sublime.cache_path() / ClaudeSublime` can be used for that as this path is part of the Sublime's resources index.

/review

petrjakub:
Good call — thanks. I went one step further than the cache dir: the two gutter icons are now **committed static PNGs** (`icons/added_line.png`, `icons/deleted_line.png`) instead of being generated at all. Sublime serves them from inside the `.sublime-package`, so the plugin no longer writes anywhere under `Packages/` and I've **dropped `.no-sublime-package`** entirely — it now installs packed like a normal package. Removed the runtime PNG generation and the unused `struct` import too.

Tip of `main` is tagged **1.0.3**. That should clear the `.no-sublime-package` warning; the only remaining one is the package name, which I'd still like to keep. Thanks for the quick reviews!

braver:
/review

> the only remaining one is the package name, which I'd still like to keep. 

That's reasonable in this case 👍🏻

braver:
Cool, I have no immediate notes otherwise, so let's just see how this goes :)

PR #9495, opened 2026-07-26
Add CompareLive

mwei-io:

[2]: https://semver.org

My package is CompareLive, a live diff comparison plugin for Sublime Text 4. It opens two tabs side by side in a dedicated two-pane window and refreshes diff highlights in real time as you type, with two-layer line + word-level intraline highlighting, synchronized scrolling, and ghost placeholder lines to keep both sides visually aligned.

Regarding the context menu (*): the package adds a single compact `CompareLive` submenu to the editor context menu and one entry to the tab context menu, since picking "this tab vs. that tab" is inherently a tab/cursor-context action. The entries are conditional: "Compare with Selected Tab" is only visible when at least two tabs are open (`is_visible`), and "End Compare" is only enabled while a compare session is active (`is_enabled`).

My package is similar to Compare Side-By-Side, FileDiffs and CompareBuff. However it should still be added because those packages produce a static diff snapshot, while CompareLive keeps the comparison alive: the diff recomputes and re-highlights on every edit, so you can fix differences on the editable side and watch highlights disappear as the two files converge — no need to re-run the comparison. The dedicated compare window also leaves the original window/layout untouched and restores it when the session ends.

### DIFF
--- a/repository/c.json
+++ b/repository/c.json
@@ -2350,7 +2350,7 @@
 					"tags": true
 				}
 			]
-		},		
+		},
 		{
 			"name": "CodeStats",
 			"details": "https://github.com/code-stats/code-stats-sublime",
@@ -3321,6 +3321,17 @@
 			"releases": [
 				{
 					"sublime_text": ">=3143",
+					"tags": true
+				}
+			]
+		},
+		{
+			"name": "CompareLive",
+			"details": "https://github.com/mwei-io/CompareLive",
+			"labels": ["diff", "compare", "productivity", "text manipulation", "utilities"],
+			"releases": [
+				{
+					"sublime_text": ">=4107",
 					"tags": true
 				}
 			]
--- a/repository/c.json
+++ b/repository/c.json
@@ -3326,7 +3326,6 @@
 			]
 		},
 		{
-			"name": "CompareLive",
 			"details": "https://github.com/mwei-io/CompareLive",
 			"labels": ["diff", "compare", "productivity", "text manipulation", "utilities"],
 			"releases": [


mwei-io:
About .no-sublime-package: it is required. This plugin writes files into its own package directory at runtime — localized menu/command files (regenerated according to the user's language setting) and a color-scheme overlay .sublime-color-scheme matching the active color scheme. With a zipped .sublime-package install the package directory is read-only, so the package must be installed unpacked.All review findings have been addressed: shared modules moved into a lib subpackage (from .lib import ...), static Default.sublime-commands and Main.sublime-menu shipped with the package, and the redundant name field removed from the entry. The v1.0.0 tag was re-pointed to the fixed commit as the package is not yet published.

kaste:
> About .no-sublime-package: it is required.

Possible solution: Try using `sublime.cache_path() / YourPackageName`.

mwei-io:
Thank you for the suggestion about `sublime.cache_path()`!
I've followed your advice and updated all runtime writes to use `sublime.cache_path()/CompareLive`. As a result:
- The plugin now writes menu/command files and color scheme overlays to `%LOCALAPPDATA%\Sublime Text\Cache\CompareLive\`, which is always writable regardless of installation mode (zip or unpacked).
- `.no-sublime-package` has been removed from the repository.
- Static English versions of menu files remain in the package archive, and runtime will automatically generate localized versions that overwrite them at cache path.
The tag `v1.0.0` has been re-pointed to this commit.
Thanks again for the helpful hint — it simplified things significantly!

braver:
Why is the "End Current Compare" command in the settings menu? I'd expect that more somewhere around the Tools menu, perhaps together with the command to initiate the comparison?

mwei-io:
Thank you for the feedback! You're absolutely right - the "End Current Compare" command should be grouped with the comparison-related commands in the Tools menu, not scattered elsewhere.I've just updated the menu structure to follow your suggestion:
Both "Compare with Selected Tab" and "End Compare" are now together under Tools → CompareLive
The settings are accessible via Preferences → Package Settings → CompareLive
The changes align with Sublime Text plugin conventions and improve discoverability. Thanks again for pointing this out!

PR #9499, opened 2026-07-27
Add 4 packages: myBasis, myHeaders, myIrix, myTest

robbwdoering:
This PR adds 4 non-syntax packages by the same author (me); a sibling PR adds my 10 syntax packages. Happy to split further into per-package PRs if preferred. All 4 share the same release setup: public GitLab repos, `v0.7.0` semver tags, protected `v*` tags (Maintainer-only), `.gitattributes` export-ignore, thorough READMEs.

- [x] I'm the packages' author and/or maintainer.
- [x] My package repos have a description and a README describing what they're for and how to use them.
- [x] My packages don't add context menu entries. *
- [x] My packages don't add key bindings. **

[2]: https://semver.org
[3]: https://git-scm.com/docs/gitattributes#_export_ignore

The packages:

| Name | Kind | What it is |
| --- | --- | --- |
| myBasis | utilities | Typed infrastructure for Sublime Text 4 plugin development — the shared base my other plugin packages build on. |
| myHeaders | utilities | Standardized hierarchical comment headers: insert, navigate, and fold code by architectonic section headers. |
| myIrix | color scheme | The Irix color palette (Radix-derived) as a Sublime color scheme, generated from the palette's single source of truth. |
| myTest | utilities | The testing half of a two-package Sublime plugin dev stack (with myBasis): run pytest against plugin code from inside Sublime. |

Regarding the only major policy violation I'm aware of: The pascalCase `my` prefix is a personal brand convention shared across the set. I've tried to follow the other rules though, especially around when to include "Sublime" and/or "Syntax". Let me know if it's a serious blocker and I can conform :)

### DIFF
--- a/repository/m.json
+++ b/repository/m.json
@@ -3037,11 +3037,44 @@
 			]
 		},
 		{
+			"name": "myBasis",
+			"details": "https://gitlab.com/doering-ai-sublime/basis",
+			"labels": ["utilities"],
+			"releases": [
+				{
+					"sublime_text": ">=4107",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Mybb Template Editor",
 			"details": "https://github.com/ionutvmi/SublimeMybbTplEditor",
 			"releases": [
 				{
 					"sublime_text": ">=3000",
+					"tags": true
+				}
+			]
+		},
+		{
+			"name": "myHeaders",
+			"details": "https://gitlab.com/doering-ai-sublime/headers",
+			"labels": ["utilities"],
+			"releases": [
+				{
+					"sublime_text": ">=4107",
+					"tags": true
+				}
+			]
+		},
+		{
+			"name": "myIrix",
+			"details": "https://gitlab.com/doering-ai-sublime/irix",
+			"labels": ["color scheme"],
+			"releases": [
+				{
+					"sublime_text": ">=4107",
 					"tags": true
 				}
 			]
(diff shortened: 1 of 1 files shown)


robbwdoering:
See explanation / AI statement / context in the sister PR.

robbwdoering:
Sorry, had to adapt to the requirement for flat GitLab repos (they have nested ones, v. fancy, come join us!). No other changes planned to either PR. 

**EDIT:** ...aaand one more change cought after moving on to the LSP PRs.

braver:
My issues with these, except perhaps for the color scheme, is how one should come to the conclusion that e.g. `myBasis` (well _your_ basis, not mine 😅) is something they need? The name doesn't mean anything and the "utilities" label doesn't help much either. Even from your explanation the myBasis + myTest sounds very much like something _you_ use for a certain workflow but it's not clear why anyone else should. Various testing suites including some for pytest already exist in the ecosystem.
 And as was already mentioned in the other PR you can't really introduce a suite of packages that interdepend in this way. I think you need to step back and reconsider beyond what "works on myMachine" and how each individual bit of functionality you want to provide would fit in the ecosystem and be useful to others.

braver:
So, yeah, we're going with "no" right now. This approach to building packages is fine for individual devs, or even inside a team or something, but not for shipping via Package Control. You're welcome to open new PR's for individual standalone packages of course though.

PR #9500, opened 2026-07-27
Add 10 syntax packages: Kitty, Logfile, Metasyntax, MystMarkdown, myCSV, myLedger, myPython, myRegex, myTaskfile, myYAML

robbwdoering:
This PR adds 10 syntax packages by the same author (me); a sibling PR adds my 4 non-syntax packages. Happy to split further into per-package PRs if preferred. All 10 share the same release setup: public GitLab repos, `v0.7.0` semver tags, protected `v*` tags (Maintainer-only), `.gitattributes` export-ignore, thorough READMEs, and syntax fixtures run in CI.

- [x] I'm the packages' author and/or maintainer.
- [x] My package repos have a description and a README describing what they're for and how to use them.
- [x] My packages don't add context menu entries. *
- [x] My packages don't add key bindings. **

[2]: https://semver.org
[3]: https://git-scm.com/docs/gitattributes#_export_ignore

The packages:

| Name | What it is |
| --- | --- |
| Kitty | Syntax for the kitty terminal's `kitty.conf` — comments, booleans, fonts, colors, key/mouse mappings, `include` — built to make large kitty configs navigable. |
| Logfile | Best-effort syntax for general unix logs (`/var/log`, `journalctl` output): timestamps, log levels, PIDs, hosts, common structural patterns, without being tied to one daemon's format. |
| Metasyntax | Ground-up reimplementation of Sublime's own `.sublime-syntax` definition highlighting, with full context/meta scoping for tooling built on scopes. |
| MystMarkdown | Syntax highlighting for MyST-flavored Markdown (directives, roles, fences, frontmatter), extending the default Markdown syntax. |
| myCSV | Extension of Sublime's built-in CSV support — richer delimiter/dialect handling and scoped fields for tooling. |
| myLedger | Syntax for Ledger/hledger plain-text accounting journals. |
| myPython | Small extension of the default Python syntax (extra scopes used by my tooling), layered on Default/Python. |
| myRegex | Syntax for Matthew Barnett's `regex`-module flavor (inline flags, named groups, recursion, etc.). |
| myTaskfile | Syntax for Task (taskfile.dev) Taskfiles, extending YAML with task/dependency/command constructs. |
| myYAML | Extensions of default YAML: a `yamlfmt` middleman handling `%YAML 1.2` documents, plus experimental YAML-subtype syntaxes (YAMLPath, BashYAML). Experimental, released early — see README. |

Similar packages: myCSV overlaps with `Advanced CSV`, `CSV`, and `rainbow_csv`; it should still be added because it extends ST4's built-in CSV syntax rather than replacing it, and its scopes are designed for downstream tooling. The other 9 have no close equivalents in Package Control. The `my` prefix is a personal brand convention shared across the set (cf. existing `myPDDL`); all names are unique in the channel.

### DIFF
--- a/repository/k.json
+++ b/repository/k.json
@@ -330,6 +330,17 @@
 			]
 		},
 		{
+			"name": "Kitty",
+			"details": "https://gitlab.com/doering-ai-sublime/kitty",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": ">=4107",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Kivy Language",
 			"details": "https://github.com/ivlevdenis/kivylng",
 			"labels": ["auto-complete", "build system", "language syntax", "snippets"],
--- a/repository/l.json
+++ b/repository/l.json
@@ -1620,6 +1620,17 @@
 			]
 		},
 		{
+			"name": "Logfile",
+			"details": "https://gitlab.com/doering-ai-sublime/logfile",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": ">=4107",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Logger Snippets",
 			"details": "https://github.com/ssrihari/sublime-logger-snippets",
 			"labels": ["snippets"],
--- a/repository/m.json
+++ b/repository/m.json
@@ -1562,6 +1562,17 @@
 			]
 		},
 		{
+			"name": "Metasyntax",
+			"details": "https://gitlab.com/doering-ai-sublime/metasyntax",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": ">=4107",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Meteor Autocomplete (TernJS)",
 			"details": "https://github.com/slava/tern-meteor-sublime",
 			"releases": [
@@ -3047,6 +3058,28 @@
 			]
 		},
 		{
+			"name": "myCSV",
+			"details": "https://gitlab.com/doering-ai-sublime/csv",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": ">=4107",
+					"tags": true
+				}
+			]
+		},
+		{
+			"name": "myLedger",
+			"details": "https://gitlab.com/doering-ai-sublime/ledger",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": ">=4107",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "myPDDL",
 			"details": "https://github.com/Pold87/myPDDL",
 			"labels": ["language syntax"],
@@ -3065,6 +3098,28 @@
 			"releases": [
 				{
 					"sublime_text": ">=3084",
+					"tags": true
+				}
+			]
+		},
+		{
+			"name": "myPython",
+			"details": "https://gitlab.com/doering-ai-sublime/python",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": ">=4107",
+					"tags": true
+				}
+			]
+		},
+		{
+			"name": "myRegex",
+			"details": "https://gitlab.com/doering-ai-sublime/regex",
+			"labels": ["language syntax"],
+			"releases": [
+				{
+					"sublime_text": ">=4107",
 					"tags": true
 				}
 			]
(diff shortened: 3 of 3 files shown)


robbwdoering:
And to answer the obvious question a bit early: no, I would not describe any of these packages (in this pull request or its sibling) as in any way vibe-coded. The quickest explanation is just to look at my GitLab contributions to these repos, which are years of slow iteration, ~12mo of more dedicated, gradually-increasing-in-scope work, and then a sprint starting ~July 1st using closely-orchestrated & bounded fleets of agents to polish my "pile of shame" ([so-to-speak](https://www.reddit.com/media?url=https%253A%252F%252Fpreview.redd.it%252Fpile-of-shame-v0-6frf5f3b33be1.jpeg%253Fauto%253Dwebp%2526s%253Dc9ce275b9a819f15879e726ab8581a4d)). I stand by everything in these repos 100% as my own work / responsibility.

Would love to chat more in depth over text or any other medium if any of the reviewers here has questions, especially since two of the repos in the other PR are intended as ambitious explorations of the future of Sublime packaging (base class APIs, plugin mocking, and Pytest integration). 

I'm entering a different phase where I'll be looking to work full time and thus don't have as much time as I did when I first made these packages, but on a more relaxed schedule I'd like to offer any services I can to help y'all with Package Control / the migration / etc.. **I'd be especially interested in taking on any backlog items related to security** -- I think there's a lot our community could stand to learn from PyPI for example, and arguably our unusually-privileged access as developers of development software calls for *more* stringent layers of swiss cheese, IMHO. Y'all are doing great already, so: open offer!

deathaxe:
1. Package names like "myPython" appear to be something rather personal/private and thus are maybe not ideal names for public packages.

2. Packages must not contain `__init__.py` in root directory. At least, it won't have any effect as Package Control removes it when installing a package to avoid conflicts.

3. It seems various packages depend on each other. 

   Metasyntax and myPython for instance fail loading, when installed individually.
   
   With Package Control not resolving such dependencies, it is very likely users running into trouble. Packages should therefore be self-contained or at most depend on always available Default packages.

4. Metasyntax extends ST's core YAML syntax which certainly fails or may cause various issues for early ST4 builds such as 4107. YAML has been refactored for being a good base syntax in later builds. 

   This raises overall questions about validity of choosen ST version specifiers  as it might also apply to other pakages in this changeset.
  
5. IMHO, most of assumptions being made about PackageDev in Metasyntax package are plain wrong.

   Despite not using modern "extends" PackageDev re-uses core YAML syntaxes in various locations. 
   
   All custom package specific scopes are applied on top of default YAML scopes. I don't see (m)any out-dated ones, which are too far away from current standards. They are just package specific choices.
   
   Could it be modernized with regards to structure and scope names? Certainly.

   Is PackageDev unmaintained to justify replacement? No.

   The reasons for it moving slowly are 
    - backward compatibility with early ST4 builds, 
    - it working well enough - despite some known edge cases
    - updating its syntaxes may likely be a comprehensive task, as various behaviors (completions, plugins) depend on current scope structures. 
    
    So moving forward without breaking too much, might require quite some considerations.

6. As we are at blaming others for being legacy and outdated ... Various syntax definitions of proposed languages use negative lookbehinds (`(?<!...)`), which triggers old/slower Onigurama regex engine. Hence modern syntax definitions don't use them anymore. Only exception is in combination with `embed...escape`, constructs, which may require them for technical reasons.

7. Backrefs also trigger Oniguma and should be avoided, whenever viable alternatives exist.

9. "myPython" contains a TOML_Jinja.sublime-syntax which is redundant with `TOML (Jinja)` provided by Jinja2 package.

   It actually seems to rely on Jinja2 package, as it includes `scope:text.jinja`.

   In case that TOML_Jinja intends to fix issues of already available `TOML (Jinja)` syntax, it is probably preferable to raise those at the original repository, instead of adding duplicate syntax.
   
   Note `TOML (Jinja)` requires at least ST4199 as it is the first build shipping a TOML syntax, which can be extended. (as an example for version dependencies when extending core syntaxes).

10. MystMarkdown seems to add a new unique syntax / markdown flavor.

    It however also copies `Text (Jinja).sublime-package` and `HTML (Jinja).sublime-syntax` from Jinja2 package. By augmenting Jinja2 package syntaxes (same name and scope), this package may cause (future) conflicts with original Jinja2 package.

    Vendoring 3rd-party syntaxes may be needed to avoid inter-package dependencies, but it should be made in ways avoiding conflicts with original packages.

    Nothing worse than looking for issues caused by packages overriding each other partly, in incompatible ways. That's especially true for syntaxes and even more when they extend each other.

----

My conclusion for the moment : Can't recommend to add those packages in their current condition.

Most packages just seem to be something private and contain too many problematic dependencies and design choices as well as apparently no considerations were made about compatibility with certain ST builds.

braver:
So, yeah, we're going with "no" right now. This approach to building packages is fine for individual devs, or even inside a team or something, but not for shipping via Package Control. You're welcome to open new PR's for individual standalone packages of course though.

FichteFoll:
I only took a look at Metasyntax (due to overlap with PackageDev), but that package declares 8 dependencies, out of which 2 are actually used (`sublime_lib` and `regex`) and a so-called `myBasis` package or library is imported but never mentioned anywhere.

I am always open to pull requests or issues for PackageDev, even if they do not get answered or reacted to immediately, but I did not see any approach in the directions indicated for the motivations of this package's README.

deathaxe:
The author's website advertises being the checker in having created 36 packages for ST using AI in some month. myBasis being one of them.

All packages were likely created with the assumption, their dependencies can be automatically resolved by Package Control.

None of the packages provides real value, despite some trivial additions, mainly extending or copying existing core or other high quality packages/syntaxes, not respecting even the basic environment's requirements or constraints.

That's most likely just AI slop, for someone trying to advertise itself.

PR #9502, opened 2026-07-29
Add Datapack Icons

bth123:

[2]: https://semver.org

*Note: Unchecked items are non-applicable to the package

My package is a file icon theme in Mincecraft inspired pixelart style

### DIFF
--- a/repository/d.json
+++ b/repository/d.json
@@ -333,6 +333,17 @@
 			]
 		},
 		{
+			"name": "Datapack Icons",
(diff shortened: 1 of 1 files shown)


braver:
Looks good! 

Can you include that `filter.py` from the package using the `.gitattributes`? It looks like it's a build step rather than an active part of the package.

bth123:
Thanks for the feedback! Didn't really notice that one script. Added it to gitattributes in the latest release.

PR #9503, opened 2026-07-30
Add RTLAssistant

jd1378:
- [ ] My package doesn't add context menu entries. * (it does, but only to tab's context menu)

[2]: https://semver.org

My package is RTLAssistant

My package is similar to [Bidirectional text support](https://packagecontrol.io/packages/Bidirectional%20text%20support) However it should still be added because that's an old package

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -2187,6 +2187,17 @@
 			]
 		},
 		{
+			"name": "RTLAssistant",
+			"details": "https://github.com/jd1378/sublime_rtl_assistant",
+			"labels": ["text manipulation", "unicode"],
+			"releases": [
+				{
+					"sublime_text": ">=4107",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "RubocopDaemon",
 			"details": "https://github.com/Lichtbit/sublime_rubocop_daemon",
 			"releases": [
--- a/repository/b.json
+++ b/repository/b.json
@@ -810,16 +810,6 @@
 			]
 		},
 		{
-			"name": "Bidirectional text support",
-			"details": "https://github.com/praveenvijayan/Sublime-Text-2-BIDI",
-			"releases": [
-				{
-					"sublime_text": "<3000",
-					"branch": "master"
-				}
-			]
-		},
-		{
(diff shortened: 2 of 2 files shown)


jd1378:
tagged [1.0.2](https://github.com/jd1378/sublime_rtl_assistant/releases/tag/1.0.2) with necessary fixes

deathaxe (approved):
Despite its github title "[Bidirectional text support](https://packagecontrol.io/packages/Bidirectional%20text%20support) is registered as ST2-only plugin and is therefore no longer relevant for the ecosystem.

We normally try to keep package index clean by replacing old unmaintained packages by:

a) using old package's name -> take over
b) renaming old package to new one via `"previous_names": ["Bidirectional text support"]`

Aspects:

- implementations are unrelated
- old package hasn't been deployed for years and is clearly marked ST2-only.
- old package's name is quite verbose

Hence, it's probably ok to keep both separate or apply strategy b), if we want to remove dead entries on package index.

---

Package itself, LGTM!

braver:
Please update your submission so that it removes the old package as Deathaxe describes!

jd1378:
it wasn't clear to me what I was expected to do, I will do now

repository/r.json
```diff
@@ -2186,6 +2186,17 @@
 				}
 			]
 		},
+		{
+			"name": "RTLAssistant",
+			"details": "https://github.com/jd1378/sublime_rtl_assistant",
+			"labels": ["text manipulation", "unicode"],
```
braver, line 2192:
```suggestion
			"labels": ["text manipulation", "unicode"],
			"previous_names": ["Bidirectional text support"],
```

What this does is replace the existing entry and update users to your new implementation. That keeps the registry fresh :) When we go forward with this it's always nice to link back to the old package from your readme, just to let users know what happened if anyone's curious. 

Thanks!

jd1378: Thank you

jd1378: This project has been written from scratch, so its only linked to that package by functionality/use-case. Will mention it in the readme

PR #9505, opened 2026-07-31
Add Semble plugin to s.json

debjan:
Added Semble plugin with details and release information.

[2]: https://semver.org

My package is plugin that integrates the [semble](https://github.com/MinishLab/semble) CLI tool for semantic code search and related-code discovery directly in editor.

### DIFF
--- a/repository/s.json
+++ b/repository/s.json
@@ -1077,6 +1077,17 @@
 			]
 		},
 		{
+			"name": "Semble",
+			"details": "https://github.com/debjan/sublime-semble",
+			"labels": ["commands", "utilities"],
+			"releases": [
+				{
+					"sublime_text": ">=4050",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Semi",
 			"details": "https://github.com/yyx990803/semi-sublime",
 			"labels": ["text manipulation"],


debjan:
1. That is already implemented: https://github.com/debjan/sublime-semble/blob/31e69ad2809314b9f1d835c3532eb958d4933528/semble.py#L229
2. It exposes the settings in https://github.com/debjan/sublime-semble/blob/main/Main.sublime-menu - let me know if adding into 'Default.sublime-commands' is necessary

braver:
> let me know if adding into 'Default.sublime-commands' is necessary

It's a strong recommendation since a lot of packages implement that pattern and makes things easier for keyboard-centric users, and it only takes you a minute to do :)

debjan:
> It's a strong recommendation since a lot of packages implement that pattern and makes things easier for keyboard-centric users, and it only takes you a minute to do :)

Done. Tag is on 1.2.0

braver:
I hadn't noticed this before, but if you want to load a JSON file from your package you have to use [load_resource](https://www.sublimetext.com/docs/api_reference.html#sublime.load_resource). Your package, as shipped by package control, will be zipped and is not directly accessible via `Path(__file__)`.

debjan:
Thanks @braver didn't know that

It's done in tag 1.3.0

PR #9508, opened 2026-08-01
Add UnicodeConverter

leibnizli:

[2]: https://semver.org

My package is a rename of the existing `ConvertChineseCharacters` package. It converts characters to Unicode escape sequences and vice versa.

My package is similar to `ConvertChineseCharacters`. However it should still be added because it is the new name for that exact same package, and I am transferring existing users over using the `previous_names` field in the JSON configuration.


braver:
I noticed the old URL already redirects to the new package, so this should end up being entirely transparent to end users. Nice one 👍🏻

PR #9509, opened 2026-08-03
Add AhuAIComplete

ahu:
- [x] I use [.gitattributes][3] to exclude files from the package: tests, demo.py, install.sh, sublime-project/workspace.

** Note:** AhuAIComplete intentionally provides default key bindings for triggering, cycling and accepting ghost-text completions. They are documented in the README and can be overridden by users via Sublime Text's key bindings preferences.

My package is an inline AI ghost-text code completion plugin for Sublime Text 4. It displays grey ghost-text suggestions as the user types, supports multiple providers (Ollama, OpenAI, OpenAI-compatible FIM), and allows cycling through candidates and partial acceptance.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -561,6 +561,19 @@
 			]
 		},
 		{
+  			"name": "AhuAIComplete",
+  			"description": "Inline AI ghost-text completion for Sublime Text (Ollama / OpenAI / FIM).",
+  			"author": "ahu",
+  			"details": "https://github.com/ahu/AIComplete",
+  			"labels": ["auto-complete", "completions", "ai", "ghost text"],
+  			"releases": [
+  				{ 
+  					"sublime_text": ">=4000", 
+  					"tags": true 
+  				}
+  			]
+		},
+		{
 			"name": "AI Bridge",
 			"author": "Jonathan Heintzeman",
 			"details": "https://github.com/MNU-497/Sublime-AI-Bridge",
--- a/repository/a.json
+++ b/repository/a.json
@@ -561,17 +561,16 @@
 			]
 		},
 		{
-  			"name": "AhuAIComplete",
-  			"description": "Inline AI ghost-text completion for Sublime Text (Ollama / OpenAI / FIM).",
-  			"author": "ahu",
-  			"details": "https://github.com/ahu/AIComplete",
-  			"labels": ["auto-complete", "completions", "ai", "ghost text"],
-  			"releases": [
-  				{ 
-  					"sublime_text": ">=4000", 
-  					"tags": true 
-  				}
-  			]
+			"name": "AhuAIComplete",
+			"description": "Inline AI ghost-text completion for Sublime Text (Ollama / OpenAI / FIM).",
+			"details": "https://github.com/ahu/AIComplete",
+			"labels": ["auto-complete", "completions", "ai", "ghost text"],
+			"releases": [
+				{
+					"sublime_text": ">=4000",
+					"tags": true
+				}
+			]
 		},
 		{
 			"name": "AI Bridge",


repository/a.json
```diff
@@ -560,6 +560,18 @@
 				}
 			]
 		},
+		{
+			"name": "AhuAIComplete",
+			"description": "Inline AI ghost-text completion for Sublime Text (Ollama / OpenAI / FIM).",
```
braver, line 565:
```suggestion
```

If you remove this the repo description (which is the same) will be used and that will allow you to change it any time you like.

braver:

Isn't this a lot like [CodeContinue](https://packages.sublimetext.com/packages/CodeContinue)?

braver:
Your comments in e.g. AhuAIComplete.sublime-settings are in Chinese. Note that Sublime Text's UI and most of its user are primarily English. All packages, unless targeting a specific audience, should have English UI and documentation.

ahu:
Thanks for the review — all three points addressed:

1. description: removed from repository/a.json, now falls back to the GitHub repo description.
2. CodeContinue overlap: spelled out in the new "How this differs" section of the README — real FIM (suffix on /completions and /api/generate), multi-candidate cycling without an `n` param, zero deps, per-view toggle.
3. Chinese comments/UI: fully translated across all files; a CJK search now returns nothing, no behaviour change (41/41 tests pass).

The CI workflow run on this fork needs a maintainer approval — could you approve when you get a chance?

repository/a.json
```diff
@@ -560,6 +560,17 @@
 				}
 			]
 		},
+		{
+			"name": "AhuAIComplete",
+			"details": "https://github.com/ahu/AIComplete",
+			"labels": ["auto-complete", "completions", "ai", "ghost text"],
```
braver:
```suggestion
			"labels": ["auto-complete", "completions", "ai"],
```

There is no other package with the "ghost text" label, and I'm not sure it's a clear explanation of what that feature is (it's the first time I've seen this completion preview described in this way).

braver:
Some feedback:

Your readme has a code block in it with "grey part is the suggestion, Tab accepts", but there is no grey text visible there (at least not in GitHub). 

You exclude the demo.py file from the package, but if if one were to install this package by simply checking it out into the Packages directory (e.g. to hack on it) it would still get loaded as a package. Better to simply not have any python you don't need in the top level dir. Similarly the install.sh seems very specific to your way of working and would be a hurdle for contributors. Sublime is built to support hacking on packages in place, and there are packages that can then help improve things like auto reloading. 

The bindings in  `Default.sublime-keymap` are not discoverable because only the OS-specific ones are linked in the menu.

Please double check that `ghost_font_size_delta` is "live" and changes to the setting are applied when the users expects.

ahu:
Thanks — all of it is addressed.

**1. The "grey part" code block in the README**
You are right: GitHub renders a fenced block as flat monospace, so the
annotation was describing something that is not visible. It is now a
text-only before/after with no colour assumption:

    typed:  return fib(n-1) + fib(
    shown:  n-2)

**2. demo.py in the package root**
Moved to examples/demo.py. Sublime imports only the .py files at the package
root, so it is no longer loaded when the repository is checked out into
Packages/. (It was already export-ignored, so it never shipped inside a
release — the problem was only the checkout case.) .gitattributes now excludes
examples/.

**3. install.sh**
Deleted, along with its export-ignore entry and the symlink caveats in the
README. The Installation section now describes the ordinary path: Package
Control for users, and for hacking, clone straight into Packages/ — Sublime
loads packages in place, so no installer and no build step is needed. It also
points at AutomaticPackageReloader for the modules under lib/, which Sublime
does not reload when they are saved.

**4. Keymap discoverability**
Correct, and it is a single-source problem now. The four generic bindings were
in Default.sublime-keymap while the platform chords were in
Default (OSX|Windows|Linux).sublime-keymap, and the menu and command palette
link the platform file — so the generic ones could not be found. The generic
file is gone; all seven bindings now live in each platform keymap, so
Preferences -> Package Settings -> AhuAIComplete -> Key Bindings shows the
complete set. Runtime behaviour is unchanged, since a given platform only ever
loads one of those files.

**5. ghost_font_size_delta**
Good catch — it was broken. It read view.settings(), but the setting lives in
AhuAIComplete.sublime-settings, which is a different Settings object, so
view.settings() never saw it and the delta was always 0. It now reads from the
package settings. It does apply live: saving the settings file fires the
existing on_change handler, which resets the engine, and the next suggestion is
drawn at the new size.

**Label**
"ghost text" is gone from the repository entry; the labels are now
["auto-complete", "completions", "ai"].

Verification: 41/41 logic tests pass, every file byte-compiles,
repository.json / the keymaps / the menu / the commands file all parse as JSON,
and a CJK search across the repository returns nothing.

braver:
Thanks!

PR #9511, opened 2026-08-07
Add PolyRun plugin details to p.json

mara1784:

[2]: https://semver.org

My package is plugin named PolyRun that enables running Python scripts in native terminals in real time.

My package is similar to Terminal. However, it should still be added because my extension saves developers time thanks to automatic saving and running of the current file.

### DIFF
--- a/repository/p.json
+++ b/repository/p.json
@@ -1989,6 +1989,11 @@
 			]
 		},
 		{
+			"name": "PolyRun",
+			"details": "https://github.com/mara1784/SublimeExtension-PolyRun",
+			"labels": ["terminal","multi-run","runner","execution"]
+		},
+		{
 			"name": "Pomodoro",
 			"details": "https://github.com/Neway6655/Sublime-Pomodoro",
 			"labels": ["timer"],


repository/p.json
```diff
@@ -1988,6 +1988,11 @@
 				}
 			]
 		},
+		{
+			"name": "PolyRun",
+			"details": "https://github.com/mara1784/SublimeExtension-PolyRun",
+			"labels": ["terminal","multi-run","runner","execution"]
```
braver, line 1994: Please specify releases, see [the example in the docs](https://docs.sublimetext.io/reference/package-control/repository.html#quick-examples).

Also please check your proposed labels against what's already out there: they're especially useful if other packages use them too. E.g. perhaps [execute](https://packages.sublimetext.com/?q=label%3A%22execute%22) instead of "execution".

braver:
Upon further review I'm unsure if this package has enough thought behind it to be useful for many users. The remark "you should change the called terminal in the code" is useless in the context of packages in general, and your code is basically only just a call to the terminal. Your package claims to work on all operating systems, but it actually only works on the version of Debian you're running.

mara1784:
I added code for windows and Macos

mara1784:
I consider as the biggest benefit of this package is that it can run more than one script on your device at the same time.

mara1784:
It is just a simple, fast way to save you time from writing python3 ...

braver:
This will need a bit more work to work as package, there are a few too many assumptions in it. Consider that Linux is more than just Debian and `x-terminal-emulator` might not be what the user wants to use. Indeed many terminal alternatives also exist on MacOS. In many systems `python` might be executable you need, not `python3`. Etc. You'll want to provide some settings here so that users can tune it to their use. 

Are you sure the save command is executed synchronously (ie. you don't have a race condition here were the executed version might be the previous version of the file)?

PR #9523, opened 2026-08-14
Add FileConverter

gerph:
A generic external-tool binary file decode/encode engine for Sublime Text. Opens files matching a configured pattern by streaming them through an external command into a new view, and can encode edited text back through a reverse command for reversible formats. Ships handlers for several RISC OS binary filetypes plus generic examples (hex dump via xxd, disassembly via objdump, media metadata via exiftool).

*We do have context menu entries* Because the window that we show isn't a real file (and we don't want to allow the user to save back) it doesn't have 'Find in sidebar', 'Open Finder window' entries (they're greyed). So we add context options - only when you're in our windows - which allow you to do those operations. We also add an explicit item to Re-encode the file to save it back, through a tool. Since Cmd-s (ctrl-s) will only offer to save the text content and you don't usually want to blindly re-write binary files. I believe this is a suitable and valid use, but if you've got better ideas about how to handle it, that'd be cool :-)

[2]: https://semver.org

My package is a generic binary presentation tool.
It can take a binary file, matched by filename, pass it through a configurable tool and show the user the output.
It can also (optionally) allow that file to be processed back to save a new file.

The closest are the PList loading tools, but I needed more control over how the files were handled, and to allow them to be saved back. And might as well make it generic so that other filetypes can be handled by people - object files being objdump'd, for example.

### DIFF
--- a/repository/f.json
+++ b/repository/f.json
@@ -624,6 +624,17 @@
 			]
 		},
 		{
+			"name": "FileConverter",
+			"details": "https://github.com/gerph/sublimetext-fileconverter",
+			"labels": ["utilities"],
+			"releases": [
+				{
+					"sublime_text": ">=3092",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "FileDialog",
 			"details": "https://github.com/shagabutdinov/sublime-file-dialog",
 			"labels": ["sublime-enhanced", "file navigation", "utilities"],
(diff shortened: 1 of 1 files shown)


repository/f.json
```diff
@@ -623,6 +623,18 @@
 				}
 			]
 		},
+		{
+			"name": "FileConverter",
+			"details": "https://github.com/gerph/sublimetext-fileconverter",
+			"labels": ["utilities"],
```
braver: Well one small thing, perhaps add the [binary](https://packages.sublimetext.io/?q=label%3A%22binary%22) label since it's used on (semi-) related packages?


```suggestion
			"labels": ["utilities", "binary"],
```

PR #9527, opened 2026-08-18
Add Limitcode

Jawuilp:

My package is an AI pair programming agent for Sublime Text. The agent only sees and touches the files you have open in the editor — no hidden filesystem scans, no shell access, no autopilot — with a side-by-side streaming chat, permission prompts before edits, and project-scoped conversation history.

** Regarding key bindings: the only bindings shipped (Enter, Shift+Enter, Escape, Up/Down) are scoped with a `setting.limitcode_chat_view` context, so they only act inside the package's own chat view and never override global keys. They are also listed in Preferences → Package Settings → Limitcode → Key Bindings.

[2]: https://semver.org

### DIFF
--- a/repository/l.json
+++ b/repository/l.json
@@ -1083,6 +1083,17 @@
 			]
 		},
 		{
+			"name": "Limitcode",
+			"details": "https://github.com/Jawuilp/Limitcode",
+			"labels": ["ai", "coding assistant"],
+			"releases": [
+				{
+					"sublime_text": "*",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "LimitTabs",
 			"details": "https://github.com/maximilize/LimitTabs",
 			"releases": [
--- a/repository/l.json
+++ b/repository/l.json
@@ -1085,7 +1085,7 @@
 		{
 			"name": "Limitcode",
 			"details": "https://github.com/Jawuilp/Limitcode",
-			"labels": ["ai", "llm", "coding assistant"],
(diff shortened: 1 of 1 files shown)


repository/l.json
```diff
@@ -1082,6 +1082,17 @@
 				}
 			]
 		},
+		{
+			"name": "Limitcode",
+			"details": "https://github.com/Jawuilp/Limitcode",
+			"labels": ["ai", "llm", "agent", "coding assistant"],
```
braver: We don't have any packages with "coding assistent"... Sublime Text mainly being about "coding" anyway that prefix is perhaps superfluous. We do have packages with "assistant" and "chat", both might apply here too.

repository/l.json
```diff
@@ -1082,6 +1082,17 @@
...
+			"releases": [
+				{
+					"sublime_text": "*",
```
braver: Saying this will run on ST3 etc might be a bit optimistic, unless of course you've tested for that. I'd personally opt for a more recent number just to avoid issues and potential headaches. It's probably fine, how many people are going to use new AI-based tools in a 5 year old editor and expect great results? But YMMV :)

repository/l.json
Jawuilp: Updated the labels to use assistant and chat instead of coding assistant. Thanks!

repository/l.json
Jawuilp: Updated the minimum Sublime Text version to >=4050, since that’s the first build with Python 3.8 plugin support.

PR #9531, opened 2026-08-23
Add GithubPullRequest

ViBiOh:
Add a reference to my plugin for reviewing a GitHub pull request from within Sublime Text.

[2]: https://semver.org

My package is ...

### DIFF
--- a/repository/g.json
+++ b/repository/g.json
@@ -743,6 +743,18 @@
 			]
 		},
 		{
+			"name": "Github Pull Request",
+			"details": "https://github.com/ViBiOh/sublime-GithubPullRequest",
+			"author": "ViBiOh",
+			"labels": ["github", "pull-request"],
+			"releases": [
+				{
+          			"sublime_text": ">=4205",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Github Search",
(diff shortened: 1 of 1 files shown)


repository/g.json
```diff
@@ -798,6 +798,17 @@
 				}
 			]
 		},
+		{
+			"name": "GitHubPullRequest",
+			"details": "https://github.com/ViBiOh/sublime-GitHubPullRequest",
+			"labels": ["github", "pull-request"],
+			"releases": [
+				{
+					"sublime_text": ">=4205",
```
kaste, line 807: Is that really the case?  It worked on my much older build, but I surely didn't touch all of it.

ViBiOh: Oops, I fixed it on the plugin's repository but not here, it's fixed now.

ViBiOh:
It seems the review is good 😃 considering the warnings as false positive like you mentioned before. Do you need any input on my side to move forward?

braver:
Cool, I think we're ready then. @kaste want to give a final LGTM?

kaste:
LGTM @braver

braver:
Thanks all!

PR #9532, opened 2026-08-24
Add Auto Maximize

mrflamingchad67:
  - N/A — the repository contains only the plugin file, README.md and LICENSE.

[2]: https://semver.org

My package is **Auto Maximize**, a small Windows-only plugin that makes Sublime Text start with its window(s) maximized instead of restored to their previous size, and maximizes any window opened later. It uses the native `ShowWindow` call on `Window.hwnd()` (hence the `>=4050` requirement and windows-only platform), so it does not interfere with fullscreen mode at all.

There are no packages like it in Package Control. The closest existing ones (e.g. AutomaticFullscreen) toggle real fullscreen mode; this plugin deliberately maximizes the OS window instead, keeping title bar, tabs and taskbar visible.

### DIFF
--- a/repository/a.json
+++ b/repository/a.json
@@ -2303,6 +2303,19 @@
 			]
 		},
 		{
+			"name": "Auto Maximize",
+			"details": "https://github.com/mrflamingchad67/sublime-auto-maximize",
+			"author": "mrflamingchad67",
+			"labels": ["utilities", "window"],
+			"releases": [
+				{
+					"sublime_text": ">=4050",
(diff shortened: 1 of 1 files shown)


kaste:
LGTM @braver 

... I did not test as I don't want the behavior to mess with all my sessions but the one-liner ctypes thing is correct AFAIK.

braver:
Indeed, the package looks simple enough :) 

@mrflamingchad67 please remove the author key from your entry, it's not needed.

PR #9533, opened 2026-08-25
Add st4-py38- tag prefix for Requester and TreeSitter plugins, for ST builds 4107 - 4200

kylebebak:
- Add st4-py38- tag prefix for Requester and TreeSitter plugins, for ST builds 4107 - 4200
- Past build 4200, there is no longer a python 3.8 plugin host, which broke both of these plugins
- Requester has already been updated for the Python 3.14 plugin host, and TreeSitter is almost ready
  - https://github.com/kylebebak/Requester/releases/tag/st4-py38-2.39.6

The following checklist may not be super relevant, because these aren't new packages. They're existing packages that are being updated for build 4205, which replaced the python 3.8 plugin host with 3.14.

[2]: https://semver.org

---

Also, it's worth noting that these errors are all pre-existing "issues" in pretty much every version of Requester that's been shipped

```
[84](https://github.com/sublimehq/package_control_channel/actions/runs/32907533352/job/97994757021?pr=9533#step:2:303)
  Error: Do not import root-level plugin module '.add_path'. Sublime Text loads root-level Python files as independent plugins; move shared code into a subpackage.
      File: core/__init__.py
      Line: 13, Column: 1
  Error: The binding ['ctrl+r'] has no context. Packages should only ship key bindings that use a specific context such as 'selector', 'setting.*', or a custom context key. If the binding defines a main entry-point to your package, move it to an example keymap instead so users can decide on their own.
      File: Default (OSX).sublime-keymap
  Error: The binding ['ctrl+alt+r'] has no context. Packages should only ship key bindings that use a specific context such as 'selector', 'setting.*', or a custom context key. If the binding defines a main entry-point to your package, move it to an example keymap instead so users can decide on their own.
      File: Default.sublime-keymap
```

### DIFF
--- a/repository/r.json
+++ b/repository/r.json
@@ -1388,7 +1388,11 @@
 					"tags": "st3-"
 				},
 				{
-					"sublime_text": ">=4107",
+					"sublime_text": "4107 - 4200",
+					"tags": "st4-py38-"
+				},
+				{
+					"sublime_text": ">=4205",
 					"tags": true
 				}
 			]
--- a/repository/t.json
+++ b/repository/t.json
@@ -3474,7 +3474,11 @@
 			"author": "kylebebak",
 			"releases": [
 				{
-					"sublime_text": ">=4107",
+					"sublime_text": "4107 - 4200",
+					"tags": "st4-py38-"
+				},
+				{
+					"sublime_text": ">=4205",
 					"tags": true
 				}
 			]


repository/r.json
```diff
@@ -1388,7 +1388,11 @@
 					"tags": "st3-"
 				},
 				{
-					"sublime_text": ">=4107",
+					"sublime_text": "4107 - 4200",
+					"tags": "st4-py38-"
+				},
+				{
+					"sublime_text": ">=4205",
 					"tags": true
```
FichteFoll, line 1396: Note that tagging 3.0.0 already pushed all users of this package to the latest 3.0.0 version that is supposedly incompatible with the Python 3.8 host with no way to automatically downgrade to 2.39.6. 

(I say that but it is unlikely for the update to have been picked up already within the last hour. Fortunately for you, I happened to have a spare minute and spotted this early.)

FichteFoll (approved):

repository/r.json
kylebebak: Ahh, you're right, I didn't catch that! Thank you for getting this merged so quickly

PR #9543, opened 2026-08-30
Add VisuBezier

mrcgrtz:

[2]: https://semver.org

My package previews CSS easing functions on hover. It is a fork of [VisuBezier](https://github.com/chriskirknielsen/visubezier) on VS Code.

### DIFF
--- a/repository/v.json
+++ b/repository/v.json
@@ -667,6 +667,17 @@
 			]
 		},
 		{
+			"name": "VisuBezier",
+			"details": "https://github.com/mrcgrtz/visubezier",
+			"labels": ["css", "preview", "hover", "utilities"],
+			"releases": [
+				{
+					"sublime_text": ">=4107",
+					"tags": true
+				}
+			]
+		},
+		{
 			"name": "Vlt",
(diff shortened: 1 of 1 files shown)


kaste:
LGTM! @braver 

@mrcgrtz It stutters a bit, did you try to increase the fps?  what happened?

In the hot loop you have:

```
view.update_popup(_popup_html(expression, reference, frames[index]))
```

but I think the content is cacheable and used in a loop, so (a) you could just pre-render or (b) (eg lru_cache) render each frame lazy once and then read from cache.

`not view.is_valid()` in that loop is likely not necessary as the next `is_popup_visible` will also answer that.  (That usually doesn't make a difference, so it's a nit.)

braver:
Might be really hard to hit 60fps with this approach. Impressive though! I think I'd personally already be happy to just see the curve (graph) rendered, without the bouncing blocks.

mrcgrtz:
Thanks @kaste & @braver, I have created https://github.com/mrcgrtz/visubezier/issues/3 and https://github.com/mrcgrtz/visubezier/issues/4 based on your feedback.

PR #9551, opened 2026-09-07
Add RTL Preview

itszux:

[2]: https://semver.org

My package is RTL Preview

My package is similar to RTL Mirror. However, it should still be added because ...
It solves the lack of Bidirectional (BiDi) text layout in a new way by providing a preview line generated with (gdi and DirectWrite/Direct2D APIs) without altering the original text like other RTL Sublime packages.


itszux:
I have fixed all the issues requested. I also pushed release 1.0.2 in the plugin repository, which includes the requested fixes for the paths, subpackages, and license.

PR #9553, opened 2026-09-08
Add SubCalc

vonbehr:
[SubCalc](https://github.com/vonbehr/SubCalc) is a Sublime Text 4 package for live, per-line calculation in plain-text notes -- write a `.calc` file mixing prose and math, and every line that evaluates to a number shows its result inline as you type, in the spirit of Soulver / NotesCalculator.

[2]: https://semver.org

My package is a live, per-line calculator for plain-text `.calc` notes.

My package is similar to [Soulver](https://soulver.app) and NotesCalculator, however it's native to Sublime Text: everything runs locally in the buffer as you type, with no separate app or window, and results are computed with a small custom parser/evaluator rather than shelling out.


braver:
Please follow and include the PR template completely.

vonbehr:
Updated the PR description to follow the submission template point by point. Also pushed two fixes to the repo: added `.gitattributes` `export-ignore` rules to keep tests/CI/tooling out of the shipped package, and added a `Preferences → Package Settings → SubCalc → Settings` menu entry (via `edit_settings`) so settings open in split view from both the menu and command palette, per the checklist. The one unchecked box (syntax + color scheme) is explained inline — the scheme is optional and never force-applied.

braver:
> sensible default for the handful of custom scopes

It's technically almost impossible for a given set of colors to be "sensible defaults" combined against any user's preferred color scheme. I don't see how a user is supposed to use this? It might be more useful to tell users how to extend any color scheme they already use, to support the specific scopes your package introduces. See https://www.sublimetext.com/docs/color_schemes.html#customization

vonbehr:
Fair point, done: removed `Calc.sublime-color-scheme` entirely and added a "Customizing colors" section to the README documenting the `.calc`-suffixed scopes, with a copy-paste snippet for adding overrides via Preferences > Customize Color Scheme. Checked that box in the description now.

braver:
Alright, we're ready to merge this as soon as you tag the most recent version!

repository/s.json
```diff
@@ -4393,6 +4393,16 @@
 				}
 			]
 		},
+		{
+			"name": "SubCalc",
+			"details": "https://github.com/vonbehr/SubCalc",
+			"releases": [
```
braver, line 4400: Although... "utilities" and "calculator" might be applicable labels!

PR #9561, opened 2026-09-14
Update Racket package in preparation for future ST 4 releases

17451k:

[2]: https://semver.org

[My package](https://github.com/17451k/racket-enhanced) is an improved syntax definition for the Racket programming language. It is similar to [the existing one](https://github.com/follesoe/sublime-racket) (last updated 7 years ago, targets ST 2), but the differences in quality of the highlighting are huge and [documented in the readme](https://github.com/17451k/racket-enhanced#why-another-racket-package). I would also like to keep improving the package in the future and add new features beside syntax definition.

Question: is it acceptable to have two plugins for the same language? An alternative, as I see it, is to try to contact the author of the existing plugin and to take over its maintenance.


braver:
Indeed, that should work 👍🏻

braver:
> No semantic version tags found

It's not great at checking prefixed tags. All the same, it would be better if a tag for st4+ exist before we merge here.

17451k:
I've added new changes and semver tag (2.0.0), should work now. LICENSE file also exists in the new version of the plugin

